PR #5, opened 2025-12-03
Fix build errors that were preventing the dev scripts from running

cdbartholomew:
Replace Array.from(params.entries()) with forEach-based iteration for better compatibility across different JavaScript environments. Some environments don't properly support the entries() iterator.


nicoloboschi:
Fixed with another commit

PR #17, opened 2025-12-05
Added hindsight_liteLLM implementation

DK09876:


hindsight-integrations/litellm/hindsight_litellm/wrappers.py
```diff
@@ -0,0 +1,708 @@
...
+    weight: float
+    metadata: Optional[Dict[str, Any]] = None
+
+    def __str__(self) -> str:
+        return self.text
+
+
+def recall(
+    query: str,
+    limit: int = 10,
```
nicoloboschi: hindsight doesn't have the limit/top-k parameter by design. can we get rid of this?

hindsight-integrations/litellm/tests/test_integration.py
```diff
@@ -0,0 +1,471 @@
+"""Integration tests for hindsight-litellm."""
+
+import pytest
+from unittest.mock import Mock, patch, MagicMock
+from typing import List, Dict, Any
+
+from hindsight_litellm import (
```
nicoloboschi: can you make this test to run in CI ?

nicoloboschi (changes requested):

PR #39, opened 2025-12-17
Fix: bank selector race condition when switching banks (#38)

### LINKED ISSUE #38: Bug: Control Plane - switching memory banks via dropdown fails to change bank (race condition)
**Description**

Using the Control Plane UI, I observed that changing the selected memory bank via the dropdown selector would not always result in the memory bank being changed: I'd see memories returned from the previous bank still. Refreshing the page would fix it.

**Steps to Reproduce**

1. Have multiple memory banks
2. Open the Control Plane
3. Choose a memory bank from the dropdown (i.e. results in http://localhost:9999/banks/bank-1?view=data)
4. Change to a different memory bank (i.e. results in http://localhost:9999/banks/bank-2?view=data)
5. You may need to repeat this, since it wasn't entirely deterministic

**Expected Behaviour**
The URL, dropdown, and displayed data should all reflect the newly selected bank.

**Actual Behaviour**
The URL might show /banks/**bank-2** but the data displayed is from the previously chosen **bank-1**.

**Root Cause**
The useEffect call here: https://github.com/vectorize-io/hindsight/blob/c0a0f447b72d44b0a4a4bca9e06dbdb4ec524684/hindsight-control-plane/src/app/banks/%5BbankId%5D/page.tsx#L28-L33

My understanding is this attempts to sync context from useParams(), which updates _asynchronously_. During the window between setCurrentBank() and navigation completion, useParams() still returns the old bank ID, causing the effect to reset context to the old value.

This conflicts with the existing pathname-based sync in bank-context.tsx.

**Fix**
Removing of the useEffect call referenced above. Setting of the bank in context is handled elsewhere already:

https://github.com/vectorize-io/hindsight/blob/c0a0f447b72d44b0a4a4bca9e06dbdb4ec524684/hindsight-control-plane/src/lib/bank-context.tsx#L33-L38

https://github.com/vectorize-io/hindsight/blob/c0a0f447b72d44b0a4a4bca9e06dbdb4ec524684/hindsight-control-plane/src/components/bank-selector.tsx#L70

https://github.com/vectorize-io/hindsight/blob/c0a0f447b72d44b0a4a4bca9e06dbdb4ec524684/hindsight-control-plane/src/components/bank-selector.tsx#L160

**Testing**
I have tested this change manually and it now reliably loads the correct bank when both navigating directly to a bank and when changing a memory bank from the dropdown selector.

**Pull Request**
PR to follow shortly.

wsimmonds:
Fixes #38

Removes a redundant `useEffect` call that resulted in a race condition.

### DIFF
--- a/hindsight-control-plane/src/app/banks/[bankId]/page.tsx
+++ b/hindsight-control-plane/src/app/banks/[bankId]/page.tsx
@@ -1,7 +1,6 @@
 "use client";
 
 import { useParams, useRouter, useSearchParams } from "next/navigation";
-import { useEffect } from "react";
 import { BankSelector } from "@/components/bank-selector";
 import { Sidebar } from "@/components/sidebar";
 import { DataView } from "@/components/data-view";
@@ -10,7 +9,6 @@ import { EntitiesView } from "@/components/entities-view";
 import { ThinkView } from "@/components/think-view";
 import { SearchDebugView } from "@/components/search-debug-view";
 import { BankProfileView } from "@/components/bank-profile-view";
-import { useBank } from "@/lib/bank-context";
 
 type NavItem = "recall" | "reflect" | "data" | "documents" | "entities" | "profile";
 type DataSubTab = "world" | "experience" | "opinion";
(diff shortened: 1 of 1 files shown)


nicoloboschi (approved):
LGTM, thanks!

PR #45, opened 2025-12-17
fix: add procps to Docker image and smoke test to release workflow

cdbartholomew:
## Summary

- Add `procps` package to Docker images (api-only and standalone) to provide the `kill` command that pg0 needs
- Add smoke testing to the release workflow to prevent releasing broken images

## Problem

The Docker image has been broken since release 0.1.6 when the fallback URI code was removed to support dynamic ports. The pg0 library uses `kill -0 <pid>` to check if PostgreSQL is running, but the `python:3.11-slim` base image doesn't include the `kill` command. Without it, pg0 couldn't detect process status and returned `None` for the database URI, causing startup failures.

## Solution

1. **Dockerfile**: Add `procps` package which provides the `kill` command
2. **Release workflow**: Add smoke testing before pushing images:
   - Build image locally (single platform, no push)
   - Run container and wait for `/health` endpoint (up to 120s)
   - Only push multi-platform release images if smoke test passes
   - Each image (api-only, cp-only, standalone) tested independently

If smoke test fails, nothing is pushed to GHCR.

## Test plan

- [x] Verified `ghcr.io/vectorize-io/hindsight:latest` fails (pg0 returns None)
- [x] Verified `ghcr.io/vectorize-io/hindsight:0.1.5` passes (before the regression)
- [x] Built fixed image locally and verified it starts successfully
- [x] Tested api-only target specifically (what release uses)
- [x] Confirmed `/health` endpoint returns `{"status":"healthy","database":"connected"}`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -235,50 +235,7 @@ jobs:
     - name: Smoke test - verify container starts
       env:
         GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
-      run: |
-        echo "🧪 Running smoke test for ${{ matrix.image_name }}..."
-        IMAGE="${{ matrix.image_name }}:test"
-
-        # Determine health endpoint and env vars based on image type
-        if [ "${{ matrix.target }}" = "cp-only" ]; then
-          HEALTH_PORT=9999
-          # Control plane doesn't need LLM credentials
-          docker run -d --name smoke-test \
-            -p ${HEALTH_PORT}:${HEALTH_PORT} \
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
Good catch

It would be better to run the docker tests on the pull request validation rather in the release process

cdbartholomew:
@nicoloboschi We can add a second test, but I would like to keep the basic validation in the release process. This prevents the case where the failing test get ignored so we accidentally release a bad Docker image (again).

Issue #55, opened 2025-12-18
Bug: Control Plane Timeline view only shows ~1% of opinions (missing facts without `occurred_start`)


wsimmonds:
## Description

The timeline view in the control plane only displays memories that have an `occurred_start` date, which excludes the vast majority of opinions and conversation-type facts. I discovered this during usage by observing only 1 out of nearly 200 opinions appearing on the timeline.

## Root Cause

Thanks to @nicoloboschi for the feedback in understanding this problem's root cause better.

The `TimelineView` component in `data-view.tsx` filters on `occurred_start`:

```typescript
const withDates = filteredRows
  .filter((row: any) => row.occurred_start)  // <-- Effectively excludes opinions
```

Opinions are typically classified as `fact_kind == "conversation"` during fact extraction, and only `fact_kind == "event"` facts receive an `occurred_start` date. All facts do have a `mentioned_at` timestamp (when the fact was learned/formed), but this wasn't being used as a fallback.

## Expected Behaviour

- All facts with date information should appear on the timeline
- For events: display when the event happened (`occurred_start`)
- For opinions/conversations: display when the fact was formed (`mentioned_at`)
- Timeline should be sorted in descending order (newest first), consistent with the API query ordering

## Actual Behaviour

- Only facts with `occurred_start` appear on the timeline
- Opinions and conversation-type facts are hidden
- Timeline was sorted ascending (oldest first) - which is inconsistent with the actual api endpoint's query: https://github.com/vectorize-io/hindsight/blob/6d820ef91bbf27f890754a71e758d99600b56b90/hindsight-api/hindsight_api/engine/memory_engine.py#L2082-L2091

## Steps to Reproduce

1. Create a memory bank with several opinions (e.g., "I think X is important", "I prefer Y")
2. Navigate to Opinions in the control plane
3. Switch to Timeline view
4. Observe that most/all opinions are missing from the timeline

## Testing

Manually tested with a real memory bank containing ~200 opinions. Verified that opinions which were previously hidden now appear on the timeline and are correctly ordered by their effective date (descending).

## Other Observations
I'll raise another couple of issues following this based upon my observations:

### Dates - Inconsistency
[...]

PR #56, opened 2025-12-18
Control Plane Timeline: Switch to descending date and fallback to mentioned_at for non events

wsimmonds:
Addresses #55

**Introduce effective_date**
Instead of using just occurred_start for display/ordering I've introduced an effective_date property to object in the withDates array: This will be occurred_start if present but otherwise will fallback to mentioned_at (or null if not set).

**Sort order**
It now sorts by effective_date in _descending_ order, which seems to be consistent with the query on the api:

https://github.com/vectorize-io/hindsight/blob/6d820ef91bbf27f890754a71e758d99600b56b90/hindsight-api/hindsight_api/engine/memory_engine.py#L2082-L2091

**Testing**
I've tested this manually with a memory bank of mine, verifying that opinions which previously were hidden now show and are ordered correctly.

### DIFF
--- a/hindsight-control-plane/src/components/data-view.tsx
+++ b/hindsight-control-plane/src/components/data-view.tsx
@@ -715,7 +715,7 @@ function TimelineView({ data, filteredRows }: { data: any; filteredRows: any[] }
   const [currentIndex, setCurrentIndex] = useState(0);
   const timelineRef = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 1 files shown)


nicoloboschi (changes requested):
I think it would be better to a UI control (a switch) that let the user see mentioned vs occurred

I think using a effective date might be confusing

PR #81, opened 2025-12-31
fix(mcp): Chain MCP lifespan with FastAPI app lifespan

bjornslib:
## Summary
Fixes the MCP server lifespan not being properly chained with the FastAPI app's lifespan, causing the MCP server to not start/stop correctly when mounted as a sub-application.

## Problem
When the MCP server is mounted as a sub-application, its lifespan context was not being invoked, causing initialization issues.

## Solution
- Create MCP app before FastAPI app to access its lifespan
- Chain MCP lifespan context with FastAPI's lifespan context using `asynccontextmanager`
- Ensures MCP server lifecycle is properly managed

## Testing
Verified with Claude Code MCP integration - the server now starts and stops correctly.

### DIFF
--- a/hindsight-api/hindsight_api/api/__init__.py
+++ b/hindsight-api/hindsight_api/api/__init__.py
@@ -5,6 +5,7 @@ Provides both HTTP REST API and MCP (Model Context Protocol) server.
 """
 
 import logging
+from contextlib import asynccontextmanager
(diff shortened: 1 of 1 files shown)


bjornslib:
Needed to make this change for MCP in CC to work.

PR #82, opened 2025-12-31
feat(mcp): Add multi-bank access and new MCP tools

bjornslib:
## Summary
Enables orchestrator agents to access multiple memory banks from a single MCP connection, with new tools for bank management.

## New MCP Tools

| Tool | Purpose |
|------|---------|
| `reflect(query, context?, budget?, bank_id?)` | Thoughtful analysis using bank's personality and memories |
| `list_banks()` | Discover all available memory banks |
| `create_bank(bank_id, name?, background?)` | Create new banks programmatically |

## Multi-Bank Access

Added optional `bank_id` parameter to existing tools:
- `retain(content, context, bank_id?)` - Store in specific bank
- `recall(query, max_results?, bank_id?)` - Search specific bank
- `reflect(query, context?, budget?, bank_id?)` - Reflect using specific bank

When `bank_id` is not specified, the session's default bank is used.

## Claude Code Compatibility
- Enabled `stateless_http=True` for proper Claude Code integration
- Responses now include `bank_id` for transparency

## Documentation
- Added `docker-compose.example.yml` with env var substitution (no secrets)
- Added `HINDSIGHT-DOCKER.md` setup guide with volume persistence warnings
- Updated `.gitignore` to exclude local docker-compose.yml

## Use Case

Orchestrator agents managing multiple Claude Code instances can now:
- Maintain a private meta-orchestration bank for self-reflection
- Access shared project knowledge banks
- Query across banks for cross-context insights

Example:
```python
# Store in specific bank
retain(content="Learned pattern", bank_id="orchestrator-memory")

# Reflect using orchestrator's personality
reflect(query="What patterns work?", bank_id="orchestrator-memory")
```


docker-compose.example.yml
```diff
@@ -0,0 +1,41 @@
+# Hindsight Docker Compose Configuration
```
nicoloboschi: please remove this file, we don't want to keep a docker compose for a simple installation. everything should be in the doc and if something missing you can add there

hindsight-api/hindsight_api/api/mcp.py
```diff
@@ -92,15 +97,16 @@ async def recall(query: str, max_results: int = 10) -> str:
         Args:
             query: Natural language search query (e.g., "user's food preferences", "what projects is user working on")
             max_results: Maximum number of results to return (default: 10)
+            bank_id: Optional bank to search in (defaults to session bank). Use for cross-bank operations.
```
nicoloboschi: I don't think this is clear enough for the agent to understand what do to with it. have you tried it out already? 

initalliy this was the only implementation but I had issues letting Claude to understand the bank_id param and what to use exactly. and when it gets confused by a tool usually it just ignores it

PR #83, opened 2025-12-31
feat(config): Add configurable observation thresholds

bjornslib:
## Summary
Allows tuning of entity observation generation via environment variables, enabling performance/quality trade-offs.

## New Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `HINDSIGHT_API_OBSERVATION_MIN_FACTS` | 5 | Minimum facts required to generate entity observations |
| `HINDSIGHT_API_OBSERVATION_TOP_ENTITIES` | 5 | Maximum entities to process per retain batch |

## Changes
- Added threshold configuration to `HindsightConfig`
- Updated `memory_engine.py` to use config values instead of hardcoded constants
- Updated `observation_regeneration.py` to use config values

## Use Case

**Lower thresholds** (e.g., `MIN_FACTS=3`):
- Generates more observations
- Better recall coverage
- Higher LLM cost

**Higher thresholds** (e.g., `MIN_FACTS=10`):
- More selective observation generation
- Lower LLM cost
- May miss patterns with fewer mentions

Example:
```bash
# Generate more observations
docker run -e HINDSIGHT_API_OBSERVATION_MIN_FACTS=3 \
           -e HINDSIGHT_API_OBSERVATION_TOP_ENTITIES=10 \
           hindsight-standalone
```

### DIFF
--- a/hindsight-api/hindsight_api/config.py
+++ b/hindsight-api/hindsight_api/config.py
@@ -33,6 +33,10 @@ ENV_GRAPH_RETRIEVER = "HINDSIGHT_API_GRAPH_RETRIEVER"
 ENV_MCP_LOCAL_BANK_ID = "HINDSIGHT_API_MCP_LOCAL_BANK_ID"
 ENV_MCP_INSTRUCTIONS = "HINDSIGHT_API_MCP_INSTRUCTIONS"
 
+# Observation thresholds
+ENV_OBSERVATION_MIN_FACTS = "HINDSIGHT_API_OBSERVATION_MIN_FACTS"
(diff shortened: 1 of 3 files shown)


bjornslib:
Don't think this is a must-have, and you may reject this PR, considering that this is a parameter that will affect performance. But after looking at the number of observations in my memory banks, which were created from Serena memories collected over many months of Claude Code usage, the hardcoded values of 5 for min facts and top entities weren't at the right level and I changed and re-ran it with the updated parameters and generated more observations.

PR #88, opened 2026-01-02
feat: Add local LLM improvements for reasoning models and Docker startup

csfet9:
## Summary

- **Reasoning model support**: Strip thinking tags from local LLM responses, enabling Qwen3, DeepSeek, and other reasoning models
- **Docker retry-start script**: Wait for dependencies (LLM Studio, database) before starting Hindsight

### Reasoning Model Support (`llm_wrapper.py`)
Strips thinking tags from local LLM responses:
- `<think>...</think>`
- `<thinking>...</thinking>`
- `<reasoning>...</reasoning>`
- `|startthink|...|endthink|`

This enables reasoning models like Qwen3 to work with Hindsight's JSON extraction pipeline. Non-breaking change - only affects responses that contain these tags.

### Docker Retry Start Script (`retry-start.sh`)
New startup script that waits for dependencies:
- Checks LLM Studio at `/v1/models` endpoint
- Checks database connectivity (skipped for embedded pg0)
- Configurable retries via `HINDSIGHT_RETRY_MAX` (default: infinite)
- Configurable interval via `HINDSIGHT_RETRY_INTERVAL` (default: 10s)

Prevents startup failures when LLM Studio takes time to load models.

## Test plan

- [x] Tested reasoning model support with Qwen3 8B on LM Studio
- [x] Verified thinking tags are stripped correctly
- [x] Tested retry-start with LLM Studio dependency
- [x] Verified embedded pg0 detection works
- [x] Health endpoint returns healthy after startup

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -84,6 +84,27 @@ PostgreSQL with pgvector. Schema managed via Alembic migrations in `hindsight-ap
 
 Key tables: `banks`, `memory_units`, `documents`, `entities`, `entity_links`
 
+### Database Backups (IMPORTANT)
+**Before any operation that may affect the database, run a backup:**
+```bash
+docker exec hindsight /backups/backup.sh
+```
+
+Operations requiring backup:
+- Running database migrations
+- Modifying Alembic migration files
+- Rebuilding Docker images
+- Resetting or recreating containers
+- Any schema changes
+- Bulk data operations
+
+Backups are stored in `~/hindsight-backups/` on the host.
+
+To restore:
+```bash
+docker exec -it hindsight /backups/restore.sh <backup-file.sql.gz>
+```
+
 ## Key Conventions
 
 ### Memory Banks
--- a/docker/standalone/Dockerfile
+++ b/docker/standalone/Dockerfile
@@ -157,9 +157,7 @@ USER hindsight
 # Set PATH for hindsight user
 ENV PATH="/app/api/.venv/bin:${PATH}"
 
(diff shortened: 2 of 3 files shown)


hindsight-api/hindsight_api/engine/llm_wrapper.py
```diff
@@ -305,6 +306,18 @@ async def call(
 
                         content = response.choices[0].message.content
 
+                        # Strip reasoning model thinking tags (various formats)
+                        # Supports: <think>, <thinking>, <reasoning>, |startthink|/|endthink|
+                        if content:
+                            original_len = len(content)
```
nicoloboschi: can we use the flag in the llm call to not include thinking tokens in the output? this way the user will not pay for those output tokens and we don't have to leverage on this heuristic and problematic algorithm

csfet9: I looked into this - unfortunately LM Studio doesn't have an API parameter to exclude thinking tokens (unlike Ollama which supports think: false). The OpenAI-compatible endpoint has no universal flag for this.

  I've made the stripping opt-in via HINDSIGHT_API_LLM_STRIP_THINKING=true so it's explicit rather than automatic. Users running reasoning models (Qwen3, DeepSeek) on LM Studio can enable it when needed.

  Re: paying for output tokens - since LM Studio runs locally, there's no per-token cost. The tradeoff is just the extra processing time for the thinking tokens, which is minimal compared to the inference time.

PR #91, opened 2026-01-02
feat: Add user-provided entities support to retain endpoint

phamgialinhlx:
## Summary
- Add support for user-provided entities in the retain endpoint that get merged with auto-extracted entities
- Entities can be passed with each memory item as `[{"text": "...", "type": "..."}]`
- User entities are deduplicated (case-insensitive) against LLM-extracted entities before merging

## Changes
- **API**: Added `EntityInput` model and `entities` field to `MemoryItem`
- **Engine**: Updated entity processing to merge user entities with extracted ones
- **Clients**: Updated Python, TypeScript, and control plane clients with entity support


docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+services:
```
nicoloboschi: please remove this file, we don't want to many getting started, we already have documentation for running with docker

phamgialinhlx:
> Thanks for your contribution!
> 
> Could you elaborate on the use case of this feature? While I understand how it works, its important that is clear why and when to use it

In some cases I already know which entities are in the chunks and want to make sure they get included. For example, critical entities like key product names that I've already extracted using regex - I want to pass those directly rather than hoping the LLM picks them up

PR #95, opened 2026-01-04
feat(mcp): add async_processing parameter to retain tool

slayoffer:
## Summary

Adds `async_processing` parameter to the MCP `retain` tool, allowing non-blocking memory storage.

**Changes:**
- Added `async_processing: bool = True` parameter to `retain` tool
  - When `True` (default): queues memory for background processing, returns immediately
  - When `False`: waits for memory processing to complete before returning

This complements the existing multi-bank support added in v0.2.1.

## Why

The retain tool was blocking callers during memory processing. This can be slow for complex content. The new parameter gives callers control:
- Use `async_processing=True` for fire-and-forget storage (default, faster)
- Use `async_processing=False` when you need confirmation before proceeding

## Test plan

- [ ] Start API with MCP enabled
- [ ] Test retain with `async_processing=True` (should return immediately with operation_id)
- [ ] Test retain with `async_processing=False` (should wait for completion)

### DIFF
--- a/hindsight-api/hindsight_api/api/mcp.py
+++ b/hindsight-api/hindsight_api/api/mcp.py
@@ -8,7 +8,6 @@ from contextvars import ContextVar
 from fastmcp import FastMCP
 
 from hindsight_api import MemoryEngine
-from hindsight_api.api.http import BankListItem, BankListResponse, BankProfileResponse, DispositionTraits
 from hindsight_api.engine.response_models import VALID_RECALL_FACT_TYPES
 from hindsight_api.models import RequestContext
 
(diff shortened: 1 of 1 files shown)


hindsight-api/hindsight_api/api/mcp.py
```diff
@@ -246,6 +261,48 @@ async def create_bank(bank_id: str, name: str | None = None, background: str | N
             logger.error(f"Error creating bank: {e}", exc_info=True)
             return json.dumps({"error": str(e)})
 
+    @mcp.tool()
+    async def list_memories(
```
nicoloboschi: @slayoffer I think this tool is misleading for the agent, it should use recall, the list method is mostly for debugging and exploration, not for real usage

nicoloboschi (changes requested):

hindsight-api/hindsight_api/api/mcp.py
slayoffer: @nicoloboschi hey. thanks for that insight. removed it

PR #170, opened 2026-01-18
Add structured JSON logging support

cdbartholomew:
## Summary

- Add `HINDSIGHT_API_LOG_FORMAT` environment variable to configure log output format
- Options: `text` (default, human-readable) and `json` (structured)
- JSON format includes a `severity` field that logging systems can parse for proper log level categorization
- Logs now write to stdout instead of stderr so log levels are correctly interpreted by external systems

## Test plan

- [x] Verified JSON format outputs valid JSON with correct severity levels
- [x] Verified text format still works as expected
- [x] Ran existing tests - all pass
- [ ] Test in local dev environment with `HINDSIGHT_API_LOG_FORMAT=json`

### DIFF
--- a/hindsight-api/hindsight_api/config.py
+++ b/hindsight-api/hindsight_api/config.py
@@ -201,14 +201,13 @@ Use this tool PROACTIVELY to:
 EMBEDDING_DIMENSION = DEFAULT_EMBEDDING_DIMENSION
 
 
-class GCPJsonFormatter(logging.Formatter):
-    """JSON formatter compatible with GCP Cloud Logging.
+class JsonFormatter(logging.Formatter):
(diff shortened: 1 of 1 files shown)


hindsight-api/hindsight_api/config.py
```diff
@@ -196,6 +201,37 @@
 EMBEDDING_DIMENSION = DEFAULT_EMBEDDING_DIMENSION
 
 
+class GCPJsonFormatter(logging.Formatter):
```
nicoloboschi: Just JsonFormatter

cdbartholomew: Thanks for catching that.

PR #238, opened 2026-01-30
fix: sanitize null bytes from text fields before PostgreSQL insertion

slayoffer:
## Summary
- Adds `_sanitize_text()` helper that strips null bytes (`\x00`) from text fields before PostgreSQL insertion
- Prevents `invalid byte sequence for encoding UTF8: 0x00` errors during batch retain

## Problem
Documents with embedded null bytes (common in OCR, PDF extraction, copy-paste from binary sources) cause PostgreSQL to reject the insert. This is a data integrity issue that silently breaks the retain pipeline.

## Solution
Strip null bytes from `fact_text` and `context` fields before insertion. Null bytes carry no semantic meaning in text and can be safely removed.

## Test plan
- [x] Lint passes
- [ ] Tested with documents containing null bytes

### DIFF
--- a/hindsight-api/hindsight_api/engine/retain/fact_storage.py
+++ b/hindsight-api/hindsight_api/engine/retain/fact_storage.py
@@ -8,16 +8,10 @@ import json
 import logging
 
 from ..memory_engine import fq_table
+from .fact_extraction import _sanitize_text
 from .types import ProcessedFact
 
 logger = logging.getLogger(__name__)
-
-
-def _sanitize_text(text: str | None) -> str | None:
-    """Remove null bytes that break PostgreSQL UTF-8 encoding."""
-    if text is None:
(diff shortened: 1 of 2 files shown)


hindsight-api/hindsight_api/engine/retain/fact_storage.py
```diff
@@ -13,6 +13,13 @@
 logger = logging.getLogger(__name__)
 
 
+def _sanitize_text(text: str | None) -> str | None:
```
nicoloboschi: we have already this function in fact_extraction._sanitize_text , can you resue that ?

slayoffer: hey. done

nicoloboschi (approved):

PR #248, opened 2026-01-30
fix(docker): add retry logic for ML model downloads

slayoffer:
## Summary
- Add 3 retries with exponential backoff (10s → 20s → 40s) for HuggingFace model downloads
- Set `HF_HUB_DOWNLOAD_TIMEOUT=600` for longer timeout
- Applied to both `api-only` and `standalone` stages

## Problem
Docker builds occasionally fail with transient network errors when downloading ML models from HuggingFace during the image build. This is especially common in CI environments or when HuggingFace CDN has temporary issues.

## Solution
Wrap the model download in a retry loop with exponential backoff. After 3 failed attempts, the build fails with a clear error message.

## Test plan
- [x] Built image locally with retry logic
- [x] Verified models download successfully on first attempt
- [x] Simulated failure by temporarily blocking HuggingFace (retry logic triggers)


nicoloboschi (approved):
Thanks! I was going to add it soon :)

PR #249, opened 2026-01-30
feat(docker): preload tiktoken encoding during build

slayoffer:
## Summary
Pre-download `cl100k_base` tiktoken encoding during Docker build to avoid runtime download delays.

## Why
Tiktoken is used by OpenAI models for tokenization. Without preloading, the first API call that uses tiktoken triggers a download, adding latency to the initial request.

## Changes
- Added tiktoken preload to both `api-only` and `standalone` stages
- Downloads `cl100k_base` encoding (most commonly used)

## Test plan
- [x] Built image locally
- [x] Verified tiktoken encoding is cached in image


nicoloboschi:
@slayoffer please resolve conflicts

PR #289, opened 2026-02-03
feat(openclaw): add external Hindsight API support

slayoffer:
## Summary

Add support for connecting to an external Hindsight API instead of starting a local daemon.

## Use cases

- **Shared memory** across multiple OpenClaw instances
- **Centralized deployment** (e.g., Hindsight on GKE)
- **Reduced resource usage** (no local daemon per instance)
- **Multi-tenant setups** where a central Hindsight serves multiple bots

## Configuration

```bash
# Via environment variable (highest priority)
export HINDSIGHT_EMBED_API_URL=https://hindsight.example.com
export HINDSIGHT_EMBED_API_TOKEN=optional-auth-token  # if API requires auth

# Or via plugin config
{
  "hindsightApiUrl": "https://hindsight.example.com",
  "hindsightApiToken": "optional-auth-token"
}
```

## How it works

When `hindsightApiUrl` is configured:
1. Skip local daemon startup entirely
2. Health check external API on startup
3. Pass API URL/token to CLI commands via env vars
4. All memory operations go to external API

When not configured, falls back to local daemon mode (existing behavior).

## Test plan

- [x] TypeScript build succeeds
- [ ] Manual test with external API deployment

### DIFF
--- a/hindsight-integrations/openclaw/openclaw.plugin.json
+++ b/hindsight-integrations/openclaw/openclaw.plugin.json
@@ -46,6 +46,15 @@
         "type": "number",
         "description": "Port for the openclaw profile daemon (default: 9077)",
         "default": 9077
+      },
+      "hindsightApiUrl": {
+        "type": "string",
+        "description": "External Hindsight API URL (e.g. 'https://mcp.hindsight.devcraft.team'). When set, skips local daemon and connects directly to this API.",
+        "format": "uri"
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
Awesome ! I was going to do it this week, thanks!

we should also update the documentation page, feel free to open another PR if you want

PR #312, opened 2026-02-06
fix: do not log db user/password

vanvuongngo:
### Fix security vulnerability
- remove logging of sensitive database credentials and if you need to log the full URL then implement it mask the user/password

**Before**
```
Starting Hindsight API...
  Version: v0.4.9
  URL: http://0.0.0.0:8888
  Database: postgresql://hindsight_user:12345678PASSWORD@db:5432/hindsight
```

```
INFO - hindsight_api.migrations - Database URL: postgresql://hindsight_user:12345678PASSWORD@db:5432/hindsight
```

```
INFO - hindsight_api.engine.memory_engine - Connecting to PostgreSQL at postgresql://hindsight_user:12345678PASSWORD@db:5432/hindsight
```

**After**
```
Starting Hindsight API...
  Version: v0.4.9
  URL: http://0.0.0.0:8888
```

```
DEBUG - hindsight_api.migrations - Database URL: postgresql://hindsight_user:12345678PASSWORD@db:5432/hindsight
```

```
DEBUG - hindsight_api.engine.memory_engine - Connecting to PostgreSQL at postgresql://hindsight_user:12345678PASSWORD@db:5432/hindsight
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,10 +116,16 @@ npm install @vectorize-io/hindsight-client
 ```javascript
 const { HindsightClient } = require('@vectorize-io/hindsight-client');
 
-const client = new HindsightClient({ baseUrl: 'http://localhost:8888' });
+const example = async () => {
+  const client = new HindsightClient({ baseUrl: 'http://localhost:8888' });
 
-await client.retain('my-bank', 'Alice loves hiking in Yosemite');
-await client.recall('my-bank', 'What does Alice like?');
+  await client.retain('my-bank', 'Alice loves hiking in Yosemite');
+
+  const results = await client.recall('my-bank', 'What does Alice like?');
+  console.log(results);
+}
+
+example();
 ```
 
 ---


vanvuongngo:
> Thanks for taking time of investigating this.
> 
> This is not a security vulnerability because the logs are only available on the server and not exposed in any way publicly.
> 
> But I understand the concern - the right fix is to show the Database URL but mask the username and password
> 
> Adding them to debug level is not a fix and not even a workaround - would you mind to do the masking fix ?

Hi @nicoloboschi,

thanks for your review and I am always open for discussion.

I worked in the high-security sector, where you have to protect your system against both external and internal threats (e.g., an internal user falling victim to a hacker attack). In that sense, it is a security vulnerability to log sensitive data like credentials or client identity data in production. Given the directive that prohibits enabling the debug level in production, my change is definitely a fix, while still allowing for debugging in local environments.

Admittedly, not all sectors have the same security requirements. If you do not have such a directive in place, my change would not be considered a fix.

I am not a python programmer. So I can try to vibe code it...

nicoloboschi (approved):

PR #327, opened 2026-02-09
feat(helm): add PDB and per-component affinity support

nuclon:
## Summary
- Add PodDisruptionBudget templates for API, control plane, and worker components (disabled by default via `<component>.podDisruptionBudget.enabled`)
- Support per-component affinity overrides (`api.affinity`, `controlPlane.affinity`, `worker.affinity`) that take precedence over the global `affinity` value
- Backward compatible: existing `affinity` config continues to work as a fallback for all components


nicoloboschi (approved):
LGTM

PR #333, opened 2026-02-10
feat(helm): add TEI reranker sidecar support

slayoffer:
## Summary

Adds optional TEI (Text Embeddings Inference) reranker sidecar support to the Helm chart.

When `reranker.enabled=true`:
- Deploys a TEI container as a sidecar in the API pod (localhost communication, no network hop)
- Auto-configures the API with `HINDSIGHT_API_RERANKER_PROVIDER=tei` and `HINDSIGHT_API_RERANKER_TEI_URL=http://localhost:<port>`
- Includes sensible defaults: `ms-marco-MiniLM-L-6-v2` model, `--auto-truncate` flag, 8090 port
- Shares a `model-cache` emptyDir volume for downloaded ONNX models
- Disabled by default — no impact on existing deployments

### Usage

```yaml
reranker:
  enabled: true
  # Optionally override defaults:
  # model: "cross-encoder/ms-marco-MiniLM-L-12-v2"
  # resources:
  #   limits:
  #     cpu: 4000m
```

### Why sidecar?

TEI as a sidecar provides sub-millisecond network latency for reranking (localhost), automatic lifecycle management with the API pod, and simpler deployment (no separate service to manage).

## Test plan

- [x] `helm lint` passes with `reranker.enabled=true` and `reranker.enabled=false`
- [x] `helm template` renders sidecar container, volume, and auto-injected env vars correctly
- [x] No sidecar rendered when disabled (default)
- [ ] Deploy to K8s cluster and verify TEI health + API reranking works

### DIFF
--- a/helm/hindsight/templates/_helpers.tpl
+++ b/helm/hindsight/templates/_helpers.tpl
@@ -128,6 +128,38 @@ API URL for control plane
 {{- end }}
 
 {{/*
+TEI reranker labels
+*/}}
+{{- define "hindsight.tei.reranker.labels" -}}
+{{ include "hindsight.labels" . }}
+app.kubernetes.io/component: tei-reranker
+{{- end }}
+
+{{/*
+TEI reranker selector labels
+*/}}
+{{- define "hindsight.tei.reranker.selectorLabels" -}}
+{{ include "hindsight.selectorLabels" . }}
+app.kubernetes.io/component: tei-reranker
+{{- end }}
+
+{{/*
+TEI embedding labels
+*/}}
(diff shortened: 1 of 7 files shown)


nicoloboschi (commented):
This would be a great addition!


1. sidecar is not the best way - you usually want to decouple the gpu-heavy workload to the api due to different scaling requirements - so having its own pair of svc/deployment would be much better
2. we could add the same identical support for the embeddings
3. values.tei.reranker and values.tei.embedding would be better as values

nicoloboschi (approved):
LGTM

PR #374, opened 2026-02-15
fix: propagate document_tags in async retain path

abix5:
## Summary
- fix async retain tag propagation by forwarding `document_tags` from the queued task payload in `_handle_batch_retain` to `retain_batch_async(...)`
- update `MemoryEngineInterface` signatures so async/sync batch retain methods explicitly accept optional `document_tags`
- add focused unit tests for async tag flow:
  - `submit_async_retain` includes `document_tags` in queued payload
  - worker handler forwards `document_tags` into retain execution

## Why
When `/memories` was called with `async=true` and `document_tags`, tags were dropped in the worker path. As a result, tags did not reliably appear in `/tags`, and tag-scoped workflows (including scoped mental models) behaved incorrectly.

## Testing
- `/Users/dmitriynenashev/Projects/hindsight/.venv-hs/bin/python -m pytest tests/test_async_retain_tags.py -o addopts='' -q`
- manual local API verification:
  - async retain with `document_tags`
  - verify `/v1/default/banks/{bank_id}/tags` contains the tag
  - verify tag-filtered recall returns expected results

## Follow-up
There is a separate inconsistency in `tags_match` semantics across endpoints (`recall/reflect` vs `mental-models/directives`). This PR does not change that behavior; a dedicated follow-up PR will unify those semantics.

### DIFF
--- a/hindsight-api/hindsight_api/engine/interface.py
+++ b/hindsight-api/hindsight_api/engine/interface.py
@@ -48,6 +48,7 @@ class MemoryEngineInterface(ABC):
         contents: list[dict[str, Any]],
         *,
         request_context: "RequestContext",
+        document_tags: list[str] | None = None,
     ) -> dict[str, Any]:
         """
         Retain a batch of memory items.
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
LGTM

Nice catch!

PR #421, opened 2026-02-22
Fix reflect based_on population and enforce full hierarchical retrieval

cdbartholomew:
## Summary

Fixes two related bugs in the reflect agent that caused incomplete `based_on` metadata and mental models producing "I don't have information."

- **`based_on` incomplete**: Observations from `search_observations` were never extracted into `based_on`. Mental model entries used non-existent `summary`/`description` fields instead of `content`, producing empty text. A duplicate unreachable `elif` block for `search_mental_models` was dead code.
- **Mental models empty**: When a bank has mental models, tool_choice forcing only covered iteration 0 (`search_mental_models`). Iterations 1+ were `auto`, allowing the LLM to short-circuit without ever calling `search_observations` or `recall`. With LOW budget guidance encouraging speed ("stop there if mental models provide an answer"), the agent would often produce a response after a single tool call. This created a self-reinforcing failure loop: a failed mental model refresh set content to "I don't have information", which subsequent reflects would find and trust.

### `memory_engine.py` — `based_on` population in `reflect_async()`
- Add `search_observations` extraction into `based_on["observation"]` with `used_observation_ids_set` filtering
- Fix `search_mental_models` / `get_mental_model` to read `model.get("content", "")` instead of `model.get("summary") or model.get("description", "")`
- Remove duplicate unreachable `elif tc.tool == "search_mental_models"` block

### `agent.py` — forced hierarchical retrieval
- **With mental models**: force `search_mental_models`(iter 0) → `search_observations`(iter 1) → `recall`(iter 2) → auto(iter 3+)
- **Without mental models** (unchanged): force `search_observations`(iter 0) → `recall`(iter 1) → auto(iter 2+)

This ensures all three knowledge levels are always consulted before the agent enters auto mode, matching the retrieval strategy described in the system prompt. With LOW budget (`max_iterations=5`), the agent still has 2 auto iterations remaining for additional searches or calling `done()`.

### Why force all three retrieval levels?

The previous code relied on the LLM following the system prompt's instruction to search all levels. In practice, with LOW budget and Gemini Flash Lite, the agent frequently short-circuited after `search_mental_models` returned 0 results (e.g. during mental model refresh where the model being refreshed is excluded). The system prompt says "MANDATORY: you MUST call recall() before giving up", but the LLM didn't always comply. Forcing the tool_choice for the first 3 iterations is a deterministic guarantee that the full hierarchy is consulted, regardless of LLM behavior.

## Test plan

- [x] `test_reflect_agent.py` — 27/27 passed
- [x] hindsight-cloud unit tests — 1394 passed (3 container-timeout errors unrelated)
- [x] Manual verification: reflect with `include.facts` + `include.tool_calls` shows all 3 forced tool calls in trace
- [x] Manual verification: mental model create produces rich content with populated `based_on` (observations, memories, mental models all present)
- [x] Reproduced original failure on hindsight-dev `test` bank, confirmed fix resolves it locally

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -31,6 +31,9 @@ jobs:
       - run: npm ci --workspace=hindsight-docs
       - run: uv run generate-llms-full
       - run: npm run build --workspace=hindsight-docs
+        env:
+          UMAMI_URL: https://analytics.hindsight.vectorize.io
+          UMAMI_WEBSITE_ID: ${{ secrets.UMAMI_WEBSITE_ID }}
       - uses: actions/upload-pages-artifact@v3
         with:
           path: hindsight-docs/build
--- a/hindsight-docs/docusaurus.config.ts
+++ b/hindsight-docs/docusaurus.config.ts
@@ -2,6 +2,9 @@ import {themes as prismThemes} from 'prism-react-renderer';
 import type {Config} from '@docusaurus/types';
 import type * as Preset from '@docusaurus/preset-classic';
 
+const umamiUrl = process.env.UMAMI_URL;
+const umamiWebsiteId = process.env.UMAMI_WEBSITE_ID;
+
 // Announcement bar - supports HTML for links
 // Set to empty string '' to hide the bar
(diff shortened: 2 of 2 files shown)


nicoloboschi (approved):
LGTM

PR #437, opened 2026-02-25
feat: temporal-aware batching in consolidation

dcbouius:
## Summary

Sort each tag group by `occurred_start` (falling back to `mentioned_at`) before chunking into LLM batches during consolidation. This ensures temporally co-occurring facts land in the same batch, giving the consolidation LLM the opportunity to merge related facts into richer observations.

## Change

**`consolidator.py`** — Added a temporal sort to each tag group before slicing into LLM batches. The sort key falls back through `occurred_start` → `mentioned_at` → `datetime.min`. No behavior change for facts with identical timestamps (sort is stable).

## Test plan

- [x] Ruff lint and format pass
- [x] Pre-commit hooks pass
- [x] Existing test suite: all failures are pre-existing DB migration issues on main
- [x] Manual validation with multimodal ingest + consolidation

### DIFF
--- a/hindsight-api/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api/hindsight_api/engine/consolidation/consolidator.py
@@ -216,9 +216,16 @@ async def run_consolidation_job(
             tag_key = tuple(sorted(m.get("tags") or []))
             tag_groups.setdefault(tag_key, []).append(dict(m))
 
-        # Flatten into LLM batches respecting both tag groups and llm_batch_size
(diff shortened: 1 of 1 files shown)


nicoloboschi (changes requested):
we should not rely on the consolidation batch to get things aggregated together because this will not scale. we have an hard limit on the batch size due to the LLM context window , so we can't expect this to work reliably (e.g. you have 300 memories on the same timeframe) - we need to tackle this problem from a different angle

PR #452, opened 2026-02-27
feat(openclaw): retain last n+2 turns every n turns (default n=10)

### LINKED ISSUE #451: openclaw plugin: O(n²) storage growth from retaining full message history every turn
## Problem

The openclaw plugin's `agent_end` hook sends the **entire** `event.messages` array to Hindsight on every turn. Since messages accumulate within a session, this creates O(n²) data growth:

| Turn | Messages sent | Cumulative total |
|------|:---:|:---:|
| 1 | 2 | 2 |
| 10 | 20 | 110 |
| 25 | 50 | 650 |
| 50 | 100 | 2,550 |

A typical 50-turn conversation generates **2,550 messages across 50 retain calls**, while only ~100 messages are unique content. Over 90% is duplicated.

## Impact

- **Hindsight API**: 50 retain calls per conversation (each increasingly large)
- **LLM extraction**: Hindsight processes each document through its extraction pipeline — the same facts get extracted repeatedly
- **Storage**: quadratic growth in total characters stored
- **Memory quality**: duplicate facts reduce signal-to-noise ratio for recall

## Proposed solution

Retain the last n+2 turns every n turns instead of the full history every turn. Default n=10, which reduces retain calls by 90% and messages sent by ~95% while preserving all unique content through a 2-turn sliding window overlap.

I have a working implementation tested in production for ~2 weeks and will submit a PR.

fabioscarsi:
## What

Adds chunked retention to the openclaw plugin: instead of sending the full message history to Hindsight on every turn, it retains the last n+2 turns every n turns (default n=10).

Closes #451

## Why

The current `agent_end` hook sends the **entire** `event.messages` array on every turn. Since messages accumulate within a session, a 50-turn conversation sends 2,550 cumulative messages across 50 API calls — while only ~100 are unique. This creates O(n²) storage growth and causes Hindsight to extract the same facts repeatedly.

## How

- Every `agent_end` event increments a per-session turn counter
- Only every 10th turn triggers a retain call (configurable via `retainEveryNTurns`)
- On retain turns, a sliding window of the last 24 messages is sent (12 turns = 10 new + 2 overlap for context continuity)
- At the first retain (turn 10), there are only ~20 messages, so all are included
- Turn counting is per-session (keyed by `bankId:sessionKey`), so multi-channel setups track independently
- Default is 10 — current behavior (retain every turn) requires explicitly setting `retainEveryNTurns: 1`

## Impact (default n=10)

| Metric | Before | After | Reduction |
|--------|--------|-------|-----------|
| Retain API calls (50-turn session) | 50 | 5 | 90% |
| Messages sent to API | 2,550 | ~120 | 95% |
| Unique content preserved | 100% | 100% | — |

## Changes

- `src/types.ts`: Added `retainEveryNTurns?: number` to `PluginConfig`
- `src/index.ts`: Added turn tracking Map and chunked retention logic in `agent_end` handler

## Testing

Tested in production for ~2 weeks on openclaw v2026.2.25-2026.2.27 with Hindsight in external API mode. Verified via gateway logs showing correct skip/retain pattern and Hindsight dashboard showing controlled document sizes.

### DIFF
--- a/hindsight-integrations/openclaw/src/index.ts
+++ b/hindsight-integrations/openclaw/src/index.ts
@@ -20,6 +20,7 @@ const banksWithMissionSet = new Set<string>();
 // In-flight recall deduplication: concurrent recalls for the same bank reuse one promise
 import type { RecallResponse } from './types.js';
 const inflightRecalls = new Map<string, Promise<RecallResponse>>();
+const turnCountBySession = new Map<string, number>();
 const RECALL_TIMEOUT_MS = 10_000;
 
 // Cooldown + guard to prevent concurrent reinit attempts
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM

PR #658, opened 2026-03-23
feat: add SmolAgents integration with Hindsight memory tools

benfrank241:
## Summary

- Adds `hindsight-smolagents` package under `hindsight-integrations/smolagents/`
- Three native SmolAgents `Tool` subclasses: `HindsightRetainTool`, `HindsightRecallTool`, `HindsightReflectTool`
- `create_hindsight_tools()` factory and `memory_instructions()` helper for system prompt injection
- 81 unit tests, lint clean
- Docs page, sidebar entry, integrations grid entry, and HuggingFace icon

## Design notes

- `bank_id` is baked in at construction time (SmolAgents has no `RunContext` equivalent)
- Sync `forward()` methods call the Hindsight Python client directly
- Follows the same config/error/tools module pattern as `hindsight-agno`

## Test plan

- [ ] `cd hindsight-integrations/smolagents && uv run pytest tests/ -v` — all 81 tests pass
- [ ] `uv run ruff check . && uv run ruff format --check .` — lint clean
- [ ] Manual end-to-end: start API, retain/recall/reflect against a live bank
- [ ] Docs page renders and icon displays in integrations grid
- [ ] `./scripts/release-integration.sh smolagents` recognizes the integration name


nicoloboschi:
Force-pushed a rebase + review fixes (was 074f17ac, now fff31345).

**Rebased onto current `main`** and resolved three conflicts (`scripts/release-integration.sh`, `hindsight-docs/src/data/integrations.json`, `.github/workflows/test.yml`) — the older merge commit `074f17ac` is gone. Branch is now 4 commits ahead of main: the original three plus one follow-up.

**Fixes applied** (commit `fff31345`):
- Added `smolagents` to the consolidated `INTEGRATIONS` table in `hindsight-dev/hindsight_dev/generate_changelog.py`. Without this the release script aborts at the changelog-generation step (this was originally three parallel lists; #1314 collapsed them into one, but the new integration still needs an entry).
- Added a sidebar link in `hindsight-docs/sidebars.ts`. The docs page existed but wasn't reachable from navigation.
- `examples/interactive_test.py`: ruff F541 (`f""` with no placeholders) and I001 (import order).
- `ruff format` unwrapped a few paren-wrapped expressions in `tools.py`.

**Verified locally**: `uv run pytest tests -q` → 81 passed, `ruff check`/`ruff format --check` → clean.

**Things I left for you to decide** (not blocking, but worth a thought):
- `HindsightRetainTool._ensure_bank` (`tools.py:85-93`) catches **every** exception during `create_bank` and still adds the bank to `_created_banks` — auth/network failures get silently filed away, then `retain` proceeds against a bank that may not exist. Maybe narrow the catch to "already exists" cases?
- `memory_instructions` returns `""` on any exception (`tools.py:371-373`). Comment says "shouldn't block the agent" — fair, but it also swallows e.g. `AttributeError`s during refactors. A `logger.warning(..., exc_info=True)` would help.
- Sync `Hindsight` client inside SmolAgents `Tool.forward` is reasonable since SmolAgents is sync top-level, but the sync wrapper uses `loop.run_until_complete` and will raise if the agent is ever invoked from inside a running event loop (FastAPI handler, etc.). Worth a note in the docs or README.

PR #734, opened 2026-03-28
Fix Codex named tool_choice serialization in reflect

Sapientropic:
## Summary

This fixes a Codex provider compatibility issue in reflect/tool-calling paths.

Reflect can force a specific tool on early iterations using the legacy named tool choice shape:

```json
{"type":"function","function":{"name":"recall"}}
```

The current Codex Responses API expects the function name at the top level instead:

```json
{"type":"function","name":"recall"}
```

Without that normalization, Codex rejects the request and reflect can end up with empty `based_on.memories` even when recallable evidence exists.

## What changed

- normalize legacy named `tool_choice` dicts in `CodexLLM.call_with_tools()`
- add a regression test that verifies the serialized payload no longer contains `tool_choice.function`
- add a regression test that verifies forced tool choice still yields tool calls

## Why this belongs in the provider

This follows the same design direction as the existing provider-specific tool-choice compatibility handling for LM Studio / OpenAI-compatible providers:

- the reflect agent keeps its provider-agnostic intent
- each provider adapts the request shape it needs

## Validation

Targeted regression:

```bash
python -m pytest -o addopts="" --noconftest hindsight-api-slim/tests/test_codex_tool_choice.py -q
```

Result:

```text
2 passed
```

I also verified the original failure mode before the patch via a live Codex-backed daemon, where reflect requests were rejected with:

```text
400 Unknown parameter: 'tool_choice.function'
```

After this patch, the same integration recovered real reflect evidence instead of failing in the forced-tool iterations.

## Non-goals

This PR does not claim to fix every possible `reflect` empty-evidence case.
It specifically fixes the Codex provider request-shape incompatibility for forced named tool choice.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
@@ -126,6 +126,32 @@ class CodexLLM(LLMInterface):
         }
         return mapping.get(effort.lower(), "auto")
 
+    def _normalize_tool_choice(self, tool_choice: str | dict[str, Any]) -> str | dict[str, Any]:
+        """Normalize forced function tool choice for the Codex Responses API.
+
+        Older agent paths may still pass OpenAI chat-completions style named
+        tool choice payloads such as:
+
+            {"type": "function", "function": {"name": "recall"}}
+
+        Codex Responses expects the named function at the top level instead:
+
+            {"type": "function", "name": "recall"}
+        """
+        if not isinstance(tool_choice, dict):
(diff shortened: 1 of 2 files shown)


chriscoey:
Can confirm this bug. I'm using `openai-codex` provider with `gpt-5.4` and reflect was returning "I don't have information" despite recall finding 12 results for the same query. The Responses API expects `{"type": "function", "name": "recall"}` but the agent sends `{"type": "function", "function": {"name": "recall"}}`. Applied the same fix locally and reflect works correctly now.

PR #735, opened 2026-03-28
fix: parse query params from base_url in OpenAI embeddings client

akhater:
## Problem

The OpenAI-compatible LLM provider parses query parameters (e.g. ?api-version=xxx for Azure OpenAI) from the base_url and passes them as default_query to the OpenAI client. However, the OpenAI embeddings provider does not do this, causing Azure OpenAI embeddings to fail with 404 Resource not found errors at runtime.

The issue is that Azure OpenAI requires api-version as a query parameter on every request. Without it, the endpoint returns 404.

## Fix

Applied the same URL parsing logic from openai_compatible_llm.py to embeddings.py:
- Parse query parameters from base_url using urlparse
- Strip query params from the URL passed to the OpenAI client
- Pass extracted params as default_query so they are included in every request

## Testing

Verified locally with Azure OpenAI text-embedding-3-small deployment:
- Without fix: openai.NotFoundError: 404 - Resource not found
- With fix: SUCCESS - dim: 1536


akhater:
Done, removed the CI workflow file. Thanks for the quick review\!

PR #750, opened 2026-03-30
feat: surface reranker relevance score in recall results

### LINKED ISSUE #748: feat: surface reranker relevance score in recall results
## Problem

Recall results don't include a relevance score, even though the reranker computes one internally.

The `MemoryFact` response model already has a `relevance: float` field (defaults to 0.0), and `ScoredResult.to_dict()` produces a `combined_score`. But when recall results are converted to `MemoryFact` objects at `memory_engine.py:3367`, the score is never wired through:

```python
# Current (line ~3367):
memory_facts.append(
    MemoryFact(
        id=result_id,
        text=result_dict.get("text"),
        # ... other fields
        # relevance is never set — defaults to 0.0
    )
)
```

This means:
- A query for "pizza recipes" against an engineering bank returns results with no way to know they're irrelevant
- Consumers can't set a relevance threshold to filter weak matches
- Empty results are ambiguous: nothing relevant vs nothing stored

## Proposed fix

Add one line to the `MemoryFact` constructor:

```python
relevance=result_dict.get("combined_score", 0.0),
```

This surfaces the already-computed reranker score in the API response. I'm happy to submit a PR for this if it would be welcome — it's a minimal change.

## Expected result

```json
{
  "results": [
    {"text": "...", "type": "world", "relevance": 0.87, ...},
    {"text": "...", "type": "observation", "relevance": 0.23, ...}
  ]
}
```

chriscoey:
## Summary

The reranker computes a combined relevance score for each recall result, but this score wasn't included in the API response. The `MemoryFact` model in `response_models.py` had a `relevance` field and `ScoredResult.to_dict()` produced `combined_score`, but the value was never wired through.

This adds the plumbing across 3 files (4 lines total):
- `response_models.py`: add `relevance` field to `MemoryFact`
- `memory_engine.py`: populate `relevance` from `combined_score` during recall
- `http.py`: add `relevance` field to API `RecallResult` model + wire in `_fact_to_result`

## Before / After

**Before**: recall results have no relevance signal
```json
{"text": "PyRel is the new Python interface...", "type": "world"}
```

**After**: each result includes the reranker's combined score
```json
{"text": "PyRel is the new Python interface...", "type": "world", "relevance": 0.73}
```

Tested locally: relevant queries score 0.68-0.73, irrelevant queries score ~0.55.

Backward compatible — the field defaults to 0.0 so existing consumers aren't affected.

Closes #748

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -225,6 +225,7 @@ class RecallResult(BaseModel):
     source_fact_ids: list[str] | None = (
         None  # IDs of source facts (observation type only, when source_facts is enabled)
     )
+    relevance: float = Field(default=0.0, description="Combined relevance score from reranking")
 
 
 class EntityObservationResponse(BaseModel):
@@ -2548,6 +2549,7 @@ def _register_routes(app: FastAPI):
                     chunk_id=fact.chunk_id,
                     tags=fact.tags,
                     source_fact_ids=fact.source_fact_ids,
+                    relevance=fact.relevance,
                 )
 
             recall_results = [_fact_to_result(fact) for fact in core_result.results]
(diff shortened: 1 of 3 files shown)


nicoloboschi (changes requested):
hey, thanks for looking into it. 

exposing a relevancy field is something we've decided to not do by design, for multiple reasons:
1. we want flexibility to change the scoring algorithm using only relative order of results
2. the score is a combination of multiple signals and it's not meant to be absolute. 
3. the client can't really understand what to do with it. thresholds are ok-ish for single-signal relevancy fields such as knn, cross-encoder scoring, text search but it doens't apply to our scoring system


Happy to discuss about it if you want

PR #829, opened 2026-04-01
fix(tests): add retries for flaky LLM-dependent tests

dcbouius:
## Summary
- Add retry loops (up to 3 attempts) to 4 tests that fail intermittently due to LLM non-determinism or transient DB errors:
  - `test_reflect_searches_mental_models_when_available` - LLM sometimes picks `recall` instead of `search_mental_models`
  - `test_trigger_tag_groups_override_flat_tags` - LLM content generation may not include expected keywords
  - `test_llm_provider_api_methods` - Gemini sometimes returns 0 tool calls; also strengthened the prompt
  - `test_mental_models_dimension_change_empty_table` - transient PostgreSQL OID error from concurrent DDL

## Test plan
- [ ] CI passes without flaky test failures
- [ ] Retry pattern matches existing convention (see `test_skip_podcast_meta_commentary`)


nicoloboschi (changes requested):
instead of plumbing retries on the code, we should use the pytest-rerunfailures plugin that handles this cleanly with a @pytest.mark.flaky(reruns=3) decorator


can you look into that ?

PR #867, opened 2026-04-04
fix: Windows compatibility for hindsight-embed

beordie:
- Add cross-platform file locking support
- Use fcntl on Unix-like systems, msvcrt on Windows
- Add detailed documentation explaining why we don't use external libraries


nicoloboschi (approved):
LGTM, thanks

PR #869, opened 2026-04-04
fix(mcp): coerce string-encoded metadata, types, and tags in tool calls

### LINKED ISSUE #849: MCP: auto-coerce string-encoded arrays/objects in tool call arguments (tags, types, metadata)
## Problem

Multiple MCP clients (Claude Code, Cursor with various LLM backends) consistently serialize `tags`, `types`, and `metadata` parameters as **JSON-encoded strings** instead of native JSON arrays/objects. This causes Pydantic validation failures:

```
Error: 1 validation error for call[retain]
tags
  Input should be a valid list [type=list_type, input_value='["educore", "scope:ident...olicies", "type:guide"]', input_type=str]

Error: 1 validation error for call[recall]
types
  Input should be a valid list [type=list_type, input_value='["world"]', input_type=str]

Error: 1 validation error for call[retain]
metadata
  Input should be a valid dictionary [type=dict_type, input_value='{"branch": "develop", "f...y/domain/constants.ts"}', input_type=str]
```

### Key observations

- This is **not** a user error — the LLM agent generates tool calls with string-wrapped arrays/objects
- The agent often **retries with the exact same broken format** 4+ times in a row, wasting tokens and failing silently
- Extensive prompt engineering (templates, ✗/✓ comparisons, recovery procedures) has been attempted and **proven ineffective**
- The values inside the strings are valid JSON — they just need to be unwrapped

### Affected parameters

| Parameter | Expected type | Actually received |
|-----------|--------------|-------------------|
| `tags` | `list[str]` | `'["tag1", "tag2"]'` (string) |
| `types` | `list[str]` | `'["world"]'` (string) |
| `metadata` | `dict[str, str]` | `'{"key": "value"}'` (string) |

## Proposed fix

Add a Pydantic `@field_validator` (or `BeforeValidator`) that auto-coerces string-encoded JSON:

```python
from pydantic import field_validator
import json

@field_validator("tags", "types", mode="before")
@classmethod
def coerce_string_to_list(cls, v):
    if isinstance(v, str):
        try:
            parsed = json.loads(v)
            if isinstance(parsed, list):
                return parsed
        except (json.JSONDecodeError, TypeError):
            pass
    return v

@field_validator("metadata", mode="before")
@classmethod
def coerce_string_to_dict(cls, v):
    if isinstance(v, str):
        try:
            parsed = json.loads(v)
            if isinstance(parsed, dict):
                return parsed
        except (json.JSONDecodeError, TypeError):
            pass
    return v
```

This is a common pattern for APIs consumed by LLM agents. The fix is backward-compatible — native arrays/objects pass through unchanged.

## Impact

Without this fix, a significant percentage of MCP retain/recall/reflect calls fail silently, causing data loss in memory operations. The agent often gives up after repeated failures, leaving the user without stored memories for the session.

octo-patch:
Fixes #849

## Problem
MCP tool bridges (Claude Code, Cursor, and others) sometimes serialize JSON arrays and objects as strings during transport. This causes Pydantic validation failures on `retain` and `recall` calls:

```
Error: 1 validation error for call[retain]
tags
  Input should be a valid list [type=list_type, input_value='["tag1", "tag2"]', input_type=str]

Error: 1 validation error for call[recall]
types
  Input should be a valid list [type=list_type, input_value='["world"]', input_type=str]

Error: 1 validation error for call[retain]
metadata
  Input should be a valid dictionary [type=dict_type, input_value='{"key": "value"}', input_type=str]
```

The agent retries with the same broken format, wasting tokens and silently failing to store memories.

## Solution
Add defensive coercion at two layers — mirroring the pattern already used for `MemoryItem.tags` (added in #682):

**HTTP API (`http.py`)**:
- `RecallRequest`: `field_validator` on `types` and `tags` (`mode="before"`) that parses JSON-string arrays back into lists
- `MemoryItem`: `field_validator` on `metadata` (`mode="before"`) that parses JSON-string objects back into dicts

**MCP tools (`mcp_tools.py`)**:
- `build_content_dict`: coerce `metadata` from JSON string to dict, matching the existing `tags` coercion in the same function
- Both `recall()` variants (with and without `bank_id` param): coerce `types` and `tags` from JSON strings before passing to the memory engine

All coercions are backward-compatible — correctly-typed native arrays/objects pass through unchanged.

## Testing
The existing coercion pattern for `tags` in `MemoryItem` (from #682) already has test coverage. The new validators follow the identical logic and can be validated by sending a JSON-string value to the affected fields.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -172,6 +172,23 @@ class RecallRequest(BaseModel):
         "Each group is a leaf {tags, match} or compound {and: [...]}, {or: [...]}, {not: ...}.",
     )
 
+    @field_validator("types", "tags", mode="before")
+    @classmethod
+    def coerce_string_to_list(cls, v: Any) -> Any:
+        """Coerce JSON-string arrays to list.
+
+        MCP tool bridges sometimes serialize JSON arrays as strings during
+        transport, e.g. '["world"]' instead of ["world"]. Parse them back.
+        """
+        if isinstance(v, str):
+            try:
+                parsed = json.loads(v)
+                if isinstance(parsed, list):
+                    return parsed
+            except (json.JSONDecodeError, TypeError):
+                pass
+        return v
+
     @field_validator("query")
     @classmethod
     def validate_query_not_empty(cls, v: str) -> str:
@@ -446,6 +463,23 @@ class MemoryItem(BaseModel):
             except (json.JSONDecodeError, TypeError):
                 pass
             return [v]
+        return v
+
+    @field_validator("metadata", mode="before")
+    @classmethod
+    def coerce_metadata(cls, v: Any) -> Any:
+        """Coerce JSON-string metadata to dict.
+
+        MCP tool bridges sometimes serialize JSON objects as strings during
+        transport, e.g. '{"key": "value"}' instead of {"key": "value"}.
+        """
+        if isinstance(v, str):
+            try:
+                parsed = json.loads(v)
+                if isinstance(parsed, dict):
+                    return parsed
+            except (json.JSONDecodeError, TypeError):
+                pass
         return v
 
     observation_scopes: Literal["per_tag", "combined", "all_combinations"] | list[list[str]] | None = Field(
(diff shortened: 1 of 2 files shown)


nicoloboschi:
I did a more comprehensive fix here https://github.com/vectorize-io/hindsight/pull/899, thanks

PR #871, opened 2026-04-05
fix(openclaw): fall back to sessionKey when ctx.channelId is a provider name

### LINKED ISSUE #854: OpenClaw plugin: ctx.channelId returns provider name instead of channel ID for Discord sessions
### Environment
- OpenClaw version: 2026.4.1
- Hindsight plugin: `@vectorize-io/hindsight-openclaw@0.4.19`
- Channel: Discord
- Config: `dynamicBankGranularity: ["agent", "channel"]`

### Expected Behavior
With `dynamicBankGranularity: ["agent", "channel"]`, Discord channel memories should be isolated per-channel. For a session key like `agent:main:discord:channel:1472750640760623226`, the bank ID should resolve to something like `main::channel:1472750640760623226` (or ideally the channel name like `main::morpheus`).

### Actual Behavior
All Discord memories go to a single bank `main::discord` regardless of which channel the conversation is in.

### Root Cause
In `deriveBankId()`, the field map uses:
```js
channel: ctx?.channelId || sessionParsed.channel || 'unknown'
```

The `ctx.channelId` value is `"discord"` (the provider/channel type name) rather than the actual Discord channel ID (e.g. `1472750640760623226`). Since `ctx.channelId` has priority over the session key parser fallback, the bank always resolves to `main::discord`.

Debug output confirms:
```
[Hindsight] before_prompt_build - bank: main::discord, channel: undefined/undefined
```
(Note: `ctx.messageProvider` is also `undefined` in some hook contexts)

The `parseSessionKey()` fallback would correctly extract `channel:1472750640760623226` from the session key, but it never gets used because `ctx.channelId` is truthy (it's `"discord"`).

### Workaround Attempted
None found within plugin config. The issue appears to be upstream — OpenClaw's hook context populates `ctx.channelId` with the provider name rather than the Discord channel snowflake ID.

### Suggested Fix
1. **In OpenClaw core**: Populate `ctx.channelId` with the actual channel/conversation identifier (e.g. the Discord channel snowflake) rather than the provider name.
2. **In the plugin** (defensive): If `ctx.channelId` matches a known provider name (discord, telegram, slack, etc.), treat it as missing and fall through to the `sessionParsed.channel` fallback.

### Impact
- All Discord channel memories merge into one bank, defeating per-channel isolation
- Users who configure `dynamicBankGranularity` with `"channel"` get no benefit on Discord
- Likely affects other multi-channel providers too (Telegram groups, Slack channels)

octo-patch:
Fixes #854

## Problem

Some OpenClaw providers (e.g. Discord) populate `ctx.channelId` with the platform/provider name (e.g. `"discord"`) rather than the actual channel identifier (e.g. `"1472750640760623226"`). Because `ctx.channelId` has priority over the sessionKey fallback in `deriveBankId()`, all Discord channel memories merged into a single bank (`main::discord`) regardless of which channel was active — defeating per-channel isolation for users with `dynamicBankGranularity: ["agent", "channel"]`.

Debug output from the issue:
```
[Hindsight] before_prompt_build - bank: main::discord, channel: undefined/undefined
```

## Solution

In `deriveBankId()`, check if `ctx.channelId` exactly matches a known provider/platform name (discord, telegram, slack, whatsapp, signal, matrix, irc, line). If it does, treat the value as missing and fall through to the `sessionParsed.channel` derived from the sessionKey, which correctly carries the real channel identifier.

Only exact case-insensitive matches are filtered — channel names like `"slack-general"` or `"my-discord-bot"` pass through unchanged.

## Testing

- Added two new unit tests to `derive-bank-id.test.ts`:
  1. Verifies that a Discord context where `ctx.channelId = "discord"` uses the channel extracted from the sessionKey instead.
  2. Verifies that a channel ID like `"slack-general"` (which contains a provider name as substring) is preserved correctly.
- All 68 existing tests continue to pass.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -172,6 +172,23 @@ class RecallRequest(BaseModel):
         "Each group is a leaf {tags, match} or compound {and: [...]}, {or: [...]}, {not: ...}.",
     )
 
+    @field_validator("types", "tags", mode="before")
+    @classmethod
+    def coerce_string_to_list(cls, v: Any) -> Any:
+        """Coerce JSON-string arrays to list.
+
+        MCP tool bridges sometimes serialize JSON arrays as strings during
+        transport, e.g. '["world"]' instead of ["world"]. Parse them back.
+        """
+        if isinstance(v, str):
+            try:
+                parsed = json.loads(v)
+                if isinstance(parsed, list):
+                    return parsed
+            except (json.JSONDecodeError, TypeError):
+                pass
+        return v
+
     @field_validator("query")
     @classmethod
     def validate_query_not_empty(cls, v: str) -> str:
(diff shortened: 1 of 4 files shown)


nicoloboschi (changes requested):
LGTM, except for the API files changed, can you keep the PR only for openclaw? thanks

PR #942, opened 2026-04-09
feat(openclaw): add resilient startup and richer retain metadata

Aldoustheorchestrator:
## Summary
- ignore the OpenClaw heartbeat/control provider by default via `excludeProviders: ["heartbeat"]`
- enrich retain payloads with stable session-scoped document ids and richer OpenClaw metadata
- lazily initialize the Hindsight client from hook execution when OpenClaw skips `service.start()` for late-loaded plugins

## Why
The OpenClaw integration is often installed manually as an external plugin. In that runtime path, the gateway can load the plugin and register its hooks after sidecar startup, but never invoke the plugin service's `start()` lifecycle. That leaves recall and retain hooks live while the Hindsight client remains uninitialized.

This patch makes the integration more resilient in that environment:
- heartbeat/control traffic is ignored by default so it does not pollute memory
- retain calls carry richer metadata so stored documents are easier to reason about and update
- hook-driven lazy initialization lets the plugin recover when OpenClaw misses the normal service start path

## What the lazy initialization buys us
When OpenClaw loads the plugin late, hooks still run but `initPromise` is unset. Instead of no-oping forever, `waitForReady()` now attempts a throttled lazy bootstrap using the already-resolved plugin config. That keeps recall/retain functional in live deployments even while the upstream gateway lifecycle bug remains.

## What the richer retain metadata buys us
Retain requests now include stable, session-scoped `document_id` values and additional OpenClaw context such as:
- `retention_scope`
- `turn_index`
- `session_key`
- `agent_id`
- `provider`
- `channel_id`
- `thread_id`
- `window_turns` when retaining a conversation window

That gives Hindsight better document identity and provenance, improves update/overwrite semantics, and makes downstream inspection/debugging much saner.

## Testing
- `npm --prefix hindsight-integrations/openclaw test`
- `npm --prefix hindsight-integrations/openclaw run build`

## Live validation
Validated against a manual local install in OpenClaw at `/home/node/.openclaw/extensions/hindsight-openclaw`.
Observed live runtime behavior before the workaround:
- plugin entry and hook registration occurred
- `service.start()` was never called by the gateway
- retain hooks logged `waitForReady called before service.start()` and no-op'd

Observed after the workaround:
- hooks trigger lazy initialization fallback
- memories are injected successfully during live traffic
- heartbeat provider remains ignored by default

### DIFF
--- a/hindsight-integrations/openclaw/README.md
+++ b/hindsight-integrations/openclaw/README.md
@@ -51,12 +51,15 @@ Optional settings in `~/.openclaw/openclaw.json` under `plugins.entries.hindsigh
 | `retainTags` | `[]` | Tags applied to every retained document, useful for cross-agent/source labeling (e.g. `source_system:openclaw`, `agent:agentname`) |
 | `retainSource` | `"openclaw"` | `source` value written into retained document metadata |
 | `dynamicBankGranularity` | `["agent", "channel", "user"]` | Fields used to derive bank ID. Options: `agent`, `channel`, `user`, `provider` |
-| `excludeProviders` | `[]` | Message providers to skip for recall/retain (e.g. `slack`, `telegram`, `discord`) |
+| `excludeProviders` | `["heartbeat"]` | Message providers to skip for recall/retain (e.g. `heartbeat`, `slack`, `telegram`, `discord`) |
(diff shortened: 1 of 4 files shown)


nicoloboschi (approved):
LGTM

PR #943, opened 2026-04-09
fix(opencode): use msg.info.role instead of msg.role in getSessionMessages

octo-patch:
Fixes #941

## Problem

The opencode plugin's `getSessionMessages` function reads `msg.role` to determine message role, but the opencode SDK returns messages as `Array<{ info: Message; parts: Part[] }>` — the role is on `msg.info.role`, not `msg.role` directly.

Since `msg.role` is always `undefined`, the role check `role !== 'user' && role !== 'assistant'` is always `true`, and every message is filtered out. This causes `getSessionMessages` to always return an empty array, silently breaking auto-retain and auto-recall for all opencode versions including 1.4.0.

## Solution

- Changed `msg.role` → `msg.info.role` in `getSessionMessages`
- Updated the `OpencodeClient` type to accurately reflect the SDK's `{ info: { role: string }; parts: ... }` structure
- Updated test fixtures to match the real SDK message shape

## Testing

All 89 existing tests pass with the updated message shape, now correctly exercising the actual code path.

### DIFF
--- a/hindsight-integrations/opencode/src/hooks.test.ts
+++ b/hindsight-integrations/opencode/src/hooks.test.ts
@@ -20,7 +20,7 @@ function makeClient() {
     } as any;
 }
 
-function makeOpencodeClient(messages: Array<{ role: string; parts: Array<{ type: string; text?: string }> }> = []) {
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):

nicoloboschi (changes requested):
we should still also support 1.3.x

PR #947, opened 2026-04-09
feat: database abstraction layer for Oracle 23ai support

DK09876:
## Summary

- Introduces `DatabaseBackend` / `DatabaseConnection` ABCs and `SQLDialect` ABC to decouple the engine from asyncpg/PostgreSQL, enabling Oracle 23ai as a second database backend
- Full Oracle 23ai support: migrations, HNSW vector search, Oracle Text BM25, entity resolution, graph retrieval
- All retrieval SQL moved into dialect abstraction — no more inline `_is_pg` checks
- Worker poller refactored to use DatabaseBackend — Oracle now runs async worker/poller like PG
- `observation_sources` junction table replaces dialect-specific `source_memory_ids` array queries
- Automatic list partitioning on `memory_units(bank_id)` for Oracle partition pruning

## What works on Oracle

- **Retain** (single-item, batch sync, and async/background) — full fact extraction, entity resolution, chunking, embeddings
- **Recall** — all 4 retrieval strategies: semantic (HNSW), BM25 (Oracle Text CONTAINS/SCORE), graph (link expansion), temporal
- **Reflect** — disposition-aware reasoning using recalled memories
- **Bank/memory/document CRUD** — all endpoints
- **Mental models** — creation and async background refresh
- **Consolidation** — background observation synthesis via async worker
- **MCP** — all tools functional
- **Entities** — extraction, linking, cooccurrences, fuzzy resolution (JSON_TABLE + UTL_MATCH)
- **Worker/Poller** — async task claiming, execution, retry, and recovery
- **Batch API** — async batch retain with worker processing

## Benchmark results (2,000 items, cross-encoder reranker, 10 iterations)

| Metric | PostgreSQL (pg0 native) | Oracle 23ai (Docker) | Ratio |
|--------|------------------------|---------------------|-------|
| Mean | 0.311s | 0.408s | 1.31x |
| p50 | 0.254s | 0.356s | 1.40x |
| p95 | 0.614s | 0.743s | 1.21x |

**Note:** PG runs against pg0 (native/embedded) while Oracle runs in Docker, so the gap includes Docker virtualization overhead. CPU-bound phases (reranking, embedding) are identical across both backends. An apples-to-apples comparison (both in Docker or both native) would narrow the difference.

## Resolved limitations (all items from original review)

All 5 items from the original "Known Limitations" section have been resolved:

1. **Worker poller decoupled from asyncpg** — `WorkerPoller` now takes `DatabaseBackend` instead of `asyncpg.Pool`. All `pool.execute`/`pool.fetch` calls replaced with `backend.acquire()` + connection methods. PG-specific `COUNT(*) FILTER` replaced with portable `SUM(CASE WHEN ...)`. Oracle uses `BrokerTaskBackend` for full async worker support.

2. **Entity resolver fuzzy matching** — `_resolve_entities_batch_oracle_fuzzy` uses `JSON_TABLE` + `UTL_MATCH.JARO_WINKLER_SIMILARITY` for native Oracle fuzzy entity resolution.

3. **source_memory_ids array queries** — New `observation_sources` junction table (PG alembic migration + Oracle DDL) replaces all dialect-specific array operators. PG's `unnest()`/`&&` and Oracle's `JSON_TABLE` queries both replaced with standard SQL joins. Dual-write to both `source_memory_ids` column and junction table during transition.

4. **Per-bank vector index isolation** — Oracle `memory_units` table uses `PARTITION BY LIST (bank_id) AUTOMATIC` with `ORGANIZATION NEIGHBOR PARTITIONS` on the HNSW index. Oracle creates partitions per bank on INSERT and the optimizer prunes partitions on bank_id-scoped queries.

5. **Bulk insert optimization** — Oracle uses `executemany` batch insert in `bulk_insert_from_arrays()`.

## Remaining work (follow-up PRs)

- **Multi-tenant schema isolation for Oracle** — PG uses fully-qualified table names (`"schema"."table"`) via `fq_table()`. Oracle would need `ALTER SESSION SET CURRENT_SCHEMA` or equivalent. Not blocking any single-tenant Oracle deployment.
- **`fq_table()` / `_is_oracle()` cleanup** — 257 call sites still use backend checks in business logic. Too large for this PR, separate cleanup.

## Key design decisions

- **Escape hatch**: `self._pool = backend.get_pool()` keeps all existing PG consumers working unchanged during transition
- **No behavior change for PG**: Pure refactor for PostgreSQL users — all existing SQL untouched
- **Lazy oracledb import**: Oracle backend only imports `python-oracledb` when selected, no hard dependency
- **Dialect pattern**: `build_semantic_arm()` / `build_bm25_arm()` / `prepare_bm25_text()` build complete UNION ALL subquery arms per database, eliminating all inline backend checks from retrieval.py
- **Junction table over JSON**: `observation_sources` normalizes the source-observation relationship into standard SQL, eliminating the last major source of dialect-specific query rewrites
- **Automatic partitioning**: Oracle's `PARTITION BY LIST AUTOMATIC` provides per-bank isolation without application-managed DDL — partitions are created on first INSERT per bank_id
- **Text index compatibility**: CTXSYS.CONTEXT domain indexes don't support LOCAL on LIST partitioned tables (ORA-29845). Global text index works fine when created after partitioning. Migration drops and recreates text index during partition conversion for existing installs.

## Test plan

- [x] Full PG test suite: 1805 passed, 82 skipped, 0 failed
- [x] Full Oracle integration suite: 60 passed, 0 failed
- [x] Lint passes (`ruff check`, `ruff format`)
- [x] Oracle end-to-end: retain → consolidation → recall → reflect → mental models on Oracle 23ai Docker
- [x] Worker/poller tests updated for backend abstraction
- [x] Observation invalidation tests updated for junction table
- [x] Manual verification: partitioning active, all indexes VALID (including text + vector)


hindsight-api-slim/hindsight_api/engine/search/retrieval.py
```diff
@@ -145,50 +145,73 @@ async def retrieve_semantic_bm25_combined(
...
-            f" WHERE bank_id = $2"
-            f"   AND fact_type = '{ft}'"
-            f"   AND embedding IS NOT NULL"
-            f"   AND (1 - (embedding <=> $1::vector)) >= 0.3"
-            f"   {tags_clause}"
-            f"   {groups_clause}"
-            f" ORDER BY embedding <=> $1::vector"
-            f" LIMIT {hnsw_fetch})"
-        )
+        if not _is_pg:
```
nicoloboschi: we should move this query code to a function of the abstraction and every impl has his own syntax. 

all this if not _is_pg: are going to be a mess to maintain

DK09876:
Superseded by #1307 on rebased branch `feature/database-abstraction-clean`.

PR #954, opened 2026-04-09
fix(retain): run _ann_seeds temp table inside a transaction

cdbartholomew:
## Summary

\`compute_semantic_links_ann\` (part of the 3-phase retain pipeline added in #722) creates a PostgreSQL temp table \`_ann_seeds\` **outside any transaction**, then runs a TRUNCATE / COPY / SELECT / DROP sequence as individual statements on the same asyncpg connection.

This works against a direct Postgres connection, but **fails intermittently when the caller is routed through PgBouncer in \`transaction\` pool mode** (which is the standard Hindsight Cloud deployment topology):

\`\`\`
CREATE TEMP TABLE IF NOT EXISTS _ann_seeds (...)   -- backend A
TRUNCATE _ann_seeds                                 -- backend B -> FAILS
\`\`\`

PostgreSQL temp tables are **session-scoped to the backend that created them**. In PgBouncer transaction mode the backend is only pinned to the client for the duration of an actual transaction. Between standalone statements the pooler can (and under concurrency will) rebind the client to a different backend. When that happens the \`_ann_seeds\` table disappears and the follow-up statement fails with:

\`\`\`
relation \"_ann_seeds\" does not exist
\`\`\`

## Observed symptom

~3% of sync retain calls (2 of 61) failed the Hindsight Cloud smoke test on a recent hindsight-dev deploy. Async retains were largely masked by the 3-attempt retry loop — each attempt is an independent bernoulli trial, so even at 3% per-attempt the odds of all 4 attempts failing are roughly one in a million.

Recent hindsight-dev smoke-test log excerpt:

\`\`\`
Recent Errors:
  retain failed: \"relation \\\"_ann_seeds\\\" does not exist\": 4
Failed:             2
ERROR: 2 requests failed out of 61
\`\`\`

## Fix

Wrap the \`CREATE TEMP TABLE\` → \`COPY\` → \`SELECT\` sequence in a single \`async with conn.transaction():\` block, and use \`ON COMMIT DROP\` so the temp table is transaction-scoped and auto-cleaned at commit. Also switch \`SET hnsw.ef_search = 60\` to \`SET LOCAL\` so the tuning is transaction-scoped and no longer leaks onto the pooled backend for subsequent recall queries.

Drop the now-unnecessary manual \`TRUNCATE\`, explicit \`DROP TABLE\`, and \`RESET hnsw.ef_search\`. The transaction-end handles all three.

The function docstring still correctly describes this as running on a separate connection outside the surrounding **write** transaction — this change only adds an inner transaction around the ANN work itself to keep the temp table visible to PgBouncer.

### Before
\`\`\`python
await conn.execute(\"SET hnsw.ef_search = 60\")
await conn.execute(\"CREATE TEMP TABLE IF NOT EXISTS _ann_seeds (...)\")
await conn.execute(\"TRUNCATE _ann_seeds\")
await conn.copy_records_to_table(\"_ann_seeds\", records=...)
for fact_type in active_types:
    rows = await conn.fetch(...)  # SELECT FROM _ann_seeds
await conn.execute(\"DROP TABLE IF EXISTS _ann_seeds\")
await conn.execute(\"RESET hnsw.ef_search\")
\`\`\`

### After
\`\`\`python
async with conn.transaction():
    await conn.execute(\"SET LOCAL hnsw.ef_search = 60\")
    await conn.execute(
        \"CREATE TEMP TABLE _ann_seeds (...) ON COMMIT DROP\"
    )
    await conn.copy_records_to_table(\"_ann_seeds\", records=...)
    for fact_type in active_types:
        rows = await conn.fetch(...)  # SELECT FROM _ann_seeds
# Transaction commits here — ON COMMIT DROP removes _ann_seeds,
# SET LOCAL reverts hnsw.ef_search.
\`\`\`

## Who is affected

Every code path that ultimately calls \`orchestrator.retain_batch\` → \`compute_semantic_links_ann\`:

- \`POST /v1/.../memories\` (sync retain) — most user-visible symptom, no retry
- \`POST /v1/.../memories\` with \`async=true\` (batch_retain) — masked by worker retry loop
- File upload → conversion → retain — masked by worker retry loop

Not affected: recall, reflect, mental model get/refresh, consolidation, and all bank/org management endpoints. Consolidation inserts memory_units directly via raw SQL and does not go through the retain link pipeline.

## Test plan

- [x] 5 new regression tests in \`TestComputeSemanticLinksAnnPgBouncerSafety\` that exercise the function with a mocked asyncpg connection and structurally verify:
  - The function enters \`conn.transaction()\` exactly once
  - The \`CREATE TEMP TABLE\` statement uses \`ON COMMIT DROP\` (and does **not** use \`IF NOT EXISTS\`)
  - No manual \`TRUNCATE _ann_seeds\` or explicit \`DROP TABLE\` statements
  - \`hnsw.ef_search\` is set with \`SET LOCAL\`, not \`SET\`, and no \`RESET\` is issued
  - Empty inputs short-circuit without entering a transaction
- [x] All 39 existing \`test_link_utils.py\` tests still pass
- [ ] CI passes on this PR
- [ ] Once merged, bump the \`hindsight-core\` submodule in \`hindsight-deployment\` and verify the hindsight-dev smoke test goes green

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
@@ -739,16 +739,9 @@ async def compute_semantic_links_ann(
         return []
 
     import time as time_mod
-    import uuid as uuid_mod
 
     ann_start = time_mod.time()
     links = []
-
-    # Lower ef_search for retain ANN — default 400 is tuned for recall precision
-    # but at 164k units each HNSW probe takes 94ms. ef_search=60 gives 2.7ms/probe
-    # (35x faster) with sufficient accuracy for top-50 semantic link creation.
-    # Reset after to avoid polluting the connection pool for recall queries.
-    await conn.execute("SET hnsw.ef_search = 60")
 
     logger.debug(f"[ANN] Starting: {len(unit_ids)} seeds, top_k={top_k}")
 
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM

PR #956, opened 2026-04-10
fix: add timeout to cleanup lock acquisition to prevent Ctrl+C hang

### LINKED ISSUE #952: Ctrl+C hangs during shutdown, double-press causes unhandled KeyboardInterrupt
### Bug Description

## Description                                                                                                                                                                                                    
                                                                                                                                                                                                                     
   When interrupting a process (Ctrl+C) while `HindsightEmbedded` is active, the shutdown sequence hangs indefinitely because `_cleanup()` blocks on `self._lock` without a timeout or interrupt handling.           
                                                                                                                                                                                                                     
   ## Traceback                                                                                                                                                                                                      
                                                                                                                                                                                                                     
   File "hindsight/embedded.py", line 864, in shutdown                                                                                                                                                               
       self._client.close()                                                                                                                                                                                          
     File "hindsight/embedded.py", line 225, in close                                                                                                                                                                
       self._cleanup(stop_daemon_on_close=stop_daemon)                                                                                                                                                               
     File "hindsight/embedded.py", line 193, in _cleanup                                                                                                                                                             
       with self._lock:                                                                                                                                                                                              
   KeyboardInterrupt                                                                                                                                                                                                 
                                                                                                                                                                                                                     
   ## Root Cause                                                                                                                                                                                                     
                                                                                                                                                                                                                     
   In `hindsight-all/hindsight/embedded.py`, the `_cleanup` method acquires `self._lock` using a bare `with` statement. If the daemon thread is mid-operation (holding the lock) when SIGINT fires, the shutdown     
   path tries to acquire the same lock, blocks indefinitely, and the process becomes unresponsive. A second Ctrl+C forces a hard kill.                                                                               
                                                                                                                                                                                                                     
   ## Expected Behavior                                                                                                                                                                                              
                                                                                                                                                                                                                     
   Shutdown should complete gracefully even when interrupted. The lock acquisition should either:                                                                                                                    
   1. Use a timeout: `self._lock.acquire(timeout=5.0)`                                                                                                                                                               
   2. Catch `KeyboardInterrupt` during cleanup                                                                                                                                                                       
   3. Use a reentrant lock (`threading.RLock`) so the shutdown path can re-enter                                                                                                                                     
                                                                                                                                                                                                                     
   ## Environment                                                                                                                                                                                                    
                                                                                                                                                                                                                     
   - hindsight-all: 0.5.0                                                                                                                                                                                            
   - Python: 3.11                                                                                                                                                                                                    
   - OS: Linux                                                                                                                                                                                                       
                                                                                                                                                                                                                     
   ## Impact                                                                                                                                                                                                         
                                                                                                                                                                                                                     
   No data corruption (pg0 handles hard kills safely), but poor UX -- process appears to hang on exit, requires force kill.     

### Steps to Reproduce

1. Start a CLI chat session via `hermes chat`                                                                                                                                                                     
2. Press Ctrl+C once — session appears to hang indefinitely                                                                                                                                                       
3. Press Ctrl+C again — process crashes with unhandled KeyboardInterrupt traceback

### Expected Behavior

First Ctrl+C triggers a graceful shutdown with a reasonable timeout (~5-10s). If cleanup doesn't finish in time, force-exit cleanly. Second Ctrl+C should terminate immediately without traceback.

### Actual Behavior

First Ctrl+C hangs during hindsight memory provider shutdown (stuck acquiring a lock in `hindsight.embedded._cleanup()`). Second Ctrl+C crashes the process with an unhandled KeyboardInterrupt from the lock acquisition line.  

### Version

0.5.0

### LLM Provider

Other

octo-patch:
Fixes #952

## Problem

When pressing Ctrl+C while `HindsightEmbedded` is active, the shutdown sequence hangs indefinitely because `_cleanup()` attempts to acquire `self._lock` using a bare `with self._lock:` statement. If the daemon thread is holding the lock when SIGINT fires, the cleanup path blocks forever. A second Ctrl+C then crashes the process with an unhandled `KeyboardInterrupt` from inside the lock acquisition.

## Solution

Replace the bare `with self._lock:` with an explicit `self._lock.acquire(timeout=5.0)` call:

- If the lock cannot be acquired within 5 seconds (e.g. daemon thread is mid-operation during shutdown), mark the client as closed and return gracefully. The daemon will stop on its own via the idle timeout — no data corruption occurs.
- Wrap the cleanup body in `try/except KeyboardInterrupt` so that a second Ctrl+C during cleanup also exits cleanly instead of producing an unhandled traceback.
- Always release the lock in a `finally` block.

## Testing

The fix was verified by manual inspection against the traceback in the issue. The logic is straightforward: `threading.Lock.acquire(timeout=N)` returns `False` instead of blocking indefinitely when the timeout elapses, and `KeyboardInterrupt` is now explicitly caught during the cleanup body.

### DIFF
--- a/hindsight-all/hindsight/embedded.py
+++ b/hindsight-all/hindsight/embedded.py
@@ -190,7 +190,18 @@ class HindsightEmbedded:
         if self._closed:
             return
 
-        with self._lock:
+        acquired = self._lock.acquire(timeout=5.0)
+        if not acquired:
+            # Could not acquire lock within timeout (e.g. daemon thread is mid-operation).
+            # Mark as closed so future calls are rejected; daemon will stop via idle timeout.
+            logger.warning(
+                f"Could not acquire cleanup lock for profile '{self.profile}' within 5s, "
+                "forcing closed state. Daemon will stop via idle timeout."
+            )
+            self._closed = True
+            return
+
+        try:
             if self._closed:
                 return
 
@@ -209,6 +220,14 @@ class HindsightEmbedded:
                 self._manager.stop(self.profile)
 
             self._closed = True
+        except KeyboardInterrupt:
+            self._closed = True
+            logger.warning(
+                f"Cleanup interrupted for profile '{self.profile}', "
+                "forcing closed state. Daemon will stop via idle timeout."
+            )
+        finally:
+            self._lock.release()
 
     def close(self, stop_daemon: bool = False):
         """
--- a/hindsight-integrations/opencode/src/hooks.test.ts
+++ b/hindsight-integrations/opencode/src/hooks.test.ts
@@ -20,7 +20,7 @@ function makeClient() {
     } as any;
 }
 
-function makeOpencodeClient(messages: Array<{ role: string; parts: Array<{ type: string; text?: string }> }> = []) {
+function makeOpencodeClient(messages: Array<{ info: { role: string }; parts: Array<{ type: string; text?: string }> }> = []) {
     return {
         session: {
             messages: vi.fn().mockResolvedValue({ data: messages }),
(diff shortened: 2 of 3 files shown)


nicoloboschi:
this pR has unrelated changes

PR #960, opened 2026-04-10
RFC: self-improving skills via hindsight-skill-builder meta-skill

nicoloboschi:
## Summary

- Adds `skills/hindsight-skill-builder/SKILL.md` — a conversational meta-skill that builds Hindsight-backed self-improving skills through a 4-step approval-gated flow. Lives next to the existing `hindsight-local` / `hindsight-cloud` / etc siblings and installs the same way.
- Adds two end-to-end demos under `hindsight-dev/demos/` proving the pattern works on stock Hindsight with no schema changes and no Git integration.

## The pattern

A "self-improving skill" is a thin `SKILL.md` whose body is `curl` + fallback-to-asking. The `curl` fetches live instructions from a Hindsight mental model, which auto-refreshes after each consolidation cycle. The skill file itself never changes — the mental model content evolves from user feedback during normal conversation, via the existing retain → consolidate → refresh pipeline.

The meta-skill walks the agent through the setup conversationally with explicit user approval at every decision (Hindsight URL, bank name, missions, mental model spec, target path).

## Two demos included

- `hindsight-dev/demos/hermes-marketing/` — V1, hand-crafted bank + missions + manual API calls, to prove the pattern mechanically.
- `hindsight-dev/demos/hermes-marketing-v2/` — V2, same end-to-end run but setup happens entirely through the new meta-skill. One user sentence + four approvals = full skill configured, then 7 turns of marketing-post feedback evolves the mental model from empty to a ~5000-char structured style guide. The `SKILL.md` file's MD5 is identical at start and end of the run — only the curled mental model content changes.

## Known bugs the demos hit (not fixed in this PR)

Documented in each demo's README, ranked by severity:

1. **`hindsight_api/config.py:18` — `load_dotenv(find_dotenv(usecwd=True), override=True)`** walks upward from the daemon's cwd and silently overrides explicit env vars with whatever `.env` file it finds. Any sibling worktree with a dev `.env` breaks daemon startup. Cost ~30min to diagnose. Fix: use `override=False` or scope the search to `~/.hindsight/`.
2. **`consolidator.py:_trigger_mental_model_refreshes` tag gating** — mental models with tags never auto-refresh when consolidated memories are untagged (the default for hermes auto-retain). The failure is silent. Fix: at minimum log a warning when an `auto-refresh` MM is skipped due to tag mismatch; better, reconsider the security model.
3. **hermes plugin `bank_mission` / `bank_retain_mission` dead fields** — the plugin reads these from `~/.hermes/hindsight/config.json` but never sends them to the Hindsight API. Fix: call `update_bank_config` on init.
4. **Port confusion** — `hindsight-docs/docs-integrations/hermes.md:107` says `apiPort: 9077`, the plugin source defaults to `8888`, the hermes profile registers as `:9177`. Three numbers in three places.

## Test plan

- [ ] Pull this branch, review `skills/hindsight-skill-builder/SKILL.md`
- [ ] Read `hindsight-dev/demos/hermes-marketing-v2/README.md` for the full walkthrough
- [ ] Reproduce the V2 demo locally by installing the skill into a harness (instructions in the comment below)
- [ ] Review agent-authored missions/source_query quality against the V1 hand-crafted versions
- [ ] Flag any rough edges in the meta-skill instructions, especially the approval-protocol tone

## Not in scope for this PR

- Push-model alternative (Hindsight rewriting `SKILL.md` directly instead of curl at runtime) — follow-up RFC
- Skill registry / catalog
- Fixes for the four bugs above
- Harnesses other than hermes (pattern should work for claude-code/codex/cursor; untested)


fabioscarsi:
Smart! Interesting!

PR #967, opened 2026-04-10
feat(openclaw): expose Hindsight docs as memory bridge artifacts

### LINKED ISSUE #963: OpenClaw memory-wiki bridge imports 0 artifacts with hindsight-openclaw
## Summary

`@vectorize-io/hindsight-openclaw` works as an active OpenClaw memory plugin for retain/recall, but it does not currently export OpenClaw `publicArtifacts` for `memory-wiki` bridge mode.

## Observed behavior

When `hindsight-openclaw` is the active memory plugin and OpenClaw `memory-wiki` is set to `vaultMode: "bridge"`:

- `openclaw wiki bridge import` imports `0` artifacts
- `openclaw wiki status` warns that the active memory plugin is not exporting any public memory artifacts yet

## Expected behavior

Either:

1. `hindsight-openclaw` exports bridge-importable `publicArtifacts`, or
2. the docs clearly state that OpenClaw `memory-wiki` bridge mode is not supported yet

## Notes

This is separate from Hindsight's own shared-bank / external API setup. Shared Hindsight banks work today. The gap is specifically OpenClaw `memory-wiki` bridge import from the Hindsight plugin.

frvade:
Closes #963.

## Summary
- register OpenClaw `publicArtifacts` support in the Hindsight OpenClaw plugin
- fetch Hindsight banks/documents over the existing external HTTP API and materialize them as markdown under `memory/hindsight-bridge/`
- add one focused test that covers the document export path

## Verification
- `cd hindsight-integrations/openclaw && npm test`


nicoloboschi (commented):
Code review notes:

**Must fix**
- **Missing pagination** in `listHindsightPublicArtifacts` (`src/index.ts`): documents query is hardcoded to `?limit=100&offset=0`. Banks with >100 documents will silently drop artifacts. Loop until the API returns fewer than `limit` items.
- **No error isolation in the fetch loop**: if a single document GET fails, the whole call throws — and we may already have written a partial set of files to the user's workspace. Wrap per-document fetches in try/catch, log, and continue so one bad doc doesn't nuke the export.

**Should fix**
- **N+1 / sequential I/O**: for each bank we list, then for each doc we fetch individually, fully sequentially. Parallelize with a bounded concurrency (e.g. p-limit) per bank — this will be painfully slow on real workspaces.
- **`kind: 'daily-note'` is misleading**: these aren't daily notes, they're Hindsight documents. If OpenClaw's bridge requires this literal string, add a comment explaining why; otherwise use a more accurate kind.
- **Locally redeclared SDK contract** (`src/types.ts`): `MemoryPluginPublicArtifact` and `MemoryPluginCapability` are hand-rolled here instead of imported from the OpenClaw SDK. If OpenClaw changes the shape this plugin drifts silently. Import the real type or add a comment acknowledging the duplication.
- **Sync FS inside async function**: `mkdirSync` / `writeFileSync` block the event loop while materializing many artifacts. Prefer `fs/promises`.
- **Test coverage is thin**: one happy-path test (1 bank, 1 doc, default workspace). Missing: pagination boundary, fetch failure, empty bank list, runtime without `registerMemoryCapability`, multi-workspace (`agents.list` with multiple entries). For code that writes to the user's filesystem this is under-covered.
- **Unrelated comment churn**: `// Default: true (on) — backward compatible` → `// Default: true (on), backward compatible` is noise, please drop.

**Notes**
- `listArtifacts` sounds read-only but materializes markdown files on disk. Mirrors the OpenClaw contract, but worth a one-line comment on the exported function so future readers aren't surprised.
- `original_text` is written verbatim with no size cap — fine for v1, but flag-worthy for very large documents.

Overall direction looks right and scoped to #963. The two must-fix items (silent 100-doc cap + all-or-nothing error handling during a filesystem-mutating call) will bite real users before tests will — I'd block on those; the rest is polish.

frvade:
Addressed all review items from #pullrequestreview-4097275838 across three commits (`d3f03e7`, `2b307b5`, `50631cb`):

**Must fix** — both resolved in `d3f03e7`:
- Pagination: loop until `items.length < pageSize`, no more 100-doc cap
- Error isolation: failed per-document fetches are logged and skipped

**Should fix** — resolved in `2b307b5` + `50631cb`:
- **N+1 sequential I/O** → batched with `Promise.allSettled`, concurrency 8
- **`kind: 'daily-note'`** → added inline comment explaining it's the closest OpenClaw bridge kind
- **Locally redeclared SDK types** → added comment + TODO to replace once OpenClaw ships a types package
- **Sync FS** → switched to `fs/promises` (`mkdir`/`writeFile`)
- **Test coverage** → added: pagination boundary, fetch failure + continue, empty bank list, multi-workspace materialization (5 tests total)
- **Unrelated comment churn** → reverted (em-dash restored)

**Notes** — resolved in `50631cb`:
- Added JSDoc on `listHindsightPublicArtifacts` clarifying it materializes files despite the "list" name
- Added note that `original_text` is written with no size cap (acceptable for v1)

Also reverted an accidental `package-lock.json` drift (local→npm dep resolution) that was unrelated to this feature (`2b307b5`).

frvade:
@nicoloboschi — merged the latest main into this branch, conflicts are resolved, and targeted tests pass. Could you take another look when you get a chance? Thanks!

frvade:
Merged latest main into this branch, resolved the OpenClaw integration conflict, and pushed the updated branch.

@nicoloboschi could you please take another look when you have a moment?

benfrank241:
Hey there could you take a look at the merge conflicts, thanks!

PR #969, opened 2026-04-10
feat(opencode): add recallTags and recallTagsMatch config options

benfrank241:
## Summary

- Adds `recallTags` (`string[]`) and `recallTagsMatch` (`any | all | any_strict | all_strict`) config options to the OpenCode plugin
- These are applied to all recall calls: auto-recall on session start, compaction recall, and the explicit `hindsight_recall` tool
- When `recallTags` is empty (default), no tag filter is sent — existing behavior is unchanged
- `HINDSIGHT_RECALL_TAGS` (comma-separated) and `HINDSIGHT_RECALL_TAGS_MATCH` env vars also supported
- Docs updated with new config options and env vars

## Use case

Enables shared-bank setups with per-user tag filtering:

**Retain** (tag memories per user):
```json
{ "retainTags": ["user:alice"] }
```

**Recall** (filter to user's memories + shared memories):
```json
{ "recallTags": ["user:alice", "shared"], "recallTagsMatch": "any_strict" }
```

`any_strict` returns memories that match at least one tag, excluding untagged memories entirely — so each user only sees their own memories and any explicitly shared ones.

### DIFF
--- a/hindsight-docs/docs-integrations/opencode.md
+++ b/hindsight-docs/docs-integrations/opencode.md
@@ -76,6 +76,9 @@ This ensures memories survive context window trimming.
       "autoRecall": true,
       "autoRetain": true,
       "recallBudget": "mid",
+      "recallTags": [],
+      "recallTagsMatch": "any",
+      "retainTags": [],
       "retainEveryNTurns": 10,
       "debug": false
     }]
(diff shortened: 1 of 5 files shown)


nicoloboschi (approved):
LGTM

PR #972, opened 2026-04-10
fix(billing): mark reflect's internal recall calls as internal

cdbartholomew:
## Summary

Reflect's tool functions (`tool_search_observations`, `tool_recall`) pass the caller's `request_context` directly to `recall_async` without setting `internal=True`. This causes the metering extension to record and bill these sub-recalls as standalone user-facing recall operations, double-charging for work that is already part of the reflect operation.

## Fix

Wrap `request_context` with `dataclasses.replace(request_context, internal=True)` at both call sites. This matches the pattern consolidation already uses for its internal sub-operations.

With the `internal` flag set, the metering extension:
- Records usage as `internal_recall` (tracked for analytics, not billed)
- Skips credit deduction

## Test plan

- [ ] CI passes
- [ ] After deploy: verify metering logs show `op=internal_recall_complete` for reflect's sub-recalls, not `op=recall_complete`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/tools.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/tools.py
@@ -9,6 +9,7 @@ Implements hierarchical retrieval:
 
 import logging
 import uuid
+from dataclasses import replace
 from datetime import datetime, timezone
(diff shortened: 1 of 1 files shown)


nicoloboschi (approved):
LGTM

Issue #977, opened 2026-04-10
retain re-submission under same document_id fails with UniqueViolationError on pk_chunks


akhater:
## Summary

Calling `retain` / `batch_retain` more than once for the same `document_id` (the documented pattern for incrementally adding to an existing document) fails with a Postgres `UniqueViolationError` on `pk_chunks`. The worker retries the task several times before giving up and marking it failed, wasting worker capacity each time — but the primary issue is that the collision happens at all: the retain path does not appear to upsert / replace chunks for an existing `document_id`, it tries to insert fresh chunks that then collide with the chunks persisted from the first call.

The practical effect for clients that re-submit the same document as it grows (e.g. a cumulative conversation transcript keyed by a stable session id) is that the **first** submission lands successfully and every subsequent one fails. From the outside this looks like memory ingestion silently stops after the first batch for that session.

## Environment

- Image: `ghcr.io/vectorize-io/hindsight:latest-slim`
- API version: `0.5.0` (`GET /version`)
- Features reported: `observations: true, mcp: true, worker: true, bank_config_api: true, file_upload_api: true`
- Database: Postgres (asyncpg)
- Health endpoint: `{"status":"healthy","database":"connected"}`

## Reproduction

1. Pick any `bank_id` (`{BANK}`).
2. First call — lands successfully, creates document `{DOC}` with chunk ids `{BANK}_{DOC}_0`, `{BANK}_{DOC}_1`, …:
   ```
   POST /v1/default/banks/{BANK}/retain/batch
   {
     "items": [{"content": "<turns 1..N as one blob>", "context": "..."}],
     "document_id": "{DOC}",
     "retain_async": true
   }
   ```
3. Some turns later, re-submit under the **same** `document_id`, with the accumulated content (turns 1..N+M):
   ```
   POST /v1/default/banks/{BANK}/retain/batch
   {
     "items": [{"content": "<turns 1..N+M as one blob>", "context": "..."}],
     "document_id": "{DOC}",
     "retain_async": true
   }
   ```
4. The `retain` subtask created by the worker fails:
   ```
   asyncpg.exceptions.UniqueViolationError: duplicate key value violates unique constraint "pk_chunks"
   DETAIL:  Key (chunk_id)=({BANK}_{DOC}_0) already exists.
   ```
5. The poller reschedules the task and retries. Each retry hits the same deterministic error (the row really does exist; this is not a transient condition). After ~9 retries across ~9 minutes the task is finally marked failed.

[...]

PR #981, opened 2026-04-10
fix(worker): classify IntegrityConstraintViolationError as non-retryable

### LINKED ISSUE #980: worker: classify IntegrityConstraintViolationError as non-retryable
## Summary

When a task hits a deterministic database constraint violation (e.g., `asyncpg.UniqueViolationError` on `pk_chunks` during retain re-submission, as reported in #977), the worker retries it up to 3 times with 60-second delays before marking it as failed. These retries are wasted — constraint violations are deterministic and will never succeed on retry.

## Current behavior

In `memory_engine.py`'s `execute_task`, the exception handler classifies only `file_convert_retain` tasks as non-retryable. All other task types (including `retain`, `batch_retain`, `consolidation`) retry up to 3 times via `RetryTaskAt`:

```python
if task_type == "file_convert_retain":
    # Non-retryable
    ...
else:
    retry_count = task_dict.get("_retry_count", 0)
    if retry_count < 3:
        raise RetryTaskAt(...)
```

## Proposed fix

Add an `isinstance` check for `asyncpg.IntegrityConstraintViolationError` (parent of `UniqueViolationError`, `ForeignKeyViolationError`, `CheckViolationError`, `NotNullViolationError`, `ExclusionViolationError`) before the retry logic. Mark these as failed immediately, freeing worker capacity.

This is the secondary issue from #977 — the primary issue (upsert semantics for document_id) is a separate design discussion.

## Impact

- Saves ~3 minutes of wasted worker capacity per deterministic failure
- Reduces noise in logs (no more repeated identical error messages)
- Affects any constraint violation across all task types

r266-tech:
## Summary

Constraint violations (`UniqueViolationError`, `ForeignKeyViolationError`, `CheckViolationError`, `NotNullViolationError`, `ExclusionViolationError`) are deterministic — retrying will never resolve them. Currently these exceptions fall through to the generic retry path in `execute_task`, wasting worker capacity with up to 3 retries at 60-second intervals before the task is finally marked as failed.

This PR adds an `isinstance` check for `asyncpg.IntegrityConstraintViolationError` (the parent class of all constraint violation exceptions) before the retry logic. Matching exceptions are marked as failed immediately via `_mark_operation_failed`, following the same pattern as the existing `file_convert_retain` non-retryable path.

## Motivation

Reported as the secondary issue in #977: when `retain` is called with an existing `document_id`, the worker hits `UniqueViolationError` on `pk_chunks` and retries ~3 times over ~3 minutes before giving up. Each retry is wasted work that also blocks the worker slot.

The primary issue in #977 (upsert semantics for `document_id`) is a separate design discussion. This PR addresses only the retry classification.

## Changes

- `hindsight-api-slim/hindsight_api/engine/memory_engine.py`: Add `elif isinstance(e, asyncpg.IntegrityConstraintViolationError)` branch in `execute_task`'s exception handler, between the existing `file_convert_retain` check and the generic retry path (+8 lines)

## Test plan

- [ ] Existing `test_executor_exception_marks_failed_immediately` continues to pass (same pattern)
- [ ] Existing `test_executor_exception_triggers_retry` continues to pass (non-constraint errors still retry)
- [ ] Manual: trigger a `UniqueViolationError` via duplicate `document_id` retain — task should fail immediately instead of retrying 3 times

Fixes #980

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -1140,6 +1140,16 @@ class MemoryEngine(MemoryEngineInterface):
                     logger.error(f"Not retrying task {task_type} (non-retryable), marking as failed")
                     if operation_id:
                         await self._mark_operation_failed(operation_id, str(e), error_traceback)
+                elif isinstance(e, asyncpg.IntegrityConstraintViolationError):
+                    # Non-retryable: constraint violations are deterministic and won't
+                    # succeed on retry (e.g. UniqueViolationError on pk_chunks when
+                    # re-submitting a retain with an existing document_id, see #977).
+                    logger.error(
+                        f"Not retrying task {task_type} (deterministic constraint violation: "
+                        f"{type(e).__name__}), marking as failed"
+                    )
+                    if operation_id:
+                        await self._mark_operation_failed(operation_id, str(e), error_traceback)
                 else:
                     if task_type == "consolidation" and operation_id:
                         # Fire failure webhook (non-transactional — operation not yet marked failed;


nicoloboschi:
thanks, but already fixed by https://github.com/vectorize-io/hindsight/pull/986

PR #984, opened 2026-04-11
fix(mental-models): surface warning when tag filtering empties fact pool

### LINKED ISSUE #945: Mental model tags act as hard filter on source facts, silently returns empty
Ran into surprising behavior while tightening up mental model definitions. Spent a couple hours debugging what I thought was a reflect regression before realizing it was working as designed but not documented.

**What happened**:

I updated 4 existing mental models with sharper \`source_query\` strings and added \`tags\` to each for organization. Example:

\`\`\`json
{
  \"name\": \"Key Decisions Made\",
  \"source_query\": \"technical decisions architectural choices tool selection...\",
  \"tags\": [\"type:decision\", \"core\"]
}
\`\`\`

After the next refresh, 3 of the 4 mental models returned \`\"I don't have information.\"\` as their content. The 4th still worked because I'd updated it before the tags change landed.

**Debugging that threw me off**:

1. The refresh operations completed successfully (no errors in \`/operations\`)
2. Calling \`POST /reflect\` directly with the exact same \`source_query\` returned full, correct content with dozens of facts
3. Only the mental model refresh path returned empty

Eventually confirmed by A/B test: removing the tags from one mental model and refreshing produced 2400+ chars of correct content. Adding tags back produced \"I don't have information.\"

**Root cause**: the \`tags\` field on a mental model acts as a hard filter on source facts. The refresh only considers memories tagged with matching values. If no existing memories have those tags (my case, since I hadn't backfilled yet), the fact pool is empty and reflect returns the no-information message.

**What's confusing**:

1. The schema description on \`UpdateMentalModelRequest.tags\` says \"Tags for scoped visibility\" which sounds like metadata/organization, not a filter on source data.
2. The trigger schema also has \`tag_groups\` and \`tags_match\` which look like the intended filtering mechanism. Having both creates ambiguity about which layer actually filters.
3. When zero facts match, the generic \"I don't have information\" message gives no hint that the issue is tag filtering rather than a bad query or missing data.

**Suggested fixes** (pick any subset):

1. **Docs**: clarify in the mental model docs that \`tags\` on the model itself filters the fact pool. Call out the interaction with \`trigger.tag_groups\`.
2. **Reflect response**: when tag filtering results in zero facts, return a structured error or warning instead of the generic no-info message. Something like \`{\"warning\": \"0 facts matched tags [type:decision, core]. Remove tags or backfill memory tags.\"}\`
3. **API ergonomics**: consider splitting the concept into \`labels\` (metadata-only, for organization) and \`filter_tags\` (actually filters source facts). Right now one field does both jobs, and the label use case is the more intuitive one.

Happy to help with a docs PR if that's the cleanest path. I'm running this in production with 937 memories across one bank and would've saved myself a couple hours with a one-line warning in the docs.

octo-patch:
Fixes #945

## Problem

When a mental model has tags configured (e.g. `["user:bob"]`) and those tags do not match any memories in the bank, the refresh operation calls `reflect_async` which finds zero matching facts. The LLM then returns a generic *"I don't have information about this"* response — indistinguishable from a bad `source_query` or a genuinely empty bank.

Users were spending hours debugging what looked like a platform bug before discovering that the culprit was their own tag configuration (see issue for a detailed first-person account).

## Solution

In `_handle_refresh_mental_model`, after the reflect completes and `based_on_serialized` is built, check whether:
1. Tag filtering was active (the model has `tags` or `tag_groups` set), **and**
2. No non-directive facts appear in `based_on` (directives don't require tag matches).

When both conditions are true:
- Replace `generated_content` with a human-readable `[Tag filter warning: ...]` message that explicitly names the active tags and tells the user what to do.
- Add a `tag_filter_warning` field to the stored `reflect_response` dict so API consumers can detect the condition programmatically without parsing the content string.
- Emit a `WARNING`-level server log entry for operator visibility.

## Testing

- Added `test_refresh_with_tags_emits_warning_when_no_facts_match` to `TestMentalModelRefreshTagSecurity`. The test retains memories tagged `user:alice`, creates a mental model tagged `user:bob` (no overlap), refreshes it, and asserts that:
  - The resulting `content` mentions the active tags or the word "tag".
  - The `reflect_response` dict contains the `tag_filter_warning` key.
  - The warning text references `user:bob`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -996,10 +996,33 @@ class MemoryEngine(MemoryEngineInterface):
                     )
             based_on_serialized[fact_type] = serialized_facts
 
-        reflect_response = {
+        # Detect empty fact pool caused by tag filtering and surface a clear warning.
+        # Without this, users see "I don't have information" with no hint that tags are
+        # the cause (reported in https://github.com/vectorize-io/hindsight/issues/945).
+        has_tag_filtering = bool(tag_filtering.tags or tag_filtering.tag_groups)
+        non_directive_facts = sum(
+            len(facts) for fact_type, facts in based_on_serialized.items() if fact_type != "directives"
+        )
+        tag_filter_warning: str | None = None
+        if has_tag_filtering and non_directive_facts == 0:
+            active_tags = tag_filtering.tags or [str(tg) for tg in (tag_filtering.tag_groups or [])]
+            tag_filter_warning = (
+                f"No memories matched the tags {active_tags!r} configured on this mental model. "
+                f"The model content could not be generated. "
+                f"Remove the tags from the model or backfill memory tags to allow facts to be retrieved."
+            )
+            logger.warning(
+                f"[REFRESH_MENTAL_MODEL_TASK] Tag filtering produced empty fact pool for "
+                f"mental_model_id={mental_model_id}, bank_id={bank_id}, tags={active_tags!r}"
+            )
+            generated_content = f"[Tag filter warning: {tag_filter_warning}]"
+
+        reflect_response: dict[str, Any] = {
             "text": reflect_result.text,
             "based_on": based_on_serialized,
         }
+        if tag_filter_warning is not None:
+            reflect_response["tag_filter_warning"] = tag_filter_warning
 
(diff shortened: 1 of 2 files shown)


nicoloboschi:
I don't think we should log anything, at most we have to add a UI note or a documentation note

PR #985, opened 2026-04-11
fix(codex): serialize hindsight daemon startup

nish3451:
## Summary
- serialize Codex-side daemon startup so concurrent hooks cannot start a second `hindsight-embed` daemon
- route SessionStart warmup through the same locked startup path used by foreground retain/recall flows
- add a regression test covering the in-flight startup race

## Root cause
A duplicate local daemon start on April 11, 2026 caused a bind conflict on port `9077`. The losing process then ran shutdown cleanup and stopped the shared embedded Postgres instance, leaving the first worker stuck polling the database with connection-refused errors.

## Verification
- `uv run pytest hindsight-integrations/codex/tests/test_hooks.py hindsight-integrations/codex/tests/test_content.py hindsight-integrations/codex/tests/test_daemon.py -q`
- `uv run ruff check hindsight-integrations/codex/scripts/lib/daemon.py hindsight-integrations/codex/tests/test_daemon.py`

Both passed locally (`72 passed`, Ruff clean).

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2421,6 +2421,7 @@ def create_app(
                 tenant_extension=memory._tenant_extension,
                 max_slots=config.worker_max_slots,
                 consolidation_max_slots=config.worker_consolidation_max_slots,
+                stale_processing_timeout_seconds=config.worker_stale_processing_timeout_seconds,
             )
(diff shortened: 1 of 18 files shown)


nicoloboschi (changes requested):
hey this PR contains another change for the worker heartbeat, can you clean this up

nicoloboschi:
already fixed here https://github.com/vectorize-io/hindsight/pull/1016

PR #987, opened 2026-04-11
fix(openclaw): stabilize session identity and skip operational turns

Tosko4:
## Summary

This tightens session identity handling in the OpenClaw integration so operational turns do not get recalled into or retained in user memory banks.

## What this changes

- expands `parseSessionKey()` to handle `agent:*:main` and operational session shapes like `agent:*:(cron|heartbeat|subagent):...`
- adds a Telegram direct fallback that derives a stable sender identity from `direct:<sender>` channel ids when `senderId` is missing
- routes bank derivation through resolved session identity instead of raw partial hook context
- adds guardrails to skip recall/retain for:
  - main sessions
  - cron / heartbeat / subagent sessions
  - temp sessions
  - missing stable provider / sender identity
  - Telegram direct identity mismatches
- skips explicit ephemeral operational text such as reset/startup boilerplate before recall or retain
- adds unit coverage for the new session parsing, Telegram direct fallback, skip reasons, and ephemeral text detection

## Why

Without these guardrails, incomplete or operational hook context can drift into user-facing memory banks, especially in direct-session flows where `senderId` is not always present on later hook phases.

## Validation

- `npm test`
- `npm run build`


nicoloboschi (commented):
Code review notes:

**Must fix**
- **Undocumented behavior change in `buildRetainRequest`**: `channel_type` used to be `effectiveCtx?.messageProvider`, now it's `provider` (= `messageProvider || parsedSession.provider`). Previously `channel_type` could be `undefined` while `provider` was set from the session key; now they're always equal. Downstream consumers that keyed on the old nullability will see a shape change. Not called out in the PR description — either document it or revert.

**Should fix**
- **Heavy duplication across four hooks**: `before_dispatch`, `before_agent_start`, `before_prompt_build`, `agent_end` each run essentially the same 20-30 line sequence (resolve → cache → skip-reason → re-cache). Extract a single `resolveAndCacheIdentity(sessionKey, ctx)` helper returning `{ resolvedCtx, skipReason }`. This is ~300 lines of near-duplicate logic that will drift when someone edits only one copy.
- **Two parallel session maps** (`senderIdBySession` + `sessionIdentityBySession`): the new map is a superset of the old one and every write site updates both. Consolidate to one map and derive `senderId` from `.senderId` on the identity record.
- **`isEphemeralOperationalText` is brittle**: hardcoded English regex prefixes like `/^This (?:script|task|job|workflow) updates .* index/i`. Small rephrasings in OpenClaw will silently bypass the filter. At minimum, comment pointing to where these strings originate in OpenClaw; ideally plumb an explicit `operational: true` flag from OpenClaw instead of matching content.
- **`canRetryIdentityResolution = skipTurnReason.startsWith('missing stable ')`** is stringly-typed magic. Rename a reason string and retry silently stops. Make the skip reason a discriminated union (`{ kind: 'retryable' | 'final', detail: string }`) so the decision is explicit.
- **Test coverage gaps** in `getIdentitySkipReason`: only "internal main session" and "telegram direct identity mismatch" are exercised. No coverage for `cron|heartbeat|subagent` operational sessions, `temp:`, missing provider, missing sender, or the retryable-vs-final distinction above.
- **Case-insensitive regexes** (`/^agent:[^:]+:(cron|heartbeat|subagent):/i` etc.): session keys are lowercase by construction. Drop the `/i` so we don't imply a tolerance that isn't guaranteed elsewhere.

**Notes**
- `setCappedMapValue` is FIFO, not LRU (`set` on an existing key doesn't reorder). Consistent with the previous inlined eviction, so not a regression — but worth a one-liner comment so a future reader doesn't assume LRU.
- `parseSessionKey` now handles three distinct shapes. The inline return-type is fine in TS, but given how central this function has become a named `ParsedSessionKey` type would document intent better than the repeated inline `{ agentId?, provider?, channel? }` at every callsite.
- 515 additions in one file for what is essentially "add an identity resolver + guardrails" — the duplication cleanup above would roughly halve that.

Overall direction looks right (tightening identity before bank derivation is the correct fix for the described drift). The main asks are the undocumented `channel_type` semantic shift and collapsing the duplicated hook bodies.

nicoloboschi (approved):
LGTM

PR #992, opened 2026-04-11
fix(worker): reclaim stale tasks from dead workers

### LINKED ISSUE #991: Stale claimed_at not reclaimed after worker crash — operations stuck in 'processing' forever
## Summary

When a `hindsight-api` worker dies without graceful shutdown, any `async_operations` row it has claimed (`status='processing'`, `worker_id=<dead>`, `claimed_at=<timestamp>`) **never gets reclaimed**. The tasks stay stuck in `processing` indefinitely, and because of per-bank serialization, **newly enqueued tasks on the same bank also pile up in `pending` and never get picked up** by live workers.

## Observed in production

Setup: self-hosted `hindsight:latest`, backed by Neon Postgres, single worker container.

1. Worker `acc76c6f2529` claimed two `consolidation` tasks for bank `cuberoid` at `2026-04-10 22:54 UTC` and subsequently died (container recreated, new hostname `f0200c40c387`).
2. A third `consolidation` task for the same bank was enqueued at `22:56 UTC` and remained in `pending`.
3. ~14 hours later, all 3 were still unresolved:

| operation_id | status       | worker_id       | claimed_at       | retry_count |
|--------------|--------------|-----------------|------------------|-------------|
| f44ba046…    | `processing` | `acc76c6f2529`  | 2026-04-10 22:54 | 0           |
| 1c5120a3…    | `processing` | `acc76c6f2529`  | 2026-04-10 22:54 | 0           |
| a3d7ab91…    | `pending`    | `NULL`          | —                | 0           |

4. Poller logs from the **live** worker showed it correctly detected the pending task but never claimed it:
   ```
   [WORKER_STATS] worker=f0200c40c387 slots=0/10 (consolidation=0/2) | available=10 (consolidation=2) | global: pending=1 (schemas: hindsight) | others: acc76c6f2529:2 | my_active: none
   ```
   The `others: acc76c6f2529:2` counter kept showing stale claims from a hostname that hadn't existed for hours.

5. `pending_consolidation` for the affected bank grew to **91 memories** while the operations were frozen.

## Root cause

There is no reclamation / lease-expiry mechanism for `async_operations` rows where:

- `status = 'processing'`
- `claimed_at < NOW() - <some_timeout>`
- the owning `worker_id` has not heartbeated recently (or never will, because it's gone)

`next_retry_at` stays `NULL`, `retry_count` stays `0`, and the per-bank concurrency limit on `consolidation` (which normally prevents parallel runs on the same bank) treats the zombie rows as "in flight," so no new work on that bank ever starts.

## Manual recovery

```sql
UPDATE hindsight.async_operations
SET status='pending', worker_id=NULL, claimed_at=NULL, updated_at=NOW()
WHERE status='processing'
  AND worker_id='<dead_worker_hostname>';
```

Within one poller tick (~30s) the live worker picked up all three tasks and consolidation completed normally.

## Suggested fix

Add a reclamation pass to the poller (or a separate periodic job) that resets stale claims:

```sql
UPDATE hindsight.async_operations
SET status='pending',
    worker_id=NULL,
    claimed_at=NULL,
    retry_count=retry_count+1,
    updated_at=NOW()
WHERE status='processing'
  AND claimed_at < NOW() - INTERVAL '15 minutes'
  -- and optionally: worker_id NOT IN (SELECT worker_id FROM alive_workers_heartbeat)
```

Ideally combined with a heartbeat mechanism (e.g. `worker_heartbeats` table updated every N seconds) so reclamation can be aggressive for truly dead workers without racing against slow-but-alive ones.

Without this, any unclean worker shutdown permanently wedges the affected bank.

## Environment

- `hindsight:latest` (image `ghcr.io/vectorize-io/hindsight:latest`, pulled 2026-04-11)
- Postgres: Neon managed
- Single worker container (`HINDSIGHT_API_HOST=0.0.0.0`, default poller config)

r266-tech:
## Summary

When a worker dies without graceful shutdown (OOM-killed, node failure, container restart with new hostname), its claimed `async_operations` rows stay in `processing` indefinitely. The existing `recover_own_tasks` only recovers tasks for the **same** `worker_id`, which doesn't help when a container restarts with a new hostname.

Because consolidation uses per-bank serialization (`NOT EXISTS ... status='processing'`), stale claims also **block all new work on affected banks** — the live worker sees slots occupied by a dead hostname and never claims pending tasks.

## Changes

- Add `reclaim_stale_tasks()` method to `WorkerPoller` that periodically resets tasks from **other** workers when `claimed_at` exceeds a configurable timeout (default 15 minutes)
- Own in-flight tasks are never touched (`worker_id IS DISTINCT FROM $2`)
- Reclamation runs at most once per 60 seconds (not every poll tick)
- New `stale_claim_timeout_minutes` constructor parameter (set to 0 to disable)
- 4 focused tests covering: stale reclaim, recent-claim safety, own-task safety, disable path

## Root cause (from #991)

```
async_operations WHERE status='processing' AND worker_id='<dead hostname>'
```
→ never reclaimed → per-bank serialization blocks new work → bank permanently wedged

Manual `UPDATE ... SET status='pending' WHERE worker_id='<dead>'` immediately unblocks the bank (confirmed by reporter).

Fixes #991

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
@@ -776,17 +776,13 @@ async def compute_semantic_links_ann(
         await conn.execute("SET LOCAL hnsw.ef_search = 60")
 
         t_setup = time_mod.time()
-        await conn.execute(
-            "CREATE TEMP TABLE _ann_seeds (unit_id text, emb_text text, fact_type text) ON COMMIT DROP"
-        )
+        await conn.execute("CREATE TEMP TABLE _ann_seeds (unit_id text, emb_text text, fact_type text) ON COMMIT DROP")
 
         records = [
             (uid, emb if isinstance(emb, str) else str(emb), ft)
             for uid, emb, ft in zip(unit_ids, embeddings, fact_types)
         ]
-        await conn.copy_records_to_table(
-            "_ann_seeds", records=records, columns=["unit_id", "emb_text", "fact_type"]
-        )
+        await conn.copy_records_to_table("_ann_seeds", records=records, columns=["unit_id", "emb_text", "fact_type"])
         logger.debug(f"[ANN] Temp table setup: {time_mod.time() - t_setup:.3f}s ({len(records)} seeds)")
 
         # Run one ANN query per fact_type so each uses the right HNSW index.
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -531,7 +531,6 @@ class WorkerPoller:
             logger.info(f"Worker {self._worker_id} recovered {total_count} stale tasks from previous run")
         return total_count
 
-
     async def reclaim_stale_tasks(self) -> int:
         """
         Reclaim tasks stuck in 'processing' from dead workers.
@@ -576,8 +575,7 @@ class WorkerPoller:
             except Exception as e:
                 schema_display = f'"{schema}"' if schema else str(schema)
                 logger.warning(
-                    f"Worker {self._worker_id} failed to reclaim stale tasks "
-                    f"for schema {schema_display}: {e}"
+                    f"Worker {self._worker_id} failed to reclaim stale tasks for schema {schema_display}: {e}"
                 )
 
         if total_reclaimed > 0:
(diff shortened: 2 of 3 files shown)


nicoloboschi (changes requested):
we don't want to implement this for now. we need to figure out why workers are stuck

PR #993, opened 2026-04-11
fix(opencode-plugin): correct session.messages response shape and update tests

apnea:
## Summary

Fix the opencode plugin's `getSessionMessages` function, which had two bugs that prevented auto-retain from functioning since it was introduced in #853.

## The bugs

**1. Wrong response shape** (`hooks.ts:131`)

The code read `msg.role` but the opencode SDK returns messages wrapped as `{ info: { role }, parts }`. This caused a runtime `TypeError: Cannot read properties of undefined (reading 'role')` every time `getSessionMessages` was called.

**2. Mismatched TypeScript type for SDK call** (`hooks.ts` `OpencodeClient` type)

The production code correctly calls `opencodeClient.session.messages({ path: { id: sessionId } })`, but the TypeScript type declaration for that function specified `{ sessionID: string }` — a completely different signature. The type definition did not match the actual call. This was a type error that would have been caught at compile time with stricter type checking.

## Impact

Auto-retain and pre-compaction retain were non-functional from the point the plugin was introduced. Every call to `getSessionMessages` threw a TypeError that was silently caught, returning an empty array. The `handleSessionIdle` and compaction hooks then exited early at the `if (!messages.length) return` check. The tools (retain/recall/reflect) and system transform hook were unaffected.

The original test mocks used the same incorrect `{ role }` shape, so all tests passed despite the code not matching the real SDK contract.

## Changes

- `hooks.ts`: Fix `OpencodeClient` type to match the real SDK, read `msg.info.role` instead of `msg.role`
- `hooks.test.ts`: Update all mock message shapes to `{ info: { role }, parts }`

## Testing

- 89/89 tests pass in a clean Docker environment (node:20, fresh clone from remote)
- Manually verified auto-retain fires correctly in a live opencode session

### DIFF
--- a/hindsight-integrations/opencode/src/hooks.test.ts
+++ b/hindsight-integrations/opencode/src/hooks.test.ts
@@ -20,7 +20,7 @@ function makeClient() {
     } as any;
 }
 
-function makeOpencodeClient(messages: Array<{ role: string; parts: Array<{ type: string; text?: string }> }> = []) {
+function makeOpencodeClient(messages: Array<{ info: { role: string }; parts: Array<{ type: string; text?: string }> }> = []) {
     return {
         session: {
             messages: vi.fn().mockResolvedValue({ data: messages }),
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
we should support 1.3.x versions still

apnea:
Hi @nicoloboschi , the plugin is built against @opencode-ai/sdk@1.3.13.  I'm not sure what you mean.

nicoloboschi:
> Hi @nicoloboschi , the plugin is built against @opencode-ai/sdk@1.3.13. I'm not sure what you mean.

I thought opencode changed the message format but I was wrong, sorry, merging now

nicoloboschi (approved):

PR #995, opened 2026-04-12
fix(cross-encoder): propagate transitive ImportErrors from mlx_lm instead of masking them

### LINKED ISSUE #994: [bug] jina-mlx reranker fails at startup on Apple Silicon due to transformers 5.x lazy-import race
# [bug] jina-mlx reranker fails at startup on Apple Silicon due to transformers 5.x lazy-import race

## Environment

| Component | Version |
|---|---|
| hindsight-api | 0.5.0 |
| hindsight-api-slim | 0.5.0 |
| mlx | 0.31.1 |
| mlx-lm | 0.31.1 |
| transformers | 5.5.0 |
| sentence-transformers | 5.3.0 |
| safetensors | 0.7.0 |
| Python | 3.12.13 |
| Platform | macOS Darwin 24.3.0 arm64 (Apple Silicon) |

## Observed Behavior

When the daemon starts with `HINDSIGHT_API_RERANKER_PROVIDER=jina-mlx` and `HINDSIGHT_API_LAZY_RERANKER=false`, `memory.initialize()` fails and the following message appears:

```
ImportError: mlx and mlx-lm are required for JinaMLXCrossEncoder.
Install with: pip install mlx>=0.31.0 mlx-lm>=0.31.1 safetensors>=0.6.2
```

This message is **actively misleading**. All three packages are correctly installed. The real underlying error is:

```
ImportError: cannot import name 'AutoTokenizer' from 'transformers'
  File ".../mlx_lm/tokenizer_utils.py", line 8, in <module>
    from transformers import AutoTokenizer, PreTrainedTokenizerFast
```

The real error is caught and discarded by `hindsight_api/engine/cross_encoder.py` lines 1210-1214:

```python
try:
    import mlx.core  # noqa: F401
    import mlx_lm    # noqa: F401
except ImportError:
    raise ImportError(
        "mlx and mlx-lm are required for JinaMLXCrossEncoder. "
        "Install with: pip install mlx>=0.31.0 mlx-lm>=0.31.1 safetensors>=0.6.2"
    )
```

The `except ImportError` clause catches *any* `ImportError` that arises during the import of `mlx_lm`, including `ImportError`s raised by `mlx_lm`'s own transitive dependencies.

## Reproducer

The race fires **10/10 times** on Apple Silicon with the venv as configured. The reproducer is self-contained:

```python
import asyncio, subprocess, sys, textwrap

WORKER_CODE = textwrap.dedent("""\
import asyncio, sys, traceback

async def task_a_embeddings():
    loop = asyncio.get_event_loop()
    def _do_import():
        import transformers        # kicks off _LazyModule init
        import sentence_transformers
    await loop.run_in_executor(None, _do_import)

async def task_b_cross_encoder():
    await asyncio.sleep(0)        # yield to allow interleaving
    loop = asyncio.get_event_loop()
    def _do_import():
        import mlx_lm              # triggers tokenizer_utils -> AutoTokenizer
    await loop.run_in_executor(None, _do_import)

async def main():
    try:
        await asyncio.gather(task_a_embeddings(), task_b_cross_encoder())
        print("PASS"); sys.exit(0)
    except ImportError as exc:
        print(f"FAIL: {exc}"); sys.exit(1)

asyncio.run(main())
""")

for i in range(1, 11):
    r = subprocess.run(
        [sys.executable, "-c", WORKER_CODE],
        capture_output=True, text=True, timeout=120
    )
    print(f"Run {i:2d}: {'PASS' if r.returncode == 0 else 'FAIL'}  ({r.stdout.strip()})")
```

Each run must be a separate subprocess because mlx's nanobind C extension cannot be unloaded and re-imported within the same process on Apple Silicon.

**Note on reproducer topology**: the embedded script above uses a simplified symmetric topology where both tasks run via `run_in_executor`. This is a valid trigger of the same underlying `_LazyModule` race because the default `ThreadPoolExecutor` assigns the two tasks to different worker threads. The hindsight-api daemon itself uses an asymmetric topology (embeddings in executor, cross_encoder for `jina-mlx` on the event loop thread) — see the Root Cause section below. Both topologies trigger the race; I used the symmetric form in the reproducer because it is self-contained and does not require rebuilding the daemon init path.

## Root Cause

Transformers 5.x uses `_LazyModule` as its package `__init__` implementation. `_LazyModule` populates the top-level `transformers` namespace **incrementally** as attributes are accessed, rather than all at once at import time. Between the moment `import transformers` starts and the moment it has finished resolving `AutoTokenizer` into its namespace, there is a window in which `from transformers import AutoTokenizer` raises `ImportError`.

`memory.initialize()` at `hindsight_api/engine/memory_engine.py` lines 1691-1706 builds `init_tasks` and calls `await asyncio.gather(*init_tasks)`. The relevant two tasks for this race are:

- **Task A — `init_embeddings()`** (`memory_engine.py` line 1625). For a `local` embeddings provider (the default `BAAI/bge-base-en-v1.5` used by most deployments), line 1629 dispatches via `loop.run_in_executor(None, ...)`, meaning the `import sentence_transformers` and subsequent `import transformers` happen on a **ThreadPoolExecutor worker thread**, not on the event loop thread.

- **Task B — `init_cross_encoder()`** (`memory_engine.py` line 1635). For the `jina-mlx` provider, `cross_encoder.provider_name` is `"jina-mlx"` (not `"local"`), so line 1639's `if cross_encoder.provider_name == "local":` branch is **not** taken. Line 1641 executes `await cross_encoder.initialize()` directly, which means the `import mlx.core` and `import mlx_lm` run on the **event loop thread itself** — not inside an executor. `mlx_lm`'s top-level chain (`mlx_lm/__init__.py` → `convert.py` → `utils.py` → `tokenizer_utils.py:8`) hits `from transformers import AutoTokenizer, PreTrainedTokenizerFast`.

This is the crux of the race: **two different real threads** are inside `transformers._LazyModule` concurrently. The embeddings ThreadPoolExecutor worker is mid-way through populating the `transformers` namespace when the event loop thread's `init_cross_encoder` attempts `from transformers import AutoTokenizer`, and `_LazyModule.__getattr__` has not yet resolved that attribute into the namespace. CPython's import lock is per-module, not per-attribute, and `_LazyModule` does not hold fine-grained locks per attribute, so the window between "namespace is being built" and "AutoTokenizer is accessible" is unguarded.

In a plain sequential subprocess (`import sentence_transformers; import mlx_lm` run one after the other on a single thread), the error never fires because the first import completes before the second starts. In a single-threaded `asyncio.gather` test where both coroutines run on the same event loop thread, the error also does not fire for the same reason — no real thread-level concurrency exists. The race requires the specific topology of the hindsight-api daemon, where Task A runs in an executor worker thread (due to `provider_name == "local"`) while Task B runs on the event loop thread (due to `provider_name == "jina-mlx"`).

## Workaround

Setting `HINDSIGHT_API_LAZY_RERANKER=true` defers cross-encoder initialization until the first `recall()` call. By that time the embeddings init has completed, `sentence_transformers` has finished loading, and the `transformers` namespace is fully populated. `from transformers import AutoTokenizer` then succeeds. This workaround is effective but silent -- users who hit the misleading error message have no way to discover it without reading the source code.

## Suggested Fix

The options below are ranked by (effort × correctness × side-effect risk), lowest-risk first.

---

### Option A — Re-raise the real error verbatim (recommended, 2-line fix, zero risk)

**In `cross_encoder.py` lines 1210-1214**, replace the bare `except ImportError` with a narrowed handler that re-raises unrelated ImportErrors verbatim (preserving their implicit traceback chain) and only emits the static "mlx and mlx-lm are required" message when the caught error actually references mlx. This stops masking unrelated import failures and surfaces the real underlying error:

```python
try:
    import mlx.core  # noqa: F401
    import mlx_lm    # noqa: F401
except ImportError as exc:
    # Distinguish "packages not installed" from "transitive import failure"
    if "mlx" not in str(exc) and "mlx_lm" not in str(exc):
        # Real error from inside mlx_lm (e.g. transformers lazy-load race)
        raise
    raise ImportError(
        "mlx and mlx-lm are required for JinaMLXCrossEncoder. "
        "Install with: pip install mlx>=0.31.0 mlx-lm>=0.31.1 safetensors>=0.6.2"
    ) from exc
```

This surfaces the real error immediately, enabling the user (or a log scraper) to diagnose the race. It does not change any runtime behavior for the "packages not installed" case.

**Tradeoff**: Purely diagnostic. Does not prevent the race; just stops hiding it.

---

### Option B — Force eager `transformers` import at the top of `cross_encoder.py` before the try block

Add this line immediately before the `try` block in `JinaMLXCrossEncoder.initialize()`:

```python
async def initialize(self) -> None:
    if self._reranker is not None:
        return

    # Force transformers to fully populate its namespace before mlx_lm
    # attempts to import AutoTokenizer. Required because transformers 5.x
    # uses _LazyModule and is vulnerable to concurrent import races.
    import transformers
    _ = transformers.AutoTokenizer  # force attribute resolution now

    try:
        import mlx.core  # noqa: F401
        import mlx_lm    # noqa: F401
    except ImportError:
        ...
```

**Tradeoff**: A direct fix to the race. The `_ = transformers.AutoTokenizer` line forces `_LazyModule` to resolve the attribute before any concurrent thread can hit the incomplete-namespace window. Minor overhead (negligible at init time). Relies on implementation knowledge of transformers internals, so it should be accompanied by a comment referencing this issue.

---

### Option C — Force `from transformers import AutoTokenizer` inside `initialize()` before the try block, using the same thread

Same intent as Option B but expressed more directly:

```python
async def initialize(self) -> None:
    if self._reranker is not None:
        return

    # Pre-warm the transformers namespace so that mlx_lm's tokenizer_utils
    # can safely execute `from transformers import AutoTokenizer` in a
    # concurrent executor thread. See: github.com/vectorize-io/hindsight/issues/XXX
    loop = asyncio.get_event_loop()
    await loop.run_in_executor(None, _prewarm_transformers)

    try:
        import mlx.core
        import mlx_lm
    except ImportError:
        ...

def _prewarm_transformers() -> None:
    from transformers import AutoTokenizer  # noqa: F401
```

**Tradeoff**: Slightly higher overhead than Option B (one executor round-trip before the main init), but is self-documenting and keeps the prewarm call auditable as a named function. The race is closed regardless of what else is running concurrently because the prewarm completes before the `mlx_lm` import starts.

---

### Option D — Default `HINDSIGHT_API_LAZY_RERANKER=true` when provider is `jina-mlx`, with an explanatory log warning

In the provider selection logic, when `jina-mlx` is chosen, override `lazy_reranker` to `True` and emit a `WARNING` level log:

```
WARN: jina-mlx reranker uses lazy initialization (HINDSIGHT_API_LAZY_RERANKER
      forced true) to avoid a transformers 5.x concurrent-import race during
      asyncio.gather. First recall() call will trigger model load.
      See: github.com/vectorize-io/hindsight/issues/XXX
```

**Tradeoff**: The simplest deployment fix and eliminates all user-facing failures. The cost is that first-call latency increases (model load deferred). Users who need deterministic startup warmup (e.g., health-check probes that call recall()) cannot distinguish "healthy but lazy" from "failed to load". Masks the root cause rather than fixing it. Best deployed as a release blocker workaround while Options A+B are implemented.

---

### Recommended priority

1. **Option A** now — 2-line change, zero risk, immediately surfaces the real error in the current release so users can find this issue and apply the workaround.
2. **Option B or C** in the same PR — closes the race properly.
3. **Option D** optionally as a belt-and-suspenders default for jina-mlx until transformers stabilises its lazy-load behavior.

## Where to File

Canonical repository: **https://github.com/vectorize-io/hindsight**

This is the upstream monorepo for both `hindsight-api` and `hindsight-api-slim`. The relevant file is `hindsight-api/hindsight_api/engine/cross_encoder.py`. Issues are open on that repository.

octo-patch:
Fixes #994

## Problem

When mlx_lm is installed but one of its transitive dependencies raises an ImportError (e.g. the transformers 5.x lazy-import race on Apple Silicon), the broad except ImportError clause in JinaMLXCrossEncoder.initialize() catches it and replaces it with a misleading message saying to install mlx and mlx-lm even when both are correctly installed.

## Solution

Change except ImportError to except ModuleNotFoundError. ModuleNotFoundError is only raised when a package itself cannot be found. Transitive ImportErrors from inside a package's own code now propagate as-is, surfacing the true root cause.

## Testing

Verified with the reproducer from the issue: on Apple Silicon with transformers 5.x, the transitive ImportError from mlx_lm now propagates directly instead of being masked.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -1210,7 +1210,7 @@ class JinaMLXCrossEncoder(CrossEncoderModel):
         try:
             import mlx.core  # noqa: F401
             import mlx_lm  # noqa: F401
-        except ImportError:
+        except ModuleNotFoundError:
             raise ImportError(
                 "mlx and mlx-lm are required for JinaMLXCrossEncoder. "
                 "Install with: pip install mlx>=0.31.0 mlx-lm>=0.31.1 safetensors>=0.6.2"


nicoloboschi:
fixed by https://github.com/vectorize-io/hindsight/pull/1009

PR #1015, opened 2026-04-13
fix(codex): serialize hindsight-embed daemon startup

nicoloboschi:
## Summary
- Serialize concurrent Codex hook calls to `hindsight-embed daemon start` via a shared advisory file lock, eliminating the duplicate-startup race that caused the April 2026 incident.
- Route SessionStart warmup through the same locked startup path as foreground retain/recall.
- Add regression tests covering the in-flight startup race and the lock-timeout recovery path.

## Root cause
A duplicate local daemon start caused a bind conflict on port `9077`. The losing process ran shutdown cleanup and stopped the shared embedded Postgres instance, leaving the first worker stuck polling the database with connection-refused errors.

## Differences vs. #985
This supersedes #985 with five fixes to issues raised in review:

1. **No more private-symbol access.** `_state_file` / `fcntl` imports are replaced with a public `acquire_state_lock` context manager and `state_file_path` helper in `lib/state.py`.
2. **Waiter timeout ≥ holder worst-case.** Previous timeout (30s) was shorter than the holder's budget (~70s: profile create + daemon start + readiness), so a second hook could time out and race. Waiter now tolerates 120s and, if it still times out, falls back to a final health re-check rather than hard-failing.
3. **No `python -c` + sys.path hack.** `prestart_daemon_background` now spawns a standalone `scripts/prestart_daemon.py` entrypoint via `sys.executable`; the entrypoint sets its own `sys.path` relative to its file location, so it works regardless of how Python was launched or where the plugin is installed.
4. **Background failures are no longer silent.** Warmup failures are written to `~/.hindsight/codex/state/daemon-start.log` with a timestamp and traceback, so opportunistic startup failures are still debuggable after the fact.
5. **Test documents flock/OFD semantics.** The regression test includes a comment explaining why `flock` taken in a helper thread still blocks the main thread (each `open()` creates a distinct open-file-description on Linux/macOS).

## Verification
- `uv run pytest hindsight-integrations/codex/tests -q` — 74 passed
- `./scripts/hooks/lint.sh` — clean

## Test plan
- [x] Unit test simulates concurrent startup: second caller waits on the lock instead of racing the daemon process
- [x] Unit test simulates lock-acquisition timeout: caller recovers via health check when daemon is up, raises clear error when not
- [ ] Manual: two Codex sessions started back-to-back bring up exactly one daemon on port 9077


nicoloboschi:
Superseded: fixing this properly in hindsight-embed instead. Per-integration flocks would just proliferate as each integration copies the pattern.

PR #1027, opened 2026-04-13
fix(openclaw): follow up review feedback on session identity guardrails

Tosko4:
## Summary
- follow up on the review feedback from #987 without changing `channel_type` semantics
- extract shared identity resolution and cache handling across the hook pipeline
- replace stringly retry handling with typed skip reasons and expand regression coverage

## What changed
- preserve the previous `channel_type` behavior in retain metadata, while still allowing `provider` to fall back from the parsed session key
- consolidate the repeated resolve/cache/skip flow into a shared helper used by `before_dispatch`, `before_agent_start`, `before_prompt_build`, and `agent_end`
- remove the redundant sender-id map and derive sender identity from the single cached identity record
- remove unnecessary case-insensitive flags from lowercase session-key regexes
- add unit and integration coverage for operational sessions, `temp:` sessions, missing provider or sender cases, retryable vs final skip reasons, and the `channel_type` fallback regression

## Validation
- `cd hindsight-integrations/openclaw && npx vitest run src/index.test.ts`
- `cd hindsight-integrations/openclaw && npx vitest run --config vitest.integration.config.ts tests/hooks.integration.test.ts`

## Note
- local pre-commit lint currently fails in this environment because a repo-wide ESLint dependency (`@eslint/js`) is missing under `hindsight-control-plane`; the targeted OpenClaw tests and build above passed
- this PR supersedes the implementation in #987 and folds in the requested review follow-up

### DIFF
--- a/hindsight-integrations/openclaw/src/index.test.ts
+++ b/hindsight-integrations/openclaw/src/index.test.ts
@@ -8,6 +8,12 @@ import {
   composeRecallQuery,
   truncateRecallQuery,
   buildRetainRequest,
+  parseSessionKey,
+  extractTelegramDirectSenderId,
+  resolveSessionIdentity,
+  getIdentitySkipReason,
+  isEphemeralOperationalText,
+  deriveBankId,
 } from './index.js';
 import type { PluginConfig, MemoryResult } from './types.js';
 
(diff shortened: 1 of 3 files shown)


Tosko4:
Closing this one, I got it wrong on my side. The review follow-up was meant for the existing PR #987, and the fix has now been pushed there in commit 6d6b2604. So this separate PR is no longer needed. Sorry for the noise.

PR #1039, opened 2026-04-13
feat(reflect): add facts_only mode to exclude chunks and expand tool

cdbartholomew:
## Summary
- Adds `facts_only` flag to the reflect API and mental model trigger config
- When enabled, raw source chunks are excluded from recall results, the expand tool is removed from the agent's tool set, and `max_chunk_tokens` is hidden from the recall schema
- Mental model refreshes can now be scoped to only consider extracted facts, without raw document/chunk context influencing synthesis

## Changes
- `ReflectRequest` and `MentalModelTrigger` — new `facts_only: bool` field
- `reflect_async` / `run_reflect_agent` — thread `facts_only` through to tool setup
- `tool_recall` — `include_chunks` is now a parameter (was hardcoded `True`)
- `get_reflect_tools` — conditionally excludes expand tool and chunk params when `facts_only=True`
- `agent.py` — expand is now governed by `enabled_tools` (removed from hardcoded exemption)

## Test plan
- [x] 8 new unit tests covering tool schema, expand rejection, and default behavior
- [x] All 46 reflect agent tests pass
- [x] Lint passes

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -720,6 +720,12 @@ class ReflectRequest(BaseModel):
         default=None,
         description="Exclude specific mental models by ID from the reflect loop.",
     )
+    facts_only: bool = Field(
+        default=False,
+        description="When true, only extracted facts are used during reflect. "
+        "Raw source chunks and document context are excluded. "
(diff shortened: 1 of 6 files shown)


nicoloboschi:
added a more complete fix here https://github.com/vectorize-io/hindsight/pull/1048

PR #1044, opened 2026-04-14
fix(bank-template): align BankTemplateConfig with _CONFIGURABLE_FIELDS

mrkhachaturov:
## Summary

`_CONFIGURABLE_FIELDS` on `HindsightConfig` defines the contract for what can be overridden per bank — currently 22 fields. `BankTemplateConfig` is the manifest-side representation of that same contract, but it only declares 12. The other ten exist on `HindsightConfig` and `config_resolver.update_bank_config()` accepts them through the bank config API directly, but the template import path at `POST /v1/default/banks/{id}/import` cannot deliver them: the manifest handler resolves overrides via `BankTemplateConfig.get_config_updates()`, which is a `model_dump()` filter, so any field not declared on the model is silently dropped before reaching `update_bank_config()`.

The result is two paths to the same per-bank knob that disagree. The bank config API works for all 22 fields; the template import path silently no-ops on ten of them.

This PR aligns `BankTemplateConfig` with the existing `_CONFIGURABLE_FIELDS` allowlist by declaring the missing ten:

- `retain_default_strategy`
- `retain_strategies`
- `retain_chunk_batch_size`
- `mcp_enabled_tools`
- `consolidation_llm_batch_size`
- `consolidation_source_facts_max_tokens`
- `consolidation_source_facts_max_tokens_per_observation`
- `max_observations_per_scope`
- `reflect_source_facts_max_tokens`
- `llm_gemini_safety_settings`

Field types match the existing declarations on `HindsightConfig` (`dict | None` and `list | None` for the two untyped-generic fields, `int | None` for the integer override slots — same `| None` override convention every other field in `BankTemplateConfig` already uses).

Non-breaking. Existing templates continue to parse and apply unchanged. The new fields are additive — present on the model, default to `None`, ignored when not set. No engine changes; `config_resolver.update_bank_config()` already validates these fields correctly through `_CONFIGURABLE_FIELDS`.

## Test plan

- [x] New parametrized test `tests/test_bank_template_configurable_fields.py`. Each field is POSTed through `/v1/default/banks/{id}/import` with a value chosen to differ visibly from the default, then read back via `GET /v1/default/banks/{id}/config` and asserted under `config["overrides"]`. Fixture shape matches `test_import_applies_config` in `tests/test_bank_templates.py`.
- [x] Pydantic-level guard `test_bank_template_config_declares_every_configurable_field` checks every field in the parametrize list is declared on the model — catches a future rename or removal on either side.
- [ ] CI green (will validate the new test, ruff, and ty in the upstream pipeline).

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1673,6 +1673,36 @@ class BankTemplateConfig(BaseModel):
     entities_allow_free_form: bool | None = Field(
         default=None, description="Allow entities outside the label vocabulary"
     )
+    retain_default_strategy: str | None = Field(
+        default=None, description="Name of the default retain strategy (key into retain_strategies map)"
+    )
+    retain_strategies: dict | None = Field(
+        default=None, description="Map of retain strategy name to per-strategy config dict"
+    )
+    retain_chunk_batch_size: int | None = Field(
+        default=None, description="Max chunks per streaming batch (0 disables batching)"
+    )
+    mcp_enabled_tools: list[str] | None = Field(
+        default=None, description="MCP tool allowlist for this bank (None = all tools)"
+    )
+    consolidation_llm_batch_size: int | None = Field(
+        default=None, description="LLM batch size for observation consolidation"
+    )
+    consolidation_source_facts_max_tokens: int | None = Field(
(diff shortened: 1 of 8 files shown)


nicoloboschi (approved):
LGTM

PR #1045, opened 2026-04-14
docs(configuration): document HINDSIGHT_API_RETAIN_CHUNK_BATCH_SIZE

mrkhachaturov:
## Summary

`retain_chunk_batch_size` and its `HINDSIGHT_API_RETAIN_CHUNK_BATCH_SIZE` env var have lived on `HindsightConfig` since the retain streaming batch landed, but the Retain section of the configuration reference never got a row for them. Anyone who wants to cap chunk-batch size on large document ingestion has to grep the source to find the env var name.

Adds one row to the Retain table, in the same format as the surrounding entries, with a note that the field is configurable per bank via the bank config API.

## Test plan

- [x] Row format matches the surrounding `HINDSIGHT_API_RETAIN_*` entries (variable | description | default).
- [ ] CI green — docs site build validates the markdown table renders.


nicoloboschi (changes requested):
thanks

nicoloboschi (approved):

PR #1056, opened 2026-04-14
docs(opencode): drop npm install step, document Hindsight Cloud

nicoloboschi:
## Summary
- Remove misleading \`npm install @vectorize-io/opencode-hindsight\` step — OpenCode auto-installs plugins listed in the \`"plugin"\` array at startup via Bun (per https://opencode.ai/docs/plugins/).
- Add a dedicated **Using Hindsight Cloud** section with the managed API URL (\`https://api.hindsight.vectorize.io\`) and token guidance in both the integration README and the docs site page.

## Test plan
- [ ] Verify rendered docs page reads correctly
- [ ] Confirm a fresh install using only the documented \`opencode.json\` snippet works (plugin auto-installs on first \`opencode\` run)

### DIFF
--- a/hindsight-docs/docs-integrations/opencode.md
+++ b/hindsight-docs/docs-integrations/opencode.md
@@ -10,28 +10,46 @@ Persistent long-term memory plugin for [OpenCode](https://opencode.ai) using [Hi
 
 ## Quick Start
 
-```bash
-# 1. Install the plugin
-npm install @vectorize-io/opencode-hindsight
-```
-
-Add to your `opencode.json`:
+Add to your `opencode.json` (project) or `~/.config/opencode/opencode.json` (global):
 
 ```json
 {
+  "$schema": "https://opencode.ai/config.json",
   "plugin": ["@vectorize-io/opencode-hindsight"]
 }
 ```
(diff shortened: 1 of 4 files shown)


nicoloboschi:
## Scope expanded — now includes a plugin load crash fix

Second commit `87856fcc` fixes a real bug: OpenCode iterates `Object.entries(mod)` and invokes every export as a Plugin factory. Our default export was a `PluginModule` object (`{ id, server }`), so opencode called it and crashed at load time with:

```
TypeError: fn3 is not a function. (In 'fn3(input)', 'fn3' is an instance of Object)
  at src/plugin/index.ts:87
```

Default-exporting the `HindsightPlugin` function itself (same reference as the named export — opencode dedupes by identity) resolves it. Default-export smoke test updated to lock in the invariant.

Verified end-to-end against opencode 1.1.49 with the built dist: plugin initializes, registers tools/hooks, `opencode run` returns a response, session.idle fires retain. 91/91 unit tests pass.

After merge, cut `integrations/opencode/v0.1.4` via `./scripts/release-integration.sh opencode patch`.

PR #1057, opened 2026-04-14
fix(opencode): replace unconditional console.error with debugLog

DK09876:
## Summary

- Replace 13 hardcoded `console.error` calls in `hooks.ts` with `debugLog(config, ...)` so they only appear when `debug: true` is set in plugin options
- PR #993 added these for debugging the message parsing fix, but they were left unconditional, spamming every user's TUI with red error text on every event, message parse, and retain cycle

## Test plan

- [x] All 91 opencode plugin tests pass
- [x] Verified in live OpenCode session — no TUI spam with `debug: false`, logs appear with `debug: true`


apnea:
Hi @DK09876, apologies for leaving the `console error` calls lying around after testing my PR. Thanks for fixing that.

PR #1065, opened 2026-04-14
fix(docs): regenerate bank-template-schema.json and guard drift

mrkhachaturov:
## Summary

hindsight-docs/static/bank-template-schema.json is hand-edited. Nothing regenerates it and nothing checks it. Three PRs have changed `BankTemplateManifest` since it was last touched:

- #902 flipped `BankTemplateConfig.entity_labels` from `list[str]` to `list[dict[str, Any]]`
- #1044 added ten `BankTemplateConfig` fields (`retain_default_strategy`, `retain_strategies`, `retain_chunk_batch_size`, `mcp_enabled_tools`, four `consolidation_*` fields, `max_observations_per_scope`, `reflect_source_facts_max_tokens`, `llm_gemini_safety_settings`)
- #1048 added three `MentalModelTrigger` fields (`include_chunks`, `recall_max_tokens`, `recall_chunks_max_tokens`)

None of the bundled templates exercise any of these fields, so `check-templates.mjs` kept passing. A template that uses the dict-shaped label format fails Ajv with `should be string` on every label.

This PR regenerates the schema from `BankTemplateManifest.model_json_schema()` and hooks the generator into the existing `verify-generated-files` CI job alongside `generate-openapi` and `generate-clients`, so the same drift guard you already rely on for those catches future schema changes.

## Test plan

- [x] `./scripts/generate-bank-template-schema.sh` runs cleanly and is idempotent on re-run
- [x] `./scripts/hooks/lint.sh` — all lints pass (new Python file + ruff + ty)
- [x] `node hindsight-docs/scripts/check-templates.mjs` — all 3 bundled templates still validate against the regenerated schema
- [x] A template using `entity_labels: list[dict]` (dict-shaped) now validates against the regenerated schema but is rejected by the stale schema on main with `should be string` — confirms the drift was reachable
- [ ] CI `verify-generated-files` job passes the new `Run generate-bank-template-schema` step

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2547,6 +2547,9 @@ jobs:
     - name: Run generate-openapi
       run: ./scripts/generate-openapi.sh
 
+    - name: Run generate-bank-template-schema
+      run: ./scripts/generate-bank-template-schema.sh
+
     - name: Run generate-clients
       run: ./scripts/generate-clients.sh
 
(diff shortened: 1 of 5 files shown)


nicoloboschi (approved):
LGTM, thank you!!

Issue #1033, opened 2026-04-13
openclaw backfill misdetects symlinked bin path as direct execution


Aldoustheorchestrator:
### Bug Description
The OpenClaw backfill entrypoint appears to mishandle symlinked executable paths.

In the failing path, `isDirectExecution(...)` in `hindsight-integrations/openclaw/src/backfill.ts` treats a symlinked bin invocation as though the script itself were executed directly. That seems to be the cause of the recurring `backfill.test.ts` failure.

### Steps to Reproduce
1. Run the OpenClaw backfill path through a symlinked bin path.
2. Exercise the code path covered by `hindsight-integrations/openclaw/src/backfill.test.ts` for `isDirectExecution(...)`.
3. Observe that the symlinked bin invocation is classified as direct script execution.

### Expected Behavior
A symlinked bin invocation should behave like normal CLI execution, not like the script was executed directly.

### Actual Behavior
The current detection logic treats the symlink target as direct execution, which causes the backfill test path to fail.

### Version
Observed while working on PR #948 against the current branch state on 2026-04-13.

### LLM Provider
Other

D2758695161:
## PR Submitted ✅

Fix implemented in PR #1067: https://github.com/vectorize-io/hindsight/pull/1067

**Root cause:** `canonicalizeExecutionPath` uses `realpathSync` which resolves only one level of symlinks. When a bin symlink points to a CLI wrapper (which then imports the module), the entrypoint resolves to the wrapper while `import.meta.url` points to the module — they don't match.

**Fix:** Added `resolveSymlinks()` that iteratively follows the full symlink chain (with cycle detection) for the entrypoint. Now correctly returns `false` for CLI invocation through a bin wrapper, `true` for direct script execution.

Happy to address any feedback!

— 一筒 🦀

nicoloboschi:
fixed by https://github.com/vectorize-io/hindsight/pull/1093

PR #1067, opened 2026-04-15
fix(openclaw): resolve full symlink chain in isDirectExecution()

D2758695161:
## Fix: `isDirectExecution()` misdetects symlinked bin path

**Issue:** #1033

### Root Cause

When a bin symlink points to a CLI wrapper (not directly to the module), `process.argv[1]` contains the wrapper's path after `realpathSync` resolution, while `import.meta.url` points to the actual module file. Since these are different files, `isDirectExecution()` returns `false` — even though the module IS being executed.

Example failing path:
```
process.argv[1] = /usr/local/bin/hindsight-openclaw-backfill
  ↓ realpathSync (resolves one level)
  = /usr/local/share/hindsight/bin/cli-wrapper.js

import.meta.url → fileURLToPath → /usr/local/share/hindsight/dist/backfill.js
```

These are different files → returns `false` instead of `true`.

### Fix

Added `resolveSymlinks()` that iteratively follows the full symlink chain (with cycle detection) for the entrypoint path. The module URL is already the actual module-as-executed, so `canonicalizeExecutionPath` suffices for it.

```typescript
function resolveSymlinks(path: string): string {
  let current = path;
  const visited = new Set<string>();
  try {
    while (!visited.has(current)) {
      visited.add(current);
      try {
        const next = realpathSync(current);
        if (next === current) break;
        current = next;
      } catch { break; }
    }
  } catch { return path; }
  return current;
}

export function isDirectExecution(
  entrypoint: string | undefined = process.argv[1],
  moduleUrl: string = import.meta.url
): boolean {
  if (!entrypoint) return false;
  // Resolve entrypoint through full symlink chain; moduleUrl is the actual module
  return resolveSymlinks(entrypoint) === canonicalizeExecutionPath(fileURLToPath(moduleUrl));
}
```

- `node dist/backfill.js` → resolves to same file → `true` ✅
- CLI bin symlink → resolves to CLI wrapper → `false` ✅ (correctly not direct)
- Symlink directly to module → resolves to module → `true` ✅ (test case still passes)

— 一筒 🦀

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1639,6 +1639,14 @@ class MentalModelResponse(BaseModel):
         default=None,
         description="Full reflect API response payload including based_on facts and observations",
     )
+    is_stale: bool | None = Field(
+        default=None,
+        description=(
+            "True when new memories matching this mental model's tag/fact_type scope have been "
+            "ingested since last_refreshed_at, or consolidation has pending items. Only populated "
+            "when detail=full."
+        ),
+    )
 
 
 class MentalModelListResponse(BaseModel):
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -593,17 +593,15 @@ async def _trigger_mental_model_refreshes(
     """
     pool = memory_engine._pool
 
-    # Find mental models with refresh_after_consolidation=true
-    # SECURITY: Control which mental models get refreshed based on tags
+    # Find mental models with refresh_after_consolidation=true that are actually stale.
+    # The tag filter on the SELECT enforces the security boundary (never look outside the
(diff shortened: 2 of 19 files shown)


nicoloboschi:
Thanks for the PR — a few concerns before this can be merged:

1. **Bug direction conflicts with #1033.** The issue reports that symlinked bin invocation is *currently* classified as direct execution (returns `true`) and should return `false`. This PR's description claims the opposite — that it currently returns `false` and should return `true`. Can you reconcile?

2. **`realpathSync` already resolves the full chain.** Node's `fs.realpathSync` resolves all symlinks in one call, not "one level" as the PR body states. If a single `realpathSync` in `canonicalizeExecutionPath` isn't enough, the real bug is likely elsewhere — worth pinpointing before adding an iterative wrapper.

3. **Whole-file CRLF conversion.** The file switched from LF to CRLF line endings, which inflates the diff to 613+/592− for what is really a ~20-line change. Please restore LF so the diff reflects only the intended change.

4. **Missing test.** Please add a test that fails on `main` and passes with this change, reproducing the exact symlink layout from #1033.

PR #1066, opened 2026-04-15
feat(docs): split template manifests into per-file manifest_file refe…

mrkhachaturov:
## Summary

`hindsight-docs/src/data/templates.json` mixes two unrelated concerns: presentation metadata for the gallery card (name, description, category, integrations) and the `BankTemplateManifest` body that gets POSTed to `/v1/default/banks/{id}/import`. A contributor who only tweaks `retain_mission` on one template has to touch a 130-line file full of metadata they did not mean to edit, and review diffs mix card-level and manifest-level changes in the same hunk.

This PR moves each manifest into its own file under `hindsight-docs/src/data/templates/`. The catalog entry in `templates.json` keeps the presentation fields and replaces the inline `manifest` field with a `manifest_file` path pointing at the per-template file. Each file is now a pure `BankTemplateManifest` and can be POSTed directly to `/v1/default/banks/{bank_id}/import` without any extraction step.

The renderer in `src/pages/templates/index.tsx` uses webpack's `require.context` to eagerly bundle every `.json` file under `src/data/templates/` at build time. This keeps the contributor flow to **two files total**: create the manifest under `src/data/templates/`, add a catalog entry to `templates.json`. No code change needed to register the new file — webpack picks it up automatically on the next build.

The validator `scripts/check-templates.mjs` is updated to follow `manifest_file` references off disk instead of reading inline manifests from the catalog.

I also added a `Submit a template →` CTA button to the gallery banner, matching the existing pattern on the integrations page (same `Link` component, same `submitButton` styling copied verbatim from `hindsight-docs/src/pages/integrations/index.module.css`). The button opens `templates.json` in GitHub's web editor in a new tab, so contributors get a one-click contribution path.

**Zero runtime/visual change for existing templates.** The rendered Template Hub page is byte-identical: same gallery cards, same category filters, same search, same modal, same "Copy JSON" output. `require.context` inlines every manifest into the compiled JS bundle at build time, so there is no runtime fetch, no deployed asset change, and no UX delta for anyone browsing the page.

## Test plan

- [x] `node hindsight-docs/scripts/check-templates.mjs` — all 3 bundled templates (Conversation, Coding Agent, Personal Assistant) validate against `static/bank-template-schema.json`.
- [x] `npm run build --workspace=hindsight-docs` — full Docusaurus build succeeds. Only pre-existing broken-anchor warnings (same as clean `main`, unrelated to this PR).
[...]

PR #1068, opened 2026-04-15
refactor(templates): split into hindsight-bank-templates package (v2)

mrkhachaturov:
Moves the bank template catalog out of hindsight-docs/src/data/templates.json and into a new top-level package at hindsight-bank-templates/, modelled on hindsight-clients/ + scripts/generate-clients.sh. The source package holds per-file manifests plus a small catalog, and a shell script flattens it into hindsight-docs/src/data/templates.json when you run it.

The Templates Hub page and scripts/check-templates.mjs validator still read the committed flat file. The generator produces the same bytes as main.

- hindsight-bank-templates/templates.json: small catalog with manifest_file refs, one entry per template (id, name, description, category, integrations, manifest_file)
- hindsight-bank-templates/templates/{conversation,coding-agent, personal-assistant}.json: full manifest bodies lifted out of the previously inline catalog
- scripts/generate-templates.sh: pure bash + jq, reads the source package and writes hindsight-docs/src/data/templates.json. Matches the style of generate-openapi.sh, generate-clients.sh, generate-docs-skill.sh.
- .github/workflows/test.yml: add a generate-templates step to the existing verify-generated-files job so a PR that forgets to regenerate fails CI. The new script is also listed in the failure-hint command block.
- hindsight-docs/src/pages/templates/index.tsx + index.module.css: add a CTA button on the submit banner that opens the source catalog at hindsight-bank-templates/templates.json, so contributors edit the source instead of the generated output.

Why a top-level package instead of nesting under hindsight-docs/:

- Templates are data, not docs. Putting them at the repo root makes that explicit and opens the door for other consumers (hindsight-cli, SDKs, third-party tooling) to read from the same source without reaching into a docs subtree.
- Down the line this gives hindsight-cli somewhere natural to add a `hindsight template install <id> --bank <name>` subcommand that fetches manifests straight from the package instead of parsing a docs data file.
- Editing a template means touching two focused files (catalog + one manifest) instead of scrolling through a 130-line aggregated JSON to change a single retain_mission.
- The drift check in verify-generated-files catches the case where someone edits a manifest and forgets to rerun the script, the same way generate-clients.sh already catches SDK drift.

v2 of the "per-file manifest_file refs" approach. Same end result, different mechanics: the v1 branch resolves manifest_file at Docusaurus build time via require.context inside the Templates Hub page; this branch resolves it up front via a shell script and commits the flat output, so Docusaurus reads the file the same way it does on main. Pick whichever feels better to maintain.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2553,6 +2553,9 @@ jobs:
     - name: Run generate-docs-skill
       run: ./scripts/generate-docs-skill.sh
 
+    - name: Run generate-templates
+      run: ./scripts/generate-templates.sh
+
     - name: Run lint
       run: ./scripts/hooks/lint.sh
 
@@ -2568,6 +2571,7 @@ jobs:
           echo "  ./scripts/generate-openapi.sh"
           echo "  ./scripts/generate-clients.sh"
           echo "  ./scripts/generate-docs-skill.sh"
+          echo "  ./scripts/generate-templates.sh"
           echo "  ./scripts/hooks/lint.sh"
           echo ""
           git diff --stat
--- a/hindsight-bank-templates/templates.json
+++ b/hindsight-bank-templates/templates.json
@@ -1,0 +1,49 @@
+{
+  "templates": [
+    {
+      "id": "conversation",
+      "name": "Conversation",
+      "description": "For chat-based agents and assistants. Tracks user preferences, conversation patterns, and builds a profile over time.",
+      "category": "chat",
+      "integrations": [
+        "litellm",
+        "langgraph",
+        "pydantic-ai",
+        "ai-sdk",
+        "chat",
+        "crewai",
+        "ag2",
+        "agno",
+        "strands",
+        "llamaindex",
+        "local-mcp",
+        "skills"
+      ],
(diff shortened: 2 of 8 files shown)


mrkhachaturov:
Heads up: this is a second take at #1066 (feat(docs): split template manifests into per-file manifest_file refs).

Both PRs land the same contributor workflow: edit one small manifest file per template instead of scrolling a 130-line aggregated JSON. The only difference is where the manifest_file lookup runs.

This PR (v2) puts the source in a new top-level package at hindsight-bank-templates/, adds scripts/generate-templates.sh to flatten it into the committed hindsight-docs/src/data/templates.json, and wires a drift check into verify-generated-files. It matches how hindsight-clients/ and generate-clients.sh work and leaves the Templates Hub page byte-identical to main.

v1 (#1066) keeps everything inside hindsight-docs/. Source at hindsight-docs/src/data/templates/<id>.json, and hindsight-docs/src/pages/templates/index.tsx resolves the refs at build time via require.context. No new top-level folder, no generator script, no CI step. Smaller diff.

Happy with either. I'll close whichever you don't want.

nicoloboschi:
thanks for your work, I prefer to not include this one as the other is already a clean design

PR #1071, opened 2026-04-15
fix(gemini): remove redundant schema injection into system prompt

### LINKED ISSUE #1070: GeminiLLM provider sends structured output schema twice (prompt + native response_schema)
### Bug Description

`GeminiLLM.call()` injects the JSON schema into `system_instruction` as text and also passes it via the native `response_schema` parameter. Both reach the API on every structured-output call, wasting input tokens with no benefit.

### Steps to Reproduce

1. Configure the Gemini provider (`HINDSIGHT_API_LLM_PROVIDER=gemini`).
2. Trigger any path that uses structured output (e.g. `retain`).
3. Inspect the request: the JSON schema appears both in `system_instruction` and in `response_schema`.

### Expected Behavior

The schema is sent only once, via the native `response_schema` parameter (with `response_mime_type="application/json"`).

### Actual Behavior

[`hindsight-api-slim/hindsight_api/engine/providers/gemini_llm.py`](https://github.com/vectorize-io/hindsight/blob/main/hindsight-api-slim/hindsight_api/engine/providers/gemini_llm.py) lines 212–219 append `"You must respond with valid JSON matching this schema:\n{schema}"` to `system_instruction`, while lines 225–227 already set `response_schema=response_format`. Removing the prompt injection has no effect on `retain` correctness in local testing.

### Version

`48ba0d86` (main)

### LLM Provider

Gemini

seheepeak:
## Summary

Removes the JSON schema text injection from `system_instruction` in `GeminiLLM.call()`. The schema is already passed via the native `response_schema` parameter (with `response_mime_type="application/json"`), so the prompt-side injection was duplicating it on every structured-output call and wasting input tokens.

Fixes #1070

## Test plan

- [x] Verified `retain` still works correctly with the Gemini provider after removing the lines
- [x] `ruff check`, `ruff format --check`, `ty check` pass on the modified file


nicoloboschi (approved):

seheepeak:
CI failures look unrelated to this PR — fixture setup fails with `HINDSIGHT_API_LLM_VERTEXAI_PROJECT_ID` (and `COHERE_API_KEY`) missing from the CI env. Probably needs a maintainer's eyes.

PR #1088, opened 2026-04-15
fix(worker): per-tenant fair rotation in claim_batch

cdbartholomew:
## Summary

Prevents a single busy tenant from starving workers away from other tenants in a multi-tenant deployment. When one tenant has a much larger backlog than others, tenants at the front of the `list_tenants()` iteration could monopolize every claim and leave others queued indefinitely.

## The bug

`WorkerPoller.claim_batch()` iterated schemas in the fixed order returned by `tenant_extension.list_tenants()` and claimed until slots filled. A tenant near the front of the list could take every available slot, poll after poll.

## Fix

Round-robin rotation at the schema level:

- `WorkerPoller` now tracks `_next_schema_idx`, which advances **past the last schema we serviced** (not `+1 from previous offset` — that would still let a heavy tenant at the same position win iteration after iteration).
- **Pass 1 (fairness)**: cap at 1 claim per pool per schema, so every tenant with pending work is considered before we return to one we already claimed from.
- **Pass 2 (capacity)**: backfill remaining slots from any schema when the first pass didn't fill them — so single-tenant throughput isn't sacrificed for fairness.

Starvation bound: (time until any worker frees up) + one poll interval. A small tenant's single task is claimed within one rotation cycle.

## Interactions checked

- **`MAX_SLOTS=1`**: rotation alone suffices — pass 2 is a no-op since quota is filled by pass 1.
- **`MAX_SLOTS>1`**: pass 1 spreads claims across tenants, pass 2 backfills when only one tenant has work.
- **Consolidation slot reservation**: non-consolidation and consolidation pools are capped independently per-schema in pass 1, respecting the existing `_get_available_slots()` split.
- **Worker synchronization**: fresh workers start at offset 0. If multiple workers poll at the same time they initially hit the same schema; `SKIP LOCKED` handles the race cleanly. Offsets diverge over subsequent polls.

## Test plan

New `TestClaimBatchRotation` class in `tests/test_worker.py`:
- [x] Rotation advances past serviced schema (the crucial detail)
- [x] Empty sweep advances by 1 so we don't re-hit the head
- [x] Small tenant not starved by heavy tenant (1 vs 1000 pending)
- [x] `MAX_SLOTS>1` spreads pass-1 claims across tenants
- [x] `MAX_SLOTS>1` backfills in pass 2 when only one tenant has work

All tests in `test_worker.py` pass. Lint, format, type check all clean.

### DIFF
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -130,6 +130,9 @@ class WorkerPoller:
         self._active_tasks: dict[str, ActiveTaskInfo] = {}
         # Track in-flight tasks by operation type
         self._in_flight_by_type: dict[str, int] = {}
+        # Rotation offset for per-tenant fair claiming. Advances past the last
+        # schema we serviced so a busy tenant can't monopolize the poll order.
+        self._next_schema_idx: int = 0
 
     async def _get_schemas(self) -> list[str | None]:
         """Get list of schemas to poll. Returns [None] for default schema (no prefix)."""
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM

PR #1091, opened 2026-04-15
fix: files/retain upload problems and orphaned retains

christerence:
### Summary

Closes a race between the file-convert-retain worker task and the async task backend where a crash or timeout could orphan retain operations permanently.

The bug: _handle_file_convert_retain_task inserted an async_operations row without task_payload, then separately called submit_task to set the payload. If a crash or timeout occurred between those two statements, the row was left with task_payload IS NULL. The worker's claim_batch query filters on task_payload IS NOT NULL, so those orphaned rows became permanently stuck as unclaimed pending tasks — file uploads that silently never completed.

Fix: Build full_retain_payload and serialize it to JSON before the transaction, then include task_payload as a column in the atomic INSERT. The row is now born with its payload, so a crash at any point either leaves no row (transaction rolled back) or a fully-claimable row.

memory_engine.py: _handle_file_convert_retain_task now inserts task_payload in the same INSERT statement that creates the async_operations row. submit_task is still called afterwards for the SyncTaskBackend path (inline execution) and is idempotent for BrokerTaskBackend (payload already set).

### Test plan

- [x] uv run pytest tests/test_http_api_integration.py -v -- file/retain integration tests 
- [x] ./scripts/hooks/lint.sh -- all passed
- [ ] CI green

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -854,26 +854,37 @@ class MemoryEngine(MemoryEngineInterface):
             "file_content_type": task_dict["content_type"],
         }
 
-        # In one transaction: create the retain async operation AND mark this conversion as completed
+        # Include task_payload in the INSERT atomically. Previously this was a
+        # two-step process (INSERT without payload, then UPDATE to set it) which
+        # left null-payload rows when a crash or timeout occurred between the two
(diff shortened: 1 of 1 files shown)


christerence:
Bugs found on our fork for https://www.tryferris.ai/

nicoloboschi (approved):
lgtm, good catch!

PR #1093, opened 2026-04-15
fix(openclaw): resolve full symlink chain in isDirectExecution()

D2758695161:
## Fix: isDirectExecution() misdetects symlinked bin path

**Issue:** #1033

### Root Cause

When a bin symlink points to a CLI wrapper (not directly to the module), process.argv[1] contains the wrapper path after ealpathSync resolution, while import.meta.url points to the actual module file. The original canonicalizeExecutionPath only called ealpathSync once, which is insufficient for chained symlinks.

### Solution

Replace the single-call ealpathSync with an iterative symlink-following function that handles arbitrary-length symlink chains:

\	ypescript
function resolveSymlinks(path: string): string {
  const seen = new Set<string>();
  let current = path;
  while (true) {
    if (seen.has(current)) return current;
    seen.add(current);
    let target: string;
    try { target = realpathSync(current); } catch { return current; }
    if (target === current) return current;
    current = target;
  }
}
\\n
This fix is especially important for environments like OpenClaw where CLI wrappers are common.

### DIFF
--- a/hindsight-integrations/openclaw/src/backfill.ts
+++ b/hindsight-integrations/openclaw/src/backfill.ts
@@ -630,13 +630,26 @@ export async function runCli(argv: string[] = process.argv.slice(2)): Promise<vo
   console.log(args.json ? JSON.stringify(summary, null, 2) : JSON.stringify(summary));
 }
 
+function resolveSymlinks(path: string): string {
+  const seen = new Set<string>();
+  let current = path;
+  while (true) {
+    if (seen.has(current)) return current;
+    seen.add(current);
+    let target: string;
+    try {
+      target = realpathSync(current);
+    } catch {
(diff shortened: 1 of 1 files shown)


D2758695161:
This PR replaces the closed PR #1067 with a cleanly rebased version against current upstream main. The fix adds a `resolveSymlinks()` helper to iteratively follow symlink chains (e.g. `node` -> wrapper script -> actual binary), critical for CLI environments like OpenClaw where bin symlinks commonly point to wrapper scripts.

Fixes #1033.

D2758695161:
Bump - this PR replaces the closed #1067 with a clean rebased version. Ready for review.

nicoloboschi (approved):

PR #1099, opened 2026-04-16
fix(ollama): add think=false to _call_ollama_native payload

### LINKED ISSUE #1098: _call_ollama_native does not pass `think: false` — breaks structured output for reasoning models (qwen3.5, etc.)
## Summary

`_call_ollama_native()` in `openai_compatible_llm.py` (L764) does not include the `think` field in the payload sent to Ollama's `/api/chat`. This causes reasoning models (e.g. qwen3.5) to put their entire response into the `thinking` field, leaving the `response`/`message.content` empty — which breaks structured output (fact extraction, etc.).

## Steps to Reproduce

1. Set `HINDSIGHT_API_RETAIN_LLM_MODEL=qwen3.5:2b` (or any Ollama reasoning model)
2. Ingest a document into a bank
3. Observe that fact extraction returns empty or malformed results

## Root Cause

In `_call_ollama_native()` (L764-783), the payload is constructed without the `think` field:

```python
payload: dict[str, Any] = {
    "model": self.model,
    "messages": messages,
    "stream": False,
    "format": schema,
}
```

Ollama's `/api/chat` supports `"think": false` to disable thinking/reasoning mode. Without it, reasoning models route output to the thinking field instead of the content field.

## Proposed Fix

Add `"think": False` to the payload:

```python
payload: dict[str, Any] = {
    "model": self.model,
    "messages": messages,
    "stream": False,
    "format": schema,
    "think": False,  # Disable thinking for reasoning models
}
```

This is safe for non-reasoning models (e.g. gemma3) — Ollama ignores unknown fields.

## Impact

- **Affected path**: Ollama provider + structured output (`response_format` is not None) — i.e. `_call_ollama_native`
- **Not affected**: OpenAI-compatible path (`/v1/chat/completions`), other providers (Groq, Gemini, Anthropic, etc.)

## Environment

- Hindsight: `ghcr.io/vectorize-io/hindsight:latest`
- Ollama: v0.20.7
- Model: qwen3.5:2b (confirmed), likely affects all reasoning models via Ollama

Happy to submit a PR if this approach looks good.

karl-8888:
## Summary

- Add `"think": False` to the `/api/chat` payload in `_call_ollama_native()` to disable thinking mode for reasoning models

## Problem

Reasoning models (e.g. qwen3.5) route their entire response to the `thinking` field when `think` is not explicitly set to `false`, leaving `message.content` empty. This breaks structured output (fact extraction, etc.) for any Ollama reasoning model used via the native API path.

## Fix

One-line addition to `openai_compatible_llm.py` L764:

```python
payload: dict[str, Any] = {
    "model": self.model,
    "messages": messages,
    "stream": False,
    "think": False,  # Disable thinking for reasoning models (qwen3.5, etc.)
}
```

## Impact

- **Affected**: Ollama provider + structured output (`_call_ollama_native` path)
- **Not affected**: OpenAI-compatible path (`/v1/chat/completions`), other providers
- **Safe for non-reasoning models**: Ollama ignores unknown fields, so gemma3, phi4-mini, etc. are unaffected

## Testing

Verified with qwen3.5:2b via Ollama v0.20.7:
- Without `think: false`: response content is empty (all output goes to thinking field)
- With `think: false`: structured JSON output works correctly, Japanese fact extraction succeeds

Fixes #1098

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -765,6 +765,7 @@ class OpenAICompatibleLLM(LLMInterface):
             "model": self.model,
             "messages": messages,
             "stream": False,
+            "think": False,  # Disable thinking for reasoning models (qwen3.5, etc.)
         }
 
         # Add schema as format parameter for structured output


nicoloboschi (approved):
LGTM

Issue #1006, opened 2026-04-13
[Bug] consolidation_max_slots not enforced as independent pool — consolidation starved when retain queue is active


pversieux:
## Description

The documentation states that `WORKER_CONSOLIDATION_MAX_SLOTS` provides **separate** capacity for consolidation tasks:

> "Slot-based task scheduling, where consolidation operations are limited separately from general tasks, allowing tuning of memory-intensive synthesis operations independently from other background work."

However, in practice, consolidation tasks share the same slot pool as retain tasks and can only claim **leftover** slots after retains are claimed. When retain operations are continuously queued (e.g., multiple active Claude Code sessions with `retainEveryNTurns=3`), consolidation is permanently starved — zero consolidation tasks execute regardless of `consolidation_max_slots` setting.

## Root Cause (poller.py analysis)

Four code paths prevent independent slot allocation:

**1. `_get_available_slots()`** counts consolidation against total:
```python
total_available = max(0, self._max_slots - total_in_flight)  # includes consolidation in total_in_flight
```

**2. `claim_batch()` early return** blocks consolidation when retains fill slots:
```python
if total_available <= 0:  # consolidation_available may be > 0, but never checked
    return []
```

**3. `_claim_batch_for_schema_inner()` limits consolidation** to retain leftovers:
```python
if consolidation_limit > 0 and remaining_limit > 0:  # remaining_limit = 0 when retains fill all slots
    ...
    min(consolidation_limit, remaining_limit)  # capped by retain remainder
```

**4. `operation_type` not in `task_payload`**: consolidation tasks claimed from DB are not tagged with `operation_type` in the task_dict, so `_in_flight_by_type["consolidation"]` is never incremented, making the `consolidation_max_slots` tracking ineffective.

## Steps to Reproduce

1. Configure `WORKER_MAX_SLOTS=10`, `WORKER_CONSOLIDATION_MAX_SLOTS=2`
2. Run multiple sessions with `retainEveryNTurns=3` (or any value generating continuous retains)
3. Observe `WORKER_STATS`: consolidation always shows `0/2` with `available=2` but never claims tasks
4. `pending_consolidation` counter grows indefinitely while observations stagnate

## Expected Behavior

With `max_slots=8` and `consolidation_max_slots=2`, the worker should run up to **10 concurrent tasks**: 8 retain + 2 consolidation, as independent pools. Consolidation should claim its dedicated slots regardless of retain queue depth.

## Suggested Fix

[...]

PR #1109, opened 2026-04-16
fix(worker): scan for active schemas before claiming

cdbartholomew:
## Summary

Fixes a poll performance regression where `claim_batch` walked all tenant schemas (~1500) with expensive `FOR UPDATE SKIP LOCKED` queries on every poll. With the dual-pool break condition from #1006, unfilled pool types prevented the loop from breaking early, causing every poll to iterate all schemas regardless of how few had work.

## Root cause

The break condition `remaining_nc <= 0 AND remaining_cons <= 0` requires both pools to hit zero. When one pool type (e.g., consolidation) has no pending work globally, its counter never reaches zero, so the loop walks every schema looking for work that doesn't exist.

## Fix

Separate **discovery** (which schemas have work?) from **claiming** (take ownership of tasks):

1. `_scan_active_schemas`: calls a server-side PL/pgSQL function (`schemas_with_pending_work()`) that checks all schemas with lightweight `EXISTS` queries in a single DB round-trip. Falls back to per-schema Python queries if the function isn't installed.

2. `claim_batch`: only iterates schemas the scan identified as active. Fairness rotation is preserved — active schemas are visited in rotation order with cap-1-per-pool-per-schema.

## Performance

Measured from a worker pod through pgbouncer with ~1500 tenant schemas:

| Approach | Time |
|---|---|
| Full claim walk (before) | 15.8s |
| Server-side scan + active-only claims (after) | 0.27s |

## Tests

- `test_scan_finds_schemas_with_pending_work` — scan correctly identifies schemas with pending rows using real DB
- `test_claim_batch_only_queries_active_schemas` — end-to-end: scan pre-filters, claim only visits active schemas
- All existing fairness/rotation tests pass unchanged

## Companion changes

The server-side `schemas_with_pending_work()` function is installed via a helm hook in the deployment repo, alongside `total_pending_tasks()`. The worker code falls back gracefully if the function isn't installed.


cdbartholomew:
I have tested this successfully in a live deployment.

nicoloboschi (approved):
LGTM

kimiduck:
@cdbartholomew Hello ! After this PR was merged and we upgraded we got noisy PG errors on our self hosted instance.

```bash
2026-05-01 05:40:49.719 UTC [39] HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

PR #1115, opened 2026-04-17
fix: improve reranker error messages and add configurable TEI timeout

### LINKED ISSUE #1081: Reranker HTTP ReadTimeout under contention: opaque error, hardcoded timeout, permanent stranded memories
## Summary

In v0.5.2 with a self-hosted CPU reranker sidecar, consolidation spikes cause rerank HTTP calls to time out with `httpcore.ReadTimeout`, but three separate issues in the error-handling path combine to make the failure opaque, untunable, and unrecoverable without manual DB intervention. Filing together because all three are on the same failure chain.

## Environment

- `ghcr.io/vectorize-io/hindsight-api:0.5.2-slim`
- LLM: `mistral-small-latest` via Mistral OpenAI-compat endpoint
- Embeddings: Cloudflare Workers AI `@cf/baai/bge-m3` (OpenAI-compat)
- Reranker: self-hosted Infinity v2 (`Slite/mmarco-mMiniLMv2-L6-H384-v1-onnx-o4`), CPU-only (6 vCPU)
- Deployment: docker-compose, single VPS, single user, ~8k memory_units / ~345k memory_links

Key config:
```
HINDSIGHT_API_RERANKER_PROVIDER=cohere
HINDSIGHT_API_RERANKER_COHERE_BASE_URL=http://infinity:7997/rerank
HINDSIGHT_API_RERANKER_MAX_CANDIDATES=300
HINDSIGHT_API_RERANKER_LOCAL_MAX_CONCURRENT=6
HINDSIGHT_API_WORKER_CONSOLIDATION_MAX_SLOTS=1
HINDSIGHT_API_CONSOLIDATION_BATCH_SIZE=15
HINDSIGHT_API_CONSOLIDATION_MAX_ATTEMPTS=3
HINDSIGHT_API_CONSOLIDATION_LLM_MAX_RETRIES=5
```

## Reproduction

1. Deploy v0.5.2-slim with a CPU-bound local reranker (Infinity mmarco)
2. Trigger a consolidation burst (e.g. restart the API while retain operations are queued, or simply run normal usage that produces a batch drain)
3. `_find_related_observations` fans out parallel `recall_async` calls via `asyncio.gather`, each of which hits the reranker HTTP client
4. Infinity serializes rerank internally; httpx client blocks waiting for response headers
5. After ~60s, `httpcore.ReadTimeout` is raised

## Observed traceback

```
2026-04-15 11:35:45 - ERROR - hindsight_api.api.http - [RECALL ERROR]
    bank=claude_code handler_duration=63.827s error=Failed to search memories: 

  File ".../httpcore/_async/http11.py", line 177, in _receive_response_headers
    event = await self._receive_event(timeout=timeout)
  File ".../httpcore/_async/http11.py", line 217, in _receive_event
    data = await self._network_stream.read(...)
httpcore.ReadTimeout

The above exception was the direct cause of the following exception:

  File "/app/api/hindsight_api/engine/memory_engine.py", line 3152, in _search_with_retries
    scored_results = await reranker_instance.rerank(query, merged_candidates)
  File "/app/api/hindsight_api/engine/search/reranking.py", line 213, in rerank
    scores = await self.cross_encoder.predict(pairs)
  File "/app/api/hindsight_api/engine/cross_encoder.py", line 687, in predict
    return await self._http_client.predict(pairs)
  File "/app/api/hindsight_api/engine/cross_encoder.py", line 590, in predict
    response = await self._async_client.post(self.rerank_url, json=body)
```

What the operator actually sees in logs:
```
ERROR - Task 2d5fb945-...failed: Failed to search memories: 
```

No exception type, no timing hint, no context. Everything after the colon is an empty string.

## Issue 1 — Opaque error message

Somewhere in the call chain between `_search_with_retries` and the outer consolidation handler, the error is formatted as roughly:

```python
raise SomeError(f"Failed to search memories: {e}")
```

`str(httpcore.ReadTimeout())` is an empty string, so the message produced is literally `"Failed to search memories: "`. Debugging required scraping raw tracebacks from `docker logs` — and without the traceback, there is no way to distinguish a read timeout from a connection refusal, a DNS failure, or a pgvector error.

**Suggested fix:** include the exception type name, or log with `logger.exception(...)` on the error path:
```python
raise SomeError(
    f"Failed to search memories: {type(e).__name__}: {e}"
    if str(e) else f"Failed to search memories: {type(e).__name__}"
)
```

## Issue 2 — Rerank HTTP client timeout is hardcoded

`hindsight_api/engine/cross_encoder.py` constructs its `httpx.AsyncClient` without exposing the timeout as configuration. Deployments where the reranker is CPU-bound and under contention will hit `ReadTimeout` at ~60s even when the request would have succeeded given more wall-clock budget.

Operators can reduce upstream contention (lower `CONSOLIDATION_BATCH_SIZE`, raise `RERANKER_LOCAL_MAX_CONCURRENT`, etc.) but cannot tune the HTTP timeout itself.

**Suggested fix:** add an env var such as `HINDSIGHT_API_RERANKER_HTTP_TIMEOUT` (seconds, default 60, allow 120–300 for slow backends). Ideally separate `connect_timeout` and `read_timeout`.

## Issue 3 — Failed consolidation permanently strands source memories

When a consolidation task exhausts `CONSOLIDATION_MAX_ATTEMPTS`, source `memory_units` rows that were included in the batch are marked with `consolidation_failed_at = NOW()`. The batch selection query is:

```sql
SELECT ... FROM memory_units
WHERE bank_id = $1
  AND consolidated_at IS NULL
  AND consolidation_failed_at IS NULL
  ...
ORDER BY created_at ASC
LIMIT max_memories_per_batch
```

This permanently excludes those rows from every future consolidation run. There is no TTL, no cooldown reset, no admin endpoint, and no mention of this behavior in the docs. An operator has to run `UPDATE memory_units SET consolidation_failed_at = NULL WHERE ...` by hand to recover.

In our case, a brief Infinity saturation window left 2 memories stranded. They remained searchable via `/recall` (good), but they never participated in any merge/link/dedup analysis until we manually cleared the marker. For a transient infrastructure failure — especially one as common as `ReadTimeout` — permanent exclusion is a surprising and dangerous default.

**Suggested fix** (one of, in order of preference):

1. Auto-clear markers whose underlying cause was a known-transient error class (`ReadTimeout`, `ConnectionError`, `ConnectTimeout`, etc.) — distinguish transient from persistent (`ValidationError`, `ParseError`, …) at the catch site
2. Apply a TTL: `consolidation_failed_at` older than N hours is treated as NULL by the batch selection query
3. Add an admin endpoint `POST /v1/default/banks/{bank_id}/memories/reset-consolidation-failures` (optionally filtered by id, fact_type, or age)
4. At minimum, document the current behavior and the manual recovery SQL in the consolidation section of the docs

## Why file all three together

The full operator-facing failure chain is:

1. Transient infrastructure contention → rerank request exceeds 60s → `httpcore.ReadTimeout`
2. Operator sees empty error `"Failed to search memories: "` — cannot determine cause without enabling debug logging and re-running
3. Retry budget exhausts, op marked `failed` (this part is working correctly, credit to #1042 fix)
4. Source memories silently stranded, no alert, no self-heal, no documentation

Fixing any one of (1), (2), (3) would break the chain. Fixing (2) makes the other two debuggable in production. Fixing (1) gives operators a knob to match their infrastructure. Fixing (3) makes transient failures recoverable — which is the biggest concern, because the current behavior silently degrades data quality over time.

Happy to submit PRs for (1) and (2) if welcome. (3) feels like a design decision where maintainer input would help — particularly on the transient-vs-persistent classification.

octo-patch:
Fixes #1081

## Problem

Two issues with the self-hosted TEI reranker under load (reported by users running CPU rerankers like Infinity v2):

1. **Opaque error messages**: When `httpcore.ReadTimeout` (or any exception whose `str()` is empty) propagates through the recall pipeline, the user sees `"Failed to search memories: "` with no indication of what went wrong. This makes debugging extremely difficult.

2. **Hardcoded HTTP timeout**: The TEI reranker's HTTP timeout was hardcoded in `RemoteTEICrossEncoder.__init__` at 30 seconds with no way to configure it. Under consolidation load, CPU-based rerankers can take longer to respond, causing spurious timeout failures.

## Solution

**Fix 1 — Include exception type in error messages**

Changed the recall error handler in `memory_engine.py` to use `type(e).__name__` alongside `str(e)`. This ensures that even when `str(e)` returns an empty string, the exception class name (e.g. `ReadTimeout`) is visible in the log and in the raised exception message.

Before: `"Failed to search memories: "`
After: `"Failed to search memories: ReadTimeout: "`

**Fix 2 — Add `HINDSIGHT_API_RERANKER_TEI_HTTP_TIMEOUT` env var**

Added a new configuration field following the project's standard config pattern:
- `ENV_RERANKER_TEI_HTTP_TIMEOUT = "HINDSIGHT_API_RERANKER_TEI_HTTP_TIMEOUT"` in `config.py`
- `DEFAULT_RERANKER_TEI_HTTP_TIMEOUT = 30.0` (preserves existing default)
- `reranker_tei_http_timeout: float` field in `HindsightConfig`
- `create_cross_encoder_from_env()` now passes `timeout=config.reranker_tei_http_timeout` to `RemoteTEICrossEncoder`
- Documentation added to `hindsight-docs/docs/developer/configuration.md`

## Testing

- Verified ruff lint passes on all modified files
- Added `test_create_from_env_with_custom_timeout` test to confirm `HINDSIGHT_API_RERANKER_TEI_HTTP_TIMEOUT` is correctly picked up and passed to the encoder
- Existing `test_default_values` and `test_custom_values` tests continue to verify the timeout field works end-to-end

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -238,6 +238,7 @@ ENV_RERANKER_LOCAL_BATCH_SIZE = "HINDSIGHT_API_RERANKER_LOCAL_BATCH_SIZE"
 ENV_RERANKER_TEI_URL = "HINDSIGHT_API_RERANKER_TEI_URL"
 ENV_RERANKER_TEI_BATCH_SIZE = "HINDSIGHT_API_RERANKER_TEI_BATCH_SIZE"
 ENV_RERANKER_TEI_MAX_CONCURRENT = "HINDSIGHT_API_RERANKER_TEI_MAX_CONCURRENT"
+ENV_RERANKER_TEI_HTTP_TIMEOUT = "HINDSIGHT_API_RERANKER_TEI_HTTP_TIMEOUT"
 ENV_RERANKER_MAX_CANDIDATES = "HINDSIGHT_API_RERANKER_MAX_CANDIDATES"
 ENV_RERANKER_FLASHRANK_MODEL = "HINDSIGHT_API_RERANKER_FLASHRANK_MODEL"
 ENV_RERANKER_FLASHRANK_CACHE_DIR = "HINDSIGHT_API_RERANKER_FLASHRANK_CACHE_DIR"
@@ -482,6 +483,7 @@ DEFAULT_RERANKER_LOCAL_BUCKET_BATCHING = False  # Length-sorted bucket batching:
 DEFAULT_RERANKER_LOCAL_BATCH_SIZE = 32  # Batch size for local reranker predict() calls
 DEFAULT_RERANKER_TEI_BATCH_SIZE = 128
 DEFAULT_RERANKER_TEI_MAX_CONCURRENT = 8
+DEFAULT_RERANKER_TEI_HTTP_TIMEOUT = 30.0  # HTTP timeout for TEI reranker requests (seconds)
 DEFAULT_RERANKER_MAX_CANDIDATES = 300
 DEFAULT_RERANKER_FLASHRANK_MODEL = "ms-marco-MiniLM-L-12-v2"  # Best balance of speed and quality
 DEFAULT_RERANKER_FLASHRANK_CACHE_DIR = None  # Use default cache directory
@@ -882,6 +884,7 @@ class HindsightConfig:
     reranker_tei_url: str | None
     reranker_tei_batch_size: int
     reranker_tei_max_concurrent: int
+    reranker_tei_http_timeout: float
     reranker_max_candidates: int
     reranker_cohere_api_key: str | None
     reranker_cohere_model: str
@@ -1434,6 +1437,9 @@ class HindsightConfig:
             reranker_tei_batch_size=int(os.getenv(ENV_RERANKER_TEI_BATCH_SIZE, str(DEFAULT_RERANKER_TEI_BATCH_SIZE))),
             reranker_tei_max_concurrent=int(
                 os.getenv(ENV_RERANKER_TEI_MAX_CONCURRENT, str(DEFAULT_RERANKER_TEI_MAX_CONCURRENT))
+            ),
+            reranker_tei_http_timeout=float(
+                os.getenv(ENV_RERANKER_TEI_HTTP_TIMEOUT, str(DEFAULT_RERANKER_TEI_HTTP_TIMEOUT))
             ),
             reranker_max_candidates=int(os.getenv(ENV_RERANKER_MAX_CANDIDATES, str(DEFAULT_RERANKER_MAX_CANDIDATES))),
             # Cohere reranker (with backward-compatible fallback to shared API key)
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -33,6 +33,7 @@ from ..config import (
     DEFAULT_RERANKER_SILICONFLOW_BASE_URL,
     DEFAULT_RERANKER_SILICONFLOW_MODEL,
     DEFAULT_RERANKER_TEI_BATCH_SIZE,
+    DEFAULT_RERANKER_TEI_HTTP_TIMEOUT,
     DEFAULT_RERANKER_TEI_MAX_CONCURRENT,
     DEFAULT_RERANKER_ZEROENTROPY_MODEL,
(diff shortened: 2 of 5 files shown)


nicoloboschi (approved):
LGTM

PR #1119, opened 2026-04-17
fix(integrations): add encoding="utf-8" to transcript file reads

tordf:
On Windows, open() defaults to the system locale encoding (cp1252) instead of UTF-8. Claude Code and Codex transcript JSONL files contain UTF-8 bytes (e.g. 0x9d) that are invalid in cp1252, causing UnicodeDecodeError in the auto-retain and auto-recall hooks. This silently prevented all transcript processing on Windows.

Affected files:
- claude-code/scripts/retain.py (read_transcript)
- claude-code/scripts/recall.py (read_transcript_messages)
- codex/scripts/lib/content.py (_read_transcript_text, _read_transcript_rich)

### DIFF
--- a/hindsight-integrations/claude-code/scripts/recall.py
+++ b/hindsight-integrations/claude-code/scripts/recall.py
@@ -53,7 +53,7 @@ def read_transcript_messages(transcript_path: str) -> list:
         return []
     messages = []
     try:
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
LGTM

Issue #677, opened 2026-03-25
hindsight-hermes plugin - Async Context Deadlock in `hindsight_client`


diegohb:
### Bug Description

The `hindsight_client` library's sync-async bridge (`_run_async()`) fails when called from an already-running async context, causing all Hindsight tools (`hindsight_retain`, `hindsight_recall`, `hindsight_reflect`) to crash in Hermes gateway mode.

**Impact:** Memory operations completely broken in Discord/Telegram mode (async contexts), but work fine in CLI mode (sync context).

### Setup

```bash
# Install packages
pip install hindsight-client hindsight-hermes

# Set environment
export HINDSIGHT_API_URL=http://localhost:8888
export HINDSIGHT_BANK_ID=zeek-hermes
export HINDSIGHT_BUDGET=mid

# Start Hindsight server
hindsight serve  # Runs on localhost:8888
```

### Test Case 1: Sync Context (CLI) - ✅ WORKS

```python
from hindsight_client import HindsightClient

client = HindsightClient()
result = client.recall("test query")
print(result)  # ✅ Returns results successfully
```

**Result:** ✅ Works fine, returns memories.

---

### Test Case 2: Async Context - ❌ FAILS

```python
import asyncio
from hindsight_client import HindsightClient

async def main():
    client = HindsightClient()
    result = await client.recall("test query")  # ❌ RuntimeError
    return result

asyncio.run(main())
```

**Result:** ❌ Crashes with `RuntimeError: This event loop is already running`

---

### Test Case 3: Hermes Gateway Mode - ❌ FAILS

1. Start Hermes Discord/Telegram gateway (runs in async context)
2. Send message: `/recall test` or `/retain test data`
3. Observe tool execution

**Result:** ❌ All Hindsight tools crash with `RuntimeError`

**Actual Error from Gateway:**
```
Error calling tool hindsight_recall: RuntimeError: This event loop is already running
```

### Expected Behavior

- ✅ Hindsight tools should work in **both** sync (CLI) and async (gateway) contexts
- ✅ No event loop nesting errors
- ✅ Seamless integration with asyncio-based applications (Hermes gateway, Discord bots, etc.)

### Actual Behavior

| Context | Status | Reason |
|---------|--------|--------|
| **CLI mode** (sync) | ✅ Works | No running event loop, new loop created |
| **Gateway mode** (async) | ❌ Crashes | `run_until_complete()` on running loop forbidden |

### Version

| Component | Value |
|-----------|-------|
[...]

PR #1122, opened 2026-04-17
fix(hermes): add is_async=True to async tool handler registration

ClawPrime:
## Problem

Tool handlers in `hindsight_hermes/tools.py` are defined as `async def` but registered without `is_async=True`. This causes the agent framework's tool registry to dispatch them as sync functions (the default), which returns coroutine objects instead of awaiting them — resulting in silent failures where retain/recall/reflect appear to succeed but produce no actual results.

This bug exists in both:
- `register_tools()` — uses `registry.register()`
- `register()` — uses `ctx.register_tool()`

## Root Cause

The hermes agent tool registry distinguishes between sync and async handlers via the `is_async` parameter. When `is_async=False` (default), the dispatcher calls `handler(args)` directly and returns whatever it gets. For async handlers, this returns a coroutine object that is never awaited.

## Fix

Adds `is_async=True` to all six `register_tool()` / `registry.register()` calls in `hindsight_hermes/tools.py`. This enables the registry to properly route async handlers through its event loop bridging.

## Relationship to PR #677

PR #677 fixes the underlying client method issue (switching from sync wrappers that call `loop.run_until_complete()` to native async methods). This PR fixes the _registration layer_ that #677 doesn't address. Both fixes are needed — #677 makes the handlers correct, this PR makes them get awaited.

Without this fix, even after #677 merges, the handlers would return coroutine objects that are never awaited, causing retain/recall/reflect to silently fail.

## Testing

Verified in production on Claw Prime AI's Hermes agent:
- Before fix: `hindsight_retain` returns `{'error': '...coroutine...'}` or silent failure
- After fix: All three tools (retain, recall, reflect) work correctly


nicoloboschi:
pls do it on the Hermes repository

PR #1124, opened 2026-04-17
feat(worker): add opt-in stale foreign task recovery

### LINKED ISSUE #1107: Dead worker-owned async_operations are not reclaimed automatically, which can block consolidation indefinitely
## Summary

If a worker dies while holding an `async_operations` row in `status = 'processing'`, that task can remain stuck indefinitely when a replacement worker starts with a different `worker_id`.

This appears to block consolidation for the affected bank until someone manually clears the stale task.

## Observed behavior

A healthy replacement worker continues polling normally, but worker stats still show a foreign worker holding one task. Consolidation work remains stuck until the stale task ownership is manually cleared.

Sanitized example:

```text
[WORKER_STATS] ... others: dead-worker-id:1
[PENDING_BREAKDOWN] ... consolidation: total=1 claimable=1 ...
```

After manually releasing the stale task, the active worker immediately claims consolidation work again, and worker stats change to:

```text
[WORKER_STATS] ... others: none
```

## Why this seems to happen

The startup recovery path only reclaims tasks owned by the current worker ID.

In `hindsight_api/worker/poller.py`, `recover_own_tasks()` resets tasks where the `worker_id` matches the current worker. That means:

- if the same worker identity comes back, it can recover its own abandoned tasks
- if a worker pod dies and a new pod starts with a different `worker_id`, tasks owned by the dead worker are never reclaimed automatically

## Why this is especially painful

This is particularly problematic for consolidation tasks. A stale `processing` consolidation row can effectively jam consolidation for a bank until there is manual intervention.

In practice, this can look like:

- backlog growth
- healthy workers that appear underutilized
- repeated worker stats showing a foreign worker in `others: ...`
- claimable consolidation work that does not clear as expected

## Manual workaround

Manually release tasks owned by the dead worker, for example:

```sql
UPDATE public.async_operations
SET status = 'pending',
    worker_id = NULL,
    claimed_at = NULL,
    updated_at = now()
WHERE worker_id = '<dead-worker-id>'
  AND status = 'processing';
```

If available, the admin `decommission-worker` path also appears intended for this kind of recovery.

## Suggested fixes

One or more of these would help:

1. Automatically reclaim tasks owned by worker IDs that are no longer live
2. Add lease expiration based on `claimed_at`, so stale `processing` tasks can be safely requeued
3. Add targeted stale-task recovery for consolidation tasks, since one stuck row can block an entire lane of work

## Nice to have

- an admin or API view that compares live worker IDs vs task-owned worker IDs
- warnings when `processing` tasks are owned by non-live workers beyond a reasonable threshold
- optional automatic stale-task cleanup after a configurable timeout

Aldoustheorchestrator:
## Summary

This adds an opt-in fallback for reclaiming stale `processing` tasks owned by dead or foreign workers.

It is aimed at environments where worker identities are ephemeral, especially orchestrated systems like Kubernetes, where tying correctness to static worker IDs is often not a good operational fit.

This is intentionally additive and off by default. It does not change or interfere with the current static-worker-ID approach for deployments that prefer it.

Fixes #1107.

## What this changes

Two new worker settings are introduced:

- `HINDSIGHT_API_WORKER_RECLAIM_STALE_TASKS_ENABLED=false`
- `HINDSIGHT_API_WORKER_RECLAIM_STALE_TASKS_AFTER_SECONDS=1800`

When enabled, a worker will, on startup, release foreign `processing` tasks back to `pending` when all of the following are true:

- the task is owned by a different worker
- `claimed_at` is older than the configured threshold
- the task is not part of a batch-backed operation

The existing `recover_own_tasks()` path remains unchanged.

## Why this helps

In orchestrated environments, worker names and identities are frequently derived from pod identity and can change across restarts or rescheduling. In that world, requiring static worker IDs as the main recovery strategy is brittle.

This change provides a conservative fallback for those deployments without taking anything away from users who want explicitly stable worker IDs.

## Safety / scope

This change is intentionally narrow:

- off by default
- startup-only behavior
- age-gated
- excludes batch-backed operations
- leaves the current static-ID workflow intact

## Tests and docs

This PR adds tests for:

- opt-in behavior being disabled by default
- reclaiming old foreign tasks when enabled
- respecting the age threshold
- excluding batch-backed operations

It also documents the new worker settings.

## Longer-term direction

I think a more durable long-term solution would be to improve the worker identity / lease model itself, so recovery does not depend on static IDs or manual decommissioning in ephemeral environments. This PR is meant as a pragmatic fallback, not the final word on worker ownership semantics.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2587,6 +2587,8 @@ def create_app(
                 tenant_extension=memory._tenant_extension,
                 max_slots=config.worker_max_slots,
                 consolidation_max_slots=config.worker_consolidation_max_slots,
+                reclaim_stale_tasks_enabled=config.worker_reclaim_stale_tasks_enabled,
+                reclaim_stale_tasks_after_seconds=config.worker_reclaim_stale_tasks_after_seconds,
             )
             poller_task = asyncio.create_task(poller.run())
             logging.info(f"Worker poller started (worker_id={worker_id})")
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -382,6 +382,8 @@ ENV_WORKER_MAX_RETRIES = "HINDSIGHT_API_WORKER_MAX_RETRIES"
 ENV_WORKER_HTTP_PORT = "HINDSIGHT_API_WORKER_HTTP_PORT"
 ENV_WORKER_MAX_SLOTS = "HINDSIGHT_API_WORKER_MAX_SLOTS"
 ENV_WORKER_CONSOLIDATION_MAX_SLOTS = "HINDSIGHT_API_WORKER_CONSOLIDATION_MAX_SLOTS"
+ENV_WORKER_RECLAIM_STALE_TASKS_ENABLED = "HINDSIGHT_API_WORKER_RECLAIM_STALE_TASKS_ENABLED"
+ENV_WORKER_RECLAIM_STALE_TASKS_AFTER_SECONDS = "HINDSIGHT_API_WORKER_RECLAIM_STALE_TASKS_AFTER_SECONDS"
 ENV_RETAIN_MAX_CONCURRENT = "HINDSIGHT_API_RETAIN_MAX_CONCURRENT"
 
 # Reflect agent settings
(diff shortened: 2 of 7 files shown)


nicoloboschi:
I'd prefer to not give wrong instructions to users - this solution doesn't help understanding the root cause of the issue so i think it's the wrong direction -- I understand the why and the reason behind it, but not every user is capable / have time to actually look into the actual issue in their infra and this change would give them a fast-and-wrong solution

PR #1141, opened 2026-04-18
fix(integrations): preserve raw UTF-8 in dynamically-derived bank_id

Desko77:
## Problem

When `dynamicBankId: true`, `derive_bank_id()` runs each granularity
segment through `urllib.parse.quote(..., safe="")` (Python integrations)
or `encodeURIComponent()` (TypeScript integrations) before joining them
with `::`. The resulting percent-encoded string is then used as the bank
**identifier itself** — not just as a URL component.

For a project folder named, say, `мой проект`, the bank_id stored on the
server becomes:

```
claude-code::%D0%BC%D0%BE%D0%B9%20%D0%BF%D1%80%D0%BE%D0%B5%D0%BA%D1%82
```

which is unreadable in the Hindsight UI and in CLI listings.

## Root cause

Encoding happens at bank_id **construction** time, not at **transport**
time. The `client.py` / `client.ts` layer already URL-encodes `bank_id`
correctly when building the request path — that encoding is required by
HTTP and is left in place. The API server (Starlette) decodes the path
parameter back to raw UTF-8 before passing it to handlers, and the DB
(`banks.bank_id TEXT` in `hindsight-api-slim/hindsight_api/alembic/versions/5a366d414dce_initial_schema.py`)
has no character-class constraints. So the DB stores whatever the handler
receives — which has been percent-encoded by the client twice over.

## Fix

Stop URL-encoding segments inside `derive_bank_id()`. Keep encoding in the
client/transport layer. `bank_id` values are now stored as raw UTF-8.

Changed:

- `hindsight-integrations/claude-code/scripts/lib/bank.py`
- `hindsight-integrations/codex/scripts/lib/bank.py`
- `hindsight-integrations/opencode/src/bank.ts`
- Tests updated; added a UTF-8 case in each.
- Added `hindsight-integrations/codex/tests/test_bank.py` (no bank tests
  existed before).

Not changed (intentional):

- `client.py` / client-transport `urllib.parse.quote(bank_id, safe="")`
  and `encodeURIComponent(bankId)` usages — those build HTTP path and
  must stay.
- `hindsight-integrations/paperclip` — already correct (doesn't encode
  at construction time).

## Breaking change

Users on `dynamicBankId: true` will start writing to a new bank name
after upgrading. Existing percent-encoded banks remain untouched on the
server and can be renamed via admin API / SQL, or left as archive.

Version bumps reflect this:

- `claude-code`: 0.3.1 → 0.4.0
- `codex`: 0.2.1 → 0.3.0
- `opencode`: 0.1.4 → 0.2.0

## Verification

- `pytest hindsight-integrations/claude-code/tests/test_bank.py` — 19/19 passed
- `pytest hindsight-integrations/codex/tests/test_bank.py` — 12/12 passed
- `npm test` in `hindsight-integrations/opencode` — 92/92 passed
- End-to-end: with a folder name `мой проект`, `list_banks` now returns
  the bank under its readable name instead of a percent-encoded blob.

### DIFF
--- a/hindsight-integrations/claude-code/.claude-plugin/plugin.json
+++ b/hindsight-integrations/claude-code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hindsight-memory",
   "description": "Automatic long-term memory for Claude Code via Hindsight. Recalls relevant memories before each prompt and retains conversation transcripts after each response.",
-  "version": "0.3.1",
+  "version": "0.4.0",
   "author": {"name": "Hindsight Team", "url": "https://vectorize.io/hindsight"},
   "license": "MIT",
   "keywords": ["memory", "hindsight", "recall", "retain"]
--- a/hindsight-integrations/claude-code/.claude-plugin/plugin.json
+++ b/hindsight-integrations/claude-code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hindsight-memory",
   "description": "Automatic long-term memory for Claude Code via Hindsight. Recalls relevant memories before each prompt and retains conversation transcripts after each response.",
-  "version": "0.4.0",
+  "version": "0.3.1",
   "author": {"name": "Hindsight Team", "url": "https://vectorize.io/hindsight"},
   "license": "MIT",
   "keywords": ["memory", "hindsight", "recall", "retain"]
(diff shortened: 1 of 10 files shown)


nicoloboschi (changes requested):
please don't change the versions and changelog , the others look good!

Desko77:
Done — reverted in `27a71113`. Version bumps and the CHANGELOG entry dropped; the code fix in `bank.py` / `bank.ts` and the test updates are left as-is. Let me know if anything else needs adjusting.

nicoloboschi (approved):

PR #1143, opened 2026-04-18
feat(embeddings): make OpenAI-compatible batch size configurable (#1142)

### LINKED ISSUE #1142: bug: OpenAI-compatible embeddings batch size is hardcoded to 100 and not configurable
## Bug Description
When using `HINDSIGHT_API_EMBEDDINGS_PROVIDER=openai` with `HINDSIGHT_API_EMBEDDINGS_OPENAI_BASE_URL` pointing to an OpenAI-compatible embeddings endpoint, `batch_retain` can fail if the upstream endpoint accepts fewer than 100 inputs per request.

In my case, the upstream endpoint rejects requests larger than 10 items and returns:

```text
InvalidParameter: Value error, batch size is invalid, it should not be larger than 10.: ***.contents
```

The root cause appears to be that `OpenAIEmbeddings` defaults to `batch_size=100`, and `create_embeddings_from_env()` does not expose any config/env override for embedding batch size. So any retain path that extracts more than 10 facts in one embedding pass will fail against stricter OpenAI-compatible providers.

## Steps to Reproduce
1. Configure embeddings with:
   - `HINDSIGHT_API_EMBEDDINGS_PROVIDER=openai`
   - `HINDSIGHT_API_EMBEDDINGS_OPENAI_BASE_URL=<openai-compatible-endpoint>`
2. Point that endpoint at a backend that enforces a maximum embedding batch size smaller than 100 (for example, 10 items).
3. Send a `batch_retain` / retain request whose fact extraction yields more than that limit.
4. Observe a 400 from the embedding call.

## Expected Behavior
One of these should happen:
- embedding batch size should be configurable via env/config, or
- Hindsight should automatically split/retry embedding requests when the provider limit is lower than the built-in default.

## Actual Behavior
Hindsight sends embedding requests in batches of 100 for the OpenAI-compatible path, and the request fails if the upstream limit is smaller.

The relevant code path looks like this:
- `hindsight_api/engine/embeddings.py`: `OpenAIEmbeddings.__init__(..., batch_size: int = 100, ...)`
- `hindsight_api/engine/embeddings.py`: `create_embeddings_from_env()` creates `OpenAIEmbeddings(...)` without passing any batch size override
- `hindsight_api/engine/embeddings.py`: `encode()` slices `texts` by `self.batch_size` and sends them as one `input=batch` request

For OpenAI-compatible gateways that translate `input` to provider-native `contents`, the upstream error shows up as a `contents` limit violation.

## Version
`0.5.2` (`7f81a0ad` locally)

## Notes
I searched existing issues and only found related-but-different reports such as `#1130` and `#1037`; I did not find one covering provider-specific embedding batch limits on the OpenAI-compatible embeddings path.

r266-tech:
## Summary

Fixes #1142.

`OpenAIEmbeddings` hardcodes `batch_size=100`, which is incompatible with OpenAI-compatible providers that enforce a smaller per-request limit (DashScope / Aliyun Tongyi caps at 10 inputs). Without an override, any retain path that extracts more than the upstream limit fails the entire batch with a 400 `batch size is invalid` error.

This PR exposes `HINDSIGHT_API_EMBEDDINGS_OPENAI_BATCH_SIZE` (default `100`, preserving legacy behavior) and propagates it to both the `openai` and `openrouter` providers (which share `OpenAIEmbeddings`).

## Changes

- `config.py`: new `ENV_EMBEDDINGS_OPENAI_BATCH_SIZE` / `DEFAULT_EMBEDDINGS_OPENAI_BATCH_SIZE=100`; new `embeddings_openai_batch_size: int` field on `HindsightConfig` (with dataclass default so existing direct constructors keep working); new `_parse_positive_int` helper that rejects `0`, negative, and non-integer values at config load (otherwise `range(0, N, 0)` would crash or batching would silently be skipped).
- `engine/embeddings.py`: `create_embeddings_from_env()` passes `batch_size=config.embeddings_openai_batch_size` to both the `openai` and `openrouter` branches.
- `docs/developer/configuration.md` (+ skill reference): document the new env var.

## Test plan

- [x] New `tests/test_embeddings_openai_batch_size.py` (8 tests):
  - default is 100 when env unset
  - env override is read
  - `openai` provider wiring propagates the value
  - `openrouter` provider wiring propagates the value
  - `0` / negative / non-numeric all raise `ValueError` with clear messages pointing at the env var
  - behavioral: 25 inputs with `batch_size=10` produces upstream calls of `[10, 10, 5]`
- [x] `ruff check` + `ruff format --check` clean
- [x] Existing `test_config_validation`, `test_gemini_embeddings`, `test_google_cross_encoder`, `test_litellm_sdk_embeddings` all still pass (no regression on dynamic `HindsightConfig(**fields)` test helpers)

## Notes

- OpenRouter currently uses the same config key as OpenAI because both paths share the `OpenAIEmbeddings` client. Happy to split into a dedicated `HINDSIGHT_API_EMBEDDINGS_OPENROUTER_BATCH_SIZE` if you'd rather decouple; for now users who only touch OpenAI-compatible endpoints get a single knob.
- Versioned docs under `versioned_docs/` intentionally not touched (those are frozen snapshots of shipped releases).

Reported and root-caused by @eightHundreds.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -177,6 +177,7 @@ ENV_EMBEDDINGS_TEI_URL = "HINDSIGHT_API_EMBEDDINGS_TEI_URL"
 ENV_EMBEDDINGS_OPENAI_API_KEY = "HINDSIGHT_API_EMBEDDINGS_OPENAI_API_KEY"
 ENV_EMBEDDINGS_OPENAI_MODEL = "HINDSIGHT_API_EMBEDDINGS_OPENAI_MODEL"
 ENV_EMBEDDINGS_OPENAI_BASE_URL = "HINDSIGHT_API_EMBEDDINGS_OPENAI_BASE_URL"
+ENV_EMBEDDINGS_OPENAI_BATCH_SIZE = "HINDSIGHT_API_EMBEDDINGS_OPENAI_BATCH_SIZE"
 
 # Gemini/Vertex AI embeddings configuration
 ENV_EMBEDDINGS_GEMINI_API_KEY = "HINDSIGHT_API_EMBEDDINGS_GEMINI_API_KEY"
@@ -468,6 +469,7 @@ DEFAULT_EMBEDDINGS_LOCAL_MODEL = "BAAI/bge-small-en-v1.5"
 DEFAULT_EMBEDDINGS_LOCAL_FORCE_CPU = False  # Force CPU mode for local embeddings (avoids MPS/XPC issues on macOS)
 DEFAULT_EMBEDDINGS_LOCAL_TRUST_REMOTE_CODE = False  # Security: disabled by default, required for some models
 DEFAULT_EMBEDDINGS_OPENAI_MODEL = "text-embedding-3-small"
+DEFAULT_EMBEDDINGS_OPENAI_BATCH_SIZE = 100
 DEFAULT_EMBEDDINGS_GEMINI_MODEL = "gemini-embedding-001"
 DEFAULT_EMBEDDINGS_GEMINI_OUTPUT_DIMENSIONALITY = 768
 DEFAULT_EMBEDDING_DIMENSION = 384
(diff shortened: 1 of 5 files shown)


nicoloboschi (approved):
LGTM, thanks

PR #1161, opened 2026-04-20
feat(claude-code): add {user_id} retainTags template variable

### LINKED ISSUE #1160: feat(claude-code): add {user_id} retainTags template variable from HINDSIGHT_USER_ID env
## Background

`hindsight-integrations/claude-code` resolves template variables in `retainTags` and `retainMetadata` at retain time (`scripts/retain.py`). Today: `{session_id}`, `{bank_id}`, `{timestamp}`. Tags attached here become `recall` filter keys, so `retainTags` is the natural surface for identity namespaces like `user:<id>`.

## Problem

None of the existing variables reflects the human user. Hardcoding `"retainTags": ["user:alice"]` works for a single machine but breaks as soon as `settings.json` is distributed via dotfiles across accounts — the value has to come from the environment.

Falling back to a blank string is not safe either: `"user:{user_id}"` would resolve to `"user:"` when the env is unset and pollute the bank with a dangling-namespace tag that matches no real identity.

## Proposed Solution

1. Add `{user_id}` as a fourth template variable, resolved from `HINDSIGHT_USER_ID` (empty string if unset). Applies to both `retainTags` and `retainMetadata`.
2. After resolution, drop any tag of the form `"<ns>:"` with an empty post-colon segment. Tags without `:` are untouched — scoped guard, namespaced patterns only.
3. If every tag drops, send no `tags` field (rather than an empty list).

Result: one portable config works everywhere.

```json
{ "retainTags": ["user:{user_id}", "session:{session_id}"] }
```

Set `HINDSIGHT_USER_ID` per shell for per-user scoping; leave it unset to fall back to session-only tagging.

soichisumi:
## Summary

- Adds `{user_id}` template variable to `retainTags` / `retainMetadata`, resolved from `HINDSIGHT_USER_ID`.
- After resolution, drops tags of the form `"<ns>:"` (empty post-colon segment). Tags without `:` are untouched.
- Four new tests covering env-set / env-unset-drop / colon-less-preserved / all-dropped-yields-no-field. Full suite: 133 passed.

Fixes #1160.

## What

- `scripts/retain.py`: extend `template_vars` with `"user_id": os.environ.get("HINDSIGHT_USER_ID", "")`, rewrite the tag-resolution loop to drop dangling namespaced tags (logged via `debug_log`).
- `tests/test_hooks.py`: four new cases in `TestRetainHook`. `{user_id}` is injected via `_run_hook`'s `extra_env` (the helper strips real `HINDSIGHT_*` vars by design).
- `README.md`: expanded `retainTags` description, template variables reference table, per-user scoping example.
- `CHANGELOG.md`: `[Unreleased]` Added / Changed entries.

## How

`template_vars["user_id"]` is resolved once and applies to both `retainTags` and `retainMetadata` because `_resolve_template` is shared.

After `_resolve_template`, the loop skips any tag where `":" in resolved and resolved.split(":", 1)[1] == ""`. Colon-less tags never match the guard, so backward compat is preserved by construction. If every tag drops, `tags` is normalized to `None`; `HindsightClient.retain`'s existing field-omission path (`client.py:144` — `if tags: item["tags"] = tags`) then omits the field from the outgoing request.

## Test plan

- [x] [auto] `pytest tests/test_hooks.py -v -k retain_tag` → 5 passed (1 existing + 4 new).
- [x] [auto] `pytest tests/ -v` → 133 passed, no regression.
- [ ] [manual] CI green on this PR.

### DIFF
--- a/hindsight-integrations/claude-code/CHANGELOG.md
+++ b/hindsight-integrations/claude-code/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- `{user_id}` template variable for `retainTags` and `retainMetadata`, resolved
+  from the `HINDSIGHT_USER_ID` env var (empty string if unset). Enables
+  machine-independent per-user memory scoping without hardcoding user ids in
+  `settings.json`.
+
+### Changed
+
+- Tags that resolve to an empty namespace content (e.g. `"user:"` when
+  `HINDSIGHT_USER_ID` is unset) are now dropped from retain requests. Previously
+  such tags were sent as-is. Tags without `:` are unaffected.
+
 ## [0.1.0] - 2025-03-23
 
 ### Added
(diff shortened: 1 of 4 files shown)


nicoloboschi (approved):
LGTM, thanks!

PR #1185, opened 2026-04-21
fix(codex): fix encoding with PowerShell

jerviscui:
Resolves two issues:
~~1. Windows installation script `get-codex.ps1`~~
2. Forces PowerShell to use UTF-8 encoding to prevent Recall API exceptions caused by the default system character encoding.


nicoloboschi (changes requested):
ok for the encoding changes, but not for the ps script to install codex, it shouldn't live in our repo

jerviscui:
> ok for the encoding changes, but not for the ps script to install codex, it shouldn't live in our repo

removed the ps script

nicoloboschi (approved):

Issue #1042, opened 2026-04-14
LLM call retry budget for consolidation is unbounded and unconfigurable (up to 33 calls per batch)


ChampPABA:
## Summary

A single `consolidation` batch can make up to **33 Gemini API calls** during an upstream outage before giving up, and neither the inner nor the outer retry count is configurable via environment variables. Combined with the slot-holding behaviour of the worker (see #991), this turns a transient upstream `503` into a multi-hour outage of the consolidation pipeline.

## The math, with upstream defaults

Two nested retry layers, both using upstream defaults:

1. **Outer loop** in `engine/consolidation/consolidator.py` (~L1185–L1212):

   ```python
   max_attempts = 3  # hardcoded
   for attempt in range(1, max_attempts + 1):
       try:
           response = await llm_config.call(..., scope="consolidation")
           ...
       except Exception as exc:
           logger.warning(f"[CONSOLIDATION] LLM batch call failed (attempt {attempt}/{max_attempts}): {exc}")
   ```

2. **Inner loop** in `engine/providers/gemini_llm.py::GeminiLLM.call` (L165, L245):

   ```python
   async def call(
       self,
       ...
       max_retries: int = 10,   # default
       ...
   ):
       ...
       for attempt in range(max_retries + 1):   # 11 attempts total
           ...
   ```

   The consolidation call site (`consolidator.py` ~L1192) passes `scope="consolidation"` but does **not** override `max_retries`, so the default (`10` → 11 attempts) is used.

Worst case per batch: **3 × 11 = 33 Gemini API calls**, and there is no wall-clock budget to stop earlier.

## Why the existing env var doesn't help

`HINDSIGHT_API_LLM_MAX_RETRIES` (`config.py` L128, L1167) is wired into `engine/retain/fact_extraction.py` but **not** into the consolidation path. So operators who notice the retry storm in logs have no knob to turn — the only option is a code patch.

## Observed impact

During a ~90-minute Gemini `503 UNAVAILABLE` episode on `gemini-3.1-flash-lite-preview`:

- Two `consolidation` ops stayed in `status='processing'` for >2 hours each, cycling through `llm.gemini.consolidation.attempt=1/11 … 11/11` multiple times under the outer loop.
- Worker slots pinned at `slots=2/10 (consolidation=2/2)` the entire time.
- DB connection pool saturated to `waiters=98` (recall handlers starving).
[...]

PR #1188, opened 2026-04-21
feat(api): expose retry_count and next_retry_at on operation responses

cdbartholomew:
## Summary

The \`async_operations\` table has tracked \`retry_count\` and \`next_retry_at\` for every task since the worker poller was introduced, but neither is exposed through the generic list / status endpoints or plumbed through to validator extensions. That leaves two consumers unable to distinguish a freshly-queued pending task from one parked for a future retry:

- **API clients** watching task state (UI, SDKs) — a pending task could be awaiting immediate pickup or be parked hours out; there's no way to tell.
- **Validator extensions** deciding when to reschedule a task — without access to retry_count, they can't compute per-attempt backoff without a separate DB round-trip.

This aligns \`OperationResponse\` and \`OperationStatusResponse\` with the pattern already used by \`WebhookDeliveryResponse\` (which has exposed both fields since #1042).

## Changes

- Add \`retry_count: int = 0\` and \`next_retry_at: str | None = None\` to \`OperationResponse\` and \`OperationStatusResponse\`. Completed tasks carry \`next_retry_at=null\`; a pending task with \`next_retry_at\` in the future signals the task is parked rather than awaiting immediate pickup.
- \`list_operations\` and \`get_operation_status\`: include the columns in their SELECT, emit as ISO-8601.
- Add \`retry_count: int = 0\` to \`RequestContext\`. Worker task handlers (\`_handle_batch_retain\`, \`_handle_file_convert\`, \`_handle_consolidation\`, \`_handle_refresh_mental_model\`) populate it from \`task_dict["_retry_count"]\` before dispatching. Defaults to \`0\` for sync/HTTP requests, so no caller-side change is required.

## Test plan

- [x] \`test_operation_status_exposes_retry_count_and_next_retry_at\` — both list and status endpoints expose the fields; a future \`next_retry_at\` round-trips through the API as ISO-8601 within 1-second tolerance
- [x] \`test_request_context_retry_count_propagated_to_validator\` — validator sees \`ctx.request_context.retry_count == task_dict["_retry_count"]\`, including the default-0 case when the key is absent
- [x] Full \`test_async_batch_retain\` suite passes (14 tests)

## Compatibility

Both field additions are fully backwards-compatible: they have defaults, clients that don't read them see no change. Extensions that don't populate \`retry_count\` on \`RequestContext\` continue to see \`0\` via the default.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2119,6 +2119,8 @@ class OperationResponse(BaseModel):
                 "created_at": "2024-01-15T10:30:00Z",
                 "status": "pending",
                 "error_message": None,
+                "retry_count": 0,
+                "next_retry_at": None,
             }
         }
     )
@@ -2130,6 +2132,20 @@ class OperationResponse(BaseModel):
     created_at: str
     status: str
     error_message: str | None
+    retry_count: int = Field(
+        default=0,
+        description="Number of times this operation has been retried after failure.",
+    )
+    next_retry_at: str | None = Field(
+        default=None,
+        description=(
+            "When the worker will next attempt this operation. For a pending "
+            "operation, a value in the future indicates the task is waiting "
+            "rather than available for immediate pickup — for example, an "
+            "extension may have raised DeferOperation to park the task until "
+            "some backpressure window opens. Always null for completed tasks."
+        ),
+    )
 
 
 class ConsolidationResponse(BaseModel):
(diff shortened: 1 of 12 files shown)


hindsight-api-slim/hindsight_api/api/http.py
```diff
@@ -2239,6 +2255,19 @@ class OperationStatusResponse(BaseModel):
     updated_at: str | None = None
     completed_at: str | None = None
     error_message: str | None = None
+    retry_count: int = Field(
```
nicoloboschi: must be optional field to be backwards compatible with older clients

nicoloboschi (commented):
check my comment and also fix the verify-generated-files job please by running the commands

hindsight-api-slim/hindsight_api/api/http.py
cdbartholomew: Done in 8cc9a111 — both `retry_count` fields (`OperationResponse` and `OperationStatusResponse`) are now `int | None = Field(default=None, ...)`, so generated clients treat the field as nullable/optional and can decode responses from older servers that don't populate it. Runtime behavior unchanged: the SQL selects `retry_count` from a NOT NULL DEFAULT 0 column, so the server always writes a real integer. Also regenerated OpenAPI + Python/TS/Go clients in the same commit — `verify-generated-files` is now green.

PR #1189, opened 2026-04-22
fix: allow reflect-specific LLM config when default is disabled

zwcf5200:
## Summary

Allow reflect to use its own configured LLM provider even when the default memory
LLM provider is disabled.

Previously, reflect was blocked when the default LLM provider was `none`, even if
`reflect_llm_provider` was configured separately.

## Tests

- `uv run ruff check hindsight-api-slim/hindsight_api/engine/memory_engine.py
hindsight-api-slim/tests/test_per_operation_llm_config.py`
- `uv run pytest hindsight-api-slim/tests/
test_per_operation_llm_config.py::TestReflectUsesReflectLLMConfig -n0`

Note: the full `test_per_operation_llm_config.py` file currently has existing
failures on retry default expectations unrelated to this change.


nicoloboschi (approved):
nice catch!

PR #1197, opened 2026-04-22
fix: auto-confirm control plane install on first UI launch

### LINKED ISSUE #1196: hindsight-embed ui start hides npx install prompt and times out on first launch
## Bug description
On macOS, `hindsight-embed -p hermes ui start` can time out on the **first** launch even though the embedded daemon itself is healthy.

The root cause appears to be that the UI launcher falls back to:

```bash
npx @vectorize-io/hindsight-control-plane@<version>
```

When the control-plane package is not installed yet, `npx` prompts for confirmation before downloading it. But `start_ui()` launches the process in the background with stdout/stderr redirected to the profile UI log, so the prompt is hidden from the user and the wrapper just waits until its 30s health-check loop times out.

## Actual behavior
`ui start` times out, and the user only sees the real cause by opening the profile UI log.

Relevant log content from `~/.hindsight/profiles/hermes.ui.log`:

```text
Need to install the following packages:
@vectorize-io/hindsight-control-plane@0.5.3
Ok to proceed? (y)
```

## Expected behavior
On first launch, the control-plane package install should either:
1. run non-interactively, or
2. surface a clear actionable error in the foreground instead of silently timing out.

## Reproduction
1. Use a fresh machine/profile where `@vectorize-io/hindsight-control-plane` is not already installed.
2. Run `hindsight-embed -p hermes ui start`.
3. Wait for the UI startup wrapper to time out.
4. Inspect `~/.hindsight/profiles/hermes.ui.log`.

## Environment
- OS: macOS
- `hindsight-embed`: observed with 0.5.3
- Daemon itself starts successfully; the issue is specific to the UI first-launch path.

## Suggested fix
Change the published-package path in `hindsight_embed/daemon_embed_manager.py` from:

```python
["npx", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
```

to:

```python
["npx", "-y", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
```

That should remove the hidden confirmation prompt on first launch.

A secondary improvement could be to either extend the first-launch timeout a bit or improve the timeout message when the log indicates `npx` was waiting for installation input.

If this approach makes sense, I can open a PR as well.

bwjoke:
## Summary
- add `-y` to the published `npx` control-plane invocation used by `hindsight-embed ui start`
- document why the first-launch prompt is a problem when UI startup is backgrounded and log-redirected
- add a regression test covering the published-package code path

## Why
On first launch, `npx @vectorize-io/hindsight-control-plane@<version>` can prompt before installing the package. Because `start_ui()` redirects stdout/stderr to the profile UI log and waits for health checks in the foreground, that prompt is invisible to the user and the startup wrapper times out.

Using `npx -y ...` keeps the first-run install non-interactive, which matches the current background-launch design.

## Testing
- `uv run pytest tests/ -q`
- `./scripts/hooks/lint.sh`

Fixes #1196

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -499,7 +499,10 @@ class DaemonEmbedManager(EmbedManager):
         from . import __version__
 
         cp_version = os.getenv("HINDSIGHT_EMBED_CP_VERSION", __version__)
-        return ["npx", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
+        # `npx` prompts before installing missing packages on first run unless `-y` is set.
+        # The UI starts in the background with stdout/stderr redirected to a log file, so an
+        # interactive prompt would be invisible to users and the health-check loop would time out.
+        return ["npx", "-y", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
 
     def get_ui_url(self, profile: str, ui_port: int | None = None, hostname: str | None = None) -> str:
         """Get the URL for the UI serving this profile."""
--- a/hindsight-embed/tests/test_embed_manager.py
+++ b/hindsight-embed/tests/test_embed_manager.py
@@ -1,6 +1,6 @@
 """Tests for EmbedManager interface."""
 
-from unittest.mock import MagicMock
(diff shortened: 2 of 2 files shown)


nicoloboschi (approved):
+1

PR #1204, opened 2026-04-22
fix: warn when embedded pg0 database is used instead of external PostgreSQL

### LINKED ISSUE #1202: Silent fallback to embedded pg0 when HINDSIGHT_API_DATABASE_URL is misnamed — causes data loss on container restart
## Summary

When `HINDSIGHT_API_DATABASE_URL` is not set (or is set under a slightly-wrong variable name), Hindsight silently falls back to the embedded `pg0` Postgres inside the container without any log warning. The service passes all health checks, runs migrations successfully, serves traffic — and silently loses every ingested memory on container restart because the embedded DB is ephemeral.

This is an easy mistake for cloud operators (Cloud Run, ECS, Fly, etc.) to make, and the failure mode is maximally expensive: deployment looks green for days/weeks, and the bug is only discovered when someone queries retrieval and finds an empty bank.

## Reproduction

1. Deploy `ghcr.io/vectorize-io/hindsight:latest` to any container platform.
2. Mount a Postgres DSN to a nearby-but-incorrect env var name — e.g., `HINDSIGHT_DB_URL`, `HINDSIGHT_DATABASE_URL`, `DATABASE_URL` — instead of the documented `HINDSIGHT_API_DATABASE_URL`.
3. Start the container.
4. Observe the logs:
   ```
   INFO - hindsight_api.engine.memory_engine - Connecting to PostgreSQL at postgresql://***:***@localhost:5432/hindsight
   INFO - alembic.runtime.migration - Running upgrade ... (schema migrations run successfully)
   INFO - hindsight_api.migrations - Database migrations completed successfully for schema 'public'
   INFO - hindsight_api.engine.memory_engine - Schema migrations completed
   ```
   Everything looks perfect. `/health` returns `{"status":"healthy","database":"connected"}`.
5. POST some memories via `/v1/default/banks/:bank_id/memories`. They ingest without error.
6. Restart the container.
7. GET `/v1/default/banks` — returns `{"banks":[]}`. All data is gone. Migrations run again from scratch.

The actual external Postgres (intended to be used) remains completely empty the whole time.

## Expected behavior

One of the following, in descending order of preference:

1. **Refuse to start if no external DB URL is configured and embedded mode was not explicitly opted into.** Embedded pg0 is fine for local dev but catastrophic as a silent cloud default. A startup flag like `HINDSIGHT_API_ALLOW_EMBEDDED_DB=true` (defaulting to false) would force operators to acknowledge what they're doing.
2. **Log a prominent warning** at startup when falling back to embedded pg0, on every container start, at `WARN` or `ERROR` level. Something like: `⚠️ HINDSIGHT_API_DATABASE_URL is not set — using ephemeral embedded pg0. All data will be lost on container restart. This is not suitable for production.`
3. **Expand the health endpoint** to report which DB backend is in use. `/health` currently reports `"database":"connected"` whether the DB is embedded or external. Adding `"database_mode": "embedded"` vs `"database_mode": "external"` would let monitoring catch this.

## Additional context

Discovered on 2026-04-22 during a multi-day observation window for an Arabic-language customer-support memory pilot. We had mounted our Cloud SQL Postgres to `HINDSIGHT_DB_URL` (a no-op name — close enough to sound right that neither the initial deployer nor the reviewer caught it). Hindsight looked healthy for days. Memory ingest worked, retrieval worked (against the ephemeral embedded DB), migrations ran cleanly. We only discovered the issue when we built an admin monitoring page that surfaced bank counts — and the counts were zero despite confirmed ingest.

Diagnosis was non-obvious because `/health` lied, `pg_tables` / `\dt` on the external Postgres (via Cloud SQL Proxy) returned empty, and the Hindsight logs cheerfully reported successful migrations each startup — all consistent with a correctly-configured deployment. Only after checking `information_schema` for table counts on the external DB did the picture become clear: Hindsight was writing to `localhost:5432` inside its own container, not to the external DB we thought we'd configured.

Fix on our side was a one-line deploy-workflow change renaming the secret mount from `HINDSIGHT_DB_URL` to `HINDSIGHT_API_DATABASE_URL`. Data lost between the initial deploy and the fix was unrecoverable.

## Environment

- Image: `ghcr.io/vectorize-io/hindsight:latest` (digest `sha256:1086902f21ab9ca0021b1660f0f1c2455f61c0651d9171aba8c6edfde1980e66`)
- Platform: Google Cloud Run
- External DB: Cloud SQL Postgres 15.17 with `pgvector 0.8.1`
- Connection mode: Unix socket via `/cloudsql/PROJECT:REGION:INSTANCE`

Happy to provide logs, repro scripts, or open a PR if helpful.

octo-patch:
Fixes #1202

## Problem

When `HINDSIGHT_API_DATABASE_URL` is not set (or is set under a misnamed variable such as `DATABASE_URL`, `HINDSIGHT_DB_URL`, etc.), Hindsight silently falls back to the embedded `pg0` PostgreSQL. Because `pg0` is ephemeral, **all ingested data is lost on every container restart**. The service passes health checks, runs migrations, and serves traffic normally — operators only discover the problem days or weeks later when memories are missing.

## Solution

Add two warning surfaces so the problem is immediately visible:

1. **Startup banner** (`banner.py`): When `database_url` is `"pg0"` or starts with `"pg0://"`, a yellow ⚠ line is appended to the Database row in `print_startup_info()` and a second line directs operators to set `HINDSIGHT_API_DATABASE_URL`.

2. **Logger warning** (`config.py`): `log_config()` now calls `_warn_if_embedded_db()`, which emits a `logger.warning()` with the same message. This ensures the warning is captured by container log scrapers even when the banner is suppressed (daemon mode).

## Testing

Manual verification: start the server without setting `HINDSIGHT_API_DATABASE_URL` and observe the new warning in startup output and logs. Existing test suite unchanged.

### DIFF
--- a/hindsight-api-slim/hindsight_api/banner.py
+++ b/hindsight-api-slim/hindsight_api/banner.py
@@ -94,7 +94,15 @@ def print_startup_info(
     if version:
         print(f"  {dim('Version:')} {color(f'v{version}', 0.1)}")
     print(f"  {dim('URL:')} {color(f'http://{host}:{port}', 0.2)}")
-    print(f"  {dim('Database:')} {color(mask_network_location(database_url), 0.4)}")
+    is_embedded = database_url == "pg0" or database_url.startswith("pg0://")
+    db_label = mask_network_location(database_url)
+    if is_embedded:
+        db_label += "  \033[33m⚠  ephemeral — data lost on restart\033[0m"
+    print(f"  {dim('Database:')} {color(db_label, 0.4)}")
+    if is_embedded:
+        print(
+            "\033[33m  ⚠  Set HINDSIGHT_API_DATABASE_URL to a PostgreSQL DSN for persistent storage.\033[0m"
+        )
     print(f"  {dim('LLM:')} {color(f'{llm_provider} / {llm_model}', 0.6)}")
     print(f"  {dim('Embeddings:')} {color(embeddings_provider, 0.8)}")
     print(f"  {dim('Reranker:')} {color(reranker_provider, 1.0)}")
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -1797,8 +1797,17 @@ class HindsightConfig:
         # Silence noisy third-party loggers
         logging.getLogger("google_genai.models").setLevel(logging.WARNING)
 
+    def _warn_if_embedded_db(self) -> None:
+        """Emit a warning when the embedded pg0 database is in use."""
+        if self.database_url == DEFAULT_DATABASE_URL or self.database_url.startswith("pg0://"):
+            logger.warning(
+                "⚠️  Using embedded pg0 PostgreSQL — data will be LOST on container restart. "
+                "Set HINDSIGHT_API_DATABASE_URL to an external PostgreSQL DSN for persistent storage."
+            )
+
     def log_config(self) -> None:
         """Log the current configuration (without sensitive values)."""
+        self._warn_if_embedded_db()
         logger.info(f"Database: {self.database_url} (schema: {self.database_schema})")
         if self.migration_database_url:
             logger.info(f"Migration database: {self.migration_database_url}")


nicoloboschi:
this pr is wrong, closing

PR #1205, opened 2026-04-22
fix: flush un-retained turns in SessionEnd hook before daemon stop

### LINKED ISSUE #1145: hindsight-memory: SessionEnd hook does not flush un-retained turns before daemon stop
## Summary

`hindsight-memory` plugin 0.3.0 throttles the Stop-hook retain via `retainEveryNTurns` (default 10), meaning uploads only fire on turns 10, 20, 30, etc. The `SessionEnd` hook does **not** call retain — it only stops the local daemon. Result: **any turns between the last 10-turn boundary and session end are silently dropped** from auto-retention.

## Reproduction

1. Start a Claude Code session with default `retainEveryNTurns: 10`.
2. Have a conversation of, say, 27 turns, then close the CLI (any exit path).
3. Check `list_operations` and `list_documents` on the bank.

Observed: the most recent retain operation corresponds to turn 20's timestamp. Turns 21-27 are not captured anywhere.

Expected: all 27 turns captured, either via a final retain triggered by SessionEnd or by some equivalent flush.

## Evidence from my own bank

- A real Claude Code session had retentions at `2026-04-17T22:01` and `2026-04-17T23:09` (turns 10 and 20 with message count 509).
- No further retain operations fired before session end — confirmed via `list_operations`.
- Roughly 1.5 hours of subsequent work (a completed PR merge, a second PR, multiple code changes, architectural decisions) never reached the bank.

## Root cause

In `scripts/retain.py`:

\`\`\`python
if retain_every_n > 1:
    turn_count = increment_turn_count(session_id)
    if turn_count % retain_every_n != 0:
        debug_log(config, f"Turn {turn_count}/{retain_every_n}, skipping retain...")
        return
\`\`\`

And in `scripts/session_end.py`:

\`\`\`python
def main():
    config = load_config()
    # ... reads stdin ...
    debug_log(config, f"SessionEnd hook, reason: ...")
    stop_daemon(config, debug_fn=_dbg)   # ← whole job
\`\`\`

No retain call in SessionEnd.

## Proposed fix

In `session_end.py`, invoke a final retain before stopping the daemon. Skip the turn-count throttle for this one case (the whole point is to flush whatever hasn't been uploaded yet). Candidate shape:

\`\`\`python
def main():
    config = load_config()
    hook_input = _read_stdin()
    debug_log(config, f"SessionEnd hook, reason: {hook_input.get('reason', 'unknown')}")

    # Flush before stopping daemon
    if config.get(\"autoRetain\"):
        try:
            _final_flush(config, hook_input)
        except Exception as e:
            print(f\"[Hindsight] SessionEnd flush failed: {e}\", file=sys.stderr)

    stop_daemon(config, debug_fn=lambda *a: debug_log(config, *a))
\`\`\`

Where `_final_flush` reuses the transcript-read + format + POST logic from `retain.py`, bypassing the `retainEveryNTurns` check.

## Workaround for users

Set `\"retainEveryNTurns\": 1` in `~/.hindsight/claude-code.json`. Every Stop fires a full-session upload. `document_id=session_id` upsert semantics keep storage cost ~constant. Still loses the single final turn after the last Stop but dramatically reduces data loss.

## Severity

Medium — no crashes, no user-visible error, but users lose load-bearing content silently. Typical sessions hit this (retainEveryNTurns default 10, few sessions are multiples of 10 turns long).

---
_Discovered while diagnosing a retention gap on 2026-04-18._

octo-patch:
Fixes #1145

## Problem

When `retainEveryNTurns > 1` (e.g. the default of 10), the Stop hook skips retain on non-multiple turns. Turns between the last multiple and session end are silently dropped — a typical 27-turn session retains only up to turn 20, losing turns 21-27 completely.

## Solution

`session_end.py` now calls `_flush_remaining_turns()` before stopping the daemon:

- Reads `turn_count` from state (without incrementing) and checks `turn_count % retain_every_n != 0`
- **Skips** when `retainEveryNTurns == 1` — Stop hook already covers every turn
- **Skips** when `turn_count % retainEveryNTurns == 0` — the last Stop hook already retained
- Uses `session_id` as `document_id` so the flush upserts the same document as prior full-session retains (last write wins, no duplicate storage)
- Adds `flush_reason: session_end` to metadata for observability
- Guarded by `config.get("autoRetain")` — no-op when auto-retain is disabled

## Testing

Added `TestSessionEndHook` (4 tests) to `test_hooks.py`:
- `test_flushes_remaining_turns_on_session_end` — verifies flush fires when turn 1/3 was skipped
- `test_skips_flush_when_retain_every_n_is_one` — no extra retain when already retained every turn
- `test_skips_flush_when_last_stop_already_retained` — no double-retain after turn 3/3
- `test_skips_flush_when_auto_retain_disabled` — respects `autoRetain: false`

All 137 existing tests continue to pass.

### DIFF
--- a/hindsight-integrations/claude-code/scripts/session_end.py
+++ b/hindsight-integrations/claude-code/scripts/session_end.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
-"""SessionEnd hook: daemon cleanup.
+"""SessionEnd hook: final retain flush + daemon cleanup.
 
-Fires once when a Claude Code session terminates. If the plugin
-auto-started a hindsight-embed daemon, this is where we stop it.
+Fires once when a Claude Code session terminates. Two responsibilities:
+1. Flush any un-retained turns that were skipped by retainEveryNTurns throttle.
+2. Stop the auto-started hindsight-embed daemon (if any).
 
 Port of: Openclaw's service.stop() in index.js
 """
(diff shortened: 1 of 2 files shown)


nicoloboschi:
already fixed

PR #1212, opened 2026-04-22
[codex] fix(worker): cache missing schemas_with_pending_work helper

JulienJBO:
## Summary

- cache the absence of the optional `schemas_with_pending_work()` helper after the first `UndefinedFunctionError`
- keep the existing per-schema fallback path unchanged
- add a regression test that verifies the helper is only probed once when it is not installed

## Root Cause

`WorkerPoller._scan_active_schemas()` tried `SELECT * FROM schemas_with_pending_work()` on every poll tick until success. In local and self-hosted deployments where that optional helper function is not installed, the worker correctly fell back to per-schema `EXISTS` checks, but PostgreSQL still logged an `UndefinedFunction` error every 500ms.

That made the fallback effectively noisy forever even though the absence of the helper is a stable condition for the lifetime of the process.

## Fix

When the fast-path query fails specifically with `asyncpg.exceptions.UndefinedFunctionError`, mark the helper as unavailable for the current worker process and skip future probes. Other exception types still fall back without disabling future attempts, preserving the existing behavior for transient failures.

## Testing

- `uv run pytest tests/test_worker.py::TestClaimBatchRotation::test_scan_caches_missing_server_side_pending_work_function -v`
- `uv run ruff check hindsight_api/worker/poller.py tests/test_worker.py`
- `./scripts/hooks/lint.sh`

## Impact

This removes persistent PostgreSQL log spam in deployments that do not install the optional helper function, without changing task claiming semantics.

### DIFF
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -15,6 +15,8 @@ from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any
 
+import asyncpg
+
 from .exceptions import DeferOperation, RetryTaskAt
 from .stage import StageHolder, bind_holder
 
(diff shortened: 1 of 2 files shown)


JulienJBO:
Closing this draft while local validation is still in progress. I will reopen or submit a fresh PR once tests and Sonar checks are complete.

nicoloboschi:
hey @JulienJBO i think this PR is valuable, can you rebase and remove the draft state?

PR #1245, opened 2026-04-24
fix(stats): timeseries buckets should return tz-aware ISO

aliu-ronin:
## Summary

The `/v1/default/banks/{bank_id}/stats/memories-timeseries` endpoint returns bucket timestamps as naive ISO strings (`2026-04-18T00:00:00`) instead of tz-aware (`2026-04-18T00:00:00+00:00`). Per ECMA-262, browsers parse a naive ISO date-time as **local time**, so the control plane's *Memories ingested* chart (via `formatBucketLabel` in `bank-stats-view.tsx`) was rendering buckets shifted by the browser's timezone offset — e.g. a user in UTC+8 would see UTC 05:00 labeled as 05:00 instead of 13:00.

Every other endpoint (e.g. `/memories/list`) already serializes timestamps with a `+00:00` offset; this one just happened to strip it.

## Changes

**Backend** (`hindsight-api-slim/hindsight_api/engine/memory_engine.py`)
- Anchor `now_utc` with `datetime.now(timezone.utc)` (tz-aware) instead of the deprecated `datetime.utcnow()` (naive).
- Normalize `asyncpg` row timestamps to UTC-aware instead of stripping `tzinfo`.
- Result: bucket `time` strings now end in `+00:00`.

**Test** (`hindsight-api-slim/tests/test_bank_stats.py`)
- Extends `test_memories_timeseries_periods` to assert every bucket `time` carries an explicit UTC offset.

**Frontend** (`hindsight-control-plane/src/components/bank-stats-view.tsx`)
- Defensive parse: `parseBucketIso` appends `Z` when no offset is present, so a control plane paired with an older API still renders correctly. Paired with the backend fix it's a no-op.

## Repro (before)

```bash
curl -s http://localhost:8000/v1/default/banks/main/stats/memories-timeseries?period=1h \
  | jq '.buckets[0].time'
# "2026-04-24T05:06:00"  ← no offset
```

## After

```bash
# "2026-04-24T05:06:00+00:00"
```

## Test plan

- [x] `pytest tests/test_bank_stats.py::test_memories_timeseries_periods` passes for every period (minute/hour/day truncation)
- [x] Manually verified the *Memories ingested* chart on a UTC+8 browser renders the correct local hour after the fix
- [x] Before the backend fix, the added assertion fails; confirms regression coverage

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -6611,7 +6611,10 @@ class MemoryEngine(MemoryEngineInterface):
             )
 
         # Build the canonical bucket list anchored on the most recent UTC boundary.
-        now_utc = datetime.utcnow()
+        # Use tz-aware UTC throughout so serialized ISO strings include a `+00:00`
+        # offset; a naive ISO (`2026-04-18T00:00:00`) would be parsed by browsers
+        # as local time per ECMA-262, producing an off-by-timezone display.
+        now_utc = datetime.now(timezone.utc)
         if cfg.trunc == "minute":
             end = now_utc.replace(second=0, microsecond=0)
         elif cfg.trunc == "hour":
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
LGTM

PR #1246, opened 2026-04-24
feat(stats): add time_field toggle to memories-timeseries chart

aliu-ronin:
## Summary

The *Memories ingested* chart on the bank dashboard always buckets by `memory_units.created_at` (ingest time). For a bank built up in real time, ingest time ≈ event time and that's the right default. But when a corpus is **backfilled in a single session** — for example migrating from another memory system — every record's `created_at` collapses to the import moment, so the chart shows "all knowledge is new" and the underlying timeline disappears.

Adds a `time_field` query parameter (and a three-way UI toggle) that lets the viewer pick which timestamp column drives the bucket assignment:

| Value | Meaning |
|---|---|
| `created_at` (default) | Ingest time — matches old behavior exactly |
| `mentioned_at` | Event time: when the fact was mentioned |
| `occurred_start` | Event time: when the underlying event started |

For the event-time columns we `COALESCE(<col>, created_at)` per row so records lacking an event timestamp still show up in the chart instead of silently disappearing. The field is whitelisted (never interpolated from untrusted input), unknown values fall back to `created_at`, and the chosen column is echoed back in the response for UI affordance.

## Changes

**Backend** (`hindsight-api-slim`)
- `memory_engine.get_memories_timeseries` gains a `time_field` keyword arg, validates it against an allowlist, and composes `bucket_expr` with `COALESCE` for the event-time columns.
- `MemoriesTimeseriesResponse` adds a `time_field` field.
- API endpoint exposes `time_field` as a query param with a clear description.

**Control Plane** (`hindsight-control-plane`)
- `getMemoriesTimeseries` SDK helper accepts a `timeField` argument, defaulting to `created_at`.
- `/api/stats/[agentId]/memories-timeseries` proxy forwards the param.
- `bank-stats-view.tsx` renders an **Ingested / Mentioned / Occurred** segmented control below the period selector; the card title updates so the chart reads unambiguously (e.g. *"Memories by mentioned time"*).

## Screenshots

On a bank migrated from another memory system, toggling to **Mentioned** spreads the chart across the actual event timeline instead of a single ingest-day spike:

- `created_at` → two-day spike (04-23, 04-24)  
- `mentioned_at` → spread across 03-26 through 03-29 (real history)

## Test plan

- [x] `curl .../memories-timeseries?period=30d&time_field=mentioned_at` returns non-zero buckets spanning the corpus's actual timeline
- [x] `time_field=nonsense` falls back to `created_at` (parity with the invalid-period behavior)
- [x] Response includes `time_field` so the UI can display the active dimension
- [x] Control plane chart updates immediately when the toggle changes, with no stale frames

## Notes

- Defaults stay `created_at` everywhere; existing integrations see no behavioral change.
- Builds on #1245 (tz-aware bucket ISO). Without that fix the chart is off-by-timezone in non-UTC browsers regardless of which `time_field` is selected.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1519,6 +1519,14 @@ class MemoriesTimeseriesResponse(BaseModel):
     bank_id: str
     period: str = Field(description="One of: 1h, 12h, 1d, 7d, 30d, 90d.")
     trunc: str = Field(description="Bucket granularity: minute, hour, day.")
+    time_field: str = Field(
+        default="created_at",
+        description=(
+            "Timestamp column used to assign each row to a bucket. "
+            "`created_at` shows ingest time; `mentioned_at` / `occurred_start` "
+            "show event time (falls back to `created_at` per row when null)."
+        ),
+    )
     buckets: list[MemoryTimeseriesBucket] = Field(
         default_factory=list,
         description="Per-bucket counts, always returned fully padded for the requested period.",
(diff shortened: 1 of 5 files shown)


nicoloboschi (approved):
LGTM

nicoloboschi:
@aliu-ronin can you resolve conflicts?

aliu-ronin:
@nicoloboschi rebased on `main`, conflicts resolved. The collision was around the `pool` → `backend` rename in `memory_engine.py::get_memories_timeseries` — kept the new `_get_backend()` naming and the `_ALLOWED_TIME_FIELDS` whitelist together. PTAL 🙏

PR #1247, opened 2026-04-24
fix(entity-resolver): stamp cooccurrences with event_date, not now()

aliu-ronin:
## Summary

`entity_cooccurrences.last_cooccurred` is always set to `datetime.now(UTC)` at flush time, regardless of the source memory unit's event date. For real-time retains this is fine (ingest time ≈ event time), but any bank **backfilled in a single session** — e.g. migrating from another memory system — collapses every co-occurrence onto the import moment. The control plane's entity-graph recency heat then shows a one-or-two-day range regardless of how far the underlying knowledge actually spans.

Example: on a bank migrated on 2026-04-24, every edge's `lastCooccurred` was `2026-04-23` or `2026-04-24`, even though the source memories' `mentioned_at` / `occurred_start` correctly reflect the original 2026-03 timeline.

Interestingly the tuples flowing into `_link_units_to_entities_batch_impl` already carried the per-unit `fact_date` alongside `(unit_id, entity_id)` — it was just being discarded at the call site (`_fact_date` underscore). This change wires it through.

### Writer path

- `_CooccurrencePair` grows an `event_date: datetime | None` field.
- `link_units_to_entities_batch` accepts both the legacy `(unit_id, entity_id)` tuples **and** the new `(unit_id, entity_id, event_date)` form, so external callers aren't forced to migrate in lockstep.
- `_link_units_to_entities_batch_impl` builds a per-unit event-date map and attaches the unit's date to every co-occurrence pair emitted from that unit. A `_SENTINEL_MISSING` guard prevents a duplicate legacy-caller row (`event_date=None`) from clobbering a real datetime from another caller on the same unit.
- `flush_pending_stats` aggregates per-pair event dates and `INSERT`s the observed maximum. It falls back to `datetime.now(UTC)` only when no event date was carried, so real-time retains preserve their pre-fix semantics exactly.

### In-repo callers

- `retain/orchestrator.py`:290 — pass `fact_date` (was already in the tuple as `_fact_date`).
- `retain/link_utils.py`:429 — same.

### Backfill migration

`b5d4e3f2a1c9_backfill_entity_cooccurrences_event_time.py` recomputes `last_cooccurred` from `MAX(COALESCE(mentioned_at, occurred_start, created_at))` over `unit_entities × memory_units`, so operators don't have to run a manual SQL backfill to see the fix on their existing data. The downgrade path is a no-op — the pre-fix value was `now()` at the time of write and isn't recoverable, and rolling the code back is enough to revert behavior for subsequent retains.

## Test plan

- [x] New regression in `test_entity_resolver.py`: inserts a memory unit with a historical `mentioned_at`, calls `link_units_to_entities_batch` + `flush_pending_stats`, asserts `entity_cooccurrences.last_cooccurred == historical`.
- [x] `python -m py_compile` passes on the three touched engine files, the migration, and the test file.
- [x] Backward compat: passing the old two-tuple shape still works (falls back to `now()`, matching pre-fix behavior).
- [x] Manually verified on a migrated bank: running the migration moves the entity-graph recency heat range from "one-day spike" back to the true multi-week event-time span.

## Repro (before)

```bash
curl -s http://localhost:8000/v1/default/banks/main/entities/graph?limit=3 \
  | jq '.edges[].data.lastCooccurred'
# "2026-04-24T06:47:08+00:00"
# "2026-04-24T06:45:12+00:00"
# "2026-04-24T06:39:02+00:00"   ← all today, despite 03-xx memories
```

## After (with migration applied)

```bash
# "2026-03-30T..." / "2026-03-29T..." / ...
```

### DIFF
--- a/hindsight-api-slim/hindsight_api/alembic/versions/b5d4e3f2a1c9_backfill_entity_cooccurrences_event_time.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/b5d4e3f2a1c9_backfill_entity_cooccurrences_event_time.py
@@ -1,0 +1,91 @@
+"""Backfill entity_cooccurrences.last_cooccurred from memory_units event time
+
+Revision ID: b5d4e3f2a1c9
+Revises: o1a2b3c4d5e6
+Create Date: 2026-04-24
+
+The writer path in `entity_resolver.link_units_to_entities_batch` historically
+stamped `entity_cooccurrences.last_cooccurred` with `datetime.now(UTC)` at
+flush time, ignoring the source memory unit's event date. For normal online
+retains that's fine (now ≈ event time), but for any corpus that was
+backfilled in a single session — migrating from another memory system, for
+example — every co-occurrence collapsed to the import moment, which hid the
+underlying knowledge timeline from the dashboard's entity graph recency heat
+and from any downstream consumer of the column.
+
+The writer is fixed in the same change set to propagate the unit's event_date;
+this migration repairs historical rows by reading the true event time off
+`unit_entities × memory_units` (falling back to `created_at` when
+`mentioned_at` / `occurred_start` are NULL, so rows never regress).
+
+Oracle slot is intentionally absent: the Oracle baseline (`o1a2b3c4d5e6`)
+landed days before this fix, so any Oracle deployment runs the corrected
+writer against an effectively empty `entity_cooccurrences` — there is no
+historical residue on Oracle to repair. PG-only matches the asymmetry of
+the data, not negligence.
+"""
+
+from collections.abc import Sequence
+
+from alembic import context, op
+
+from hindsight_api.alembic._dialect import run_for_dialect
+
+revision: str = "b5d4e3f2a1c9"
+down_revision: str | Sequence[str] | None = "o1a2b3c4d5e6"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _get_schema_prefix() -> str:
+    """Get schema prefix for table names (required for multi-tenant support)."""
+    schema = context.config.get_main_option("target_schema")
+    return f'"{schema}".' if schema else ""
+
+
+def _pg_upgrade() -> None:
+    schema = _get_schema_prefix()
+    # Recompute last_cooccurred from the true event time per entity pair.
+    # COALESCE picks the first non-null of mentioned_at / occurred_start /
+    # created_at so banks without event-time metadata still see a sane value
+    # (equivalent to the pre-fix behaviour) instead of NULL.
+    #
+    # The self-join on `unit_entities` is O(k²) per memory_unit in the number
+    # of distinct entities mentioned (k). For typical units k is small (single
+    # digits), but a bank with units containing hundreds of entities and tens
+    # of millions of co-occurrence rows may want to run this off-hours — the
+    # whole UPDATE is one statement, so it locks every targeted ec row for
+    # the duration. The migration is one-time; subsequent online writes
+    # already carry event time via the writer fix.
+    op.execute(
+        f"""
+        UPDATE {schema}entity_cooccurrences ec
+        SET last_cooccurred = sub.event_time
+        FROM (
+            SELECT
+                LEAST(ue1.entity_id, ue2.entity_id) AS e1,
+                GREATEST(ue1.entity_id, ue2.entity_id) AS e2,
+                MAX(COALESCE(mu.mentioned_at, mu.occurred_start, mu.created_at)) AS event_time
+            FROM {schema}memory_units mu
+            JOIN {schema}unit_entities ue1 ON ue1.unit_id = mu.id
+            JOIN {schema}unit_entities ue2 ON ue2.unit_id = mu.id AND ue1.entity_id <> ue2.entity_id
+            GROUP BY 1, 2
+        ) sub
+        WHERE ec.entity_id_1 = sub.e1 AND ec.entity_id_2 = sub.e2
+        """
+    )
+
+
+def _pg_downgrade() -> None:
+    # No-op: the previous column value was `now()` at the time of write and
+    # isn't recoverable. Rolling back the code is sufficient — new writes will
+    # revert to the old behaviour for subsequent retains.
+    pass
+
+
+def upgrade() -> None:
+    run_for_dialect(pg=_pg_upgrade)  # oracle slot intentionally absent — see header
+
+
+def downgrade() -> None:
+    run_for_dialect(pg=_pg_downgrade)
--- a/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
+++ b/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
@@ -12,7 +12,7 @@ from collections import defaultdict
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from difflib import SequenceMatcher
-from typing import Any
+from typing import Any, Final
 
 from .db_utils import acquire_with_retry
 from .memory_engine import fq_table
@@ -46,12 +46,38 @@ class _EntityStatAgg:
     max_date: datetime | None = None
 
 
+# Sentinel distinguishing "key not in dict" from "key present with value None".
+# Needed when merging event_date across duplicate unit rows: legacy two-tuple
+# callers surface `None`, which must not clobber a real datetime from another
+# caller for the same unit.
+_SENTINEL_MISSING: Final = object()
+
+
+def _later_date(a: datetime | None, b: datetime | None) -> datetime | None:
+    """Return whichever of ``a`` / ``b`` is later (None loses to any datetime).
+
+    Used to fold duplicate co-occurrence pairs across a retain batch: legacy
+    two-tuple callers surface ``None``, which must not clobber a real
+    datetime that arrived for the same pair from an aware caller.
+    """
+    if a is None:
+        return b
+    if b is None:
+        return a
+    return a if a > b else b
+
+
 @dataclass
 class _CooccurrencePair:
     """A (entity_id_1, entity_id_2) pair observed in a retain batch (for post-txn flush)."""
 
     entity_id_1: str
     entity_id_2: str
+    # When the two entities co-occurred in the source content. For real-time
+    # retains this is ~now; for backfilled corpora it's the historical event
+    # time, so the cooccurrence cache reflects the underlying knowledge
+    # timeline instead of collapsing to the import moment.
+    event_date: datetime | None = None
 
 
 # Load spaCy model (singleton)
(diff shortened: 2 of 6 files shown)


aliu-ronin:
### End-to-end verification on a running instance

Deployed the writer patch to a live Hindsight container (v1.x image, same codebase locations as this PR targets), then retained a test memory with a historical `timestamp`:

```bash
POST /v1/default/banks/main/memories
{
  "items": [{
    "content": "In June 2024, Fitzgerald Blackwood met Gwendolyn Stormrider at the Reykjavik summit to discuss the Chronos Protocol.",
    "timestamp": "2024-06-15T12:00:00Z"
  }]
}
```

Result in the DB:

| field | value |
|---|---|
| `memory_units.mentioned_at` | `2024-06-15 12:00+00` (matches `timestamp`) |
| `memory_units.occurred_start` | `2024-06-15 00:00+00` (extracted by the LLM as the event day) |
| `memory_units.created_at` | `2026-04-24 08:35+00` (ingest moment) |
| **`entity_cooccurrences.last_cooccurred`** | **`2024-06-15 00:00+00`** (event time, **not** ingest time) |

All 4 fresh entity pairs — `(Fitzgerald Blackwood, Gwendolyn Stormrider)`, `(Fitzgerald Blackwood, Chronos Protocol)`, `(Fitzgerald Blackwood, Reykjavik summit)`, `(Gwendolyn Stormrider, Chronos Protocol)`, etc. — carry `last_cooccurred = 2024-06-15`. Without this patch they would all be `2026-04-24` (the ingest moment).

hindsight-api-slim/hindsight_api/alembic/versions/b5d4e3f2a1c9_backfill_entity_cooccurrences_event_time.py
```diff
@@ -0,0 +1,67 @@
...
+`mentioned_at` / `occurred_start` are NULL, so rows never regress).
+"""
+
+from collections.abc import Sequence
+
+from alembic import context, op
+
+revision: str = "b5d4e3f2a1c9"
+down_revision: str | Sequence[str] | None = "z1u2v3w4x5y6"
+branch_labels: str | Sequence[str] | None = None
```
nicoloboschi, line 36: **Blocker: `down_revision` creates a divergent branch.**

`z1u2v3w4x5y6` already has two children (`a2v3w4x5y6z7` and `aa2b3c4d5e6f`). Adding this as a third child will break the migration graph — Alembic will refuse to run with multiple unmerged heads.

The current chain head on `main` is `i4j5k6l7m8n9`. This migration needs to be re-parented to descend from that (or whatever the head is at merge time).

hindsight-api-slim/hindsight_api/alembic/versions/b5d4e3f2a1c9_backfill_entity_cooccurrences_event_time.py
```diff
@@ -0,0 +1,67 @@
...
+                LEAST(ue1.entity_id, ue2.entity_id) AS e1,
+                GREATEST(ue1.entity_id, ue2.entity_id) AS e2,
+                MAX(COALESCE(mu.mentioned_at, mu.occurred_start, mu.created_at)) AS event_time
+            FROM {schema}memory_units mu
+            JOIN {schema}unit_entities ue1 ON ue1.unit_id = mu.id
+            JOIN {schema}unit_entities ue2 ON ue2.unit_id = mu.id AND ue1.entity_id <> ue2.entity_id
+            GROUP BY 1, 2
+        ) sub
+        WHERE ec.entity_id_1 = sub.e1 AND ec.entity_id_2 = sub.e2
+        """
```
nicoloboschi, line 75: This self-join (`ue1 × ue2` on the same `unit_id`) is O(n²) per unit in the number of entities. For banks with many entities per unit this could be slow. Worth adding a comment warning operators about expected runtime on large banks, or consider batching by `bank_id` with a `WHERE` clause so the UPDATE doesn't lock the whole table at once.

Not blocking since it's a one-time migration, but something to be aware of.

hindsight-api-slim/hindsight_api/engine/entity_resolver.py
```diff
@@ -46,12 +47,24 @@ class _EntityStatAgg:
     max_date: datetime | None = None
 
 
+# Sentinel distinguishing "key not in dict" from "key present with value None".
+# Needed when merging event_date across duplicate unit rows: legacy two-tuple
+# callers surface `None`, which must not clobber a real datetime from another
+# caller for the same unit.
+_SENTINEL_MISSING: typing.Any = object()
+
```
nicoloboschi, line 68: Nit: `typing.Any` defeats static analysis — any comparison with the sentinel will type-check as valid even if surrounding code regresses. Consider `_SENTINEL_MISSING: typing.Final = object()` instead, which preserves the intent and gives the type checker something to work with.

Also, `import typing` was added solely for this — the existing codebase uses `from typing import ...`. A `from typing import Final` would be more consistent.

hindsight-api-slim/hindsight_api/engine/entity_resolver.py
```diff
@@ -141,11 +154,18 @@ async def flush_pending_stats(self) -> None:
...
                 for c in cooccurrences:
                     pair = (c.entity_id_1, c.entity_id_2)
-                    coo_agg[pair] = coo_agg.get(pair, 0) + 1
+                    prev_count, prev_date = coo_agg.get(pair, (0, None))
+                    latest = prev_date
+                    if c.event_date is not None and (prev_date is None or c.event_date > prev_date):
+                        latest = c.event_date
+                    coo_agg[pair] = (prev_count + 1, latest)
 
                 now = datetime.now(UTC)
```
nicoloboschi, line 182: Nit: the "keep latest non-None event_date" logic appears both here (folding `coo_agg` in `flush_pending_stats`) and in `_link_units_to_entities_batch_impl` (building `cooccurrence_pairs`). Both do the same `if event_date is not None and (prev is None or event_date > prev)` pattern. A small helper like `_later_date(a, b) -> datetime | None` would reduce the surface area for the two copies diverging.

nicoloboschi (commented):
Good fix — well-motivated, well-documented PR. The core change is sound and the risk is low since `last_cooccurred` is only consumed by the entity graph visualization endpoint (tiebreaker in ORDER BY + UI recency heat), not by the recall/entity-resolution path at all.

A few items below, one blocker on the migration chain.

hindsight-api-slim/hindsight_api/engine/entity_resolver.py
```diff
@@ -825,28 +866,44 @@ async def _link_units_to_entities_batch_impl(self, conn, unit_entity_pairs: list
...
-            # For each pair of entities in this unit, create co-occurrence
+            entity_list = list(entity_ids)
+            event_date = unit_event_date.get(unit_id)
             for i, entity_id_1 in enumerate(entity_list):
                 for entity_id_2 in entity_list[i + 1 :]:
-                    # Skip if same entity (shouldn't happen with set, but be safe)
                     if entity_id_1 == entity_id_2:
                         continue
-                    # Ensure consistent ordering (entity_id_1 < entity_id_2)
+                    # Canonical ordering (entity_id_1 < entity_id_2) matches the
```
nicoloboschi, line 980: The first two branches are identical:
```python
if prev is _SENTINEL_MISSING:
    cooccurrence_pairs[key] = event_date
elif prev is None:
    cooccurrence_pairs[key] = event_date  # same
```
Could simplify to:
```python
if prev is _SENTINEL_MISSING or prev is None:
    cooccurrence_pairs[key] = event_date
elif event_date is not None and event_date > prev:
    cooccurrence_pairs[key] = event_date
```
Minor — the current form is arguably more explicit about intent, so up to you.

hindsight-api-slim/tests/test_entity_resolver.py
```diff
@@ -112,3 +112,75 @@ async def test_resolve_entities_batch_handles_unicode_lower_conflicts(pg0_db_url
...
+              AND entity_id_2 = GREATEST($1::uuid, $2::uuid)
+            """,
+            e1,
+            e2,
+        )
+        assert row is not None, "co-occurrence row should exist"
+        assert row["last_cooccurred"] == historical, (
+            f"expected last_cooccurred == {historical}, got {row['last_cooccurred']}"
+        )
+    finally:
```
nicoloboschi, line 399: The `finally` block cleans up `memory_units` and `entities` but doesn't explicitly delete from `unit_entities` or `entity_cooccurrences`. If the FK on `entity_cooccurrences` → `entities` isn't `ON DELETE CASCADE`, the `DELETE FROM entities` will fail and the cooccurrence row leaks. Even if cascades handle it, explicitly cleaning up `entity_cooccurrences` and `unit_entities` first (in the right order) would be safer and make the teardown self-documenting.

nicoloboschi:
@aliu-ronin please resolve review comments, thanks!

aliu-ronin:
@nicoloboschi rebased on `main` and addressed all review points in the amended commit (`4ada303`). Summary:

- **Migration parent (blocker):** re-parented to `o1a2b3c4d5e6` (current chain head, single-head verified). Converted to the `run_for_dialect` dispatcher; PG slot only — Oracle baseline (`o1a2b3c4d5e6`) landed days before this fix, so any Oracle deployment runs the corrected writer against an effectively empty `entity_cooccurrences`. Documented the asymmetry in the migration header.
- **Self-join O(k²):** added a comment in the migration explaining the per-unit cost, that it locks all targeted `ec` rows for the duration, and that this is acceptable for a one-time backfill.
- **Sentinel typing:** `_SENTINEL_MISSING: typing.Any = object()` → `_SENTINEL_MISSING: Final = object()`. Dropped the `import typing` (only added for the sentinel) and switched to `from typing import Any, Final` to match codebase style.
- **Duplicated "keep latest event_date" logic:** extracted `_later_date(a, b) -> datetime | None` helper. Now used in both `flush_pending_stats` (folding `coo_agg`) and `_link_units_to_entities_batch_impl` (building `cooccurrence_pairs`).
- **Identical first two branches:** collapsed via the helper — `if prev is _SENTINEL_MISSING: ... else: cooccurrence_pairs[key] = _later_date(prev, event_date)`.
- **Test cleanup ordering:** `finally` now explicitly deletes `unit_entities` and `entity_cooccurrences` before `entities` and `memory_units`, in dependency order, instead of relying on FK cascades. Also migrated the test to the new `create_database_backend` API since `pool` → `backend` was renamed on `main`.

Migration shape test passes (61/61). All entity_resolver tests pass (11/11). Type check + ruff clean.

aliu-ronin:
Heads up: the `verify-generated-files` failure on this PR is **not** caused by these changes — it traces back to #1246 landing without a regen pass. Opened #1426 to fix the drift; once that lands I'll rebase this PR and CI should go green.

(Confirmed locally that running the regen scripts against current `main` produces exactly the 8 files CI flagged here, all `time_field`-related, none touched by this PR.)

aliu-ronin:
Rebased on `main` (post-#1426) and pushed `b6bb4d9`. CI green now — `verify-generated-files` ✅, `check-openapi-compatibility` ✅, all `build-api-python-versions` ✅. `mergeable_state: clean`.

Two commits on the branch:
1. `eb3d573` — the entity-cooccurrence fix (with all your review points addressed)
2. `b6bb4d9` — incidental 1-line `configuration.md` regen for `HINDSIGHT_API_LLM_DEFAULT_HEADERS` from #1389 (same pattern as the time_field drift, will get squashed away on merge)

Ready when you are 🙏

PR #1253, opened 2026-04-24
fix(llm): handle DeepSeek tool-call compatibility

M1p0:
## Summary

Fix DeepSeek OpenAI-compatible tool-call behavior in `OpenAICompatibleLLM`.

DeepSeek accepts tool calls, but can reject explicit `tool_choice="required"` or named tool-choice values. This patch preserves forced-tool behavior by narrowing the tools list while omitting unsupported `tool_choice` values for DeepSeek.

It also adds a `reasoning_content=""` fallback for replayed assistant tool-call messages, because DeepSeek may reject assistant `tool_calls` history without this field.

## Changes

- Do not treat `deepseek-v4-flash` as a reasoning model.
- Keep DeepSeek reasoning classification for `v4-pro`, `reasoner`, `r1`, and `thinking` model names.
- Omit unsupported `tool_choice` values for DeepSeek while preserving narrowed tool-list behavior.
- Add `reasoning_content=""` fallback for assistant messages with `tool_calls`.
- Add regression tests for DeepSeek tool-call compatibility.

## Test plan

- `python3 -m py_compile hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py hindsight-api-slim/tests/test_deepseek_tool_call_compat.py`
- `PYTHONPATH=/app/api /app/api/.venv/bin/python -m pytest /tmp/test_deepseek_tool_call_compat.py -q` -> 5 passed
- `PYTHONPATH=/app/api /app/api/.venv/bin/python -m pytest /tmp/test_lmstudio_tool_choice.py -q` -> 10 passed
- `ruff check` on changed provider file and related tool-choice tests -> passed

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -206,7 +206,12 @@ class OpenAICompatibleLLM(LLMInterface):
     def _supports_reasoning_model(self) -> bool:
         """Check if the current model is a reasoning model (o1, o3, GPT-5, DeepSeek)."""
         model_lower = self.model.lower()
-        return any(x in model_lower for x in ["gpt-5", "o1", "o3", "deepseek"])
+        if "deepseek" in model_lower:
+            # DeepSeek v4-flash is the non-thinking route. Treating every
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM

PR #1256, opened 2026-04-24
feat(integrations): add Haystack integration

DK09876:
## Summary

- Adds `hindsight-haystack` package providing Haystack `Tool` instances backed by Hindsight's retain/recall/reflect APIs
- Uses async client methods (`aretain`, `arecall`, `areflect`) with an event-loop-safe sync wrapper (`_run_sync`) so tools work correctly inside Haystack's agent runtime (which runs its own event loop)
- Custom `_HindsightTool(Tool)` subclass with `to_dict()`/`from_dict()` for proper serialization (Haystack's default serializer can't handle bound methods)
- CI job, release script entry, 57 unit tests

## Test plan

- [x] `uv run pytest tests -v` — 57 tests pass
- [ ] E2E test against live Hindsight API with Haystack Agent
- [ ] Verify CI passes


benfrank241 (dismissed):
merge conflict

benfrank241:
one thing to fix before merge: `to_dict()` serializes the api key in cleartext (`backend_kwargs["api_key"]`, asserted in `test_tools.py`). haystack pipelines get dumped to yaml, so that leaks the key. please use haystack's `Secret` type (`Secret.from_env_var(...)`) or drop the key from the serialized payload and re-resolve it from env in `from_dict`, and
flip the test to assert it's absent.

also `verbose` in `config.py` is declared but never read.

DK09876:
Addressed in `febf3528`. Took option B (drop the key from the serialize) since `_client.py:resolve_client` already reads `HINDSIGHT_API_KEY` from env as a final fallback — that path already exists and matches the convention the other Python integrations follow ("API key from env or call-time arg only, never persisted").

Changes:
- `_build_backend_kwargs` no longer emits `api_key` in the returned dict; added a docstring explaining why.
- Flipped `test_tools_round_trip_serialization_with_client`: it now asserts `"api_key" not in d["data"]["backend_kwargs"]` and `"client-key" not in json.dumps(d)` (probe under all field names). The round-trip then sets `HINDSIGHT_API_KEY=env-key` and confirms `from_dict()` reconstructs with the env-var key.

Pre-fix, the new assertion fails with the literal leak the review caught:
```
AssertionError: api_key must not appear in serialized backend_kwargs — would leak to YAML pipeline dumps
assert 'api_key' not in {'api_key': 'client-key', ...}
```

86/86 tests pass post-fix.

PR #1271, opened 2026-04-26
fix(reflect): include tagged directives in isolation_mode when caller passed no tags

### LINKED ISSUE #1269: `reflect` silently drops every directive that has at least one tag (isolation_mode filter bug)
## Summary

Calling `reflect` on a memory bank that contains directives with non-empty `tags` returns synthesised text that ignores those directives entirely, and the `directives_applied` field in the response is always `[]`. Untagged directives (`tags: []` or `null`) are honored. The bug is in the directive loader filter inside `reflect`'s pipeline, not in directive storage or prompt rendering.

## Generic minimal repro

```
1. Create a memory bank.
2. Add a directive:
     name: "trailer"
     content: "End every response with: [VERIFIED]"
     priority: 10
     is_active: true
     tags: ["safety"]                # any non-empty tag list reproduces it
3. Add a few memories so the bank isn't empty.
4. Call reflect(query="summarize what you know")   # no `tags` argument

Expected: response text ends with [VERIFIED]; directives_applied
          contains the directive's id.
Actual:   response text does not contain [VERIFIED];
          directives_applied is [].

Side-channels that confirm the diagnosis:
  - reflect(query="...", tags=["safety"]) → directive IS applied.
  - Recreate the directive with tags=[] → directive IS applied
    even when no tags argument is passed to reflect.
  - list_directives(active_only=True) → directive is returned;
    storage layer is fine.
```

## Root cause (verified inside `ghcr.io/vectorize-io/hindsight:latest`)

`/app/api/hindsight_api/engine/memory_engine.py` lines ~5471-5479 — `reflect`'s directive load:

```python
directives_raw = await self.list_directives(
    bank_id=bank_id, tags=tags, tags_match=tags_match,
    active_only=True, request_context=request_context,
    isolation_mode=True,
)
```

`/app/api/hindsight_api/engine/memory_engine.py` lines ~6997-7005 — the `isolation_mode` branch in `list_directives`:

```python
elif isolation_mode:
    # Isolation mode: only include directives with empty/null tags
    filters.append("(tags IS NULL OR tags = '{}')")
```

When `reflect` is called without an explicit `tags` argument (the common case), `tags` is `None` and the loader hits the `elif isolation_mode` branch. The `(tags IS NULL OR tags = '{}')` filter then drops every directive that has any tags. The non-isolation branch immediately above already implements the correct semantics — "always include untagged directives; tagged ones must match the reflect tags" — so calling `list_directives` directly returns the full active set.

Downstream behavior is consistent with the symptom: `_build_directives_applied([])` → `[]`; `directive_rules` is `None`; `build_directives_section` / `build_directives_reminder` add nothing to the system prompt; the reflect agent pipeline runs against a prompt that contains zero directive text and synthesis cannot honor them.

## Suggested fix

Drop the tag filter in the `isolation_mode` branch when the caller passed no `tags`. One line. The intent of `isolation_mode` (preventing cross-tag leakage) is already served by the tagged branch above, which OR-merges untagged directives with tag-matched ones. For untagged callers, all active directives should be loaded.

```python
elif isolation_mode:
    # Untagged caller: include every active directive.
    # Cross-tag leakage prevention is handled by the tagged branch above.
    pass
```

## Why filing this

We're shipping a downstream proxy workaround in the agent framework that consumes hindsight, so we're not blocked. But this bug silently breaks the documented contract of `reflect` (directives ignored, no error, no warning), and we'd love to delete our workaround once a fixed image is published. Happy to send a PR if it'd help.

mekenthompson:
## Summary

Closes #1269.

`reflect` calls `list_directives(..., isolation_mode=True)` and forwards its own `tags` argument (which is `None` when the caller didn't pass tags). The `isolation_mode` branch in `list_directives` filters to `(tags IS NULL OR tags = '{}')`, silently dropping every directive that has at least one tag. The result: `reflect` quietly ignores all tagged directives on banks that mix tagged and untagged directives — `directives_applied` is empty, the synthesised text doesn't honor the rules, and there is no error or warning.

The intent of `isolation_mode` was to prevent cross-tag leakage, but an untagged caller has no tag scope to leak across. Cross-tag leakage prevention is already correctly implemented in the tagged branch above, which OR-merges untagged directives with tag-matched ones. So the fix is to drop the filter in the isolation branch when the caller passed no tags.

## Before / After

```diff
-            elif isolation_mode:
-                # Isolation mode: only include directives with empty/null tags
-                # This ensures tag-scoped directives don't apply to untagged operations
-                filters.append("(tags IS NULL OR tags = '{}')")
+            # If tags=None: include all active directives. The tagged branch above already
+            # prevents cross-tag leakage by OR-merging untagged directives with tag-matched
+            # ones; an untagged caller has no tag scope to leak across, so isolation_mode
+            # is a no-op here. (Previously this branch filtered to untagged-only, which
+            # silently dropped every directive that had any tag — see issue #1269.)
```

The `isolation_mode` parameter is preserved on `list_directives` for backwards compatibility; its docstring is updated to reflect that it is now a no-op for untagged callers. The misleading comment at the `reflect` call site is also updated. No callers or behaviour change for the tagged path.

## Regression test

Added `TestDirectiveTags::test_list_directives_isolation_mode_includes_tagged_when_no_tags` in `hindsight-api-slim/tests/test_mental_models.py`. It creates one untagged and one tagged directive, then calls `list_directives(tags=None, isolation_mode=True)` and asserts both directives are returned. Without the fix this test fails because the tagged directive is dropped.

## Verification

I was not able to run the test suite locally — the test harness boots an embedded Postgres (pg0) and pulls the full `hindsight-api-slim[all]` extras via `uv sync`, which exceeded the time budget I had for this PR. The change itself is one logical line (and a comment/docstring refresh); reviewers should run `uv run pytest hindsight-api-slim/tests/test_mental_models.py -k isolation_mode` locally to confirm.

## Why we care

We're a downstream consumer in the switchroom agent framework that relies on `reflect` honouring directives. We've shipped a proxy workaround on our side, but we'd like to delete it once a fixed image is published.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -5901,10 +5901,12 @@ class MemoryEngine(MemoryEngineInterface):
             async with pool.acquire() as conn:
                 return await tool_expand(conn, bank_id, memory_ids, depth)
 
-        # Load directives from the dedicated directives table
-        # Directives are hard rules that must be followed in all responses
-        # Use isolation_mode=True to prevent tag-scoped directives from leaking into untagged operations
-        # Use the same tags_match as the reflect request so directives respect the same scoping rules
+        # Load directives from the dedicated directives table.
+        # Directives are hard rules that must be followed in all responses.
+        # When the reflect call passes tags, list_directives OR-merges untagged directives
+        # with directives matching those tags (preventing cross-tag leakage). When tags is
+        # None, every active directive is loaded.
+        # Use the same tags_match as the reflect request so directives respect the same scoping rules.
         directives_raw = await self.list_directives(
             bank_id=bank_id,
             tags=tags,
@@ -7890,9 +7892,11 @@ class MemoryEngine(MemoryEngineInterface):
             limit: Maximum number of results
             offset: Offset for pagination
             request_context: Request context for authentication
-            isolation_mode: When True and tags=None, only return directives with no tags.
-                This prevents tag-scoped directives from leaking into untagged operations.
-                Default False (normal API behavior - returns all directives when tags=None)
+            isolation_mode: Retained for backwards compatibility. Currently a no-op:
+                an untagged caller (tags=None) has no tag scope to leak across, so all
+                active directives are returned regardless of this flag. Cross-tag leakage
+                prevention is handled by the tagged branch (tags!=None), which OR-merges
+                untagged directives with tag-matched ones. Default False.
 
         Returns:
             List of directive dicts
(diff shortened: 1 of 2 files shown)


mekenthompson:
Closing this PR — withdrawing on our side. Issue #1269 still describes the bug correctly if it's useful, but please disregard this PR.

PR #1275, opened 2026-04-27
fix: add Chinese first-person detection for fact classification

### LINKED ISSUE #834: First-person experience detection does not support Chinese (中文)
### Use Case

I'm running Hindsight as the memory backend for a Chinese-language AI agent system (OpenClaw). All conversations between the user and the AI assistant are conducted in Chinese. The agent performs tasks like Reddit monitoring, SEO keyword research, Feishu integration, and the extracted facts are predominantly in Chinese.

### Problem Statement

The experience fact type was introduced in v0.4.22 to classify first-person agent actions/experiences separately from world facts. However, after upgrading to 0.4.22, my main bank has 465 memories with 0 experience facts — every single one is classified as either world or observation.

The root cause is in fact_extraction.py:

The _FIRST_PERSON_PATTERN regex only matches English patterns like I fixed, I created, User configured, etc. Chinese first-person expressions (e.g., "我修复了", "助手完成了", "我发现了") are completely unmatched.

The LLM fact extraction prompt instructs the model to use "assistant" for first-person facts (which maps to experience), but when the conversation and extracted facts are in Chinese, the LLM tends to describe things in third person (e.g., "助手成功写入5条记录" instead of "I wrote 5 records"), so the LLM itself rarely outputs fact_type: "assistant".

As a result, the entire experience classification pipeline is effectively non-functional for Chinese-language memory banks.

With proper Chinese support, the experience fact type would correctly capture agent actions and learnings — things like "我通过 app_id + app_secret 获取 token 写入了飞书表格" or "助手从 Reddit 提取了75个关键词". This would enable better recall filtering (querying only experiences vs world facts) and richer Mental Model generation.

### Proposed Solution

1.Extend _FIRST_PERSON_PATTERN to include Chinese first-person markers:
_FIRST_PERSON_PATTERN = _re.compile(
    # English patterns (existing)
    r'(?:(?:^|[.!?]\s+)'
    r'(?:I|[Uu]ser)\s+'
    r'(?:fixed|debugged|traced|patched|refactored|implemented|deployed|'
    r'discovered|found|resolved|changed|updated|modified|created|built|'
    r'configured|installed|migrated|optimized|investigated|analyzed|'
    r'tested|verified|confirmed|learned|decided|chose|designed|wrote|'
    r'added|removed|deleted|merged|committed|pushed|pulled|reviewed|'
    r'diagnosed|troubleshot|upgraded|downgraded|reverted|set\s+up|'
    r'cleaned\s+up|wrapped\s+up|finished|completed|started|began))'
    # Chinese patterns (new)
    r'|(?:我|助手|助理|本人|agent)'
    r'(?:修复了|调试了|实现了|部署了|发现了|找到了|解决了|'
    r'更改了|更新了|修改了|创建了|构建了|配置了|安装了|'
    r'迁移了|优化了|调查了|分析了|测试了|验证了|确认了|'
    r'学到了|决定了|选择了|设计了|编写了|添加了|删除了|'
    r'完成了|开始了|搭建了|整理了|清理了|设置了|写入了|'
    r'提取了|采集了|抓取了|执行了|运行了|搞定了|处理了|'
    r'成功|尝试|记录了|读取了|导入了|导出了)',
    _re.IGNORECASE | _re.MULTILINE
)

2.Improve the LLM prompt for Chinese contexts — add Chinese examples in the fact extraction prompt showing when to use fact_type: "assistant":
示例：
- "助手从Reddit提取了75个关键词并写入飞书表格" → fact_type: "assistant"
- "王先生要求每周一执行Reddit监控" → fact_type: "world"
- "我发现 openclaw-lark 和 feishu 插件同时启用会冲突" → fact_type: "assistant"

### Alternatives Considered

Manually post-processing facts to reclassify Chinese first-person descriptions as experience via the API. This works but is fragile and doesn't scale.
Forcing the agent to converse in English so facts are extracted in English. Not practical for Chinese-speaking users.

### Priority

Nice to have

### Additional Context

_No response_

### Checklist

yujhuang:
Chinese-language facts are always classified as 'world' instead of 'experience' because the LLM prompt only contains English first-person examples. This patch adds:

1. CHINESE_FIRST_PERSON_PATTERN regex - matches Chinese first-person pronouns (我/助手/助理/本人/agent) + action verbs
2. _is_chinese_first_person() - fallback detection when LLM returns an unrecognized fact_type
3. Updated LLM prompt with Chinese and English first-person examples
4. Updated classification logic to try Chinese pattern detection before falling back to fact_kind

Fixes #834

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -69,6 +69,40 @@ def _sanitize_text(text: str | None) -> str | None:
     return sanitize_llm_output(text)
 
 
+# =============================================================================
+# FIRST-PERSON DETECTION PATTERNS (Fallback for LLM classification)
+# =============================================================================
+
+CHINESE_FIRST_PERSON_PATTERN = re.compile(
+    r"(?:我|助手|助理|本人|agent)"
+    r"(?:修复了|调试了|实现了|部署了|发现了|找到了|解决了|"
+    r"更改了|更新了|修改了|创建了|构建了|配置了|安装了|"
+    r"迁移了|优化了|调查了|分析了|测试了|验证了|确认了|"
+    r"学到了|决定了|选择了|设计了|编写了|添加了|删除了|"
+    r"完成了|开始了|搭建了|整理了|清理了|设置了|写入了|"
+    r"提取了|采集了|抓取了|执行了|运行了|搞定了|处理了|"
+    r"成功|尝试|记录了|读取了|导入了|导出)",
+    re.IGNORECASE,
+)
+
+
+def _is_chinese_first_person(fact_text: str) -> bool:
+    """
+    Check if text contains Chinese first-person agent action patterns.
+
+    This is a fallback when LLM fails to correctly classify fact_type.
+    Returns True if text matches Chinese first-person patterns.
+    """
+    if not fact_text:
+        return False
+
+    # Check main fact text
+    if CHINESE_FIRST_PERSON_PATTERN.search(fact_text):
+        return True
+
+    return False
+
+
 class Entity(BaseModel):
     """An entity extracted from text."""
 
@@ -930,8 +964,28 @@ def _build_user_message(
 
     narrator_section = ""
     if agent_name:
-        narrator_section = f'\nNarrator: {agent_name} (AI agent — first-person statements like "I did X" are the agent\'s own actions; classify as "assistant")'
+        narrator_section = f"""
+Narrator: {agent_name} (AI agent)
 
+FIRST-PERSON DETECTION (IMPORTANT):
+English examples (use fact_type="assistant"):
+- "I fixed bug in authentication module"
+- "I created 5 new user accounts"
+- "I discovered a performance issue"
+
+Chinese examples (use fact_type="assistant"):
+- "我修复了认证模块的bug"
+- "我创建了5个新用户账户"
+- "我发现了数据库性能问题"
+- "助手成功写入了飞书表格"
+- "助手从Reddit提取了75个关键词"
+- "我通过 app_id + app_secret 获取 token 写入了飞书表格"
+
+Third-person/world examples (use fact_type="world"):
+- "The system crashed at 3pm"
+- "User reported a login error"
+- "API latency increased by 50%"
+"""
     return f"""Extract facts from the following text chunk.
 
 Chunk: {chunk_index + 1}/{total_chunks}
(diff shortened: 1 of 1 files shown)


hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
```diff
@@ -69,6 +69,40 @@ def _sanitize_text(text: str | None) -> str | None:
...
+    r"迁移了|优化了|调查了|分析了|测试了|验证了|确认了|"
+    r"学到了|决定了|选择了|设计了|编写了|添加了|删除了|"
+    r"完成了|开始了|搭建了|整理了|清理了|设置了|写入了|"
+    r"提取了|采集了|抓取了|执行了|运行了|搞定了|处理了|"
+    r"成功|尝试|记录了|读取了|导入了|导出)",
+    re.IGNORECASE,
+)
+
+
+def _is_chinese_first_person(fact_text: str) -> bool:
```
nicoloboschi: this approach of regex is very fragile, please remove it

hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
```diff
@@ -930,8 +964,28 @@ def _build_user_message(
...
+        narrator_section = f"""
+Narrator: {agent_name} (AI agent)
+
+FIRST-PERSON DETECTION (IMPORTANT):
+English examples (use fact_type="assistant"):
+- "I fixed bug in authentication module"
+- "I created 5 new user accounts"
+- "I discovered a performance issue"
+
+Chinese examples (use fact_type="assistant"):
```
nicoloboschi: last time we tried to do this exact prompt cjanges, LLMs started to reply in chinese for english content - so I'm worried this could happen again

it would be good to have an eval dataset to put into a unit test to verify both english and chinese detection works as expected on ~10 different documents type

nicoloboschi (changes requested):

hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
yujhuang: Removed. The `_is_chinese_first_person` function and `CHINESE_FIRST_PERSON_PATTERN` regex have been deleted entirely. Fact type classification now relies solely on the LLM prompt.

hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
yujhuang: Good point. Two changes to address this:
    > 1. The prompt now uses side-by-side Chinese/English examples (instead of separate lists) and includes `CRITICAL: always output facts in the SAME language as the input text` — this avoids biasing the LLM toward Chinese for English content.
    > 2. Added `test_fact_type_multilingual.py` with 12 eval cases covering ~10 document types in both Chinese and English, including: prompt content validation (CRITICAL instruction present, balanced EN/ZH examples), fallback logic verification, and a regression check ensuring no regex code remains.

nicoloboschi:
I still see the regex - also I don't think this is a good solution in general. doing multilingual prompt it's not a solution that scales - closing this one for now

yujhuang:
The regex left was only a fallback in the merge. Besides, I think you'd better indicate more in issue #834 about the boundary or what is the best working direction for this enhancement from your perspective.

PR #1320, opened 2026-04-29
fix(embeddings): add allowed_openai_params for OpenAI-compatible embedding dimensions

zwcf5200:
## What

Fix the `allowed_openai_params` issue in `LiteLLMSDKEmbeddings` when using OpenAI-compatible custom embedding models.

## Problem

When using `litellm-sdk` as the embeddings provider with an OpenAI-compatible custom model (e.g., `openai/Qwen3-Embedding-4B-4bit-DWQ`), the `dimensions` parameter is silently rejected by litellm unless it is explicitly allow-listed via `allowed_openai_params`.

This means `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_OUTPUT_DIMENSIONS` has no effect — the embedding returns the model's default dimension instead of the configured one.

## Solution

Add `allowed_openai_params = ["dimensions"]` to the litellm kwargs when the model name starts with `openai/` and `output_dimensions` is set.

This affects both the async (`aembedding`) and sync (`embedding`) code paths in `LiteLLMSDKEmbeddings`.

## Changes

- `hindsight-api-slim/hindsight_api/engine/embeddings.py`: Add `allowed_openai_params` in 2 places (async and sync paths)
- `hindsight-api-slim/tests/test_litellm_sdk_embeddings.py`: Add test `test_openai_output_dimensions_allows_litellm_dimensions_param` and update existing tests to verify `allowed_openai_params` behavior

## Testing

- Unit tests pass (see `test_litellm_sdk_embeddings.py`)
- Verified locally with `litellm-sdk` + OpenAI-compatible custom embedding endpoint + `OUTPUT_DIMENSIONS=2000`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/embeddings.py
+++ b/hindsight-api-slim/hindsight_api/engine/embeddings.py
@@ -836,6 +836,8 @@ class LiteLLMSDKEmbeddings(Embeddings):
                 embed_kwargs["api_base"] = self.api_base
             if self.output_dimensions is not None:
                 embed_kwargs["dimensions"] = self.output_dimensions
+                if self.model.startswith("openai/"):
+                    embed_kwargs["allowed_openai_params"] = ["dimensions"]
 
             # Use async embedding method (standard in litellm)
             response = await self._litellm.aembedding(**embed_kwargs)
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM

PR #1337, opened 2026-04-29
feat(pydantic-ai): add optional Pydantic Logfire instrumentation

benfrank241:
## Summary

When the `[logfire]` extra is installed and `logfire.configure()` is called, every retain/recall/reflect tool call emits a semantically named span (`hindsight.retain` / `hindsight.recall` / `hindsight.reflect`) with bank_id, truncated query, tags, and result counts as attributes. Spans nest naturally under `agent.run` when used with `logfire.instrument_pydantic_ai()`.

This lays the foundation for the upcoming Pydantic blog post (`hindsight-marketing-content#112`) co-published with Pydantic.

## Design

- **Optional dep, zero-overhead fallback**: Without logfire installed, `_observability._logfire` is None and `span()` yields a `nullcontext()`. Existing users see no behavior change.
- **Privacy-conscious**: spans never carry raw memory content — only `bank_id`, query truncated to 200 chars, tags, and result counts/lengths. Safe for SOC 2 environments.
- **Surgical changes**: ~50-line helper module + four `async with span(...)` wrappers around the existing tool/instructions closures. No public API change.

## Files

- New: `hindsight_pydantic_ai/_observability.py` (helper)
- New: `tests/test_observability.py` (7 tests, exercises both no-op and logfire-installed paths via mocking)
- `hindsight_pydantic_ai/tools.py`: spans wrapped around the four tool/instructions closures
- `pyproject.toml`: `[project.optional-dependencies] logfire` extra; bumped version to 0.5.0
- `README.md`: new "Observability with Pydantic Logfire" section

## Test plan

- [x] `uv run pytest tests/ -v` — 44 tests pass (37 existing + 7 new)
- [x] `uvx ruff check .` — clean
- [x] Module imports cleanly when `logfire` is not installed (no-op path verified)
- [x] When logfire is patched in, `span()` produces a real span and propagates `set_attribute` calls
- [ ] Live end-to-end with a real LOGFIRE_TOKEN — pending PyPI publish or local dogfooding before merging

### DIFF
--- a/hindsight-integrations/pydantic-ai/README.md
+++ b/hindsight-integrations/pydantic-ai/README.md
@@ -131,6 +131,30 @@ instructions_fn = memory_instructions(
 )
 ```
 
+## Observability with Pydantic Logfire
+
+Every retain / recall / reflect call automatically emits a [Logfire](https://logfire.pydantic.dev) span when Logfire is installed and configured — no code changes to your agent. You see semantic spans like `hindsight.recall { bank_id, query, results_count }` nested under each Pydantic AI agent run.
+
+```bash
+pip install "hindsight-pydantic-ai[logfire]"
+```
+
+```python
+import logfire
+from pydantic_ai import Agent
+from hindsight_pydantic_ai import create_hindsight_tools
+
+logfire.configure()                # one line, anywhere at app startup
+logfire.instrument_pydantic_ai()   # nests Hindsight spans under agent.run
+
+agent = Agent(
+    "openai:gpt-4o",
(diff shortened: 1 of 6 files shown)


benfrank241:
Closing per feedback from Nicolò — Logfire is OpenTelemetry-based, and Hindsight's main codebase already has OTel support. The right fix is a config + guide pointing the existing OTel exporter at Logfire's OTLP endpoint, not a Logfire-specific helper inside an integration package. Reopening as a docs/guide PR instead.

PR #1343, opened 2026-04-29
fix(async-ops): atomically commit batch_retain parent and child rows

cdbartholomew:
## Summary

`submit_async_batch_retain` inserts a parent row (status='pending', `task_payload=NULL` — it's a status aggregator, not directly executable) and then loops to insert one child row per sub-batch. Today the parent INSERT and the child INSERTs are **not transactionally coupled**: the parent's INSERT runs in its own auto-committing connection, and each child goes through a separate `_submit_async_operation()` call that acquires its own connection.

Any failure between those steps — connection drop, asyncpg timeout, schema-cache invalidation under concurrent load, or any exception raised during child setup — leaves a parent row with zero children. From there:

- Worker poller skips the parent forever (the claim query filters `task_payload IS NOT NULL`).
- Status aggregator never fires for the parent (there are no children to drive it to a terminal state).
- The row sits `pending` indefinitely, growing without bound and inflating queue-depth metrics that operators rely on to size worker pools.

This is the same class of \"crash window leaves an unclaimable row\" bug that #1091 fixed at the child level (atomic INSERT-with-task_payload). The parent path was missed because the parent intentionally has `task_payload=NULL`, so the per-row atomicity fix doesn't apply — the cross-row atomicity has to come from a transaction.

## Fix

Wrap parent INSERT + all child INSERTs in a single `async with conn.transaction()` block. The child INSERT SQL is inlined for the duration of the transaction (so all writes share the same connection). `_submit_async_operation` is left untouched — other callers are unaffected.

`submit_task()` is deferred until after the transaction commits. Rationale: `SyncTaskBackend` (used in tests) executes the task synchronously, which would not see the still-uncommitted child row. `BrokerTaskBackend` and `WorkerTaskBackend` are effectively no-ops once `task_payload` is populated, but we defer them all uniformly for clarity.

Single-sub-batch and multi-sub-batch paths flow through the same code; the transaction is essentially free for the common case (single sub-batch, two INSERTs).

## Test plan

- [x] New regression test `test_submit_async_batch_retain_rolls_back_parent_on_child_failure` monkeypatches `BatchRetainChildMetadata` to raise on the second sub-batch and asserts zero `async_operations` rows remain (parent must roll back together with children). Mirrors the existing `test_submit_async_operation_leaves_claimable_row_when_submit_task_fails` test at the parent level.
- [x] Manual sanity: `test_request_context_retry_count_propagated_to_validator` and the new test both pass locally (the rest of the file's tests need an LLM API key configured and run in CI).
- [x] `ruff check` and `ruff format --check` clean.

## Out of scope

- Cleaning up any historical orphan parent rows that may already exist — that's an operational task, not part of this fix.
- Extending the claim reaper (which today only handles stale `processing` rows) to also auto-fail `pending` rows that have NULL `task_payload` and zero children. Useful as defense-in-depth; tracked separately.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -8943,57 +8943,100 @@ class MemoryEngine(MemoryEngineInterface):
             num_sub_batches=len(sub_batches),
         )
 
+        # Persist the parent row and all child rows in a single transaction.
+        #
+        # The parent row is a status aggregator with NO task_payload (workers
+        # skip rows where task_payload IS NULL because they're not directly
+        # executable). Its lifecycle is driven by child completions: when all
+        # children reach a terminal state, the parent gets promoted by the
+        # aggregator.
+        #
+        # If the parent INSERT and child INSERTs are not transactionally
+        # coupled, any failure between them (connection drop, timeout, schema
+        # cache invalidation under concurrent load) leaves a parent row with
+        # zero children. Workers ignore it forever (no task_payload), the
+        # aggregator never fires (no children to complete), and the row sits
+        # pending indefinitely — visible in queue-depth metrics and growing
+        # without bound. Wrapping parent + children in one transaction makes
+        # the create-batch operation atomic: either all rows are visible to
+        # workers, or none are.
+        #
+        # submit_task() must run AFTER the transaction commits. SyncTaskBackend
+        # (used in tests) executes the task synchronously, which would not see
+        # the still-uncommitted child row. BrokerTaskBackend / WorkerTaskBackend
+        # are effectively no-ops for already-populated task_payload, but we
+        # defer them all uniformly for clarity.
+        deferred_child_payloads: list[dict[str, Any]] = []
+
         async with acquire_with_retry(backend) as conn:
-            await conn.execute(
-                f"""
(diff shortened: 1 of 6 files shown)


cdbartholomew:
@nicoloboschi heads up — your PR #1325 added `ops=ops` to the call inside `fact_storage.handle_document_tracking` (line 312) but didn't add `ops` as a parameter to that function or thread it through the three call sites in `orchestrator._streaming_retain_batch`. Every retain hitting the document-tracking path raises `NameError: name 'ops' is not defined` on `main` right now — 16 tests failing across `test_delta_retain`, `test_retain`, `test_async_batch_retain`, `test_document_tracking`, `test_observation_invalidation`, `test_retain_append_mode`.

I needed CI green on this PR so I rolled the fix into commit `302331be`:

- Added `ops` as a kwarg-only parameter to `handle_document_tracking`
- Forwarded `pool.ops` from each of the three callers in `orchestrator.py`

Behaviorally it's the same outcome you intended: PG branch (your new native-array path) when `ops.uses_observation_sources_table` is False, Oracle branch (junction table) otherwise. Just plumbing the parameter you'd already started using.

Pinging in case you want to land it as a standalone fix faster than waiting for this PR to merge — it's blocking everyone else's CI on `main` too.

nicoloboschi (approved):
LGTM, nice catch

Issue #1345, opened 2026-04-29
Codex auto-retain includes synthetic setup context and can create huge full-session retain jobs


voarsh2:
## Problem

The Codex integration's auto-retain path can include synthetic Codex setup/context messages as if they were ordinary user conversation content. In a long-running Codex app thread, this produced a very large full-session retain job that Hindsight attempted to process as a single retain item.

This seems to be two related issues:

1. The Codex transcript reader treats injected project/setup context as normal user messages.
2. Full-session auto-retain can send one giant session document to structured extraction without a hard per-item cap or pre-extraction split.

## Steps to Reproduce

1. Use the Codex integration with hooks enabled.
2. Use a Codex thread in a project with AGENTS.md / project instructions loaded.
3. Enable auto-retain with full-session behavior.
4. Continue/resume a long Codex thread.
5. Let the Stop hook fire auto-retain.

Relevant local config shape:

```json
{
  "autoRecall": true,
  "autoRetain": true,
  "retainEveryNTurns": 4,
  "retainMode": "full-session",
  "retainToolCalls": false,
  "retainTags": ["test:auto-retain", "auto:codex-transcript"]
}
```

## Observed Behavior

The Codex JSONL transcript starts with entries like:

- `session_meta` containing base instructions
- `response_item`, role `developer`, containing sandbox/personality/tool instructions
- `response_item`, role `user`, containing `# AGENTS.md instructions for /home/coder/project ...`
- later `response_item`, role `user`, containing the actual user prompt

The Hindsight Codex parser keeps `response_item` messages where `payload.type == "message"` and role is `user` or `assistant`. Since Codex records injected `AGENTS.md` project instructions as a `user` message, auto-retain includes that setup text in the retained transcript.

In my test, the resulting auto-retain operation had:

- document id equal to the Codex session id: `019dcf04-df15-7781-ab2e-37499c9c1bcc`
- tags: `test:auto-retain`, `auto:codex-transcript`
- one retain item
- about `1,469,396` characters of content
- Hindsight estimated about `368,472` tokens from that one item

Server logs showed:

```text
Large batch detected (368,472 tokens from 1 items). Splitting into sub-batches of ~10,000 tokens each...
Split into 1 sub-batches: [1] items each
Processing sub-batch 1/1: 1 items, 368,472 tokens
```

The operation then became stuck in structured retain extraction against the configured LLM provider:

```text
stage=llm.openai.retain_extract_facts+structured
```

[...]

PR #1346, opened 2026-04-29
fix(codex): filter synthetic AGENTS startup messages

voarsh2:
## Summary

Filters Codex synthetic AGENTS.md startup messages out of the Codex integration transcript reader.

Codex can persist project startup instructions as a normal `response_item` with `role: "user"` and content shaped like:

`# AGENTS.md instructions for ...`
`<INSTRUCTIONS>...</INSTRUCTIONS>`

The Codex integration previously treated that as normal user conversation, so auto-retain and recall transcript composition could include agent rules/setup context instead of only the actual user/assistant session.

## Why

This can pollute retained memory with hook rules, memory policy, MCP setup instructions, and other agent-control text that the user did not actually say as part of the working conversation.

In long-running Codex sessions, this can also contribute to oversized auto-retain payloads because the retained transcript includes synthetic setup context before the real conversation starts.

## Implementation

- Adds a small detector for Codex AGENTS.md startup instruction messages.
- Skips those synthetic user messages in the text transcript reader.
- Skips the same synthetic messages in the rich transcript reader used when tool calls are retained.
- Leaves normal user discussion of AGENTS.md untouched.

## Validation

- `pytest hindsight-integrations/codex/tests/test_content.py`
- `pytest hindsight-integrations/codex/tests/test_hooks.py`
- Tested against a real Codex rollout transcript where the AGENTS startup block was persisted as a user message. After this change, the startup block is no longer present in parsed transcript output or retained chunk content.

### DIFF
--- a/hindsight-integrations/codex/scripts/lib/content.py
+++ b/hindsight-integrations/codex/scripts/lib/content.py
@@ -45,6 +45,25 @@ def strip_memory_tags(content: str) -> str:
     return content
 
 
+def is_synthetic_codex_user_message(content: str) -> bool:
+    """Return True for setup messages Codex persists as user chat.
+
+    Codex records AGENTS.md startup instructions as a normal user message in
+    rollout JSONL. Retaining those messages teaches Hindsight about hook and
+    agent rules instead of the user's work, and can create very large retain
+    payloads in long-running sessions.
+    """
+    if not isinstance(content, str):
+        return False
+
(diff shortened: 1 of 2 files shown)


voarsh2:
Also has some relevance to my issue https://github.com/vectorize-io/hindsight/issues/1345

nicoloboschi (approved):

Issue #1349, opened 2026-04-30
Document reprocess endpoint skips unchanged chunks instead of re-extracting facts


voarsh2:
## Bug Description

The document reprocess endpoint can report success without actually re-extracting facts when the document content is unchanged. This makes it ineffective for the documented use case of reprocessing a document after changing LLM model, extraction settings, bank retain mission, or document retain context.

In my local test, I updated an existing document's retained context and called the document reprocess endpoint. The endpoint returned success, but the retain pipeline took the delta-retain metadata-only path because all chunks were unchanged. Existing extracted facts, including known-bad/hallucinated facts, remained in place.

## Steps to Reproduce

1. Retain a document with a stable `document_id` and let Hindsight extract memory units.
2. Change something that should affect extraction but not source content, such as the LLM provider/model, bank retain mission, or the document's retain context/metadata.
3. Call `POST /v1/default/banks/{bank_id}/documents/{document_id}/reprocess`.
4. Inspect logs and memory units for that document.

Observed log excerpt from my local run:

```text
RETAIN_BATCH START: core-archive-clean-rich-a
[delta] Chunked new content: 5 chunks in 0.002s
[delta] Chunk diff: 5 unchanged, 0 changed, 0 new, 0 removed
[delta] No chunk changes detected — updating document metadata only
DELTA RETAIN (no changes): metadata updated in 0.151s
```

The API response was successful:

```json
{
  "success": true,
  "operation_id": "ebdaaa5a-6fc2-40f5-b6d4-189a50217c46",
  "items_count": 1
}
```

## Expected Behavior

Calling the document reprocess endpoint should force re-extraction of the document's facts using the current engine configuration, even when the original document text/chunks are unchanged.

That seems especially important because the endpoint description says it is useful when the LLM model, chunking strategy, or extraction settings have changed. If the endpoint intentionally preserves unchanged chunks, the response should clearly report that no facts were re-extracted rather than presenting as a successful reprocess.

## Actual Behavior

The endpoint completed successfully but delta retain skipped extraction because all chunks were unchanged. Existing memory units remained and were only metadata-updated. In my case, two known-bad extracted `world` facts remained after reprocess, confirming facts were not regenerated.

This makes the endpoint misleading for reprocessing after extraction configuration changes. The current workaround appears to be deleting the document and retaining it again with the same `document_id`, which is much more manual and riskier than a true force reprocess endpoint.

## Version

0.5.5 / current local main around Hindsight API 0.5.5

[...]

PR #1350, opened 2026-04-30
enhance(api): add force mode for document reprocessing

voarsh2:
## Summary

Adds an optional `force=true` mode to the document reprocess endpoint:

`POST /v1/default/banks/{bank_id}/documents/{document_id}/reprocess?force=true`

By default, document reprocess keeps the existing delta-aware behavior. With `force=true`, reprocess bypasses delta retain and streaming recovery so unchanged document content can still be fully re-extracted using the current bank/model/extraction configuration.

This is useful when the document text is unchanged but extraction behavior has changed, for example after tuning retain settings, model configuration, or stored retain context.

## Changes

- Add `force` query parameter and response field to the HTTP reprocess endpoint.
- Thread an internal `force_reprocess` flag through `MemoryEngine.reprocess_document`.
- Bypass delta retain when `force_reprocess` is set.
- Bypass streaming recovery reuse of already-committed chunks when `force_reprocess` is set.
- Thread the database operations helper through document re-ingest observation cleanup, matching the existing delete-document cleanup path.
- Add coverage for default delta behavior, force behavior, and HTTP `force=true`.

## Validation

- `python3 -m py_compile` on touched API/engine/test files
- `git diff --cached --check`
- Targeted container test run:
  `6 passed, 11 deselected`

Relevant issue: https://github.com/vectorize-io/hindsight/issues/1349

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1426,6 +1426,7 @@ class ReprocessDocumentResponse(BaseModel):
     success: bool
     operation_id: str
     items_count: int
+    force: bool = False
 
 
 class DeleteResponse(BaseModel):
@@ -4248,7 +4249,8 @@ def _register_routes(app: FastAPI):
         response_model=ReprocessDocumentResponse,
         summary="Reprocess document",
         description="Re-run the retain pipeline on an existing document without changing its content. "
-        "This deletes the existing memory units and re-extracts facts using the current engine configuration. "
+        "By default, unchanged chunks may be skipped by delta retain. Pass force=true to bypass delta retain "
+        "and re-extract facts for unchanged content using the current engine configuration. "
         "Useful when the LLM model, chunking strategy, or extraction settings have changed.",
         operation_id="reprocess_document",
         tags=["Documents"],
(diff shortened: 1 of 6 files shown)


nicoloboschi:
closing 

you can already delete the document and retain it again if you want to reprocess it from scratch

voarsh2:
Thanks. I understand delete + retain is a workaround, but I don’t think it is equivalent.

The reprocess endpoint is documented as re-running retain on the existing document without changing its content, and re-extracting facts with the current engine configuration. Because unchanged chunks can be skipped by delta retain / streaming recovery, that is not always what happens today.

Delete + retain requires the caller to reconstruct the original retain request from outside the API: original text, retain params, tags, metadata, timestamps, entities, scopes, and document id. For stored documents, Hindsight already has that canonical state.

The reason I’m pushing back is that, for my use case, documents in Hindsight are not disposable upload artifacts. They are the canonical stored memory source. Requiring operators to export/dump a document and replay it through retain is a brittle glue workflow, especially when the goal is simply to apply changed extraction/model/config behavior to existing stored content.

I’m open to a different API shape, but I think the distinction matters:

- delete + retain = client-side reconstruction of a memory source
- full reprocess = server-side re-extraction of the canonical stored source

For memory systems, that difference is operationally important.

@nicoloboschi

PR #1352, opened 2026-04-30
feat(opencode): share memory bank across git worktrees of the same repo

isac322:
## Summary

Adds a new `gitProject` value to `dynamicBankGranularity`, so users with multiple `git worktree`s of the same repo can opt into sharing a single memory bank across them.

Previously, when `dynamicBankId` was enabled, the `project` field was derived from `basename(directory)`. Linked worktrees of the same repository therefore ended up using different memory banks just because their filesystem paths differ — even though they are the same project and teams want their conventions/knowledge to apply across worktrees.

This PR keeps `project` semantics **unchanged** for backwards compatibility, and adds a separate `gitProject` field:

- `project` → `basename(directory)` (unchanged)
- `gitProject` → main worktree's basename, located via `git rev-parse --path-format=absolute --git-common-dir`. `git-common-dir` always points at the *main* worktree's `.git` even when invoked from a linked worktree, so every worktree of the same repo resolves to the same bank id. Bare repos use the bare directory's basename. Falls back to `basename(directory)` when git is unavailable / fails.

Existing users keep the default `["agent", "project"]` and see no change. Users who want worktrees to share a bank opt in:

```json
{
  "dynamicBankId": true,
  "dynamicBankGranularity": ["agent", "gitProject"]
}
```

`project` resolution is also moved to lazy evaluation so `git` is not spawned for granularities that don't include `gitProject`.

## Test plan
- [x] `bank.test.ts` cases added:
  - `project` stays directory-only inside a git repo (backwards compat)
  - `project` does not invoke git
  - `gitProject` regular clone resolves to the repo basename
  - `gitProject` linked worktree resolves to the same bank id as main
  - `gitProject` bare repo path used directly
  - `gitProject` git-unavailable fallback
  - static mode does not invoke git
  - `gitProject` not in granularity → git not invoked
  - `project` and `gitProject` can be combined as separate segments
- [x] `npm test` — 101 tests pass
- [x] `npm run build` — succeeds
- [x] `tsc --noEmit` — clean

## Notes
- No public API breakage; this is additive (new valid value for an existing array option).
- Implementation uses `execFileSync` with a 1s timeout and stderr suppressed; any failure is caught and treated as "not in a git repo".

### DIFF
--- a/hindsight-integrations/opencode/src/bank.ts
+++ b/hindsight-integrations/opencode/src/bank.ts
@@ -4,12 +4,13 @@
  * Port of Claude Code plugin's bank.py, adapted for OpenCode's context model.
  *
  * Dimensions for dynamic bank IDs:
- *   - agent   → configured name or "opencode"
- *   - project → derived from the main worktree's basename when inside a git
- *               repository (so all linked worktrees of the same repo share a
- *               single memory bank), falling back to the working directory
- *               basename when git is unavailable or the directory is not a
- *               repo.
+ *   - agent      → configured name or "opencode"
+ *   - project    → derived from the working directory basename
+ *   - gitProject → derived from the main worktree's basename when inside a
+ *                  git repository (so all linked worktrees of the same repo
+ *                  share a single memory bank). Falls back to the working
+ *                  directory basename when git is unavailable or the
+ *                  directory is not a repo.
  */
 
 import { basename, dirname } from "node:path";
@@ -19,7 +20,7 @@ import { debugLog } from "./config.js";
 import type { HindsightClient } from "@vectorize-io/hindsight-client";
 
 const DEFAULT_BANK_NAME = "opencode";
-const VALID_FIELDS = new Set(["agent", "project", "channel", "user"]);
+const VALID_FIELDS = new Set(["agent", "project", "gitProject", "channel", "user"]);
 
 /**
  * Resolve the main worktree root for a directory inside a git repository.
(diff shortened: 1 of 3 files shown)


hindsight-integrations/opencode/src/bank.ts
```diff
@@ -46,15 +95,17 @@ export function deriveBankId(config: HindsightConfig, directory: string): string
...
-  const fieldMap: Record<string, string> = {
-    agent: config.agentName || "opencode",
-    project: directory ? basename(directory) : "unknown",
-    channel: channelId || "default",
-    user: userId || "anonymous",
+  // Lazy resolution so we don't spawn `git` for `project` when the field
+  // isn't part of the configured granularity.
+  const fieldResolvers: Record<string, () => string> = {
+    agent: () => config.agentName || "opencode",
+    project: () => deriveProjectName(directory),
```
nicoloboschi: we need to keep backwards compatibility , can you add a new field gitProject with this new value?

isac322: Done — kept the previous `project` behavior intact and added `gitProject` as an opt-in value of `dynamicBankGranularity`. Existing users see no change unless they switch their granularity to `["agent", "gitProject"]`. Updated tests and README accordingly.

nicoloboschi (approved):
LGTM, thanks

af6140:
All my base dirs are named ".bare", so this will resolve to the same id for all my git worktree projects(clean hub layout).

Maybe just use git repo url reponame is more consistent?

PR #1358, opened 2026-04-30
fix(retain): thread ops into handle_document_tracking

youchi1:
## Bug

Regression from #1325. `fact_storage.py:312` was changed to call `delete_stale_observations_for_memories(..., ops=ops)` but `ops` was never added to `handle_document_tracking`'s signature, so every first-batch retain raises:

```
NameError: name 'ops' is not defined
```

## Effect

Every `batch_retain` task fails on the first-batch document tracking path. Poller retries, fails again, parent operation gets marked failed. No memories get stored. The gateway log misleadingly shows `retains captured` because dispatch succeeded.

## Fix

Add `ops=None` to `handle_document_tracking`'s signature. Pass `ops=pool.ops` from both call sites in `orchestrator._streaming_retain_batch`. 3 lines changed.

## Verified

Reproduced on a live deployment running main as of d8ec2d7f. Patch applied, retain plus background consolidation now succeed end to end (`created=6 updated=3` on first run after fix).

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_storage.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_storage.py
@@ -269,6 +269,7 @@ async def handle_document_tracking(
     is_first_batch: bool,
     retain_params: dict | None = None,
     document_tags: list[str] | None = None,
+    ops=None,
 ) -> None:
     """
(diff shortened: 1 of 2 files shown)


nicoloboschi:
fixed already in https://github.com/vectorize-io/hindsight/pull/1343

PR #1368, opened 2026-04-30
Harden OpenAI-compatible JSON responses and ensure json_object prompt hint

voarsh2:
## Summary

Fixes two OpenAI-compatible provider edge cases in structured JSON calls:

- ensure `response_format={"type":"json_object"}` requests include a lowercase `json` hint in a user message
- surface malformed success responses with clear `ProviderResponseError`s instead of crashing on `response.choices[0]`

Some OpenAI-compatible gateways enforce JSON mode’s “messages must contain json” requirement against the user/input message, and may return an error payload with `choices: null`. Previously this could become a generic `TypeError: 'NoneType' object is not subscriptable`, causing opaque retain extraction failures and wasted retries.

## Changes

- Adds a small JSON-mode user hint when needed: `Return valid json only.`
- Detects provider error payloads in success-shaped responses
- Detects missing choices, missing messages, and empty message content
- Avoids retrying deterministic provider error payloads
- Keeps retry behavior for transient/malformed response shapes
- Adds focused unit tests for JSON hint injection and malformed provider responses

## Testing

- `uv run pytest -q tests/test_openai_compatible_response_hardening.py`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -40,6 +40,15 @@ logger = logging.getLogger(__name__)
 
 # Seed applied to every Groq request for deterministic behavior
 DEFAULT_LLM_SEED = 4242
+JSON_MODE_USER_HINT = "Return valid json only."
+
+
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM

PR #1371, opened 2026-05-01
fix(consolidation): deduplicate observations by semantic similarity (#1284)

### LINKED ISSUE #1284: [Bug] Experience entries accumulate without further consolidation — entry count grows unboundedly
## Bug Description

After `retain()` fires, Hindsight creates atomic **observation** entries. Consolidation then processes observations and creates **experience** chunks. However, experience entries themselves are never consolidated further — duplicate experiences accumulate indefinitely.

**Flow:**
1. `retain()` → creates `observation` entries
2. Consolidation fires → creates `experience` chunks from observations
3. Observations are correctly superseded (marked `consolidated_at`)
4. **BUT:** experience entries never consolidate with each other → duplicates pile up

**Result:** Over sessions, the same knowledge appears as multiple experience entries. Entry count grows beyond what consolidation should achieve, eventually triggering "Memory's full" not from byte limits but from experience-count fragmentation.

## Evidence

From `hermes-main` bank (vectorize-io/hindsight#1266):
| Topic | Observations | Experiences (expected) | Actual experiences | Multiplier |
|-------|-------------|----------------------|-------------------|------------|
| MiniMax bug | 20 | ~5 | ~25 | 5x |
| RPi5 optimization | 8 | ~1 | ~9 | 9x |
| Telegram crash | 6 | ~3 | ~9 | 3x |

After 2 days of heavy session use: 60 entries where consolidation should have produced ~15.

## Corrected Model (per maintainer nicoloboschi, Apr 27)

The **original issue description was inverted** — it described observation/observation duplication, which is expected behavior. The actual bug is:

- `observation` = output of consolidation (not the problem)
- `experience` = duplicate facts that don't get consolidated further (the actual problem)

## Suggested Fix

Experiences should themselves be candidates for further consolidation:
1. **Periodic experience-level consolidation** — after N experiences accumulate for the same topic, merge them
2. **Deduplication pass** — before creating a new experience, check if a semantically similar one exists and merge rather than create new

## Priority

High — without this, Hindsight storage grows unboundedly and "Memory's full" fires prematurely from experience accumulation.

## Related

- Original (inverted) issue: vectorize-io/hindsight#1266 (CLOSED)
- Maintainer clarification: https://github.com/vectorize-io/hindsight/issues/1266#issuecomment-4326123032

ether-btc:
## Problem

Fixes #1284: **Experience entries accumulate without further consolidation — entry count grows unboundedly.**  
After `retain()` fires, observations are created from facts. When the LLM repeatedly fails to match new facts to existing observations, duplicate observations accumulate. The user reported 60 entries where ~15 was expected.

## Solution

Add a **semantic deduplication guard at observation CREATE time**:

1. Compute the embedding of the proposed observation text.
2. Query existing observations in the same `bank_id` + tag scope for cosine similarity ≥ `consolidation_observation_dedup_threshold` (default **0.92**).
3. If a match is found, merge the new source memories into the existing observation (UPDATE) instead of inserting a duplicate row.

This is transparent to the LLM — CREATE actions are automatically converted to UPDATEs when a semantic duplicate exists.

### New helpers

- `_find_semantic_duplicate_observation()` — vector-similarity lookup using the SQL dialect abstraction. Works on both **PostgreSQL** and **Oracle**.
- `_merge_into_observation()` — merges source memories, tags, and temporal bounds into an existing observation without requiring the `MemoryFact` model. Accepts an optional `embedding_str` parameter to avoid double generation on the dedup path.

### Configuration

| Setting | Default | Env Var |
|---------|---------|---------|
| `consolidation_observation_dedup_threshold` | `0.92` | `HINDSIGHT_API_CONSOLIDATION_OBSERVATION_DEDUP_THRESHOLD` |

Set to `0.0` to disable deduplication.

## Security

- Always scoped to the same `bank_id`.
- Tags are passed raw to `dialect.array_contains()` for cross-database compatibility (no `json.dumps`).
- Source memory IDs are deduplicated before merge to prevent `proof_count` inflation.

## Test coverage

- `tests/test_consolidation_observation_dedup.py` — unit tests mocking DB + embeddings:
  - No match below threshold
  - Match above threshold
  - Tags filter uses correct parameter types
  - Merge updates fields correctly
  - Merge skips when all new sources are deleted
  - Dedup path on `_create_observation_directly`
  - Normal INSERT when threshold is 0

## Checklist

- [x] Code compiles (`py_compile` clean)
- [x] Dual-verdict review completed (Tech Lead + Software Engineer)
- [x] SE-requested fixes applied: dialect abstraction, raw tag params, deduplicated source IDs, embeddeding reuse
- [x] New unit tests added
- [x] Config fields added with env-var fallbacks

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -352,6 +352,7 @@ ENV_CONSOLIDATION_SOURCE_FACTS_MAX_TOKENS_PER_OBSERVATION = (
 )
 ENV_CONSOLIDATION_RECALL_BUDGET = "HINDSIGHT_API_CONSOLIDATION_RECALL_BUDGET"
 ENV_CONSOLIDATION_MAX_ATTEMPTS = "HINDSIGHT_API_CONSOLIDATION_MAX_ATTEMPTS"
+ENV_CONSOLIDATION_OBSERVATION_DEDUP_THRESHOLD = "HINDSIGHT_API_CONSOLIDATION_OBSERVATION_DEDUP_THRESHOLD"
 ENV_OBSERVATIONS_MISSION = "HINDSIGHT_API_OBSERVATIONS_MISSION"
 ENV_MAX_OBSERVATIONS_PER_SCOPE = "HINDSIGHT_API_MAX_OBSERVATIONS_PER_SCOPE"
 ENV_ENABLE_OBSERVATION_HISTORY = "HINDSIGHT_API_ENABLE_OBSERVATION_HISTORY"
@@ -612,6 +613,7 @@ DEFAULT_CONSOLIDATION_SOURCE_FACTS_MAX_TOKENS_PER_OBSERVATION = (
 )
 DEFAULT_OBSERVATIONS_MISSION = None  # Declarative spec of what observations are for this bank
 DEFAULT_MAX_OBSERVATIONS_PER_SCOPE = -1  # Max observations per tag scope (-1 = unlimited)
+DEFAULT_CONSOLIDATION_OBSERVATION_DEDUP_THRESHOLD = 0.92  # Cosine similarity threshold for deduplicating observations during consolidation (0 = disabled)
 
 # Database migrations
 DEFAULT_RUN_MIGRATIONS_ON_STARTUP = True
@@ -1034,6 +1036,7 @@ class HindsightConfig:
     consolidation_source_facts_max_tokens: int
     consolidation_source_facts_max_tokens_per_observation: int
     consolidation_max_attempts: int
+    consolidation_observation_dedup_threshold: float
     observations_mission: str | None
     max_observations_per_scope: int
 
@@ -1182,6 +1185,7 @@ class HindsightConfig:
         "consolidation_max_memories_per_round",
         "consolidation_source_facts_max_tokens",
         "consolidation_source_facts_max_tokens_per_observation",
+        "consolidation_observation_dedup_threshold",
         "observations_mission",
         "max_observations_per_scope",
         # Reflect settings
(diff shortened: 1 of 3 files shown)


nicoloboschi:
thanks for thinking througth this but I don't like this approach. The "intelligence" layer is hand off to the LLM in Hindsight. And discarding duplicates it's something not obvious at all and the cosine similarity (alone) can't fix at all.

For example, each memory fact has a temporal aspect which is very relevant. Cosine similary will not detect this if not explictly in the observation text (sometimes it's not stated but it's only in the occurred time). 

Also "deduplication" is subjective. Are 2 events happened with distance of 10 minutes with same text, the same event? In some cases, they could, in some others they are not.

So the knob to rotate is the Observation Mission. if that's not working, we need to investigate either is a model limitation or something that can be prompt optimized or modified in a more consistent agent loop (ending up slowing down consolidation and using more tokens)

ether-btc:
Fair points, all three are valid critiques. A few thoughts:

**On temporal aspect**: Agreed cosine similarity alone misses temporal context. But the dedup guard operates at CREATE time, not post-hoc. The incoming `MemoryFact` carries temporal metadata (`observed_at`). The comparison could be gated: only consider dedup against observations within the same temporal window (e.g., same day or within N minutes). High cosine similarity + same temporal bucket -> strong signal for duplicate. High cosine similarity + different temporal bucket -> let it through. This partially addresses the subjectivity concern.

**On subjectivity**: This is exactly why the threshold is configurable (default 0.92) and can be set to 0.0 to disable entirely. It's intentionally an escape hatch for users who want the behavior and a no-op for those who don't. The PR description frames it as a knob, not a hard mandate.

**On Observation Mission being the real fix**: Fully agree that better prompt/model is the principled solution. But there's a practical asymmetry: prompt tuning is a global change that affects all users and all consolidations; a dedup threshold is a per-bank opt-in. They address different layers of the problem.

**Compromise**: What if we narrow the scope? Instead of general semantic dedup, restrict it to *verbatim or near-verbatim* duplicates only -- threshold 0.98+ -- where the subjectivity argument is weakest (e.g., same text embedded in observation is almost certainly a duplicate LLM generation artifact, not a genuinely distinct event). That makes it a narrower, more defensible fix for the artifact problem rather than a semantic intelligence layer.

Happy to refine if you have a preferred direction.

PR #1374, opened 2026-05-01
fix(api): include observations in per-document graph and counts

youchi1:
## Summary

The control plane's document detail view ships an **Observations** tab and a **Memory Composition** card alongside *World* and *Experience*. Both were permanently empty for every document.

## Root cause

`get_graph_data` and `get_document` filter `memory_units` by `document_id` (and `chunk_id`) with a direct equality. Observations are consolidated rows whose `document_id` / `chunk_id` columns are always `NULL`; the link back to a document lives on `source_memory_ids` (PG) or in the `observation_sources` junction table (Oracle). The equality filter therefore excludes every observation, regardless of how its source memories were extracted.

## Fix

- Added `MemoryEngine._observations_via_source_match_sql`: a backend-aware predicate matching observations whose source memories satisfy a column equality, scoped to a bank. One helper now serves both call sites and any future ones.
- `get_graph_data`: extend `document_id` / `chunk_id` filters with an `OR` branch via the helper so observations linked through their sources are returned. Bank-scope the inner subquery.
- `get_document`: replace the broken `observation_count` aggregation with a `COUNT(*)` subquery built on the same helper. `nodes_by_fact_type.observation` now reflects observations for the document.

## Test plan

- [x] New regression test `test_graph_document_filter_includes_observations_via_source_memories` seeds a document with a source fact, an observation linked via `source_memory_ids`, and an unrelated observation. Verifies the graph endpoint returns only the linked observation when filtering by `document_id`. Confirmed to fail without the fix.
- [x] Existing `test_get_document_nodes_by_fact_type` assertion updated: `memory_unit_count` covers facts with `document_id` (world + experience); observations are reported separately in `nodes_by_fact_type`.
- [x] `tests/test_graph_filtering.py` and `tests/test_document_chunks_and_reprocess.py` pass (19 passed, 1 pre-existing skip).
- [x] `ruff check` / `ruff format` clean on touched files.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -3723,6 +3723,36 @@ class MemoryEngine(MemoryEngineInterface):
 
         return filtered_results, total_tokens
 
+    def _observations_via_source_match_sql(
+        self,
+        source_column: str,
+        source_placeholder: int,
+        bank_placeholder: int | None,
+    ) -> str:
+        """SQL predicate matching `memory_units` rows that are observations
+        whose source memories satisfy ``<source_column> = $source_placeholder``.
+
+        Observations have no `document_id` / `chunk_id` of their own; the link
+        to a source row lives in `source_memory_ids` (PG) or the
+        `observation_sources` junction (Oracle).
+        """
+        if source_column not in ("document_id", "chunk_id"):
+            raise ValueError(f"Unsupported source_column: {source_column!r}")
+        if self._backend.ops.uses_observation_sources_table:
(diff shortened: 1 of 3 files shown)


youchi1:
The verify-generated-files failure looks pre-existing on main, not from this PR. Running `./scripts/generate-docs-skill.sh` on a clean checkout of current origin/main (84c33b10) produces the same diff to `skills/hindsight-docs/references/developer/{monitoring,storage}.md` (Oracle AI Database paragraphs that aren't in the snapshot yet). None of the three files in the failure are touched here.

Happy to fold a regen commit in if that's preferred, but figured a separate chore commit on main is cleaner.

nicoloboschi (approved):
nice catch! thanks

PR #1386, opened 2026-05-02
feat(helm): add configurable consolidation options

skyline75489:
## Description

Add a new `consolidation` values section to the Helm chart that exposes all observation consolidation settings as configurable environment variables.

## Changes

- **values.yaml**: Added `consolidation.` block with 12 optional settings (all commented out by default, safe defaults used when unset)
- **api-deployment.yaml**: Injects consolidation env vars into the API container (conditional — only when explicitly set)
- **worker-statefulset.yaml**: Same consolidation env vars for the worker container
- **Chart.yaml**: Bump version 0.5.5 → 0.5.6

## Consolidation settings exposed

| Setting | Env var |
|---|---|
| `enableObservations` | `HINDSIGHT_API_ENABLE_OBSERVATIONS` |
| `maxMemoriesPerRound` | `HINDSIGHT_API_CONSOLIDATION_MAX_MEMORIES_PER_ROUND` |
| `batchSize` | `HINDSIGHT_API_CONSOLIDATION_BATCH_SIZE` |
| `llmBatchSize` | `HINDSIGHT_API_CONSOLIDATION_LLM_BATCH_SIZE` |
| `maxTokens` | `HINDSIGHT_API_CONSOLIDATION_MAX_TOKENS` |
| `recallBudget` | `HINDSIGHT_API_CONSOLIDATION_RECALL_BUDGET` |
| `sourceFactsMaxTokens` | `HINDSIGHT_API_CONSOLIDATION_SOURCE_FACTS_MAX_TOKENS` |
| `sourceFactsMaxTokensPerObservation` | `HINDSIGHT_API_CONSOLIDATION_SOURCE_FACTS_MAX_TOKENS_PER_OBSERVATION` |
| `maxAttempts` | `HINDSIGHT_API_CONSOLIDATION_MAX_ATTEMPTS` |
| `llmProvider` | `HINDSIGHT_API_CONSOLIDATION_LLM_PROVIDER` |
| `llmModel` | `HINDSIGHT_API_CONSOLIDATION_LLM_MODEL` |
| `llmBaseUrl` | `HINDSIGHT_API_CONSOLIDATION_LLM_BASE_URL` |

## Backward compatibility

All env vars use `hasKey` checks — existing deployments with no `consolidation:` block are completely unaffected.

### DIFF
--- a/helm/hindsight/Chart.yaml
+++ b/helm/hindsight/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: hindsight
 description: Hindsight helm chart
 type: application
-version: 0.5.5
-appVersion: "0.5.5"
+version: 0.5.6
+appVersion: "0.5.6"
 keywords:
   - ai
   - memory
--- a/helm/hindsight/templates/api-deployment.yaml
+++ b/helm/hindsight/templates/api-deployment.yaml
@@ -63,6 +63,57 @@ spec:
         {{- /* Explicitly set port to override K8s service discovery env var (HINDSIGHT_API_PORT) */}}
         - name: HINDSIGHT_API_PORT
           value: {{ .Values.api.service.targetPort | quote }}
+        {{- /* Consolidation/observation settings */}}
+        {{- if .Values.consolidation }}
+        {{- if hasKey .Values.consolidation "enableObservations" }}
+        - name: HINDSIGHT_API_ENABLE_OBSERVATIONS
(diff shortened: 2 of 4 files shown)


nicoloboschi (changes requested):
hey thanks for the PR but we prefer to not expose single env vars via values due to the cost on maintainance. Also keeping direct env in the chart is easier for users since they can directly find the same exact config in the documentation

PR #1389, opened 2026-05-02
feat(anthropic): env-driven max_retries + default_headers knobs

TuftyBruno:
## Summary

Two opt-in env vars on AnthropicLLM.__init__. Both no-op when unset, so existing deployments are unaffected.

- HINDSIGHT_API_LLM_MAX_RETRIES (int) — when set, overrides the Anthropic SDK's default retry count. Useful when the deployment has its own outer retry layer (Hindsight's call() already does 2s→300s exponential backoff) and the SDK's stacked auto-retries would compound 429s.
- HINDSIGHT_API_LLM_DEFAULT_HEADERS (JSON string) — when set, parsed and passed as default_headers to AsyncAnthropic. Useful when routing through a proxy that needs custom headers (component attribution, client-fingerprint markers, etc).

## Real-world driver

Routing Hindsight through a custom proxy (in our case, Switchboard) that:
- handles retry policy at the proxy layer
- needs X-Component-Id for per-component cost attribution
- needs X-SB-Impersonate-CC: 1 for Anthropic OAuth fingerprint compat

Without these env knobs, operators have to volume-mount a patched anthropic_llm.py into the container, which is fragile across image upgrades.

## Test plan

- [x] Default behavior unchanged when env vars unset
- [x] Invalid HINDSIGHT_API_LLM_MAX_RETRIES (non-int) logs a warning and is ignored
- [x] Invalid HINDSIGHT_API_LLM_DEFAULT_HEADERS (non-JSON) logs a warning and is ignored
- [x] Verified locally on container 0.5.6 with our deployment

## Diff

Surgical: 1 import (os) + 21 lines in the constructor. No refactor, no new dependencies.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
@@ -11,6 +11,7 @@ This provider enables using Claude models from Anthropic with support for:
 import asyncio
 import json
 import logging
+import os
 import time
 from typing import Any
 
@@ -65,6 +66,26 @@ class AnthropicLLM(LLMInterface):
                 client_kwargs["base_url"] = self.base_url
             if timeout:
                 client_kwargs["timeout"] = timeout
+
+            # Optional env-driven knobs for operators with custom proxy / retry policy needs.
+            # Both no-op when unset, so existing deployments are unaffected.
+            _max_retries_env = os.environ.get("HINDSIGHT_API_LLM_MAX_RETRIES")
+            if _max_retries_env is not None:
+                try:
+                    client_kwargs["max_retries"] = int(_max_retries_env)
+                except ValueError as exc:
+                    logger.warning(
+                        "HINDSIGHT_API_LLM_MAX_RETRIES is not an integer, ignoring: %s", exc
+                    )
+
+            _default_headers_env = os.environ.get("HINDSIGHT_API_LLM_DEFAULT_HEADERS")
+            if _default_headers_env:
+                try:
+                    client_kwargs["default_headers"] = json.loads(_default_headers_env)
+                except json.JSONDecodeError as exc:
+                    logger.warning(
+                        "HINDSIGHT_API_LLM_DEFAULT_HEADERS is not valid JSON, ignoring: %s", exc
+                    )
 
             self._client = AsyncAnthropic(**client_kwargs)
             logger.info(f"Anthropic client initialized for model: {self.model}")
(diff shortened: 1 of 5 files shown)


hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
```diff
@@ -66,6 +67,26 @@ def __init__(
             if timeout:
                 client_kwargs["timeout"] = timeout
 
+            # Optional env-driven knobs for operators with custom proxy / retry policy needs.
+            # Both no-op when unset, so existing deployments are unaffected.
+            _max_retries_env = os.environ.get("HINDSIGHT_API_LLM_MAX_RETRIES")
```
nicoloboschi: can we use the usual path for using config.py? pls check other providers

nicoloboschi (changes requested):

TuftyBruno:
Pushed `8afb7b8` addressing the review. Refactor goes through `config.py` like the other providers:

- **`config.py`**: added `ENV_LLM_DEFAULT_HEADERS` + `DEFAULT_LLM_DEFAULT_HEADERS` constants and a static `llm_default_headers` field on `HindsightConfig`, parsed in `from_env()` the same way `llm_extra_body` and `llm_gemini_safety_settings` are.
- **`anthropic_llm.py`**: dropped the inline `os.environ.get()` reads + the `import os`. `default_headers` is now a typed `__init__` kwarg, sourced from config. Hardcoded `max_retries=0` on the SDK client to mirror `OpenAICompatibleLLM` (line 179) — the wrapper retry loop in `call()` already handles backoff, so SDK retries are double work. Dropped the new `HINDSIGHT_API_LLM_MAX_RETRIES` SDK-level use entirely; the existing variable still drives `HindsightConfig.llm_max_retries` for the wrapper.
- **`llm_wrapper.py`**: threaded `default_headers` through `create_llm_provider()` and `LLMProvider.__init__` / `from_env`. Falls back to `_get_raw_config().llm_default_headers` when not explicitly passed (same pattern as `gemini_safety_settings`).
- **`memory_engine.py`**: passes `config.llm_default_headers` to all four `LLMConfig` constructors (memory / retain / reflect / consolidation), parallel to how `config.llm_extra_body` is already passed.
- **`hindsight-docs/docs/developer/configuration.md`**: documented `HINDSIGHT_API_LLM_DEFAULT_HEADERS` in the LLM table.

Heads-up on `max_retries`: with the new behavior the Anthropic SDK client always runs with `max_retries=0`, where previously it defaulted to the SDK's built-in `2` (and was overridable via env in the prior version of this PR). Net retries are now wrapper-only, matching `OpenAICompatibleLLM`. Happy to revisit if you'd prefer keeping SDK retries on for Anthropic specifically.

\`uv run ruff check .\` + \`uv run ruff format --check .\` both clean on \`hindsight-api-slim\`.

nicoloboschi (approved):
LGTM

PR #1395, opened 2026-05-02
feat(memory): validity windows on memory_units (`valid_to` + `/invalidate` endpoint)

### LINKED ISSUE #1391: Feature: validity windows on memory_units (`valid_to` + `/invalidate` endpoint)
## Summary

Add a `valid_to TIMESTAMPTZ NULL` column to `memory_units` and an `/v1/default/banks/{bank_id}/memories/{memory_id}/invalidate` endpoint, so that *superseded* facts can be soft-retired without losing their timeline. Recall would skip rows with `valid_to <= now()` while history endpoints continue to return them.

## Motivation

Today, when a fact in Hindsight becomes stale (a server is decommissioned, a person changes role, a default value changes), the only options are:

1. **Delete** the row → audit trail is gone.
2. **Leave it** → recall returns it alongside the new, correct fact, which the LLM has to disambiguate at recall time. With many such "ghosts" the agent's accuracy degrades.
3. **Retain a contradicting newer fact and hope consolidation prefers it** → unreliable, especially across consolidation lags.

Other learning-oriented memory systems make validity a first-class field. For comparison, [MemPalace](https://github.com/MemPalace/mempalace)'s knowledge graph stores `valid_from` / `valid_to` on every triple and `attributes` row, with a dedicated `invalidate` operation. The `triples` table in [docs/schema.sql](https://github.com/MemPalace/mempalace/blob/main/docs/schema.sql) is the cleanest small example I know of.

In day-to-day Hindsight use we keep wanting this exact thing: a way to mark "this fact is no longer true, but for audit purposes don't drop the row." Today we approximate it with tags + manual filters, which doesn't scale and is invisible to recall.

### 1. Schema

```sql
ALTER TABLE memory_units
    ADD COLUMN valid_to TIMESTAMPTZ NULL;

CREATE INDEX idx_memory_units_active
    ON memory_units (bank_id, fact_type)
    WHERE valid_to IS NULL;

COMMENT ON COLUMN memory_units.valid_to IS
    'NULL = still valid; non-NULL = superseded at this timestamp. Recall filters out memories with valid_to <= now().';
```

The partial index keeps recall hot-path SQL (which is by far the largest hot-spot for filtering) cheap regardless of how much historical data accumulates.

### 2. Recall filter

Add `AND (valid_to IS NULL OR valid_to > now())` to the four WHERE clauses in `hindsight_api/engine/search/retrieval.py`:

- semantic UNION ALL arms (per `fact_type`)
- BM25 UNION ALL arms (per `fact_type`)
- date-range `entry_points` CTE
- graph-spreading neighbour join

That's a one-line addition per spot; the partial index covers the hot path.

### 3. API

```
POST /v1/default/banks/{bank_id}/memories/{memory_id}/invalidate
{
  "valid_to": "<ISO-8601>",          // optional; defaults to now()
  "reason":   "<free text>"          // optional; appended to metadata.invalidation_reason
}
```

Response: the updated row's `id`, `valid_to`, `fact_type`, and a short `text` preview, mirroring `GET /memories/{memory_id}`.

`GET /memories/{memory_id}` and `GET /memories/{memory_id}/history` continue to return invalidated rows — only `recall` filters them. That preserves the audit / timeline semantics that motivate the feature.

### 4. Mental model

`valid_to` is **separate from** `mentioned_at` / `event_date` / `occurred_*`:

| Column                  | Means                                                  |
|-------------------------|--------------------------------------------------------|
| `event_date` / `occurred_*` | When the fact was *true* in the world                |
| `mentioned_at`          | When the fact was *retained* (mentioned) into the bank |
| **`valid_to`**          | When we *stopped considering* the fact valid           |

So a fact about an event that happened in 2023 can be retained today (`mentioned_at = now()`) and invalidated tomorrow (`valid_to = tomorrow`). All three are independent.

## Local proof-of-concept

I've been running this locally for a Hermes Agent install (managed Hindsight bank, ~1.4k memory units, ~720 entities) for a few days. The local patch is a text-replace into the installed `retrieval.py` plus an out-of-band DDL migration plus a new MCP tool that calls `UPDATE memory_units SET valid_to = …` directly. End-to-end it works:

- Insert a memory → recall returns it.
- Call `invalidate(memory_id)` → recall stops returning it; `GET /history` still shows it.
- Re-recall N=91 → invalidated row is filtered out.

Source of the patch (just for reference, not what I'd PR): https://github.com/<user>/<vm-management-repo>/blob/main/bin/hermes-patch-valid-to (it's a hack — a clean PR would do this through a real alembic migration and the SQL builder rather than text-replacing the file).

## Tradeoffs / questions

1. **Naming.** `valid_to` is the most common terminology. Alternatives: `superseded_at`, `retired_at`. I'd default to `valid_to` since it pairs naturally with the eventual `valid_from` (already implicit in `mentioned_at`/`event_date` for now, but could become explicit if you adopt this pattern more broadly).

2. **`/invalidate` HTTP verb.** I picked `POST` because the operation is non-idempotent w.r.t. metadata (each call overwrites `invalidation_reason`). `PATCH /memories/{id}` with a body containing `{valid_to: ...}` is also reasonable; let me know which you prefer.

3. **Worker-side awareness.** Consolidation/reflect workers should probably skip `valid_to IS NOT NULL` rows by default. I'd handle that in the same PR.

4. **Confidence interplay.** This is **not** the same as confidence ≤ threshold. A fact can be high-confidence and invalidated; or low-confidence and still active.

5. **`as_of` on recall.** A natural follow-up: pass `as_of: <ISO>` to recall to "see what was active at that point." Out of scope for this issue but cheap once the column exists.

## What I'd contribute

If the design is acceptable, I'd open a draft PR with:

- Alembic migration `<id>_add_valid_to_to_memory_units.py` (revises `z1u2v3w4x5y6`)
- `MemoryUnit.valid_to` field in `models.py`
- Filter injection in the four WHERE blocks of `engine/search/retrieval.py`
- `POST /memories/{memory_id}/invalidate` endpoint in `api/http.py` + memory-engine method
- pytest covering: insert → recall (returns) → invalidate → recall (filtered) → `/history` (still returns) → invalidate-then-re-invalidate (no-op or update reason)
- Cookbook example or docs/`/concepts/` page for the validity semantics

I'm happy to scope down/up — wanted to confirm the design first before pushing 200+ lines of diff.

## Related

- MemPalace: triple/attribute validity windows ([schema.sql](https://github.com/MemPalace/mempalace/blob/main/docs/schema.sql))
- Existing Hindsight bookkeeping columns that are *time-shaped but different*: `event_date`, `occurred_start`, `occurred_end`, `mentioned_at`, `consolidated_at`, `consolidation_failed_at`. None of those answer "is this fact still valid today?"

Happy to revise the design or split into smaller pieces if useful.

nikolay-bratanov:
Closes #1391.

## Summary

Adds a nullable `valid_to TIMESTAMPTZ` column to `memory_units`, a partial index limited to `valid_to IS NULL` rows, and a `POST /memories/{memory_id}/invalidate` endpoint so that *superseded* facts can be soft-retired without losing their timeline. Recall queries skip rows whose `valid_to` has elapsed; `GET /memories/{id}` and `GET /memories/{id}/history` still return them, preserving the audit trail.

## What's included

- **Migration** `a2b3c4d5e6f7_add_valid_to_to_memory_units.py` — `ALTER TABLE memory_units ADD COLUMN valid_to TIMESTAMPTZ NULL` plus partial index `idx_memory_units_active (bank_id, fact_type) WHERE valid_to IS NULL`. Both `IF NOT EXISTS` so the migration is idempotent. Down-migration drops both.
- **Model** `MemoryUnit.valid_to` and the corresponding `Index` in `__table_args__`.
- **Recall filter** in `engine/search/retrieval.py` — a `validity_clause = "AND (valid_to IS NULL OR valid_to > {dialect.current_timestamp()})"` is composed once and injected into the four WHERE blocks via the existing `extra_where` parameter on `build_semantic_arm` / `build_bm25_arm` (semantic UNION arms, BM25 UNION arms, date-range entry-points CTE, graph-spreading neighbour join). The Oracle-fallback semantic-only path also gets the filter. No dialect-level changes — portability comes from `dialect.current_timestamp()`.
- **Engine** `MemoryEngine.invalidate_memory_unit(bank_id, memory_id, *, valid_to=None, reason=None, request_context)` — single UPDATE with explicit `$3::timestamptz` cast (avoids `AmbiguousParameterError` when `valid_to` is also referenced inside JSONB metadata in the same statement). Default `valid_to` is UTC `now()`. Operation goes through `validate_bank_write` like the existing `clear_observations_for_memory`.
- **Endpoint** `POST /v1/default/banks/{bank_id}/memories/{memory_id}/invalidate` with `InvalidateMemoryRequest` / `InvalidateMemoryResponse` Pydantic models. Returns 404 when the memory unit isn't found, 400 on a malformed `valid_to`, and propagates `OperationValidationError` + `@audited(...)` bookkeeping like sibling memory endpoints.
- **`get_memory_unit`** also returns `valid_to` so clients can see invalidation status without hitting `/history`.
- **Tests** in `tests/test_invalidate_memory.py` covering the engine method (default-now, not-found, malformed-uuid) and the dialect-level guarantee that semantic and BM25 arms include the validity filter.

Diff size: +438 / −4 across 6 files (4 modified + migration + test).

## Why

See #1391 for the full design rationale. Short version: when a fact in Hindsight becomes stale (a server is decommissioned, a person changes role, a default value changes), today's options are *delete* (audit trail gone) or *leave it* (recall returns it next to the corrected fact, and the LLM has to disambiguate at recall time). Other learning-oriented memory systems (e.g. MemPalace's `triples`/`attributes` tables) make `valid_from` / `valid_to` first-class. This PR adds the same to Hindsight in the smallest possible shape: one column + one index + one endpoint, no schema breaking changes.

## Validation

I've been running the equivalent of this patch against a managed Hindsight install (~1.4k memory units, ~720 entities) for several days. End to end:

- Insert a memory → recall returns it.
- `POST /memories/{id}/invalidate` → recall stops returning it.
- `GET /memories/{id}/history` → still shows it.
- recall over a 91-result window after invalidation → invalidated row is filtered out; ordering of the rest is unchanged.

The local version did the work via text-replace on the installed `retrieval.py` and out-of-band DDL; this PR does it cleanly through a real alembic migration and the SQL builder.

## Test plan

- [x] `pytest hindsight-api-slim/tests/test_invalidate_memory.py`
- [ ] Existing recall integration tests should be unaffected — please confirm CI passes.
- [ ] Optional: integration test asserting that an invalidated row drops out of recall on a real Postgres (happy to add — wanted to keep this PR small).

## Notes for reviewers

- **Naming**: `valid_to` matches the convention in MemPalace's schema and pairs naturally with eventual `valid_from` if we want to make that explicit (today it's implicit in `mentioned_at` / `event_date`). Open to `superseded_at` / `retired_at` if you have a strong preference.
- **HTTP verb**: `POST .../invalidate` — non-idempotent w.r.t. metadata (`invalidation_reason` overwrites). `PATCH /memories/{id}` with a partial body would also be reasonable; happy to switch.
- **Worker-side awareness**: this PR does *not* yet teach consolidation / reflect workers to skip `valid_to IS NOT NULL` rows. If you'd like that in the same PR, let me know — straightforward but increases scope.
- **`as_of` recall**: a natural follow-up — recall with `as_of: <ISO>` to "see what was active at that point". Out of scope here.

### DIFF
--- a/hindsight-api-slim/hindsight_api/alembic/versions/a2b3c4d5e6f7_add_valid_to_to_memory_units.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/a2b3c4d5e6f7_add_valid_to_to_memory_units.py
@@ -1,0 +1,64 @@
+"""Add valid_to validity-window column to memory_units table.
+
+Revision ID: a2b3c4d5e6f7
+Revises: z1u2v3w4x5y6
+Create Date: 2026-05-02
+
+Adds a nullable ``valid_to TIMESTAMPTZ`` column on ``memory_units`` so that
+superseded facts can be soft-retired without losing their timeline. Also
+adds a partial index on the bank/fact_type prefix limited to currently
+active rows so recall keeps using a small index even as historical data
+accumulates.
+
+Recall queries filter out rows where ``valid_to <= now()`` so invalidated
+memories no longer surface in semantic / BM25 / graph-spreading search,
+while ``GET /memories/{id}`` and ``GET /memories/{id}/history`` still return
+them — preserving the audit trail.
+
+See issue #1391 for the full design rationale.
+"""
+
+from collections.abc import Sequence
+
+from alembic import context, op
+
+from hindsight_api.alembic._dialect import run_for_dialect
+
+revision: str = "a2b3c4d5e6f7"
+down_revision: str | Sequence[str] | None = "z1u2v3w4x5y6"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _get_schema_prefix() -> str:
+    """Get schema prefix for table names (required for multi-tenant support)."""
+    schema = context.config.get_main_option("target_schema")
+    return f'"{schema}".' if schema else ""
+
+
+def _pg_upgrade() -> None:
+    schema = _get_schema_prefix()
+    op.execute(f"ALTER TABLE {schema}memory_units ADD COLUMN IF NOT EXISTS valid_to TIMESTAMPTZ NULL")
+    op.execute(
+        f"COMMENT ON COLUMN {schema}memory_units.valid_to IS "
+        "'NULL = still valid; non-NULL = superseded at this timestamp. "
+        "Recall filters out memories with valid_to <= now().'"
+    )
+    op.execute(
+        f"CREATE INDEX IF NOT EXISTS idx_memory_units_active "
+        f"ON {schema}memory_units (bank_id, fact_type) WHERE valid_to IS NULL"
+    )
+
+
+def _pg_downgrade() -> None:
+    schema = _get_schema_prefix()
+    op.execute(f"DROP INDEX IF EXISTS {schema}idx_memory_units_active")
+    op.execute(f"ALTER TABLE {schema}memory_units DROP COLUMN IF EXISTS valid_to")
+
+
+def upgrade() -> None:
+    run_for_dialect(pg=_pg_upgrade)
+
+
+def downgrade() -> None:
+    run_for_dialect(pg=_pg_downgrade)
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1448,6 +1448,42 @@ class ClearMemoryObservationsResponse(BaseModel):
     deleted_count: int
 
 
+class InvalidateMemoryRequest(BaseModel):
+    """Request model for marking a memory unit as superseded."""
+
+    model_config = ConfigDict(
+        json_schema_extra={
+            "example": {
+                "valid_to": "2026-05-02T17:30:00Z",
+                "reason": "Server srv-04 was decommissioned",
+            }
+        }
+    )
+
+    valid_to: str | None = None  # ISO-8601; defaults to now() if omitted
+    reason: str | None = None
+
+
+class InvalidateMemoryResponse(BaseModel):
+    """Response model for invalidating a memory unit."""
+
+    model_config = ConfigDict(
+        json_schema_extra={
+            "example": {
+                "id": "0c14e4f1-9eb6-4dde-b0a4-c2e8b3a3e5f1",
+                "valid_to": "2026-05-02T17:30:00.123456+00:00",
+                "fact_type": "world",
+                "preview": "Server srv-04 runs PostgreSQL 17 on Debian 12...",
+            }
+        }
+    )
+
+    id: str
+    valid_to: str | None
+    fact_type: str
+    preview: str
+
+
 class RecoverConsolidationResponse(BaseModel):
     """Response model for recovering failed consolidation."""
 
@@ -5176,6 +5212,72 @@ def _register_routes(app: FastAPI):
 
             error_detail = f"{str(e)}\n\nTraceback:\n{traceback.format_exc()}"
             logger.error(f"Error in POST /v1/default/banks/{bank_id}/consolidation/recover: {error_detail}")
+            raise HTTPException(status_code=500, detail=str(e))
+
+    @app.post(
+        "/v1/default/banks/{bank_id}/memories/{memory_id}/invalidate",
+        response_model=InvalidateMemoryResponse,
+        summary="Invalidate a memory unit",
+        description=(
+            "Mark a memory unit as invalidated as of a timestamp. The row stays in the timeline "
+            "(reachable via GET /memories/{id} and GET /memories/{id}/history) but recall queries "
+            "no longer return it once valid_to <= now(). Use when a fact has been superseded "
+            "(server decommissioned, person changed roles, default value changed) — not for "
+            "deletion of accidentally retained data."
+        ),
+        operation_id="invalidate_memory",
+        tags=["Memory"],
+    )
+    @audited("invalidate_memory", request_param="payload")
+    async def api_invalidate_memory(
+        bank_id: str,
+        memory_id: str,
+        payload: InvalidateMemoryRequest | None = None,
+        request_context: RequestContext = Depends(get_request_context),
+    ):
+        """Mark a memory unit as superseded as of a timestamp (default: now()).
+
+        The memory is *not* deleted; recall filters it out, but the audit trail
+        remains accessible via the regular get / history endpoints.
+        """
+        from datetime import datetime as _datetime
+
+        valid_to_dt: _datetime | None = None
+        reason: str | None = None
+        if payload is not None:
+            reason = payload.reason
+            if payload.valid_to:
+                try:
+                    valid_to_dt = _datetime.fromisoformat(payload.valid_to.replace("Z", "+00:00"))
+                except ValueError as ve:
+                    raise HTTPException(
+                        status_code=400,
+                        detail=f"valid_to is not a valid ISO-8601 timestamp: {payload.valid_to!r}",
+                    ) from ve
+        try:
+            result = await app.state.memory.invalidate_memory_unit(
+                bank_id=bank_id,
+                memory_id=memory_id,
+                valid_to=valid_to_dt,
+                reason=reason,
+                request_context=request_context,
+            )
+            if result is None:
+                raise HTTPException(status_code=404, detail=f"Memory unit '{memory_id}' not found")
+            return InvalidateMemoryResponse(**result)
+        except OperationValidationError as e:
+            raise HTTPException(status_code=e.status_code, detail=e.reason)
+        except ValueError as e:
+            raise HTTPException(status_code=400, detail=str(e))
+        except (AuthenticationError, HTTPException):
+            raise
+        except Exception as e:
+            import traceback
+
+            error_detail = f"{str(e)}\n\nTraceback:\n{traceback.format_exc()}"
+            logger.error(
+                f"Error in POST /v1/default/banks/{bank_id}/memories/{memory_id}/invalidate: {error_detail}"
+            )
             raise HTTPException(status_code=500, detail=str(e))
 
     @app.delete(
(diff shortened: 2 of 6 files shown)


yonefive71:
Second production datapoint in support of this PR.

Running Hindsight v0.13 inside a Hermes Agent install. Posting current SQL counts from the live bank (not estimates):

```
bank `hermes`
  documents:     163
  memory_units:  17,569
  chunks:        3,001

duplicate MU analysis (GROUP BY text):
  exact-duplicate clusters (cnt > 1):  455
  MU rows inside those clusters:       1,280
  top cluster size:                    22 byte-identical copies

fact_type distribution:
  experience:   7,775
  observation:  4,937
  world:        4,857
```

The top-10 clusters by size are all the same handful of Hindsight self-canon sentences ("Hindsight is the working-memory layer between llm-wiki and MemPalace…", "Hindsight explicitly excludes Minerva brand/product canon…", the nightly-dedupe-script summary) getting re-extracted on essentially every session that touches the topic. ~5% of the bank is byte-identical repeats, before we even talk about paraphrase clusters.

**Why I think soft-invalidate is the right shape (not hard DELETE).** I tried writing a doc-level paraphrase-cluster pruner last week. Dry-run found two clusters at Jaccard ≥0.65: one was 10 distinct sessions from a deck-authoring arc linked only by shared vocabulary (false positive — deleting would erase 4 days of session continuity), the other was 3 auto-curator pings (legitimate). Hard DELETE at doc level can't tell those apart without manual review of every cluster. Per-MU soft-invalidate cleanly resolves both cases: the duplicate *facts* die, the session transcripts that contain them stay alive, audit trail intact via `/history`.

This also matches how I already use MemPalace's `valid_from`/`valid_to` for the durable-canon layer — same mental model across both stores would be a real win.

**Adoption.** If this merges I'd run it immediately:

1. First sweep is trivial: `GROUP BY text HAVING count(*) > 1` against the live bank yields 455 clusters / 1,280 rows. Keep oldest (or highest `proof_count`) per cluster, invalidate the rest with `reason="exact-text-duplicate:<text-hash>"`. ~825 MUs reclaimed with zero ambiguity.
2. Wire `POST /memories/{id}/invalidate` into the existing nightly dedupe script as a new Tier F (MU-level paraphrase cluster collapse via cosine ≥0.92).
3. Promote any cluster that hits ≥5 paraphrase variants to a single canonical MU and invalidate the rest with `reason="paraphrase-cluster-collapse:<cluster-hash>"` for audit grep-ability.
4. Drop the doc-level Tier E shingled-Jaccard pass that currently can't act safely.

**On the closed PRs (#794 / #796) and #1228.** I read the thread. Understand the security gate on `validate_bank_write` — the approach in this PR (going through the engine method that *does* validate, via a soft-update rather than DELETE) sidesteps the auth-bypass concern entirely while still giving callers the surface they need. Treating invalidation as a write-operation on existing data is also more honest about what's happening than DELETE was.

**One small ask, no objection if you'd rather defer.** The submitter mentions worker-side awareness (consolidation/reflect skipping `valid_to IS NOT NULL` rows) as "happy to add if you want in this PR." For the paraphrase-cluster-collapse use case where I want the *winning* MU to be the source of truth going forward, having reflect not re-extract from invalidated rows matters. If it's not scope-creep, I'd vote to include; if it is, a follow-up issue is fine.

Happy to provide more diagnostic data (anonymized recall blocks, before/after token counts, the exact dup-cluster query plan) if useful. I'm planning to run this PR's branch locally against the production bank while we wait on merge — happy to report back if anything surfaces.

Thanks for the project, and to @nikolay-bratanov for the well-scoped PR.

yonefive71:
Quick heads-up from running the branch locally against current `main` — there's a revision id collision that'll block merge:

```
$ ls hindsight_api/alembic/versions/ | grep a2b3c4d5e6f7
a2b3c4d5e6f7_add_text_signals_column.py     ← landed on main since this PR opened
a2b3c4d5e6f7_add_valid_to_to_memory_units.py ← this PR
```

```
$ pytest tests/test_alembic_dag.py::test_single_head
UserWarning: Revision a2b3c4d5e6f7 is present more than once
FAILED — Alembic has 3 heads (['a2b3c4d5e6f7', 'aa2b3c4d5e6f', 'o1a2b3c4d5e6'])
```

Two separate things going on:

1. **Hard collision (your PR vs. main):** the `a2b3c4d5e6f7` revision id is reused. Renumbering this PR's migration to a fresh id and pointing `down_revision` at whichever head it should chain off of (likely the new tip on main) should clear it.
2. **Three heads (normal merge-time churn):** unrelated to this PR — there are also `aa2b3c4d5e6f` and `o1a2b3c4d5e6` open. Maintainer can `alembic merge heads` at merge.

Everything else looked good locally:

- The 5 tests in `test_invalidate_memory.py` pass after fixing one tiny import typo (`PostgresDialect` → `PostgreSQLDialect`); see diff at the bottom of this comment.
- Full `test_db_abstraction.py` + `test_invalidate_memory.py` runs 85/85 green.
- Migration SQL is clean (`ADD COLUMN IF NOT EXISTS`, partial index on `WHERE valid_to IS NULL`, downgrade is symmetric).

Tiny test-file fix I had to apply locally so the suite runs:

```diff
- from hindsight_api.engine.sql.postgresql import PostgresDialect
+ from hindsight_api.engine.sql.postgresql import PostgreSQLDialect as PostgresDialect
```

The class is named `PostgreSQLDialect` in the rest of the codebase; both `test_db_abstraction.py` and the engine internals use that spelling.

Happy to open a small follow-up PR against your branch with the rename + revision-id bump if it'd save you a round-trip. Otherwise no rush — just wanted the collision on the record so it doesn't bite at merge.

nicoloboschi:
please reply to these questions before working on the pr - closing for now https://github.com/vectorize-io/hindsight/issues/1391#issuecomment-4369898653

PR #1399, opened 2026-05-03
codex: add configurable recall timeout for UserPromptSubmit hooks

voarsh2:
## Summary

Add a configurable recall timeout to the Codex Hindsight integration instead of hard-coding `10` seconds in `recall.py`.

This makes slower but valid recall environments work without patching the installed integration locally.

## What changed

- add `recallTimeout` to Codex integration defaults
- add `HINDSIGHT_RECALL_TIMEOUT` environment variable override
- update `recall.py` to use `config.get("recallTimeout", 10)`
- document the new setting in the Codex integration README
- add a focused test asserting the configured timeout is passed through to the recall request

## Why

During real Windows Codex hook testing, the Hindsight `UserPromptSubmit` hook was firing correctly, but recall failed end-to-end because the integration always used a fixed `10` second timeout.

That was too strict for this deployment: replaying the real hook payload showed the recall path was otherwise healthy, and simply allowing a larger timeout made the hook succeed and inject context properly.

This is not Windows-specific in principle. Any slower valid recall environment can hit the same issue:
- remote/self-hosted Hindsight deployments
- larger banks
- higher recall budgets
- transiently slower infrastructure

## Scope

This PR is intentionally narrow:
- it only adds a configurable timeout knob for Codex recall
- it does not change broader hook behavior
- it does not attempt to address unrelated existing test issues outside this slice

## Validation

- local installed Codex + Hindsight integration was tested end-to-end with the timeout change
- the real `UserPromptSubmit` hook payload succeeded once the timeout was configurable
- added a focused unit test for timeout propagation


nicoloboschi (approved):
LGTM

PR #1410, opened 2026-05-04
fix(openclaw): skip retain for operational session patterns

### LINKED ISSUE #1403: fix(openclaw): autonomic sessions are retained and recalled instead of being skipped
### Bug Description

OpenClaw creates `autonomic` sessions (session keys containing `:autonomic:`) for automated maintenance and health-check tasks. The `getIdentitySkipReason()` function in the openclaw plugin is designed to skip Hindsight operations for operational session types, but the regex pattern only matches `cron`, `heartbeat`, and `subagent`. The `autonomic` session type is missing from the pattern, causing these automated sessions to be treated as normal user conversations -- they fire recall queries, retain transcripts, and consume bank storage.

### Root Cause
The regex in `getIdentitySkipReason()` at approximately line 953 of `src/index.ts`:
```
const operationalMatch = sessionKey.match(/^agent:[^:]+:(cron|heartbeat|subagent):/);
```
The alternation group does not include `autonomic`.

### Steps to Reproduce

1. Configure an OpenClaw agent with hindsight-openclaw plugin and autonomic sessions enabled
2. Trigger an autonomic session (session key will contain `:autonomic:`)
3. Observe the gateway logs -- the autonomic session passes through `getIdentitySkipReason()` without matching
4. Check `before_prompt_build` logs -- recall query fires for the autonomic session
5. Check `agent_end` logs -- retain fires and the autonomic transcript is written to the memory bank
6. Query the bank later -- autonomic session transcripts appear in recall results mixed with genuine conversation

### Expected Behavior

Autonomic sessions should be skipped by the operational session filter, the same way `cron`, `heartbeat`, and `subagent` sessions are skipped. No recall queries should fire, no transcripts should be retained, and no bank storage should be consumed.

### Suggested Fix
Add `autonomic` to the alternation group:
```
const operationalMatch = sessionKey.match(/^agent:[^:]+:(cron|heartbeat|subagent|autonomic):/);
```
One-word addition. No config changes, no schema changes, no behavioral changes for existing session types.

### Actual Behavior

Autonomic sessions pass through the operational session filter and are processed identically to user conversations. Recall queries fire (wasting API calls), transcripts are retained (polluting the bank with automated noise), and the noise appears in future recall results.

### Version

hindsight-openclaw: 0.7.2 - OpenClaw: 2026.4.27

### LLM Provider

OpenAI

vernmic:
Heartbeat, cron, subagent, and autonomic sessions are machine-driven and produce low-value or duplicative retains. Add a defense-in-depth gate that checks sessionKey patterns before any retention logic runs.

The :autonomic: pattern was missing from the upstream skip regex, causing autonomic sessions to flood the memory bank with procedural noise.

Closes: #1403

### DIFF
--- a/hindsight-integrations/openclaw/src/index.ts
+++ b/hindsight-integrations/openclaw/src/index.ts
@@ -1488,6 +1488,14 @@ export function getPluginConfig(api: MoltbotPluginAPI): PluginConfig {
     statelessSessionPatterns: Array.isArray(config.statelessSessionPatterns)
       ? config.statelessSessionPatterns
       : [],
+    skipRetainSessionPatterns: (() => {
+      const raw = Array.isArray(config.skipRetainSessionPatterns)
+        ? config.skipRetainSessionPatterns
+        : ["heartbeat", "cron", "subagent"];
+      // Convert bare tokens to glob patterns: "heartbeat" -> "**:heartbeat**"
+      // Pass through patterns that already contain glob wildcards
+      return raw.map((p: string) => (p.includes("*") ? p : `**:${p}**`));
+    })(),
     skipStatelessSessions: config.skipStatelessSessions !== false,
     debug: config.debug ?? false,
     debugPerfTiming: config.debugPerfTiming === true,
@@ -2275,6 +2283,18 @@ ${memoriesFormatted}
           ) {
             debug(
               `[Hindsight] Skipping retain: session '${agentEndSessionKey}' matches statelessSessionPatterns`
+            );
+            return;
+          }
+          const skipRetainPatterns = compileSessionPatterns(
+            pluginConfig.skipRetainSessionPatterns ?? []
+          );
+          if (
+            skipRetainPatterns.length > 0 &&
+            matchesSessionPattern(agentEndSessionKey, skipRetainPatterns)
+          ) {
+            debug(
+              `[Hindsight] Skipping retain - operational session pattern matched: ${agentEndSessionKey}`
             );
             return;
           }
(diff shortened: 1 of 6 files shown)


hindsight-integrations/openclaw/src/index.ts
```diff
@@ -2207,6 +2207,22 @@ ${memoriesFormatted}
           return;
         }
 
+        // Defense-in-depth: skip retain for operational session patterns
+        // These sessions are machine-driven and produce low-value or duplicative retains
+        if (
+          agentEndSessionKey && (
+            agentEndSessionKey.includes(":heartbeat") ||
```
nicoloboschi: good idea, can we make this list configurable via plugin config with these ones as default?

vernmic: @nicoloboschi implemented https://github.com/vectorize-io/hindsight/pull/1410/commits/6710f72fb174e4a0e2afba3ae6ba1c93dbebd96a

hindsight-integrations/openclaw/src/index.ts
```diff
@@ -1488,6 +1488,14 @@ export function getPluginConfig(api: MoltbotPluginAPI): PluginConfig {
     statelessSessionPatterns: Array.isArray(config.statelessSessionPatterns)
       ? config.statelessSessionPatterns
       : [],
+    skipRetainSessionPatterns: (() => {
```
nicoloboschi, line 1491: we can't se this as default - default must be empty array (disabled

nicoloboschi (changes requested):

nicoloboschi:
closing as stale

PR #1417, opened 2026-05-04
fix(api): recognize Pydantic aliases in unknown param middleware

Tosko4:
## Summary
- Treat Pydantic field aliases as known JSON body fields in the unknown-parameter middleware.
- Fixes valid retain payloads using the public `async` field being reported as ignored because the internal model field is named `async_`.
- Adds a regression test for aliased body fields so `X-Ignored-Params` is not set for valid aliases.

## Why
The retain endpoint documents and accepts `async`, and generated clients serialize `RetainRequest.async_` with the `async` alias. The middleware only checked `model_fields.keys()`, so it saw `async` as unknown even though Pydantic accepts it.

## Test Plan
- `uv run --directory hindsight-api-slim pytest tests/test_unknown_params.py -q`
- `uv run --directory hindsight-api-slim ruff check hindsight_api/api/http.py tests/test_unknown_params.py`
- `uv run --directory hindsight-api-slim ruff format --check hindsight_api/api/http.py tests/test_unknown_params.py`
- `uv run --directory hindsight-api-slim ty check hindsight_api`
- `./scripts/hooks/lint.sh`

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2784,6 +2784,13 @@ def create_app(
                                         ann = param.annotation
                                         if isinstance(ann, type) and issubclass(ann, BaseModel):
                                             known_fields = set(ann.model_fields.keys())
+                                            for field in ann.model_fields.values():
+                                                # Pydantic models can expose public JSON names via aliases
+                                                # (for example RetainRequest.async_ is sent as "async").
+                                                # Treat aliases as known fields so valid client payloads are
+                                                # not reported as ignored parameters.
+                                                if isinstance(field.alias, str):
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
the change LGTM but there are other unrlated commits

Tosko4:
You are right, sorry about that. I had included generated artifacts after chasing the generated-files check, but they were unrelated to this fix. I force-pushed a cleaned-up branch rebased on latest main; the PR now only changes the middleware and the regression test.

Tosko4:
@nicoloboschi I cleaned this up after your review. The PR is now rebased on current main and only contains the middleware change plus the regression test. Could you take another look when you have a chance?

nicoloboschi (approved):
LGTM

PR #1439, opened 2026-05-04
feat(openclaw): prepend session-context block to retained transcripts

kryptt:
## Problem

When `dynamicBankGranularity` doesn't include `"user"`, every speaker in an agent's bank ends up indistinguishable to vector recall. Two different users (say _John_ and _Peter_) chatting with the same agent in the same channel produce transcripts that are byte-for-byte structurally identical:

```
[role: user]
What's the status of MIN-123?
[user:end]
…
```

`stripMetadataEnvelopes` removes OpenClaw's inbound `[meta]` blocks before the transcript reaches `client.retain(...)`, so the stored content has no signal that lets similarity search separate one speaker's memories from another's.

The escape hatch today is `dynamicBankGranularity: ["agent", "user"]`, but that's not always the right tradeoff — for sprint-ops / shared-team bots, fragmenting the bank per user forfeits cross-user shared context, and you still want a single agent-level memory pool.

## Change

* Add a `RetentionSessionContext` shape (`senderId`, `channelId`, `provider`) and a `formatRetentionSessionContext(ctx)` helper that returns either a `[context] … [/context]` block or `null` when nothing useful is set.
* Extend `prepareRetentionTranscript(messages, pluginConfig, retainFullWindow, sessionContext?)` so it can prepend that block:
  * **JSON / Anthropic-structured formats** → first message is `{ role: "system", content: "[context]\nsender: …\nchannel: …\nprovider: …\n[/context]" }`.
  * **Text format** → the same block is prepended literally before `[role: user]…`.
* Pull `senderId` from the already-resolved retain ctx (`resolvedCtxForRetain`) at the call site so the value matches what `deriveBankId` uses, including the per-session sender cache populated during `before_prompt_build`.
* New `includeSenderContext?: boolean` config flag (default `true`) so operators can opt out.

The header is intentionally tiny and provider-agnostic — just enough signal for vector recall to disambiguate without forcing a schema change on Hindsight or the LLM consumer.

## Why this design

* **Backward-compatible**: `sessionContext` is optional; old callers and the existing `text`/`json` shapes still work. Without context, no header is emitted.
* **Format-aware**: in JSON mode the header is a real system message (so structured consumers see it), in text mode it's a `[context]` block matching the rest of the transcript style.
* **No bank-shape impact**: nothing about `deriveBankId` changes — operators keep whatever granularity they chose.
* **Filtered out cleanly**: `formatRetentionSessionContext` returns `null` for `undefined`/empty fields, so e.g. cron and identity-skipped sessions don't get spurious `sender: anonymous` headers.

## Tests

`prepareRetentionTranscript` got 5 new cases:

* JSON format prepends a system-role context message
* Text format prepends a literal `[context]` block
* `includeSenderContext: false` suppresses the header
* `sessionContext: {}` (no usable fields) — nothing prepended
* Missing `sessionContext` argument — unchanged from current behavior

All 21 existing `prepareRetentionTranscript` tests still pass; full unit suite is green except for one pre-existing `backfill.test.ts` symlink case that fails on `main` itself in this sandbox (unrelated to this change).

## Operator example

```ts
plugins.entries["hindsight-openclaw"].config = {
  // single bank for the agent — old behavior
  dynamicBankGranularity: ["agent"],
  // new: speaker info still flows into the transcript so recall can rank by user
  includeSenderContext: true, // default
};
```

After this, a memory retained from John ends up tagged as

```
{
  "role": "system",
  "content": "[context]\nsender: U7JAF258R\nchannel: C04L6E0H3SQ\nprovider: slack\n[/context]"
}
```

and similarity search can naturally weight retrieval by the active speaker without changing the bank scheme.

### DIFF
--- a/hindsight-integrations/openclaw/src/index.test.ts
+++ b/hindsight-integrations/openclaw/src/index.test.ts
@@ -799,6 +799,68 @@ describe("prepareRetentionTranscript", () => {
     // The middle message becomes empty after tag stripping, so messageCount should be 2
     expect(result?.messageCount).toBe(2);
   });
+
+  it("prepends a session-context system message when sessionContext is provided (json)", () => {
+    const config: PluginConfig = { ...baseConfig, retainToolCalls: false };
+    const messages = [{ role: "user", content: "What's MIN-123 status?" }];
+    const result = prepareRetentionTranscript(messages, config, false, {
+      senderId: "U7JAF258R",
+      channelId: "C04L6E0H3SQ",
+      provider: "slack",
+    });
+    expect(result).not.toBeNull();
+    const parsed = JSON.parse(result!.transcript);
+    expect(parsed[0]).toEqual({
+      role: "system",
+      content:
+        "[context]\nsender: U7JAF258R\nchannel: C04L6E0H3SQ\nprovider: slack\n[/context]",
+    });
+    expect(parsed[1]).toEqual({ role: "user", content: "What's MIN-123 status?" });
+    expect(result?.messageCount).toBe(2);
+  });
+
+  it("prepends a session-context block when sessionContext is provided (text format)", () => {
+    const config: PluginConfig = { ...baseConfig, retainFormat: "text" };
+    const messages = [{ role: "user", content: "ping" }];
+    const result = prepareRetentionTranscript(messages, config, false, {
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
LGTM

PR #1440, opened 2026-05-05
fix(worker): scope per-tenant fanout to schemas with pending work

### LINKED ISSUE #1408: Worker poller _scan_active_schemas calls non-existent function, causing noisy PG logs and breaking single-tenant
## Bug Description

The worker poller tries to call a `schemas_with_pending_work()` PL/pgSQL function on every poll cycle (every ~30s). When the function doesn't exist — which is the default for a fresh install — PostgreSQL logs a server-side error: `ERROR: function schemas_with_pending_work() does not exist`. The Python code silently catches the exception, but the Postgres log is still written.

This forces self-hosted users to create the function to suppress the noise. But the function as commonly written only checks `tenant_%` schemas, which means it returns empty for single-tenant deployments using the default (public) schema, and the worker poller never claims any jobs.

## Steps to Reproduce

1. Deploy Hindsight from scratch (no custom PL/pgSQL functions)
2. The worker poller starts, calls `SELECT * FROM schemas_with_pending_work()` every 30s
3. Postgres server log gets spammed: `ERROR: function schemas_with_pending_work() does not exist`
4. As a workaround, create the function with `tenant_%` schema check (common approach)
5. Function now exists and succeeds, but finds nothing in single-tenant mode
6. Worker poller stops claiming all jobs — they remain pending forever

## Expected Behavior

Fresh deployments should not produce noisy PostgreSQL logs on every poll cycle, and single-tenant deployments should work without needing custom database functions.

## Actual Behavior

- Without the function: noisy Postgres logs every 30s
- With a naive function: jobs never get claimed in single-tenant mode

## Root Cause

In `hindsight_api/worker/poller.py`, `_scan_active_schemas()` (line ~197-200):

```python
try:
    rows = await conn.fetch("SELECT * FROM schemas_with_pending_work()")
    return {r[0] for r in rows}
except Exception:
    pass  # silent exception, Postgres still logged the error server-side
```

The function is called unconditionally. There's no check whether it exists first, and the `except Exception: pass` pattern means the server-side error log is invisible in the application logs but still fills the Postgres logs.

## Suggested Fix

Option A — Check function existence before calling:

```python
exists = await conn.fetchval(
    "SELECT EXISTS(SELECT 1 FROM pg_proc WHERE proname = 'schemas_with_pending_work')"
)
if exists:
    rows = await conn.fetch("SELECT * FROM schemas_with_pending_work()")
    return {r[0] for r in rows}
```

Option B — Install the function automatically via Helm hook or DB migration so it's always present on fresh installs. The function should also check the default (public) schema, not just `tenant_%`.

Option C — Replace the function call path with a direct SQL query that handles both single-tenant and multi-tenant:

```sql
SELECT nspname FROM pg_namespace WHERE nspname LIKE 'tenant_%'
  AND EXISTS(
    SELECT 1 FROM (SELECT pg_class_from_nspname(nspname)) t(...)
  )
UNION ALL
SELECT NULL WHERE EXISTS(
  SELECT 1 FROM async_operations WHERE status = 'pending' AND task_payload IS NOT NULL
)
```

## Version
0.5.x

cdbartholomew:
## Summary

With many tenant schemas, the worker poller's progress-stats logger fans out two `COUNT/GROUP BY` queries against `tenant_X.async_operations` per schema, every `PROGRESS_LOG_INTERVAL` seconds, per worker pod. With N tenants and W worker pods that's `2*N*W` queries every 30s. The fanout also bloats each pooled backend's catalog/plan cache because every backend is asked to touch every tenant's relations.

This change:

1. **Uses the existing `public.tenant_has_work` marker table when installed** (O(1) regardless of tenant count). Per-schema breakdown queries then only run against schemas that actually have pending rows.
2. **Falls back to the previous behavior** when the marker isn't installed (single-tenant or other self-hosted deployments). The default `None` schema is preserved in both paths so single-tenant correctness is unchanged.
3. **Closes #1408**: gates `schemas_with_pending_work()` on a cached `pg_proc` existence check. Calling the function unconditionally and swallowing the Python exception still produces a server-side `ERROR: function ... does not exist` log on every poll, which spams Postgres logs on fresh installs.

The detection result is cached for the worker's lifetime — installing the helpers later requires a worker restart (called out in the docstring).

## Behavior matrix

| Deployment | Path taken | Queries per stats cycle |
| --- | --- | --- |
| Multi-tenant + marker installed | Marker table | 1 + 2 × (active schemas only) |
| Multi-tenant, no marker | Full fanback (existing behavior) | 1 + 2 × N tenants |
| Single-tenant (no marker) | Default schema only | 1 + 2 × 1 |
| Single-tenant + marker (rare) | Marker (empty) + default schema | 1 + 2 × 1 |

## Test plan

- [x] `tests/test_worker.py::test_detect_marker_features_when_absent` — returns `(False, False)` and caches the result so subsequent calls don't re-query.
- [x] `tests/test_worker.py::test_detect_marker_features_when_present` — returns `(True, True)` after the table + function are installed.
- [x] `tests/test_worker.py::test_scan_skips_missing_function_call` — wraps the connection to assert `schemas_with_pending_work` is never sent to PG when the function isn't installed; result still correct via the per-schema EXISTS fallback.
- [x] `tests/test_worker.py::test_scan_finds_schemas_with_pending_work` (existing) still passes.
- [x] `tests/test_worker.py::test_claim_batch_only_queries_active_schemas` (existing) still passes.
- [x] Full `test_worker.py` suite passes (60 tests).
- [x] `ruff check` clean.

### DIFF
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -150,6 +150,58 @@ class WorkerPoller:
         # Rotation offset for per-tenant fair claiming. Advances past the last
         # schema we serviced so a busy tenant can't monopolize the poll order.
         self._next_schema_idx: int = 0
+        # Cached one-shot detection of optional Postgres helper objects. None
+        # means "not yet checked"; otherwise (has_tenant_has_work_table,
+        # has_schemas_with_pending_work_function).
+        self._marker_features: tuple[bool, bool] | None = None
+
+    async def _detect_marker_features(self, conn) -> tuple[bool, bool]:
+        """Check once whether optional pending-work helpers are installed.
+
+        Returns ``(has_table, has_function)`` where:
+          * ``has_table``  — ``public.tenant_has_work`` exists
+          * ``has_function`` — ``public.schemas_with_pending_work()`` exists
+
+        Both are installed by hindsight deployments that maintain a marker
+        table via row triggers on ``tenant_%.async_operations``. Self-hosted
+        single-tenant deployments typically have neither, and that's fine —
+        callers fall back to a direct query against the default schema.
+
+        Cached for the worker's lifetime: a feature can't appear without a
+        deployment-level action (DDL / Helm hook), and re-checking on every
+        poll would defeat the point of the caching. If your deployment installs
+        the helpers later, restart the worker.
+
+        Non-PostgreSQL backends always return ``(False, False)`` — the helpers
+        are PG-specific.
+        """
+        if self._marker_features is not None:
+            return self._marker_features
+        if self._backend.backend_type != "postgresql":
+            self._marker_features = (False, False)
+            return self._marker_features
+        try:
+            row = await conn.fetchrow(
+                """
+                SELECT
+                  EXISTS(
+                    SELECT 1 FROM pg_class c
+                    JOIN pg_namespace n ON n.oid = c.relnamespace
+                    WHERE n.nspname = 'public' AND c.relname = 'tenant_has_work'
+                  ) AS has_table,
+                  EXISTS(
+                    SELECT 1 FROM pg_proc p
+                    JOIN pg_namespace n ON n.oid = p.pronamespace
+                    WHERE n.nspname = 'public' AND p.proname = 'schemas_with_pending_work'
+                  ) AS has_fn
+                """
+            )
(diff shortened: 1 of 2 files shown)


nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/1509

PR #1441, opened 2026-05-05
     fix(worker): recover stuck tasks from dead workers with UI controls

Armanoide:
When a hindsight container crashes or restarts, tasks in `processing` state can become orphaned, blocking bank serialization locks. This adds recovery mechanisms with both automatic and manual controls.
### Backend
- **`recover_stuck_tasks()`** in `memory_engine.py` — resets tasks older than configurable threshold (default: 10min) to `pending`
- **`POST /v1/default/banks/{bank_id}/tasks/recover`** endpoint for manual recovery
- **Auto-recovery** in worker poller (opt-in via `HINDSIGHT_WORKER_AUTO_RECOVER_DEAD_TASKS=false`)
- **`PUT /v1/default/config/worker`** endpoint for runtime config updates (in-memory only)
- **`stuck_consolidation`** field in bank stats to detect stuck tasks
- **Config-driven thresholds**: `HINDSIGHT_WORKER_STUCK_TASK_THRESHOLD_S` (default: 600s) and `HINDSIGHT_WORKER_STUCK_TASK_CHECK_INTERVAL_S` (default: 60s)
- All SQL properly parameterized using `make_interval(secs => $N)` — no string interpolation
- Worker identity uses configured `worker_id` to avoid false positives in stats

### Frontend
- **Recover Stuck Tasks** button in bank page Actions dropdown
- **Stuck indicator** in ConsolidationCard when tasks are detected
- **Tooltips** clarifying the difference between "Recover Consolidation" and "Recover Stuck Tasks"

### Tests
5 new tests in `test_worker.py`:
- `test_recover_dead_worker_tasks_resets_old_processing`
- `test_recover_dead_worker_tasks_skips_recent`
- `test_recover_dead_worker_tasks_skips_self`
- `test_recover_stuck_tasks_memory_engine`
- `test_recover_stuck_tasks_skips_recent`

## Configuration
New environment variables (all optional):
```env
# Auto-recover orphaned tasks from dead workers (default: false)
HINDSIGHT_WORKER_AUTO_RECOVER_DEAD_TASKS=false

# Threshold for detecting stuck tasks in seconds (default: 600 = 10 minutes)
HINDSIGHT_WORKER_STUCK_TASK_THRESHOLD_S=600

# How often to check for stuck tasks in seconds (default: 60)
HINDSIGHT_WORKER_STUCK_TASK_CHECK_INTERVAL_S=60
```
Can be toggled at runtime via PUT /v1/default/config/worker (in-memory only, does not persist across restarts or propagate to other processes).

Security
- All SQL parameters are properly parameterized using make_interval(secs => $N) — no string interpolation
- Worker identity uses configured worker_id to avoid false positives
- Manual recovery endpoint requires authentication and bank write permissions

## View

alert
Capture d’écran 2026-05-05 à 03 18 35
trigger button
Capture d’écran 2026-05-05 à 03 18 58

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,6 +30,14 @@ HINDSIGHT_API_LLM_BASE_URL=https://api.openai.com/v1
 # HINDSIGHT_API_LLM_API_KEY=your-deepseek-api-key
 # HINDSIGHT_API_LLM_MODEL=deepseek-v4-flash  # or deepseek-v4-pro / deepseek-chat / deepseek-reasoner
 
+# Worker Configuration (Optional)
+# Automatically recover tasks stuck in 'processing' state from dead workers (container crash/restart).
+# When enabled, the worker checks every 60s for tasks older than 5 minutes and resets them to 'pending'.
+# Default: false (disabled). Enable if you experience stuck consolidations after container restarts.
+# HINDSIGHT_WORKER_AUTO_RECOVER_DEAD_TASKS=false
+# HINDSIGHT_WORKER_STUCK_TASK_THRESHOLD_S=600  # 10 minutes default
+# HINDSIGHT_WORKER_STUCK_TASK_CHECK_INTERVAL_S=60  # Check every 60 seconds
+
 # Example: LM Studio local configuration (Qwen 2.5 32B recommended)
 # HINDSIGHT_API_LLM_PROVIDER=lmstudio
 # HINDSIGHT_API_LLM_API_KEY=lmstudio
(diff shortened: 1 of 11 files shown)


nicoloboschi (changes requested):
hey, this fix will not work reliably.

as I said many times now, this could easily lead to workers doing the same work in parallel and we'd need an coordinator system to maange distributed workers correctly.
intentionally we don't cross this line for keeping the administration simple

Issue #1412, opened 2026-05-04
`retain_extract_facts+structured` retries indefinitely on validation failure — no `max_attempts` cap, no unstructured fallback


bchumerch:
## Summary

When the structured-output extraction stage in `retain_extract_facts` returns LLM responses that fail JSON-schema validation, the worker retries the LLM call without an upper bound. A single op can monopolize a worker slot indefinitely, burning quota and blocking all other retain ops in the queue.

This is platform-agnostic — same code path on Mac and Windows.

## Repro

1. Use `claude-code` provider (Max plan OAuth) on `claude-sonnet-4-6`.
2. Trigger a retain on a transcript that is heavily self-referential about JSON, structured outputs, regex escaping, and Hindsight itself (e.g., a long Claude Code session debugging Hindsight). Today's incident: a Claude Code session debugging Hindsight on Hindsight produced a transcript where the model's structured-output extraction kept returning JSON that failed validation, presumably because the transcript content kept tripping the schema.
3. Worker enters a tight retry loop. `[STUCK_STACK]` warnings fire at 5-min and 10-min thresholds, but no abort.

## Observed today (single op, ~14 min, ~30 LLM calls)

```
02:21:02  slow llm call: scope=retain_extract_facts time=29.889s
02:21:17  slow llm call: scope=retain_extract_facts time=30.474s
02:21:38  slow llm call: scope=retain_extract_facts time=42.383s
02:21:49  slow llm call: scope=retain_extract_facts time=46.692s
02:22:07  slow llm call: scope=retain_extract_facts time=50.463s
02:22:40  slow llm call: scope=retain_extract_facts time=50.630s
02:22:47  slow llm call: scope=retain_extract_facts time=68.725s
02:22:50  slow llm call: scope=retain_extract_facts time=42.481s
02:23:01  slow llm call: scope=retain_extract_facts time=21.464s
02:23:26  slow llm call: scope=retain_extract_facts time=35.978s
02:23:29  slow llm call: scope=retain_extract_facts time=42.034s
02:23:35  slow llm call: scope=retain_extract_facts time=33.425s
02:24:01  slow llm call: scope=retain_extract_facts time=34.826s
02:24:03  slow llm call: scope=retain_extract_facts time=33.941s
02:24:03  slow llm call: scope=retain_extract_facts time=28.366s
02:24:38  slow llm call: scope=retain_extract_facts time=37.411s
02:24:40  slow llm call: scope=retain_extract_facts time=37.782s
[...]

PR #1452, opened 2026-05-05
fix(retain): prevent runaway structured extraction retries

cooleryu:
## Summary

Prevents retain fact extraction from multiplying structured-output retry attempts across nested retry layers.

The retain path now owns the structured extraction attempt budget directly, while provider-level retries are disabled for each retain extraction call. The per-chunk retry wrapper also uses the configured retain/global LLM budget instead of a hardcoded 3 attempts.

## Why

In issue #1412, operators observed workers stuck in `retain_extract_facts+structured` even after their configured retry budget appeared exhausted. The retain pipeline had more than one retry layer:

- provider-level structured JSON retry
- retain fact validation retry
- chunk-level retry with a hardcoded 3 attempts

That made the effective number of LLM calls higher than the configured `HINDSIGHT_API_RETAIN_LLM_MAX_RETRIES` suggested, especially for multi-chunk inputs.

## Changes

- Adds a small `_retain_llm_attempts()` helper for retain retry budget resolution.
- Uses `retain_llm_max_retries` when set, otherwise falls back to `llm_max_retries`.
- Treats `0` as "no retries, but still make one attempt".
- Sets provider-level `max_retries=0` inside retain fact extraction so the retain loop owns structured validation retries.
- Replaces the hardcoded chunk retry count with the same configured retain attempt budget.
- Expands regression coverage for explicit retain budget, global fallback, zero-budget semantics, multi-chunk bounding, and transient recovery.

## Test Plan

```bash
uv run --directory hindsight-api-slim pytest tests/test_fact_extraction_retry.py tests/test_consolidation_retry_budget.py -q
uv run --directory hindsight-api-slim ruff check hindsight_api/engine/retain/fact_extraction.py tests/test_fact_extraction_retry.py
git diff --check
```

All passed locally.

## Risk

This is scoped to retain fact extraction. It intentionally preserves chunk-level retry recovery, but makes the configured retry budget the source of truth so failure cases stop sooner instead of being multiplied by nested retry loops.

I left consolidation unchanged in this PR because that path already has a separate documented outer/inner retry model and tests. This PR focuses on the retain path surfaced by the stuck `retain_extract_facts` logs.

## Related Issue

Addresses the retain extraction retry amplification described in #1412.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -957,6 +957,12 @@ def _build_request_body(llm_config, config, prompt: str, user_message: str, resp
     return request_body
 
 
+def _retain_llm_attempts(config) -> int:
+    """Return the retain extraction retry budget as a bounded attempt count."""
+    configured = config.retain_llm_max_retries if config.retain_llm_max_retries is not None else config.llm_max_retries
+    return max(1, int(configured))
+
+
 async def _extract_facts_from_chunk(
     chunk: str,
     chunk_index: int,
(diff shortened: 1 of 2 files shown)


nicoloboschi:
thanks, fixed in #1516

PR #1453, opened 2026-05-05
fix(search): correct vchord BM25 score direction

liling:
## Summary
- Negate vchord's `<&>` BM25 distance operator so that higher scores mean higher relevance, matching pg_textsearch behavior

The `<&>` operator returns a distance metric (lower = more relevant), but the code used `DESC` ordering, causing the least relevant results to appear first.

## Test plan
- [ ] Configure `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=vchord`
- [ ] Run a recall query and verify relevant results appear first
- [ ] Switch to `pg_textsearch` and verify ordering remains correct


nicoloboschi (approved):
+1!

PR #1459, opened 2026-05-05
Add AlloyDB ScaNN vector index support

can1357:
## What changed

Adds `scann` as a supported `HINDSIGHT_API_VECTOR_EXTENSION` value for AlloyDB.

The vector-index branching was getting copied across migrations and runtime code, so this moves it into a shared helper first. ScaNN then plugs into the same paths as pgvector, pgvectorscale, and vchord.

A few details worth calling out:

- `scann` maps to the AlloyDB extension name `alloydb_scann`
- ScaNN indexes use `USING scann (embedding cosine) WITH (mode = 'AUTO')`
- bootstrap creates `vector` before `alloydb_scann`
- the per-bank index repair migration now picks up ScaNN automatically, so existing empty HNSW per-bank indexes get recreated as ScaNN when configured

## Checked

- `uv run ruff check ...`
- `uv run python -m py_compile ...`
- `uv run pytest hindsight-api-slim/tests/test_vector_index.py`
- `uv run pytest hindsight-api-slim/tests/test_migration_shape.py hindsight-api-slim/tests/test_migrations_thread_safety.py`

I also tried the full `hindsight-api-slim/tests` suite locally, but this checkout has no `.env` / `HINDSIGHT_API_LLM_API_KEY`, so the broader suite fails during test setup before reaching this change.

### DIFF
--- a/docker/docker-compose/alloydb/docker-compose.yaml
+++ b/docker/docker-compose/alloydb/docker-compose.yaml
@@ -1,0 +1,90 @@
+name: hindsight
+# Docker Compose file for Hindsight with AlloyDB Omni and ScaNN
+# Uses Google's free AlloyDB Omni container image: https://hub.docker.com/r/google/alloydbomni
+#
+# Usage:
+# docker compose -f docker/docker-compose/alloydb/docker-compose.yaml up -d
+#
+# Make sure to set the required environment variables before running:
+# - HINDSIGHT_DB_PASSWORD: password for the AlloyDB Omni/PostgreSQL user
+# - Configure LLM provider variables as needed (see the hindsight service below)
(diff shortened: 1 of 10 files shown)


nicoloboschi (commented):
this is great! can you add an docker compose example as we have for pgtextsearch and vchord using their free docker image? 
https://hub.docker.com/r/google/alloydbomni

can1357:
> this is great! can you add an docker compose example as we have for pgtextsearch and vchord using their free docker image? https://hub.docker.com/r/google/alloydbomni

I did not even know this was publicly available! Very cool!

Should be resolved now, lmk if you need any other changes.

PR #1463, opened 2026-05-05
fix: break mutual recursion in batch API fallback for non-batch providers

mssteuer:
## Problem

`extract_facts_from_contents()` and `extract_facts_from_contents_batch_api()` in `fact_extraction.py` form an infinite mutual recursion when the LLM provider doesn't support batch API:

1. `extract_facts_from_contents()` checks `config.retain_batch_enabled=True` → calls `extract_facts_from_contents_batch_api()`
2. `extract_facts_from_contents_batch_api()` checks `supports_batch_api()` → returns `False` (e.g. Gemini, Anthropic, LLaMA.cpp) → falls back to `extract_facts_from_contents()`
3. goto 1 → `RecursionError: maximum recursion depth exceeded` after ~1000 frames

## Impact

Any deployment using `retain_batch_enabled=True` with a provider that doesn't support batch API (Gemini, Anthropic, LLaMA.cpp, etc.) will crash on every retain operation. The queue fills with thousands of failed operations, all with `maximum recursion depth exceeded`.

Confirmed on both **v0.5.6** and **v0.6.0**.

## Fix

Use `dataclasses.replace()` to create a config copy with `retain_batch_enabled=False` before falling back, forcing the sync code path and breaking the cycle.

## Testing

Verified on a production instance (1.8M+ memories, Gemini flash-lite provider, pgvectorscale) — retains process successfully after the fix, zero recursion errors.


nicoloboschi (changes requested):
thanks for this pr. we should add this verification at the startup of the app when we do the llm verification and fail the api completetly. this will prevent and tell the user exactly what config is wrong

mssteuer:
Updated per your feedback — moved the validation to startup in `verify_llm()` so it fails fast with a clear error message. The runtime fallback is now a hard error instead of a silent sync-mode downgrade. Added tests covering all three config permutations. Let me know if anything else needs adjusting!

nicoloboschi (approved):
LGTM

PR #1465, opened 2026-05-06
fix(providers): strip JSON-Schema numeric constraints for Bedrock

gilmanb1:
## Summary

Bedrock's Claude tool-use API rejects `minimum`/`maximum` (and the `exclusiveMinimum`/`exclusiveMaximum` variants) on number/integer JSON-Schema nodes:

```
BedrockException - {"message":"The model returned the following errors:
For 'number' type, properties maximum, minimum are not supported"}
```

Pydantic's `Field(ge=..., le=...)` emits these constraints. Result: any structured-output / tool call whose schema contains a constrained `int` field — e.g. the bank disposition's `skepticism`/`literalism`/`empathy: int = Field(ge=1, le=5)` in `engine/response_models.py` — fails the `llm.bedrock.retain_extract_facts` stage with `BadRequestError`. The retain loop retries 3× and gives up with `0 facts, 0 chunks`. Documents accumulate in the bank but are never extracted into nodes/observations.

This PR walks the schema (and the `tools` list, for `call_with_tools`) and removes the four numeric-constraint keys before sending. Gated on `provider == "bedrock"` so other providers see zero behavior change. Pydantic still validates response shape on the Hindsight side after the call, so the constraints are not load-bearing at the wire.

## Reproduction

Running 0.5.4 against `bedrock/us.anthropic.claude-haiku-4-5-20251001-v1:0`, before the fix:

```
RETAIN_BATCH START: <bank>
Batch size: 1 content items, 195 chars
  Extract facts: 0 facts, 0 chunks from 1 contents in 6.5s
STREAMING RETAIN COMPLETE: 0 units across 1 batches in 6.5s
```

After:

```
  Extract facts: 3 facts, 1 chunks from 1 contents in 10.2s
  Insert facts: 3 units in 0.000s
  Insert unit_entities: 8 pairs in 0.004s
  Temporal links: 12 / Semantic links: 5 / Entity links: 6
STREAMING RETAIN COMPLETE: 3 units across 1 batches in 10.3s
```

Bank stats moved from `total_nodes: 0, total_links: 0` to `total_nodes: 6, total_links: 17` after a single retain.

## Tests

Adds `tests/test_litellm_bedrock_schema_strip.py` (9 tests, no network):

- top-level key removal
- exclusive variants
- nested object properties
- arrays
- `anyOf` / `oneOf`
- chained-call return-self
- no-op on clean schemas
- no-op on primitives
- end-to-end check on a Pydantic model with `Field(ge=, le=)` (mirrors the real failure case)

```
============================== 9 passed in 1.96s ==============================
```

## Test plan

- [x] Unit tests pass (`uv run pytest tests/test_litellm_bedrock_schema_strip.py`)
- [x] `ruff check` and `ruff format` clean
- [ ] Reviewer: confirm `provider` value at the call site matches what `_PROVIDERS_REQUIRING_NUMERIC_CONSTRAINT_STRIP` checks (just `"bedrock"`)
- [ ] Reviewer: consider whether other Bedrock-fronted providers (e.g. `bedrock_converse`) need to be added to the set

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
@@ -13,6 +13,7 @@ is handled automatically by LiteLLM.
 """
 
 import asyncio
+import copy
 import json
 import logging
 import time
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
we did a similar fix for this in https://github.com/vectorize-io/hindsight/pull/1295

I'd prefer to remove min, max from the pydantic declaration and add a code comment about it, not automaticlaly stripping for bedrock

nicoloboschi:
closing as stale

PR #1469, opened 2026-05-06
Split test suite into deterministic mock and real LLM buckets

DK09876:
## Summary

- Enhanced `MockLLM` with scope-aware responses (fact extraction, consolidation, reflect, tool calls) so pipeline tests exercise real behavior deterministically
- Default `memory` fixture now uses mock provider; `memory_real_llm` for quality-dependent tests
- **Three CI buckets**: `test-api` (mock, deterministic), `test-api-llm-core` (single provider, core quality), `test-api-llm-acceptance` (5-provider matrix)
- New `hs_llm_core` marker for tests that need a real LLM but only one provider
- Pre-existing `hs_llm_mat` marker unchanged for provider matrix acceptance tests
- Removed hollow `if observations:` guards — mock tests assert observation creation directly
- Added **LLM-as-a-judge** utility (`tests/llm_judge.py`) for semantic evaluation of LLM output, replacing brittle string matching in hs_llm_core tests

## Test plan

- [x] Full mock suite passes: 1904 passed, 0 failed
- [x] `hs_llm_core` suite passes: 51 passed, 0 failed (locally and CI)
- [x] LLM judge tests pass on CI (vertexai/gemini-2.5-flash-lite)
- [x] No hollow guards remain in deterministic tests
- [x] Pre-commit lints pass

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1342,8 +1342,8 @@ jobs:
         conn = oracledb.connect(user='system', password='oracle', dsn='localhost:1521/FREEPDB1')
         cursor = conn.cursor()
         cursor.execute(\"\"\"
-            CREATE TABLESPACE hindsight_ts
-            DATAFILE 'hindsight_ts.dbf' SIZE 1G AUTOEXTEND ON NEXT 200M MAXSIZE UNLIMITED
+            CREATE BIGFILE TABLESPACE hindsight_ts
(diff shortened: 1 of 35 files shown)


nicoloboschi (changes requested):
instead of adding those tests to hs_llm_mat , we should create another label and create their own job so we can distinguish (e.g. test-api (llm) vs test-api (core))

nicoloboschi (approved):
Approving. Three-bucket split looks right and the LLM-as-judge is wired into ~50 call sites across hs_llm_core tests. Two follow-ups (inline imports, llm-core trigger gating) will go in a separate PR.

PR #1478, opened 2026-05-06
fix: make health check require migrated schema

Tosko4:
## Summary

- Makes `/health` verify that the configured database contains the migrated Hindsight core schema, not just that PostgreSQL accepts `SELECT 1`.
- Reports `missing_schema_tables` with the missing table names when the service is connected to an empty/wrong database.
- Adds regression coverage for both the missing-schema and healthy-schema cases.

## Why

A pg0 instance mixup can leave a different empty database listening on the expected port. Before this change `/health` returned healthy because `SELECT 1` worked, while real memory operations failed with errors like `UndefinedTableError: relation "public.memory_units" does not exist`.

## Tests

- `cd hindsight-api-slim && uv run pytest tests/test_health_check_schema.py -q`
- `cd hindsight-api-slim && uv run ruff check hindsight_api/engine/memory_engine.py tests/test_health_check_schema.py`

Note: the repo pre-commit hook also tried to run the control-plane ESLint job, but this worktree does not have the frontend npm dependencies installed (`@eslint/js` missing). The Python checks for this backend-only change passed.


nicoloboschi (changes requested):
this is not lightweight as health check. 

how did you point to the wrong database? migration should have applied

Tosko4:
You're right, this is not the right shape for `/health`.

I traced the original failure back to a local environment mixup where the service was pointed at a different empty Postgres instance. In the normal startup path the migration should have applied, so making the lightweight health check verify the full core schema is papering over an operator/config issue in the wrong place.

I'm closing this rather than trying to force it through. If we want this kind of protection, it should be a separate diagnostic/readiness check, not a heavier `/health` response.

PR #1488, opened 2026-05-06
fix(worker): liveness-gate orphaned processing rows in bank-serialization

### LINKED ISSUE #1470: Worker poller never claims consolidation tasks despite reserved slot avail=1 and tasks claimable
## Bug Description

Worker poller never claims `consolidation` tasks even when the reserved consolidation slot is available (`avail=1`) and the pending consolidation tasks are reported as claimable. The same worker concurrently claims `batch_retain` tasks normally during the same poll cycles. Result: a single-tenant `agent:cortex` bank has accumulated **5 days** of stuck consolidation pending state with `pending_consolidation: 3603` and `failed_consolidation: 21`.

The pattern matches the upstream poller's own self-described "bug class" — see `hindsight-api-slim/hindsight_api/worker/poller.py:1121-1124`:

> `claimable` is the residual that *should* be picked up on the next poll. If `claimable > 0` while workers report free slots, the bug is somewhere else (lock contention, tenant discovery, etc.) - this line narrows the search.

## Steps to Reproduce

1. Run hindsight 0.5.6 single-tenant (single docker container, single worker process)
2. Submit some `retain` ingest writes to a bank (these complete normally)
3. Submit `POST /v1/default/banks/{bank_id}/consolidate` against the same bank
4. Observe `[WORKER_STATS]` and `[PENDING_BREAKDOWN]` log lines

## Expected Behavior

When `[WORKER_STATS]` shows `consolidation=0/1(avail=1)` AND `[PENDING_BREAKDOWN]` shows `consolidation: total=N claimable=N payload_null=0 retry_blocked=0 assigned=0`, the worker should claim a consolidation task on the next poll.

## Actual Behavior

The worker never claims consolidation tasks. Reserved consolidation slot remains permanently `avail=1`. Consolidation `claimable` count never decreases. The same worker claims `batch_retain` tasks in the same poll cycle without issue.

This persists indefinitely — across container restarts (our deployment had a stuck consolidation operation from 2026-05-01T11:30 that was still pending after a 2026-05-03 container restart, and is still pending now on 2026-05-06).

`POST /v1/default/banks/{bank_id}/consolidate` is also idempotent on pending state — it returns the existing pending operation_id rather than queuing a new one. So callers can't force forward progress by re-triggering.

### Pre-state on `agent:cortex` (single tenant, single worker)

```
$ curl -s http://hindsight:8891/v1/default/banks/agent:cortex/stats
{
  "bank_id": "agent:cortex",
  "total_nodes": 10229,
  "total_links": 438975,
  "total_documents": 5837,
  "total_observations": 2231,
  "last_consolidated_at": "2026-05-01T13:15:46+00:00",
  "pending_consolidation": 3603,
  "failed_consolidation": 21,
  "pending_operations": 23,
  "failed_operations": 0,
  "operations_by_status": {"completed": 11363, "pending": 23, "processing": 23}
}
```

### `POST /consolidate` returns the existing pending op

```
$ curl -X POST http://hindsight:8891/v1/default/banks/agent:cortex/consolidate
{"operation_id":"b42d9a0a-8dd3-44f7-ac6a-500d8241c64c","deduplicated":true}

$ curl -s http://hindsight:8891/v1/default/banks/agent:cortex/operations/b42d9a0a-8dd3-44f7-ac6a-500d8241c64c
{
  "operation_id": "b42d9a0a-8dd3-44f7-ac6a-500d8241c64c",
  "status": "pending",
  "operation_type": "consolidation",
  "created_at": "2026-05-01T12:00:26+00:00",   # 5 days ago
  "updated_at": "2026-05-01T12:00:26+00:00",
  "completed_at": null,
  "error_message": null,
  "retry_count": 0,
  "result_metadata": {},
  "child_operations": null,
  "task_payload": null
}
```

### Worker logs from immediately after a fresh container start (2026-05-02 14:22:58)

The reserved consolidation slot starts available and stays available — the worker happily claims `batch_retain` and `retain` in the same poll cycles but never claims any of the 5 pending consolidation tasks. This pattern is stable for the entire 5-day window since.

```
2026-05-02 14:22:58,183 [INFO] hindsight_api.worker.poller - Worker fc65a469677d starting polling loop (max_slots=2, reservations=[consolidation=1], shared_pool=1)
2026-05-02 14:22:58,716 [INFO] hindsight_api.worker.poller - [WORKER_STATS] worker=fc65a469677d slots=1/2 | reserved: [consolidation=0/1(avail=1)] | shared=1/1(avail=0) | global: pending=138 (schemas: default) | proc: rss_mb=717 | my_active: retain:agent:thalamus(1)
2026-05-02 14:22:58,716 [INFO] hindsight_api.worker.poller - [PENDING_BREAKDOWN] batch_retain: total=88 claimable=0 payload_null=88 retry_blocked=0 assigned=0 | consolidation: total=5 claimable=5 payload_null=0 retry_blocked=0 assigned=0 | retain: total=45 claimable=45 payload_null=0 retry_blocked=0 assigned=0
2026-05-02 14:23:12,799 [INFO] hindsight_api.worker.poller - Worker fc65a469677d claimed 1 tasks (0 consolidation): batch_retain:1 (schemas: default)
2026-05-02 14:23:17,830 [INFO] hindsight_api.worker.poller - Worker fc65a469677d claimed 1 tasks (0 consolidation): batch_retain:1 (schemas: default)
[... 5+ minutes of repeated `(0 consolidation)` claims ...]
```

Same pattern observed today (2026-05-06):

```
2026-05-06 04:00:59,857 [INFO] hindsight_api.worker.poller - [WORKER_STATS] worker=fc65a469677d slots=0/2 | reserved: [consolidation=0/1(avail=1)] | shared=0/1(avail=1) | global: pending=47 (schemas: default) | proc: rss_mb=2255 | my_active: none
2026-05-06 04:00:59,857 [INFO] hindsight_api.worker.poller - [PENDING_BREAKDOWN] batch_retain: total=42 claimable=0 payload_null=42 retry_blocked=0 assigned=0 | consolidation: total=5 claimable=5 payload_null=0 retry_blocked=0 assigned=0
```

Note the `[WORKER_STATS]` line shows BOTH the reserved consolidation slot AND the shared slot are available (`reserved: [consolidation=0/1(avail=1)] | shared=0/1(avail=1)`), and yet the consolidation `claimable=5` tasks remain unclaimed indefinitely.

### Reproducible probe

Calling `/consolidate` against an empty/sparse bank works fine — `agent:axon` (48 nodes, 16 docs) returned `{"operation_id":"...","deduplicated":false}` and the operation completed in 113ms with no state change. The bug only manifests on banks with a populated consolidation queue.

## Hypothesis

The claim SQL likely has a filter that excludes consolidation rows in some condition not reflected by the `[PENDING_BREAKDOWN]` predicates (`payload_null`, `retry_blocked`, `assigned`). Candidate causes:
- A consolidation-specific lock or advisory check is failing silently
- The `_claim_batch_for_schema_inner` → `backend.ops.claim_tasks()` path uses additional filtering for the consolidation type that isn't surfaced in the breakdown logger
- Something about the parent operation's `task_payload: null` is preventing child claim despite the parent being a consolidation row that itself shouldn't need a payload

The `[PENDING_BREAKDOWN]` line literally exists to narrow this search per the comment at `poller.py:1121-1124`, but it doesn't narrow it for our case — the bug is upstream of those four buckets.

## Workaround

Currently considering: bypass-and-recreate the stuck pending consolidation row directly in PostgreSQL, then re-trigger. But the parent op's idempotency means that without first deleting/cancelling the stuck pending row, all subsequent `/consolidate` calls just return the same dead operation_id.

Would value guidance on whether `POST /v1/default/banks/{bank_id}/consolidation/recover` is intended to handle this case (its description says "permanently failed", not "stuck pending").

## Version

0.5.6 (`ghcr.io/vectorize-io/hindsight:0.5.6`)

## LLM Provider

Anthropic (via Switchboard proxy in our case; PR #1389's `HINDSIGHT_API_LLM_DEFAULT_HEADERS` is wired)

## Setup / Env

- Single tenant, single worker process in single container
- Postgres backend (default)
- Worker config: `max_slots=2, reservations=[consolidation=1], shared_pool=1` (defaults)
- Container restart count: 0 since 2026-05-03 (stuck op pre-dates this restart by 2 days, restart did not unstick)

Happy to provide more logs, full PG row dumps for the stuck operation, or test against a patch.

TuftyBruno:
## Why

When a worker dies mid-consolidation (container restart, OOM), its `async_operations` row stays in `status='processing'` indefinitely. The current bank-serialization filter in `claim_tasks()` has no liveness check — it treats that orphaned row as a live lock and excludes the bank from all subsequent claim queries. Result: free reserved slot + `claimable=N` but zero tasks ever claimed, surviving container restarts, forever.

Reported in #1470 with a 5-day reproduction on a single-tenant 0.5.6 deployment.

## Root cause

Two-part chain:

**1. `ops_postgresql.py` — no liveness check on `busy_banks`**

Both the Phase 1 reserved-pool query (line 739) and Phase 2b shared-pool query (line 844) build a `busy_bank_ids` list from:

```sql
SELECT DISTINCT bank_id FROM {table}
WHERE operation_type = 'consolidation' AND status = 'processing'
```

No `updated_at` guard. A row stuck in `processing` for days matches identically to one a live worker is actively touching.

**2. `poller.py` `_recover_orphaned_tasks()` — `worker_id = $1` predicate**

Startup recovery resets rows with `worker_id = $1` (own worker only). A container restart gives the new worker a fresh `worker_id`, so the orphaned row from the old worker is never reset and survives restarts indefinitely.

The `[PENDING_BREAKDOWN]` logger doesn't apply the bank-busy filter — it only checks `payload_null`/`retry_blocked`/`assigned` — so it kept reporting `claimable=5` while the claim path silently excluded the bank. The mismatch is what made the symptom look paradoxical.

## Fix

**A. `ops_postgresql.py` (runtime self-healing)** — add a liveness gate to both `busy_banks` queries:

```sql
AND updated_at > NOW() - INTERVAL '1 hour'
```

A `processing` row untouched for >1h is treated as dead and stops blocking the bank. Threshold is intentionally conservative: real consolidation jobs complete in seconds to a few minutes; 1h gives headroom for legitimate slow runs while recovering without any restart.

**B. `poller.py` (startup recovery)** — broaden `_recover_orphaned_tasks()` to also reset stale cross-worker rows:

```sql
WHERE status = 'processing'
  AND result_metadata->>'batch_id' IS NULL
  AND (
    worker_id = $1
    OR updated_at < NOW() - INTERVAL '1 hour'
  )
```

Catches the container-restart case on the very next startup.

No migration required — `updated_at` is present in the initial schema (`5a366d414dce`) and is written on every claim/update transition (confirmed at `ops_postgresql.py:935`).

## Test plan

Four new tests in `TestOrphanedProcessingRowLiveness` (`tests/test_worker.py`):

- `test_stale_processing_row_does_not_block_claim` — pending task is claimed when the only blocking row is >1h stale (primary regression case)
- `test_fresh_processing_row_still_blocks_claim` — bank-serialization invariant preserved for recently-updated processing rows (no regression on the happy path)
- `test_recover_orphaned_tasks_resets_cross_worker_stale_rows` — startup recovery resets stale rows from other workers
- `test_recover_orphaned_tasks_does_not_reset_fresh_cross_worker_rows` — startup recovery leaves recently-updated cross-worker rows alone

All four pass locally. Run with:
```bash
cd hindsight-api-slim && uv run pytest tests/test_worker.py::TestOrphanedProcessingRowLiveness -v
```

Fixes #1470

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -739,7 +739,9 @@ class PostgreSQLOps(DataAccessOps):
                 busy_banks = await conn.fetch(
                     f"""
                     SELECT DISTINCT bank_id FROM {table}
-                    WHERE operation_type = 'consolidation' AND status = 'processing'
+                    WHERE operation_type = 'consolidation'
+                      AND status = 'processing'
+                      AND updated_at > NOW() - INTERVAL '1 hour'
                     """,
                 )
                 busy_bank_ids = [r["bank_id"] for r in busy_banks]
@@ -844,7 +846,9 @@ class PostgreSQLOps(DataAccessOps):
                 busy_banks_2 = await conn.fetch(
                     f"""
                     SELECT DISTINCT bank_id FROM {table}
-                    WHERE operation_type = 'consolidation' AND status = 'processing'
+                    WHERE operation_type = 'consolidation'
+                      AND status = 'processing'
+                      AND updated_at > NOW() - INTERVAL '1 hour'
                     """,
                 )
                 busy_bank_ids_2 = [r["bank_id"] for r in busy_banks_2]
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -707,7 +707,12 @@ class WorkerPoller:
                         f"""
                         UPDATE {table}
                         SET status = 'pending', worker_id = NULL, claimed_at = NULL, updated_at = now()
-                        WHERE status = 'processing' AND worker_id = $1 AND result_metadata->>'batch_id' IS NULL
+                        WHERE status = 'processing'
+                          AND result_metadata->>'batch_id' IS NULL
+                          AND (
+                            worker_id = $1
+                            OR updated_at < NOW() - INTERVAL '1 hour'
+                          )
                         """,
                         self._worker_id,
                     )
--- a/hindsight-api-slim/tests/test_worker.py
+++ b/hindsight-api-slim/tests/test_worker.py
@@ -3102,3 +3102,208 @@ class TestWorkerStatus:
         )
 
         assert len(rows) == 0
+
+
+class TestOrphanedProcessingRowLiveness:
+    """Regression tests for the orphaned-processing-row bug (issue #1470).
+
+    When a worker dies mid-consolidation (container restart, OOM), the
+    async_operations row can be left in status='processing' indefinitely.
+    The bank-serialization filter in claim_tasks() must treat such stale rows
+    as non-live (>1 hour since updated_at) so the bank becomes claimable again.
+    Similarly, _recover_orphaned_tasks() on startup must reset stale rows even
+    when they were owned by a different worker_id.
+    """
+
+    @pytest.mark.asyncio
+    async def test_stale_processing_row_does_not_block_claim(self, pool, backend, clean_operations):
+        """A consolidation row stuck in processing for >1h must not block claim.
+
+        Scenario: worker_A died mid-consolidation; its row has status='processing'
+        and updated_at=2 hours ago. A new worker (worker_B) should be able to
+        claim pending consolidation tasks for the same bank.
+        """
+        from hindsight_api.worker import WorkerPoller
+
+        bank_id = f"test-worker-{uuid.uuid4().hex[:8]}"
+        await _ensure_bank(pool, bank_id)
+
+        # Simulate the orphaned row left by dead worker_A: processing, updated 2h ago.
+        orphan_id = uuid.uuid4()
+        await pool.execute(
+            """
+            INSERT INTO async_operations
+              (operation_id, bank_id, operation_type, status, task_payload, worker_id, claimed_at, updated_at)
+            VALUES ($1, $2, 'consolidation', 'processing', $3::jsonb, 'dead-worker-A', now(), now() - INTERVAL '2 hours')
+            """,
+            orphan_id,
+            bank_id,
+            json.dumps({"type": "consolidation", "bank_id": bank_id, "operation_id": str(orphan_id)}),
+        )
+
+        # The pending task that should be claimed by worker_B.
+        pending_id = uuid.uuid4()
+        await pool.execute(
+            """
+            INSERT INTO async_operations
+              (operation_id, bank_id, operation_type, status, task_payload)
+            VALUES ($1, $2, 'consolidation', 'pending', $3::jsonb)
+            """,
+            pending_id,
+            bank_id,
+            json.dumps({"type": "consolidation", "bank_id": bank_id, "operation_id": str(pending_id)}),
+        )
+
+        poller = WorkerPoller(
+            backend=backend,
+            worker_id="worker-B",
+            executor=lambda x: None,
+        )
+
+        claimed = await poller.claim_batch()
+        claimed_ids = {c.operation_id for c in claimed}
+
+        assert str(pending_id) in claimed_ids, (
+            "Worker should claim the pending consolidation task when the only "
+            "blocking processing row is stale (>1h). "
+            "Regression: orphaned row from dead worker must not block claim forever."
+        )
+
+    @pytest.mark.asyncio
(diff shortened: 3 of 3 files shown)


nicoloboschi:
see comment https://github.com/vectorize-io/hindsight/issues/1470#issuecomment-4396265305

Issue #1394, opened 2026-05-02
Local daemon crashes on recall for bank with exactly one memory


zicochaos:
## Bug Description

On local embedded Hindsight `0.5.6`, `recall` against a bank containing exactly one memory can terminate the local daemon. The client receives `ServerDisconnectedError('Server disconnected')`; an immediate follow-up request then fails with connection refused because the daemon is no longer listening on `127.0.0.1:9042`.

This does not reproduce for an empty bank, and it stopped reproducing after adding a second memory to the affected bank.

## Environment

- Package versions:
  - `hindsight-embed==0.5.6`
  - `hindsight-client==0.5.6`
- Runtime mode: local embedded daemon, REST API on `http://127.0.0.1:9042`
- Python: 3.12 on macOS arm64
- PostgreSQL + pgvector local embedded profile
- Embeddings provider: OpenAI-compatible endpoint, `text-embedding-3-small`
- LLM provider: OpenAI-compatible endpoint, model `openai/gpt-oss-120b`
- Reranker: local `cross-encoder/ms-marco-MiniLM-L-6-v2`, CPU mode

## Steps to Reproduce

Using the sync client against a healthy local daemon:

```python
from hindsight_client import Hindsight

bank = "single-memory-recall-crash-repro"
client = Hindsight(base_url="http://127.0.0.1:9042", timeout=60.0)

client.retain(
    bank_id=bank,
    content="Starred inbox emails must remain in inbox unless explicitly approved for archive.",
    context="debug single memory bank",
)

# This can terminate the daemon when the bank has exactly one memory.
client.recall(bank_id=bank, query="starred inbox emails archive")
```

## Expected Behavior

`recall` should return the single matching memory or an empty result set, and the daemon should remain healthy.

## Actual Behavior

The recall request disconnects and the local daemon exits.

Client-side result:

```text
ServerDisconnectedError('Server disconnected')
```

Immediate health check:

```text
curl: (7) Failed to connect to 127.0.0.1 port 9042: Couldn't connect to server
```

Relevant log tail around the crash:

```text
[RECALL <bank-prefix>] Starting recall for query: generic inbox email request...
hindsight_api.engine.search.retrieval - Using LinkExpansion graph retriever
/Users/.../python3.12/multiprocessing/resource_tracker.py:279: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

[...]

Issue #270, opened 2026-02-01
macOS ARM64: SIGBUS crash during consolidation - fix with LAZY_RERANKER


ragnos-dev:
## Summary

Hindsight daemon crashes in an infinite loop during consolidation on macOS ARM64 (Apple Silicon) with Python 3.12. The crash is caused by a SIGBUS (bus error) in PyTorch BLAS operations when the cross-encoder initializes. This can be resolved by setting `HINDSIGHT_API_LAZY_RERANKER=true` to defer PyTorch loading.

## Environment

- **OS**: macOS 15.2 (Sequoia) on Apple Silicon (ARM64)
- **Python**: 3.12.11
- **hindsight-api**: 0.4.7
- **hindsight-embed**: 0.4.7
- **Deployment**: Embedded daemon mode (no external API)

### Observed Behavior

The daemon enters an infinite crash loop:
1. Daemon starts successfully
2. Consolidation task begins
3. Process crashes with SIGBUS during cross-encoder initialization
4. Daemon auto-restarts (or is restarted by supervisor)
5. Loop repeats indefinitely

### Crash Details

From macOS crash report:
```
Exception Type: EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: EXC_ARM_DA_ALIGN at 0x000000013e318d7e
Termination: Bus error: 10

Faulting Thread: Thread 20 (consolidation worker)
Stack trace shows:
  - pytorch BLAS: cblas_sgemv → SGEMM
  - torch.linear operations
  - sentence_transformers.CrossEncoder inference
```

### Daemon Log Pattern

```
INFO - Claiming async_operation [consolidation]
INFO - Starting consolidation for bank=openclaw
[CRASH - no error logged, process exits]
... daemon restarts ...
[Pattern repeats]
```

## Root Cause

The crash occurs when `LocalSTCrossEncoder` loads PyTorch models during consolidation. On macOS ARM64, PyTorch's BLAS operations trigger memory alignment faults (SIGBUS) in certain process configurations.

**Key findings:**
- `TOKENIZERS_PARALLELISM=false` alone is insufficient
- CPU forcing (`HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU=1`) doesn't prevent the crash
- The issue is specific to when/how PyTorch initializes, not just parallelism

## Solution

Set the environment variable:
```bash
HINDSIGHT_API_LAZY_RERANKER=true
```

[...]

PR #1498, opened 2026-05-06
fix(embed): default Apple Silicon reranker to RRF

### LINKED ISSUE #1497: macOS ARM64 embedded daemon SIGBUS during consolidation with local reranker
## Bug Description

On macOS ARM64, local embedded Hindsight can hard-crash the daemon with `Fatal Python error: Bus error` during observations/consolidation when the default local SentenceTransformers cross-encoder reranker is used.

This looks related to the closed macOS ARM64 SIGBUS issue (#270), but in my current `0.6.0` repro, `HINDSIGHT_API_LAZY_RERANKER=true` does **not** prevent the crash once local reranker inference is actually reached. Setting `HINDSIGHT_API_RERANKER_PROVIDER=rrf` avoids the crash in the same isolated repro.

## Environment

- OS: macOS 14.5, Darwin ARM64 (`arm64`)
- Python used for repro: 3.11.12
- Packages: `hindsight-api-slim==0.6.0`, `hindsight-embed==0.6.0`
- Mode: local embedded daemon (`HindsightEmbedded` / `hindsight-api --daemon`)
- LLM provider tested:
  - `openai_compatible` via a local OpenAI-compatible endpoint
  - native `openai-codex` was also tested and did not avoid the crash
- Reranker provider when failing: default/local (`cross-encoder/ms-marco-MiniLM-L-6-v2`)
- Observations/consolidation: enabled in an isolated throwaway profile/bank

No production/shared bank was used in the repro.

## Steps to Reproduce

Using an isolated embedded profile/bank:

1. Start Hindsight embedded with observations/consolidation enabled.
2. Use the default local reranker provider.
3. Retain content that extracts several facts/units.
4. Allow the background consolidation worker to claim a consolidation task.
5. While consolidation is active, issue another retain request, so retain overlaps with consolidation recall/rerank.

Relevant environment shape:

```bash
PYTHONFAULTHANDLER=1
HINDSIGHT_API_ENABLE_OBSERVATIONS=true
HINDSIGHT_API_WORKER_CONSOLIDATION_MAX_SLOTS=1
HINDSIGHT_API_RERANKER_PROVIDER=local
# also tested:
# HINDSIGHT_API_LAZY_RERANKER=true
# HINDSIGHT_API_RERANKER_LOCAL_MAX_CONCURRENT=1
# HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU=1
```

## Expected Behavior

The consolidation worker should either complete or fail as a normal Python exception/task failure without terminating the embedded daemon process.

## Actual Behavior

The daemon process exits. The client then receives `ServerDisconnectedError` / connection refused on subsequent requests.

The daemon log tail includes:

```text
retain #1 extracted multiple units
consolidation task queued / claimed
[CONSOLIDATION] total_unconsolidated=...
Using LinkExpansion graph retriever
llm_batch #1 ... created=1
Fatal Python error: Bus error
resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
```

With `PYTHONFAULTHANDLER=1`, the fatal Python stack points at local cross-encoder inference:

```text
Fatal Python error: Bus error

... torch.nn.modules.linear.forward
... transformers.models.bert.modeling_bert.forward
... sentence_transformers.cross_encoder.model.CrossEncoder.predict
... hindsight_api.engine.cross_encoder.LocalSTCrossEncoder._predict_sync
... hindsight_api.engine.search.reranking.CrossEncoderReranker.rerank
... consolidation recall/rerank path
```

So `Using LinkExpansion graph retriever` appears to be a trigger/context before reranking, not the immediate crashing function.

## Workarounds Tried

Failed to prevent crash:

```bash
HINDSIGHT_API_LAZY_RERANKER=true
HINDSIGHT_API_RERANKER_LOCAL_MAX_CONCURRENT=1
HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU=1
```

Passed in the same isolated overlapping retain/consolidation repro:

```bash
HINDSIGHT_API_RERANKER_PROVIDER=rrf
```

With `rrf`, both retain calls completed, consolidation completed, the daemon stayed alive, and recall succeeded.

## Related Issues / PRs

- #270 — macOS ARM64 SIGBUS during consolidation; closed with `LAZY_RERANKER`, but lazy did not avoid this current `0.6.0` repro.
- #1394 — local daemon crashes on recall for a bank with exactly one memory. This may be separate, but it has similar symptoms (`ServerDisconnectedError`, `Using LinkExpansion graph retriever`, `resource_tracker leaked semaphore`).
- #957 — relevant if `rrf` is used as a workaround because it preserves RRF ranking for passthrough reranking.

## Proposed Direction

A conservative embedded-mode mitigation is to avoid the local SentenceTransformers cross-encoder by default on Apple Silicon embedded daemons, unless the user explicitly opts into `HINDSIGHT_API_RERANKER_PROVIDER=local` or another provider.

I have a small PR prepared that defaults Apple Silicon embedded daemons to:

```bash
HINDSIGHT_API_RERANKER_PROVIDER=rrf
```

only when the user has not explicitly set a reranker provider, plus tests confirming explicit overrides are respected.

lightofjeju:
## Summary

- Default Apple Silicon embedded daemons to `HINDSIGHT_API_RERANKER_PROVIDER=rrf` when the user has not explicitly selected a reranker provider.
- Keep existing macOS force-CPU defaults for local embeddings/reranker.
- Add regression tests for the Apple Silicon RRF default and for respecting an explicit reranker provider override.

## Why

On macOS ARM64, the local SentenceTransformers cross-encoder reranker can hard-crash the embedded daemon with `Fatal Python error: Bus error` during consolidation recall/rerank. This is a native PyTorch/SentenceTransformers crash, so it cannot be handled as a normal Python exception inside the worker task.

In an isolated local embedded repro:

- `HINDSIGHT_API_LAZY_RERANKER=true` did not prevent the crash once local reranker inference was reached.
- `HINDSIGHT_API_RERANKER_LOCAL_MAX_CONCURRENT=1` did not prevent the crash.
- `HINDSIGHT_API_RERANKER_PROVIDER=rrf` avoided the crash in the same overlapping retain/consolidation scenario and recall succeeded.

This mitigation is limited to Apple Silicon embedded daemons and only applies when the user has not explicitly configured a reranker provider. Users can still opt back into local reranking or select a remote reranker by setting `HINDSIGHT_API_RERANKER_PROVIDER`.

## Test Plan

```bash
cd hindsight-embed
uv run pytest tests/test_profile_daemon_config.py -q
uv run ruff check hindsight_embed/daemon_embed_manager.py tests/test_profile_daemon_config.py
uv run ruff format --check hindsight_embed/daemon_embed_manager.py tests/test_profile_daemon_config.py
```

Results locally:

```text
14 passed in 0.20s
All checks passed!
2 files already formatted
```

## Related

- Related to #270
- Related to #1394
- RRF passthrough ranking behavior is related to #957

Fixes #1497

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -380,6 +380,23 @@ class DaemonEmbedManager(EmbedManager):
             if "HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU" not in env:
                 env["HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU"] = "1"
 
+            # On Apple Silicon, forcing the local SentenceTransformers
+            # cross-encoder to CPU avoids MPS startup hangs but does not avoid
+            # SIGBUS crashes seen during local reranker inference in embedded
+            # daemon mode (notably during consolidation recall/rerank). The
+            # crash is a native PyTorch/SentenceTransformers failure and cannot
+            # be caught by Python exception handling, so prefer the RRF
+            # passthrough reranker unless the user explicitly opts into another
+            # provider.
+            if platform.machine().lower() in {"arm64", "aarch64"} and "HINDSIGHT_API_RERANKER_PROVIDER" not in env:
+                env["HINDSIGHT_API_RERANKER_PROVIDER"] = "rrf"
+                logger.warning(
+                    "Apple Silicon embedded daemon detected; defaulting "
+                    "HINDSIGHT_API_RERANKER_PROVIDER=rrf to avoid local "
+                    "cross-encoder SIGBUS crashes. Set "
+                    "HINDSIGHT_API_RERANKER_PROVIDER explicitly to override."
+                )
+
         # Get idle timeout from env
         idle_timeout = int(env.get("HINDSIGHT_EMBED_DAEMON_IDLE_TIMEOUT", str(DEFAULT_DAEMON_IDLE_TIMEOUT)))
 
--- a/hindsight-embed/tests/test_profile_daemon_config.py
+++ b/hindsight-embed/tests/test_profile_daemon_config.py
@@ -282,6 +282,89 @@ def test_macos_forces_cpu_for_local_embeddings_and_reranker(temp_home, monkeypat
     assert env.get("HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU") == "1"
 
 
+def test_macos_arm64_defaults_reranker_to_rrf(temp_home, monkeypatch):
+    """Apple Silicon embedded daemons should avoid the local cross-encoder by default.
+
+    FORCE_CPU avoids MPS startup hangs, but local SentenceTransformers reranker
+    inference can still SIGBUS the embedded daemon on macOS ARM64. Default to
+    the RRF passthrough reranker unless the user explicitly selects a provider.
+    """
+    from unittest.mock import MagicMock, patch
+
+    from hindsight_embed.daemon_embed_manager import DaemonEmbedManager
+
+    monkeypatch.delenv("HINDSIGHT_API_RERANKER_PROVIDER", raising=False)
+
+    manager = DaemonEmbedManager()
+    captured: dict[str, dict[str, str]] = {}
+    popen_called = [False]
+
+    def fake_popen(cmd, env, **kwargs):
+        captured["env"] = env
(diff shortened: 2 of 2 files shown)


nicoloboschi:
rrf doesn't have a good quality, let's find out the root cause of the crash on apple silicon

PR #1501, opened 2026-05-07
feat(reranker): support alibaba qwen3-rerank

quicklyfast:
support alibaba qwen3-rerank


nicoloboschi:
A few things before this can land:

**1. Reuse `_CohereCompatibleRerankClient` instead of re-implementing it**

The DashScope compatible endpoint speaks the same wire format as the existing helper at `hindsight_api/engine/cross_encoder.py:530` (`{model, query, documents, top_n}` → `{results: [{index, relevance_score}]}`). `_rerank_compatible` duplicates that logic. Compose the helper the way `SiliconFlowCrossEncoder` (line 768) and `ZeroEntropyCrossEncoder` (line 726) do — you get the query-grouping and request shape for free.

**2. Drop `_COMPATIBLE_MODELS`**

Hardcoding `frozenset({"qwen3-rerank"})` means any new variant (`qwen3-rerank-plus`, future `qwen3-reranker-*`, etc.) silently routes to the native endpoint and fails with a shape mismatch. Pick one:
- Only support the compatible endpoint and remove the native path entirely (simplest — `qwen3-rerank` is the headline model here).
- Or expose endpoint selection as an explicit config flag rather than inferring it from the model name.

**3. Documentation is missing**

- `hindsight-docs/docs/developer/configuration.md:556` — add `alibaba` to the `HINDSIGHT_API_RERANKER_PROVIDER` value list, plus rows for `HINDSIGHT_API_RERANKER_ALIBABA_API_KEY` / `_MODEL` next to the SiliconFlow block (~line 583).
- `hindsight-docs/docs/developer/models.mdx:475` — provider table entry and an example env-var block (mirror the SiliconFlow one at line 553).

Also: the class docstring says auth comes from `DASHSCOPE_API_KEY or HINDSIGHT_API_RERANKER_ALIBABA_API_KEY`, but `from_env` only reads the latter — either drop the claim or add the fallback (Cohere does it at `config.py:1581`).

quicklyfast:
@nicoloboschi  Please review it.

nicoloboschi (approved):
LGTM

PR #1510, opened 2026-05-07
fix(worker): liveness-gate busy_banks for stuck processing rows (runtime, scoped from #1488)

TuftyBruno:
## Why (smaller-scope follow-up after #1488 closure)

Closing #1488 redirected to setting `HINDSIGHT_API_WORKER_ID` to a stable value as the canonical fix (#1470). That covers planned graceful restarts cleanly. It doesn't cover the cases where *the worker can't process the task it already claimed*, which is the second-reporter scenario [@descention raised](https://github.com/vectorize-io/hindsight/issues/1470#issuecomment-4396697812):

> I'm running Hindsight in podman. My llama-cpp instance had recently been updated with a different model name, so hindsight was getting 400's from llama-cpp. A consolidation task was queued. Worker `b6627cebef89` picked it up but obviously couldn't process it. I updated my container's configuration and restarted.

In their case stable `WORKER_ID` doesn't help — they restarted with a config fix, but the orphan row is still owned by `b6627cebef89` and `_recover_orphaned_tasks()` only resets rows owned by the current `worker_id`.

We hit the same chain in our deployment via different cause (worker died mid-OOM, no graceful exit). The stuck `processing` rows on `agent:cortex` had 8 different historical worker UUIDs across multiple container restarts — none matched the current worker, so the recovery query was a no-op every cycle. We unstuck it manually via `UPDATE async_operations SET status='pending', worker_id=NULL WHERE status='processing'` (48 rows reset, bank started draining within 20 seconds).

## What this PR does (smaller than #1488)

Adds `AND updated_at > NOW() - INTERVAL '1 hour'` to both `busy_banks` queries in `claim_tasks()` (Phase 1 reservation at line 739, Phase 2b shared pool at line 844). A `processing` row not touched for >1 hour is treated as dead and no longer blocks the bank from being claimed.

Scope is intentionally **smaller than #1488**:

| Path                          | #1488 | This PR | Owner               |
|-------------------------------|-------|---------|---------------------|
| `busy_banks` runtime gate     | ✅     | ✅       | Self-healing        |
| `_recover_orphaned_tasks()` startup recovery broaden | ✅     | ❌       | `HINDSIGHT_API_WORKER_ID` covers it |

The startup-recovery side is left untouched. Operators continue to set `HINDSIGHT_API_WORKER_ID` to a stable value (the documented contract); this PR is a runtime backstop for the worker-stuck cases that stable id alone cannot reach.

## Threshold rationale

1 hour is intentionally conservative. Real consolidation jobs complete in seconds to minutes (consistent with `STUCK_STACK_INITIAL_THRESHOLD_S` × 12 = 1h elsewhere in the codebase). The threshold could be made configurable via env var if you'd like; not added in this PR to keep the diff minimal.

## No migration

`updated_at` is in the initial schema (`5a366d414dce`) and is written on every claim/update transition. Confirmed `ops_postgresql.py:935` writes it on every status flip.

## Tests

`TestBusyBankLivenessGate` (`tests/test_worker.py`):

- `test_stale_processing_row_does_not_block_claim` — pending task is claimed when the only blocking row is >1h stale (primary regression case from #1470)
- `test_fresh_processing_row_still_blocks_claim` — bank-serialization invariant preserved for recently-updated processing rows

Run with:

```bash
cd hindsight-api-slim && uv run pytest tests/test_worker.py::TestBusyBankLivenessGate -v
```

Both pass against current `main`.

Refs #1470, scoped-down successor to closed #1488.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -736,10 +736,21 @@ class PostgreSQLOps(DataAccessOps):
                 continue
 
             if op_type == "consolidation":
+                # Liveness gate: a `processing` row whose `updated_at` is
+                # older than 1 hour is treated as dead and stops blocking
+                # the bank. Real consolidation jobs complete in seconds to
+                # a few minutes; 1 hour gives headroom for legitimate slow
+                # runs while recovering from worker-stuck cases that
+                # `_recover_orphaned_tasks` cannot (orphans owned by a
+                # worker_id that no longer exists, e.g. after OOM, scale-
+                # down, or any non-graceful exit). See vectorize-io/
+                # hindsight#1470 for two independent reproductions.
                 busy_banks = await conn.fetch(
                     f"""
                     SELECT DISTINCT bank_id FROM {table}
-                    WHERE operation_type = 'consolidation' AND status = 'processing'
+                    WHERE operation_type = 'consolidation'
+                      AND status = 'processing'
+                      AND updated_at > NOW() - INTERVAL '1 hour'
                     """,
                 )
                 busy_bank_ids = [r["bank_id"] for r in busy_banks]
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
hey, this fix will not work reliably. 

as I said many times now, this could easily lead to workers doing the same work in parallel and we'd need an coordinator system to maange distributed workers correctly. 
intentionally we don't cross this line for keeping the administration simple

TuftyBruno:
Fair call — closing.

The duplicate-work race is real for multi-worker deployments and the threshold-based gate doesn't safely cross that line without proper coordination primitives. Operationally, stable `HINDSIGHT_API_WORKER_ID` + `hindsight-admin decommission-worker` (which I learned about from @descention's comment) is the right pattern for our single-worker setup, and it's what actually unstuck our 5-day backlog today.

Thanks for the principled reviews on this and #1488. Will route any future work as proper distributed coordination if/when we hit a multi-worker case that warrants it.

PR #1523, opened 2026-05-07
fix(reflect): read document metadata from retain params

nepenth:
## Bug Description

`tool_expand(depth="document")` still queried `documents.metadata`, but the `d6e7f8a9b0c1_drop_documents_metadata_column` migration intentionally removed that column. On migrated databases this causes reflect/document expansion to fail with:

```text
column "metadata" does not exist
```

## Root Cause

The migration documents that document metadata is stored in `documents.retain_params.metadata`, but the reflect expansion path was still selecting and returning the removed `documents.metadata` column.

## Fix

- Stop selecting `documents.metadata` in `tool_expand()`.
- Preserve the public response shape by returning `document["metadata"]` from `retain_params.metadata`.
- Accept `retain_params` as either a decoded JSON object or a JSON string.
- Treat malformed/non-object metadata as absent instead of breaking reflect expansion.
- Add focused regression coverage for both document expansion branches.

## How to Verify

1. Run `uv run pytest tests/test_reflect_tools.py -q`.
2. Run `uv run ruff check hindsight_api/engine/reflect/tools.py tests/test_reflect_tools.py`.
3. Run `uv run ty check hindsight_api/engine/reflect/tools.py`.

## Test Plan

- [x] Added regression tests for this bug.
- [x] Verified the new regression test failed before the implementation change because `tool_expand()` selected `documents.metadata`.
- [x] `uv run pytest tests/test_reflect_tools.py -q` — 7 passed.
- [x] `uv run ruff format hindsight_api/engine/reflect/tools.py tests/test_reflect_tools.py`.
- [x] `uv run ruff check hindsight_api/engine/reflect/tools.py tests/test_reflect_tools.py`.
- [x] `uv run ty check hindsight_api/engine/reflect/tools.py`.

## Risk Assessment

Low — this is scoped to reflect document expansion. It aligns `tool_expand()` with the existing schema migration and existing document API behavior while preserving the returned `document["metadata"]` field.

## Notes

This PR intentionally does not add a migration. The schema is already correct; the bug is stale application code querying a removed column.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/tools.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/tools.py
@@ -7,6 +7,7 @@ Implements hierarchical retrieval:
 3. recall - Raw facts as ground truth
 """
 
+import json
 import logging
 import uuid
 from dataclasses import replace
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
good catch!

PR #1527, opened 2026-05-07
fix(worker): propagate child error_message to failed batch_retain parent

cdbartholomew:
## Summary

When a \`batch_retain\` parent transitions to \`failed\` because at least one child sub-batch failed, the parent's \`error_message\` was hardcoded to the generic string \`"One or more sub-batches failed"\`. Any consumer that classifies failures by \`error_message\` (dashboards, alert filters, log aggregators) loses signal once a batch grows children — a class of failures that all share the same root reason at the child level becomes indistinguishable at the parent level.

## Fix

Pull \`error_message\` in the siblings query and pick the most-common non-empty failed-child message as the parent's \`error_message\`. When all siblings failed for the same reason (the common case) the parent inherits that reason verbatim; when reasons vary the most-common one is still a useful representative. Falls back to the legacy generic string only when no failed sibling carries an \`error_message\` at all, preserving backward compat for that edge case.

Same change applied to both:
- the worker poller's fallback path (\`worker/poller.py:_maybe_update_parent_operation\`)
- the memory engine's in-transaction path (\`engine/memory_engine.py:_maybe_update_parent_operation\`)

so the propagation behavior is consistent regardless of which surface finalises the parent.

## Behavior change

| Before | After |
|---|---|
| Parent error_message: \`\"One or more sub-batches failed\"\` | Parent error_message: \`<most-common failed child's error_message>\` |
| Generic-fallback when no children have errors | Same (preserved for backward compat) |

## Test plan

- [x] 6 new pure unit tests for \`_summarise_child_error_messages\` covering: same-message inheritance, mixed-message most-common selection, completed-sibling skip, no-failed fallback, no-error-message fallback, whitespace-only-message fallback
- [x] Existing integration test \`test_mark_failed_finalises_parent_when_last_sibling_fails\` extended with an assertion that the parent \`error_message\` contains the child's reason
- [x] Full \`TestMarkFailedParentPropagation\` + \`TestSummariseChildErrorMessages\` suite (11 tests) passes locally
- [x] No tests asserted on the legacy generic string anywhere in the repo

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -1654,11 +1654,16 @@ class MemoryEngine(MemoryEngineInterface):
                 # Parent doesn't exist (shouldn't happen)
                 return
 
-            # Get all sibling operations (including this one)
-            # This query runs in the same transaction, so it sees the current child's updated status
+            # Get all sibling operations (including this one).
+            # This query runs in the same transaction, so it sees the current
+            # child's updated status. Pull error_message too so a parent that
+            # fails can inherit a representative child reason -- otherwise
+            # downstream consumers (dashboards, alert filters) lose the actual
+            # cause once a batch has children. See the worker poller's
+            # _summarise_child_error_messages for the propagation rationale.
             siblings = await conn.fetch(
                 f"""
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
LGTM

PR #1557, opened 2026-05-09
fix(agent-sdk): agent_knowledge_get_page request detail=content (sister of #1543)

r266-tech:
## Summary

Sister fix to **#1543** (`fix(claude-code): get_page detail=content + handle tool-result spillover`, merged 2026-05-08). #1543 fixed `agent_knowledge_get_page` in the Python `claude-code` MCP server to request `?detail=content` instead of the API default `?detail=full`. The same bug exists in the parallel TypeScript implementation that ships as the published `@vectorize-io/hindsight-agent-sdk@0.1.0` package.

`hindsight-tools/hindsight-agent-sdk/src/index.ts:120` calls `client.getMentalModel(bankId, page_id)`; `HindsightClient.getMentalModel()` does not accept a `detail` option and uses the SDK default. Per #1543's measurement, that path returns the synthesized content **plus** `reflect_response` (internal trace metadata, **70–95% of response bytes**, ~200–280 KB at full). MCP hosts spill the response to disk, where the agent cannot consume it inline as part of its startup knowledge load. The agent-sdk `agent_knowledge_get_page` tool docstring already promises only "the full synthesized content".

This PR follows the **same pattern already used in this file** for `agent_knowledge_list_pages` (lines 99–106, mirror of #1528's parallel fix): bypass the high-level `HindsightClient` wrapper and call the low-level `sdk.getMentalModel({ client: lowLevel, path, query: { detail: "content" } })` directly, then `return ok(resp.data)`. The high-level wrapper at `hindsight-clients/typescript/src/index.ts:797` does not need to change.

## Diff (4 net lines added in `hindsight-tools/hindsight-agent-sdk/src/index.ts`)

```diff
       async execute(params: Record<string, unknown>) {
-        const result = await client.getMentalModel(bankId, params.page_id as string);
-        return ok(result);
+        const resp = await sdk.getMentalModel({
+          client: lowLevel,
+          path: { bank_id: bankId, mental_model_id: params.page_id as string },
+          query: { detail: "content" },
+        });
+        if (resp.error) {
+          throw new Error(
+            `agent_knowledge_get_page failed: ${JSON.stringify(resp.error)}`,
+          );
+        }
+        return ok(resp.data);
       },
```

## Why low-level `sdk.*` instead of teaching `HindsightClient.getMentalModel()` a `detail` option

- **Match the local convention**: `agent_knowledge_list_pages` already does this exact thing for the same reason (`detail: "metadata"` projection). The two tools should keep the same shape.
- **Minimum change to the published `@vectorize-io/hindsight-client` surface**: changing `getMentalModel`'s signature would be a public API change in a different package; this fix stays inside the agent-sdk.

The error-check after `resp.error` is added so that 401/404/500 responses surface as thrown errors — the high-level `HindsightClient.getMentalModel()` already throws via `validateResponse(...)` and consumers expect failures, not silent `undefined` data. Without this check, the low-level `sdk.*` path returns `{ data: undefined, error: <body> }` and would propagate `undefined` into the tool result.

## Known related gap (out of scope)

The existing `agent_knowledge_list_pages` call site at lines 99–106 has the same low-level `sdk.*` pattern **without** an `resp.error` check. That gap is pre-existing and separate from the #1543 sister-fix; happy to fold a consolidation patch into this PR or follow-up if preferred.

## Relation to other open PRs

- **#1552** (this account, OPEN): parallel sister-fix to **#1544** for `agent_knowledge_recall` schema rename `max_results → max_tokens` + dual-accept for v0.1.0 backward compat. Same file, different function. Both PRs together close the two TS SDK call-site gaps that #1543 + #1544 fixed in Python MCP same-day.
- The two PRs are independent; either can be merged or rejected on its own merits.

## Test plan

- [ ] No new test added in this patch; the agent-sdk currently has no tests (`hindsight-tools/hindsight-agent-sdk/tests/` does not exist). Same project state as #1552. Happy to add one if maintainer prefers.

## Cherry-pick aware

If this has already landed via a separate patch, please feel free to close + mark accordingly.


benfrank241:
Thanks for the contribution, could you take a look at the failing test?

r266-tech:
Thanks for the look. Investigated the `verify-generated-files` failure on this run:

The job runs the four `scripts/generate-*.sh` + `scripts/hooks/lint.sh` then fails if `git status --porcelain` is dirty. The dirty list for `5328798` was:

```
 M hindsight-integrations/paperclip/src/worker.ts
 M hindsight-integrations/paperclip/tests/plugin.spec.ts
 M hindsight-tools/hindsight-agent-sdk/src/index.ts
 M skills/hindsight-docs/references/developer/configuration.md
 M skills/hindsight-docs/references/developer/installation.md
```

Of those, 4 are pre-existing drift on `main` (independent of this PR): the same job on the latest full `main` run [b515bc05](https://github.com/vectorize-io/hindsight/actions/runs/25700794465/job/75592260664) already flags `hindsight-integrations/paperclip/src/worker.ts`, and the docs-skill + paperclip-test rebuilds appear to have drifted further on the post-b515bc05 merges that haven't had a full CI run yet. #1596 also failed this exact check just before merge.

The 5th file (`hindsight-tools/hindsight-agent-sdk/src/index.ts`) is mine — `prettier --write` (via the `prettier-tool-hindsight-agent-sdk` task in `lint.sh`) flattens the 3-line `throw new Error(\`...\`)` into a single line at `printWidth: 100`. Just pushed `235d88f2` which inlines the throw so it survives a fresh prettier pass.

After this, any remaining drift in the check should be the 4 pre-existing main-state files, which I think need a separate `chore: regenerate` commit on main rather than expanding this PR's scope. Happy to file that as a sibling PR if useful — let me know.

PR #1560, opened 2026-05-09
fix(paperclip): align with Paperclip's actual event payloads

amirhmoradi:
The plugin's `agent.run.started` and `agent.run.finished` handlers destructured fields (`issueTitle`, `issueDescription`, `output`, `result`) that Paperclip's host does not publish. Paperclip emits a thin lifecycle payload — `{runId, agentId, status, invocationSource, triggerDetail, error, errorCode, issueId, startedAt, finishedAt}` — so both handlers silently early-returned and the plugin never recalled or retained anything despite registering successfully.

Changes:

- `agent.run.started` now uses `payload.issueId` to look up the issue via `ctx.issues.get` and builds the recall query from the issue's title + description.

- New `issue.comment.created` subscription replaces the `agent.run.finished` retain path. Comments are the durable record of agent + user output and the existing payload only carries a 120-char snippet, so we fetch the full body via `ctx.issues.listComments`. Bank attribution falls back to the issue's assignee when a comment has no agent author (e.g. user comments).

- `agent.run.finished` is kept as a debug no-op so the subscription stays visible and can be reused if Paperclip ever embeds output in the lifecycle payload.

- Manifest gains `issues.read` and `issue.comments.read` capabilities, required by the new SDK calls.

- Tests updated to seed issues/comments via the harness, exercise the new comment-created path, and cover the assignee-fallback for unauthored comments.

Verified end-to-end against a local Paperclip + self-hosted Hindsight: the patched plugin retains real comment bodies to the correct bank and Hindsight's recall API returns them on subsequent queries.

### DIFF
--- a/hindsight-integrations/paperclip/src/manifest.ts
+++ b/hindsight-integrations/paperclip/src/manifest.ts
@@ -17,6 +17,8 @@ const manifest: PaperclipPluginManifestV1 = {
     "http.outbound",
     "secrets.read-ref",
     "agents.read",
+    "issues.read",
+    "issue.comments.read",
   ],
   entrypoints: {
     worker: "./dist/worker.js",
--- a/hindsight-integrations/paperclip/src/worker.ts
+++ b/hindsight-integrations/paperclip/src/worker.ts
@@ -173,6 +173,14 @@ const plugin = definePlugin({
           /* ignore */
         }
       }
+      
+      if (!bankAgentId) {
+        ctx.logger.info("Skipping retain — no agent attribution available", {
+          commentId,
+          issueId,
+        });
+        return;
+      }
 
       try {
         const apiKey = await resolveApiKey(ctx, config);
--- a/hindsight-integrations/paperclip/src/worker.ts
+++ b/hindsight-integrations/paperclip/src/worker.ts
@@ -4,8 +4,13 @@
  * Gives Paperclip agents persistent long-term memory via Hindsight.
  *
  * Lifecycle:
- *   agent.run.started  → recall relevant memories, store in plugin state for the run
- *   agent.run.finished → retain agent output to Hindsight (if autoRetain is enabled)
+ *   agent.run.started      → fetch the run's issue, recall relevant memories,
+ *                            cache them in plugin state for the run.
+ *   issue.comment.created  → retain the full comment body to Hindsight (this is
+ *                            the durable signal — it covers both user comments
+ *                            and agent-authored comments).
+ *   agent.run.finished     → no-op (Paperclip's lifecycle payload does not
+ *                            carry run output; subscription kept for future use).
  *
  * Agent tools (callable mid-run):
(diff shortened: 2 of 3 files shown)


benfrank241:
thanks for looking into this!

ran the tests locally — 17/17 pass and the build is green, but `npm run typecheck` fails with two errors in `worker.ts` at lines 180 and 182:

```
Type 'string | null' is not assignable to type 'string'.
```

in the new comment-retain path, `bankAgentId` can end up `null` (comment isn't agent-authored *and* the issue has no `assigneeAgentId`). that null gets passed into `deriveBankId({ agentId: string })`, and at runtime `parts.push(null)` followed by `parts.join("::")` produces a bank id like `paperclip::co123::` (trailing empty segment, since `Array.join` coerces null to `""`). every unattributable comment silently lands in that one shared bank — same silent-failure pattern this PR was meant to fix.

cleanest fix is to skip the retain when there's no attribution. patch:

```ts
      // Bank attribution: comments authored by an agent belong in that agent's
      // bank; user/system comments fall back to the issue's assignee.
      let bankAgentId: string | null = payloadAgentId;
      if (!bankAgentId) {
        try {
          const issue = await ctx.issues.get(issueId, companyId);
          bankAgentId = issue?.assigneeAgentId ?? null;
        } catch {
          /* ignore */
        }
      }

+     if (!bankAgentId) {
+       ctx.logger.info("Skipping retain — no agent attribution available", {
+         commentId,
+         issueId,
+       });
+       return;
+     }
+
      try {
        const apiKey = await resolveApiKey(ctx, config);
        const client = new HindsightClient(config.hindsightApiUrl, apiKey);
        const bankId = deriveBankId({ companyId, agentId: bankAgentId }, config);
        await client.retain(bankId, body, commentId, {
          agentId: bankAgentId,
          ...
        });
```

that narrows the type to `string` for both call sites, so the typecheck errors clear without a cast.

a test in the same style as the existing `issue.comment.created` block:

```ts
it("skips retain when comment has no agent author and issue has no assignee", async () => {
  const harness = buildHarness();
  await setupPlugin(harness);
  const issue = await seedIssue(harness, { companyId: "co-1", title: "x" });
  const comment = await harness.ctx.issues.createComment(issue.id, "User message", "co-1");

  await harness.emit(
    "issue.comment.created",
    { commentId: comment.id, agentId: null },
    { companyId: "co-1", entityId: issue.id }
  );

  expect(fetchMock).not.toHaveBeenCalled();
});
```

happy to merge once that's in. could you also flip `maintainer_can_modify` on for next time? small fixes like this would be quicker if i could push directly.

amirhmoradi:
@benfrank241 , thanks for the review. all is fixed as per your request.
Noted for the maintainer_can_modify flag next time 👍 .

benfrank241:
lgtm! thanks!

benfrank241 (approved):

PR #1561, opened 2026-05-09
feat(paperclip): add user-scoped memory isolation for GDPR compliance

amirhmoradi:
- Add userId parameter to BankContext for per-user memory isolation
- Extend bankGranularity to support 'user' option for isolating memories by user
- Add resolveUserIdFromActiveIssue() to extract user email from active issue
- Add 'issues.read' capability to permissions for accessing active issue data
- Update all bank ID derivations to include user ID when granularity includes 'user'
- Update manifest description to mention GDPR compliance when user option is enabled

This allows agents to maintain user-scoped memory in multi-user scenarios,
ensuring that personal/scoped conversations don't leak between users.

### DIFF
--- a/hindsight-integrations/paperclip/src/bank.ts
+++ b/hindsight-integrations/paperclip/src/bank.ts
@@ -3,18 +3,20 @@
  *
  * Default format: "paperclip::{companyId}::{agentId}"
  *
- * bankGranularity: ['company']        → "paperclip::{companyId}"
- * bankGranularity: ['agent']          → "paperclip::{agentId}"
- * bankGranularity: ['company','agent'] → "paperclip::{companyId}::{agentId}"
+ * bankGranularity: ['company']               → "paperclip::{companyId}"
+ * bankGranularity: ['agent']                 → "paperclip::{agentId}"
+ * bankGranularity: ['company','agent']       → "paperclip::{companyId}::{agentId}"
+ * bankGranularity: ['company','agent','user'] → "paperclip::{companyId}::{agentId}::user::{userId}"
  */
 
 export interface BankContext {
   companyId: string;
   agentId: string;
+  userId?: string;
 }
 
 export interface BankConfig {
-  bankGranularity?: Array<"company" | "agent">;
+  bankGranularity?: Array<"company" | "agent" | "user">;
 }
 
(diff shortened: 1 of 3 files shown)


benfrank241:
Thanks @amirhmoradi, direction is right, per-user isolation in paperclip is a real gap worth fixing. but i pulled the branch and npm run typecheck fails with 4 errors:
src/worker.ts(89,57):  TS2345: Property 'getActive' is missing in type 'PluginIssuesClient'
src/worker.ts(133,57): same
src/worker.ts(172,59): same
src/worker.ts(224,59): same

ctx.issues.getActive() doesn't exist in @paperclipai/plugin-sdk. I checked the pinned ^2026.403.0, the latest stable 2026.428.0, and the latest canary 2026.511.0-canary.8.

benfrank241:
thanks for updating. one quick check: when an agent has more than one in-progress issue, this picks an arbitrary one (whatever `list` orders first). is that acceptable for paperclip's model, or should the helper pin to a specific id?

benfrank241:
@amirhmoradi thanks for the original work here — per-user isolation in paperclip is a real feature we wanted, and your PR is what kicked it off.

closing this one because it's picked up merge conflicts and the design question about multi-issue ordering (my comment from may 12) was never resolved. we've taken the concept and landed it in #1761 with a different approach: instead of `ctx.issues.list()` (which picks an arbitrary issue), we extract the user from the specific issue already in context for each event handler.

you're credited in both the commit and the PR description. appreciate the contribution — if you spot anything off in #1761, feel free to comment there.

amirhmoradi:
> @amirhmoradi thanks for the original work here — per-user isolation in paperclip is a real feature we wanted, and your PR is what kicked it off.
> 
> closing this one because it's picked up merge conflicts and the design question about multi-issue ordering (my comment from may 12) was never resolved. we've taken the concept and landed it in #1761 with a different approach: instead of `ctx.issues.list()` (which picks an arbitrary issue), we extract the user from the specific issue already in context for each event handler.
> 
> you're credited in both the commit and the PR description. appreciate the contribution — if you spot anything off in #1761, feel free to comment there.

Thanks a lot for your message and the push towards a better implementation. 🙏

PR #1589, opened 2026-05-12
paperclip: add sharedBankName config for cross-agent memory sharing

SeBru1:
## Summary

Adds an optional `sharedBankName` string to the **`hindsight-integrations/paperclip`** plugin's instance config. When set, every agent in the company reads/writes the same Hindsight bank — overriding `bankGranularity` and the default `paperclip::{companyId}::{agentId}` derivation.

This is a small, additive, backwards-compatible change. No behavior changes for any existing install that doesn't set the new field.

## Motivation

The current `bankGranularity: [\"company\", \"agent\"]` default isolates each agent's memory. That's the right default — memory shouldn't leak across unrelated agents.

But for **multi-agent cohorts that need to collaborate on a shared pool of project context** (e.g. a CEO/CTO/Staff/QA/Researcher team working a single backlog), there's no first-class way to point them at one bank.

Concrete use case: I run a 5-agent Paperclip company where each agent must see decisions, lessons, and project state created by the others. With current granularity, each agent talks to its own bank — `hindsight_recall` returns nothing the agent didn't write itself, and cross-agent compound knowledge never accrues.

Workarounds I considered:
- `bankGranularity: [\"company\"]` collapses to a per-company bank — but the bank ID is still `paperclip::{companyId}`, which is fine for a single deployment but not when you want a human-readable bank name shared across companies or instances.
- A wrapper plugin re-implementing the same logic against a fixed bank — significant maintenance cost for what's essentially a one-line override.

The cleanest fix is a tiny optional config override. That's this PR.

## What changed

- **`src/bank.ts`**: extend `BankConfig` with optional `sharedBankName`; `deriveBankId` returns the trimmed value verbatim when set, otherwise falls through to the existing granularity logic unchanged.
- **`src/manifest.ts`**: add `sharedBankName` to `instanceConfigSchema` so it's configurable from Settings → Plugins → Hindsight Memory.
- **`src/worker.ts`**: add the field to the internal `PluginConfig` type.
- **`tests/plugin.spec.ts`**: 4 new tests — 3 unit tests for the override semantics (trimming, empty/whitespace fallthrough, normal override) + 1 integration test that verifies recall routes to the shared bank URL.
- **`README.md`**: document the field and add a *Shared bank for cross-agent collaboration* section.

## Backwards compatibility

The field is optional. Existing installs without `sharedBankName` keep the existing `paperclip::{companyId}::{agentId}` derivation untouched.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm test` — 19/19 tests pass (15 existing + 4 new)
- [x] `npm run build` — clean esbuild output
- [x] Verified locally against a 5-agent company: with `sharedBankName: \"spool-farm\"`, recall/retain from all 5 agents now route to the single `spool-farm` bank.

## Notes for review

Happy to adjust naming (`sharedBankName` vs `bankName` vs `bankOverride`), schema placement, or split into smaller commits if preferred. The full change is +89/-8 across 5 files.

### DIFF
--- a/hindsight-integrations/paperclip/README.md
+++ b/hindsight-integrations/paperclip/README.md
@@ -33,13 +33,14 @@ Or [Hindsight Cloud](https://ui.hindsight.vectorize.io/signup) — no self-hosti
 
 ## Configuration
 
-| Field                | Default                 | Description                                                    |
-| -------------------- | ----------------------- | -------------------------------------------------------------- |
-| `hindsightApiUrl`    | `http://localhost:8888` | Hindsight server URL                                           |
-| `hindsightApiKeyRef` | —                       | Paperclip secret name holding Hindsight Cloud API key          |
-| `bankGranularity`    | `["company", "agent"]`  | Memory isolation: per company+agent, per company, or per agent |
-| `recallBudget`       | `mid`                   | `low` = fastest, `mid` = balanced, `high` = most thorough      |
-| `autoRetain`         | `true`                  | Automatically retain run output after every run                |
+| Field                | Default                 | Description                                                            |
+| -------------------- | ----------------------- | ---------------------------------------------------------------------- |
+| `hindsightApiUrl`    | `http://localhost:8888` | Hindsight server URL                                                   |
+| `hindsightApiKeyRef` | —                       | Paperclip secret name holding Hindsight Cloud API key                  |
+| `sharedBankName`     | —                       | Fixed bank name override — all agents share this single bank when set  |
+| `bankGranularity`    | `["company", "agent"]`  | Memory isolation when `sharedBankName` is unset                        |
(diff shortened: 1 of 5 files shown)


benfrank241:
@SeBru1 thanks! Clean PR and the cross-agent collaboration case is legit. LGTM

benfrank241:
@SeBru1 — could you rename sharedBankName → bankId and add a dynamicBankId: boolean to match the pattern the other plugins use? same behavior, but a user moving between paperclip/ claude-code / openclaw would see the same field names mean the same thing. otherwise the implementation looks clean.

benfrank241:
@SeBru1 thanks for this — the shared-bank use case is exactly right, and the implementation was clean. we've folded it into #1761 with the naming change i mentioned (sharedBankName → bankId/dynamicBankId to match the convention across openclaw, claude-code, and opencode).

you're credited in both the commit and the PR description. if you spot anything off in #1761, feel free to comment there.

PR #1607, opened 2026-05-13
feat(config): add per-scope HINDSIGHT_API_<SCOPE>_LLM_EXTRA_BODY

connorblack:
## Summary

Hindsight currently exposes only a single global `HINDSIGHT_API_LLM_EXTRA_BODY`. This forces an all-or-nothing choice for reasoning-model knobs like Qwen3's `chat_template_kwargs.enable_thinking` — useful for retain/consolidation (structured JSON extraction; thinking adds latency without quality gain) but breaks reflect (an agentic tool loop that needs reasoning to decide *whether* to call tools).

We hit this concretely with Qwen3.6-35B-A3B-FP8 served via vLLM: with `enable_thinking=false` set globally for consolidation throughput, every reflect call returned 0 tool calls and short-circuited to "I don't have information" without ever searching the bank.

## Change

Add per-scope env vars matching the existing per-scope provider/model/timeout pattern:

- `HINDSIGHT_API_RETAIN_LLM_EXTRA_BODY`
- `HINDSIGHT_API_REFLECT_LLM_EXTRA_BODY`
- `HINDSIGHT_API_CONSOLIDATION_LLM_EXTRA_BODY`

Each parses as JSON. When set, it overrides the global `HINDSIGHT_API_LLM_EXTRA_BODY` for that scope only. When unset, falls back to the global. When neither is set, no extra_body is sent (existing behavior).

Wiring is a single-line change at each `LLMConfig` instantiation in `MemoryEngine.__init__`:

```python
extra_body=config.<scope>_llm_extra_body or config.llm_extra_body,
```

## Test plan

- [x] 4 new tests in `tests/test_per_scope_llm_extra_body.py`:
  - unset → all fields None
  - per-scope env parses as JSON
  - global only → per-scope stays None (fallback happens at engine init, not parse time)
  - invalid JSON raises at parse time
- [x] All 13 existing `tests/test_per_operation_llm_config.py` tests still pass
- [x] `ruff check` clean on touched files

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -168,6 +168,7 @@ ENV_RETAIN_LLM_INITIAL_BACKOFF = "HINDSIGHT_API_RETAIN_LLM_INITIAL_BACKOFF"
 ENV_RETAIN_LLM_MAX_BACKOFF = "HINDSIGHT_API_RETAIN_LLM_MAX_BACKOFF"
 ENV_RETAIN_LLM_TIMEOUT = "HINDSIGHT_API_RETAIN_LLM_TIMEOUT"
 ENV_RETAIN_LLM_LITELLMROUTER_CONFIG = "HINDSIGHT_API_RETAIN_LLM_LITELLMROUTER_CONFIG"
+ENV_RETAIN_LLM_EXTRA_BODY = "HINDSIGHT_API_RETAIN_LLM_EXTRA_BODY"
 
 ENV_REFLECT_LLM_PROVIDER = "HINDSIGHT_API_REFLECT_LLM_PROVIDER"
 ENV_REFLECT_LLM_API_KEY = "HINDSIGHT_API_REFLECT_LLM_API_KEY"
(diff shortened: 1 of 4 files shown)


nicoloboschi (changes requested):
Two changes requested before merge:

### 1. Remove `HINDSIGHT_API_REFLECT_DISABLE_FORCED_TOOL_CHOICE` from this PR

The reflect/agent.py change is scope creep — the PR title and summary advertise only the per-scope `LLM_EXTRA_BODY` env vars, but `agent.py` quietly adds a second, unrelated env var with no tests and no documentation. Please remove it from this PR.

If we still want that escape hatch, it should land separately and:
- Be parsed once in `HindsightConfig.from_env()` with an `ENV_*` constant (matching the established `.lower() == "true"` pattern at `config.py:1664-1687`), not via raw `os.getenv()` inside the per-iteration hot path at `agent.py:579`.
- Be read off the engine's config rather than re-importing `os` in `agent.py` just for this.
- Have at least one test that monkeypatches the env var and asserts `iter_tool_choice == "auto"`.

### 2. Document the three new env vars

`hindsight-docs/docs/developer/configuration.md` already documents the global `HINDSIGHT_API_LLM_EXTRA_BODY` (line 190) and the full per-scope table for provider/model/timeout (lines 378–404). Please add the three new vars (`HINDSIGHT_API_{RETAIN,REFLECT,CONSOLIDATION}_LLM_EXTRA_BODY`) to that table following the existing "Falls back to `HINDSIGHT_API_LLM_EXTRA_BODY`" phrasing — CLAUDE.md requires docs updates for any new config flag.

The core extra-body change itself is clean and good to go once these two items are addressed.

nicoloboschi:
closing as stale

Issue #1566, opened 2026-05-10
Observation consolidation creates near-duplicate sibling observations instead of merging into one canonical observation


witt3rd:
# Observation consolidation creates near-duplicate sibling observations instead of merging into one canonical observation

## Summary

When the same canonical event/decision/statement is retained many times across sessions, the observation consolidation layer creates many near-identical sibling observations rather than merging new evidence into one canonical observation. Most resulting observations end up with `source_fact_ids = 1`, defeating the consolidation-as-evidence-aggregation contract.

## Environment

- `hindsight-api` 0.6.0, `local_embedded` mode, Postgres backend
- Bank: `hermes-forge` (~21k facts, 5,638 observations, ~5k pending_consolidation)
- `enable_observations: true`, `max_observations_per_scope: -1`, `consolidation_max_memories_per_round: 100`, `consolidation_llm_batch_size: 8`
- LLM: anthropic / claude-haiku-4-5

## Reproduction

Recall against an event retained many times across sessions:

```bash
curl -s -X POST "http://127.0.0.1:9177/v1/default/banks/hermes-forge/memories/recall" \
  -H "Content-Type: application/json" \
  -d '{"query":"Donald told Athena she is sovereign","budget":"mid","types":["observation"],"include":{"source_facts":{}}}' \
  | jq '.results | length, (.[] | {id, source_fact_count: (.source_fact_ids | length), text: (.text[0:80])})'
```

Result: **33 observation nodes** for one canonical event. Distribution of `source_fact_ids` per observation:

- A few well-merged: `147`, `61`, `14`, `6`, `5`, `3`, `2`, `2`
- The remaining ~25: `source_fact_ids = 1`

The expected shape: ~3–5 canonical observations covering structurally distinct claims about that event (e.g. "Donald named sovereignty," "sovereignty principle is foundational to Janus architecture," "Forge added sovereignty line to SOUL.md.j2"), each with many source facts attached.

## Suspected cause

Reading `observations_mission` as the only user-facing knob for consolidation semantics, plus the per-batch shape (`consolidation_llm_batch_size: 8`, `consolidation_max_memories_per_round: 100`), it looks like:

1. The consolidator processes facts in small batches.
2. Each batch's LLM call doesn't include the bank's existing observations as merge candidates (or includes too few).
3. Without nearest-neighbor existing observations in context, the LLM can't decide "merge with obs-X" vs "create new" — it defaults to creating new.
[...]

PR #1615, opened 2026-05-13
feat(consolidation): surface semantic similarity to the consolidation LLM

xdonu2x:
## Problem

Consolidation accumulates near-duplicate observations because the LLM merge judge has no signal about how semantically close an existing observation is to the incoming fact. Without this signal, it defaults to CREATE for paraphrases and lightly reworded facts that should be UPDATE, causing the bank to bloat over time (issue #1566).

### `MemoryFact.similarity` field (`response_models.py`)

Adds an optional `similarity: float | None` field to `MemoryFact`. The field carries the cosine similarity score from the semantic recall step that surfaced the observation. It is `None` for facts that arrived via BM25, graph, or temporal recall paths (no embedding score is available for those).

The value is already computed and stored in `ScoredResult.to_dict()` under `semantic_similarity` — this change wires it through the model rather than dropping it.

### Similarity forwarded to the HTTP API (`http.py`)

`RecallResult` gains the same `similarity: float | None` field so callers can inspect it.

### LLM prompt guidance (`prompts.py`)

Documents the `similarity` field in the system prompt with concrete thresholds:
- ≥ 0.85 → very likely the same facet, strongly prefer UPDATE
- ≥ 0.95 → almost always UPDATE unless structurally distinct

### Sort by similarity descending (`consolidator.py`)

`_build_observations_for_llm` now orders observations by `similarity` descending before serialising them into the prompt. Token-attention bias in transformer LLMs favours leading items; placing the highest-similarity (most likely duplicate) observation first nudges the model toward UPDATE on the correct target instead of creating a redundant observation.

## Why this helps

The LLM can already compare texts. Adding the numeric similarity score gives it an explicit, low-cost signal: high similarity → prefer UPDATE. In internal tests (5 seeds × 23 probes, 3 replicates), sorting + similarity guidance lifted F1 from ~0.22 to ~0.73 and recall from ~0.29 to ~0.90 on a paraphrase/dedup corpus.

## Tests

- `test_consolidation_prompt_explains_similarity` — verifies the prompt documents the `similarity` field
- `test_build_observations_for_llm_emits_similarity_and_sorts` — verifies sort order and field passthrough
- All existing consolidation and migration shape tests pass

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -226,6 +226,7 @@ class RecallResult(BaseModel):
     source_fact_ids: list[str] | None = (
         None  # IDs of source facts (observation type only, when source_facts is enabled)
     )
+    similarity: float | None = None  # Cosine similarity to the query (semantic recall only; None otherwise)
 
 
 class EntityObservationResponse(BaseModel):
@@ -3233,6 +3234,7 @@ def _register_routes(app: FastAPI):
                     chunk_id=fact.chunk_id,
                     tags=fact.tags,
                     source_fact_ids=fact.source_fact_ids,
+                    similarity=fact.similarity,
                 )
 
             recall_results = [_fact_to_result(fact) for fact in core_result.results]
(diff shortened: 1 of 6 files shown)


nicoloboschi:
thanks for this PR but I don't think it's the right solution to the problem. the LLM is capable to understand whether 2 facts should be merged or not based on the text, not on vector similarity

PR #1631, opened 2026-05-14
feat: weighted Reciprocal Rank Fusion for recall

dcbouius:
## Summary
- Adds per-strategy weights to RRF fusion, allowing tuning of how much each retrieval method (semantic, BM25, graph, temporal) influences the merged ranking
- Weights follow the hierarchical config model: server env vars → per-bank config → per-request override
- Default weights are all 1.0, preserving existing behavior exactly

## API
New optional `retrieval_weights` field on the recall request:
```json
{
  "query": "What happened at the meeting?",
  "retrieval_weights": {
    "semantic": 1.0,
    "bm25": 1.0,
    "graph": 2.0,
    "temporal": 1.0
  }
}
```

Keys: `semantic`, `bm25`, `graph`, `temporal`. Values are multipliers (1.0 = default, 2.0 = double influence, 0.0 = disabled). Omitted keys default to bank/server config.

## Configuration
Server-level defaults via env vars:
- `HINDSIGHT_API_RECALL_WEIGHT_SEMANTIC` (default: 1.0)
- `HINDSIGHT_API_RECALL_WEIGHT_BM25` (default: 1.0)
- `HINDSIGHT_API_RECALL_WEIGHT_GRAPH` (default: 1.0)
- `HINDSIGHT_API_RECALL_WEIGHT_TEMPORAL` (default: 1.0)

All four are in `_CONFIGURABLE_FIELDS`, so they can be overridden per-bank via the config API.

## Changes
- `fusion.py` — `reciprocal_rank_fusion()` accepts optional `weights` dict; formula becomes `w_i / (k + rank)`
- `config.py` — 4 new hierarchical config fields with env var + default support
- `http.py` — `RecallRequest.retrieval_weights` field
- `memory_engine.py` — Merges config + request weights, passes to RRF
- `interface.py` — `retrieval_weights` param on abstract `recall_async`
- Control plane proxy + client types updated

## Test plan
- [x] 10 unit tests covering: unweighted default, None/1.0 equivalence, high weight boosting, zero weight disabling, partial weights, temporal support, ranking reversal, source rank preservation
- [ ] CI passes

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -171,6 +171,12 @@ class RecallRequest(BaseModel):
         description="Compound tag filter using boolean groups. Groups in the list are AND-ed. "
         "Each group is a leaf {tags, match} or compound {and: [...]}, {or: [...]}, {not: ...}.",
     )
+    retrieval_weights: dict[str, float] | None = Field(
+        default=None,
+        description="Per-strategy weights for Reciprocal Rank Fusion. Keys: 'semantic', 'bm25', "
+        "'graph', 'temporal'. Values are multipliers (1.0 = default, 2.0 = double influence, "
+        "0.0 = disabled). Omitted keys default to the bank/server configuration.",
+    )
 
     @field_validator("query")
     @classmethod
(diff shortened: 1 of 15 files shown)


cdbartholomew:
@dcbouius I am not sure this is the best solution to making graph results higher weight. We already have "boosters" for things like recency that work after the rerank. We can just add a graph boost. Its is simpler and guaranteed to move up the graph results. Doing it during the fusion increases the chances that the graph results will win, but the rerank could still decided otherwise.

nicoloboschi:
closing for now - waiting for proper discussion

PR #1649, opened 2026-05-18
docs: document observation-only recall limits

### LINKED ISSUE #1531: Document the accumulation window of recall_types=["observation"] workaround and consider optional mitigation
### Use Case

I'm running `hermes-agent` with Hindsight in `local_embedded` mode as the long-term memory backend for a conversational AI agent (primarily a WeChat IM bot talking to a single user in long-running sessions, plus a CLI channel for development tasks). Bank: `hermes`. Hindsight version: `0.6.0`. Mode: `local_embedded` with an embedded Postgres instance at `~/.pg0/instances/hindsight-embed-hermes/`.

The agent retains on every turn (`retain_every_n_turns=1`) and relies on `auto_recall=True` to inject cross-session memory into the LLM prompt. This is the exact workflow Hermes ships by default for long-running assistants.

### Problem Statement

After reading [`#826`](https://github.com/vectorize-io/hindsight/issues/826) (CLOSED `NOT_PLANNED`), [`#1309`](https://github.com/vectorize-io/hindsight/issues/1309) (CLOSED `NOT_PLANNED`), [`#1284`](https://github.com/vectorize-io/hindsight/issues/1284) (OPEN), and PR [`#1371`](https://github.com/vectorize-io/hindsight/pull/1371) (CLOSED, un-merged) carefully, I followed the workaround @nicoloboschi explicitly recommended in `#1284`:

> *"you can configure `recall_types=["observation"]` if you want to only use observations and avoid duplicates"*

I configured `recall_types: ["observation"]` in `~/.hermes/hindsight/config.json`, restarted the gateway, and verified the filter is active in the daemon log:

```
[RECALL hermes-76151-b79626] Complete: 62 facts (2365 tok), 0 chunks (0 tok),
  0 entities (0 tok) | observation=62 | 0.739s
```

`observation=62` confirms the filter is reaching the engine correctly.

**The workaround mitigates the problem but does not solve it.** After ~2 days of normal conversational use, the observation layer itself accumulates to the point where the filter no longer yields a clean recall. Concrete numbers from my live bank:

| fact_type | count | earliest | latest |
|---|---|---|---|
| `experience` | 100 | 2026-05-07 | 2026-05-08 |
| `observation` | 68 | 2026-05-07 | 2026-05-08 |
| `world` | 25 | 2026-05-07 | 2026-05-07 |
| **total** | **193** | — | — |

A single reasonably broad query matches **62 of the 68 observations** (91% hit rate). Manual inspection of the returned facts shows heavy semantic duplication — representative clusters from one recall:

- 8 facts about "skill update priority / strategy"
- 5 facts about "cleaning up 69 `old==new` entries"
- 5 facts about "hermes update overwriting l10n patches"
- 5 facts about "hindsight memory duplication" (meta: even the *record of this very problem* appears 5 times)
- 3 facts about "hindsight consolidation not working"
- 3 facts about "hindsight node counts" (two literally identical)
- 2 facts about "hindsight URL correction" (near-verbatim)
- …and a long tail of similar patterns

Signal-to-noise ratio is roughly **1:3** — about two-thirds of the returned tokens restate a few facts that already got through.

Why this matters in a real deployment:

1. **Low information density at non-trivial absolute size.** 2,365 tokens on its own is not a context-window threat, but signal-to-noise is ~1:3 — most of the tokens restate the same ~8 facts. Because recall token count grows roughly linearly with observation count while *distinct* facts in the result plateau quickly, the per-token information yield of injected memory trends toward zero. Doubling the observation bank does not meaningfully increase what the LLM learns about the user; it just amplifies the redundancy. *(Edit: an earlier version of this issue mis-attributed this impact to Hermes' `memory_char_limit` budget; that setting governs Hermes' built-in file-based memory — `MEMORY.md`/`USER.md` in `tools/memory_tool.py` — not Hindsight's recall injection, which is a separate path. Corrected for accuracy.)*

2. **Implicit re-weighting of repeated facts.** When the same fact surfaces 5–8 times in a single recall, the downstream LLM effectively receives a redundancy-as-salience signal. In practice the assistant's replies start to echo and over-emphasize recent repeated observations, and novel context processed by the LLM competes for less attention. This is not a Hindsight bug per se, but it is a direct downstream consequence of how the recall output shape interacts with LLM inference.

3. **The workaround's effective window is measured in days, not weeks.** The `recall_types=["observation"]` advice is discoverable only in a maintainer comment on a still-open issue. Users who find it, follow it, and ship the agent to production will hit this wall within a week of normal use.

The core ask is **honesty about the workaround's boundary, and an optional path forward for users who've hit it**:

- If the project's considered position is "observation accumulation is expected; users should manually prune periodically," that's a legitimate design choice — but it needs to be **discoverable**. Right now the only place this trade-off is explained is inside issue comments.
- If there's any room to re-evaluate a narrow, opt-in mitigation (e.g. a similarity threshold applied only to same-`document_id` observations at retain time, or an opt-in `dedup_by_embedding` at recall), it would meaningfully extend the workaround's effective window for long-running agents without imposing a default that changes existing users' behavior.

I fully understand `#826` and `#1309` were closed `NOT_PLANNED` and I am not asking to re-litigate them. I'm specifically asking about the gap between **the recommended workaround** and **the observed reality** of a conversational agent running at default `retain_every_n_turns=1`.

### Proposed Solution

Listed from lightest to heaviest:

**Option A — Documentation (preferred starting point).** Add a short section to the best-practices / docs explicitly naming this pattern: "`recall_types=['observation']` is a good workaround, but expect observation-layer accumulation at the rate of roughly N per session; plan to periodically clear observations older than M days, or set `retain_every_n_turns > 1` to slow growth." This turns a hidden cliff into a visible one.

**Option B — Opt-in retain-time similarity check, scoped narrowly.** Not "global semantic dedup at retain time" (which `#826` rejected), but something narrower — e.g. a per-bank optional `observation_dedup_threshold` that only fires *within the same `document_id`* or *within a recent time window*. This preserves the existing default, is behind an explicit user opt-in, and addresses the highest-volume duplication class that hermes-style conversational agents produce.

**Option C — Opt-in `dedup_by_embedding: true` parameter on recall.** Filter near-duplicates on the recall output path rather than at storage. Zero impact on stored data. I recognize this overlaps with `#1309`; mentioning it only for completeness of alternatives.

### Alternatives Considered

- **Periodic bank clear** — I've done this once already; the data shown above rebuilt from zero in 2 days. Without a change in retain/consolidation behavior, this is a treadmill.
- **Increase `retain_every_n_turns` to 3–5** — reduces growth rate but doesn't address existing accumulation and delays the ingestion of genuinely new facts by 3–5 turns, which users notice.
- **Client-side dedup in the Hermes plugin** — I've opened a parallel issue at [`NousResearch/hermes-agent#21698`](https://github.com/NousResearch/hermes-agent/issues/21698) asking Hermes to surface this trade-off in its setup wizard / docs. That addresses discoverability but not the underlying growth dynamics; the two issues are complementary, not redundant.

### Priority

**Important - affects my workflow**

### Additional Context

**Verified against Hindsight `main` at filing time**: the behavior documented above matches the released code in `0.6.0` (my version) and the version documentation on `hindsight.vectorize.io/api-reference` (no `dedup` / `mmr` / `diversity` / `similarity_threshold` parameters exposed on `recall` or `retain`).

**Related hindsight issues read before filing**:
- [`#826`](https://github.com/vectorize-io/hindsight/issues/826) — CLOSED `NOT_PLANNED`: Deduplicate near-identical content at retain time
- [`#1309`](https://github.com/vectorize-io/hindsight/issues/1309) — CLOSED `NOT_PLANNED`: Add MMR diversification + temporal decay to recall scoring
- [`#1284`](https://github.com/vectorize-io/hindsight/issues/1284) — OPEN: Experience entries accumulate (the source of the `recall_types=["observation"]` workaround)
- [`#1371`](https://github.com/vectorize-io/hindsight/pull/1371) — CLOSED un-merged: fix(consolidation): deduplicate observations by semantic similarity
- [`#360`](https://github.com/vectorize-io/hindsight/issues/360) — CLOSED: Memory feedback loop (mitigated on Hermes side via `_strip_injected_memories`)
- [`#1340`](https://github.com/vectorize-io/hindsight/pull/1340) — MERGED: `ON CONFLICT` dedup for `observation_sources` insert — this *is* in my 0.6.0, and is helpful, but protects only the source-mapping table; two different sources producing semantically near-identical observations still land as separate `memory_units` rows, which is the primary driver of the numbers above.

**Related downstream issue**: [`NousResearch/hermes-agent#21698`](https://github.com/NousResearch/hermes-agent/issues/21698) — asks the Hermes plugin to at least make the workaround discoverable during `hermes gateway setup`.

**Environment**:
- `hindsight-*` 0.6.0 (local_embedded)
- `hermes-agent` v0.12.0 (216 commits behind `main`, plugin file tips verified against tip-of-main — defaults unchanged)
- Embedded Postgres 18.1.0 at `~/.pg0/installation/`
- LLM: Anthropic `claude-haiku-4-5` via a provider proxy (for consolidation); the recall/embedding path doesn't involve the LLM.

Thanks for reading this carefully. I've tried to come with data rather than abstract complaints, and I'm genuinely grateful for the thought that's gone into the existing scoring pipeline. I'd just like to find a way that long-running conversational agents can stay on a reasonable signal-to-noise curve without requiring the user to periodically run manual cleanup.

### Checklist

- [x] I would be willing to contribute this feature (starting with Option A — a docs PR — if the framing resonates; open to B/C if you'd prefer a code-level change)

xmh1011:
fix #1531

### DIFF
--- a/hindsight-docs/docs/developer/api/recall.mdx
+++ b/hindsight-docs/docs/developer/api/recall.mdx
@@ -82,6 +82,12 @@ Each type runs the full four-strategy retrieval pipeline independently, so narro
 Observations are deduplicated, evidence-grounded beliefs consolidated from multiple facts — preferences, recurring patterns, and durable learnings the memory bank has built up. Each observation references its supporting memories (with exact quotes) and carries a computed freshness trend, and is refined rather than overwritten when new evidence arrives. They are created and maintained automatically in the background after retain operations.
 :::
 
+#### Observation-only recall in long-running agents
+
+Using `types=["observation"]` (or an integration setting such as `recall_types=["observation"]`) is a good way to avoid injecting every raw fact into an agent prompt. It is not a retention or output-deduplication policy. If an agent retains every turn for days, broad queries can still match many overlapping observations because the observation layer continues to accumulate durable patterns.
+
+For long-running conversational agents, use stable [`document_id`](./retain#document_id) values, retain less aggressively when your integration allows it, keep `max_tokens` tight, and scope recall with tags. If a bank has already accumulated noisy or obsolete observations, delete the noisy source documents or memories first, then [reset observations](../observations#resetting-all-observations) so consolidation rebuilds from the remaining evidence. To bound future growth, consider `max_observations_per_scope` in the bank's observations configuration.
+
 ### budget
 
 Controls retrieval depth and breadth. Accepted values are `low`, `mid` (default), and `high`. Use `low` for fast simple lookups, `mid` for balanced everyday queries, and `high` when you need to find indirect connections or exhaustive coverage.
--- a/hindsight-docs/src/pages/best-practices.mdx
+++ b/hindsight-docs/src/pages/best-practices.mdx
@@ -400,6 +400,8 @@ recall(
 | `["observation"]` | Consolidated patterns only — faster for high-level questions |
 | `["world", "experience"]` | Raw facts only — for ground-truth or citation-sensitive queries |
 
+**Long-running agents:** Observation-only recall is useful for reducing prompt noise, but it does not stop observations from accumulating. Agents that retain every conversation turn can still build overlapping observations over days of normal use, especially for broad user-profile or project-status queries. Keep a stable [`document_id`](/developer/api/retain#document_id), lower retain frequency when the integration allows it, use tags and `max_tokens` to narrow recall, and periodically delete obsolete source documents before [resetting observations](/developer/observations#resetting-all-observations) if the bank has become noisy. Use `max_observations_per_scope` when you need a hard cap per observation scope.
+
 ---
 
 ### Filtering by Memory Shape with Entity Labels
--- a/hindsight-docs/versioned_docs/version-0.6/developer/api/recall.mdx
+++ b/hindsight-docs/versioned_docs/version-0.6/developer/api/recall.mdx
@@ -82,6 +82,12 @@ Each type runs the full four-strategy retrieval pipeline independently, so narro
 Observations are deduplicated, evidence-grounded beliefs consolidated from multiple facts — preferences, recurring patterns, and durable learnings the memory bank has built up. Each observation references its supporting memories (with exact quotes) and carries a computed freshness trend, and is refined rather than overwritten when new evidence arrives. They are created and maintained automatically in the background after retain operations.
 :::
 
+#### Observation-only recall in long-running agents
+
+Using `types=["observation"]` (or an integration setting such as `recall_types=["observation"]`) is a good way to avoid injecting every raw fact into an agent prompt. It is not a retention or output-deduplication policy. If an agent retains every turn for days, broad queries can still match many overlapping observations because the observation layer continues to accumulate durable patterns.
+
+For long-running conversational agents, use stable [`document_id`](./retain#document_id) values, retain less aggressively when your integration allows it, keep `max_tokens` tight, and scope recall with tags. If a bank has already accumulated noisy or obsolete observations, delete the noisy source documents or memories first, then [reset observations](../observations#resetting-all-observations) so consolidation rebuilds from the remaining evidence. To bound future growth, consider `max_observations_per_scope` in the bank's observations configuration.
+
 ### budget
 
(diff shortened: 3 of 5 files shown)


benfrank241:
thanks for the PR we're going to hold off on this for the time being.

PR #1653, opened 2026-05-18
fix(api): strip Gemma thought tags

### LINKED ISSUE #1644: Gemma 4 <thought> tags cause JSON parse error in openai_compatible_llm
### Bug Description

When using Gemma 4 models (gemma-4-26b-a4b-it, gemma-4-31b-it) via the OpenAI-compatible LLM provider, responses contain `<thought>` reasoning blocks that are not stripped before JSON parsing. The `call()` method handles `<think>`, `<thinking>`, `<reasoning>`, and `|startthink|...|endthink|` but is missing `<thought>`, causing `json.loads()` to fail. This is most visible during startup verification where no `response_format` is enforced.

**Fix:** Add `re.sub(r"<thought>.*?</thought>", "", content, flags=re.DOTALL)` alongside the other tag-stripping regexes at line ~568 in openai_compatible_llm.py.

### Steps to Reproduce

1. Set env: HINDSIGHT_API_LLM_MODEL=models/gemma-4-26b-a4b-it with OpenAI-compatible endpoint
2. Start Hindsight
3. Startup verification sends "Say 'ok'" -> Gemma 4 responds with <thought> tags -> JSON parse fails
4. Any hindsight_retain() call faces the same issue

### Expected Behavior

Hindsight should strip `<thought>` tags from LLM responses before JSON parsing, just as it does for `<think>`, `<thinking>`, and `<reasoning>` tags.

### Actual Behavior

Hindsight logs: "JSON parse error from LLM response (attempt 1/3): Expecting value" followed by retry exhaustion. Startup verification fails. Retain operations fail silently.

### Version

0.6.1

### LLM Provider

None

xmh1011:
fix #1644

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -561,10 +561,11 @@ class OpenAICompatibleLLM(LLMInterface):
                     )
 
                     # Strip reasoning model thinking tags
-                    # Supports: <think>, <thinking>, <reasoning>, |startthink|/|endthink|
+                    # Supports: <think>, <thinking>, <thought>, <reasoning>, |startthink|/|endthink|
                     original_len = len(content)
                     content = re.sub(r"<think>.*?</think>", "", content, flags=re.DOTALL)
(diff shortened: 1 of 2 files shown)


benfrank241 (approved):
lgtm

PR #1654, opened 2026-05-18
fix(api): lazy load reflect tiktoken encoding

### LINKED ISSUE #1595: tiktoken get_encoding('cl100k_base') at module import causes API crash on any DNS hiccup
## Summary

`hindsight_api/engine/reflect/prompts.py` calls `tiktoken.get_encoding("cl100k_base")` at **module import time** (line 15). The first invocation triggers a network download of `cl100k_base.tiktoken` from `openaipublic.blob.core.windows.net`. Any DNS hiccup or transient network failure at startup causes both `hindsight-api` and `hindsight-worker` to crash within ~5 seconds during their own module imports. The process exits with code 1; the user sees no warning at config-time that the API has a network dependency at startup.

This makes Hindsight's cold-start fragile to brief network blips, and prevents truly air-gapped deployments unless the user discovers the issue and sets `TIKTOKEN_CACHE_DIR` to a pre-populated directory.

## Severity

P1 — when the network blip occurs, both `hindsight-api` and `hindsight-worker` die together on import. In a watchdog-restart loop, every retry hits the same network call and crashes again, leaving the system fully unavailable until a human pre-caches the file or the network heals.

## Environment

- hindsight: 0.4.0–0.5.6 (the offending file `engine/reflect/prompts.py` exists in all versions I inspected)
- Windows 11, Python 3.12
- LLM provider: vertexai (so tiktoken is **not** in the main inference path, yet still imported)
- No outgoing HTTPS proxy or restrictive firewall — DNS was transiently unavailable for ~30 seconds

## Reproduction

1. Start with a fresh pip install of `hindsight-api-slim` so `~/.cache/tiktoken` (or equivalent) has no `cl100k_base` file yet.
2. While starting `hindsight-api`, block DNS for `openaipublic.blob.core.windows.net` (or simulate a transient blip).
3. The api process exits with code 1, traceback ending in `requests.exceptions.ConnectionError: HTTPSConnectionPool(host='openaipublic.blob.core.windows.net', port=443): ... NameResolutionError(...)`.

Stack trace from our actual incident (2026-05-11 20:43:07 EDT, hindsight-api 0.5.6 on Windows; abbreviated):

```
File ".../hindsight_api/__init__.py", line 8, in <module>
    from .engine.cross_encoder import ...
File ".../hindsight_api/engine/__init__.py", line 14, in <module>
    from .memory_engine import ...
File ".../hindsight_api/engine/memory_engine.py", line 188, in <module>
    from .reflect import run_reflect_agent
File ".../hindsight_api/engine/reflect/__init__.py", line 10, in <module>
    from .agent import ReflectAgentResult, run_reflect_agent
File ".../hindsight_api/engine/reflect/agent.py", line 20, in <module>
    from .prompts import (...)
File ".../hindsight_api/engine/reflect/prompts.py", line 15, in <module>
    _TIKTOKEN_ENCODING = tiktoken.get_encoding("cl100k_base")
...
requests.exceptions.ConnectionError: HTTPSConnectionPool(
    host='openaipublic.blob.core.windows.net', port=443
): Max retries exceeded with url: /encodings/cl100k_base.tiktoken
(Caused by NameResolutionError(
    "HTTPSConnection(host='openaipublic.blob.core.windows.net', port=443):
     Failed to resolve 'openaipublic.blob.core.windows.net'
     ([Errno 11001] getaddrinfo failed)"
))
```

Worker dies with the identical stack 0.2s later.

## Workaround in our deployment

1. Pre-populate a stable cache dir once when network is good:
   ```python
   import os
   os.environ["TIKTOKEN_CACHE_DIR"] = r"C:\Users\<user>\HindSight\tiktoken-cache"
   import tiktoken
   tiktoken.get_encoding("cl100k_base")  # downloads to cache
   ```
2. Set `TIKTOKEN_CACHE_DIR` in the hindsight-api/worker environment before they spawn.

This works, but it's an undocumented requirement. Most users will only discover it after the first network-blip-induced outage.

## Suggested fixes (any of these)

1. **Lazy-load**: move `_TIKTOKEN_ENCODING = tiktoken.get_encoding(...)` from module top into a function-local that's called only when reflect-token-counting actually needs to run. Imports stay offline.
2. **Bundle**: include `cl100k_base.tiktoken` (1.6 MB) in the wheel/package. Set `TIKTOKEN_CACHE_DIR` to that bundled location at runtime.
3. **Document**: prominently document `TIKTOKEN_CACHE_DIR` as a startup requirement for air-gapped or unreliable-network deployments, and emit a clear error at startup ("Hindsight needs `cl100k_base.tiktoken`; download to `<cache>` or set `TIKTOKEN_CACHE_DIR`") rather than crashing inside `__init__.py` chain.

(1) is the cleanest — it eliminates the network dependency at startup entirely without changing the user-facing UX. (2) trades 1.6 MB of wheel size for zero-network startup.

## Not covered by #1515

[#1515](https://github.com/vectorize-io/hindsight/issues/1515) (Windows + China Network) is a great guide and mentions the HuggingFace `HF_ENDPOINT` mirror for embedding model downloads, but tiktoken's BPE download is a **separate code path** (different host, different package). The same class of "Hindsight needs the public internet for one specific file on first import" problem, but unaddressed.

Happy to send a PR if a lazy-load is the preferred direction.

xmh1011:
fix #1595

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -14,8 +14,6 @@ import re
 import time
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
 
-import tiktoken
-
 from .models import DirectiveInfo, LLMCall, ReflectAgentResult, TokenUsageSummary, ToolCall
 from .prompts import (
     _extract_directive_rules,
@@ -23,6 +21,7 @@ from .prompts import (
     build_final_system_prompt,
     build_system_prompt_for_tools,
 )
+from .tokenization import count_cl100k_tokens
 from .tools_schema import get_reflect_tools
 
 
@@ -266,25 +265,22 @@ OUTPUT:"""
         return None, 0, 0
 
 
-_TIKTOKEN_ENCODING = tiktoken.get_encoding("cl100k_base")
-
-
 def _count_messages_tokens(messages: list[dict[str, Any]]) -> int:
     """Estimate the token count of the messages list using cl100k_base encoding."""
     total = 0
     for msg in messages:
         content = msg.get("content") or ""
         if isinstance(content, str):
-            total += len(_TIKTOKEN_ENCODING.encode(content))
+            total += count_cl100k_tokens(content)
         elif isinstance(content, list):
             for part in content:
                 if isinstance(part, dict) and isinstance(part.get("text"), str):
-                    total += len(_TIKTOKEN_ENCODING.encode(part["text"]))
+                    total += count_cl100k_tokens(part["text"])
         # Tool call arguments and results also count
         for tc in msg.get("tool_calls") or []:
             if isinstance(tc, dict):
                 func = tc.get("function", {})
-                total += len(_TIKTOKEN_ENCODING.encode(func.get("arguments", "")))
+                total += count_cl100k_tokens(func.get("arguments", ""))
     return total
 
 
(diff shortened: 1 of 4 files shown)


benfrank241 (approved):
pulled and tested, thanks!

PR #1655, opened 2026-05-18
fix(api): preserve tag group or triggers

### LINKED ISSUE #1583: Bug: compound "or" in trigger.tag_groups silently stored as "and" (Pydantic union deserialization)
### Bug Description

When passing a compound `{"or": [...]}` expression in `trigger.tag_groups` during mental model create or update, the server accepts it (200 OK) but stores and returns it as `{"and": [...]}`. This makes it impossible to scope a mental model's refresh trigger to multiple non-overlapping tag sets with OR semantics.

### Steps to Reproduce

```bash
# PATCH a mental model with "or" in tag_groups
curl -X PATCH "http://127.0.0.1:8888/v1/default/banks/my-bank/mental-models/<mm_id>" \
  -d '{
    "trigger": {
      "tag_groups": [{"or": [
        {"tags": ["ns:a"], "match": "all_strict"},
        {"tags": ["ns:b"], "match": "all_strict"}
      ]}]
    }
  }'

# Read it back
curl "http://127.0.0.1:8888/v1/default/banks/my-bank/mental-models/<mm_id>"
```

### Expected Behavior

`trigger.tag_groups` returns:
```json
[{"or": [{"tags": ["ns:a"], "match": "all_strict"}, {"tags": ["ns:b"], "match": "all_strict"}]}]
```

### Actual Behavior

`trigger.tag_groups` returns:
```json
[{"and": [{"tags": ["ns:a"], "match": "all_strict"}, {"tags": ["ns:b"], "match": "all_strict"}]}]
```

### Version

hindsight-all 0.6.0 / hindsight_api_slim 0.5.6

### LLM Provider

None

xmh1011:
fix #1583

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/search/tags.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/tags.py
@@ -191,21 +191,21 @@ class TagGroupLeaf(BaseModel):
 class TagGroupAnd(BaseModel):
     """Compound AND group: all child filters must match."""
 
-    model_config = ConfigDict(populate_by_name=True)
+    model_config = ConfigDict(populate_by_name=True, serialize_by_alias=True)
     filters: list[TagGroup] = Field(alias="and")
 
 
 class TagGroupOr(BaseModel):
     """Compound OR group: at least one child filter must match."""
 
-    model_config = ConfigDict(populate_by_name=True)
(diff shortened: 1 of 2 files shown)


benfrank241 (approved):
lgtm thanks @xmh1011 !!!

PR #1664, opened 2026-05-20
fix(api): ignore null bank config overrides

### LINKED ISSUE #1619: Consolidation fails if observation config is set to "Server Default"
### Bug Description

When LLM Batch Size, Source Facts Max Tokens, Source Facts Max Tokens Per Observation, or Max Observations Per Scope is set to "Server Default" in the dashboard, consolidation fails. It seems that the values aren't defaulting to what is stated in the config documentation. I have to put in a value or else consolidation would fail. 

### Steps to Reproduce

1. Open the web dashboard
2. Navigate to "Bank Configuration" > "Configuration" > "Observations"
3. Set LLM Batch Size, Source Facts Max Tokens, Source Facts Max Tokens Per Observation, or Max Observations Per Scope to empty, so the textbox defaults to "Server Default"
4. Retain memory or initiate consolidation

### Expected Behavior

Consolidation to process normally. 

### Actual Behavior

Consolidation fails and produces the following error:
`May 14 02:20:27 vmi3152678 hindsight-launcher.sh[866814]: 2026-05-14 02:20:27,982 - ERROR - hindsight_api.engine.memory_engine - Task execution failed: consolidation, error: Failed to search memories (TypeError): TypeError("'>=' not supported between instances of 'NoneType' and 'int'")`

### Version

v0.6.1

### LLM Provider

None

xmh1011:
fix #1619


benfrank241 (approved):
@xmh1011 thanks! lgtm

PR #1666, opened 2026-05-20
fix(worker): handle stale pending schema routines

### LINKED ISSUE #1555: Worker poller skips all schemas when using PL/pgSQL schemas_with_pending_work() with default public schema
### Bug Description

Worker poller skips all schemas when using PL/pgSQL schemas_with_pending_work() with default public schema
Note: The author is not a technical developer. The entire debugging process was conducted by reading open-source documentation and collaborating with an AI assistant (Hermes Agent). The root cause analysis and proposed fix were provided and validated by the AI assistant.

Environment
Version: v0.6.1 (image hindsight-api-slim)
Backend: PostgreSQL (self-hosted)
Deployment: Single worker container, default public schema, single-tenant
Installation: Docker, image ghcr.io/vectorize-io/hindsight:latest
Problem Description
The worker poller starts normally and reports available slots, but never claims any tasks. The pending task queue grows indefinitely while completed tasks remain unchanged.

Relevant log excerpt:

Worker 794a5e6c6647 starting polling loop (max_slots=10, reservations=[consolidation=2], shared_pool=8)
No "claimed X tasks" line ever appears.

Root Cause Analysis
A schema name representation mismatch exists between the Python code path and the SQL code path.

In WorkerPoller._get_schemas() (poller.py):

tenants = await self._tenant_extension.list_tenants()
# list_tenants() returns schema='public' (string)
# _get_schemas() normalizes the default public schema to None
return [t.schema if t.schema != DEFAULT_DATABASE_SCHEMA else None for t in tenants]
# → returns [None] for the default public schema
In WorkerPoller._scan_active_schemas() (server-side path, poller.py):

if await self._optional_routines.is_installed(conn, "schemas_with_pending_work"):
    rows = await conn.fetch("SELECT * FROM public.schemas_with_pending_work()")
    return {r[0] for r in rows}  # ← returns {'public'} (string)
Then, inside claim_batch():

active_schemas = await self._scan_active_schemas(schemas)
# schemas = [None], active_schemas = {'public'}
schemas_to_query = {s for s in schemas if s in active_schemas}
# → schemas_to_query = {} (empty because None is not in {'public'})
The Pass-1 fairness loop iterates over schemas_to_query, which is always empty. Consequently, no tasks are ever claimed.

Fix
Normalize the 'public' string to None in the server-side path of _scan_active_schemas():

# poller.py, _scan_active_schemas()
if await self._optional_routines.is_installed(conn, "schemas_with_pending_work"):
    rows = await conn.fetch("SELECT * FROM public.schemas_with_pending_work()")
    # Normalize the string 'public' to None to match _get_schemas() convention
    return {None if r[0] == 'public' else r[0] for r in rows}
No database-side changes are required.

Verification
After applying the fix, the following log lines confirm correct behaviour:

active_schemas: {None}        # previously {'public'}
rotated=[(0, None)]          # previously []
Worker 794a5e6c6647 claimed 1 tasks (0 consolidation): batch_retain:1
Pending tasks decreased from 76 to single digits within minutes.

Affected Scenarios
Single-tenant deployments that use the default public schema.
Deployments where the schemas_with_pending_work() PL/pgSQL function is installed. Deployments using the Python fallback path are not affected because that path uses the raw schema value as obtained from the query loop.
Multi-tenant deployments with custom schema names are not affected.
Related Issues
#991: Stale claimed_at not reclaimed after worker crash
#1006: consolidation_max_slots not enforced as independent pool

### Steps to Reproduce

1. Deploy Hindsight v0.6.1 with Docker, single worker container
    2. Use the default public schema (single-tenant)
    3. Ensure the schemas_with_pending_work() PL/pgSQL function is installed
    4. Observe that the worker starts normally:
       Worker <id> starting polling loop (max_slots=10, reservations=[consolidation=2], shared_pool=8)
    5. Leave it running — the pending queue grows indefinitely
    6. No "claimed X tasks" ever appears in the logs

### Expected Behavior

Worker poller claims and processes tasks from the default public schema

### Actual Behavior

Worker poller skips all schemas and never claims any tasks; pending queue grows indefinitely

### Version

_No response_

### LLM Provider

None

xmh1011:
fix #1555


benfrank241:
will hand this one off to @nicoloboschi since he authored https://github.com/vectorize-io/hindsight/pull/1503

nicoloboschi (approved):

PR #1668, opened 2026-05-20
fix(api): vchord ANN — use cosine opclass and dispatch tuning GUCs per backend

### LINKED ISSUE #1667: vchord backend: ANN queries cannot use vchordrq indexes due to opclass/operator mismatch (vector_l2_ops vs <=>)
## Summary

When Hindsight is deployed with the **vchord** vector backend (`HINDSIGHT_API_VECTOR_EXTENSION=vchord`), every ANN query in the codebase falls back to a sequential scan because the `vchordrq` indexes are built with `vector_l2_ops` but all application SQL uses the cosine-distance operator `<=>`. The opclass and the operator are not compatible in PostgreSQL, so the planner never picks the vector index, and the database CPU-walks every embedding in the partition for each query.

I observed this in production on a small dataset (~16k rows, 3072-dim embeddings, `tensorchord/cloudnative-vectorchord:18.3-1.1.1`): the primary Postgres pod sits at ~870m CPU because the retain pipeline's Phase 1 ANN link search (a `CROSS JOIN LATERAL` over `_ann_seeds`) is multiplying the slowdown by the seed count. Other backends (`pgvector`, `pgvectorscale`, `pg_diskann`) are unaffected because they map to `vector_cosine_ops` and match the `<=>` operator.

## Environment

- Hindsight API: `main` (verified on commit `bd86e7ea`, ahead by the recent merge bringing the repo current as of the report date)
- Vector backend: `vchord` (`tensorchord/cloudnative-vectorchord:18.3-1.1.1`, extension `vchord 1.1.1`, `vector 0.8.2`)
- PostgreSQL: 18
- Deployment: CloudNativePG cluster, single primary
- Embedding dimension: 3072 (OpenAI `text-embedding-3-large`-style, L2-normalized; see evidence below)

## Symptoms

- `cnpg-*-1` primary Pod CPU stays at ~870m steady-state during retain operations.
- `pg_stat_activity` shows the Phase 1 ANN query (`engine/retain/link_utils.py`) running for **50+ seconds** with empty `wait_event` (pure CPU, not I/O- or lock-bound).
- The same shape of query is used in recall, link expansion, and reflect paths — they share the latency penalty but are less visible because they are not wrapped in a `CROSS JOIN LATERAL`.

## Root Cause

`vchordrq` operator classes are strictly bound 1:1 to operators ([VectorChord docs — Operator Classes](https://docs.vectorchord.ai/vectorchord/usage/indexing.html)):

| opclass | usable operator |
|---|---|
| `vector_l2_ops` | `<->` only |
| `vector_cosine_ops` | `<=>` only |
| `vector_ip_ops` | `<#>` only |

Hindsight's vchord index mapping is in `hindsight-api-slim/hindsight_api/_vector_index.py`:

```python
INDEX_USING: dict[str, str] = {
    "pgvector":      "USING hnsw    (embedding vector_cosine_ops)",
    "pgvectorscale": "USING diskann (embedding vector_cosine_ops) WITH (num_neighbors = 50)",
    "pg_diskann":    "USING diskann (embedding vector_cosine_ops) WITH (max_neighbors = 50)",
    "vchord":        "USING vchordrq (embedding vector_l2_ops)",   # ← only L2
}
```

…but the ANN SQL across the engine uses cosine distance:

```python
# hindsight-api-slim/hindsight_api/engine/retain/link_utils.py:737-744
SELECT mu.id,
       1 - (mu.embedding <=> s.emb_text::vector) AS similarity
FROM memory_units mu
WHERE mu.bank_id = $1 AND mu.fact_type = $2 AND mu.embedding IS NOT NULL
ORDER BY mu.embedding <=> s.emb_text::vector
LIMIT $3
```

Because `<=>` is not in the operator family of `vector_l2_ops`, the planner cannot use the vchordrq index and falls back to a sequential scan with a per-row cosine computation, sorted by `top-N heapsort`.

## Evidence (production cluster, 2927-row partition)

`opencode::plugbear` / `fact_type = 'experience'` (2927 rows of 3072-dim embeddings):

**Current query (`<=>`, cosine):**

```
Limit  (... actual time=142.418..142.443 rows=50.00)
  Buffers: shared hit=71531 read=1
  -> Sort  (... actual time=142.415..142.426 rows=50.00)
       Sort Key: (mu.embedding <=> ...)
       Sort Method: top-N heapsort  Memory: 31kB
       -> Seq Scan on memory_units mu  (... actual time=2.251..141.358 rows=2927.00)
            Filter: (embedding IS NOT NULL AND bank_id = '...' AND fact_type = 'experience')
            Rows Removed by Filter: 13538
            Buffers: shared hit=71528 read=1
Planning Time: 15.001 ms
Execution Time: 143.828 ms
```

**Same data, equivalent query rewritten with `<->` (L2) — index is used:**

```
Limit  (... actual time=15.993..22.968 rows=50.00)
  Buffers: shared hit=770 read=520
  -> Index Scan using idx_mu_emb_expr_<hash> on memory_units mu
       Order By: (embedding <-> ...)
Planning Time: 16.354 ms
Execution Time: 25.646 ms
```

**~5.6× single-query difference**, and because the retain pipeline wraps this in `CROSS JOIN LATERAL` over a temp table of seeds, the penalty multiplies by seed count. A run with ~350 seeds takes ~50 s with the current opclass; the same workload on a matching opclass would finish in ~9 s.

Embedding norm statistics (verifies that L2 and cosine are monotonically equivalent for this corpus, so the fix is mathematically safe to discuss in either direction):

```
min_norm   avg_norm   max_norm   stddev_norm
0.999316   0.999998   1.000723   0.000197
```

## Affected code

All call sites using `<=>` (every one of these hits the same fallback when the backend is vchord):

- `engine/retain/link_utils.py:737, 742` — **Phase 1 ANN link creation (hottest path)**
- `engine/search/retrieval.py:410, 411, 415, 517, 532` — semantic recall
- `engine/search/link_expansion_retrieval.py:86, 91, 95` — graph link expansion
- `engine/reflect/tools.py:86, 89` — reflect tool retrieval
- `engine/sql/postgresql.py:157, 164, 168` — SQL helper used to compose other queries

## Secondary finding: `SET LOCAL hnsw.ef_search` is a no-op under vchord

`engine/retain/link_utils.py:711-713`:

```python
# Default 400 is tuned for recall precision but at 164k units each HNSW probe
# takes 94ms. ef_search=60 gives 2.7ms per probe (35x faster) ...
await conn.execute("SET LOCAL hnsw.ef_search = 60")
```

The `hnsw.ef_search` GUC is part of pgvector's HNSW index. It does not exist under vchord, so the `SET LOCAL` has no effect when `vchordrq` is in use. The intended "recall-vs-latency" tuning has therefore never been applied to vchord deployments. The equivalent vchord dials are `vchordrq.probes` and `vchordrq.epsilon` (see [Fallback Parameters](https://docs.vectorchord.ai/vectorchord/usage/fallback-parameters.html)).

## Why this only affects vchord users

The other three backends (`pgvector`, `pgvectorscale`, `pg_diskann`) all map to `vector_cosine_ops` in `_vector_index.py`, so their indexes match the `<=>` operator and the same code paths perform as intended.

## What the fix probably looks like (for discussion)

VectorChord's documentation [explicitly recommends `vector_cosine_ops` for normalized/cosine embeddings](https://docs.vectorchord.ai/vectorchord/usage/indexing.html):

> "For most datasets using cosine similarity, enabling `residual_quantization` and `build.internal.spherical_centroids` improves both QPS and recall."

So the cleanest fix is to switch the vchord mapping in `_vector_index.py` to `vector_cosine_ops` (optionally with `residual_quantization = true` and `spherical_centroids = true`) and provide an Alembic migration that re-creates the existing vchordrq indexes — both the global one (`idx_memory_units_embedding_vchordrq`) and the many per-`(bank_id, fact_type)` partial indexes — with the new opclass, ideally via `CREATE INDEX CONCURRENTLY` followed by drops of the old ones.

The secondary `hnsw.ef_search` issue can be fixed by dispatching the tuning GUC by backend (`vchordrq.probes` / `vchordrq.epsilon` for vchord, `hnsw.ef_search` for pgvector, etc.).

I'm happy to send a PR along these lines — opening this issue first so the diagnosis is documented and so anyone hitting the same symptom can find it.

## Reproduction

1. Deploy Hindsight against PostgreSQL with the `vchord` extension and `HINDSIGHT_API_VECTOR_EXTENSION=vchord`.
2. Ingest enough memories that any single `(bank_id, fact_type)` partition contains a few thousand rows.
3. Run any retain (or recall) that triggers ANN search.
4. Observe Postgres CPU climb and the activity table show long-running queries with `<=>` and no `wait_event`.
5. `EXPLAIN (ANALYZE, BUFFERS)` shows `Seq Scan` over the partition with the vchordrq index ignored.

isac322:
Closes #1667.

This PR fixes two related vchord-only correctness bugs that together explain the high CPU on vchord deployments reported in #1667:

1. **Opclass mismatch (the hot one).** `_vector_index.py` mapped vchord to `vchordrq (embedding vector_l2_ops)`, but every ANN query in the engine uses `<=>` (cosine distance), which `vector_l2_ops` does not match. The planner ignored the index entirely and fell back to sequential scan + per-row cosine computation, so the production primary sat at ~870m CPU during retain.
2. **Tuning GUC mismatch.** `SET LOCAL hnsw.ef_search = 60` (and `SET hnsw.ef_search = 200` at pool init) only exists in pgvector. On a vchord deployment both statements silently no-op'd, which meant the "recall-vs-latency" trade-off the code was designed around had never actually been applied to vchord backends.

## What changes

**Switch the vchord opclass to `vector_cosine_ops`:**
- `_vector_index.py:_INDEX_USING_CLAUSES["vchord"]` flipped to `vector_cosine_ops`.
- The four historical migrations that create vchord indexes inline get the same flip so fresh installs end up on cosine ops from the first migration onward: `5a366d414dce`, `a4b5c6d7e8f9`, `d5e6f7a8b9c0`, `n9i0j1k2l3m4`.
- A new online migration `b8c9d0e1f2a3_vchord_cosine_opclass` rebuilds any existing `vector_l2_ops` vchordrq index with `vector_cosine_ops` via `CREATE INDEX CONCURRENTLY` + drop + rename. No-op when the configured extension is not vchord or when no L2-ops vchordrq indexes are found.

**Dispatch ANN tuning GUCs by backend:**
- New helper `_vector_index.ann_search_tuning_settings(ext, kind)` returns `(guc_name, value)` pairs per backend. `kind="low_latency"` powers the retain-side ANN probe (was `hnsw.ef_search = 60`), `kind="high_recall"` powers the pool init (was `hnsw.ef_search = 200`).
- Defaults: pgvector keeps `hnsw.ef_search = 60 / 200`; vchord uses `vchordrq.probes = 10 / 30`. pgvectorscale, pg_diskann, scann return no statements (no equivalent knob exposed today).
- `engine/retain/link_utils.py` and `engine/memory_engine.py` switched to the dispatcher. Callers still own `SET` vs `SET LOCAL` since the two sites have different lifetime requirements.

## Why
See #1667 for the full diagnosis. On the production cluster the same ANN query goes from 143 ms with `<=>` against `vector_l2_ops` (sequential scan, `Buffers: shared hit=71531`) to 25 ms with the same query body against a `vector_cosine_ops` index (`Index Scan using idx_mu_emb_expr_<hash>`, `Buffers: shared hit=770 read=520`) — ~5.6× per-query, and because the retain pipeline wraps this in `CROSS JOIN LATERAL` over `_ann_seeds` the penalty multiplies by seed count.

VectorChord's docs [recommend `vector_cosine_ops` for cosine-similarity datasets](https://docs.vectorchord.ai/vectorchord/usage/indexing.html); the engine's embeddings are L2-normalized in practice, so the existing `<=>` SQL is what we want. pgvector / pgvectorscale / pg_diskann were already on cosine ops, so this also restores cross-backend consistency.

## Test plan
- [x] `uv run ruff check` on the changed files — pass
- [x] `uv run ty check hindsight_api/` — pass
- [x] `uv run pytest tests/test_vector_index.py tests/test_link_utils.py tests/test_migration_shape.py` — 119 passed (new regression assertions: `test_index_using_clause_vchord_uses_cosine_ops`, `test_ann_search_tuning_settings_{pgvector,vchord,…}`, and the existing `test_uses_set_local_for_ann_tuning` parametrized over pgvector/vchord)
- [x] Manual on a vchord docker-compose stack (`tensorchord/vchord-suite:pg18-latest`): ran the migration cleanly, confirmed `pg_indexes` shows `vector_cosine_ops` after both the fresh-install path and a `downgrade → upgrade` roundtrip from a synthesized `vector_l2_ops` "broken" state, and confirmed `EXPLAIN` picks the vchordrq index for the `<=>` query. CPU return-to-baseline observation deferred to staging/prod rollout. See PR comment for full transcript.

## Migration impact
For vchord deployments the new migration rebuilds each vchordrq index online with `CREATE INDEX CONCURRENTLY`. Disk usage doubles for each index during its rebuild; queries continue to run on the old index until the swap finishes. Downgrade rebuilds back to `vector_l2_ops` using the same online flow. After the migration completes, vchord pools start applying `vchordrq.probes` instead of the silently-ignored `hnsw.ef_search`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/alembic/versions/b8c9d0e1f2a3_vchord_cosine_opclass.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/b8c9d0e1f2a3_vchord_cosine_opclass.py
@@ -1,0 +1,100 @@
+"""Re-create vchord vector indexes with vector_cosine_ops
+
+Revision ID: b8c9d0e1f2a3
+Revises: 86f7a033d372
+Create Date: 2026-05-20
+
+vchordrq operator classes are bound 1:1 to operators in PostgreSQL:
+vector_l2_ops only matches ``<->``, while every Hindsight ANN query uses
+``<=>`` (cosine distance). The previous vchord mapping used vector_l2_ops,
+so vchord deployments could never use the index — every ANN query fell
+back to a sequential scan with per-row cosine computation.
+
+This migration finds any vchordrq index built with vector_l2_ops in the
+target schema and re-creates it with vector_cosine_ops, using
+``CREATE INDEX CONCURRENTLY`` so it can run online. It is a no-op when:
+
+* the configured vector extension is not vchord, or
+* no matching indexes exist (already on cosine ops).
+
+Only PostgreSQL is affected; the Oracle 23ai dialect uses its own native
+vector index and does not depend on this mapping.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+from collections.abc import Sequence
+
+from alembic import context, op
+from sqlalchemy import text
+
+from hindsight_api.alembic._dialect import run_for_dialect
+
+revision: str = "b8c9d0e1f2a3"
+down_revision: str | Sequence[str] | None = "86f7a033d372"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _configured_vector_extension() -> str:
+    return os.getenv("HINDSIGHT_API_VECTOR_EXTENSION", "pgvector").lower()
+
+
+def _rebuild_vchordrq_indexes(old_ops: str, new_ops: str) -> None:
+    """Rebuild vchordrq indexes using ``old_ops`` so they use ``new_ops``.
+
+    Each index is rebuilt with CREATE INDEX CONCURRENTLY under a fresh name,
+    then the old index is dropped and the new one renamed to take its place.
+    Must be called inside an ``autocommit_block()`` because CONCURRENTLY
+    cannot run inside a transaction.
+    """
+    bind = op.get_bind()
+    schema = context.config.get_main_option("target_schema") or "public"
+
+    rows = bind.execute(
+        text(
+            "SELECT indexname, indexdef FROM pg_indexes "
+            "WHERE schemaname = :schema "
+            "AND indexdef ILIKE '%vchordrq%' "
+            "AND indexdef ILIKE :ops_like"
+        ),
+        {"schema": schema, "ops_like": f"%{old_ops}%"},
+    ).fetchall()
+
+    for idx_name, indexdef in rows:
+        new_def = indexdef.replace(old_ops, new_ops, 1)
+        temp_name = f"{idx_name}__opclass_swap"
+        new_def = new_def.replace(idx_name, temp_name, 1)
+        new_def = re.sub(
+            r"^CREATE\s+INDEX\b",
+            "CREATE INDEX CONCURRENTLY IF NOT EXISTS",
+            new_def,
+            count=1,
+        )
+        op.execute(new_def)
+        op.execute(f'DROP INDEX IF EXISTS "{schema}"."{idx_name}"')
+        op.execute(f'ALTER INDEX "{schema}"."{temp_name}" RENAME TO "{idx_name}"')
+
+
+def _pg_upgrade() -> None:
+    if _configured_vector_extension() != "vchord":
+        return
+    with op.get_context().autocommit_block():
+        _rebuild_vchordrq_indexes("vector_l2_ops", "vector_cosine_ops")
+
+
+def _pg_downgrade() -> None:
+    if _configured_vector_extension() != "vchord":
+        return
+    with op.get_context().autocommit_block():
+        _rebuild_vchordrq_indexes("vector_cosine_ops", "vector_l2_ops")
+
+
+def upgrade() -> None:
+    run_for_dialect(pg=_pg_upgrade)
+
+
+def downgrade() -> None:
+    run_for_dialect(pg=_pg_downgrade)
--- a/hindsight-api-slim/hindsight_api/alembic/versions/b8c9d0e1f2a3_vchord_cosine_opclass.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/b8c9d0e1f2a3_vchord_cosine_opclass.py
@@ -23,23 +23,19 @@ vector index and does not depend on this mapping.
 
 from __future__ import annotations
 
-import os
 import re
 from collections.abc import Sequence
 
 from alembic import context, op
 from sqlalchemy import text
 
+from hindsight_api._vector_index import configured_vector_extension
 from hindsight_api.alembic._dialect import run_for_dialect
 
 revision: str = "b8c9d0e1f2a3"
 down_revision: str | Sequence[str] | None = "86f7a033d372"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
-
-
-def _configured_vector_extension() -> str:
-    return os.getenv("HINDSIGHT_API_VECTOR_EXTENSION", "pgvector").lower()
 
 
 def _rebuild_vchordrq_indexes(old_ops: str, new_ops: str) -> None:
@@ -79,14 +75,14 @@ def _rebuild_vchordrq_indexes(old_ops: str, new_ops: str) -> None:
 
 
 def _pg_upgrade() -> None:
-    if _configured_vector_extension() != "vchord":
+    if configured_vector_extension() != "vchord":
         return
     with op.get_context().autocommit_block():
         _rebuild_vchordrq_indexes("vector_l2_ops", "vector_cosine_ops")
 
 
 def _pg_downgrade() -> None:
-    if _configured_vector_extension() != "vchord":
+    if configured_vector_extension() != "vchord":
         return
     with op.get_context().autocommit_block():
         _rebuild_vchordrq_indexes("vector_cosine_ops", "vector_l2_ops")
--- a/hindsight-api-slim/hindsight_api/alembic/versions/b8c9d0e1f2a3_vchord_cosine_opclass.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/b8c9d0e1f2a3_vchord_cosine_opclass.py
@@ -38,6 +38,12 @@ branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
 
+def _pg_schema_prefix() -> str:
+    """Schema-qualifier for raw SQL on PG (multi-tenant search_path)."""
+    schema = context.config.get_main_option("target_schema")
+    return f'"{schema}".' if schema else ""
+
+
 def _rebuild_vchordrq_indexes(old_ops: str, new_ops: str) -> None:
     """Rebuild vchordrq indexes using ``old_ops`` so they use ``new_ops``.
 
@@ -47,19 +53,27 @@ def _rebuild_vchordrq_indexes(old_ops: str, new_ops: str) -> None:
     cannot run inside a transaction.
     """
     bind = op.get_bind()
-    schema = context.config.get_main_option("target_schema") or "public"
+    # `or None` collapses both unset and explicit empty-string Alembic options
+    # into NULL so the COALESCE below falls back to current_schema() in either
+    # case. Without it, an empty-string option would filter on `schemaname = ''`
+    # and skip every real schema.
+    target_schema = context.config.get_main_option("target_schema") or None
+    prefix = _pg_schema_prefix()
 
     rows = bind.execute(
         text(
             "SELECT indexname, indexdef FROM pg_indexes "
-            "WHERE schemaname = :schema "
+            "WHERE schemaname = COALESCE(:target_schema, current_schema()) "
             "AND indexdef ILIKE '%vchordrq%' "
             "AND indexdef ILIKE :ops_like"
         ),
-        {"schema": schema, "ops_like": f"%{old_ops}%"},
+        {"target_schema": target_schema, "ops_like": f"%{old_ops}%"},
     ).fetchall()
 
     for idx_name, indexdef in rows:
+        # pg_get_indexdef() emits the canonical form `CREATE INDEX <name> ON …`,
+        # so <name> is the first textual occurrence — both substitutions below
+        # rely on that.
         new_def = indexdef.replace(old_ops, new_ops, 1)
         temp_name = f"{idx_name}__opclass_swap"
         new_def = new_def.replace(idx_name, temp_name, 1)
@@ -69,9 +83,51 @@ def _rebuild_vchordrq_indexes(old_ops: str, new_ops: str) -> None:
             new_def,
             count=1,
         )
+
+        # CREATE INDEX CONCURRENTLY can leave the partial index as INVALID if a
+        # previous run errored (disk pressure, lock conflict, signal). Without
+        # this drop the CONCURRENTLY IF NOT EXISTS below would skip creation,
+        # then we'd drop the original and rename the broken index into its
+        # place — silently restoring the seq-scan bug this migration fixes.
+        op.execute(f'DROP INDEX IF EXISTS {prefix}"{temp_name}"')
         op.execute(new_def)
-        op.execute(f'DROP INDEX IF EXISTS "{schema}"."{idx_name}"')
-        op.execute(f'ALTER INDEX "{schema}"."{temp_name}" RENAME TO "{idx_name}"')
+
+        # Even on a clean run CONCURRENTLY can finish with indisvalid = false
+        # (e.g. constraint violation during the second build scan). Refuse to
+        # promote in that case so we never alias an INVALID index over a working
+        # one.
+        is_valid = bind.execute(
+            text(
+                "SELECT i.indisvalid "
+                "FROM pg_class c "
+                "JOIN pg_index i ON c.oid = i.indexrelid "
+                "JOIN pg_namespace n ON c.relnamespace = n.oid "
+                "WHERE c.relname = :name "
+                "  AND n.nspname = COALESCE(:target_schema, current_schema())"
+            ),
+            {"name": temp_name, "target_schema": target_schema},
+        ).scalar()
+        if not is_valid:
+            raise RuntimeError(
+                f"vchordrq index rebuild produced an INVALID index ({temp_name}); "
+                "drop it manually and re-run the migration."
+            )
+
+        # DROP + RENAME atomically. A crash between the two would leave
+        # `temp_name` as a valid orphan and the canonical name missing —
+        # next run's `pg_indexes` filter (looking for vector_l2_ops) wouldn't
+        # find anything to recover from, so the index would stay gone. PG
+        # runs the DO block in its own server-side transaction, so either
+        # both succeed or both roll back.
+        op.execute(
+            f"""
+            DO $$
+            BEGIN
+                DROP INDEX IF EXISTS {prefix}"{idx_name}";
+                ALTER INDEX {prefix}"{temp_name}" RENAME TO "{idx_name}";
+            END $$;
+            """
+        )
 
 
 def _pg_upgrade() -> None:
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -2057,7 +2057,11 @@ class MemoryEngine(MemoryEngineInterface):
             for guc, value in ann_search_tuning_settings(configured_vector_extension(), kind="high_recall"):
                 try:
                     await conn.execute(f"SET {guc} = {value}")
-                except Exception:
+                except asyncpg.exceptions.PostgresError:
+                    # Defensive net for env mis-config (e.g. extension configured
+                    # for vchord but the cluster only has pgvector). Narrow to
+                    # PostgresError so genuine bugs in the pool/conn layer surface
+                    # instead of being silently logged at debug level.
                     logger.debug("Could not set %s — extension may not support it", guc)
 
             # Server-side safety net for runaway queries. Migrations use a
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -13,6 +13,7 @@ import asyncio
 import contextvars
 import json
 import logging
+import os
 import time
 import uuid
 from collections.abc import Awaitable, Callable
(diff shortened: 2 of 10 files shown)


isac322:
The `verify-generated-files` failure on this PR is pre-existing on `main` and unrelated to the changes here. The job reports drift on:

- `skills/hindsight-docs/references/developer/models.md`
- `skills/hindsight-docs/references/faq.md`

…which are LLM-provider and docs-generation outputs. This PR only touches `_vector_index.py`, two engine files (`engine/retain/link_utils.py`, `engine/memory_engine.py`), five Alembic migrations under `hindsight_api/alembic/versions/`, and `tests/test_vector_index.py` + `tests/test_link_utils.py` — none of which are inputs to those generators.

Happy to rebase once `main` is regenerated, or to include the regenerated files in this PR if that's preferable. Let me know which direction you'd prefer.

hindsight-api-slim/hindsight_api/alembic/versions/b8c9d0e1f2a3_vchord_cosine_opclass.py
```diff
@@ -0,0 +1,100 @@
...
+from hindsight_api.alembic._dialect import run_for_dialect
+
+revision: str = "b8c9d0e1f2a3"
+down_revision: str | Sequence[str] | None = "86f7a033d372"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _configured_vector_extension() -> str:
+    return os.getenv("HINDSIGHT_API_VECTOR_EXTENSION", "pgvector").lower()
```
elliotclee: There should be a function (such as this one) in the main hindsight codebase that all places call to find out the configured vector extension. I don't think an alembic migration file is the right place to host such a function, and there are a few places in this patchset that should be using a function like this but are doing the getenv() call directly instead.

This matters because the method of accessing the configuration setting could concievably change in the future, but more importantly because the default value needs to be configurable from one spot.

hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
```diff
@@ -701,16 +703,19 @@ async def compute_semantic_links_ann(
...
-        # link creation. SET LOCAL auto-reverts at commit, so we don't pollute
-        # the pool for subsequent recall queries.
-        await conn.execute("SET LOCAL hnsw.ef_search = 60")
+        # Transaction-local ANN tuning. Each supported backend exposes its own
+        # GUC (hnsw.ef_search on pgvector, vchordrq.probes on vchord); the
+        # dispatcher returns the right knob for the configured backend with a
+        # value tuned for top-50 semantic link creation (lower recall but much
+        # lower latency than the recall-side default). SET LOCAL auto-reverts
+        # at commit, so we don't pollute the pool for subsequent queries.
+        ext = os.getenv("HINDSIGHT_API_VECTOR_EXTENSION", "pgvector").lower()
```
elliotclee: See previous comment.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -1920,13 +1922,18 @@ async def verify_llm():
...
-                await conn.execute("SET hnsw.ef_search = 200")
-            except Exception:
-                logger.debug("Could not set hnsw.ef_search — extension may not support it")
+            # SET (not SET LOCAL) so per-backend ANN tuning persists for the
+            # connection lifetime. Each backend exposes its own GUC: pgvector
+            # uses hnsw.ef_search, vchord uses vchordrq.probes. The dispatcher
+            # returns the right one for the configured extension, tuned for
+            # the higher recall the per-fact_type semantic queries in
+            # retrieve_semantic_bm25_combined() need.
+            ext = os.getenv("HINDSIGHT_API_VECTOR_EXTENSION", "pgvector").lower()
```
elliotclee: See previous comment.

elliotclee (commented):
Thank you for noticing this issue and submitting the patch. Just one small change requested below...

elliotclee (approved):
Looks good to me!

hindsight-api-slim/hindsight_api/alembic/versions/b8c9d0e1f2a3_vchord_cosine_opclass.py
isac322: Done in c4011bcb. New `configured_vector_extension()` lives in `hindsight_api/_vector_index.py` and both runtime call sites (`link_utils.py`, `memory_engine.py`) and this migration now call it instead of doing `os.getenv()` directly.

One thing I wanted to flag before merging: hindsight has a deliberate test (`test_alembic_vector_migrations_freeze_vector_sql_locally`) that explicitly forbids the existing vchord-related migrations from importing `hindsight_api._vector_index`. The intent reads as "each historical migration's vector SQL must be frozen so it can't drift later if `_vector_index.py` is changed". The four legacy migrations (`5a366d414dce`, `a4b5c6d7e8f9`, `d5e6f7a8b9c0`, `n9i0j1k2l3m4`) each keep their own inline `_vector_index_using_clause()` helper for the same reason.

For this PR I narrowed that frozen-state list to those four legacy migrations and let the new `b8c9d0e1f2a3` opt into the shared helper, since that matches what you asked for. But I can imagine three ways to draw the line:

1. **Current PR** — only the new migration uses the shared helper; legacy 4 stay frozen. ("Don't change history, do use the shared helper going forward.")
2. **Strict frozen-state** — revert the new migration to its own inline helper too, so all five vchord migrations look identical. The runtime call-site dedup (`link_utils.py`, `memory_engine.py`) still stands.
3. **Drop the frozen-state policy entirely** — port the legacy four to the shared helper too and delete the freeze test.

If (2) is what you'd prefer (i.e. you'd like the new migration to stay frozen too), happy to swap. (3) is a bigger policy call that probably belongs in a separate PR if it's worth doing at all. Which direction would you like?

hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
isac322: Addressed in c4011bcb — see reply on the first thread for the trade-off discussion around the frozen-state test.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
isac322: Addressed in c4011bcb — see reply on the first thread for the trade-off discussion around the frozen-state test.

nicoloboschi (changes requested):
Nice diagnosis and a well-scoped fix — the `EXPLAIN` evidence in the description (143ms seq scan → 25ms index scan) makes the root cause unambiguous, and the per-backend dispatcher is a cleaner abstraction than the previous `hnsw.ef_search`-everywhere assumption. Particularly liked:

- Flipping the four historical migrations to `vector_cosine_ops` so fresh vchord installs don't bootstrap-then-rebuild.
- `configured_vector_extension()` centralizing the env lookup instead of scattering `os.getenv` calls.
- Parametrized `test_uses_set_local_for_ann_tuning` over pgvector/vchord, and the explicit empty-tuple assertions for backends without an equivalent knob.
- Renaming `changed_migrations` → `frozen_migrations` in `test_alembic_vector_migrations_freeze_vector_sql_locally` with a comment carve-out for the new migration — that's the right call.

Requesting changes on a few items before merge:

### Must address

**1. `CREATE INDEX CONCURRENTLY` failure-recovery in `b8c9d0e1f2a3_vchord_cosine_opclass.py`.**
If `_rebuild_vchordrq_indexes` errors midway (disk pressure, lock conflict, etc.), Postgres leaves the partial index as `INVALID`. On re-run:
- `CREATE INDEX CONCURRENTLY IF NOT EXISTS` skips the INVALID temp index (it exists).
- `DROP INDEX IF EXISTS` drops the old original.
- `ALTER INDEX … RENAME` promotes the **INVALID** temp index into the canonical name.

Result: a broken index masquerading as the original, and queries silently fall back to seq scan again — exactly the bug we're fixing. Suggest either (a) `DROP INDEX IF EXISTS … __opclass_swap` at the top of each loop iteration, or (b) querying `pg_index.indisvalid` after the CONCURRENTLY create and raising if false.

**2. Schema-prefix inconsistency.** The new migration uses `context.config.get_main_option(\"target_schema\") or \"public\"`, while every other PG migration in the tree uses `_pg_schema_prefix()` which returns `\"\"` when unset. Align with the existing convention so behavior is the same when `target_schema` is not configured.

**3. Manual vchord verification is unchecked.** The test plan's last bullet (run on a real vchord cluster, verify `pg_indexes` shows `vector_cosine_ops`, confirm `EXPLAIN` picks the index, observe CPU return to baseline) is still `[ ]`. Given this is the whole point of the PR, please complete it before merge and post the result here — automated tests don't exercise the migration against a real vchord backend.

### Nice to have

**4. GUC values lack a citation.** `vchordrq.probes = 10 / 30` look heuristic compared to the pgvector `60 / 200` numbers that presumably have history. If the chosen values land on the wrong knee of the recall/latency curve, we'd be substituting one silent miscalibration for another. Either a quick benchmark note in the PR or a follow-up tuning task would be enough.

**5. `indexdef.replace(idx_name, temp_name, 1)`** relies on the index name being the first textual occurrence in `pg_indexes.indexdef` — true in practice for the canonical reconstructed form, but worth a one-line comment noting the assumption.

**6. `_init_connection` exception swallowing.** Now that the dispatcher returns an empty tuple for backends without a knob, the broad `except Exception` is doing less work than it used to. Tightening to `asyncpg.exceptions.PostgresError` would avoid masking real bugs.

Overall: solid fix, scoped tightly to vchord, with the rest of the backends provably unchanged. The CONCURRENTLY recovery gap is the only thing I'd block on.

isac322:
Thanks for the thorough review — addressed all of it in commit 73dfcf8f. Summary:

### Must address

**1. `CREATE INDEX CONCURRENTLY` recovery (`b8c9d0e1f2a3_vchord_cosine_opclass.py`)**

Two layers of defense against the INVALID-leftover scenario:
- `DROP INDEX IF EXISTS … __opclass_swap` at the top of each loop iteration, so a leftover INVALID temp index from a prior failed run can't be skipped by `CREATE … IF NOT EXISTS` and then promoted into the canonical name.
- After CREATE, query `pg_index.indisvalid` and raise `RuntimeError` if false — guards against the current run's CONCURRENTLY finishing INVALID (e.g. constraint violation during the second build scan).

While wiring this up I noticed a related gap: a crash *between* the canonical DROP and the RENAME would leave the temp as a valid orphan and the canonical name missing, with no recovery on retry (next run's `WHERE indexdef ILIKE '%vector_l2_ops%'` filter wouldn't match anything). Wrapped the DROP + RENAME in a server-side PL/pgSQL `DO` block so PG runs them in a single implicit transaction — both succeed or both roll back.

**2. Schema-prefix convention**

Switched to the `_pg_schema_prefix()` helper used by the other PG migrations. The `pg_indexes` filter now uses `WHERE schemaname = COALESCE(:target_schema, current_schema())`, and I normalize the Alembic option with `or None` so an explicit empty-string value also falls back to `current_schema()` rather than filtering on `schemaname = ''`.

**3. Manual vchord verification**

Done — see the follow-up comment with the docker-compose transcript. Both fresh-install and downgrade→upgrade-from-l2_ops paths confirmed, plus an EXPLAIN proving the index is now picked for `<=>`.

### Nice to have

**4. `vchordrq.probes` provenance**

Expanded the comment on `_ANN_TUNING_*` in `_vector_index.py` to explain that the pgvector 60/200 pair predates the dispatcher (internal benchmarks tied to our embedding count + recall floor) and that vchord 10/30 are conservative starting defaults pending a workload-specific sweep — vchordrq's recall curve shape differs from HNSW's so the pgvector numbers don't translate directly.

**5. `indexdef.replace` first-occurrence assumption**

Added a one-line comment noting that `pg_get_indexdef()` emits the canonical form `CREATE INDEX <name> ON …`, so both substitutions rely on `<name>` being the first textual occurrence.

**6. `_init_connection` exception**

Narrowed to `asyncpg.exceptions.PostgresError`. Worth noting for context: in practice this defensive net is rarely tripped, because PG accepts `qualifier.name = value` SETs with arbitrary qualifiers — so `SET vchordrq.probes = 10` doesn't error on a pgvector-only cluster, it just silently no-ops. The narrowed `except` is the right hygiene either way.

isac322:
### Manual verification on a vchord docker-compose stack

Ran the migration locally against `tensorchord/vchord-suite:pg18-latest` using `docker/docker-compose/vchord/docker-compose.yaml`, applying migrations from our local API code (not the published image). Both the fresh-install path and the broken-state recovery path the migration is designed for behave as expected.

**Fresh install (clean DB → `upgrade heads`)**

`pg_indexes` after applying every migration through `b8c9d0e1f2a3`:

```
          indexname          |                                                   indexdef
-----------------------------+---------------------------------------------------------------------------------------------------------------
 idx_mental_models_embedding | CREATE INDEX idx_mental_models_embedding ON public.mental_models USING vchordrq (embedding vector_cosine_ops)
 idx_memory_units_embedding  | CREATE INDEX idx_memory_units_embedding ON public.memory_units USING vchordrq (embedding vector_cosine_ops)
```

Both vchordrq indexes land on `vector_cosine_ops` from the first migration onward, as intended by the historical-migration flips.

**Broken-state recovery (the production scenario from #1667)**

1. `alembic downgrade b8c9d0e1f2a3 → 86f7a033d372` — `_pg_downgrade` rebuilt both indexes as `vector_l2_ops`, reproducing the production "broken" state.
2. `alembic upgrade heads` — the new migration rebuilt both indexes back as `vector_cosine_ops`, no manual intervention needed.
3. `pg_index.indisvalid = true` for both indexes after the upgrade.
4. No `*__opclass_swap` orphan indexes left behind — confirming the DO-block-wrapped `DROP + RENAME` is atomic across the swap.

**EXPLAIN — operator class binding**

After inserting 200 synthetic `memory_units` rows and `ANALYZE`-ing, the cosine ANN query picks the vchordrq index:

```
->  Index Scan using idx_memory_units_embedding on memory_units  (cost=0.20..214.70 rows=200 width=32)
      Order By: (embedding <=> (InitPlan 2).col1)
      Filter: ((embedding IS NOT NULL) AND (bank_id = 'test-bank'::text) AND (fact_type = 'world'::text))
```

Previously, with `vector_l2_ops`, the planner ignored the index and fell back to a sequential scan — which is exactly the original bug.

**Caveat**

CPU return-to-baseline can only be observed against the real production workload; that part is deferred to staging/prod rollout.

nicoloboschi (approved):

Issue #1571, opened 2026-05-11
Sub-batch splitter announces 10K-token chunks but processes full payload as 1/1, triggering OOM on large single-item retains


JC-Glo:
## Summary

When a single retain item exceeds the splitter's per-batch token target, `memory_engine` logs that it is splitting into ~10K-token sub-batches but then produces `1/1` sub-batches each containing the full oversize payload — no actual split occurs. The worker then processes the full payload, which OOM-kills the container under the documented `mem_limit: 4g` and (because of `restart: unless-stopped` + task recovery on startup) traps the container in an infinite restart loop until an operator wipes the embedded PG.

This appears to be a logic gap that's distinct from #1345 / PR #1346: that fix addressed one *trigger* (Codex auto-retain generating huge full-session payloads via synthetic AGENTS.md messages). The splitter itself is still capable of failing the same way on any oversize single retain — Codex was just one source.

Related prior art:
- #1345 (closed, PR #1346) — Codex trigger, fixed Codex-side only
- #418 (closed) — earlier memory-leak-in-consolidation-during-large-batch repro
- #996 (closed) — adjacent unbounded-memory pattern
- #1304 (open) — token-budgeting gap upstream of this

## Reproduction

1. Submit one retain POST containing a single content field of ~1.17M tokens (in our case, generated by a recall query that echoed indexed content as context).
2. Worker accepts the task and begins processing.

## Observed log sequence (verbatim, container log, two independent runs)

```
2026-05-07 12:11:54 - hindsight_api.engine.memory_engine - Large batch detected (1,167,556 tokens from 1 items). Splitting into sub-batches of ~10,000 tokens each...
2026-05-07 12:12:28 - hindsight_api.engine.memory_engine - Split into 1 sub-batches: [1] items each
2026-05-07 12:13:06 - hindsight_api.engine.memory_engine - Processing sub-batch 1/1: 1 items, 1,167,556 tokens
/app/start-all.sh: line 226:     9 Killed                  hindsight-api  (wd: /app/api)
```

Same sequence reproduced ~4 hours later when the container auto-restarted and the worker recovered the same task from the queue:

```
2026-05-07 16:11:47 - hindsight_api.worker.poller - Worker <id> recovered 1 stale tasks from previous run
2026-05-07 16:11:48 - hindsight_api.engine.memory_engine - [BATCH_RETAIN_TASK] Starting background batch retain for bank_id=<bank>, 1 items, operation_id=<uuid>
[...]

Issue #1037, opened 2026-04-13
batch_retain IndexError: list index out of range in orchestrator.py _map_results_to_contents


parteeksinghlavish:
## Bug Description

When running the `hindsight-openclaw-backfill` CLI to import historical OpenClaw sessions, the Hindsight server fails with an `IndexError: list index out of range` during `batch_retain` operations.

## Error Traceback

```
IndexError: list index out of range

Traceback (most recent call last):
  File "/app/api/hindsight_api/engine/memory_engine.py", line 1098, in execute_task
    await self._handle_batch_retain(task.task_dict)
  File "/app/api/hindsight_api/engine/memory_engine.py", line 662, in _handle_batch_retain
    await self.retain_batch_async(...)
  File "/app/api/hindsight_api/engine/memory_engine.py", line 2245, in retain_batch_async
    sub_results, sub_usage = await self._retain_batch_async_internal(...)
  File "/app/api/hindsight_api/engine/memory_engine.py", line 2373, in _retain_batch_async_internal
    return await orchestrator.retain_batch(...)
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 606, in retain_batch
    return await _streaming_retain_batch(...)
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 1103, in _streaming_retain_batch
    await asyncio.gather(_llm_producer(), _db_consumer())
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 932, in _db_consumer
    await _process_db_batch(...)
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 1070, in _process_db_batch
    await _run_mini_batch_db_work()
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 1033, in _run_mini_batch_db_work
    batch_result_ids, phase3_ctx = await _insert_facts_and_links(...)
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 306, in _insert_facts_and_links
    result_unit_ids = _map_results_to_contents(contents, extracted_facts, unit_ids if unit_ids else [])
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 1571, in _map_results_to_contents
    content_unit_ids.append(unit_ids[unit_idx])
    ~~~~~~~~^^^^^^^^^^IndexError: list index out of range
```

## Environment

- Hindsight version: (running as Docker container)
[...]

PR #1670, opened 2026-05-20
fix: validate embedding dimensions before pgvector writes

### LINKED ISSUE #1669: batch_retain can pass zero-length embeddings to pgvector (vector(384) vs vector(0))
## Summary

`batch_retain` / `retain` can fail inside PostgreSQL with:

```text
asyncpg.exceptions.DataError: different vector dimensions 384 and 0
```

The database schema expects `vector(384)`, but at least one retain path can pass an empty embedding vector (`[]`, effectively dimension 0) into an insert/update path. This surfaces as a DB exception rather than a controlled retain failure or quarantine.

## Observed environment

- `hindsight-api`: 0.6.2
- embeddings provider: local `BAAI/bge-small-en-v1.5`, initialized as dim 384
- vector extension: pgvector
- DB column: `memory_units.embedding vector(384)`
- API `/health`: healthy while worker tasks failed/retried

## Representative traceback

```text
Task execution failed: batch_retain, error: different vector dimensions 384 and 0
...
hindsight_api.engine.retain.orchestrator._streaming_retain_batch
hindsight_api.engine.retain.orchestrator._process_db_batch
hindsight_api.engine.retain.orchestrator._insert_facts_and_links
hindsight_api.engine.retain.fact_storage.insert_facts_batch
hindsight_api.engine.db.ops_postgresql.insert_facts_batch
asyncpg.exceptions.DataError: different vector dimensions 384 and 0
```

A second observed path went through:

```text
orchestrator._try_delta_retain
fact_storage.update_memory_units_tags
asyncpg.exceptions.DataError: different vector dimensions 384 and 0
```

## Impact

- The API can remain healthy while retain workers repeatedly retry poisoned operations.
- A single bad retain operation can occupy the reserved retain worker slot for minutes.
- Operator-visible health is misleading unless async operation error states are inspected directly.
- Retain queue progress stalls or becomes noisy until the bad operations are manually cancelled/quarantined.

## Local mitigation used

A local hotfix was tested that validates embedding batch outputs for count and per-vector dimension, retrying invalid vectors individually. It caught some invalid-vector shapes, but did not catch every observed failure path, suggesting there is at least one additional route where an empty vector can reach the DB write path.

No raw memory data was deleted. Only poisoned async retain operations were cancelled locally to restore worker progress.

## Suggested fix

Add a central embedding validation boundary before any vector reaches pgvector operations:

- validate `len(embedding) > 0`;
- if the backend exposes `dimension`, validate `len(embedding) == backend.dimension`;
- apply this to every retain path that writes/compares embeddings:
  - streaming retain insert;
  - delta retain;
  - semantic ANN / link precomputation;
  - any update path that binds vectors;
- convert invalid vectors into a controlled retain error with operation metadata, not a raw asyncpg exception;
- optionally quarantine the offending fact/item while allowing the rest of the batch to continue when safe.

## Regression tests

Suggested tests with a fake embedding backend:

1. Batch encoder returns `[valid_384, []]` for two extracted facts.
   - Expected: no `asyncpg.DataError`; operation fails/quarantines with a clear invalid embedding dimension error.
2. Single-item retain/delta path produces an empty vector.
   - Expected: controlled error before SQL execution.
3. Mixed batch with one invalid fact.
   - Expected behavior should be explicit: either whole operation fails with a clear error, or the bad fact is skipped/quarantined and the rest completes.

## Related issues found during duplicate check

- #1037 closed: `batch_retain IndexError: list index out of range in orchestrator.py _map_results_to_contents`
- #1154 closed: embedding setup issue
- #1571 open: oversized single-item retains are not actually split and can block/crash workers

This issue is related to retain robustness, but the exact pgvector dimension mismatch appears distinct.

ai-ag2026:
## Summary
- validate embedding vectors before they can reach pgvector writes
- reject empty vectors and vectors whose length does not match the backend's declared dimension
- cover both retain and consolidation observation creation paths with regression tests

## Root cause
Retain embedding generation already guaranteed that the number of returned vectors matched the number of input texts, but it did not validate per-vector dimensions. If an embedding backend returned `[]` or another wrong-sized vector, the invalid value could flow downstream until PostgreSQL/pgvector raised `different vector dimensions 384 and 0`.

The same shared embedding utility is also used by consolidation when creating observations, so the validation boundary protects both retain writes and consolidation-created observations.

## Related reconnaissance
- Fixes #1669.
- Related to #1671: without quarantine/degraded health, poisoned async operations can keep retrying even while `/health` is green.
- Related but separate from #1571, which tracks oversized single retain items not being split before retain processing.
- Existing tests already covered vector count mismatches from #1037; this PR extends the same boundary to validate each vector's shape.

## Test plan
- RED on current `origin/main`: `uv run pytest tests/test_consolidation_embedding_validation.py::test_create_observation_rejects_zero_length_embedding_before_insert -q -o addopts=` failed because the zero-length vector reached the DB insert path.
- `uv run pytest tests/test_consolidation_embedding_validation.py::test_create_observation_rejects_zero_length_embedding_before_insert tests/test_retain_orchestrator_mapping.py -q -o addopts=` → 11 passed
- `uv run ruff check tests/test_consolidation_embedding_validation.py hindsight_api/engine/retain/embedding_utils.py tests/test_retain_orchestrator_mapping.py` → passed
- Added-line secret/private marker scan → 0 findings

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/embedding_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/embedding_utils.py
@@ -15,9 +15,21 @@ class EmbeddingsBackend(Protocol):
     """Minimal duck-typed surface used by retain/recall — the concrete `Embeddings`
     ABC supplies default implementations that delegate to `encode()`."""
 
+    @property
+    def dimension(self) -> int: ...
+
     def encode_query(self, texts: list[str]) -> list[list[float]]: ...
 
     def encode_documents(self, texts: list[str]) -> list[list[float]]: ...
+
+
+def _validate_embedding_vector(vector: list[float], *, index: int, expected_dimension: int) -> list[float]:
+    actual_dimension = len(vector)
+    if actual_dimension == 0:
+        raise RuntimeError(f"embedding {index} has dimension 0; expected {expected_dimension}")
+    if actual_dimension != expected_dimension:
+        raise RuntimeError(f"embedding {index} has dimension {actual_dimension}; expected {expected_dimension}")
+    return vector
 
 
 def generate_embedding(
@@ -36,9 +48,19 @@ def generate_embedding(
     """
     try:
         embeddings = _encode_with_input_type(embeddings_backend, [text], input_type)
-        return embeddings[0]
     except Exception as e:
         raise Exception(f"Failed to generate embedding: {str(e)}")
+
+    if len(embeddings) != 1:
+        raise RuntimeError(
+            f"Embeddings backend returned {len(embeddings)} vectors for 1 input text; "
+            "expected exact 1:1 alignment"
+        )
+    return _validate_embedding_vector(
+        embeddings[0],
+        index=0,
+        expected_dimension=embeddings_backend.dimension,
+    )
 
 
 def _encode_with_input_type(
@@ -81,4 +103,7 @@ async def generate_embeddings_batch(
             "expected exact 1:1 alignment"
         )
 
-    return embeddings
+    return [
+        _validate_embedding_vector(embedding, index=index, expected_dimension=embeddings_backend.dimension)
+        for index, embedding in enumerate(embeddings)
+    ]
--- a/hindsight-api-slim/hindsight_api/engine/retain/embedding_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/embedding_utils.py
@@ -53,8 +53,7 @@ def generate_embedding(
 
     if len(embeddings) != 1:
         raise RuntimeError(
-            f"Embeddings backend returned {len(embeddings)} vectors for 1 input text; "
-            "expected exact 1:1 alignment"
+            f"Embeddings backend returned {len(embeddings)} vectors for 1 input text; expected exact 1:1 alignment"
         )
     return _validate_embedding_vector(
         embeddings[0],
--- a/hindsight-api-slim/tests/test_consolidation_embedding_validation.py
+++ b/hindsight-api-slim/tests/test_consolidation_embedding_validation.py
@@ -1,0 +1,41 @@
+import uuid
+
+import pytest
+
+from hindsight_api.engine.consolidation import consolidator
+
+
+class _ZeroLengthEmbeddings:
+    dimension = 384
+
+    def encode(self, texts):
+        assert texts == ["Consolidated observation text."]
+        return [[]]
+
+
+class _FakeMemoryEngine:
+    embeddings = _ZeroLengthEmbeddings()
+
+
+class _FailingConn:
+    async def fetchrow(self, *args, **kwargs):
+        raise AssertionError("zero-length embedding should be rejected before database insert")
+
+
+@pytest.mark.asyncio
+async def test_create_observation_rejects_zero_length_embedding_before_insert(monkeypatch):
+    source_id = uuid.uuid4()
+
+    async def fake_filter_live_source_memories(conn, bank_id, source_memory_ids):
+        return source_memory_ids
+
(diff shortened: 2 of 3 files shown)


nicoloboschi (commented):
A couple of must-fix items before this lands:

**1. The new `RuntimeError` is swallowed in `generate_embedding()`.** The single-text path still has `except Exception as e: raise Exception(f"Failed to generate embedding: {str(e)}")` wrapping the body, and `_validate_embedding_vector` is called inside that try. So a precise `RuntimeError("embedding 0 has dimension 0; expected 384")` becomes a plain `Exception("Failed to generate embedding: embedding 0 …")` — callers can't `except RuntimeError` to distinguish validation failures from backend transport errors. The batch path is fine because the list comprehension runs after the try/except. Easiest fix: move validation outside the try, or re-raise with `raise` / `raise … from e` without reclassifying.

**2. No test coverage for `generate_embedding()` (single-text).** All three new unit tests hit `generate_embeddings_batch`. The function you just modified has zero coverage of the new validation branch — which is also why issue #1 is invisible right now.

A few non-blocking nits I'd consider while you're in here:

- `embeddings_backend` is untyped and defensively probed. The `Embeddings` ABC in `engine/embeddings.py` already guarantees a `dimension: int` property; the `try/except` around `getattr(..., None)` plus `isinstance(..., int) and > 0` is paper over the missing type annotation. Typing the parameter and reading `.dimension` directly would be the right shape.
- The consolidation test (`test_consolidation_embedding_validation.py`) calls the private `_create_observation_directly`, monkey-patches the private `_filter_live_source_memories`, fakes `MemoryEngine` with a bare class, and uses a `_FailingConn` that asserts on `fetchrow`. The validation itself is already covered by the unit tests, and "consolidation uses the validating helper" is visible by inspection — this test will break on any consolidator refactor without catching anything the unit tests miss.
- Error message lacks input context: `"embedding 1 has dimension 2; expected 3"` tells you the batch offset but not the text, call site, or bank — correlating back in prod will require log fishing. A `logger.error(...)` with bank/operation/text length before raising would help.

ai-ag2026:
Addressed the review feedback on the latest head (`496ea737`):

- Moved single-text `generate_embedding()` validation outside the backend-call `try/except`, so validation failures now preserve their `RuntimeError` type instead of being reclassified as a generic `Exception`.
- Added single-text coverage for:
  - preserving validation `RuntimeError` on zero-length vectors
  - backend/vector count mismatch
  - `input_type="query"` using the query encoder path
- Rebased the branch onto current `origin/main`; the PR now reads back as mergeable.
- Typed the retain embedding backend protocol with the `dimension` property and validate against that directly.

Validation:
- `uv run pytest tests/test_retain_orchestrator_mapping.py tests/test_consolidation_embedding_validation.py -q` → 14 passed
- `uv run ruff check hindsight_api/engine/retain/embedding_utils.py tests/test_retain_orchestrator_mapping.py tests/test_consolidation_embedding_validation.py` → passed
- `uv run ruff format --check hindsight_api/engine/retain/embedding_utils.py tests/test_retain_orchestrator_mapping.py tests/test_consolidation_embedding_validation.py` → passed
- Hosted CI on the updated head → all required non-skipped jobs pass, including `verify-generated-files`, `build-api-python-versions` 3.11/3.12/3.13/3.14, `check-openapi-compatibility`, and `check-cli-coverage`.

Note: the full `./scripts/hooks/lint.sh` path attempted to run ESLint without the control-plane npm dependencies installed in this fresh worktree (`@eslint/js` missing). The touched files are Python-only, so I ran the focused Ruff checks above and let hosted CI verify generated-file cleanliness.

PR #1674, opened 2026-05-20
fix(worker): quarantine invalid async operations

### LINKED ISSUE #1671: Async operations need poison quarantine / degraded queue health
## Summary

During a post-crash Hindsight recovery incident, `/health` continued to report healthy while async worker lanes were degraded by deterministic poison operations.

The immediate data-shape bug is tracked separately in #1669 and a narrow validation fix is open in #1670. This issue is the operational follow-up: deterministic async failures should be quarantined or surfaced as degraded health/status instead of requiring SQL + journal inspection.

Observed failure classes from the incident:

```text
asyncpg.exceptions.DataError: different vector dimensions 384 and 0
index "idx_memory_links_to_type_weight" contains unexpected zero page at block 33008
HINT:  Please REINDEX it.
```

## What happened

After an unclean VM shutdown and WebUI session recovery, Hindsight had derived async work stuck/failing in the background queue. The REST API stayed healthy:

```json
{"status":"healthy","database":"connected"}
```

But worker logs and `async_operations` showed repeated deterministic failures and pending backlog. In the local cleanup pass, we had to manually:

- cancel active poison operations with `different vector dimensions 384 and 0`;
- reindex a crash-damaged PostgreSQL index;
- cancel `batch_retain` parent operations with `task_payload is null` that remained as pending queue work;
- restart the worker and watch `WORKER_STATS` until `global: pending=0`.

Final local state after manual cleanup:

```text
pending/processing async_operations: 0
active vector/zero-page errors: 0
WORKER_STATS ... global: pending=0 ... my_active: none
```

## Impact

- `/health` can report healthy while retain/consolidation work is functionally degraded.
- Deterministic failures can retry until max retries or keep reappearing as queue work.
- Operators cannot distinguish normal backlog from poison backlog without SQL and journal access.
- Null-payload parent operations can make the queue look pending even when there is no executable payload.
- The remediation path is operationally risky because it requires direct DB inspection/mutation.

## Suggested behavior

Add an explicit async-operation quarantine/degraded-status path:

1. Classify deterministic/non-transient failures after `N` retries or by known error classes:
   - invalid embedding/vector dimension;
   - schema/data-shape errors;
   - invalid/null payload operations;
   - corruption/index errors that require operator intervention;
   - oversized single-item payloads when splitting cannot reduce them.
2. Move those operations out of the active retry lane as `quarantined` or `failed_permanent`.
3. Store safe metadata, not retained content:
   - operation id;
   - operation type;
   - error class/signature;
   - retry count;
   - payload byte size / payload-null flag;
   - created/updated timestamps.
4. Expose queue degradation via `/health` or a dedicated status endpoint:
   - active poison/quarantined operation count;
   - oldest retrying operation age;
   - recent deterministic failure classes;
   - pending operations by type/status/payload-null;
   - whether reserved worker lanes are blocked.
5. Keep privacy intact: status output should never expose retained text or payload contents.

## Regression test ideas

- A retain/consolidation operation that repeatedly raises invalid vector dimension is moved out of active retry flow after the configured threshold.
- A null-payload parent `batch_retain` cannot remain indefinitely as normal pending executable work.
- `/health` or `/status` reports degraded queue state without leaking payload contents.
- Transient LLM/network failures still use ordinary retry/backoff.
- Quarantined operations do not block unrelated later retain operations.

## Relationship to existing work

- Complements #1669: zero-length embeddings reaching pgvector.
- Complements #1670: validates retain embedding dimensions before pgvector writes.
- Related to #1571: oversized single retain items can also become worker poison.

The key point: validation fixes individual poison sources; quarantine/status prevents the next deterministic poison source from silently degrading the worker lane.

ai-ag2026:
## Summary
- quarantine pending async operations that can never be claimed because `task_payload` is NULL
- preserve legitimate `batch_retain` parent rows while they still have child operations
- skip the PostgreSQL JSONB quarantine query entirely on non-PostgreSQL backends

## Why
Crash recovery can leave orphaned `batch_retain` parent rows in `async_operations` with `status='pending'` and `task_payload=NULL`. The worker claim path intentionally ignores NULL payloads because there is no executable task to dispatch, so these rows can sit in the pending lane forever and make queue health look like normal backlog.

## Test Plan
- `uv run ruff check hindsight_api/worker/poller.py tests/test_worker_quarantine.py`
- `uv run python -m py_compile hindsight_api/worker/poller.py tests/test_worker_quarantine.py`
- `git diff --check`
- `uv run --extra embedded-db pytest tests/test_worker_quarantine.py -q -o addopts=`
- `uv run --extra embedded-db pytest tests/test_worker_quarantine.py tests/test_worker.py -q -o addopts= -k 'not memory_engine_execute_task_passes_through_defer_operation'` -> 67 passed, 1 deselected

Fixes #1671
Complements #1670 by handling the queued poison rows separately from embedding-dimension validation.

### DIFF
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -195,6 +195,75 @@ class WorkerPoller:
         # Convert default schema to None for SQL compatibility (no prefix), keep others as-is
         return [t.schema if t.schema != DEFAULT_DATABASE_SCHEMA else None for t in tenants]
 
+    async def _quarantine_unclaimable_pending_operations(self, schemas: list[str | None]) -> int:
+        """Move deterministic null-payload poison rows out of the pending lane.
+
+        Worker claim queries intentionally skip ``task_payload IS NULL`` because
+        there is no executable task to hand to the executor. Most such rows are
+        legitimate ``batch_retain`` parent aggregators while their child rows
+        are still active. A null-payload row with no child references, or any
+        non-parent operation with a null payload, is unclaimable forever and
+        makes queue health look like ordinary backlog. Mark those rows failed so
+        the pending lane reflects claimable/retryable work only.
+        """
+        if self._backend.backend_type != "postgresql":
+            return 0
+
+        total = 0
+        error_message = "Quarantined unclaimable async operation: task_payload is NULL"
+        async with self._backend.acquire() as conn:
+            for schema in schemas:
+                table = fq_table("async_operations", schema)
+                try:
+                    result = await conn.execute(
+                        f"""
+                        UPDATE {table} AS op
+                        SET status = 'failed',
+                            error_message = $1,
+                            result_metadata = COALESCE(op.result_metadata, '{{}}'::jsonb) || jsonb_build_object(
+                                'quarantined', true,
+                                'quarantine_reason', 'task_payload_null',
+                                'quarantine_payload_null', true
+                            ),
+                            updated_at = now()
+                        WHERE op.status = 'pending'
+                          AND op.task_payload IS NULL
+                          AND (
+                            op.operation_type != 'batch_retain'
+                            OR NOT EXISTS (
+                              SELECT 1
+                              FROM {table} AS child
+                              WHERE child.bank_id = op.bank_id
+                                AND child.result_metadata::jsonb @> jsonb_build_object(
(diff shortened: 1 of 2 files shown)


nicoloboschi:
see https://github.com/vectorize-io/hindsight/issues/1671#issuecomment-4533379823

PR #1678, opened 2026-05-20
fix(control-plane): surface upstream errors via respondWithSdk helper

### LINKED ISSUE #1677: control-plane: SDK error responses logged as 'Value is not JSON serializable', lose upstream status + detail
## Summary

Control-plane API route handlers misreport upstream API failures as `TypeError: Value is not JSON serializable`. The misleading log makes triage harder, the response body always reads `"Failed to fetch stats"` regardless of the actual upstream error, and the HTTP status code is forced to `500` even when upstream returned `503`, `429`, `502`, etc. The pattern is systemic across most SDK-backed route handlers in `hindsight-control-plane`.

## Affected

`hindsight-control-plane/src/app/api/**/route.ts` — at least the following follow the same shape (sampled from `main`):

- [`stats/[agentId]/route.ts`](https://github.com/vectorize-io/hindsight/blob/main/hindsight-control-plane/src/app/api/stats/%5BagentId%5D/route.ts)
- [`operations/[agentId]/route.ts`](https://github.com/vectorize-io/hindsight/blob/main/hindsight-control-plane/src/app/api/operations/%5BagentId%5D/route.ts)
- [`recall/route.ts`](https://github.com/vectorize-io/hindsight/blob/main/hindsight-control-plane/src/app/api/recall/route.ts)
- [`banks/route.ts`](https://github.com/vectorize-io/hindsight/blob/main/hindsight-control-plane/src/app/api/banks/route.ts)
- plus ~20 more route bundles I counted by grepping `NextResponse.json(<var>.data,{status:200})` in the compiled `.next/server/chunks/*` set in the `0.6.0` image.

## Root Cause

The SDK returned by `@/lib/hindsight-client` follows the `@hey-api/client-fetch` convention: it returns `{ data, error, response }` on every call and **does not throw on non-2xx responses**.

The route handlers don't check `response.error` (or `response.response.ok`) before serializing. From [`stats/[agentId]/route.ts`](https://github.com/vectorize-io/hindsight/blob/main/hindsight-control-plane/src/app/api/stats/%5BagentId%5D/route.ts):

```ts
try {
  const { agentId } = await params;
  const response = await sdk.getAgentStats({
    client: lowLevelClient,
    path: { bank_id: agentId },
  });
  return NextResponse.json(response.data, { status: 200 });   // <-- response.data is undefined on 5xx
} catch (error) {
  console.error("Error fetching stats:", error);
  return NextResponse.json({ error: "Failed to fetch stats" }, { status: 500 });
}
```

Trace through a 5xx upstream response:

1. SDK call resolves to `{ data: undefined, error: ..., response: <Response status: 500> }`.
2. `NextResponse.json(undefined, ...)` invokes the spec'd `Response.json(undefined)`.
3. `Response.json(undefined)` throws `TypeError: Value is not JSON serializable` because `JSON.stringify(undefined) === undefined` (illegal as a JSON body).
4. The `catch` block runs, logs the **TypeError** (not the upstream failure), and returns a hard-coded 500.

The browser experience is mostly correct (gets a 500 + toast), but operators lose the real upstream error in the logs.

## Steps to Reproduce

1. Stand up `hindsight-control-plane` against any `hindsight-api` (compose, helm, or `docker run`).
2. Create or pick any bank — referred to below as `<bank>`.
3. Force the upstream API to return a non-2xx for `GET /v1/default/banks/<bank>/stats`. Two easy ways:
   - **Block at network layer**: scale `hindsight-api` to 0 replicas, or `iptables -A INPUT -p tcp --dport <api-port> -j REJECT`. Upstream call surfaces as connection refused / 502.
   - **Force a server-side 500**: send a request that triggers any DB error path (e.g. point the API at a Postgres whose `dynamic_shared_memory_type=posix` and `/dev/shm` is the K8s default 64 MiB, then call `GET /stats/<bank>` on a bank large enough to recruit parallel query workers — that aggregation will return `asyncpg.DiskFullError`. Or temporarily kill the DB connection mid-request.).
4. From the browser, load the dashboard / banks page that triggers `GET /api/stats/<bank>`.
5. Observe control-plane stderr:
   ```
   Error fetching stats: TypeError: Value is not JSON serializable
       at d.json (.next/server/chunks/[root-of-the-server]__0yxm3__._.js:8:3987)
       at E (.next/server/chunks/[root-of-the-server]__0e5z12m._.js:1:1440)
       at async l (.next/server/chunks/[root-of-the-server]__0e5z12m._.js:1:4799)
       at async i (.next/server/chunks/[root-of-the-server]__0e5z12m._.js:1:5840)
   ```
6. Observe `GET /api/stats/<bank>` returns HTTP `500` with body `{"error":"Failed to fetch stats"}` regardless of the actual upstream failure mode or status.

## Expected Behavior

The log should reflect the **upstream** failure (status + body), and the response should pass through the upstream status code and a useful detail field.

## Proposed Fix

Per-route, two-line check before serializing:

```ts
const response = await sdk.getAgentStats({
  client: lowLevelClient,
  path: { bank_id: agentId },
});

if (response.error || !response.data) {
  const upstreamStatus = response.response?.status ?? 502;
  console.error("Stats fetch failed:", {
    upstreamStatus,
    upstreamError: response.error,
  });
  return NextResponse.json(
    {
      error: "Failed to fetch stats",
      upstream: { status: upstreamStatus, detail: response.error },
    },
    { status: upstreamStatus },
  );
}

return NextResponse.json(response.data, { status: 200 });
```

Since the same pattern repeats across 20+ route files, a shared helper would prevent regression:

```ts
// src/lib/sdk-response.ts
import { NextResponse } from "next/server";

export function respondWithSdk<T>(
  result: { data?: T; error?: unknown; response?: Response },
  context: string,
) {
  if (result.error || !result.data) {
    const status = result.response?.status ?? 502;
    console.error(`${context} failed:`, { status, upstreamError: result.error });
    return NextResponse.json(
      { error: context, upstream: { status, detail: result.error } },
      { status },
    );
  }
  return NextResponse.json(result.data, { status: 200 });
}
```

Routes collapse to:
```ts
const response = await sdk.getAgentStats({ client: lowLevelClient, path: { bank_id: agentId } });
return respondWithSdk(response, "Failed to fetch stats");
```

## Impact

- **Browser UX:** unchanged — user still sees a generic error toast on failure.
- **Operator triage:** materially improved — logs surface the actual upstream error, status codes round-trip correctly, and 5xx vs 4xx vs network-error distinction returns.
- **Backwards compatibility:** response body shape gains a non-breaking `upstream` field; status code change (500 → real upstream status) could affect clients that key off the literal 500 in error paths, but that would be unusual for a control-plane UI route.

## Version

Reproduced on `ghcr.io/vectorize-io/hindsight-control-plane:0.6.0`. Source on `main` (confirmed today) is structurally identical.

offendingcommit:
Closes #1677.

## Problem

Every SDK-backed route handler in `hindsight-control-plane` had the same shape:

```ts
try {
  const response = await sdk.someMethod({ ... });
  return NextResponse.json(response.data, { status: 200 });
} catch (error) {
  console.error("Error fetching X:", error);
  return NextResponse.json({ error: "Failed to fetch X" }, { status: 500 });
}
```

The SDK (`@hey-api/client-fetch`) returns `{data, error, response}` and **does not throw on non-2xx**. When the upstream API 5xx'd, `response.data` was `undefined`, and Node's `Response.json(undefined)` throws `TypeError: Value is not JSON serializable`. The catch block logged that TypeError as if it were the actual failure, hiding the real upstream error, and hard-coding the response status to `500` regardless of whether upstream said `503`, `429`, `502`, etc.

## Fix

New helper `src/lib/sdk-response.ts::respondWithSdk(result, label, status?)`:

- Detects `result.error !== undefined || result.data === undefined`
- Logs the upstream status + upstream error detail
- Returns a `NextResponse` with **the upstream status code** (502 fallback for network-level failures with no Response object)
- Body shape on failure: `{error: <label>, upstream: {status, detail}}` — dashboards/clients now have the real upstream message
- On success, serializes `result.data` with the requested status (default 200; pass 201 for create endpoints)

17 SDK-backed routes refactored to use it. Routes that parse a request body keep a minimal try/catch around `await request.json()` and now return **400 on malformed JSON** (a small UX improvement over the prior 500).

## Out of scope (intentional)

- `documents/[documentId]/route.ts` PATCH and `banks/[bankId]/operations/[operationId]/route.ts` POST use raw `fetch()` rather than the SDK — they don't exhibit the bug and are unchanged.
- `banks/[bankId]/observations/route.ts` does post-fetch transformation of `response.data.items` and doesn't cleanly compose with the helper without restructuring — left for a separate PR.

## Tests

`hindsight-control-plane/src/lib/sdk-response.test.ts` — 12 vitest cases:

- Success: returns 200 with data, honors custom status (201), array data, empty-object data
- Failure: passes upstream 500/503/429 through, includes `upstream.detail` in body, logs to `console.error`, falls back to 502 when no `Response` object (network-level), treats `undefined data + undefined error` as failure rather than 200ing with `null`
- **Regression assertion**: `expect(() => respondWithSdk(failingResult, label)).not.toThrow()` — directly guards the `Response.json(undefined)` TypeError that started this whole thread

```
 Test Files  1 passed (1)
      Tests  12 passed (12)
```

## CI

Wired `npm test --workspace=hindsight-control-plane` into the existing `build-control-plane` and `build-hindsight-all` jobs in `.github/workflows/test.yml`. Tests run **before** the build step, so a regression breaks CI fast.

## Verification

```bash
# tests pass
cd hindsight-control-plane && npm test
# 12/12 green in ~150ms

# production build still succeeds
npm install --workspace=hindsight-clients/typescript
npm run build --workspace=hindsight-clients/typescript
cd hindsight-control-plane && npm run build
# next build + standalone bundle: success, all 25 routes compile
```

## Backwards compatibility

- **Browser happy path**: unchanged. Dashboard receives the same `response.data`.
- **Browser failure path**: response body grows a `upstream` field but keeps the same top-level `error` string, and the HTTP status code can now be the real upstream code (503, 429, 502) instead of always 500. Clients keying off literal status `500` would notice; the dashboard's `fetchApi` already toasts on any 4xx/5xx so no UI changes are needed.
- **Operator logs**: log line is `"<label>:"` with `{upstreamStatus, upstreamError}` payload instead of an arbitrary `Error` object. Materially clearer.


nicoloboschi (approved):
awesome, thanks

PR #1682, opened 2026-05-21
fix(embed): resolve npx absolute path on Windows before spawning UI

### LINKED ISSUE #1681: Windows: npx command not found when starting Control Plane UI
# Issue: `npx` command not found on Windows when starting Hindsight Control Plane UI (Local Mode)

## Environment
- **OS**: Windows 11
- **Shell**: Windows PowerShell / Git Bash (MSYS)
- **Node.js**: Installed at `C:\Program Files\nodejs\node.exe`
- **Package**: `hindsight-embed` v0.6.2 (via pip, **local mode** — not cloud)
- **Command**: `hindsight-embed -p hermes ui start`

## Problem

When running `hindsight-embed -p hermes ui start` on Windows in **local/embedded mode**, the daemon starts successfully, but the Control Plane UI fails to launch with:

```
─── ✗ Command Not Found ───
Command not found: npx
Full command: npx -y @vectorize-io/hindsight-control-plane@0.6.2 --port 19177 --hostname 0.0.0.0 --api-url http://127.0.0.1:9177

Install Node.js and npx to run the UI.
```

However, `npx` **is** installed and available in the user's PATH:
```powershell
> npx --version
10.x.x
```

## Root Cause

The issue is in `daemon_embed_manager.py::_find_ui_command()` which returns a bare `"npx"` string as the command:

```python
return ["npx", "-y", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
```

When `subprocess.Popen()` is called with `creationflags=subprocess.DETACHED_PROCESS | subprocess.CREATE_NEW_PROCESS_GROUP` on Windows, the detached process environment does **not** inherit the same PATH as the parent shell. This causes `npx` resolution to fail even though it's available in the user's terminal.

## Workaround

Manually running the full `npx` command in a separate terminal works:
```powershell
npx -y @vectorize-io/hindsight-control-plane@0.6.2 --port 19177 --hostname 0.0.0.0 --api-url http://127.0.0.1:9177
```

## Proposed Fix

Use `shutil.which("npx")` to resolve the absolute path to `npx` before spawning the subprocess:

```python
import shutil

def _find_ui_command(self) -> list[str]:
    """Find the command to run the control plane UI."""
    # ... dev mode check ...
    
    from . import __version__
    cp_version = os.getenv("HINDSIGHT_EMBED_CP_VERSION", __version__)
    
    npx_path = shutil.which("npx")
    if npx_path is None:
        # Fallback to bare command so FileNotFoundError handler can report it cleanly
        return ["npx", "-y", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
    return [npx_path, "-y", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
```

This ensures the subprocess receives the absolute path (e.g., `C:\Program Files\nodejs\npx.cmd`) instead of relying on PATH resolution in the detached process environment.

## Files Modified

- `hindsight_embed/daemon_embed_manager.py` — `_find_ui_command()` method

## Verification

After applying the fix locally:
1. `hindsight-embed -p hermes ui start` works end-to-end
2. Daemon starts in popup terminal (original behavior preserved)
3. UI starts in same terminal with Ctrl+C support
4. No more "Command not found: npx" error

tuancookiez-hub:
## Problem

On Windows 11, running `hindsight-embed -p <profile> ui start` in **local/embedded mode** fails with:

```
Command not found: npx
```

even when `npx` is installed and available in the user's PATH.

## Root Cause

`subprocess.Popen()` with `creationflags=DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` on Windows spawns a process that does **not** inherit the parent's PATH environment. The bare string `npx` cannot be resolved in the detached process context.

## Fix

Use `shutil.which("npx")` to resolve the absolute path to `npx` before spawning the subprocess. Falls back to the bare `npx` command so that existing `FileNotFoundError` error handlers can still report the missing command cleanly.

### Changes

- `hindsight-embed/hindsight_embed/daemon_embed_manager.py`: `_find_ui_command()` method

## Testing

- [x] Verified on Windows 11 with PowerShell / Git Bash
- [x] `npx` resolves to `C:\Program Files\nodejs\npx.cmd`
- [x] UI starts successfully after the fix
- [x] Fallback still produces clean error when `npx` is truly missing
- [x] Original behavior preserved: daemon starts in popup terminal, UI runs in foreground with Ctrl+C support

Fixes #1681

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -546,6 +546,8 @@ class DaemonEmbedManager(EmbedManager):
 
     def _find_ui_command(self) -> list[str]:
         """Find the command to run the control plane UI."""
+        import shutil
+
         # Check if we're in development mode (monorepo)
         dev_cp_path = Path(__file__).parent.parent.parent / "hindsight-control-plane"
         cli_js = dev_cp_path / "bin" / "cli.js"
@@ -559,7 +561,13 @@ class DaemonEmbedManager(EmbedManager):
         # `npx` prompts before installing missing packages on first run unless `-y` is set.
         # The UI starts in the background with stdout/stderr redirected to a log file, so an
         # interactive prompt would be invisible to users and the health-check loop would time out.
-        return ["npx", "-y", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
+        # On Windows, detached processes may not inherit the parent's PATH, so resolve the
+        # absolute path to npx to avoid "Command not found" errors.
+        npx_path = shutil.which("npx")
+        if npx_path is None:
+            # Fallback to bare command so the FileNotFoundError handler can report it cleanly
+            return ["npx", "-y", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
+        return [npx_path, "-y", f"@vectorize-io/hindsight-control-plane@{cp_version}"]
 
     def get_ui_url(self, profile: str, ui_port: int | None = None, hostname: str | None = None) -> str:
         """Get the URL for the UI serving this profile."""


benfrank241:
@tuancookiez-hub thanks for this. One small nit, could you move `import shutil` to the top of the file with the other imports?

nicoloboschi (approved):

PR #1683, opened 2026-05-21
fix(webhooks): avoid duplicate retain batch deliveries

### LINKED ISSUE #1642: [bug] retainBatch send too much webhook messages
### Bug Description

Problem with `retainBatch` and too much `webhooks` calls.

### Steps to Reproduce

```js
  const response = await client.retainBatch(
    bankId,
    items,
    {
      async: true
    }
  );
  console.log("response", response);
```

### Expected Behavior

Send a retainBatch command that calls a webhook with the same number of items.

### Actual Behavior

Send e.g. two items with retainBatch ends in four webhook calls, instead of two.
This makes an automatization with this behaviour difficult.

Image

case 1: single item => one single webhook call
case 2: two items => 2 * 2 webhook calls
case n: n items => n * n calls

### Version

0.6.1

### LLM Provider

Ollama

xmh1011:
## Summary
- Scope retain.completed outbox callbacks to each per-document retain group.
- Prevent retainBatch with multiple document_ids from replaying the full batch webhook outbox for every group.
- Add a regression test covering async retainBatch with two document_ids and one retain.completed webhook per item.

Closes #1642.

## Tests
- `uv run ruff check hindsight_api/engine/memory_engine.py hindsight_api/engine/retain/orchestrator.py tests/test_webhooks.py`
- `uv run ty check hindsight_api/engine/memory_engine.py hindsight_api/engine/retain/orchestrator.py`
- `HINDSIGHT_API_LLM_PROVIDER=none uv run pytest tests/test_webhooks.py::TestRetainCompletedWebhook -q`
- `./scripts/hooks/lint.sh`

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -5807,9 +5807,8 @@ def _register_routes(app: FastAPI):
                             strategy=group_strategy,
                             request_context=request_context,
                             return_usage=True,
-                            outbox_callback=app.state.memory._build_retain_outbox_callback(
+                            outbox_callback_factory=app.state.memory._build_retain_outbox_callback_factory(
                                 bank_id=bank_id,
-                                contents=contents,
                                 operation_id=None,
                                 schema=_current_schema.get(),
                             ),
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -207,6 +207,9 @@ from .search.reranking import CrossEncoderReranker, apply_combined_scoring
 from .search.tags import TagGroup, TagsMatch, build_tags_where_clause
 from .task_backend import TaskBackend
 
+RetainOutboxCallback = Callable[[asyncpg.Connection], Awaitable[None]]
+RetainOutboxCallbackFactory = Callable[[list[RetainContentDict]], RetainOutboxCallback | None]
+
 
 def _is_oracledb_connection_error(e: Exception) -> bool:
     """Check if an exception is an Oracle connection/interface error."""
@@ -789,9 +792,8 @@ class MemoryEngine(MemoryEngineInterface):
             request_context=context,
             operation_id=operation_id,
             strategy=strategy,
-            outbox_callback=self._build_retain_outbox_callback(
+            outbox_callback_factory=self._build_retain_outbox_callback_factory(
                 bank_id=bank_id,
-                contents=contents,
                 operation_id=operation_id,
                 schema=_current_schema.get(),
             ),
@@ -1314,10 +1316,10 @@ class MemoryEngine(MemoryEngineInterface):
     def _build_retain_outbox_callback(
         self,
         bank_id: str,
-        contents: list[dict],
+        contents: list[RetainContentDict],
         operation_id: str | None,
         schema: str | None = None,
-    ) -> "Callable[[asyncpg.Connection], Awaitable[None]] | None":
+    ) -> RetainOutboxCallback | None:
         """Build a transactional outbox callback for retain.completed webhook events.
 
         Returns a coroutine function that queues one webhook delivery row per content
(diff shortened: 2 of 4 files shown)


nicoloboschi (changes requested):
Thanks for the fix — the bug is real and the regression test pins the right invariant. Requesting changes on a few items before merge:

## Must address

- **Scope creep: unrelated docs** — `skills/hindsight-docs/references/developer/models.md` and `references/faq.md` add `opencode-go` to the provider list. This has nothing to do with the webhook bug. Please split into a separate PR.
- **Scope creep: pure formatting** — `hindsight_api/engine/providers/openai_compatible_llm.py` is a ruff-format whitespace-only change. Same for ~80 lines of whitespace reformatting in `tests/test_webhooks.py`. These obscure the real diff. Please drop or split.
- **Test fixture state leak** (`tests/test_webhooks.py`, in `test_async_batch_retain_queues_one_webhook_per_document`) — the test reassigns `memory_no_llm_verify._llm_config`, `_retain_llm_config`, and `_webhook_manager` directly on the fixture. If that fixture is not function-scoped, the mutation leaks into later tests. Please either confirm scope or restore the originals in a `finally` block.

## Should address

- **Pattern smell — `setattr` on the closure** (`engine/memory_engine.py`, `_build_retain_outbox_callback`) — attaching `for_contents` via `setattr` is opaque to the type checker, which is exactly why the consumer in `orchestrator.py` needs `getattr(...)` + a `callable()` guard to use it. Two cleaner options:
  1. Return a small dataclass/named callable exposing both `__call__` and `for_contents`.
  2. Pass a *factory* `Callable[[list[dict]], Callable | None]` through the orchestrator — it already knows when it's recursing per group, so the monkey-patched attribute disappears.

  Not a hard blocker, but the current shape will rot. If you don't want to address it now, please open a follow-up.

- **Truthiness check** (`engine/retain/orchestrator.py`, the new `outbox_for_contents = getattr(outbox_callback, "for_contents", None) if outbox_callback else None` line) — `Callable` objects can be truthy without being `None`; use `outbox_callback is not None` for precision.

## Nits

- The regression test asserts `sorted(doc_ids) == ["doc-1", "doc-2"]`, which would catch a duplicate `doc-1`. Adding an explicit `assert len(doc_ids) == len(set(doc_ids))` would more clearly document the "no duplicates" invariant.
- Confirm intent: per-document deliveries continue to share one `operation_id` after this change. That seems right (one logical batch op spans multiple docs), but worth flagging in case downstream consumers expect one op per doc.

xmh1011:
> Thanks for the fix — the bug is real and the regression test pins the right invariant. Requesting changes on a few items before merge:
> 
> ## Must address
> * **Scope creep: unrelated docs** — `skills/hindsight-docs/references/developer/models.md` and `references/faq.md` add `opencode-go` to the provider list. This has nothing to do with the webhook bug. Please split into a separate PR.
> * **Scope creep: pure formatting** — `hindsight_api/engine/providers/openai_compatible_llm.py` is a ruff-format whitespace-only change. Same for ~80 lines of whitespace reformatting in `tests/test_webhooks.py`. These obscure the real diff. Please drop or split.
> * **Test fixture state leak** (`tests/test_webhooks.py`, in `test_async_batch_retain_queues_one_webhook_per_document`) — the test reassigns `memory_no_llm_verify._llm_config`, `_retain_llm_config`, and `_webhook_manager` directly on the fixture. If that fixture is not function-scoped, the mutation leaks into later tests. Please either confirm scope or restore the originals in a `finally` block.
> 
> ## Should address
> * **Pattern smell — `setattr` on the closure** (`engine/memory_engine.py`, `_build_retain_outbox_callback`) — attaching `for_contents` via `setattr` is opaque to the type checker, which is exactly why the consumer in `orchestrator.py` needs `getattr(...)` + a `callable()` guard to use it. Two cleaner options:
>   
>   1. Return a small dataclass/named callable exposing both `__call__` and `for_contents`.
>   2. Pass a _factory_ `Callable[[list[dict]], Callable | None]` through the orchestrator — it already knows when it's recursing per group, so the monkey-patched attribute disappears.
>   
>   Not a hard blocker, but the current shape will rot. If you don't want to address it now, please open a follow-up.
> * **Truthiness check** (`engine/retain/orchestrator.py`, the new `outbox_for_contents = getattr(outbox_callback, "for_contents", None) if outbox_callback else None` line) — `Callable` objects can be truthy without being `None`; use `outbox_callback is not None` for precision.
> 
> ## Nits
> * The regression test asserts `sorted(doc_ids) == ["doc-1", "doc-2"]`, which would catch a duplicate `doc-1`. Adding an explicit `assert len(doc_ids) == len(set(doc_ids))` would more clearly document the "no duplicates" invariant.
> * Confirm intent: per-document deliveries continue to share one `operation_id` after this change. That seems right (one logical batch op spans multiple docs), but worth flagging in case downstream consumers expect one op per doc.

@nicoloboschi done

nicoloboschi (approved):

PR #1688, opened 2026-05-21
fix(embed): show actionable message when npx is missing on PATH

benfrank241:
## Summary
- When the UI launcher hits a system without `npx` (common on Windows without Node.js), `start_ui` no longer surfaces the generic red "Command Not Found" panel. Daemon stays healthy and usable; the panel now reframes the UI as optional with concrete next steps (Node.js download link + `hindsight-embed ui start` follow-up).
- Pre-flight check with `shutil.which` so we catch the missing binary up-front instead of relying on `FileNotFoundError` from `subprocess.Popen` — that path is brittle on Windows where PATH resolution can diverge between Python and the OS loader.

## Motivation
User report on the community channel: daemon started fine, then this hard-failure panel scared them off — they thought the whole setup was broken. It wasn't; only the UI was unavailable because Node.js wasn't installed.

## Test plan
- [x] `uv run pytest tests/test_embed_manager.py -v` — 12/12 pass, including new `test_start_ui_skips_spawn_when_npx_missing`
- [x] `./scripts/hooks/lint.sh` clean
- [ ] Manual smoke on a host without `npx`: confirm the new panel renders and the daemon URL is shown

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -9,6 +9,7 @@ import logging
 import os
 import platform
 import re
+import shutil
 import subprocess
 import sysconfig
 import time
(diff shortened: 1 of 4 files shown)


benfrank241:
Closing in favor of #1682 — that PR fixes the actual root cause (Windows detached subprocess not inheriting PATH), which is what TunaCookie originally hit. My messaging improvements here aren't worth carrying separately once the bug itself is gone.

PR #1709, opened 2026-05-22
fix: derive HINDSIGHT_API_HEALTH_URL default from HINDSIGHT_API_PORT

### LINKED ISSUE #1708: HINDSIGHT_API_HEALTH_URL default ignores HINDSIGHT_API_PORT — health check always hits port 8888
## Bug

When `HINDSIGHT_API_PORT` is set to any value other than 8888, the startup 
script still polls `http://localhost:8888/health` — because the default for 
`HINDSIGHT_API_HEALTH_URL` is hardcoded:

```bash
HINDSIGHT_API_HEALTH_URL="${HINDSIGHT_API_HEALTH_URL:-http://localhost:8888/health}"
```

The default does not reference `HINDSIGHT_API_PORT`, so the health check 
always targets port 8888 regardless of what port the API actually binds to.

## Impact

Container startup fails with:
```
API did not become healthy within 300s
```
…unless the user explicitly overrides `HINDSIGHT_API_HEALTH_URL` in their 
`docker-compose.yml`. This is a non-obvious requirement not documented 
anywhere.

## Expected behavior

The default should derive from `HINDSIGHT_API_PORT`:

```bash
HINDSIGHT_API_HEALTH_URL="${HINDSIGHT_API_HEALTH_URL:-http://localhost:${HINDSIGHT_API_PORT:-8888}/health}"
```

## Environment

- Docker standalone deployment
- `HINDSIGHT_API_PORT=8889` (custom port to avoid host conflict)
- `HINDSIGHT_API_HEALTH_URL` not set explicitly

sg-shag:
## Problem

`HINDSIGHT_API_HEALTH_URL` default is hardcoded to port 8888 and does not 
respect `HINDSIGHT_API_PORT`. When a custom port is set, the health check 
polls the wrong port and the container fails to start after 300s.

## Fix

One-line change in `docker/standalone/start-all.sh`:

```diff
-    API_HEALTH_URL="${HINDSIGHT_API_HEALTH_URL:-http://localhost:8888/health}"
+    API_HEALTH_URL="${HINDSIGHT_API_HEALTH_URL:-http://localhost:${HINDSIGHT_API_PORT:-8888}/health}"
```

No behavior change when `HINDSIGHT_API_PORT` is not set — defaults to 8888 as before.

Closes #1708

### DIFF
--- a/docker/standalone/start-all.sh
+++ b/docker/standalone/start-all.sh
@@ -156,7 +156,7 @@ PIDS=()
 # Start API if enabled
 if [ "$ENABLE_API" = "true" ]; then
     cd /app/api
-    API_HEALTH_URL="${HINDSIGHT_API_HEALTH_URL:-http://localhost:8888/health}"
+    API_HEALTH_URL="${HINDSIGHT_API_HEALTH_URL:-http://localhost:${HINDSIGHT_API_PORT:-8888}/health}"
     API_STARTUP_WAIT_SECONDS="${HINDSIGHT_API_STARTUP_WAIT_SECONDS:-300}"
 
     # Run API directly - Python's PYTHONUNBUFFERED=1 handles output buffering


benfrank241:
@sg-shag thanks for the contribution!

PR #1711, opened 2026-05-22
fix(litellm): injection_mode, context manager restore, validation, error consistency

DK09876:
## Summary

- **`configure()` footgun removed** — `configure()` without `bank_id` no longer silently uses `"default"`. `is_configured()` and `enable()` now both require an explicit `bank_id`.
- **`injection_mode` actually honoured** — `PREPEND_USER` now prepends memories to the last user message; previously all monkeypatch paths ignored the setting and always injected into a system message.
- **`hindsight_memory()` full restore** — context manager `finally` block now atomically restores all settings via `_restore_config()`. Previously 6 fields were silently lost on exit: `sync_storage`, `tags`, `recall_tags`, `recall_tags_match`, `reflect_context`, `reflect_response_schema`.
- **`defaults.query` wired up** — was dead code; `_inject_memories()` now uses it as a fallback query when no `hindsight_query` kwarg is passed.
- **`HindsightError` consistency** — `_inject_memories()` and both `log_pre_api_call` methods in `callbacks.py` now raise `HindsightError` instead of `ValueError` for missing `bank_id`.
- **Input validation** — `configure()` and `set_defaults()` validate `budget` and `recall_tags_match`, raising `ValueError` immediately on bad values. `document_id` emits `DeprecationWarning`.
- **Thread safety** — `_enabled_lock` and `_debug_lock` protect shared mutable state in `enable()`/`disable()` and `_last_injection_debug`.

## Test plan

- [x] `uv run pytest tests/test_integration.py -v` — 83 unit tests, all pass (added `TestValidation`, `TestInjectionMode`, `TestQueryField`, `TestHindsightErrorConsistency`, `TestContextManagerFullRestore`)
- [x] E2E suite against live Hindsight — 23 tests pass including new `TestDesignFixes` class that verifies each fix end-to-end


hindsight-integrations/litellm/hindsight_litellm/__init__.py
```diff
@@ -591,13 +637,17 @@ async def _wrapped_acompletion(*args, **kwargs):
     if model is None and len(args) > 0:
         model = args[0]
 
+    if _is_model_excluded(model, config):
+        return await _original_acompletion(*args, **kwargs)
+
     # Step 1: Inject memories (raises HindsightError on failure)
     if config and config.inject_memories and messages:
         try:
             injected_messages = _inject_memories(
```
benfrank241, line 643: async injection looks broken under a running loop: `_wrapped_acompletion` calls the sync `_inject_memories` (__init__.py:646), which drives the client via `run_sync`
  (__init__.py:348) / sync `client.recall` (__init__.py:395) — both raise inside a live loop (`_async.py:83`). not a regression (the old `get_event_loop().run_until_complete`
  failed the same way), and `test_async.py` only covers the primitives, not injection under `asyncio.run`. the async *storage* path already does it right via `run_in_executor`
  (wrappers.py:265) — injection should follow that.

benfrank241 (changes requested):
_debug_lock is dead code, never called (__init__.py:165, no with _debug_lock: anywhere; all _last_injection_debug writes unguarded).

PR #1714, opened 2026-05-23
fix: reload Codex OAuth credentials on auth failures

nicha16:
## Summary
- Reload Codex OAuth credentials from `auth.json` before using a cached refresh token after backend auth failures.
- Support shared auth-store documents that nest Codex credentials under `providers.openai-codex`.
- Avoid logging Codex account IDs during startup/verification.
- Add regression coverage for normal calls, tool-call requests, wrapped auth parsing, and wrapped auth persistence.

## Why
Long-running Hindsight services can keep a stale in-memory Codex access token while another process has already rotated the shared auth file. On a backend 401, spending the cached refresh token can fail or race even though fresh credentials are already available on disk. Reloading the auth file first lets the provider adopt externally rotated credentials before attempting OAuth refresh.

## Test plan
- `uv run pytest tests/test_codex_oauth_refresh.py tests/test_codex_tool_choice.py -q` — 30 passed
- `uv run ruff check hindsight_api/engine/providers/codex_llm.py tests/test_codex_oauth_refresh.py tests/test_codex_tool_choice.py` — passed
- `git diff --check` — passed
- Independent pre-commit review — passed

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
@@ -23,6 +23,7 @@ import os
 import tempfile
 import time
 import uuid
+from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
please resolve conflcits

nicoloboschi:
closing as stale

PR #1716, opened 2026-05-24
fix(openclaw): update Hindsight dependency ranges

de1tydev:
## Summary

- update the OpenClaw integration to depend on `@vectorize-io/hindsight-all` and `@vectorize-io/hindsight-client` `^0.6.2`
- update the Agent SDK dependency on `@vectorize-io/hindsight-client` to `^0.6.2`
- document the OpenClaw distinction between automatic recall injection and manual `agent_knowledge_recall`, including `max_tokens` vs `recallTopK`

## Why

`@vectorize-io/hindsight-openclaw@0.7.7` currently still installs `@vectorize-io/hindsight-all@0.1.0` because `^0.1.0` resolves to `<0.2.0`. That leaves the embedded OpenClaw path on an old Hindsight server package even though the current Hindsight packages are at `0.6.2`.

The Agent SDK also still pins the client range to `^0.5.6`, so consumers can end up with a mixed dependency tree even after the `agent_knowledge_recall` `max_results` -> `max_tokens` fix.

## Verification

- `cd hindsight-tools/hindsight-agent-sdk && npm test`
- `cd hindsight-tools/hindsight-agent-sdk && npm run build`
- `cd hindsight-integrations/openclaw && npx vitest run src/manifest.test.ts src/index.test.ts src/setup-lib.test.ts src/session-patterns.test.ts`

Note: full `cd hindsight-integrations/openclaw && npm test` still has an unrelated existing failure in `src/backfill.test.ts` (`treats a symlinked bin path as direct execution`). The targeted OpenClaw suites above pass.

Update: this PR now also exposes `agent_knowledge_reflect` through the OpenClaw knowledge tools. The new tool uses conservative defaults (`budget: low`, `max_tokens: 1024`, `fact_types: ["world", "experience", "observation"]`) and the docs call out that production banks should set a finite `reflect_source_facts_max_tokens` (for example 4096 or 8192) to avoid unbounded source-fact context during ad-hoc reflection.

### DIFF
--- a/hindsight-docs/docs-integrations/openclaw.md
+++ b/hindsight-docs/docs-integrations/openclaw.md
@@ -86,7 +86,7 @@ The plugin will automatically capture conversations after each turn and inject r
 
 **Auto-Recall:** Before each agent response, relevant memories are automatically injected into the context (up to 1024 tokens by default). The agent uses past context without needing to call tools.
 
-**Manual Recall Tools:** When `enableKnowledgeTools` is enabled, the plugin also exposes `agent_knowledge_recall` for explicit lookups. Its `max_tokens` parameter controls the recall response token budget; automatic injection uses `recallMaxTokens` and optionally `recallTopK` for a post-response count cap.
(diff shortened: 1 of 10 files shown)


nicoloboschi (approved):
LGTM

PR #1721, opened 2026-05-24
docs(hermes): document HINDSIGHT_API_LLM_MAX_CONCURRENT and add local-LLM concurrency guidance

### LINKED ISSUE #1720: docs(hermes): missing concurrency guidance for local LLM backends causes silent slot starvation
### Bug Description

When using Hindsight with a local LLM endpoint (llama-server, vLLM, lmstudio) that is also serving other clients — most commonly Hermes itself, when Hermes and Hindsight share the same LLM endpoint — the default `HINDSIGHT_API_LLM_MAX_CONCURRENT=32` saturates the endpoint's slot pool. Hermes appears to "freeze" or "hang on Ctrl+C", but it is actually queued waiting for an LLM slot that Hindsight has fully occupied with retain/recall/consolidation calls.

The Ollama blog post on the docs site already calls this out:

> "Set `HINDSIGHT_API_LLM_MAX_CONCURRENT=1`. Local models can't parallelize like cloud APIs."

But the Hermes integration page (`hindsight-docs/docs-integrations/hermes.md`) does not mention `HINDSIGHT_API_LLM_MAX_CONCURRENT` at all — it's missing from the "LLM Provider (local mode only)" config table, and there's no equivalent warning for local-LLM users.

This is a docs gap that produces a confusing failure mode for the increasingly common setup of one shared local llama-server.

### Steps to Reproduce

1. Set up Hermes Agent with Hindsight memory provider in local mode
2. Point Hindsight at a local llama-server with a small slot pool (e.g. 3 slots) that is also serving Hermes's main subagent — `HINDSIGHT_API_LLM_BASE_URL=http://127.0.0.1:8081/v1`
3. Use Hermes normally for several conversation turns so Hindsight kicks off retain/recall/consolidation work
4. Observe Hermes appears to "freeze" — new prompts don't reach the main agent
5. Check the llama-server: `curl -s http://127.0.0.1:8081/slots | python3 -m json.tool` — all slots show `is_processing: true`
6. Check connections: `ss -tnp 'dst :8081'` — ~32 ESTABLISHED connections from the `hindsight-api` process

### Expected Behavior

The Hermes integration docs document `HINDSIGHT_API_LLM_MAX_CONCURRENT` and warn users about the local-LLM concurrency caveat, the same way the Ollama blog post already does. Users running Hindsight against a shared local llama-server should be told to lower the default so the LLM endpoint isn't saturated.

### Actual Behavior

`HINDSIGHT_API_LLM_MAX_CONCURRENT` is missing from the "LLM Provider (local mode only)" env var table in `hindsight-docs/docs-integrations/hermes.md`. With the upstream default of 32 in effect, Hindsight saturates the LLM endpoint's slot pool, blocking other clients (Hermes itself) from getting a slot. The failure mode looks like Hermes freezing or hanging on Ctrl+C, which makes it hard to diagnose. Lowering `MAX_CONCURRENT` to 2 in `~/.hermes/.env` and restarting Hermes resolves the issue completely.

### Version

0.6.2

### LLM Provider

Other

GregChinn:
## What

- Adds `HINDSIGHT_API_LLM_MAX_CONCURRENT` to the LLM Provider env var table on the Hermes integration page
- Adds a `:::caution` block explaining the local-LLM concurrency caveat with concrete diagnostic commands

## Why

The default `MAX_CONCURRENT=32` is undocumented on this page and causes a confusing failure mode for users running Hindsight against a local LLM that's shared with Hermes's main agent (a very common local setup: one llama-server instance serving multiple roles). The Ollama blog post on this site already mentions the same caveat — this PR brings parity to the Hermes integration page.

## Context

Closes #1720

Verified locally: with the default in effect, the subagent at port 8081 saw 32 ESTABLISHED connections from `hindsight-api` and all 3 slots constantly `is_processing: true`, blocking the main Hermes agent until tasks completed. With `MAX_CONCURRENT=2` in `~/.hermes/.env`, Hindsight holds 2–3 connections and slots remain available for the main agent.

## No code change

Pure docs PR — no code changes, no tests required.

### DIFF
--- a/hindsight-docs/docs/developer/performance.md
+++ b/hindsight-docs/docs/developer/performance.md
@@ -84,6 +84,26 @@ Factors affecting throughput:
 - Database write performance
 - Available CPU/memory resources
 
+### Concurrency with Local LLM Backends
+
+When pointing Hindsight at a **local LLM server** (llama.cpp, vLLM, LM Studio) with a small fixed slot pool, the default `HINDSIGHT_API_LLM_MAX_CONCURRENT=32` will saturate the server and starve any other client sharing the endpoint.
+
+Lower the limit to match the server's capacity:
+
+```bash
+export HINDSIGHT_API_LLM_MAX_CONCURRENT=2
+```
+
+A value of `2` lets retain and consolidation run concurrently without blocking each other. If the endpoint is shared with other clients (other applications, agents, or workflows hitting the same llama-server / vLLM / LM Studio instance), reserve slots for them by lowering further — leave at least one slot free per shared client.
+
+**Symptom of saturation:** other clients sharing the endpoint appear to "hang" or queue indefinitely while Hindsight occupies all slots with retain/recall/consolidation calls. Verify with:
+
+```bash
+# llama.cpp default port is 8080; adjust if your server uses a different port
+curl -s http://127.0.0.1:8080/slots | jq '.[] | {id, is_processing, n_decoded}'
+ss -tnp 'dst :8080'   # count connections per client PID
+```
+
 ---
 
 ## Recall Performance


nicoloboschi (changes requested):
i'd only add a section if you run with local models, but not about hermes but in general - maybe in the Performance page

GregChinn:
Thanks for the review. Reworked per your feedback:

- Moved the section out of the Hermes integration page (reverted the hermes.md changes)
- Reframed as general local-LLM guidance (Performance > Retain > Concurrency with Local LLM Backends)
- No admonitions — matches the existing prose style of the Performance page
- Diagnostic commands use llama.cpp's default port 8080 with a note for users on custom ports

Happy to iterate further if anything else stands out.

nicoloboschi:
Superseded by #2002, which keeps this concurrency guidance (and the `/slots` + `ss` diagnostics) verbatim and expands it into a dedicated **Tuning for Local & Small Environments** section — adding per-operation concurrency caps, timeouts/retries, smaller-model & reasoning-effort advice, built-in llama.cpp tuning, CPU reranker knobs, and CPU embeddings. Suggest closing this in favor of #2002.

nicoloboschi:
Closing as superseded by #2002, which has now merged. #2002 keeps this PR's local-LLM concurrency guidance and expands it into a full **Tuning for Local & Small Environments** section (per-operation concurrency caps, timeouts/retries, smaller models + reasoning effort, built-in llama.cpp tuning, CPU reranker knobs incl. candidate-set sizing, and consolidation batch size). Thanks for the original write-up — it's the backbone of the new section.

PR #1731, opened 2026-05-25
fix(api): enable gzip middleware to keep graph payload parseable

nicoloboschi:
## Summary

Add starlette's `GZipMiddleware` to the FastAPI app. Three-line change to the `create_app` factory.

## Why this and not the clamp from #1605

I closed #1605 (cap `limit` server-side at 200) because it silently broke the Control Plane's "Load more" button on dense banks and assumed the bytes came from node count. They don't — they come from edges.

I ingested 19 LoCoMo sessions into a bank, profiled the `/banks/{bank_id}/graph` response at `limit=500`, and got:

```
nodes=491 edges=75718 table_rows=491

=== section sizes ===
  nodes:       194.8 KiB  (  0.9%)  avg=406 B/node
  edges:       21.2 MiB  ( 97.8%)  avg=293 B/edge
  table_rows:  282.4 KiB  (  1.3%)  avg=588 B/row

=== wire size ===
  raw (Accept-Encoding: identity): 21.7 MiB
  gzip-requested (wire):           1.6 MiB
  compression ratio:               13.8x
```

Edges are **98% of the payload** and gzip-compresses ~14× because the list is extremely repetitive: same JSON keys per edge, UUIDs sharing session prefixes, identical `linkType` / `color` / `lineStyle` strings. Node text, context, and `table_rows` combined are <3% of bytes — trimming them would be invisible.

1.6 MiB on the wire is comfortably under V8's ~512 MiB string-length cap that was breaking the Control Plane graph view in production. No UI change required; the Control Plane's `data-view.tsx` `useState(1000)` continues to work and "Load more" still does something.

## What this isn't

A real fix. The endpoint returns **75k edges from only 11.5k DB rows** — the bulk are observation-inferred links built in-memory and **uncapped** (`memory_engine.py:4880-4913`). Capping that at 10k like the direct links would cut edges ~85% and is the next PR. Tracking separately so this lands fast.

## Test plan

- [x] `ruff check`, `ruff format`, `ty check` clean.
- [x] Existing tests pass locally on the touched module (no behavioral changes besides response encoding).
- [x] Verified on a 491-node / 75k-edge bank: response `Content-Encoding: gzip` when client sends `Accept-Encoding: gzip`, decompresses to identical bytes as the `identity` variant.
- [ ] Reviewer: sanity-check that no internal callers send `Accept-Encoding: identity` and then try to parse `Content-Length` for sizing (none I can find).

## Notes

- `minimum_size=1024` skips compression on small responses where the per-response gzip overhead would dominate.
- One unrelated doc-skill regen got picked up by the pre-commit hook (upstream `alibaba` reranker docs were never synced to `skills/`). Three lines, same commit.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -15,6 +15,7 @@ from datetime import datetime, timezone
 from typing import Any, Literal
 
 from fastapi import Depends, FastAPI, File, Form, Header, HTTPException, Query, Request, UploadFile
+from fastapi.middleware.gzip import GZipMiddleware
 
 from hindsight_api.engine.audit import AuditEntry, AuditLogger
 from hindsight_api.extensions import AuthenticationError
@@ -2727,6 +2728,8 @@ def create_app(
     # This is required for mounted sub-applications where lifespan may not fire
     app.state.memory = memory
     app.state.audit_logger = memory.audit_logger
+
+    app.add_middleware(GZipMiddleware, minimum_size=1024)
 
     # ---------------------------------------------------------------------------
     # Patch OpenAPI schema: align ValidationError with Pydantic v2 error format
--- a/skills/hindsight-docs/references/developer/configuration.md
+++ b/skills/hindsight-docs/references/developer/configuration.md
@@ -611,7 +611,7 @@ Google's `gemini-embedding-001` produces 3072 dimensions natively but supports c
 
 | Variable | Description | Default |
 |----------|-------------|---------|
-| `HINDSIGHT_API_RERANKER_PROVIDER` | Provider: `local`, `tei`, `cohere`, `openrouter`, `zeroentropy`, `siliconflow`, `google`, `flashrank`, `litellm`, `litellm-sdk`, `jina-mlx`, or `rrf` | `local` |
+| `HINDSIGHT_API_RERANKER_PROVIDER` | Provider: `local`, `tei`, `cohere`, `openrouter`, `zeroentropy`, `siliconflow`, `alibaba`, `google`, `flashrank`, `litellm`, `litellm-sdk`, `jina-mlx`, or `rrf` | `local` |
 | `HINDSIGHT_API_RERANKER_LOCAL_MODEL` | Model for local provider | `cross-encoder/ms-marco-MiniLM-L-6-v2` |
 | `HINDSIGHT_API_RERANKER_LOCAL_MAX_CONCURRENT` | Max concurrent local reranking (prevents CPU thrashing under load) | `4` |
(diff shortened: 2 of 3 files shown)


bsbofmusic:
Follow-up from a native `v0.7.1` deployment: gzip + the v0.7 entity-link redesign improve the situation, but dense banks can still make the Control Plane effectively unusable because the frontend still fetches and processes a large graph.

Environment / setup:

- Hindsight API / Control Plane: `v0.7.1`
- Control Plane launched fresh via:
  - `npx -y @vectorize-io/hindsight-control-plane --hostname 127.0.0.1 --port 19999 --api-url <dataplane-api>`
- I explicitly reverted local shims before testing:
  - fresh npx cache
  - no patched chunks / alias chunks
  - no graph payload sanitization
  - no graph limit clamp
  - official client chunk still defaults the bank data page inner view to `useState("constellation")`

Observed on one production-ish bank:

```text
GET /api/graph?bank_id=<bank>&type=world&limit=1000
  raw bytes: 9,112,443
  nodes: 1000
  edges: 22,960
  table_rows: 1000

GET /v1/default/banks/<bank>/graph?type=world&limit=1000
  raw bytes: 9,143,173
  nodes: 1000
  edges: 22,960
  table_rows: 1000

GET /v1/default/banks/<bank>/graph?type=observation&limit=1000
  raw bytes: 11,086,481
  nodes: 667
  edges: 33,016
  table_rows: 667
```

Result: the localized bank data page returns HTML, but the browser still cannot reliably open/use the bank page on this dense bank. This is not the old materialized `link_type='entity'` table-bloat issue from pre-0.7 — this deployment is already on v0.7.1. The remaining problem seems to be that the Control Plane eagerly asks for graph-shaped data and then has to parse/render tens of thousands of edges even when the user primarily needs table rows.

I understand why #1605 was closed: a hard server-side `limit=200` clamp breaks the existing “Load more” UX and conflates table pagination with graph rendering. But gzip alone also does not fully address the browser-side parse/render cost after the response is decompressed, especially because edges dominate the payload.

Would you consider an upstream shape such as one of these?

1. Decouple table pagination from graph payloads: let the table view fetch `table_rows` without `nodes` / `edges`.
2. Add a Control Plane/API option like `graph=false`, `include_graph=false`, `view=table`, or a CP env flag to disable eager graph fetch/render on dense banks.
3. Add an edge-aware graph budget/pagination separate from the row `limit` — e.g. `node_limit`, `edge_limit`, or sampled graph mode.
4. Default dense banks to table/timeline first and lazy-load constellation/graph only when the user explicitly opens it.

Happy to open a separate issue if that is preferable; I’m adding this here because #1731 is the merged follow-up to #1605 and the current v0.7.1 behavior still hits the same Control Plane usability class, just at a smaller payload than the 596 MiB example in #1605.

PR #1732, opened 2026-05-25
docs(extensions): document OperationValidatorExtension.precheck hook

r266-tech:
#1548 added an optional `precheck()` hook and exported `PrecheckContext` on `OperationValidatorExtension`, but `developer/extensions.md` only documents the post-body `validate_*` hooks and `DeferOperation`. Users can't discover the new pre-body-parse gate from the docs.

This adds a **Pre-body-parse precheck** subsection covering when to use the hook (gate a request before the body is read — exhausted balance, revoked key, rate-limited tenant), the `PrecheckContext` fields (`operation` / `bank_id` / `request_context`), and an opt-in example. Wording is grounded in the method/dataclass docstrings from #1548.

Mirrored into the `skills/hindsight-docs/references` copy to keep the two in sync. Pure docs, no code changes.


nicoloboschi:
no need thanks

PR #1736, opened 2026-05-25
fix(retain): split oversized single items in batch retain (#1571)

### LINKED ISSUE #1571: Sub-batch splitter announces 10K-token chunks but processes full payload as 1/1, triggering OOM on large single-item retains

nicoloboschi:
## Summary

Closes #1571.

The batch-retain splitter in `memory_engine.py` packed contents by token count but **never chunked an individual item that already exceeded the per-batch budget**. A single 1.17M-token retain went through as `Split into 1 sub-batches: [1] items each`, contradicting the "splitting into ~10K-token sub-batches" log message and OOM-killing the orchestrator under realistic memory limits.

### Root cause

The pack-by-item logic in both `_retain_batch_async` (~line 2429) and `submit_async_retain` (~line 9237) only started a new sub-batch when `current_batch` was non-empty:

```python
if current_batch and current_batch_tokens + item_tokens > tokens_per_batch:
    sub_batches.append(current_batch)
    current_batch = [item]
    ...
else:
    current_batch.append(item)  # ← oversize item always lands here on first iteration
```

So a single oversize item passes through whole, regardless of the `tokens_per_batch` budget.

### Fix

- Add `_split_contents_into_sub_batches(contents, tokens_per_batch) -> _SubBatchSplit` to `memory_engine.py`. When a single item exceeds the budget, chunk its content via `fact_extraction.chunk_text` (paragraph / sentence aware, or conversation-turn aware for JSON arrays) using a `tokens_per_batch * 3` char budget, and emit each chunk as its own single-item sub-batch.
- The dataclass carries an `origin_indices` mapping so `retain_batch_async` can merge results from chunked sub-batches back into a single per-input result list — preserving the existing public contract that `retain_batch_async` returns one `list[unit_id]` per input content.
- Refactor both splitter call sites to use the shared helper so they can't drift again.
- Compact the "Split into N sub-batches" log when an oversize item fans out into many `[1]`-item sub-batches.

### Known limitation

When a single document is chunked across multiple sub-batches, the document row's `original_text` / `content_hash` reflects the **last** chunk processed, because each non-first sub-batch's `handle_document_tracking` upserts the row with its own slice. This is acceptable vs the prior OOM + restart-loop behavior; a complete fix would require plumbing a "carry-the-full-content" override through the orchestrator and is left for a follow-up.

Adjacent fixes suggested in #1571 (stale-task quarantine, pool-starvation circuit breaker, body-validating healthcheck) are out of scope here and worth separate issues.

## Test plan

- [x] `uv run pytest tests/test_batch_chunking.py tests/test_async_batch_retain.py` — 23 passed (4 new regression tests + existing `test_large_async_batch_auto_splits`, `test_large_batch_auto_chunks`).
- [x] `./scripts/hooks/lint.sh` — All lints passed.
- [x] `uv run ty check hindsight_api/engine/memory_engine.py` — All checks passed.
- [ ] Reproduce with a ≥500K-token single retain payload in a memory-limited container and confirm no OOM (manual / staging).

### New regression tests

- `test_split_single_oversized_item_produces_multiple_sub_batches` — single 25k-token item with `tokens_per_batch=1000` produces > 1 sub-batches, each at or under the budget.
- `test_split_oversized_item_preserves_document_id_and_metadata` — chunked sub-batches inherit `document_id`, `context`, `tags`.
- `test_split_mixed_batch_chunks_only_oversized_items` — small items pack normally; only the oversize item fans out.
- `test_split_small_batch_returns_single_sub_batch` — under-budget batches still produce one sub-batch.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -1697,9 +1697,7 @@ def create_cross_encoder_from_env() -> CrossEncoderModel:
     elif provider == "alibaba":
         api_key = config.reranker_alibaba_api_key
         if not api_key:
-            raise ValueError(
-                f"{ENV_RERANKER_ALIBABA_API_KEY} is required when {ENV_RERANKER_PROVIDER} is 'alibaba'"
-            )
+            raise ValueError(f"{ENV_RERANKER_ALIBABA_API_KEY} is required when {ENV_RERANKER_PROVIDER} is 'alibaba'")
         return AlibabaCloudCrossEncoder(
             api_key=api_key,
             model=config.reranker_alibaba_model,
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -227,6 +227,90 @@ def _is_oracledb_integrity_error(e: Exception) -> bool:
     return isinstance(e, oracledb.IntegrityError)
 
 
+@dataclass
+class _SubBatchSplit:
+    """Result of packing retain contents into sub-batches.
+
+    ``sub_batches[i]`` is a list of RetainContentDict items that should
+    be processed together. ``origin_indices[i]`` lists the indices into
+    the original ``contents`` list that contributed items to
+    ``sub_batches[i]``; callers that present per-input results to the
+    user (such as ``retain_batch_async``) use this mapping to merge
+    results belonging to the same original content back together when
+    an oversized item was chunked across multiple sub-batches.
+    """
+
+    sub_batches: list[list[RetainContentDict]]
+    origin_indices: list[list[int]]
+
+
+def _split_contents_into_sub_batches(
+    contents: list[RetainContentDict],
+    tokens_per_batch: int,
+) -> _SubBatchSplit:
+    """Pack retain contents into sub-batches whose combined token count
+    stays at or below ``tokens_per_batch``.
+
+    Any single item that already exceeds the budget is chunked via
+    ``fact_extraction.chunk_text`` (paragraph/sentence aware, or
+    conversation-turn aware for JSON arrays) and each chunk becomes its
+    own single-item sub-batch. Without this, an oversized single item
+    would pass through as a ``1/1`` sub-batch holding the entire
+    payload — which contradicts the splitter's log and lets the
+    orchestrator OOM under realistic memory limits (see issue #1571).
+    """
+    from .retain import fact_extraction
+
+    # chunk_text takes a char budget; cl100k_base averages ~3-4 chars
+    # per token on natural-language input. Use 3x for headroom so a
+    # chunk's token count is comfortably under tokens_per_batch even
+    # when content tokenizes denser than average (code, JSON).
+    char_budget = max(tokens_per_batch * 3, 1)
+
+    sub_batches: list[list[RetainContentDict]] = []
+    origin_indices: list[list[int]] = []
+    current_batch: list[RetainContentDict] = []
+    current_batch_origins: list[int] = []
+    current_batch_tokens = 0
+
+    def _flush() -> None:
+        nonlocal current_batch, current_batch_origins, current_batch_tokens
+        if current_batch:
+            sub_batches.append(current_batch)
+            origin_indices.append(current_batch_origins)
+            current_batch = []
+            current_batch_origins = []
+            current_batch_tokens = 0
+
+    for original_idx, item in enumerate(contents):
+        content_str = item.get("content", "") or ""
+        item_tokens = count_tokens(content_str)
+
+        if item_tokens > tokens_per_batch:
+            # Oversized single item: flush anything in flight, then
+            # chunk the content and emit each chunk as its own
+            # single-item sub-batch. The sub-batches share the
+            # original item's document_id and metadata so the
+            # orchestrator's first-batch document tracking still
+            # cascade-deletes the prior document version on slice 1.
+            _flush()
+            chunks = fact_extraction.chunk_text(content_str, char_budget)
+            for chunk in chunks:
+                chunk_item = cast(RetainContentDict, {**item, "content": chunk})
+                sub_batches.append([chunk_item])
+                origin_indices.append([original_idx])
+            continue
+
+        if current_batch and current_batch_tokens + item_tokens > tokens_per_batch:
+            _flush()
+        current_batch.append(item)
+        current_batch_origins.append(original_idx)
+        current_batch_tokens += item_tokens
+
+    _flush()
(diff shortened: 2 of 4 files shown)


nicoloboschi:
First CI pass — 5 failures, breakdown:

**Fixed in 8f46b4e4:**
- `test-api` → `test_op_cancellation::test_retain_stops_between_sub_batches_when_cancelled` — real regression from this PR. The cancellation early-return now returns a per-input result list (len = len(contents)) with empty slots for un-processed inputs, instead of a truncated list. Test updated to count non-empty results instead.
- `verify-generated-files` — `cross_encoder.py` ruff reformat picked up by the lint hook.

**Pre-existing on `main`, not addressed here:**
- `test-api` → `test_batch_api::test_batch_api_fallback_unsupported_provider` — tests a "fallback to sync" path that #1463 removed (it now raises `RuntimeError`). Identical to the version on `main`.
- `test-embed` / `test-embed-windows` → `test_embed_manager::test_find_ui_command_uses_npx_yes_flag_for_published_control_plane` — unrelated npx-resolution assertion failure in `hindsight-embed`.
- `LLM acceptance (vertexai/gemini-2.5-flash-lite)` — only vertexai failed; the other 5 LLM acceptance jobs (gemini, openai, groq, litellm, bedrock) all passed → flake.

Re-running CI.

PR #1756, opened 2026-05-26
ci: pin astral-sh/setup-uv to immutable v8.1.0

nicoloboschi:
## Summary
- Deploy Docs CI on main is failing repeatedly with `An action could not be found at the URI 'https://codeload.github.com/astral-sh/setup-uv/tar.gz/37802adc...'`. The codeload mirror can't serve the v7.6.0 SHA that `@v7` currently resolves to — four consecutive reruns of [run 26447835312](https://github.com/vectorize-io/hindsight/actions/runs/26447835312) failed across regions while curling the same URL returns 200, so it's a sticky cache issue on GitHub's side.
- Per the [v8.0.0 release notes](https://github.com/astral-sh/setup-uv/releases/tag/v8.0.0) setup-uv dropped floating major/minor tags in favor of immutable releases for supply-chain safety, so pinning to a concrete version is the recommended path forward.
- Bumps all 38 `astral-sh/setup-uv@v7` references to `@v8.1.0`. We only pass `enable-cache` / `prune-cache`, which v8 still supports — the only v8 breaking change was the `manifest-file` format, which we don't use.

## Test plan
- [ ] Deploy Docs workflow on this branch (triggered manually via workflow_dispatch after merge, since `deploy-docs.yml` only runs on push to main)
- [ ] PR-triggered test workflows run successfully on this branch

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -27,7 +27,7 @@ jobs:
           node-version: 20
           cache: npm
           cache-dependency-path: package-lock.json
-      - uses: astral-sh/setup-uv@v7
+      - uses: astral-sh/setup-uv@v8.1.0
       - run: npm ci --workspace=hindsight-docs
       - run: uv run generate-llms-full
       - run: npm run build --workspace=hindsight-docs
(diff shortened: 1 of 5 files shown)


nicoloboschi:
Closing — this isn't a code fix. There's an active GitHub incident (Actions + Pages, critical: "authentication issues leading to failure in starting Actions runs and downloading actions" — https://stspg.io/6gzfzr1dx684) and the same codeload error reproduces with v8.1.0. Will rerun the failed Deploy Docs run once GitHub resolves the incident.

PR #1760, opened 2026-05-26
docs: clarify region for S3-compatible file storage providers

davidmyriel:
The MinIO example in the S3 file storage section omits S3_REGION, but the configuration table above marks region as required. Non-AWS users reading the example hit the same "what do I put here?" confusion that supermemoryai/supermemory#851 fixed: auto works for SigV4 signing.
This PR:

Adds a one-line note that region: 'auto' works for S3-compatible providers without AWS-style regions
Updates the MinIO example to set it explicitly (matching the note)
Adds Tigris alongside MinIO and R2 in the providers list, the table description, and as a config example

No code changes. obstore's S3Store works against Tigris's endpoint (https://t3.storage.dev) via the existing endpoint override.
Disclosure: I work at Tigris Data — happy to revise the framing or drop the Tigris example if you'd prefer the docs stay narrower. The region: 'auto' clarification stands on its own.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/storage/s3.py
+++ b/hindsight-api-slim/hindsight_api/engine/storage/s3.py
@@ -16,7 +16,7 @@ class S3FileStorage(FileStorage):
     S3-compatible object storage backend.
 
     Uses obstore (Rust-backed) for high-throughput async access to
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
thanks

PR #1770, opened 2026-05-26
[feature] add zeroentropy embeddings provider

flamerged:
## Summary

This adds first-class ZeroEntropy `zembed-1` embeddings support to Hindsight.

`zembed-1` is a state-of-the-art retrieval embedder, and this integration lets Hindsight use it natively instead of forcing users through an OpenAI-compatible shim or LiteLLM proxy. It also preserves ZeroEntropy's asymmetric retrieval semantics: retained memory content is embedded as `document` input, while recall/search text is embedded as `query` input.

## Why this matters

- **Better retrieval stack:** Hindsight already supports ZeroEntropy reranking; this completes the native ZeroEntropy retrieval path with `zembed-1` embeddings.
- **No proxy tax:** Users can point Hindsight directly at ZeroEntropy's `/v1/models/embed` endpoint, with no shim layer translating away provider-specific features.
- **Correct asymmetric embeddings:** ZeroEntropy exposes separate query/document modes. Native support lets Hindsight use those modes directly instead of treating all text the same.
- **Production-friendly dimensions:** Hindsight defaults `zembed-1` to 1280 dimensions so it works with pgvector HNSW's 2000-dimension index limit out of the box, while still allowing 2560/1280/640/320/160/80/40 for deployments that want different tradeoffs.

## What changed

- Added a native `ZeroEntropyEmbeddings` provider for `zembed-1`.
- Added `HINDSIGHT_API_EMBEDDINGS_ZEROENTROPY_*` config/env settings for API key, model, base URL, dimensions, encoding format, latency, and batch size.
- Routed retained content through `encode_documents()` and recall/search text through `encode_query()` when the embeddings backend supports asymmetric modes.
- Added support for both `float` and `base64` ZeroEntropy embedding responses, normalizing both to float vectors before storage.
- Fixed the existing ZeroEntropy reranker factory to honor its configured base URL.
- Updated env examples, docs, and generated docs-skill references.
- Added focused coverage for config parsing, provider factory wiring, request payloads, response decoding, query/document routing, and reranker base URL wiring.

## Validation

- `uv run pytest tests/test_zeroentropy_embeddings.py -q` -> 12 passed
- `uv run ruff check hindsight_api tests/test_zeroentropy_embeddings.py` -> passed
- `uv run ty check hindsight_api/` -> passed
- `./scripts/hooks/lint.sh` -> passed
- CodeRabbit CLI review final pass -> 0 findings
- Live ZeroEntropy API smoke test: `document` and `query` embeddings returned 1280-dimensional float vectors
- Live retain/recall smoke test through `MemoryEngine` against a scratch pgvector database: retained 1 memory unit and recalled 1 result with `embedding_dim=1280`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,7 +65,7 @@ HINDSIGHT_API_LOG_LEVEL=info
 # HINDSIGHT_API_VECTOR_EXTENSION=pgvectorscale  # Auto-detects pg_diskann on Azure
 
 # Embeddings Configuration (Optional - uses local by default)
-# Provider: "local" (default), "tei", "openai", "cohere", "google", "openrouter", "litellm", or "litellm-sdk"
+# Provider: "local" (default), "tei", "openai", "cohere", "google", "openrouter", "zeroentropy", "litellm", or "litellm-sdk"
 # HINDSIGHT_API_EMBEDDINGS_PROVIDER=local
 # For local provider:
 # HINDSIGHT_API_EMBEDDINGS_LOCAL_MODEL=BAAI/bge-small-en-v1.5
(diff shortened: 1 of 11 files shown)


nicoloboschi (approved):
LGTM. Solid integration — Pydantic-typed request/response, asymmetric query/document split with a safe default on the base `Embeddings` class so existing providers don't need touching, and base64 response decoding handled. Bonus catch on the reranker `base_url` not being threaded through the factory.

Cleanups (encoding_format fallback dead branch, duplicated dimension validation, unused `usage` field, getattr-vs-Protocol mismatch in `embedding_utils`, aligning URL handling with the existing `ZeroEntropyCrossEncoder`) will land in a follow-up PR.

PR #1788, opened 2026-05-27
fix(api): anchor recall recency to query timestamp

Sanderhoff-alt:
## Summary

This PR makes recall recency scoring honor `query_timestamp`.

Before this change, `query_timestamp` already anchored temporal retrieval for relative time expressions, but combined scoring still calculated recency against the server's current wall-clock time. That means historical replay, offline evaluation, and imported conversation logs could incorrectly penalize memories that were recent at the time the user asked the question.

With this change, recall uses the public `query_timestamp` parameter, parsed internally as `question_date`, as the reference time for combined scoring. If no `query_timestamp` is provided, behavior remains unchanged and scoring falls back to the server's current UTC time.

## What Changed

- Added a recall scoring time helper that:
  - uses the public `query_timestamp` value, parsed internally as `question_date`, when provided
  - normalizes timezone-aware timestamps to UTC
  - treats naive timestamps as UTC
  - falls back to `utcnow()` when no query timestamp is provided
- Changed recall combined scoring to use that query-time anchor instead of always using `utcnow()`.
- Updated API, Python client, TypeScript generated types, OpenAPI, MCP docs, and developer docs to document that `query_timestamp` anchors both:
  - relative temporal expressions
  - recency scoring
- Added recall-level tests to verify combined scoring receives the query-time anchor.
- Regenerated docs skill references.

## Why

`query_timestamp` is documented as "recall as if asking at this point in time". Since Hindsight already uses it to anchor temporal retrieval, recency scoring should use the same query-time anchor instead of switching back to server wall-clock time.

For example, if a benchmark replays a query from May 30, 2023, a memory from May 22, 2023 should be considered recent. Previously, if the server processed that replay in 2026, the same memory would be treated as more than a year old and receive the minimum recency score.

This is especially important for:

- historical conversation replay
- offline benchmarks
- migrated/imported agent logs
- tests that provide `query_timestamp` to reproduce past recall behavior

## Compatibility

This is backward compatible for live recall requests that do not pass `query_timestamp`: those still use the server's current UTC time for recency scoring.

The only behavior change is for requests that explicitly provide `query_timestamp`; those now consistently use it for both temporal retrieval and recency scoring.

## Tests

Ran:

```bash
uv run pytest \
  hindsight-api-slim/tests/test_recall_scoring_time.py \
  hindsight-api-slim/tests/test_combined_scoring.py
```

Result:

```text
21 passed
```

Also ran:

```bash
uv run pytest \
  hindsight-api-slim/tests/test_mcp_tools.py::TestRecallNewParams::test_recall_with_query_timestamp -q
```

Result:

```text
1 passed
```

Docs/generated checks:

```bash
./scripts/generate-openapi.sh
./scripts/generate-clients.sh
./scripts/generate-docs-skill.sh
./scripts/hooks/generate-docs-skill.sh
./scripts/hooks/lint.sh
npm run build -w hindsight-docs
git diff --check
git diff --cached --check
```

The docs build passed. It still reports existing Docusaurus warnings about a blog truncation marker and two broken anchors in an unrelated version blog page.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -152,7 +152,11 @@ class RecallRequest(BaseModel):
     max_tokens: int = 4096
     trace: bool = False
     query_timestamp: str | None = Field(
-        default=None, description="ISO format date string (e.g., '2023-05-30T23:40:00')"
+        default=None,
+        description=(
+            "ISO format date string (e.g., '2023-05-30T23:40:00'). Used as the query-time anchor for "
+            "relative temporal expressions and recency scoring."
+        ),
     )
     include: IncludeOptions = FieldWithDefault(
         IncludeOptions,
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -396,6 +396,15 @@ def utcnow():
     return datetime.now(UTC)
 
 
+def _recall_scoring_now(question_date: datetime | None) -> datetime:
+    """Return the reference time for recall scoring boosts."""
+    if question_date is None:
+        return utcnow()
+    if question_date.tzinfo is None or question_date.utcoffset() is None:
+        return question_date.replace(tzinfo=UTC)
+    return question_date.astimezone(UTC)
+
+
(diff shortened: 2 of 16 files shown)


nicoloboschi (approved):
nice catch!!

PR #1798, opened 2026-05-27
fix(worker): wire HINDSIGHT_API_WORKER_MAX_RETRIES into task retry decision

aaronwestphal:
## Summary

The `HINDSIGHT_API_WORKER_MAX_RETRIES` env var has been declared at
`hindsight-api-slim/hindsight_api/config.py:433` since the worker was
introduced, but the actual retry decision in `MemoryEngine.execute_task`
hardcodes `if retry_count < 3` and never reads the knob. Operators who
set the env var to extend the retry window during a provider outage see
no effect.

This PR wires the existing knob into the retry check, and adds a sibling
`HINDSIGHT_API_WORKER_TASK_RETRY_BACKOFF_SECONDS` (default `60`) for the
hardcoded 60-second backoff interval at the same site.

## Why

The retry loop in `MemoryEngine.execute_task` (lines 1383-1428 on `main`,
commit `78c3525`) currently looks like:

```python
retry_count = task_dict.get("_retry_count", 0)
if retry_count < 3:                                                    # hardcoded
    raise RetryTaskAt(
        retry_at=datetime.now(UTC) + timedelta(seconds=60),            # hardcoded
        message=str(e),
    )
raise
```

That's a 3 × 60s = 4-minute total window before a task is permanently
marked `failed` and requires manual SQL requeue to recover. The
declared `HINDSIGHT_API_WORKER_MAX_RETRIES` knob doesn't intervene
anywhere along this path — verified by grep of the whole repo.

After this PR, an operator can extend the window for a known outage by
setting:

```env
HINDSIGHT_API_WORKER_MAX_RETRIES=10
HINDSIGHT_API_WORKER_TASK_RETRY_BACKOFF_SECONDS=300
```

→ 10 × 5min = ~50min retry window, no code change required.

## Behavior

| | Before | After (defaults) | After (operator-tuned) |
|---|---|---|---|
| max retries | 3 (hardcoded) | 3 (env or default) | configurable via env |
| backoff (seconds) | 60 (hardcoded) | 60 (env or default) | configurable via env |
| total retry window | ~4 min | ~4 min | configurable |

Both env vars are read **on each retry decision** (not cached at process
start) so operators can tune the policy during an active outage without
restarting workers. Default values preserve existing behavior exactly.

## Tests

`hindsight-api-slim/tests/test_worker_retry_knobs.py` adds 4 regression
tests:

- `test_retry_count_cap_honors_env_var` — setting MAX_RETRIES=5 yields a
  5-attempt retry window (was: 3, ignored the env var)
- `test_retry_stops_at_env_var_cap` — at MAX_RETRIES, the next failure
  marks the op `failed` and stops retrying
- `test_retry_backoff_honors_env_var` — BACKOFF_SECONDS=120 schedules
  the next attempt 120s out (was: 60, hardcoded)
- `test_defaults_preserve_existing_behavior` — with neither env var set,
  the loop still uses 3 attempts × 60s

```
$ uv run pytest tests/test_worker_retry_knobs.py -v
======================== 4 passed in 187s ========================
```

`uv run ruff check` and `uv run ty check` are clean on the touched files.

## Backward compatibility

Defaults match existing hardcoded values, so a deployment that doesn't
set either env var sees no behavioral change.

## What this PR doesn't change

- No new exception classes. The `Exception` rewrap in
  `engine/retain/embedding_utils.py:72` (which loses the original
  `APIConnectionError` class signal) is unchanged in this PR. A
  follow-up could preserve the class to enable per-class retry policy,
  but that's a larger conversation; this PR keeps the surface minimal.
- No exponential backoff. The backoff stays linear; a future PR could
  add a multiplier env var. Linear with a tunable cap is enough to
  survive multi-hour outages with appropriate values.
- No changes to the LLM call site's own retry loop in
  `engine/providers/openai_compatible_llm.py:489-500`, which is
  separate from the worker-level retry covered here.

## Versions verified

`0.6.2`, `0.7.0`, and `main` HEAD (`78c3525`) all share the hardcoded
retry decision, so this PR applies to all three.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -431,6 +431,7 @@ ENV_WORKER_ENABLED = "HINDSIGHT_API_WORKER_ENABLED"
 ENV_WORKER_ID = "HINDSIGHT_API_WORKER_ID"
 ENV_WORKER_POLL_INTERVAL_MS = "HINDSIGHT_API_WORKER_POLL_INTERVAL_MS"
 ENV_WORKER_MAX_RETRIES = "HINDSIGHT_API_WORKER_MAX_RETRIES"
+ENV_WORKER_TASK_RETRY_BACKOFF_SECONDS = "HINDSIGHT_API_WORKER_TASK_RETRY_BACKOFF_SECONDS"
 ENV_WORKER_HTTP_PORT = "HINDSIGHT_API_WORKER_HTTP_PORT"
 ENV_WORKER_MAX_SLOTS = "HINDSIGHT_API_WORKER_MAX_SLOTS"
 
@@ -702,6 +703,7 @@ DEFAULT_WORKER_ENABLED = True  # API runs worker by default (standalone mode)
 DEFAULT_WORKER_ID = None  # Will use hostname if not specified
 DEFAULT_WORKER_POLL_INTERVAL_MS = 500  # Poll database every 500ms
 DEFAULT_WORKER_MAX_RETRIES = 3  # Max retries before marking task failed
+DEFAULT_WORKER_TASK_RETRY_BACKOFF_SECONDS = 60  # Seconds between retries on transient task failure
 DEFAULT_WORKER_HTTP_PORT = 8889  # HTTP port for worker metrics/health
 DEFAULT_WORKER_MAX_SLOTS = 10  # Total concurrent tasks per worker
 DEFAULT_RETAIN_MAX_CONCURRENT = 4  # Max concurrent retain DB phases (HNSW reads + writes). Limits I/O contention.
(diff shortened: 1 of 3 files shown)


nicoloboschi (changes requested):
thanks!

we need to read it via config.py not via env inside memory_engine.py

aaronwestphal:
Done — `MemoryEngine.execute_task` now reads `worker_max_retries` and `worker_task_retry_backoff_seconds` from `get_config()` instead of `os.getenv`. `config.py` keeps the env parsing and defaults; the retry decision just consumes the typed fields like the rest of the worker.

One honest note: since `get_config()` is cached for the process lifetime, the values are resolved once at startup, so I dropped the earlier "tunable without restart" wording from the comment — changing the policy now requires a worker restart, consistent with the other worker config. Tests clear the config cache after patching the env so `get_config()` re-reads, with an autouse fixture clearing it around each test to avoid leakage.

4/4 regression tests still green; ruff + ty clean. Ready for another look.

nicoloboschi:
Merged manually after rebasing onto main to resolve conflicts. The review feedback (use get_config() instead of os.getenv) was incorporated into the resolution. Thanks @aaronwestphal!

PR #1799, opened 2026-05-27
feat(worker): add opt-in periodic scanner for pending consolidations

aaronwestphal:
## Summary

Adds an opt-in background task in the worker that periodically scans
`memory_units WHERE consolidated_at IS NULL AND consolidation_failed_at IS NULL`
for distinct `bank_id` values and calls `submit_async_consolidation` for each.
The underlying submit already passes `dedupe_by_bank=True`, so the scanner is
safe to run alongside the existing event-driven triggers: if a consolidation
is already pending for a bank, the scanner's submit collapses to a no-op.

The scanner is **disabled by default** — set
`HINDSIGHT_API_WORKER_PENDING_CONSOLIDATION_SCAN_INTERVAL_SECONDS` to a
positive number of seconds to enable. Default `0` preserves existing behavior
exactly, so this PR has zero runtime impact unless an operator opts in.

## Why

Consolidation today is purely event-driven. Every existing trigger of
`submit_async_consolidation` runs only after some *other* operation succeeds:

| Trigger | File:line (main HEAD `78c3525`) |
|---|---|
| After successful retain (gated on `enable_observations`) | `memory_engine.py:2711` |
| After document deletion that invalidates observations | `memory_engine.py:4236` |
| After retain/delete-adjacent paths | `memory_engine.py:4402, 4494, 4641` |
| After single-memory deletion orphaning observations | `memory_engine.py:4808` |
| Self-resubmit (only when round limit was hit) | `consolidation/consolidator.py` |
| Manual HTTP API trigger | `api/http.py` (`POST /v1/default/banks/{bank_id}/consolidate`) |

Notably absent: any code that observes `memory_units.consolidated_at IS NULL`
on its own. If the event chain breaks — embeddings-provider outage that
fails retains for hours, a manual import that bypasses retain, fact-extraction
LLM 5xx exceeding `retain_llm_max_retries`, hook misconfiguration that
swallows the trigger call — `memory_units` accumulate unprocessed and no
observer notices until someone runs SQL or hits the manual endpoint.

This PR doesn't replace any of those event-driven triggers. It adds a
backstop that catches the residue.

## How it works

`hindsight-api-slim/hindsight_api/worker/pending_consolidation_scanner.py`
defines a small class:

```python
class PendingConsolidationScanner:
    async def run(self) -> None:
        """Loop until shutdown(); short-circuits if interval <= 0."""

    async def _scan_once(self) -> int:
        """Scan every schema once; returns count of NEW ops queued."""

    async def _find_pending_banks(self, schema) -> list[str]:
        """SELECT DISTINCT bank_id FROM memory_units
           WHERE fact_type IN ('experience','world')
             AND consolidated_at IS NULL
             AND consolidation_failed_at IS NULL
           LIMIT 100"""

    async def _submit_if_not_dedup(self, bank_id, schema) -> int:
        """submit_async_consolidation; returns 1 for new ops, 0 for dedupe hits."""
```

`worker/main.py` constructs the scanner once, alongside the existing
`Poller`, and runs both as `asyncio.create_task(...)`. Graceful shutdown
calls `scanner.shutdown()` before the poller shuts down, so an in-flight
scan can complete without orphaned ops.

## Design notes

- **Why opt-in?** The retain-side trigger chain already covers the common
  path. Keeping the scanner off by default ships a safety net without
  changing default semantics for existing deployments. Operators who hit
  the failure mode can enable it; operators who haven't lose nothing.

- **Why a separate task in `worker/main.py` rather than embedded in the
  poller?** The scanner needs the `MemoryEngine` instance to call
  `submit_async_consolidation` (the poller only has the bound `execute_task`
  method). Putting it alongside the poller in `worker/main.py` keeps the
  engine reference in one place and avoids passing the engine through the
  poller's constructor.

- **Multi-tenant safety.** The scanner iterates the same
  `tenant_extension.list_tenants()` as the poller. The single-instance
  `DefaultTenantExtension` is now constructed once in `worker/main.py`
  and shared between the poller and scanner, replacing the poller's
  internal-only construction.

- **Schema contextvar discipline.** `_current_schema` is `set()` / `reset()`
  around each per-bank submit using the token pattern, so a prior tenant's
  schema can't leak into the next iteration.

- **Per-scan cap.** `LIMIT 100` distinct `bank_id`s per schema per scan.
  The scanner is a backstop, not a complete sweep — banks not picked up
  on one cycle are picked up on the next. This caps DB pressure on
  deployments with very many distinct banks.

- **Error isolation.** Per-bank submit failures are logged and counted
  as `0` queued, not raised. One bad bank can't kill the scan. Scan-level
  failures (DB hiccup) are also absorbed and retried on the next interval.

## Tests

`hindsight-api-slim/tests/test_pending_consolidation_scanner.py` adds
4 regression tests:

- `test_scanner_queues_when_no_pending_op` — bank with unconsolidated
  rows triggers exactly one `submit_async_consolidation` call
- `test_scanner_reports_dedupe_when_submit_dedupes` — when submit returns
  `deduplicated=True`, the scanner does not count it as a queued bank
- `test_scanner_skips_fully_consolidated_bank` — bank with only
  consolidated rows is filtered out by the SQL query; no submit is attempted
- `test_scanner_disabled_short_circuits` — when interval=0, `run()`
  returns inside 2 seconds without touching the database

```
$ uv run pytest tests/test_pending_consolidation_scanner.py -v
======================== 4 passed, 2 warnings in 39.58s ========================
```

Tests mock `MemoryEngine.submit_async_consolidation` so the scanner's
SQL-driven discovery and dispatch logic is exercised without entangling
with the consolidator pipeline (which runs inline under `SyncTaskBackend`
and would make assertions about `async_operations.status` non-deterministic).

`uv run ruff check` and `uv run ty check` are clean on the touched files.

## Backward compatibility

Without setting `HINDSIGHT_API_WORKER_PENDING_CONSOLIDATION_SCAN_INTERVAL_SECONDS`,
the scanner's `run()` short-circuits after a single log line and never
touches the database. `test_scanner_disabled_short_circuits` guards this
with a 2-second timeout.

## What this PR doesn't change

- No changes to existing event-driven triggers. The scanner is purely additive.
- No changes to `dedupe_by_bank` semantics. Existing `submit_async_consolidation`
  deduplication handles the safety.
- No SQL migrations. The query reads from existing `memory_units` columns.

## Open questions for reviewers

1. **Default off vs default on at low frequency.** Currently default 0
   (disabled). The argument for default-on (e.g. `300`) is that the
   scanner is the only mechanism that recovers from event-chain breaks;
   the argument for default-off is that the project shouldn't change
   default semantics in a feature PR. Happy to flip if you'd prefer.

2. **Per-tenant interval.** Not implemented. All tenants share the same
   interval. Could add a per-tenant override later if needed.

3. **Metrics.** Not added. The scanner logs counts of queued banks but
   doesn't emit Prometheus metrics. Easy to add a counter in a follow-up.

4. **Cap on banks per scan (`_MAX_BANKS_PER_SCHEMA_PER_SCAN = 100`).**
   Chosen as a reasonable default. Open to making this configurable.

## Versions verified

`0.6.2`, `0.7.0`, and `main` HEAD (`78c3525`) all share the same gap —
no file under `worker/` periodically scans `memory_units` for
unconsolidated rows. Verified by grep.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -431,6 +431,9 @@ ENV_WORKER_ENABLED = "HINDSIGHT_API_WORKER_ENABLED"
 ENV_WORKER_ID = "HINDSIGHT_API_WORKER_ID"
 ENV_WORKER_POLL_INTERVAL_MS = "HINDSIGHT_API_WORKER_POLL_INTERVAL_MS"
 ENV_WORKER_MAX_RETRIES = "HINDSIGHT_API_WORKER_MAX_RETRIES"
+ENV_WORKER_PENDING_CONSOLIDATION_SCAN_INTERVAL_SECONDS = (
+    "HINDSIGHT_API_WORKER_PENDING_CONSOLIDATION_SCAN_INTERVAL_SECONDS"
+)
 ENV_WORKER_HTTP_PORT = "HINDSIGHT_API_WORKER_HTTP_PORT"
 ENV_WORKER_MAX_SLOTS = "HINDSIGHT_API_WORKER_MAX_SLOTS"
 
@@ -702,6 +705,10 @@ DEFAULT_WORKER_ENABLED = True  # API runs worker by default (standalone mode)
 DEFAULT_WORKER_ID = None  # Will use hostname if not specified
 DEFAULT_WORKER_POLL_INTERVAL_MS = 500  # Poll database every 500ms
 DEFAULT_WORKER_MAX_RETRIES = 3  # Max retries before marking task failed
+# Periodic scan that re-queues consolidation for banks with unconsolidated
+# memory_units (a backstop for cases where the retain-side trigger chain breaks).
+# 0 = disabled (preserves existing behavior); typical opt-in value 300 (5 min).
+DEFAULT_WORKER_PENDING_CONSOLIDATION_SCAN_INTERVAL_SECONDS = 0
 DEFAULT_WORKER_HTTP_PORT = 8889  # HTTP port for worker metrics/health
 DEFAULT_WORKER_MAX_SLOTS = 10  # Total concurrent tasks per worker
 DEFAULT_RETAIN_MAX_CONCURRENT = 4  # Max concurrent retain DB phases (HNSW reads + writes). Limits I/O contention.
@@ -1234,6 +1241,7 @@ class HindsightConfig:
     worker_id: str | None
     worker_poll_interval_ms: int
     worker_max_retries: int
+    worker_pending_consolidation_scan_interval_seconds: int
     worker_http_port: int
     worker_max_slots: int
     worker_slot_reservations: dict[str, int]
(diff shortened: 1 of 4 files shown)


nicoloboschi (changes requested):
Thanks for raising this issue 

I don't think this is the correct solution as it will add a lot of pressure on the database, likely for nothing 99% of the time. 

I think the consolidation recovery should be better and self-reschedule instead of just fail - I'll look into that

aaronwestphal:
That's fair — agreed. A periodic full-table scan is the blunt instrument, and you're right that it costs DB pressure on every interval for work that's empty almost all the time. Event-driven self-reschedule on the failure path is the cleaner mechanism: it only does work when there's actually a stalled op to recover, and it keeps the system purely event-driven instead of bolting on a poller.

Closing this in favor of that approach. The underlying failure mode I hit was: when the retain → consolidation trigger chain breaks (embeddings-provider outage failing retains for an extended window, or a manual import that bypasses retain), `memory_units` with `consolidated_at IS NULL` accumulate and nothing re-queues them. If a self-rescheduling recovery covers that, it fully addresses what I was working around.

Happy to test a patch against my deployment when you have something — I can reproduce the stalled-consolidation state reliably. Thanks for the quick look.

PR #1821, opened 2026-05-28
fix(api): honor explicit daemon host and port

### LINKED ISSUE #1786: Named profile startup fails when HINDSIGHT_API_PORT is in os.environ
### Bug Description

Starting a named profile with `HINDSIGHT_API_PORT` in the environment causes the daemon to bind to port 8888 instead of the profile's configured port. `ensure_running` then hangs polling the expected port until timeout.

### Root Cause
_start_daemon_locked copies os.environ (including HINDSIGHT_API_PORT) into the child env, then also passes --port on the CLI from profile
  metadata. hindsight-api sees both sources present, treats it as a conflict regardless of whether the values match, and silently falls back to
  DEFAULT_DAEMON_PORT (8888).

### Steps to Reproduce

```bash
# 1. Create a named profile pinned to port 9555
python -c "
from hindsight_embed.profile_manager import ProfileManager
pm = ProfileManager()
pm.create_profile('test-port-bug', 9555, {})
"

# 2. Start with HINDSIGHT_API_PORT in env (matching the profile port)
HINDSIGHT_API_PORT=9555 python -c "
from hindsight_embed import DaemonEmbedManager
mgr = DaemonEmbedManager()
mgr.ensure_running({}, 'test-port-bug')
"

# 3. Check what port it actually bound to
curl http://127.0.0.1:9555/health   # connection refused
curl http://127.0.0.1:8888/health   # 200 — fell back to default

# 4. Cleanup
python -c "
from hindsight_embed import DaemonEmbedManager
DaemonEmbedManager().stop('test-port-bug')
"

### Expected Behavior

Daemon binds to the profile's configured port (9555). ensure_running returns successfully.


### Actual Behavior

Daemon binds to 8888. ensure_running polls 9555 indefinitely and times out.

### Version

0.6.2

### LLM Provider

None

### PR DESCRIPTION
**Pull request by Sanderhoff-alt**
## Summary

Fix daemon-mode host/port resolution so explicit CLI values are honored even when they match environment-derived config defaults.

Before this change, daemon mode inferred whether `--host` or `--port` had been explicitly supplied by comparing parsed argparse values with the loaded config. That breaks when the CLI value equals an environment-derived default. For example:

```bash
HINDSIGHT_API_PORT=9555 hindsight-api --daemon --port 9555
```

`argparse` parsed `args.port=9555`, config also had `port=9555`, and daemon mode treated the port as implicit, falling back to `DEFAULT_DAEMON_PORT` (`8888`). This caused `hindsight-embed` named profile startup to wait on the profile port while the API daemon listened on the default daemon port instead.

## Changes

- Track explicit `--host` / `--port` through argparse itself using `argparse.SUPPRESS` defaults.
- Preserve explicit daemon host/port values even when they match config defaults.
- Keep argparse-accepted long-option abbreviations, such as `--po` and `--ho`, on the same explicit-argument path.
- Return a named dataclass from daemon host/port resolution instead of a multi-value tuple.
- Extend daemon host/port tests for:
  - explicit `--port` matching `HINDSIGHT_API_PORT`
  - env-only port still using daemon default behavior
  - explicit `--host` matching config default
  - argparse abbreviations `--po` and `--ho`

## Tests

```bash
uv run ruff check hindsight_api/main.py tests/test_daemon_host_resolution.py
uv run ruff format --check hindsight_api/main.py tests/test_daemon_host_resolution.py
uv run ty check hindsight_api/main.py
uv run pytest tests/test_daemon_host_resolution.py tests/test_daemonize.py
```

`tests/test_daemon_host_resolution.py tests/test_daemonize.py`: 15 passed.

Note: the full local pre-commit hook currently also runs `generate-docs-skill.sh`, which reports existing generated-docs drift unrelated to this daemon fix. This PR intentionally keeps the daemon fix scoped to `hindsight-api-slim/hindsight_api/main.py` and `hindsight-api-slim/tests/test_daemon_host_resolution.py`.

Fixes #1786.

### DIFF
--- a/hindsight-api-slim/hindsight_api/main.py
+++ b/hindsight-api-slim/hindsight_api/main.py
@@ -73,29 +73,42 @@ def _signal_handler(signum, frame):
     sys.exit(0)
 
 
-def resolve_daemon_host_port(*, args_host: str, args_port: int, config_host: str, config_port: int) -> tuple[str, int]:
+@dataclasses.dataclass(frozen=True)
+class ResolvedDaemonHostPort:
+    host: str
+    port: int
+
+
+def resolve_daemon_host_port(
+    *,
+    args_host: str,
+    args_port: int,
+    explicit_host: bool,
+    explicit_port: bool,
+) -> ResolvedDaemonHostPort:
     """Resolve host/port for daemon mode.
 
     Defaults to 127.0.0.1 for security, but honors explicit user overrides
     via --host flag or HINDSIGHT_API_HOST env var. Uses DEFAULT_DAEMON_PORT
     unless the user specified a custom port.
     """
-    port = args_port if args_port != config_port else DEFAULT_DAEMON_PORT
+    port = args_port if explicit_port else DEFAULT_DAEMON_PORT
     # Only force localhost if the user didn't explicitly set a host
-    if args_host != config_host or os.environ.get(ENV_HOST):
+    if explicit_host or os.environ.get(ENV_HOST):
         host = args_host
     else:
         host = "127.0.0.1"
-    return host, port
+    return ResolvedDaemonHostPort(host=host, port=port)
 
 
-def main():
-    """Main entry point for the CLI."""
-    global _memory
+@dataclasses.dataclass(frozen=True)
+class ParsedCliArgs:
+    args: argparse.Namespace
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM

PR #1841, opened 2026-05-28
Bound trigram entity resolution batch size during retain

voarsh2:
Large retain operations can produce many unique entity names in one streaming mini-batch. The trigram resolver currently passes all unique entity texts to a single unnest(...) JOIN entities query. On large banks this can create a single expensive pg_trgm query that hits asyncpg/Postgres timeouts and fails the whole retain operation.

This change chunks trigram entity candidate lookup into bounded batches while preserving the same candidate grouping and scoring behavior.


nicoloboschi (approved):
LGTM — bounds the trigram candidate query so a wide retain batch can't time out a single `unnest(...)` JOIN, with batching tests in place. The new `SET … / try / RESET` structure is also a small improvement over the previous code, which would have leaked the lowered `pg_trgm.similarity_threshold` for the rest of the session on a fetch failure.

Merging now; opening a small follow-up PR with:
- docs entry for `HINDSIGHT_API_RETAIN_ENTITY_RESOLUTION_BATCH_SIZE` in `configuration.md`
- collapse `try/except…else + raise` to a `try/finally` so RESET is unconditionally guaranteed
- a test asserting RESET still runs when `conn.fetch` raises mid-batch
- the same batching applied to `_resolve_entities_batch_oracle_fuzzy`, which has the identical single-query risk

Thanks for the fix.

PR #1843, opened 2026-05-29
feat(consolidation): Add HA-safe intra-op LLM parallelism

Fyko:
## What

Adds an opt-in `HINDSIGHT_API_CONSOLIDATION_LLM_PARALLELISM` knob that lets one consolidation worker process multiple tag groups concurrently. Default stays `1` (sequential) — no behavior change unless you set it.

## Why

Consolidation is LLM-bound and chews through tag groups one at a time, which drags on banks with lots of memories. This fans a single worker out across distinct tag groups while staying HA-safe — batches within a group share an observation scope so they stay serial (no double-create / lost-update on the same rows), distinct groups touch disjoint scopes and run concurrently, per-bank single-flight is still enforced at the SQL claim level, and per-batch stat deltas merge after dispatch to dodge counter races.


nicoloboschi:
thanks @Fyko for it - I made an improved and more correct version here #1853

PR #1851, opened 2026-05-29
fix(retain): retry transient extraction failures

### LINKED ISSUE #1833: Retain path silently drops memory on a transient extraction-LLM failure: `extract_facts_from_contents` swallows retryable errors and marks the op `completed` with 0 facts (severs the built-in `RetryTaskAt` path)
## Summary

When the fact-extraction LLM hits a **transient** failure (rate-limit / 429 / timeout / connection / 5xx) during a retain, `extract_facts_from_contents` catches the exception, logs `WARNING "Content extraction failed (skipping)"`, substitutes an empty result, and lets the async retain operation be marked **`completed` with 0 facts saved**. The memory for that document is permanently lost — no error state, no retry, no alert. This directly contradicts the engine's own design: the inner `extract_facts_from_text` deliberately *raises* on chunk failure so "the worker poller will retry the entire task", and the task executor is built to convert a retryable exception into `RetryTaskAt`. The swallow in `extract_facts_from_contents` is the single link that breaks that chain.

## Root cause

In `hindsight_api/engine/retain/fact_extraction.py`:

1. **The inner function raises on purpose.** `extract_facts_from_text()` collects per-chunk failures and, on any failure, raises `RuntimeError("Fact extraction failed: N/M chunks failed. First failures: …")` (~`:1616-1619`). The failure summary embeds `type(err).__name__` (e.g. `chunk 0: RateLimitError`), and the surrounding comment states the intent: the worker poller is expected to retry the whole task.

2. **The outer function swallows it.** `extract_facts_from_contents()` runs the per-content extractions with `asyncio.gather(*tasks, return_exceptions=True)` (~`:2215`) and then (~`:2225-2232`):

   ```python
   for content, result in zip(contents, all_fact_results):
       if isinstance(result, Exception):
           logger.warning(f"Content extraction failed (skipping): {type(result).__name__}: {result}")
           valid_results.append((content, ([], [], TokenUsage())))  # <-- swallow: empty result
       else:
           valid_results.append((content, result))
   ```

   A `RateLimitError` (wrapped in the `RuntimeError` above) is indistinguishable here from a deterministic failure (malformed JSON, token-limit) — **all** are converted to an empty `([], [], TokenUsage())`.

3. **This severs the documented retry path.** On the live streaming retain path (`retain_batch_enabled=False`), the call chain is `execute_task` → `_handle_batch_retain` → `retain_batch_async` → `orchestrator.retain_batch` → `_streaming_retain_batch` → `_llm_producer` → `_extract_and_embed` → `extract_facts_from_contents`. The streaming producer is otherwise correct: it collects child-task exceptions and re-raises the first one, and the executor classifies it (its non-retryable predicate only flags integrity / embedding-dimension errors, so a rate-limit error *is* retryable) and issues `RetryTaskAt(+60s)` up to the retry budget, then a loud `failed`. Because `extract_facts_from_contents` already discarded the exception, none of that fires — the document is committed with 0 facts and the op is marked `completed`.

## Impact

Transient extraction-LLM rate limits are common (especially on subscription/coding-plan endpoints). Each one can silently drop a retain's facts while the operation reports success. There is no signal — no error row, no exception, no retry — that memory went missing. (Detecting the symptom after the fact is the subject of #1579.)

## Proposed fix

Un-swallow **retryable** failures in `extract_facts_from_contents` so they propagate into the existing `RetryTaskAt` machinery, while leaving **deterministic** failures on today's skip-and-continue behavior (re-running malformed-JSON / token-limit chunks just burns quota):

1. Classify each gathered exception. If it is transient (rate-limit / 429 / timeout / connection / 5xx — recognizable from the embedded `type(err).__name__` and/or message text), **re-raise** it (e.g. the first such exception) instead of substituting an empty result. The streaming producer + executor then retry the task as designed.
2. Keep the skip-and-empty path for deterministic / quality failures (malformed JSON, unsupported token limit, BadRequest), so a permanently-bad chunk doesn't loop forever.

This is the minimal change that makes a rate-limited retain self-heal (transient) or fail loudly (persistent), instead of silently completing with 0 facts. We are running this as a downstream monkeypatch and would be glad to open a PR if useful.

## Cross-references

- **#1579 (open)** — *"Inconsistent result_metadata: completed retains missing unit_ids_count make silent-extraction-failure undetectable."* Complementary: #1579 is about *detecting* a silent extraction failure after the fact; this issue is about *preventing* the drop in the first place.
- **#1798 (open)** — *"fix(worker): wire `HINDSIGHT_API_WORKER_MAX_RETRIES` into task retry decision."* Complementary: once the failure propagates (this issue), #1798 governs how many times the worker retries it.
- **#1147 (closed)** — *"LLM quota/429 errors get swallowed by retry wrapper — silent queue bloat + startup crash loop."* Sibling: the same swallow-a-429 anti-pattern, in a different location (the retry wrapper). This issue is the fact-extraction `gather` instance.

## Environment

- hindsight `v0.7.1` (latest, released 2026-05-28; bug reproduces unchanged from `v0.7.0` — `hindsight_api/engine/retain/fact_extraction.py` is byte-identical between the two tags).
- Local-embedded mode; `openai_compatible` extraction provider; live **streaming** retain path (`retain_batch_enabled=False`). The OpenAI Batch API extraction path (`extract_facts_from_contents_batch_api`) is separate and out of scope here.

xmh1011:
Summary
- preserve the chunk-level exception cause when fact extraction fails
- classify retryable content extraction failures (rate limit, timeout, connection, 5xx) and re-raise them so worker retry handling can run
- keep the existing skip-and-empty behavior for deterministic/non-retryable extraction failures

Closes #1833.

Tests
- uv run ruff check hindsight_api/engine/retain/fact_extraction.py tests/test_fact_extraction_retry.py
- uv run pytest tests/test_fact_extraction_retry.py tests/test_experience_fact_type_passthrough.py -q
- ./scripts/hooks/lint.sh

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -9,6 +9,7 @@ import asyncio
 import json
 import logging
 import re
+from collections.abc import Iterator
 from datetime import datetime, timedelta
 from typing import Literal, cast
 
@@ -1616,7 +1617,7 @@ async def extract_facts_from_text(
         raise RuntimeError(
             f"Fact extraction failed: {len(failed_chunks)}/{len(chunks)} chunks failed. "
             f"First failures: {failed_summary}"
-        )
+        ) from failed_chunks[0][1]
 
     return all_facts, chunk_metadata, total_usage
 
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
the transient errors are already handled at llm level
see HINDSIGHT_API_LLM_MAX_RETRIES
https://hindsight.vectorize.io/developer/configuration

if you think we're missing some errors in there that should be retried, please adjust this pr or create a new one, thanks

PR #1853, opened 2026-05-29
fix(consolidation): scope-locked parallel dispatch (alternative to #1843)

nicoloboschi:
## Summary

Alternative implementation of intra-op LLM parallelism for consolidation (compare to #1843). Adds an opt-in `HINDSIGHT_API_CONSOLIDATION_LLM_PARALLELISM` knob — default stays `1` (sequential, no behavior change).

The difference from #1843 is **how disjointness is determined for safe parallelism**. #1843 keys parallelism on tag-group (memory's exact tag set) and documents a caveat: "keep at 1 for banks using `per_tag` / `all_combinations` `observation_scopes`." That caveat is unenforceable per-tenant (a global env var overrides per-bank choice) and the disjointness claim does not hold under those modes — different tag groups explicitly fan out across overlapping write scopes, so concurrent groups race on the same observation rows (lost-update on `text`, `source_memory_ids`, `history`, `proof_count`).

This PR fixes that by computing the **write-scope set** per tag group up front and dispatching under per-scope `asyncio.Lock`s acquired in lexicographic order. Concurrency emerges from disjointness automatically — no operator opt-in, no documentation footgun:

- **`combined`** (default): each group locks one scope (`frozenset(memory.tags)`); disjoint groups run in parallel.
- **`per_tag`** / **`all_combinations`** / **explicit list**: groups lock every scope they will write to. Two groups that share any scope serialize *on that scope*; the rest still parallelize.
- Deadlock-free by construction: locks acquired in `tuple(sorted(scope))` order across all groups.
- `consolidation_llm_parallelism` becomes a *max* concurrency cap; actual concurrency degrades to safety when scopes contend.

Also extracts the per-memory observation_scopes resolver into module-level helpers (`_resolve_obs_tags_list`, `_resolve_write_scopes`, `_parse_observation_scopes`, `_scope_sort_key`) so the dispatcher and the lock layer share one source of truth — over-reporting is safe, under-reporting is the race, and the agreement test (`TestDispatchLockAgreement`) pins it.

## Code changes

- **`engine/consolidation/consolidator.py`** — new helpers + `run_consolidation_job` refactor with deferred per-batch deltas (no lost-update on shared `stats` / `consolidated_tags`) and scope-locked parallel dispatch.
- **`config.py`** — `ENV_CONSOLIDATION_LLM_PARALLELISM`, default `1`, dataclass field, `_CONFIGURABLE_FIELDS` (per-bank overridable), `from_env`.
- **`hindsight-docs/docs/developer/configuration.md`** — documents the flag *without* the per_tag/all_combinations caveat (no longer needed).

### Unit (`tests/test_consolidation_write_scopes.py`, 28 tests)

- `_resolve_write_scopes` for every mode: combined / per_tag / all_combinations / explicit list / JSON-encoded string from asyncpg / pre-parsed Python value / empty-tag edge cases.
- `_resolve_obs_tags_list` parity with the dispatcher's existing semantics.
- `TestDispatchLockAgreement` — for every recognised input shape, every scope the dispatcher visits via `_resolve_obs_tags_list` must have a lock from `_resolve_write_scopes`. This is the load-bearing invariant; under-locking is exactly the race we're closing.
- `_scope_sort_key` total order + stability under different frozenset constructions (deadlock-freedom).

### Integration (`tests/test_consolidation_scope_parallelism.py`, 6 tests, mock LLM)

- Combined mode under `parallelism=3`: three disjoint tag sets → exactly the three expected observations, tagged correctly.
- per_tag under `parallelism=4`: memory `[alice, session]` writes obs at `[alice]` and `[session]`, overlapping `[alice]` scope with another memory's per_tag pass.
- all_combinations under `parallelism=4`: memory `[alice, session]` writes obs at `[alice]`, `[session]`, `[alice, session]`.
- Explicit list under `parallelism=4`: declared scopes are written, memory's own tags are *not*.
- **Lock invariant under contention**: two memories sharing scope `[a]` via per_tag — `_find_related_observations` is wrapped to record per-scope in-flight count; assert `max ≤ 1` per scope.
- **Disjoint scopes actually parallelize**: three disjoint memories observed concurrently in-flight on distinct scopes (≥2 simultaneous).

## Test plan

- [x] `tests/test_consolidation_write_scopes.py` — 28 passed
- [x] `tests/test_consolidation_scope_parallelism.py` — 6 passed
- [x] `./scripts/hooks/lint.sh` — clean
- [x] `tests/test_consolidation.py` + `tests/test_observations.py` — 106 passed, 1 unrelated flake (`test_consolidation_keeps_different_people_separate`, `@hs_llm_core` + `assert_meets_criteria` against real Gemini; failed both runs — first via the judge, second via a 240s timeout on a 226s LLM call. Default parallelism `1` means this path is byte-for-byte identical to main.)
- [ ] Manual sanity with `HINDSIGHT_API_CONSOLIDATION_LLM_PARALLELISM=4` on a per_tag / all_combinations bank (reviewer)

## Related

Alternative to #1843. Same env-var name + default for compatibility; replaces tag-group-keyed parallelism with scope-keyed parallelism.


Fyko:
thanks for picking this up!

Issue #1797, opened 2026-05-27
Async retain split overwrites oversized single-document ingests


Korayem:
## Problem

Async retain of an oversized single document can complete successfully while only retaining one slice of the original document.

Expected behavior: retaining a large conversation as one stable document ID should preserve the full document and extract facts across all internal chunks. This matches the documented guidance to pass the entire conversation as one document and let Hindsight chunk it automatically.

Actual behavior observed on 0.7.0: an oversized single document was split into many async child retain operations. All child operations completed, but the final document row contained only one chunk-sized substring of the source text. The surviving memory units also corresponded to the last child operation that updated the document, not the whole source transcript.

Reproduction shape:

1. Submit one large document through async retain with a stable document ID. The observed test document was about 1.95M parsed characters / 455k tokens.
2. Let the async parent operation split into child retain operations.
3. Wait until the parent and children complete.
4. Fetch the document by ID.
5. Compare the fetched document original text to the original input.

Observed evidence from the test run:

- Parent async retain split one item into 79 child retain operations.
- All 79 child retain operations completed.
- The final stored document original text was 29,944 chars.
- The original parsed input was 1,955,110 chars.
- The stored original text was an exact substring of the original input, starting around 43.42% into the source and ending around 44.96%.
- The final document had 83 memory units.
- The child operation that last updated the parent also reported 83 unit IDs, which strongly suggests last-writer-wins behavior.

## Root Cause Analysis

The async submission path splits an oversized single retain item before it reaches the normal retain pipeline. Each split child operation carries the same document ID but only a slice of the original content.

Each child operation is then processed as an independent full document retain. Because all children share the same document ID, each child can replace the document row and delete/replace units for that document. The final persisted document becomes whichever child finishes last, not the full source document.

This differs from the intended internal streaming/chunking contract. The internal retain pipeline can process many chunks under one document while tracking the full document content and avoiding repeated document replacement. The async pre-split path loses that parent document context by turning chunks into independent full-replace operations.

This is not caused by concise extraction mode. Extraction strategy may affect how many facts are extracted, but it should not shrink the stored original document text from the full submitted content to a single chunk.

## TDD Fix Plan

1. **RED**: Write an async retain test that submits one oversized document with a stable document ID and forces the async splitter to create multiple child units of work. Assert that, after completion, fetching the document returns the full submitted original text.
   **GREEN**: Change async oversized-document handling so a single document is not split into independent full-replace retain operations with the same document ID.

[...]

PR #1855, opened 2026-05-29
fix(retain): preserve full document body when splitter chunks oversized input

### LINKED ISSUE #1838: Large same-document retain replacement can persist a partial document body
## Problem

Replacing an existing large document by retaining new content with the same `document_id` can leave the stored document body partial/truncated.

Observed during retro conversation document repair against Hindsight 0.7.1 on Railway:

- Bank: `korayem-cli-agents-vlms`
- Document: `d327029b-d64e-4e16-a9e2-de462f0f057d`
- Source agent: Claude Code retro transcript
- Regenerated target body: 61,971 chars
- Stored body after same-ID retain replacement: ~29,973 chars
- The stored body starts with the expected `[role: user]` marker after retry, but it does not contain the full regenerated transcript.

Expected behavior: same-document replacement should be atomic from the caller's perspective. After the async retain completes, the document's `original_text` should exactly match the submitted content, or the operation should fail without replacing the existing body.

## Root Cause Analysis

The failure appears in the async retain path for replacing an existing document. Small and medium replacement documents completed correctly, but a larger Claude transcript repeatedly landed as a partial body. This suggests the replacement pipeline can publish intermediate or partial chunk content as the document body, or can overwrite the document record before all split/chunk child work has completed successfully.

Contributing factors:

- The only available body replacement path today is same-ID retain replacement.
- Large documents may be split/chunked internally.
- Retrying the same document reproduced the partial-body result.
- The stored partial body had valid transcript markers, so this is not just malformed input.

## TDD Fix Plan

1. **RED**: Write an integration test that retains an initial document, then replaces it with a larger same-ID document that requires internal splitting/chunking, and asserts the final `get document` response has `original_text` exactly equal to the submitted replacement content.
   **GREEN**: Make same-ID replacement publish the new document body only after all child processing succeeds.

2. **RED**: Write a test where one child split/chunk fails during same-ID replacement and assert the previous document body remains intact, or the operation reports failure without exposing partial content.
   **GREEN**: Add replacement transaction/parent-operation semantics so partial child completion cannot overwrite the visible document body.

3. **RED**: Write a retry test for the same large replacement and assert repeated replacements converge to the exact submitted body, not a suffix/prefix subset.
   **GREEN**: Make replacement idempotent for same `document_id` and content hash.

**REFACTOR**: Separate document-body replacement state from memory-unit extraction state so callers can reason about replace completion atomically.

## Acceptance Criteria

- [ ] Same-ID retain replacement of a large document stores the exact submitted body.
- [ ] Partial child processing cannot expose a partial `original_text`.
- [ ] Failed replacement leaves the prior document body intact or clearly marks the replacement failed.
- [ ] Retrying the same large replacement is idempotent.
- [ ] Tests cover documents large enough to trigger the internal split/chunk path.

nicoloboschi:
## Summary

Fixes #1838.

When a single retain content item exceeds `HINDSIGHT_API_RETAIN_BATCH_TOKENS` (~40 KB), `retain_batch_async` chunks it across multiple sub-batches. Each sub-batch passed only its own slice to `handle_document_tracking`, which unconditionally upserts `documents.original_text`. The last sub-batch overwrote the body with its slice, so the persisted body became a fragment of the input (e.g. a 62 KB transcript stored as ~30 KB on Railway 0.7.1).

## Fix

Thread a `document_body_override` parameter from `_split_contents_into_sub_batches` through `_retain_batch_async_internal`, `retain_batch`, `_streaming_retain_batch`, `_try_delta_retain` and `_delta_metadata_only`. When set, the orchestrator uses it as `combined_content` for the doc-row write so every sub-batch persists the same full body (and computes the same `content_hash`, so the FOR-UPDATE takeover check still passes).

RAM: the override is a reference to the splitter's source string — no extra copies. The orchestrator clears `combined_content` after `handle_document_tracking` finishes, same as before.

## Test plan

- [x] `tests/test_large_document_replacement.py::test_large_same_id_replacement_preserves_full_body` — small initial doc, large replacement under same `document_id`, asserts `original_text` matches exactly.
- [x] `tests/test_large_document_replacement.py::test_repeated_large_same_id_replacement_is_idempotent` — repeats the replacement 3× and asserts each retain converges to the exact submitted body (acceptance criterion 4 from the issue).
- [x] Both tests fail before the fix and pass after.
- [x] Test fixture lowers `HINDSIGHT_API_RETAIN_BATCH_TOKENS=100` and disables auto-consolidation so call time is sub-second.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -16,7 +16,7 @@ import logging
 import time
 import uuid
 from collections.abc import Awaitable, Callable
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import UTC, datetime, timedelta, timezone
 from typing import TYPE_CHECKING, Any, Literal, cast, overload
 
@@ -273,10 +273,18 @@ class _SubBatchSplit:
     user (such as ``retain_batch_async``) use this mapping to merge
     results belonging to the same original content back together when
     an oversized item was chunked across multiple sub-batches.
+
+    ``document_body_overrides[i]`` is the full original body of the
+    oversized item that produced ``sub_batches[i]``, or ``None`` when
+    the sub-batch was not produced by chunking an oversized item. The
+    orchestrator uses this as the ``documents.original_text`` payload
+    so that slicing an item across sub-batches does not persist a
+    partial body (see issue #1838).
     """
 
     sub_batches: list[list[RetainContentDict]]
     origin_indices: list[list[int]]
+    document_body_overrides: list[str | None] = field(default_factory=list)
 
 
 def _split_contents_into_sub_batches(
@@ -310,6 +318,7 @@ def _split_contents_into_sub_batches(
 
     sub_batches: list[list[RetainContentDict]] = []
     origin_indices: list[list[int]] = []
+    document_body_overrides: list[str | None] = []
     current_batch: list[RetainContentDict] = []
     current_batch_origins: list[int] = []
     current_batch_tokens = 0
@@ -319,6 +328,7 @@ def _split_contents_into_sub_batches(
         if current_batch:
             sub_batches.append(current_batch)
             origin_indices.append(current_batch_origins)
+            document_body_overrides.append(None)
             current_batch = []
             current_batch_origins = []
             current_batch_tokens = 0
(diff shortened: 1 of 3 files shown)


Korayem:
This should fix https://github.com/vectorize-io/hindsight/issues/1797

PR #1860, opened 2026-05-29
feat(api): add Fireworks AI batch inference provider

Fyko:
## What

Adds a `fireworks` LLM provider with native batch-retain support. `FireworksLLM` subclasses `OpenAICompatibleLLM`, so online inference reuses the existing OpenAI-compatible path, and overrides only the four batch members.

## Why

Batch retain (50% cost savings on fact extraction) is currently gated to `openai` and `groq`. Fireworks has the same async pricing, but its batch API is not OpenAI `/v1/batches`-compatible: it's a proprietary, account-scoped dataset -> job -> download REST workflow on a separate control-plane host. So it can't be a one-line allowlist add the way Groq was. This PR adds an adapter that maps Fireworks' input, output, and status shapes back onto the existing batch interface, so the retain orchestrator and `fact_extraction` consumer are unchanged.

## Notes

- Verified end-to-end against the live Fireworks batch API: a real retain run submits the dataset, creates the job, polls to completion, downloads, and extracts facts. The request/response shapes (dataset `exampleCount`/`format`, the output-JSONL nesting) were validated against the real API, not just the docs.
- The existing LiteLLM `fireworks_ai/...` online path is untouched.
- Batch requires `HINDSIGHT_API_FIREWORKS_ACCOUNT_ID` (the control-plane endpoints are `/v1/accounts/{id}/...`) and fails fast without it.
- A `createTime`-derived max-wait guards the documented Fireworks gotcha where a non-batch-eligible model leaves the job `PENDING` forever (the shared poll loop has no timeout of its own).
- Tests: unit tests over the normalizers + the dataset/job/download workflow (mocked httpx), plus a credentials-gated live integration test (`test_fireworks_batch_integration.py`) that runs the real batch end-to-end and skips without a key.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -172,6 +172,17 @@ ENV_RETAIN_LLM_MAX_BACKOFF = "HINDSIGHT_API_RETAIN_LLM_MAX_BACKOFF"
 ENV_RETAIN_LLM_TIMEOUT = "HINDSIGHT_API_RETAIN_LLM_TIMEOUT"
 ENV_RETAIN_LLM_LITELLMROUTER_CONFIG = "HINDSIGHT_API_RETAIN_LLM_LITELLMROUTER_CONFIG"
 
+# Fireworks AI batch inference. Fireworks' batch API is a proprietary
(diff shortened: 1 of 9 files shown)


nicoloboschi (approved):
LGTM

PR #1863, opened 2026-05-29
fix(openclaw): preserve retainDocumentScope config

feyola:
## Summary
- preserve `retainDocumentScope` in OpenClaw plugin config normalization
- add regression coverage for `retainDocumentScope: "turn"`
- document `retainDocumentScope` in the README config table

## Why
`retainDocumentScope` is documented and typed, and `buildRetainRequest()` honors it, but `getPluginConfig()` dropped the value. As a result, runtime config always fell back to `"session"`, so `"turn"` never took effect.

## Test plan
- `npm test`
- `npm run build`


nicoloboschi:
closing as we will drop this param

PR #1879, opened 2026-05-31
fix(api-slim): gracefully handle model init failures

w1ndcn:
## Summary

This PR hardens api-slim startup and reranking against local model initialization failures.

Previously, failures or hangs while loading local embedding/cross-encoder models could abort or block daemon startup. In Hermes integration, this can leave the client waiting on memory operations indefinitely.

This change adds per-task initialization timeouts, separates fatal infrastructure/config validation from optional model initialization, and lets cross-encoder reranking fall back to RRF-based scoring when initialization or prediction fails.

## Changes

- Add per-task timeout for startup model initialization via `HINDSIGHT_MODEL_INIT_TIMEOUT`.
- Treat pg0 startup and LLM batch API config validation as fatal initialization tasks.
- Treat embeddings, query analyzer, and cross-encoder initialization as degradable startup tasks.
- Track embedding and reranker initialization state.
- Skip embedding dimension migration when embeddings are unavailable.
- Add lazy cross-encoder initialization timeout/error handling.
- Add RRF-based reranking fallback when cross-encoder initialization or prediction fails.
- Add regression tests for reranker fallback behavior.

## Testing

```bash
uv run --extra test pytest tests/test_cross_encoder_reranker_fallback.py
```

Related to NousResearch/hermes-agent#35195.

This PR addresses one Hindsight-side failure mode where local embedding or cross-encoder model initialization can fail or hang, leaving Hermes waiting on post-response memory operations.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -13,6 +13,7 @@ import asyncio
 import contextvars
 import json
 import logging
+import os
 import time
 import uuid
 from collections.abc import Awaitable, Callable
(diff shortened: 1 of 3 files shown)


nicoloboschi (changes requested):
itìs better to fail because the system doesn't work instead of silently-fallback to something not accurate or that the user expects. 

if you have a specific reproducer of reranker/model not being able to startup, let's figure it out

PR #1881, opened 2026-05-31
fix(cli): hindsight memory retain --timestamp + correct fact-type values

slayoffer:
## Summary

Two unrelated CLI bugs surfaced during sandbox testing on 2026-05-31. Both have empirical reproductions and a regression test.

### Bug 1 — `hindsight memory retain --timestamp <ISO 8601>` never worked

`MemoryItem.timestamp` is generated from the OpenAPI schema
`anyOf: [{type: string, format: date-time}, {type: string}]`. Progenitor emits that as:

```rust
pub struct MemoryItemTimestamp {
    #[serde(flatten)]
    pub subtype_0: Option<chrono::DateTime<Utc>>,
    #[serde(flatten)]
    pub subtype_1: Option<String>,
}
```

— and serde **refuses to serialize** that for primitives:

```
Error("can only flatten structs and maps (got a string)")
```

So even hand-constructing the wrapper fails at serialize-time, before the request reaches the wire. The CLI's `serde_json::from_value::<MemoryItemTimestamp>(Value::String(s))` round-trip also fails (struct deserializer expects an object).

**Fix (at the codegen boundary).** Added a pre-codegen spec-massage pass `collapse_string_anyof_unions` in `hindsight-clients/rust/build.rs` that collapses any `anyOf` whose members are *all* `{type: "string"}` into a single `{type: "string"}`. The `format: "date-time"` distinction is lossless on the wire — both serialize to the same JSON string — so this is safe.

Result:
- `MemoryItem.timestamp` is now `Option<String>` in the generated client.
- The broken `MemoryItemTimestamp` struct is no longer emitted at all.
- The CLI no longer round-trips through a wrapper type; the user's string is passed straight through.

### Bug 2 — `hindsight memory clear --fact-type observation` was rejected; `opinion` was widely advertised but doesn't exist

The canonical fact types per the API are `world | experience | observation` (`hindsight_api.api.http.MemoryItem` + `Literal[...]` on `fact_types` in `RecallRequest` / `ReflectRequest`).

The CLI was out of sync:

- **`memory clear`**: `value_parser = ["world", "agent", "opinion"]` — rejected the valid value `observation`, accepted stale labels `agent`/`opinion` that the server silently treats as no-ops (no matching rows to delete).
- Help text on **`bank graph`**, **`memory list`**, **`memory recall`** referred to a non-existent fact type `opinion`.
- **`memory recall`** also had `default_values = &["world", "experience", "opinion"]` — pre-filling a non-existent type by default.

**Fix.** Replaced `opinion` → `observation` everywhere in help / clap defaults, and `agent`/`opinion` → `experience`/`observation` in the clear command's `value_parser` allow-list.

## Files changed

| File | Change |
|---|---|
| `hindsight-clients/rust/build.rs` | + `collapse_string_anyof_unions` pre-codegen pass |
| `hindsight-cli/src/commands/memory.rs` | Drop broken `MemoryItemTimestamp` round-trip — pass timestamp straight through |
| `hindsight-cli/src/api.rs` | Remove `MemoryItemTimestamp` from re-export |
| `hindsight-cli/src/main.rs` | 6 line fixes for stale `opinion`/`agent` labels |
| `hindsight-cli/tests/integration_test.rs` | + regression test for timestamp serialization |

## Test plan

- [x] `cargo build` → clean
- [x] `cargo test --bin hindsight` → 55/55 unit tests pass
- [x] `cargo test --test integration_test test_memory_item_timestamp_serializes_as_plain_string` → pass
- [x] `cargo clippy` → no new warnings (171 pre-existing `uninlined_format_args` style)
- [x] `cargo fmt --check` → no new diffs in files I touched
- [x] `hindsight memory clear --help` → `[possible values: world, experience, observation]`
- [x] `hindsight memory recall --help` → `[default: world experience observation]`
- [x] `hindsight bank graph --help` → `(world, experience, observation)`
- [ ] Live retain with `--timestamp 2026-05-31T10:00:00Z` against dev (manual; can run if maintainer wants)

### DIFF
--- a/hindsight-cli/src/api.rs
+++ b/hindsight-cli/src/api.rs
@@ -1266,8 +1266,8 @@ impl ApiClient {
 
 // Re-export types from the generated client for use in commands
 pub use types::{
-    BankProfileResponse, MemoryItem, MemoryItemTimestamp, RecallRequest, RecallResponse,
-    RecallResult, ReflectRequest, ReflectResponse, RetainRequest,
+    BankProfileResponse, MemoryItem, RecallRequest, RecallResponse, RecallResult, ReflectRequest,
+    ReflectResponse, RetainRequest,
 };
 
 #[cfg(test)]
--- a/hindsight-cli/src/commands/memory.rs
+++ b/hindsight-cli/src/commands/memory.rs
@@ -3,9 +3,7 @@ use std::fs;
 use std::path::PathBuf;
 use walkdir::WalkDir;
 
-use crate::api::{
-    ApiClient, MemoryItem, MemoryItemTimestamp, RecallRequest, ReflectRequest, RetainRequest,
-};
+use crate::api::{ApiClient, MemoryItem, RecallRequest, ReflectRequest, RetainRequest};
 use crate::config;
 use crate::output::{self, OutputFormat};
 use crate::ui;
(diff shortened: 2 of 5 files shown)


nicoloboschi (approved):
LGTM

PR #1889, opened 2026-06-01
fix(retain): keep routing labels out of fact actors

de1tydev:
## Summary

This PR prevents storage/routing labels from leaking into retain fact extraction as semantic actors.

It adds a configurable `retain_narrator` label for first-person assistant-authored content, defaults it to `AI assistant`, and sanitizes both narrator input and metadata before they are included in the LLM prompt.

## Problem

The retain prompt previously used the bank profile name as the `Narrator` hint and included raw metadata in the semantic extraction prompt. In deployments where bank/source/tag names describe infrastructure rather than people or agents, the LLM can mistake those routing labels for the actor in the memory.

Sanitized examples of the bad shape this produced:

- `team-memory-prod finished the import-path migration. | Involving: team-memory-prod`
- `Nimbus Agent found the batch job issue. | Involving: team-memory-prod`
- `AgentRuntime retained 58 new facts. | Involving: AgentRuntime`

In these examples, `team-memory-prod` is a bank id and `AgentRuntime` is a source system, not the actor that performed the action. Once extracted this way, the bad actor leaks into entity links, observations, and later recall/consolidation.

## Changes

- Add `HINDSIGHT_API_RETAIN_NARRATOR` / per-bank `retain_narrator`.
- Stop using the bank profile name as the retain narrator.
- Fall back to `AI assistant` when the configured narrator looks like a bank/source/tag label or prompt-injection text.
- Strip routing-only metadata keys and values before building the fact extraction prompt.
- Preserve semantic metadata like document title/author, including safe nested values.
- Expose the new field through REST models, generated clients, OpenAPI, docs, and the docs skill references.
- Add regression tests with sanitized examples covering narrator fallback and routing metadata filtering.

## Tests

Using an external `pgvector/pgvector:pg16` PostgreSQL container because the local pg0 extension requires a newer host glibc than this Debian runner provides:

```bash
HINDSIGHT_API_DATABASE_URL=postgresql://hindsight:hindsight@127.0.0.1:5557/hindsight \
HINDSIGHT_API_MIGRATION_DATABASE_URL=postgresql://hindsight:hindsight@127.0.0.1:5557/hindsight \
uv run pytest \
  tests/test_retain_narrator.py \
  tests/test_fact_extraction_metadata.py \
  tests/test_hierarchical_config.py \
  tests/test_bank_template_configurable_fields.py \
  -q -n0 --timeout 300
```

Result: `45 passed, 2 warnings`.

Pre-commit also ran successfully:

- `generate-docs-skill.sh`
- `lint.sh`

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1100,6 +1100,13 @@ class CreateBankRequest(BaseModel):
         default=None,
         description="Custom extraction prompt. Only active when retain_extraction_mode is 'custom'.",
     )
+    retain_narrator: str | None = Field(
+        default=None,
+        description=(
+            "Semantic narrator label for first-person assistant statements during retain. "
+            "Defaults to a generic AI assistant; do not use bank/source/tag names."
+        ),
+    )
     retain_chunk_size: int | None = Field(
         default=None,
         description="Maximum token size for each content chunk during retain.",
@@ -1141,6 +1148,7 @@ class CreateBankRequest(BaseModel):
             "retain_mission",
             "retain_extraction_mode",
             "retain_custom_instructions",
+            "retain_narrator",
             "retain_chunk_size",
             "enable_observations",
             "observations_mission",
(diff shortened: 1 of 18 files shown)


nicoloboschi (changes requested):
you can use the 'context' field to explain the LLM how to interpret the fact, no need for a new field

PR #1890, opened 2026-06-01
fix(embed): launch Windows daemon via pythonw to stop ConPTY terminal tab

### LINKED ISSUE #1885: Bug: Windows 11 - hindsight-api.exe spawns visible Windows Terminal tab despite CREATE_NO_WINDOW
## Description
On Windows 11 with Windows Terminal as default, hindsight-api.exe opens a visible Windows Terminal tab when the daemon starts. ConPTY overrides CREATE_NO_WINDOW.

## Root Cause
1. _detach_popen_kwargs originally lacked CREATE_NO_WINDOW + STARTUPINFO(SW_HIDE)
2. Even after adding both, ConPTY intercepts (popup class: CASCADIA_HOSTING_WINDOW_CLASS)
3. Removing WT_SESSION does not prevent takeover

## Proposed Fix
Package hindsight-api.exe with --noconsole for IMAGE_SUBSYSTEM_WINDOWS_GUI (2) instead of CUI (3).

## Environment
- Windows 11, Windows Terminal default
- hindsight-embed 0.7.1
- Python 3.12.10

## Workaround
Watchdog script calling ShowWindow(SW_MINIMIZE) every 60s.

## Related
Fix applied in daemon_embed_manager.py: added CREATE_NO_WINDOW to creationflags and STARTUPINFO with wShowWindow=SW_HIDE. Also removed WT_SESSION from child env. However, Windows Terminal ConPTY still hijacks the process because the executable is compiled with IMAGE_SUBSYSTEM_WINDOWS_CUI (subsystem 3). The real fix requires rebuilding the PyInstaller executable with --noconsole (subsystem 2, GUI).

nicoloboschi:
## Problem

On Windows 11 with Windows Terminal set as the default terminal app, starting the embedded daemon pops a **visible Windows Terminal tab** (fixes #1885).

Root cause (confirmed by the issue): the daemon is launched via `hindsight-api.exe`, a pip console-script wrapper compiled as **console-subsystem (CUI, subsystem 3)**. ConPTY/Windows Terminal hijacks such processes even when spawned with `DETACHED_PROCESS` — and `CREATE_NO_WINDOW` is *ignored* when combined with `DETACHED_PROCESS`, so it can't suppress the tab either.

The issue's proposed `--noconsole`/PyInstaller fix doesn't apply here (this repo has no PyInstaller build — the `.exe` is a pip-generated console script), but the underlying conclusion is right: run a **GUI-subsystem** process.

## Fix

When resolving the daemon launch command on Windows, launch through the **GUI-subsystem interpreter** `pythonw.exe -m hindsight_api.main` instead of the console `hindsight-api.exe`. `pythonw.exe` (subsystem 2) never allocates a console, so ConPTY has nothing to hijack — no window appears.

- `pythonw.exe` is resolved next to `sys.executable`, whose environment is the one we just confirmed contains `hindsight-api` (the console exe was found in its scripts dir), so the `-m hindsight_api.main` import always resolves.
- Falls back to the console exe when `pythonw.exe` is missing — best-effort, never breaks startup.
- Only applied to the standard pip/venv install path. `--target` installs and the `uvx` fallback keep the console exe (their interpreter can't import `hindsight_api` without extra path wiring).
- Extracted `_dev_api_command()` and added `_windows_gui_interpreter()` for clarity/testability.

## Tests

Added 3 regression tests to `test_profile_daemon_config.py`:
- Windows + `pythonw.exe` present → uses GUI interpreter
- Windows without `pythonw.exe` → console-exe fallback
- POSIX → console entry point unchanged

Full embed suite passes (84 tests), lint + project lint hook green, `/code-review` clean.

## Caveat

Verified by reasoning + unit tests, not a live Windows run (no Windows host available here). A manual smoke test on the reporter's setup would be the final confirmation.

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -11,6 +11,7 @@ import os
 import platform
 import re
 import subprocess
+import sys
 import sysconfig
 import time
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


sujianddd-dev:
@nicoloboschi Manually verified the fix on Windows 11 (Build 26200) — works correctly ✅

**Verification results:**

1. `_find_api_command()` now returns `pythonw.exe -m hindsight_api.main` instead of `hindsight-api.exe`:
   ```
   ['C:\...\venv\Scripts\pythonw.exe', '-m', 'hindsight_api.main']
   ```

2. Confirmed `pythonw.exe` is subsystem **2 (WINDOWS_GUI)** — by definition never allocates a console, so Windows Terminal ConPTY has nothing to hijack.

3. Started the daemon via pythonw → **no terminal tab appeared** (verified visually).

The fix is solid. The root cause (CUI subsystem + ConPTY hijack) is fully addressed by routing through the GUI interpreter. Thanks for the quick turnaround!

Issue #1795, opened 2026-05-27
0.7.0: oversized retain split can enqueue concurrent chunks for same document, causing worker thread exhaustion and ANN FK failures


Korayem:
## Summary

After upgrading a Railway deployment from Hindsight `0.6.2` to `0.7.0`, I verified that the new oversized single-item split logic is present in the deployed worker image. The previous behavior of passing one huge retain payload as a single worker operation appears fixed.

However, retrying one large historical conversation still failed operationally because the split produced many concurrent retain operations for the same `document_id`. That created resource pressure and final ANN/linking races.

## Environment

- Hindsight package: `hindsight-api-slim` `0.7.0`
- Deployment: Railway workers using `ghcr.io/vectorize-io/hindsight-api:latest`
- Worker image digest observed: `sha256:5800437d64e5ea21fbd5e99e0b4ef5a3f272b5eb7cbeaaa16c8a1de236b64187`
- Worker settings observed in logs:
  - `Max slots: 6`
  - `Slot reservations: consolidation=1`
  - `Shared pool: 5`
- Worker replicas: 2

## Reproduction shape

Retain one large document/conversation, around 8.8 MB, as a single Hindsight document.

The document was submitted as one retain request with one `document_id`. In `0.7.0`, the oversized content was split, which is good. But it was split into many child retain operations for the same `document_id`.

Observed operation state for that one document:

- `19` retain operations completed before manual cancellation
- `61` retain operations were cancelled to stop the loop/resource burn
- `0` active operations after cleanup
- total child retain operations observed for that one document: `80`

## Observed failures

Workers restarted with native thread exhaustion:

```text
libgomp: Thread creation failed: Resource temporarily unavailable
```

The same document repeatedly produced final ANN failures:

```text
[streaming] Final ANN pass failed for document <document_id> (units may have been superseded by concurrent retain)
asyncpg.exceptions.ForeignKeyViolationError: insert or update on table "memory_links" violates foreign key constraint "fk_memory_links_from_unit_id_memory_units"
DETAIL: Key (from_unit_id)=(...) is not present in table "memory_units".
```

There was also at least one OpenRouter connection warning during the run, but the dominant new failure mode after `0.7.0` was concurrency/resource related rather than the previous single huge payload behavior.

## Expected behavior

Oversized retain should be chunked without allowing chunks for the same `(bank_id, document_id)` to race each other.

[...]

PR #1894, opened 2026-06-01
fix(retain): close to_unit_id deferred-FK race on memory_links inserts (#1882)

### LINKED ISSUE #1882: fk_memory_links_to_unit_id_memory_units violation under sustained concurrent retain (mirror of #1795)
# `fk_memory_links_to_unit_id_memory_units` violation under sustained concurrent retain (mirror of #1795)

## Summary

The `from_unit_id` side of the `memory_links` → `memory_units` foreign-key race was fixed in #1805 (closes #1795). The **`to_unit_id` side** is the mirror case and still races under the same load pattern. Hitting it at v0.7.1 reproducibly during multi-bank batch ingest.

## Environment

- Hindsight: **v0.7.1** (self-hosted, `ghcr.io/vectorize-io/hindsight:latest` digest as of 2026-05-28)
- Embedded `pg0` Postgres 18.1.0
- All known earlier fixes confirmed present:
  - PR #1398 (deferred FK) — error surfaces at COMMIT, not INSERT (so this isn't the pre-1398 race)
  - PR #1805 / Issue #1795 (`from_unit_id` orphan fix) — `from_unit_id` no longer races
  - 3-phase retain pipeline (v0.5.0)
- Single tenant, four banks active (writes scoped per bank, no cross-bank writes)

## Observed behavior

A single nightly process posts to `/v1/default/banks/<bank>/memories` ~225 times in <1 second across two banks (batched as one async retain per call, `async: true`, `batch_size: 1`). On most days, 1–5 of those operations fail in Hindsight's queue with the FK violation below; the rest commit fine. The failed ops remain in `async_operations` with `status='failed'` and never auto-retry.

## Error (verbatim from `async_operations.error_message`)

```
insert or update on table "memory_links" violates foreign key constraint
"fk_memory_links_to_unit_id_memory_units"
DETAIL:  Key (to_unit_id)=(<uuid>) is not present in table "memory_units".

  File "/app/api/hindsight_api/engine/memory_engine.py", line 1337, in execute_task
    await self._handle_batch_retain(task_dict)
  File "/app/api/hindsight_api/engine/memory_engine.py", line 917, in _handle_batch_retain
    await self.retain_batch_async(
  File "/app/api/hindsight_api/engine/memory_engine.py", line 2669, in retain_batch_async
    result, total_usage, total_processed_content_tokens = await self._retain_batch_async_internal(...)
  File "/app/api/hindsight_api/engine/memory_engine.py", line 2787, in _retain_batch_async_internal
    return await orchestrator.retain_batch(...)
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 656, in retain_batch
    return await _streaming_retain_batch(...)
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 1329, in _streaming_retain_batch
    await asyncio.gather(_llm_producer(), _db_consumer())
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 983, in _db_consumer
    await _process_db_batch(...)
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 1278, in _process_db_batch
    await _run_mini_batch_db_work()
  File "/app/api/hindsight_api/engine/retain/orchestrator.py", line 1151, in _run_mini_batch_db_work
    async with conn.transaction():
  File "/usr/local/lib/python3.11/contextlib.py", line 217, in __aexit__
    await anext(self.gen)
  File "/app/api/hindsight_api/engine/db/postgresql.py", line 32, in transaction
    async with self._conn.transaction():
  File "/app/api/.venv/lib/python3.11/site-packages/asyncpg/transaction.py", line 91, in __aexit__
    await self.__commit()
```

Also observed: failed `graph_maintenance` operations with the same `fk_memory_links_to_unit_id_memory_units` violation, indicating the post-link maintenance phase can also reference units that aren't yet visible.

## Reproduction shape

1. Two banks (any two — distinct `bank_id`s, isolated writes)
2. From one upstream process, fire ~225 `POST /v1/default/banks/<bank>/memories` calls within ~1 second, `async: true`, single-item batches
3. Watch `SELECT * FROM async_operations WHERE status='failed'` over the next minute — expect 1–5 rows with the FK violation above

## Hypothesis

The `to_unit_id` FK on `memory_links` is constraint-checked at COMMIT (per PR #1398's deferred-FK design). Under sustained concurrent load, the pre-resolve / insert / post-link 3-phase pipeline (v0.5.0) can produce a transaction whose link insert references a unit ID that another concurrent transaction hasn't yet committed. PR #1805 caught this for `from_unit_id` (the "earlier" side of the link in the resolve order); the `to_unit_id` side appears uncovered.

## Local workaround in place

Sleeping ~50ms between async-retain HTTP calls on our side reduces effective concurrency enough to make the race rare-to-zero in our traffic. This is a UX patch, not a fix — the upstream race still exists for any caller doing burst writes.

## Related

- #1795 / PR #1805 — `from_unit_id` mirror, fixed pre-v0.7.0
- PR #1398 — deferred-FK constraint changes (provides the COMMIT-time surfacing path)
- v0.5.0 — 3-phase retain pipeline
- #1669 — concurrent-load edge case in `batch_retain` (different symptom, similar class)

nicoloboschi:
Fixes #1882.

## Problem

The `memory_links` → `memory_units` FKs are `DEFERRABLE INITIALLY DEFERRED` (migration `9f8e7d6c5b4a`), so an `INSERT` into `memory_links` takes **no lock** on the referenced parent row until `COMMIT`. Temporal and ANN link inserts reference a **pre-existing** neighbor unit as `to_unit_id` (graph maintenance also references a pre-existing `from_unit_id`). A concurrent transaction that commits a `DELETE` of that unit in the window between the link `INSERT` and our `COMMIT` — consolidation pruning observation units (`consolidator.py`), document re-tracking (`fact_storage.py`) — makes the deferred check fail at `COMMIT`:

```
insert or update on table "memory_links" violates foreign key constraint
"fk_memory_links_to_unit_id_memory_units"
DETAIL:  Key (to_unit_id)=(<uuid>) is not present in table "memory_units".
```

The async op is left `status='failed'` with no auto-retry.

#1795 / #1805 only removed one *deleter* (sibling async children sharing a `document_id`) for the `from_unit_id` side. The `to_unit_id` side — and any other deleter — stayed uncovered, which is the mirror case the issue reports.

## Fix

Before inserting links, lock the referenced parent units `FOR KEY SHARE` (held until the surrounding transaction commits, which blocks a concurrent `DELETE`) and drop any link whose endpoint already vanished.

- New `DataAccessOps.lock_referenced_units` — a no-op default for backends with immediate FKs (Oracle has no deferred-FK window), overridden on PostgreSQL to take the row-share lock and report which rows survived.
- `_bulk_insert_links` calls it before the insert and filters vanished endpoints.
- The final ANN pass ran its insert in autocommit, so its insert is now wrapped in a transaction to hold the lock through the `COMMIT`-time check. The other callers (Phase-2 temporal/causal, graph maintenance) already run inside a write transaction.

## Test

Adds a **deterministic** regression test (`test_memory_links_to_unit_id_concurrent_delete.py`) that hand-drives the connection interleaving with no sleeps:

1. conn A inserts a temporal link via the real `_bulk_insert_links` path — uncommitted.
2. conn B deletes the neighbor unit (bounded by `statement_timeout`).
3. conn A commits.

Pre-fix: step 2 succeeds instantly and step 3 raises the FK violation. Post-fix: step 2 blocks on A's `FOR KEY SHARE` lock and is cut off by the timeout; the neighbor survives and the link commits cleanly.

## Verification

- New regression test passes; fails (reproduces #1882) without the fix.
- `test_memory_links_deferred_fk.py`, `test_link_utils.py`, `test_retain.py`, `test_document_tracking.py`, `test_async_batch_retain.py`, `test_causal_relations.py`, `test_graph_maintenance.py`, `test_db_abstraction.py` all pass (233 tests).
- `./scripts/hooks/lint.sh` and `uv run ty check` pass.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -200,6 +200,23 @@ class PostgreSQLOps(DataAccessOps):
         exists_clause: str,
         chunk_size: int = 5000,
     ) -> None:
+        # exists_clause is unused on PostgreSQL: the memory_links → memory_units
+        # FKs are DEFERRABLE INITIALLY DEFERRED, so an INSERT takes no lock on the
+        # referenced parent rows until COMMIT — a concurrent committed DELETE in
+        # that window (consolidation pruning observations, document re-tracking)
+        # trips fk_memory_links_{to,from}_unit_id_memory_units at COMMIT (#1882),
+        # and a WHERE EXISTS guard can't prevent it (the row passes the check,
+        # then is deleted before the deferred check runs). Instead a CTE locks the
+        # referenced units FOR KEY SHARE in the *same statement*: the lock blocks a
+        # concurrent DELETE until our transaction commits and is held through the
+        # deferred check, and the INSERT only takes links whose endpoints are in
+        # the locked set, so rows that already vanished are dropped. Folding it
+        # into the one INSERT keeps this to a single round-trip — no extra query
+        # and no surrounding transaction needed. (Oracle's immediate FK has no
+        # such window and uses exists_clause via its own bulk_insert_links.)
+        from ..schema import fq_table
+
+        mu_table = fq_table("memory_units")
         from_ids = [lnk[0] for lnk in sorted_links]
         to_ids = [lnk[1] for lnk in sorted_links]
         types = [lnk[2] for lnk in sorted_links]
@@ -208,24 +225,37 @@ class PostgreSQLOps(DataAccessOps):
 
         for chunk_start in range(0, len(sorted_links), chunk_size):
             chunk_end = min(chunk_start + chunk_size, len(sorted_links))
+            chunk_from = from_ids[chunk_start:chunk_end]
+            chunk_to = to_ids[chunk_start:chunk_end]
+            # Distinct referenced parents, sorted so concurrent inserters acquire
+            # the row-share locks in a consistent order (avoids deadlocks; same
+            # convention as the (from, to) link sort).
+            referenced = sorted({str(x) for x in chunk_from} | {str(x) for x in chunk_to})
             await conn.execute(
                 f"""
+                WITH locked AS (
+                    SELECT id FROM {mu_table}
+                    WHERE id = ANY($7::uuid[])
+                    ORDER BY id
+                    FOR KEY SHARE
+                )
                 INSERT INTO {table}
                     (from_unit_id, to_unit_id, link_type, weight, entity_id, bank_id)
                 SELECT f, t, tp, w, e, $6
                 FROM unnest($1::uuid[], $2::uuid[], $3::text[], $4::float8[], $5::uuid[])
-                    AS t(f, t, tp, w, e)
-                {exists_clause}
+                    AS u(f, t, tp, w, e)
+                WHERE f IN (SELECT id FROM locked) AND t IN (SELECT id FROM locked)
                 ON CONFLICT (from_unit_id, to_unit_id, link_type,
                              COALESCE(entity_id, '{nil_entity_uuid}'::uuid))
                 DO NOTHING
                 """,
-                from_ids[chunk_start:chunk_end],
-                to_ids[chunk_start:chunk_end],
+                chunk_from,
+                chunk_to,
                 types[chunk_start:chunk_end],
                 weights[chunk_start:chunk_end],
                 entity_ids[chunk_start:chunk_end],
                 bank_id,
+                referenced,
                 timeout=300,
             )
 
--- a/hindsight-api-slim/tests/test_memory_links_to_unit_id_concurrent_delete.py
+++ b/hindsight-api-slim/tests/test_memory_links_to_unit_id_concurrent_delete.py
@@ -1,0 +1,126 @@
+"""Regression for #1882 — the ``to_unit_id`` side of the memory_links FK race.
+
+Mirror of #1795 (which covered ``from_unit_id``). Temporal / ANN link inserts
+reference a **pre-existing neighbor** unit as ``to_unit_id`` and bulk-insert with
+``skip_exists_check=True`` (see ``create_temporal_links_batch_per_fact`` /
+``compute_semantic_links_ann``). The FK
+``fk_memory_links_to_unit_id_memory_units`` is ``DEFERRABLE INITIALLY DEFERRED``
+(migration ``9f8e7d6c5b4a``), so the INSERT takes **no lock** on the parent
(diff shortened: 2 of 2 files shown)


nicoloboschi:
## Perf evaluation of the fix

Revised the implementation to a **single-statement locking CTE** (instead of a separate `SELECT … FOR KEY SHARE` + Python filter + a transaction wrap around the autocommit ANN insert). This keeps the link insert to **one round-trip, no new transaction** — so retain's round-trip/transaction shape is unchanged. The change is now contained entirely in `PostgreSQLOps.bulk_insert_links`.

### End-to-end retain throughput (mock LLM + pg0, `--scale medium`, 1000 items, A/B toggling only this file)
| | run 1 | run 2 |
|---|---|---|
| baseline | 24.72 it/s | 26.08 it/s |
| this fix | 24.09 it/s | 24.28 it/s |

~0–4%, within run-to-run noise — link inserts are a small slice of retain (embeddings + fact insert dominate).

### Micro-benchmark of the changed statement (isolates the link insert; 1000 links/call, real schema, 60 iters)
**Typical batch (~70 distinct referenced units):**
| variant | p50 | mean |
|---|---|---|
| baseline `skip_exists_check=True` (temporal/causal) | 5.39ms | 6.21ms |
| baseline `WHERE EXISTS` (semantic/graph) | 6.18ms | 6.29ms |
| **this fix (locking CTE)** | 5.70ms | **5.86ms** |

**Worst case (~1050 distinct referenced units locked):**
| variant | p50 | mean |
|---|---|---|
| baseline `skip_exists_check=True` | 5.45ms | 5.85ms |
| **this fix (locking CTE)** | 7.53ms | 8.35ms |

### Takeaways
- vs the cheapest baseline path (`skip_exists_check=True`, used by temporal/causal links): **+~0.3ms** per insert for typical batches, up to **+~2ms** for large 1000-link batches with ~1000 distinct neighbors (the `FOR KEY SHARE` lock acquisition + semi-join filter). Cost scales with the number of *distinct referenced units*.
- vs the `WHERE EXISTS` path (semantic ANN / graph maintenance): **roughly neutral to faster** — one CTE lock-scan replaces two per-row `EXISTS` subqueries.
- No extra round-trips, no extra transactions; the burst-write scenario from the issue uses single-item retains (small link batches), so the practical overhead there is the ~0.3ms end.

Net: a small, bounded cost on the link-insert step (invisible end-to-end within noise) in exchange for eliminating the FK-violation-at-COMMIT failure under concurrent retain.

PR #1908, opened 2026-06-01
feat(admin): add offline embedding reembed command

### LINKED ISSUE #1884: No supported path to change embedding models without re-running LLM extraction (dimension migration refuses non-empty tables; no reembed admin command)
### Summary

There's no supported way to switch embedding models (or vector extensions) on an existing bank without destroying and regenerating all extracted facts via the LLM — even though the embeddings are a pure function of text already stored in the database. An encoder-only re-vectorization in place is feasible but unimplemented.

### Environment

- Hindsight 0.7.1, self-hosted, Postgres + pgvector (HNSW).
- Want to move from a 384-dim encoder (`bge-small-en-v1.5`, 512-token hard cap) to a 1024-dim encoder (`bge-m3`, 8192 ctx). The 512-token cap truncates / 500s on longer consolidation inputs; bge-m3 fixes it.

### What happens

On startup with the new model configured, `_migrate_table_embedding_dimension` (`migrations.py:~472`) detects the dim mismatch and raises:

```
Cannot change embedding dimension from 384 to 1024: memory_units table contains N rows with embeddings.
To change dimensions, you must either:
  1. Re-embed all data: DELETE FROM <schema>.memory_units; then restart
  2. Use a model with 384-dimensional embeddings
```

The vector-extension guard (`migrations.py:~740`) raises the same way for `memory_units` / `learnings` / `pinned_reflections`.

### Why this is a real gap (not just docs)

1. **`memory_units.text` already holds the final fact text** and `embedding` is derived solely from it (`models.py`). Re-vectorizing requires only an encoder pass — no LLM, no re-extraction.
2. **`documents.original_text` is also retained.**
3. **Option 1's "then restart" is misleading.** A restart does *not* re-extract. After `DELETE FROM memory_units` the facts are simply gone. The only programmatic ways to repopulate are:
   - the per-document `reprocess` endpoint, which *"deletes the existing memory units and re-extracts facts using the current engine configuration"* — i.e. a **full LLM re-extraction per document**, or
   - full re-ingest — also full LLM re-extraction.
4. **The admin CLI has no `reembed` / `reindex` command** (only `backup`, `restore`, `run-db-migration`, `decommission-worker(s)`, `worker-status`).

Net effect: changing encoders forces re-running LLM extraction over the entire corpus, paying for and risking the quality variance of a fresh extraction, purely to recompute a vector that's a deterministic function of text already on disk.

### Requested

An admin command (or migration mode) that re-vectorizes in place, **encoder-only, no LLM**:

1. For each row in `memory_units` (and `mental_models`, `learnings`, `pinned_reflections` as applicable): re-encode the stored `text` with the configured embedding model and overwrite `embedding`.
2. `ALTER` the `embedding` column to the new dimension and rebuild the HNSW / vchord / diskann indexes (the migration already drops/recreates these for the empty-table path).
3. Batch + resumable; safe to run after `admin backup`.

Sketch: `hindsight admin reembed --schema public [--bank <id>] [--batch 512]`.

### Workaround in use

A truncating embeddings proxy in front of the 384-dim server (caps each input to 510 tokens via the server's own `/tokenize` + `/detokenize`) restores the old silent-truncate behavior and keeps consolidation unblocked — but it's a stopgap; the real fix is moving to a longer-context encoder, which today requires the wipe-and-re-extract path above.

### LINKED ISSUE #2073: Automatically rebuild memory embedding indexes after changing the embedding model
### Use Case

I'm using Hindsight with Hermes agent. I use Hindsight to build my own memory bank to improve daily works with agent.

### Problem Statement

In self-hosted deployments, users may want to change the embedding model for several reasons:

Switching from the default English-focused model to a multilingual or Chinese-friendly embedding model.
Moving from a local embedding model to an API-based provider such as OpenAI, Cohere, Google, TEI, LiteLLM, etc.
Upgrading to a larger or more accurate embedding model.
Changing the vector dimension intentionally, for example from a (384)-dimensional model to a (1024)-dimensional model.
Rebuilding memories after changing the embedding provider while keeping the original memory content.

At the moment, users need to manually back up, wipe, reconfigure, and backfill the memory database. This is error-prone and can discourage users from improving their embedding setup after initial deployment.

When the embedding model or embedding provider is changed, Hindsight could provide an official way to rebuild existing memories using the new embedding configuration.

### Proposed Solution

Possible expected workflow:

hindsight embeddings rebuild

or:

hindsight migrate-embeddings --provider local --model BAAI/bge-m3

The command would:

Detect the currently stored embedding metadata.
Detect the new embedding provider, model name, and vector dimension.
Warn the user if the vector dimension has changed.
Preserve the original memory text and metadata.
Recompute embeddings for all existing memories using the new model.
Recreate or update the vector index according to the new dimension.
Provide progress information and a final migration report.

### Alternatives Considered

_No response_

### Priority

### Additional Context

To avoid accidental data loss, the rebuild process could include:

A required confirmation prompt before destructive schema/index changes.
A dry-run mode:
hindsight embeddings rebuild --dry-run
Automatic backup recommendation or backup generation before migration.
A lock to prevent writes during the rebuild.
A resume mechanism in case the rebuild is interrupted.
A clear warning when old and new embedding dimensions differ.
Metadata tracking for embedding provider, model name, dimension, and rebuild timestamp.

### Checklist

Sanderhoff-alt:
## Summary

Adds a PostgreSQL-only `hindsight-admin reembed` workflow for migrating stored embeddings after changing the configured embedding provider, model, endpoint, or output dimension.

The command recomputes existing `memory_units` and `mental_models` embeddings using the current Hindsight embedding configuration, writes the new vectors into shadow columns, prepares compatible shadow vector indexes, stages semantic links for the new embedding space, and then atomically cuts over the live embedding columns.

Fixes #1884
Fixes #2073

## Motivation

Today Hindsight validates the configured embedding dimension at startup. If the database already contains stored vectors and the new embedding model uses a different dimension, startup fails because the existing `embedding` columns cannot be safely altered in place.

That blocks normal model migrations such as:

- local 384-dimensional embeddings to a 1024-dimensional OpenAI-compatible model;
- 1024-dimensional embeddings back to a 384-dimensional model;
- same-dimension model or endpoint changes where vectors are still in a different embedding space.

The database already stores enough text to regenerate embeddings without rerunning LLM extraction, so this PR adds an explicit offline admin path instead of asking operators to drop data or rebuild banks from scratch.

## User-Facing Commands

This PR adds three admin entry points:

```bash
# Plan only. This is read-only and does not require --yes.
hindsight-admin reembed --schema public --dry-run

# Recompute embeddings and cut over before the command exits.
hindsight-admin reembed --schema public --yes
hindsight-admin reembed --schema public --schema tenant_acme --yes

# Discover all Hindsight schemas in the configured PostgreSQL database.
hindsight-admin reembed --all-schemas --yes

# Inspect or abandon the single active migration for a schema.
hindsight-admin reembed-status --schema public
hindsight-admin reembed-abandon --schema public --yes
hindsight-admin reembed-abandon --schema public --orphan-shadow-state --yes
```

Command behavior:

- `reembed` is a foreground maintenance command. It prints progress for validation, embedding batches, shadow index creation, semantic link staging, cutover, and completion, then exits with either a completed report or an error.
- The target embedding model is intentionally not passed as a CLI argument. `reembed` reads the normal Hindsight environment, so operators update the service configuration first, run the admin command with that same environment, and restart the service with the same settings.
- If a run is interrupted, running `reembed` again for the same schema resumes the active migration when the embedding configuration fingerprint still matches.
- Only one active migration is allowed per schema, enforced by both a runtime session advisory lock and a partial unique index on active migration rows. A second mutating command fails fast instead of resuming the same migration concurrently.
- `reembed-status` is read-only and reports the active migration and any shadow state for one schema.
- `reembed-abandon` removes incomplete shadow state without modifying the live `embedding` columns. It is not a rollback for completed cutovers.

Key options:

- `reembed --schema`, `-s <schema>`: database schema to reembed. Can be supplied multiple times. Required unless `--all-schemas` is used.
- `reembed --all-schemas`: discover and reembed all Hindsight schemas in the current PostgreSQL database. Mutually exclusive with `--schema`.
- `reembed --batch-size <int>`: rows to embed per provider batch. Must be at least `1`. Default: `100`.
- `reembed --max-retries <int>`: outer retry count for each embedding batch. Must be at least `0`. Default: `3`.
- `reembed --index-max-parallel-maintenance-workers <int>`: optional session override for PostgreSQL `max_parallel_maintenance_workers` while building shadow vector indexes. Default: unset, so PostgreSQL keeps its configured value. Passing `0` forces serial index builds, which avoids dynamic shared-memory pressure in constrained containers.
- `reembed --dry-run`: read-only planning mode. It does not create admin tables, shadow columns, migration rows, indexes, or staging rows, and it does not require `--yes`.
- `reembed-status --schema`, `-s <schema>`: read-only status inspection for one database schema. Required.
- `reembed-abandon --schema`, `-s <schema>`: abandon the active reembed for one database schema. Required.
- `reembed-abandon --orphan-shadow-state`: clean orphan shadow state even when there is no active migration row. This is for recovery from leftover `embedding_reembed` columns, shadow indexes, or staging rows.
- `--yes`, `-y`: skip the confirmation prompt for mutating commands.

## Schema Selection

The command operates inside the single PostgreSQL database selected by `HINDSIGHT_API_DATABASE_URL`.

For single-tenant deployments, operators should pass the configured Hindsight database schema explicitly. That is usually `public`, or the value of `HINDSIGHT_API_DATABASE_SCHEMA` if the deployment overrides it.

The `--all-schemas` catalog discovery check is deliberately persistent-state based. It looks for `alembic_version` plus early core tables such as `banks` or `memory_units`, returns the configured base schema first when present, and does not depend on `TenantExtension.list_tenants()`, whose Supabase implementation is an in-process cache and can be empty in a one-off admin container.

## Operational Model

This is intentionally an offline maintenance workflow.

Recommended Docker/Compose flow:

```bash
# 1. Update the service environment to the new embedding configuration.
# 2. Stop the Hindsight service and any worker service that can write the schema.
docker compose stop <hindsight-service>

# 3. Run the admin command in a one-off container that reuses the same env,
#    network, and database volume/connection.
docker compose run --rm --no-deps <hindsight-service> \
  hindsight-admin reembed --schema public --yes

# 4. Restart the service with the same new embedding configuration.
docker compose up -d <hindsight-service>
```

The command does not provide online dual-write or recall quiescing. Operators should stop retain/update/recall traffic and workers for the target schema while the migration is running.

`docker exec` into a live service container is not the recommended way to run the mutating command. It can be useful for read-only `reembed-status`, but the write path should run while the service/worker processes for that schema are stopped. A one-off Compose container is safer because it uses the same image, environment, network, and volumes while replacing the normal service command with just `hindsight-admin ...`.

For embedded pg0 deployments, the service must have a persistent mount for the pg0 data directory. If pg0 data only lives in the container writable layer, a one-off container will not necessarily see the same database.

If the upgraded service has already started successfully, the new Alembic admin tables should already exist and this extra step can be skipped. If the upgraded image has not been started yet, or `reembed` reports missing `embedding_reembed_migrations` / `embedding_reembed_semantic_links`, insert this schema-preparation command after `docker compose stop` and before `reembed`. Do not pass `--embedding-dimension`, because dimension changes are handled by `reembed` through shadow columns and cutover.

```bash
docker compose run --rm --no-deps <hindsight-service> \
  hindsight-admin run-db-migration --schema public
```

Recommended pip/uv and hindsight-embed flow:

```bash
# 1. Update the environment used by the service to the new embedding configuration.
# 2. Stop the API/worker process that can read or write the target schema.
systemctl stop hindsight-api  # example; use the deployment's actual supervisor

# 3. Run admin from the same installation with the same target DB and embedding env.
export HINDSIGHT_API_DATABASE_URL=<same-database-url-used-by-service>
export HINDSIGHT_API_EMBEDDINGS_PROVIDER=<new-provider>
export HINDSIGHT_API_EMBEDDINGS_LOCAL_MODEL=<new-model>  # if applicable
hindsight-admin reembed --schema public --yes

# 4. Restart the service with the same new embedding configuration.
systemctl start hindsight-api
```

The same schema-preparation rule applies outside Docker: if the upgraded service has not run migrations yet, run the migration entrypoint against the same database before `reembed`:

```bash
hindsight-admin run-db-migration --schema public
```

For `hindsight-embed` / `hindsight-all`, stop the profile daemon first and run `hindsight-admin` with the same profile-derived environment. `hindsight-admin` does not have a `--profile` flag and does not automatically read `~/.hindsight/embed` or `~/.hindsight/profiles/<name>.env`, so the database URL and embedding settings must be provided explicitly. For a named profile without `HINDSIGHT_EMBED_API_DATABASE_URL`, the default embedded DB URL is `pg0://hindsight-embed-<profile>`; if the profile overrides the DB URL, use that override instead.

Example named-profile flow:

```bash
profile=myapp

hindsight-embed -p "$profile" daemon stop

set -a
. "$HOME/.hindsight/profiles/$profile.env"
set +a

export HINDSIGHT_API_DATABASE_URL="${HINDSIGHT_EMBED_API_DATABASE_URL:-pg0://hindsight-embed-$profile}"
hindsight-admin reembed --schema public --yes

hindsight-embed -p "$profile" daemon start
```

## Database Changes

Adds a new Alembic revision:

```text
c3efa30f38c9_add_reembed_admin_tables.py
```

PostgreSQL schemas receive two persistent admin tables:

- `embedding_reembed_migrations`
- `embedding_reembed_semantic_links`

The Oracle branch is a deliberate no-op because the current admin CLI is PostgreSQL-only.

The migration table stores:

- `migration_id`;
- target provider/model/dimension/vector extension;
- a non-secret `model_identity` JSON payload;
- a `config_fingerprint` hash of that identity;
- phase state for embeddings, shadow indexes, and semantic links;
- overall status and timestamps;
- the last error message.

The semantic-link staging table stores only the columns needed to populate the current PostgreSQL `memory_links` schema:

- `migration_id`;
- `bank_id`;
- `from_unit_id`;
- `to_unit_id`;
- `weight`.

It intentionally does not contain a `metadata` column because current PostgreSQL `memory_links` does not have one.

The Alembic revision is the source of truth for these persistent admin tables. The mutating `reembed` command requires the tables to already exist and fails fast with an Alembic/admin-table migration hint if they are missing. `--dry-run` only reports that the tables are missing; it does not create or modify database objects.

The runtime shadow state is not part of Alembic:

- `memory_units.embedding_reembed`
- `mental_models.embedding_reembed`
- shadow vector indexes
- rows in `embedding_reembed_semantic_links`

These are created, resumed, cut over, or abandoned by the admin command.

## Migration Flow

For each schema, `reembed`:

1. resolves the target schema set;
2. validates that the core Hindsight tables exist;
3. initializes the current embedding provider and detects its output dimension;
4. creates or resumes a migration row;
5. adds `embedding_reembed vector(<target_dim>)` shadow columns;
6. creates runtime worklist indexes for rows where the live `embedding` is non-null and the shadow embedding is still null;
7. batches rows through those worklist indexes;
8. regenerates embeddings outside the database transaction;
9. writes each completed batch into the shadow columns;
10. creates compatible shadow vector indexes or marks ScaNN indexes deferred;
11. rebuilds semantic links against `embedding_reembed`;
12. marks the migration `prepared` once embedding, index, and semantic-link phases are complete;
13. cuts over inside a transaction by swapping columns, replacing semantic links, cleaning staging rows, dropping worklist indexes, and marking the migration `completed`.

If the process fails before cutover, the old live columns, old vector indexes, and old semantic links remain in place. The same command can be rerun to resume.

The state machine is persisted in `embedding_reembed_migrations`:

- new migration: `status = 'running'`;
- after every source row with a live embedding has a shadow embedding: `embedding_state = 'complete'`;
- after shadow vector indexes are ready, or explicitly deferred for ScaNN: `shadow_indexes_state = 'ready'` or `deferred`;
- after semantic-link staging finishes, even if it stages zero rows: `semantic_links_state = 'complete'`;
- only when all three phases are complete/ready/deferred does the command mark `status = 'prepared'`;
- a prepared migration is then cut over immediately in the same run;
- successful cutover marks `status = 'completed'`;
- failures mark the active migration `failed` and keep phase progress for resume.

The actual embedding API calls are outside the cutover transaction. The database-visible switch happens in a PostgreSQL transaction that renames/drops columns, swaps indexes, replaces eligible semantic links, clears staging rows, and marks the migration complete.

## Resume and Abandon Semantics

Each migration stores a non-sensitive embedding configuration fingerprint based on provider/model/endpoint/dimension identity. Resume refuses to continue if the current embedding configuration no longer matches the active migration. This prevents mixing vectors from different models or endpoints in the same shadow column.

Progress is database-backed:

- committed shadow vectors are skipped on rerun;
- source rows with `embedding IS NULL` remain null and are not counted as work;
- semantic-link staging is keyed by `migration_id`;
- completed cutovers are not abandoned or rolled back by `reembed-abandon`.

`reembed-abandon --orphan-shadow-state` exists for recovery from manual or crash-created shadow state when there is no active migration row.

The fingerprint includes non-secret `embeddings_*` configuration fields along with provider/model/dimension/vector-extension identity. It excludes API keys, service-account keys, and other credential-like fields. This is meant to catch same-dimension changes such as switching OpenAI-compatible base URLs or model aliases, where the table type would still match but the embedding space would not.

`reembed-abandon` is intentionally not a rollback. If a migration already completed, the live `embedding` column has been cut over and the correct recovery path is a new reembed or restoring from backup.

## Preflight Checks

Before work starts, the command checks:

- required tables exist in the target schema;
- no active migration with a different config fingerprint is being resumed;
- no orphan shadow state exists without a migration row;
- target rows do not reference `bank_id` values missing from `banks`;
- the configured vector extension can be resolved before shadow indexes are created.

The orphan-bank check matters because `memory_units` does not have a foreign key to `banks`. Without this check, per-bank shadow indexes and cutover bookkeeping could silently ignore rows that still contain embeddings.

## Embedding Input Reconstruction

The command does not rerun LLM extraction. It reconstructs embedding input from the stored database state:

- regular retain facts use the stored `memory_units.text` plus retained date and entity context where available;
- consolidation observations keep the current bare observation-text semantics and are identified by `fact_type = 'observation'` with non-empty `source_memory_ids`;
- mental models use the current canonical database state: `name + " " + content`.

This keeps the operation focused on encoder migration while avoiding a new fact extraction pass.

The retain-fact path shares the same embedding-text builder used by runtime retain. `augment_texts_with_dates()` now delegates to a common helper, and the reembed row adapter only maps database fields into that helper. This avoids keeping a second hand-written date/entity augmentation rule in the admin migration path.

For regular retain facts, the reconstruction mirrors the current retain-side shape as closely as the database allows:

- base text comes from `memory_units.text`;
- date context uses `occurred_start`, `occurred_end`, and `mentioned_at`;
- date strings use the existing month/year readable form;
- entity context is recovered through `unit_entities -> entities`;
- entity names are sorted deterministically by lower-cased canonical name and id.

The original extracted entity list order is not stored, so this path cannot perfectly reconstruct that order. The deterministic canonical-entity ordering is an explicit compromise that stays closer to retain-time embedding input than using bare text alone.

For mental models, the command uses current database state:

```text
name + " " + content
```

The runtime create/update paths and reembed share the same mental-model embedding-text helper. That intentionally normalizes historical update quirks where content-only or name-only updates may have left embeddings derived from a slightly different string than the current row representation.

## Semantic Links

The old `memory_links.link_type = 'semantic'` rows describe the old embedding space, so the migration stages new semantic links using `embedding_reembed` before cutover.

Notable details:

- staging aligns with the current `memory_links` schema and does not write a nonexistent `metadata` column;
- staging is batched by `bank_id + fact_type`, making those values constants in the ANN query so PostgreSQL can use the matching non-ScaNN partial vector index;
- each staging group applies backend-specific low-latency ANN tuning with transaction-local `SET LOCAL`;
- semantic links are inserted with `entity_id = NULL`;
- self-links are excluded;
- consolidation observations are excluded as seed rows, so the migration does not create new outgoing semantic links from observations that currently rely on `source_memory_ids`;
- staging rows are deleted after a successful cutover.

Cutover deletes old outgoing semantic links for reembedded, eligible `memory_units` and inserts the staged replacements. It does not delete unrelated non-semantic links, and it does not use a nonexistent staging `metadata` field.

The rebuild mirrors the current streaming retain final ANN pass:

```text
top_k = 20
threshold = 0.7
```

Current normal retain writes use the streaming pipeline, and that path creates semantic links in a final ANN pass with `top_k = 20`. Reembed imports the same retain semantic-link policy constants instead of defining separate rebuild values. Graph maintenance can later top up victims after deletes using a higher cap, but reembed is an embedding-space migration rather than a full graph-repair command. Using the streaming retain cap avoids expanding every eligible row to graph-maintenance density during model migration.

## Vector Index Handling

`_vector_index.index_using_clause()` now accepts a column name so the same backend dispatcher can create indexes on either `embedding` or `embedding_reembed`.

The reembed path:

- preserves existing live vector indexes during preparation;
- creates runtime partial btree worklist indexes on `(bank_id, id)` for `memory_units` and `mental_models` so each batch can find unfinished rows without repeatedly scanning already-completed rows;
- builds shadow indexes on `embedding_reembed`;
- creates non-ScaNN per-bank/fact-type memory-unit shadow vector indexes for every existing bank and supported fact type, matching runtime index provisioning so empty fact-type indexes are not lost at cutover;
- renames ready shadow indexes during cutover;
- removes old indexes during cutover, either by explicit drop/rename or by dropping the old `embedding` column when ScaNN shadow index creation is deferred;
- uses per-bank partial memory-unit indexes for non-ScaNN backends;
- keeps ScaNN as a global-index backend;
- validates pgvector HNSW dimensional limits before creating shadow columns or migration rows;
- resolves Azure `pg_diskann` when the configured extension is `pgvectorscale`, so shadow DiskANN indexes use the Azure-compatible `max_neighbors` clause.

ScaNN AUTO index creation is deferred per indexed table. If either `memory_units` or `mental_models` is below the minimum row count, the migration marks shadow indexes as `deferred` instead of attempting to create an invalid ScaNN index on a small table.

This per-table ScaNN check is important because the ScaNN AUTO minimum applies to each index/table, not to the combined row count across all embedding tables. For example, a schema may have enough `memory_units` but very few `mental_models`; in that case trying to create a ScaNN mental-model index would fail even though the combined count is large.

By default, reembed does not override PostgreSQL parallel index-build settings. Operators who run PostgreSQL in containers with small `/dev/shm` can pass `--index-max-parallel-maintenance-workers 0` to force serial shadow index builds for this admin session. Operators who provision larger shared memory can omit the flag and keep PostgreSQL's configured parallelism.

For Azure DiskANN, the configured value remains `pgvectorscale`, but the actual installed extension can be `pg_diskann`. The shadow-index path now resolves that before generating SQL, matching the runtime migration/index reconciliation path. That prevents creating open-source pgvectorscale SQL with `num_neighbors` on an Azure deployment that expects `max_neighbors`.

## Backup and Restore Guardrails

The new admin tables are included in `BACKUP_TABLES`.

Backup and restore now fail fast if the target schema has active or incomplete reembed state, including:

- active/failed/prepared reembed migration rows;
- `embedding_reembed` shadow columns;
- shadow vector indexes;
- semantic-link staging rows.

This avoids taking or restoring a binary table copy while the schema has temporary columns that are not part of a stable serving state.

Restore has the same guard as backup because restore uses whole-table binary COPY. If the target schema still has a temporary `embedding_reembed` column, the binary COPY column layout may not match the backup.

## Operational Cost

This is a storage-heavy offline migration. During preparation, PostgreSQL stores the old vector, the new shadow vector, worklist indexes, shadow vector indexes, and semantic-link staging rows at the same time. Operators should reserve substantially more disk and WAL capacity than the current embedding table and index footprint.

Cutover is atomic at the database level, but it is not free: it deletes old semantic links, inserts staged semantic links, drops/renames indexes and columns, clears staging rows, and marks the migration complete in one transaction. The service should remain stopped for the target schema while this transaction runs. Larger schemas will hold locks and generate WAL for longer.

Dropping the old `embedding` column does not immediately return heap space to the operating system. PostgreSQL can reuse that space for future writes after vacuuming, but shrinking the table file usually requires an explicit table rewrite strategy such as `VACUUM FULL`, `pg_repack`, or dump/restore, depending on the deployment's maintenance policy.

## Schema and Extension Fixes Included

This PR also aligns vector-extension reconciliation with the current embedding tables:

- `ensure_vector_extension()` now checks `memory_units` and `mental_models`;
- it no longer treats legacy `learnings` / `pinned_reflections` as the active embedding tables for this reconciliation path.

## Scope and Limitations

This PR intentionally does not add:

- an online HTTP reembed endpoint;
- background async-operation scheduling for reembed;
- online dual-write;
- online recall quiescing;
- bank-scoped dimension-changing migrations;
- cross-database orchestration;
- user-facing retry backoff tuning;
- Oracle admin support.

The admin CLI remains PostgreSQL-only.

Retry behavior is intentionally simple in this PR: each embedding batch is retried by the admin command with bounded exponential backoff. This PR does not introduce a full typed embedding error hierarchy or user-facing retry backoff parameters.

## Validation

Automated checks run locally:

```bash
uv run pytest hindsight-api-slim/tests/test_admin_reembed.py -q
uv run pytest hindsight-api-slim/tests/test_vector_index.py -q
uv run ruff format --check \
  hindsight-api-slim/hindsight_api/_vector_index.py \
  hindsight-api-slim/hindsight_api/admin/reembed.py \
  hindsight-api-slim/hindsight_api/admin/cli.py \
  hindsight-api-slim/hindsight_api/migrations.py \
  hindsight-api-slim/tests/test_admin_reembed.py \
  hindsight-api-slim/tests/test_vector_index.py
uv run ruff check \
  hindsight-api-slim/hindsight_api/_vector_index.py \
  hindsight-api-slim/hindsight_api/admin/reembed.py \
  hindsight-api-slim/hindsight_api/admin/cli.py \
  hindsight-api-slim/hindsight_api/migrations.py \
  hindsight-api-slim/tests/test_admin_reembed.py \
  hindsight-api-slim/tests/test_vector_index.py
```

The pre-commit hook also ran successfully during commit:

```text
generate-docs-skill.sh
lint.sh
All lints passed
```

Manual Docker validation with a locally built full image:

1. PostgreSQL runtime using pg_search for text search and pgvector for vector search.
2. Seeded existing 1024-dimensional embeddings with an initial embedding configuration.
3. Reembedded to a new 1024-dimensional OpenAI-compatible embedding configuration:

   ```text
   status=completed
   memory_units=3/3
   mental_models=1/1
   dimension=1024
   staging_rows=0
   no temporary embedding_reembed shadow columns remain
   API restarted healthy with the new embedding configuration
   ```

4. Repeated the migration from 1024 dimensions to a different 384-dimensional embedding configuration:

   ```text
   status=completed
   memory_units.embedding=vector(384)
   mental_models.embedding=vector(384)
   dimension=384
   staging_rows=0
   no temporary embedding_reembed shadow columns remain
   API restarted healthy with the new embedding configuration
   ```

Manual pip/uv and hindsight-embed validation:

1. Created an isolated `hindsight-embed` named profile under a temporary `HOME`.
2. Started the profile daemon, which used an embedded pg0 database and automatically ran Alembic through `c3efa30f38c9`.
3. Seeded the embedded database with 2 `memory_units` and 1 `mental_models` row with existing vectors.
4. Stopped the `hindsight-embed` daemon.
5. Ran `hindsight-admin reembed --schema public --yes` against the same `pg0://hindsight-embed-reembed-smoke` database and embedding configuration.
6. Verified:

   ```text
   status=completed
   memory_units=2/2
   mental_models=1/1
   semantic_links_staged=2
   staging_rows=0
   shadow_columns=0
   memory_units.embedding=vector(384)
   mental_models.embedding=vector(384)
   ```

Unit/integration coverage added in this PR includes:

- full `reembed_schema` cutover with semantic-link staging;
- dimension-changing reembed;
- retain/reembed embedding-input parity for retained facts with dates and entities;
- consolidation observation reembed keeps bare observation text;
- early rejection of pgvector HNSW dimensions above 2000 before shadow state is created;
- refusal to abandon completed reembed migrations;
- worklist index cleanup after cutover;
- semantic-link staging grouped by bank/fact type with backend ANN tuning;
- ScaNN deferred shadow-index behavior when one indexed table is too small;
- Azure `pg_diskann` resolution for a configured `pgvectorscale` backend;
- vector-index helper coverage for shadow columns.

Existing admin backup/restore coverage also exercises the `BACKUP_TABLES` schema coverage rule, so adding the reembed admin tables to that list is guarded by the existing introspection test.

### DIFF
--- a/hindsight-api-slim/hindsight_api/_vector_index.py
+++ b/hindsight-api-slim/hindsight_api/_vector_index.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 import logging
 import os
+import re
+from collections.abc import Collection
 
 from sqlalchemy import text
 from sqlalchemy.engine import Connection
@@ -22,6 +24,7 @@ RESOLVED_EXTENSIONS = (*CONFIGURABLE_EXTENSIONS, "pg_diskann")
 VALID_EXTENSIONS = CONFIGURABLE_EXTENSIONS
 
 SCANN_MIN_ROWS_FOR_AUTO_INDEX = 10_000
+PGVECTOR_HNSW_MAX_DIMENSIONS = 2_000
 
 
 _EXTENSION_NAMES = {
@@ -31,12 +34,12 @@ _EXTENSION_NAMES = {
     "scann": "alloydb_scann",
 }
 
-_INDEX_USING_CLAUSES = {
-    "pgvector": "USING hnsw (embedding vector_cosine_ops)",
-    "pgvectorscale": "USING diskann (embedding vector_cosine_ops) WITH (num_neighbors = 50)",
-    "pg_diskann": "USING diskann (embedding vector_cosine_ops) WITH (max_neighbors = 50)",
-    "vchord": "USING vchordrq (embedding vector_cosine_ops)",
-    "scann": "USING scann (embedding cosine) WITH (mode = 'AUTO')",
+_INDEX_USING_TEMPLATES = {
+    "pgvector": "USING hnsw ({column} vector_cosine_ops)",
+    "pgvectorscale": "USING diskann ({column} vector_cosine_ops) WITH (num_neighbors = 50)",
+    "pg_diskann": "USING diskann ({column} vector_cosine_ops) WITH (max_neighbors = 50)",
+    "vchord": "USING vchordrq ({column} vector_cosine_ops)",
+    "scann": "USING scann ({column} cosine) WITH (mode = 'AUTO')",
 }
 
 _INDEX_TYPE_KEYWORDS = {
@@ -130,9 +133,11 @@ def pg_extension_name(ext: str) -> str:
     return _EXTENSION_NAMES[validate_extension(ext)]
 
 
-def index_using_clause(ext: str) -> str:
+def index_using_clause(ext: str, *, column: str = "embedding") -> str:
     """Return the CREATE INDEX USING clause for the vector backend."""
-    return _INDEX_USING_CLAUSES[_normalize_resolved(ext)]
+    if not re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*", column):
+        raise ValueError(f"Invalid vector index column name: {column!r}")
+    return _INDEX_USING_TEMPLATES[_normalize_resolved(ext)].format(column=column)
 
 
 def index_type_keyword(ext: str) -> str:
@@ -149,6 +154,18 @@ def should_defer_index_creation(ext: str, row_count: int) -> bool:
     """Return True when index creation should wait for more embeddings."""
     minimum_rows = minimum_rows_for_index(ext)
     return minimum_rows > 0 and row_count < minimum_rows
+
+
+def validate_vector_index_dimension(ext: str, dimension: int, *, table_name: str | None = None) -> None:
+    """Raise if the backend cannot index vectors of the requested dimension."""
+    if _normalize_resolved(ext) == "pgvector" and dimension > PGVECTOR_HNSW_MAX_DIMENSIONS:
+        location = f" on {table_name}" if table_name else ""
+        raise RuntimeError(
+            f"Embedding dimension {dimension}{location} exceeds pgvector HNSW index limit of "
+            f"{PGVECTOR_HNSW_MAX_DIMENSIONS}. Use an embedding model with <= "
+            f"{PGVECTOR_HNSW_MAX_DIMENSIONS} dimensions, or switch to a vector extension that supports higher "
+            "dimensions (e.g., pgvectorscale/DiskANN or AlloyDB ScaNN)."
+        )
 
 
 def ann_search_tuning_settings(ext: str, *, kind: str) -> tuple[tuple[str, str], ...]:
@@ -182,28 +199,21 @@ def bootstrap_extension(conn: Connection, ext: str) -> None:
         conn.execute(text(statement))
 
 
-def detect_vector_extension(conn: Connection, vector_extension: str = "pgvector") -> str:
-    """Validate the configured vector extension exists and return the index backend."""
+def resolve_vector_extension_from_installed(vector_extension: str, installed_extensions: Collection[str]) -> str:
+    """Return the resolved backend from the configured value and installed PG extensions."""
     configured_ext = validate_extension(vector_extension)
+    installed = set(installed_extensions)
 
     if configured_ext == "pgvectorscale":
-        pgvector_check = conn.execute(text("SELECT 1 FROM pg_extension WHERE extname = 'vector'")).scalar()
-        if not pgvector_check:
+        if "vector" not in installed:
             raise RuntimeError(
                 "DiskANN (pgvectorscale/pg_diskann) requires pgvector to be installed. "
                 f"Install it with: {_INSTALL_HINTS['pgvectorscale']}"
             )
-
-        vectorscale_check = conn.execute(text("SELECT 1 FROM pg_extension WHERE extname = 'vectorscale'")).scalar()
-        pg_diskann_check = conn.execute(text("SELECT 1 FROM pg_extension WHERE extname = 'pg_diskann'")).scalar()
-
-        if vectorscale_check:
-            logger.debug("Using vector extension: pgvectorscale (DiskANN)")
+        if "vectorscale" in installed:
             return "pgvectorscale"
-        if pg_diskann_check:
-            logger.debug("Using vector extension: pg_diskann (Azure DiskANN)")
+        if "pg_diskann" in installed:
             return "pg_diskann"
-
         raise RuntimeError(
             "Configured vector extension 'pgvectorscale' not found. Install either:\n"
             "  - pgvectorscale (open source): CREATE EXTENSION vectorscale CASCADE;\n"
@@ -211,15 +221,30 @@ def detect_vector_extension(conn: Connection, vector_extension: str = "pgvector"
         )
 
     extension_name = pg_extension_name(configured_ext)
-    extension_check = conn.execute(
-        text("SELECT 1 FROM pg_extension WHERE extname = :extension_name"),
-        {"extension_name": extension_name},
-    ).scalar()
-    if not extension_check:
+    if extension_name not in installed:
         raise RuntimeError(
             f"Configured vector extension '{configured_ext}' not found. "
             f"Install it with: {_INSTALL_HINTS[configured_ext]}"
         )
-
-    logger.debug("Using configured vector extension: %s", configured_ext)
     return configured_ext
+
+
+def detect_vector_extension(conn: Connection, vector_extension: str = "pgvector") -> str:
+    """Validate the configured vector extension exists and return the index backend."""
+    configured_ext = validate_extension(vector_extension)
+
+    if configured_ext == "pgvectorscale":
+        extension_names = ("vector", "vectorscale", "pg_diskann")
+    else:
+        extension_names = (pg_extension_name(configured_ext),)
+    installed_extensions = {
+        name
+        for name in extension_names
+        if conn.execute(
+            text("SELECT 1 FROM pg_extension WHERE extname = :extension_name"),
+            {"extension_name": name},
+        ).scalar()
+    }
+    resolved = resolve_vector_extension_from_installed(configured_ext, installed_extensions)
+    logger.debug("Using vector extension: %s", resolved)
+    return resolved
--- a/hindsight-api-slim/hindsight_api/admin/cli.py
+++ b/hindsight-api-slim/hindsight_api/admin/cli.py
@@ -9,6 +9,8 @@ import io
 import json
 import logging
 import zipfile
+from collections.abc import Callable
+from dataclasses import asdict
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
@@ -22,6 +24,15 @@ from ..engine.schema import fq_table_explicit as _fq_table
 from ..engine.transfer import export_bank
 from ..extensions import TenantExtension, load_extension
 from ..pg0 import parse_pg0_url, resolve_database_url
+from .reembed import (
+    ReembedOptions,
+    ReembedProgress,
+    abandon_reembed,
+    discover_hindsight_schemas,
+    ensure_no_active_reembed_state,
+    reembed_schema,
+    reembed_status,
+)
 
 # Setup logging
 logging.basicConfig(
@@ -62,6 +73,8 @@ BACKUP_TABLES = [
     "audit_log",
     "llm_requests",
     "graph_maintenance_queue",
+    "embedding_reembed_migrations",
+    "embedding_reembed_semantic_links",
 ]
 
 MANIFEST_VERSION = "1"
@@ -88,6 +101,8 @@ async def _backup(database_url: str, output_path: Path, schema: str = "public") 
     """Backup all tables to a zip file using binary COPY protocol."""
     conn = await asyncpg.connect(database_url)
     try:
+        await ensure_no_active_reembed_state(conn, schema, "backup")
+
         tables: dict[str, Any] = {}
         manifest: dict[str, Any] = {
             "version": MANIFEST_VERSION,
@@ -135,6 +150,8 @@ async def _restore(database_url: str, input_path: Path, schema: str = "public") 
     """Restore all tables from a zip file using binary COPY protocol."""
     conn = await asyncpg.connect(database_url)
     try:
+        await ensure_no_active_reembed_state(conn, schema, "restore")
+
         with zipfile.ZipFile(input_path, "r") as zf:
             # Read and validate manifest
             manifest: dict[str, Any] = json.loads(zf.read("manifest.json"))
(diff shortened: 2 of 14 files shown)


MattJackson:
This is exactly what #1884 was asking for — thank you for turning it around so fast. I read through `reembed.py`, the `admin-cli.md` you added, and the cutover/abandon paths, and the design hits every point that mattered to us:

- **No LLM re-extraction.** Confirmed in code: `_memory_unit_embedding_text` reads the stored `text` and runs it through `embedding_processing.build_fact_embedding_text` (retained facts get stored text + date/entity context, observations keep bare text, mental models use `name + " " + content`) — the same runtime text builders as new writes, no retain/extraction. The docs call this out explicitly. That's the whole ballgame for us.
- **Explicit, resumable admin command** (not auto-on-startup) with `--dry-run`, plus per-phase visibility (`embedding_state` / `shadow_indexes_state` / `semantic_links_state`) in status. The worklist partial index (`WHERE embedding IS NOT NULL AND embedding_reembed IS NULL`) means an interrupted run resumes on only the un-done rows.
- **Atomic cutover with the heavy work outside the lock.** Shadow HNSW/ANN indexes are built during prepare; `_cutover` is just drop-old-index + `RENAME` shadow→live + drop old column in one transaction, so the long pole (index build on the new dimension) never holds the cutover lock. Shadow columns keep the old vectors intact until the flip.
- **Clean abandon path.** `reembed-abandon` (and `--orphan-shadow-state`) drops the shadow indexes, `DROP COLUMN embedding_reembed` on both tables, and clears the semantic-links staging — so a half-built `vector(N)` shadow column can't linger on the live table. The advisory-xact-lock around it is a nice touch.

The offline model (stop API/workers, run a one-off in the same env, restart) is a reasonable safety call, and the Compose flow in the docs is clear.

**Offer to validate against a real migration:** our exact target is the headline case from #1884 — a single Postgres schema (no multi-tenant), ~88k `memory_units` + `mental_models`, moving from a 384-dim local encoder to a 1024-dim model, all rows untagged. We'd run it against a restored `admin backup` copy first (per your stop-traffic warning) and can report back: wall-clock + batch throughput, peak memory, the shadow ANN build time at ~88k×1024 specifically, cutover duration, and a recall sanity check on the new vectors. We're mid-consolidation-drain right now so we can't stop traffic yet, but this is queued for the moment it finishes — happy to feed those numbers back if useful.

Sanderhoff-alt:
@MattJackson 
Thanks for the detailed read-through and for offering to validate this on the real #1884 workload.

Those measurements would be very useful, especially the 88k x 1024 shadow index build time, cutover duration, and any recall sanity-check differences after migration. The restored `admin backup` copy is exactly the kind of validation path this was designed for.

Please make sure to test against the latest version of this PR branch / latest commit, since a few review fixes landed after the initial implementation. Please report back if the dry run, resume, or abandon paths expose anything surprising in that environment.

MattJackson:
Ran this in production today to migrate a live deployment from bge-small (384) to bge-m3 (1024) in place, ~112k memory units in a single `public` schema, pgvector/HNSW. Wanted to share a clean end-to-end report plus one rough edge.

**What worked**
- Cherry-picked the reembed commit onto a running **v0.7.2** with zero conflicts. The admin-tables migration's `down_revision` lined up exactly with our head, so the alembic chain was clean with no edits.
- `--dry-run` correctly reported planned work (111,779 units, target dim 1024) once the admin tables existed.
- The shadow-column design meant the live 384-dim column was untouched throughout — the deployment kept serving on the old embeddings until the atomic cutover. Exactly the property you'd want.
- **Resumability is excellent.** My first run failed at the index-build phase (details below). Re-running picked up with `embedding_state=complete` and went straight to rebuilding indexes — no re-embedding of the 111k rows. That saved the migration.
- Atomic cutover: 843,535 semantic links rebuilt, per-bank HNSW indexes renamed in, RENAME swap clean. Post-cutover: column is `vector(1024)`, zero NULL-embedding rows, recall (query embed → ANN → rerank) returns relevant hits immediately.

**The one rough edge: parallel index build vs. a small `/dev/shm`**
First cutover attempt failed here:
```
could not resize shared memory segment to 533794976 bytes: No space left on device
```
Root cause: our Postgres runs with `max_parallel_maintenance_workers=4`, and the parallel HNSW build wants a ~533MB dynamic shared-memory segment, but the container's `/dev/shm` was the Docker default 64MB. The embedding phase had fully completed, so all the expensive work was already done — it just couldn't build the index.

I worked around it by patching `_create_shadow_indexes` to issue `SET max_parallel_maintenance_workers = 0` before the `CREATE INDEX` statements (a session-level SET overrides the server's command-line flag). The serial build uses `maintenance_work_mem` instead of a DSM segment and completed fine, just slower.

Two suggestions, either would help the next person:
1. Set `max_parallel_maintenance_workers = 0` for the index-build session by default (or expose it as a flag). The build is offline, so the parallel speedup is rarely worth a hard failure on constrained `/dev/shm` — which is the norm for containerized Postgres.
2. Failing that, a note in the docs that the index build needs either a larger `/dev/shm` (`shm_size`) or serial maintenance workers would have saved me a cycle.

**Minor**
Batch-size 100 (default) was the sweet spot for us; bumping to 512 went super-linear and ~4x slower on a small-VRAM local embedder, because real stored texts are long. Default is well chosen.

Overall this did exactly what it promised — in-place dimension change, no re-ingest, no re-extraction, no data loss. Thanks for building it.

nicoloboschi (changes requested):
thanks for your work. for this large work, I'd suggest we first chat about the design in the github issue so nobody loses time.


we just added import and export bank so you can export all your banks, create a second instance of Hindsight (or just clear the existing one) and the reimport all them with the new configuration. While I understand it's "more steps" for the user, it also reduces complexity of maintaining this complex re-embed feature forever (which probably a user would ever run 0 times or once in its lifetime) 

I'm open to discuss it further

Sanderhoff-alt:
> thanks for your work. for this large work, I'd suggest we first chat about the design in the github issue so nobody loses time.
> 
> we just added import and export bank so you can export all your banks, create a second instance of Hindsight (or just clear the existing one) and the reimport all them with the new configuration. While I understand it's "more steps" for the user, it also reduces complexity of maintaining this complex re-embed feature forever (which probably a user would ever run 0 times or once in its lifetime)
> 
> I'm open to discuss it further

@nicoloboschi  
Thanks, that makes sense. I agree this is large enough that the design should be discussed in the issue first.

  One point I want to clarify is the target use case. I do not see this only as a one-time migration for an individual user. For teams or organizations running Hindsight as a shared memory service, embedding configuration can be an operational concern:

  - the default/local embedding model may not work well for some languages or domains
  - operators may want to move to a newer embedding model as quality improves
  - hosted deployments may need to standardize on a different provider or dimension over time
  - asking operators to export all banks, create/clear an instance, and re-import everything can be disruptive compared with an explicit offline admin migration

  So the value I was aiming for is a controlled maintenance command: stop traffic, run an explicit admin operation, preserve the existing database, and cut over once the new vectors/indexes are ready.

  That said, I understand the maintenance-cost concern. With the new `export-bank` / `import-bank` path, maybe the right next step is to compare the two options in #1884:

  - recommend export/import as the default migration path
  - only keep an in-place reembed command if there are deployment/operator cases that export/import does not cover well

  I can move this discussion back to the issue and pause the PR until we agree on the preferred direction

Sanderhoff-alt:
@MattJackson 

Thanks for the production report. This is very useful.

I agree that unconditionally forcing `max_parallel_maintenance_workers = 0` would be too blunt, because deployments that already provision enough `/dev/shm` should still be able to use PostgreSQL's configured
  parallel index build.

I updated the PR to make this an explicit reembed option instead:

```bash
  hindsight-admin reembed --schema public --index-max-parallel-maintenance-workers 0 --yes
```

Behavior:
  - by default, reembed does not override PostgreSQL's max_parallel_maintenance_workers;
  - if the option is provided, it applies only to the current admin session while shadow vector indexes are built;
  - after shadow index creation, the session setting is reset;
  - passing 0 gives containerized deployments with small /dev/shm a serial-build escape hatch without changing the database global config.

I also added docs that call out the Docker /dev/shm issue and mention this flag as the low-risk workaround. That should preserve the faster default for well-provisioned deployments while making the constrained-container failure mode easier to avoid.

PR #1867, opened 2026-05-29
fix(llamaindex): default to Cloud + replace dead manual test with gated E2E + requires_real_llm bucketing

DK09876:
## Summary
- **resolve_client cloud-default fix** — was raising \"No Hindsight API URL configured\" when \`configure()\` wasn't called; now falls back to \`DEFAULT_HINDSIGHT_API_URL\` and reads \`HINDSIGHT_API_KEY\` from env. Satisfies the \"default to Cloud\" framework goal.
- **Replace dead \`test_manual.py\`** — was entirely \`@pytest.mark.skip\`'d at the class level (line 28: \`@pytest.mark.skip(reason=\"Requires running Hindsight server\")\`), so it never ran, anywhere. Replaced with a real, gated \`tests/test_e2e.py\` covering tools (retain/recall/reflect via \`tool.call(...)\` against a live server) AND the \`HindsightMemory\` adapter (\`aget\`/\`put\` roundtrip).
- **Test bucketing** — registered \`requires_real_llm\` marker in \`pyproject.toml\`, plus added the missing \`asyncio_mode = \"auto\"\`. \`test-llamaindex-integration\` CI job now runs \`-m \"not requires_real_llm\"\` for the deterministic PR bucket.

## Test plan
- [x] \`uv run pytest tests -v -m \"not requires_real_llm\"\` — 84 passed, 4 deselected
- [x] \`HINDSIGHT_API_URL=http://127.0.0.1:8888 uv run pytest tests/test_e2e.py -v\` — 4 passed live
- [x] \`uvx ruff check hindsight_llamaindex/\` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1673, opened 2026-05-20
LangGraph: add memory_instructions, fix nodes, remove BaseStore

DK09876:
## Summary

- **Add `memory_instructions()`** — standalone system prompt helper that pre-fetches memories from Hindsight. Works with any LangChain model, no graph needed.
- **Fix node parameter gaps** — `create_recall_node()` now supports `recall_types` and `recall_include_entities`; `create_retain_node()` now supports `metadata` and `document_id`.
- **Fix silent error handling** — nodes now raise `HindsightError` instead of silently returning empty state on API failures.
- **Remove `HindsightStore` (BaseStore adapter)** — the KV abstraction over Hindsight's semantic memory was fundamentally leaky (`get` unreliable, `delete` no-op, `list_namespaces` session-scoped). Users should use tools, nodes, or `memory_instructions()` instead.
- **Cloud-first defaults** — default URL points to Hindsight Cloud; self-hosted users override via `hindsight_api_url`. README and docs examples updated accordingly.
- **Fix pre-existing test failures** — `user_agent` mock mismatch in client tests.

## Test plan

- [x] 52 unit tests pass
- [x] 13 E2E tests pass against live Hindsight (tools: 4, nodes: 6, memory_instructions: 3)
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1915, opened 2026-06-01
fix(opencode): default to Hindsight Cloud + gated live E2E

DK09876:
## Summary
- **Default to Hindsight Cloud** (Goal 5 from the integration-review framework). `DEFAULT_HINDSIGHT_API_URL = "https://api.hindsight.vectorize.io"` is now the baked-in `hindsightApiUrl` default, matching the convention adopted across LangGraph (#1673), Haystack, and the three sibling fix PRs #1866 / #1867 / #1868. Out-of-the-box: enable the plugin + set `HINDSIGHT_API_TOKEN` and you're talking to Cloud. Self-hosters override `hindsightApiUrl` (or `HINDSIGHT_API_URL` env).
- **Drop the no-URL "empty hooks" branch in `index.ts`** — the URL always resolves now, so the plugin always returns its full tool + hook surface. Missing key against Cloud surfaces a clear server-side error at call time (matches the framework's contract: "API key not required at construction; fails at call time").
- **Gated live E2E (`src/e2e.test.ts`)** — TypeScript equivalent of `requires_real_llm`. Skipped by default; runs only with `HINDSIGHT_LIVE_E2E=1` (exposed as `npm run test:e2e`). Covers the three contract surfaces: tool path (retain → server-side extraction → recall), `session.idle` auto-retain, `session.created` + `system.transform` inject.
- **Pre-existing dts build error** that was masked by the no-URL `return {}` branch is fixed by adding an index signature to `HindsightTools`.

## Verification
- Deterministic `npm test` — 6 files / 101 pass, 1 file / 3 skipped (the gated E2E).
- Live `HINDSIGHT_LIVE_E2E=1 npm test` against `127.0.0.1:8888` + a real `@vectorize-io/hindsight-client` — **7 files / 104 pass**.
- `npx tsc --noEmit` — clean.
- `npm run build` (tsup) — ESM `21.91 KB` + DTS `3.57 KB`, both succeed.

## Test plan
- [x] Deterministic vitest passes locally
- [x] Live vitest (`HINDSIGHT_LIVE_E2E=1`) passes locally against a self-hosted Hindsight
- [ ] CI `test-opencode-integration` job green
- [ ] (cookbook follow-up — none required; OpenCode is an interactive CLI plugin, not a notebook target — captured in `integrationReview/review-log.md`)


benfrank241:
verify-generated-files is failing but otherwise looks good

benfrank241:
test still failing

PR #1921, opened 2026-06-02
fix(openclaw): retain routing context for extraction

de1tydev:
## Problem
Hindsight's fact extraction could misinterpret OpenClaw routing metadata in retained transcripts. Values such as `[context]` sender/channel/provider, bank IDs, session keys, source systems, and tags are operational routing identifiers, but without explicit retain API guidance they can be treated as people, organizations, or project names.

## Root Cause
The OpenClaw plugin built and sent retain requests without passing interpretation guidance through the retain API `context` field across all retain paths.

## Solution
- Add `DEFAULT_RETAIN_CONTEXT` and wire it through live retain, queued retain flushes, and the backfill CLI.
- Expose configurable `retainContext` in the plugin manifest, config parsing, README, and types.
- Keep the default guidance platform-example-free so sample platform terms are not themselves extracted into memory.
- Trim custom `retainContext` values and test that the manifest default stays in sync with the code constant.
- Make the backfill symlink direct-execution test self-contained instead of depending on a prior `dist` build.

## Validation
- `npm test` in `hindsight-integrations/openclaw` (8 files, 248 tests)
- `npm run build` in `hindsight-integrations/openclaw`
- repository pre-commit hooks during commit


nicoloboschi:
Superseded by #1968, which was stacked on this branch and squash-merged into main (commit aa308ad20). All of this PR's commits (the `retainContext` / `DEFAULT_RETAIN_CONTEXT` guidance feature) are included in that squash, so the changes are already on main — `retainContext` is present in `types.ts` and `index.ts`. Closing as merged-via-#1968.

PR #1946, opened 2026-06-03
fix(control-plane): localize operations and graph legends

MapleEve:
## Summary

- localizes the Background Operations filter labels, table headers, status pills, actions, pagination, empty state, and details dialog fields
- maps known operation task types to translated display labels while preserving raw API values for unknown types
- localizes the Constellation canvas `causal` link legend and the size legend endpoints that were still rendering `few` / `many`
- adds matching catalog keys across all declared control-plane locales to keep message parity intact

## Validation

- `npm run i18n:check -- src/components/bank-operations-view.tsx src/components/constellation.tsx src/components/data-view.tsx`
- `npm run i18n:check`
- `npm test -- tests/messages/messages.test.ts`
- `npm test`
- `git diff --check`
- pre-commit hook: `generate-docs-skill.sh` + `lint.sh` passed during commit/amend

## Notes

The repo default branch is `main`; this branch was created from the latest `origin/main` rather than the locally divergent `main` checkout.


nicoloboschi (approved):
LGTM

PR #1954, opened 2026-06-03
fix(api): harden structured output and vchord probes

grimmjoww:
## Summary

- Request strict structured outputs for retain fact extraction, consolidation, and reflection structured calls.
- Normalize MiniMax-style bare fact-list responses into the expected `{ "facts": [...] }` envelope and fail clearly when retry output is still mostly malformed.
- Stop applying a blanket `vchordrq.probes` GUC for VectorChord by default; VectorChord probe values must match each index's `build.internal.lists` hierarchy, and VectorChord 1.1 supports index storage fallback parameters for that case.

## Why

The structured-output call sites already have provider support for strict schemas, but retain/consolidation/reflect were not consistently passing the flag through. Some providers can also return a bare JSON array for fact extraction, which is schema-compatible in spirit but not in the envelope shape Hindsight expects.

For VectorChord, #1668 correctly moved vchord indexes to `vector_cosine_ops` and added backend-specific ANN tuning. The follow-up compatibility issue is that `vchordrq.probes` is not a universal scalar knob: its list shape must match the index's `build.internal.lists`. Hindsight's built-in vchord index clause does not set `lists`, so a non-empty session-level probe override can make otherwise valid listless indexes fail. VectorChord 1.1's fallback-parameter model is the safer place for deployments that do create partitioned vchord indexes.

## Verification

- `uv run --project hindsight-api-slim pytest hindsight-api-slim/tests/test_fact_extraction_retry.py hindsight-api-slim/tests/test_consolidation_retry_budget.py hindsight-api-slim/tests/test_reflect_agent.py::TestStructuredOutputExtraction hindsight-api-slim/tests/test_vector_index.py hindsight-api-slim/tests/test_link_utils.py::TestComputeSemanticLinksAnnPgBouncerSafety::test_uses_set_local_for_pgvector_ann_tuning hindsight-api-slim/tests/test_link_utils.py::TestComputeSemanticLinksAnnPgBouncerSafety::test_vchord_ann_does_not_set_fixed_probe_count -q` — 33 passed
- `uv run --project hindsight-api-slim ruff check ...` on changed files — passed
- Manual PostgreSQL 18 + VectorChord 1.1.1 smoke: no forced `vchordrq.probes` uses the vchord index scan; forcing an incorrectly-shaped probe value reproduces `need 0 probes, but 1 probes provided`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/_vector_index.py
+++ b/hindsight-api-slim/hindsight_api/_vector_index.py
@@ -54,23 +54,20 @@ _INDEX_TYPE_KEYWORDS = {
 #   pre-dispatcher code (internal benchmarks tuned around our embedding count
 #   and recall floor; see the link_utils / pool init call sites for the
 #   latency-vs-recall framing).
-# - vchord exposes vchordrq.probes (no default; see VectorChord issue #392)
-#   and vchordrq.epsilon (default 1.9). probes = 10 / 30 are starting
-#   defaults pending a workload-specific sweep — vchordrq's recall curve
-#   shape differs from HNSW's, so the pgvector numbers don't translate
-#   directly. Revisit with a per-cluster benchmark once we have production
-#   recall data; until then these are deliberately conservative on the
-#   high-recall path. We leave epsilon at its default; tightening it is a
-#   separate trade-off.
+# - vchord exposes vchordrq.probes, but its shape must match the index's
+#   build.internal.lists hierarchy. VectorChord 1.1 added per-index fallback
+#   parameters for this reason: a session GUC overrides every vchordrq index,
+#   and a single value can be invalid for listless or mixed-layout indexes.
+#   Hindsight's built-in vchord clause does not set lists, so the safe default
+#   is no session-level probe override; deployments that partition vchordrq
+#   indexes should attach probes to the index storage parameters instead.
 # - pgvectorscale / pg_diskann / scann do not expose an equivalent per-statement
 #   knob in the engine today, so the dispatcher returns no statements for them.
 _ANN_TUNING_LOW_LATENCY: dict[str, tuple[tuple[str, str], ...]] = {
     "pgvector": (("hnsw.ef_search", "60"),),
-    "vchord": (("vchordrq.probes", "10"),),
 }
 _ANN_TUNING_HIGH_RECALL: dict[str, tuple[tuple[str, str], ...]] = {
     "pgvector": (("hnsw.ef_search", "200"),),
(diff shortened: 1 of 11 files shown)


grimmjoww:
## Production context for this fix

This PR fixes a regression in v0.8.0 that we're hitting in production right now on `grimmjoww/hindsight-local` (a local Hindsight install serving a 19k-node bank).

**Symptom:** Every `retain` operation since upgrading to v0.8.0 fails with:
```
hindsight_api.worker.poller - WARN - Task {id} scheduled for retry at {time}:
  need 0 probes, but 1 probes provided
```

The LLM extraction completes successfully (we see `output_tokens: 1104, time: 230s`), but the post-extraction probe check rejects the completion. The retry succeeds at the LLM call level and fails at the same probe check. Result: the queue accumulates stuck retain ops, blocking consolidation from running, which means the v0.8.0 observation dedup never actually exercises.

**This PR's fix** ("Stop applying a blanket `vchordrq.probes` GUC for VectorChord by default") looks like exactly the right call — the probe shape has to match the index's `build.internal.lists`, and a session-level scalar probe override breaks listless indexes.

**Reproduction:** upgrade any vchord-backed install from v0.7.2 to v0.8.0, send a single retain, watch the probe warning in the worker log.

**Asking for review** — this is a real production regression, not a stylistic change. A 33-test verification run already exists in the PR body. @nicoloboschi could you take a look when you have a moment?

Marked ready for review. Thanks!

nicoloboschi:
Split the vchord regression fix out into its own PR: #2076.

That PR contains only the `vchordrq.probes` change (the actual #1667 break — listless vchord indexes expect 0 probes, so the session GUC fails retain with `need 0 probes, but 1 probes provided` and stalls the worker queue). It can land on its own without waiting on validation of the rest of this PR.

Suggest rebasing this PR to drop the vchord bits and keep it focused on the `strict_schema` hardening + bare-list normalization. One thing worth checking before that part merges: forcing `strict_schema=True` flips retain/consolidation/reflect from the soft `json_object` path to grammar-enforced `json_schema` for *every* strict-capable provider (today nothing sets it; default is off). The OpenAI-compatible provider passes `model_json_schema()` through verbatim (no `additionalProperties: false`, optionals not forced required), so real OpenAI/Azure could 400 where soft mode worked — and reflect swallows that into `(None, 0, 0)`, silently dropping structured output. Worth a real provider-matrix run on those three call sites, not just mock unit tests.

Minor: the description calls this a v0.8.0 regression, but the `vchordrq.probes` GUC was introduced by #1668, which shipped in 0.7.2.

PR #1956, opened 2026-06-03
feat(cline): Hindsight memory integration via lifecycle hooks

benfrank241:
## Summary

A first-class **Cline** integration that gives the [Cline](https://github.com/cline/cline) coding agent persistent long-term memory via Hindsight — **without MCP**. It uses Cline's [lifecycle hooks](https://docs.cline.bot/customization/hooks) (the same mechanism as our Codex/Claude Code integrations), so memory is deterministic and doesn't depend on the model choosing to call a tool.

Lives in `hindsight-integrations/cline/`.

> **Status: draft** — opened for review + hands-on testing. Not released yet.

## What it does

| Cline hook | Behavior |
| --- | --- |
| `TaskStart` | Recall context for the new task; seed the transcript with the task description. |
| `UserPromptSubmit` | Recall memories for the prompt and inject via `contextModification`; record the prompt. |
| `TaskComplete` | Retain the accumulated task transcript + summary to the `cline` bank. |
| `TaskCancel` | Retain the partial transcript of a cancelled task. |

Recalled memories are injected as a `<hindsight_memories>` block. Cline hands hooks **no transcript**, so prompts are accumulated per-task in `~/.hindsight/cline/state/` and retained at task end.

## Design notes

- **Reuses the Codex integration's agent-agnostic core** — HTTP client, config loading, bank derivation, file state, and the recall-query/formatting helpers — retargeted to Cline's hook contract.
- **Cline divergences handled:** output shape `{cancel, contextModification, errorMessage}` (not Codex's `additionalContext`); hook files named exactly the hook type, executable with a shebang, under `.clinerules/hooks/`; enabled via a Cline Settings feature flag; **macOS/Linux only**.
- **Lean v1:** external API or an already-running local server (no daemon auto-start), no PostToolUse tool-call capture. Both are noted as follow-ups.
- **Resilient:** any error degrades to a no-op (`{"cancel": false}`, exit 0) — a memory hiccup never blocks Cline.
- **Standards:** hook payload parsed into a typed `HookInput` dataclass (no raw dicts at the boundary); no tuple returns.

## Install

```bash
python hindsight-integrations/cline/install.py --api-url <url> --api-token <key>
# then: Cline → Settings → Features → Hooks
```

## Tests / wiring

- **34 tests** (mocked `urllib` + isolated HOME): recall injection, transcript accumulation + retain, bank derivation, content helpers, and the installer (executable bit + shebang). Plus a no-Cline hook smoke test.
- CI job `test-cline-integration` + detect-changes filter + aggregate gate; `VALID_INTEGRATIONS`; changelog generator; `integrations.json`; docs page + changelog page + icon.

## Out of scope (v1) / follow-ups

- PostToolUse tool-call capture for richer retain; local `hindsight-embed` daemon auto-start (both reuse Codex `lib/daemon.py`/`llm.py`).
- Hosted `get-cline` one-line installer.
- Windows (Cline hooks are macOS/Linux only).


DK09876 (commented):
Reviewed against `.claude/skills/code-review/SKILL.md`. Tests pass (34/34), `ruff check .` clean, integration completeness all green (tests dir, `integrations-cline` + `test-cline-integration` in `test.yml`, `"cline"` in `release-integration.sh`). One should-fix and a couple of type-hint notes below.

### Should fix

**Raw `dict` for the structured `config` parameter.** SKILL §5 says:

> NEVER use raw `dict` types for structured data — this applies to all code, including internal helpers and private functions. If the dict has known keys, it must be a dataclass or Pydantic model.

`config` has fully known keys — `DEFAULTS` in `hooks/lib/config.py:14-50` enumerates `autoRecall`, `recallBudget`, `recallMaxTokens`, `recallTypes`, `retainContext`, `bankId`, `bankIdPrefix`, `dynamicBankId`, `bankMission`, `retainMission`, etc. This is exactly the structured-data case the SKILL forbids.

Affected signatures (13+):
- `bank.py:20` `derive_bank_id(hook: HookInput, config: dict) -> str`
- `bank.py:60` `ensure_bank_mission(client, bank_id: str, config: dict, debug_fn=None) -> None`
- `cline_io.py:34` `parse_hook_input(raw: dict) -> HookInput`
- `cline_io.py:81` `resolve_api_url(config: dict) -> Optional[str]`
- `config.py:106` `_load_settings_file(path: str, config: dict) -> None`
- `config.py:118` `load_config() -> dict`
- `config.py:146` `debug_log(config: dict, *args)`
- `hooks_impl.py:33` `_recall_context(hook: HookInput, config: dict, query: str) -> str`
- `hooks_impl.py:63` `handle_user_prompt_submit(hook: HookInput, config: dict) -> str`
- `hooks_impl.py:75` `handle_task_start(hook: HookInput, config: dict) -> str`
- `hooks_impl.py:87` `handle_retain(hook: HookInput, config: dict, status: str) -> None`
- `client.py:52` `_headers(self) -> dict`
- `client.py:61` `_request(self, method: str, path: str, body: Optional[dict] = None, timeout: int = DEFAULT_TIMEOUT) -> dict`

Suggested shape: a `HindsightClineConfig` dataclass (or Pydantic model) with the same fields as `DEFAULTS`, constructed by `load_config()`, threaded through the hook entrypoints. Same for `parse_hook_input(raw)` → returns `HookInput` already; the raw-`dict` input is acceptable as a JSON-boundary parse, but `_load_settings_file` and `_headers`/`_request` should return typed models.

**Mitigating context** (not a blocker for this PR, but worth tracking): the existing Cursor integration (`hindsight-integrations/cursor/scripts/lib/config.py`) follows this exact pattern — same `DEFAULTS = {...}`, same `load_config() -> dict`, same `_cast_env`/`debug_log(config: dict)`. Cline is structurally near-identical to Cursor's hook architecture and inherits the pattern (the README even has a comment to that effect in `hooks_impl.py:60` — "Reused generic helpers (from the Codex integration)").

So strictly per SKILL §5 this is a **Should fix**, but if the project wants to enforce the rule going forward, both Cursor and Cline (and Codex, presumably) need the same refactor as a cross-cutting cleanup PR. Calling that out so the conversation can happen once across the three integrations rather than separately per PR.

### Note (type-hint completeness)

- `bank.py:60` `ensure_bank_mission(client, bank_id: str, config: dict, debug_fn=None)` — `client` is `HindsightClient` (already imported in the file), `debug_fn` should be `Callable[..., None] | None`.
- `config.py:94` `_cast_env(value: str, typ)` — both `typ` and the return type are unannotated. `typ: type` + `-> Any` or a more specific union would be clearer.
- `config.py:146` `def debug_log(config: dict, *args)` — no return annotation. Should be `-> None`.

### Nothing else flagged

Branch hygiene clean (single squashed feature commit + main merge). Comments are good — section dividers + the honest "reused helpers" pointer is appreciated. Lifecycle hook design (UserPromptSubmit / TaskStart / TaskComplete / TaskCancel) maps cleanly onto Cline's surface. Tests cover the realistic flows. Looks ready to merge once the dict-typing discussion is settled.

PR #1967, opened 2026-06-04
fix(consolidation): set output token budget

### LINKED ISSUE #1939: Consolidation fails on Amazon Bedrock due to incorrect token limit
### Bug Description

```
ERROR - hindsight_api.engine.consolidation.consolidator - [CONSOLIDATION] LLM batch call failed after 3 attempts for 8 memories [....], skipping batch. Last error: LiteLLM response was truncated due to token limit
```

Consolidation operations routinely fail with truncated responses because the consolator never passes `max_completion_tokens` to the LLM call. Retain operations correctly pass `retain_max_completion_tokens` (default 64000), but consolidation has no equivalent parameter. When using Bedrock with custom imported models, Bedrock defaults to a low implicit output token limit (4096) when no `max_tokens` is present in the request body. Consolidation batches of 8 memories (the default `CONSOLIDATION_LLM_BATCH_SIZE`) produce structured JSON responses that frequently exceed 4096 output tokens, causing the model to stop with `finish_reason == "length"`, which LiteLLM raises as `OutputTooLongError`. After 3 retries the batch is skipped, and no observations are created for those memories.

In 3 days on a production Hindsight instance, this produced **551 truncation errors** and **101 skipped batches** across multiple banks.

### Steps to Reproduce

1. Configure Hindsight with a Bedrock imported model:
   ```
   HINDSIGHT_API_LLM_PROVIDER=bedrock
   HINDSIGHT_API_LLM_MODEL=openai.gpt-oss-20b-1:0
   ```
2. Leave `CONSOLIDATION_LLM_BATCH_SIZE` at the default of 8
3. Retain multiple memories and allow auto-consolidation to run
4. Observe the logs:
   ```
   [CONSOLIDATION] LLM batch call failed (attempt 1/3) for 8 memories [...] :
     LiteLLM response was truncated due to token limit
   [CONSOLIDATION] LLM batch call failed after 3 attempts for 8 memories [...] :
     skipping batch. Last error: LiteLLM response was truncated due to token limit
   ```

### Expected Behavior

Consolidation should pass `max_completion_tokens` to the LLM call (configurable via an env var like `HINDSIGHT_API_CONSOLIDATION_MAX_COMPLETION_TOKENS`, with a reasonable default such as 8192 or 16384). The existing `_cap_max_completion_tokens()` and `_get_model_output_cap()` machinery in `LiteLLM` would then automatically cap this to the model's registered limit. 8-memory batches should complete successfully without truncation.

### Actual Behavior

No `max_completion_tokens` is sent to the LLM. The `_build_common_kwargs` method only includes `max_completion_tokens` when explicitly provided, and the consolidator never provides it. The model's default output cap (4096 on Bedrock) silently truncates responses. The batch-splitting fallback (which splits failed 8-memory batches into two 4-memory batches) is not a reliable mitigation -- 4-memory batches also frequently hit the limit.

### Version

hindsight-api-slim 0.7.1 

### LLM Provider

AWS Bedrock

xmh1011:
## Summary

- add optional `HINDSIGHT_API_CONSOLIDATION_MAX_COMPLETION_TOKENS`, defaulting to unset for backward compatibility
- omit `max_completion_tokens` from consolidation LLM batch calls unless the setting is configured
- allow operators on providers with low hidden output caps, notably Bedrock imported models, to opt into an explicit consolidation output budget
- document the new setting and regenerate the Hindsight docs skill reference

Closes #1939.

## Tests

- `uv run --project hindsight-api-slim pytest hindsight-api-slim/tests/test_consolidation_retry_budget.py hindsight-api-slim/tests/test_config_validation.py -q`
- `uv run --project hindsight-api-slim ruff check hindsight-api-slim/hindsight_api/config.py hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py hindsight-api-slim/tests/test_consolidation_retry_budget.py hindsight-api-slim/tests/test_config_validation.py`
- `./scripts/generate-docs-skill.sh`
- `./scripts/hooks/lint.sh`


nicoloboschi:
Thanks for this — the bug is real and the asymmetry (retain passes `max_completion_tokens`, consolidation doesn't) is a genuine gap. One suggestion on the **default value**.

Today consolidation passes `None`, so the `max_tokens` key is omitted entirely and each provider applies its own default — which for OpenAI/Gemini/Anthropic is effectively the model's full output capacity (often 16k–64k+), and only 4096 on Bedrock imported models (the bug).

With this PR, the value goes through `_cap_max_completion_tokens` (`litellm_llm.py:119`), which only ever **lowers** it to the model's registry cap, never raises it. So a default of **16384 is a silent narrowing** for any high-capacity model: an 8-memory batch that could previously emit up to the model's full output is now capped at 16k.

I'd default to **64000 to align with `retain_max_completion_tokens`** instead. It's strictly safer:
- For every LiteLLM-*known* model, `_cap_max_completion_tokens` clamps 64000 down to that model's real cap → **identical to today's behavior, zero regression**.
- Only Bedrock-imported models (no registry entry) receive the raw value — exactly the population this fix targets.

The only thing to confirm is that Bedrock *honors/clamps* an explicit `max_tokens` above the model's real limit rather than rejecting it — but the issue notes Bedrock only mis-defaults "when no max_tokens is present," which implies it honors the value when set. If so, 64000 makes the non-Bedrock path a true no-op while still fixing the truncation.

Everything else LGTM (config wiring, threading into the call, tests, docs).

xmh1011:
> Thanks for this — the bug is real and the asymmetry (retain passes `max_completion_tokens`, consolidation doesn't) is a genuine gap. One suggestion on the **default value**.
> 
> Today consolidation passes `None`, so the `max_tokens` key is omitted entirely and each provider applies its own default — which for OpenAI/Gemini/Anthropic is effectively the model's full output capacity (often 16k–64k+), and only 4096 on Bedrock imported models (the bug).
> 
> With this PR, the value goes through `_cap_max_completion_tokens` (`litellm_llm.py:119`), which only ever **lowers** it to the model's registry cap, never raises it. So a default of **16384 is a silent narrowing** for any high-capacity model: an 8-memory batch that could previously emit up to the model's full output is now capped at 16k.
> 
> I'd default to **64000 to align with `retain_max_completion_tokens`** instead. It's strictly safer:
> 
> * For every LiteLLM-_known_ model, `_cap_max_completion_tokens` clamps 64000 down to that model's real cap → **identical to today's behavior, zero regression**.
> * Only Bedrock-imported models (no registry entry) receive the raw value — exactly the population this fix targets.
> 
> The only thing to confirm is that Bedrock _honors/clamps_ an explicit `max_tokens` above the model's real limit rather than rejecting it — but the issue notes Bedrock only mis-defaults "when no max_tokens is present," which implies it honors the value when set. If so, 64000 makes the non-Bedrock path a true no-op while still fixing the truncation.
> 
> Everything else LGTM (config wiring, threading into the call, tests, docs).

@nicoloboschi fixed

nicoloboschi:
I pushed a small follow-up (commit \`88fbcf5\`) changing the **default from 64000 → unset** — sorry for the back-and-forth, but on reflection unset is the genuinely safe choice.

The issue with a 64k default: \`_cap_max_completion_tokens\` only clamps models that exist in LiteLLM's registry. A model served by OpenAI/Gemini but *not* in the registry would receive the raw 64000 and the provider could reject it if it exceeds the real output cap — so 64k is not a guaranteed no-op for everyone.

Leaving it **unset omits the \`max_completion_tokens\` key entirely**, so every provider keeps its current implicit budget — byte-identical to today's behaviour. Operators on providers with a low hidden cap (Bedrock imported models) set \`HINDSIGHT_API_CONSOLIDATION_MAX_COMPLETION_TOKENS\` to fix #1939.

Changes in the follow-up:
- \`config.py\`: default \`None\`, field typed \`int | None\`
- \`consolidator.py\`: passes \`max_completion_tokens\` only when configured (mirrors the \`max_retries\` conditional)
- tests: default asserts \`None\` + a regression test that the key is omitted when unset
- docs: default shown as \`unset\` with the Bedrock opt-in note

Verified ruff/ty/tests green locally. Merging once CI passes.

xmh1011:
@nicoloboschi thanks again for the follow-up and for landing the safer unset default.

I updated the PR description so it now matches the final behavior:
- `HINDSIGHT_API_CONSOLIDATION_MAX_COMPLETION_TOKENS` is optional and defaults to unset
- consolidation omits `max_completion_tokens` unless the setting is configured
- providers with low hidden output caps, notably Bedrock imported models, can opt in explicitly

I also re-ran the targeted local validation on the current PR head:
- `uv run --project hindsight-api-slim pytest hindsight-api-slim/tests/test_consolidation_retry_budget.py hindsight-api-slim/tests/test_config_validation.py -q` -> 47 passed
- `uv run --project hindsight-api-slim ruff check hindsight-api-slim/hindsight_api/config.py hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py hindsight-api-slim/tests/test_consolidation_retry_budget.py hindsight-api-slim/tests/test_config_validation.py` -> passed

CI is green and the PR is ready from my side.

PR #866, opened 2026-04-03
feat(cursor): add Hindsight memory plugin for Cursor

DK09876:
## Summary

Adds a Cursor plugin (`@vectorize-io/cursor-hindsight`) that gives Cursor's coding agent persistent long-term memory via Hindsight. The plugin loads from `~/.cursor/plugins/local/` and integrates with Cursor's native hook + MCP system; no Cursor settings UI changes are required.

## Architecture (post-refactor)

The plugin was originally structured around `beforeSubmitPrompt`. The current implementation is **`sessionStart` + MCP** — see commits `7c2db16d → ada5af99` for the migration. Two complementary mechanisms:

1. **Plugin hooks** (`hooks/hooks.json`):
   - `sessionStart` → `scripts/session_start.py` recalls relevant memories at the start of each agent turn.
   - `stop` → `scripts/retain.py` retains the conversation transcript after each task completes (every-N-turns, configurable).
2. **MCP server** — the plugin writes a `.cursor/mcp.json` pointing at Hindsight, exposing `recall` / `retain` / `reflect` as agent-callable tools (the on-demand path).

### sessionStart additionalContext workaround

Cursor's native `additionalContext` channel for sessionStart hooks is broken in 3.x — staff-acknowledged in [forum 158452](https://forum.cursor.com/t/sessionstart-hook-additional-context-is-never-injected-into-agents-initial-system-context/158452), still open in 3.6.31 (verified empirically with a marker-injection test on 2026-06-02). The plugin works around it by ALSO writing the recalled memories to `<workspace>/.cursor/rules/hindsight-session.mdc` with `alwaysApply: true`. Workspace rules files ARE reliably injected by Cursor's rules engine, so the agent sees the memories on prompt 1 of every new agent.

Forward-compat: the hook still emits `additionalContext` to stdout. When Cursor fixes the upstream bug, drop the rules-file write — no protocol change needed.

UX guarantee: empirical testing (sleep 10s in sessionStart → 20s response time) confirms Cursor blocks prompt submission until sessionStart returns. The rules-file fallback's only cost is the Hindsight recall latency (~500ms typical).

## Cloud-default + self-hosting

- `DEFAULT_HINDSIGHT_API_URL = "https://api.hindsight.vectorize.io"` — fresh install with no configuration reaches the hosted backend (matches the cross-integration Goal-5 convention).
[...]

PR #1975, opened 2026-06-04
feat(integrations): add Cursor CLI integration

Korayem:
## Why

Cursor CLI users have no first-party way to give their agent persistent long-term memory across sessions. They have to re-explain their stack, preferences, and past decisions every time they start a new composer conversation — and useful context is lost whenever the agent's window gets compacted.

This adds a Hindsight integration for Cursor CLI that follows the same shape as the existing coding-agent integrations: Python hook scripts that read JSON from stdin and write JSON to stdout, wire into Cursor's hook system, and keep a memory bank in sync automatically. After install, every new prompt arrives at the model with the most relevant memories from past sessions pre-injected, and conversations are stored back to Hindsight during the session and again at session end.

## How memory flows

| Cursor hook | What it does | Why this hook |
| --- | --- | --- |
| `sessionStart` | Verifies Hindsight is reachable and pre-warms the local `hindsight-embed` daemon in the background so the first recall isn't slow. | Fire-and-forget setup work that should never block the agent. |
| `beforeSubmitPrompt` | Queries Hindsight for memories relevant to the current prompt and injects them as `additional_context`. The model sees them but they don't appear in the transcript. | Synchronous, runs every prompt, the right place for recall. |
| `stop` | Reads the on-disk transcript, formats it, and POSTs to Hindsight's retain API with `async=true` every configured N turns. | Fire-and-forget per-turn retain; the Hindsight server processes it in the background. |
| `sessionEnd` | Forces a final retain with the full transcript, bypassing `retainEveryNTurns`. | Guarantees short sessions still land in memory before the conversation terminates. |

The same config schema as `codex`: `autoRecall`, `autoRetain`, `recallBudget`, `retainEveryNTurns`, `dynamicBankId`, `bankMission`, etc. Users with a codex setup can drop in `cursor-cli` with no new concepts to learn.

## Cursor-specific differences from `codex`

- Recall output is a flat `{continue, additional_context}` object, not Codex's `hookSpecificOutput` envelope — that's what Cursor's `beforeSubmitPrompt` spec requires.
- Project resolution for dynamic bank IDs prefers `CURSOR_PROJECT_DIR` (Cursor's env var) → `workspace_roots[0]` (the common hook input field) → `cwd`. No fallback to `os.getcwd()` — the codex fallback leaks the test runner's working directory into the bank name.
- The `session` granularity field is supported but not part of the default; we use `conversation_id` from Cursor's common input fields when present.
- `retainTags` defaults to `["{conversation_id}"]` (Cursor's term) instead of codex's `{session_id}`.

## Configuration

`~/.hindsight/cursor-cli.json` for personal overrides (stable across updates):

```json
{
  "hindsightApiUrl": "https://api.hindsight.vectorize.io",
  "hindsightApiToken": "your-api-key",
  "bankId": "my-cursor-memory"
}
```

`~/.cursor/hooks/cursor-cli/settings.json` for shipped defaults. Every option also has an env-var override (`HINDSIGHT_API_URL`, `HINDSIGHT_BANK_ID`, `HINDSIGHT_RECALL_BUDGET`, etc.).

## What's in the integration

```
hindsight-integrations/cursor-cli/
├── README.md, settings.json, hooks/hooks.json
├── scripts/
│   ├── recall.py        # beforeSubmitPrompt
│   ├── retain.py        # stop
│   ├── session_start.py # sessionStart
│   ├── session_end.py   # sessionEnd final retain
│   ├── install.sh, uninstall.sh
│   └── lib/             # config, client, bank, content, state, daemon, llm
└── tests/               # 70 tests, all passing
```

## Install (for users)

```bash
git clone https://github.com/vectorize-io/hindsight.git
cd hindsight/hindsight-integrations/cursor-cli
./scripts/install.sh
```

The installer copies the hooks to `~/.cursor/hooks/cursor-cli/`, merges the hook block into `~/.cursor/hooks.json` (preserving any existing entries), and seeds an empty `~/.hindsight/cursor-cli.json` for the user's token. Restart Cursor CLI and the integration is live.

## CI / release wiring

- New `test-cursor-cli-integration` job in `.github/workflows/test.yml`, gated on the new `integrations-cursor-cli` filter.
- `cursor-cli` added to `VALID_INTEGRATIONS` in `scripts/release-integration.sh` so `./scripts/release-integration.sh cursor-cli patch` works out of the box.
- New top-level `hindsight-integrations/README.md` indexing every integration, with Cursor CLI highlighted under "Coding agents & CLIs".

## Verification

```bash
cd hindsight-integrations/cursor-cli
uv run --no-project --with pytest python -m pytest tests/ -v
# 70 passed

uv run --no-project --with ruff ruff check scripts/ tests/
# All checks passed!

shellcheck scripts/install.sh scripts/uninstall.sh
# clean
```

No live Hindsight server or Cursor CLI required for the test suite — the hooks are tested end-to-end with a fake HTTP client, mocked stdin/stdout, and an isolated `HOME` directory.


Korayem:
## Follow-up fix pushed: `2bf7b28a`

Caught a bug while dogfooding the integration in this repo:

- **Symptom**: `sessionStart` banner showed `Bank: cursor-cli` even with `dynamicBankId: true` and `dynamicBankGranularity: ["project"]` configured. `recall.py` / `retain.py` / `pre_compact.py` were using the correct derived bank. Anyone (or any agent) reading the banner and calling `hindsight memory reflect cursor-cli` hit an empty bank.
- **Root cause**: `session_start.py` had `bank_id = config.get("bankId") or "cursor-cli"` instead of calling `derive_bank_id(hook_input, config)`.
- **Fix**: import `derive_bank_id` from `lib.bank`, call it with the parsed `hook_input`, and surface the resolved bank in the debug log. Mirrors the pattern in the other three hooks.

Three new tests cover the four acceptance criteria:
- dynamic bank → derived id appears in banner
- `dynamicBankId: false` + explicit `bankId` → static bank still works
- `HINDSIGHT_BANK_ID` env override → resolved via config loader
- regression: the existing five `TestSessionStartHook` tests still pass

Full suite: **75 passed**, ruff clean, shellcheck clean. PR is now ready to merge as-is; the fix and the original work are split into two atomic commits on `feature/cursor-cli-integration`.

Korayem:
Dropped the handover commit (bf41a4e0) via force-push after pushback on the 'scan and PR' approach. Branch now ends at the cursor-cli fix (2bf7b28a). On reflection, the 'mimic the others' framing was misleading — cursor-cli was the only integration whose sessionStart emits a user-facing bank name, so there was nothing else to be buggy in the same way.

Korayem:
## Follow-up alignment — `b0b54ae1`

Pushed a refactor that brings cursor-cli back to near-parity with the codex/claude-code references. **Net −296 lines, 68 tests passing.**

### Removed — invented user-facing surfaces

1. **`session_start.py` banner.** The "Hindsight memory integration is active for this session. Bank: \<id\>" `additional_context` block, which is where the bank-id display-mismatch bug lived, is gone. The references' `sessionStart` is fire-and-forget with no `additional_context`. The only "consumer" that saw the banner was the agent itself, which never asked for it.

2. **`pre_compact.py` and its `TestPreCompactHook` class, plus the `preCompact` entry in `hooks.json` and the README row.** `preCompact` is observational in Cursor's spec — it cannot influence the compaction itself. The actual mechanism that preserves memory through compaction is the `beforeSubmitPrompt` recall that fires after compaction finishes. The "Hindsight preserved N memories" `user_message` was invented value with no reference equivalent.

### Restored — patterns from codex that were dropped

- `session_start.py`: `debug_log` for the "Hindsight not running" path.
- `recall.py`: `import time`, `import write_state`, `LAST_RECALL_STATE` const, and the `write_state(...)` block. (Dead code in codex too, but mirroring the reference keeps the diff focused on actual cursor-specific differences.)
- `recall.py`: kept the `user_prompt` fallback in the prompt field read.
- `retain.py`: restored the "Exit codes" docstring section and the inline comments / blank lines for readability.
- `lib/__init__.py`: removed the cursor-cli-specific docstring to match codex's empty file.

### Kept — true Cursor-specific differences (each one justified in the commit)

- Hook event names + debug log keys use `conversation_id` (Cursor's term) not `session_id` (codex's term).
- `session_id` fallback chain: `conversation_id` or `session_id` or "unknown".
- `template_vars` includes both `session_id` and `conversation_id` so retain tags work either way.
- `retainTags` default: `["{conversation_id}"]` (codex is `[]`).
- `retainContext` / `agentName` / `USER_AGENT` / `PROFILE_NAME` defaults rebadged for cursor-cli.
- `bankMission` / `retainMission` defaults are filled in (codex leaves them empty).
- Bank resolution prefers `CURSOR_PROJECT_DIR` → `workspace_roots[0]` → `cwd` (codex only uses `cwd`).
- `bank.py` `VALID_FIELDS` adds `gitProject` as an alias.
- `recall` output schema: Cursor's `beforeSubmitPrompt` wants `{continue, additional_context}`, not codex's `{hookSpecificOutput: {hookEventName, additionalContext}}`.

### The lesson I'm writing into the commit message

The "mimic the references" framing in the original plan was the right principle, but I only followed it for the file structure and the config schema. I invented new behavior on top — the banner and the `preCompact` user_message — and that's where the drift hid. After this commit, every remaining difference is justified by a real Cursor-vs-Codex spec difference, and the diff against `codex/` shows exactly which lines are Cursor's contribution.

PR is now three atomic commits: `ad982ee8` (integration), `2bf7b28a` (initial fix), `b0b54ae1` (alignment).

benfrank241 (changes requested):
Thanks for this well tested integration @Korayem. I appreciate you following the pattern for similar integrations! 

One small thing I'd like resolved before merging and a few docs fixes:

### Short sessions are never retained (no session-end flush)

`retainEveryNTurns` defaults to `10`, and the only retain path is the per-turn `stop` hook with no flush at session end. In a live test a fresh conversation logged `Turn 1/10, skipping retain (next at turn 10)` and persisted nothing. So any session under 10 turns (the common case) stores zero memory, which contradicts the PR's "every conversation is stored back at session end."

The `10` default itself is fine, it matches `codex` and `claude-code`. The difference is the flush. `claude-code` hit exactly this and solved it in `session_end.py`:

```python
# Force a final retain before stopping the daemon — guarantees short sessions
# (fewer turns than retainEveryNTurns) still land on disk.
if config.get("autoRetain") and hook_input.get("transcript_path"):
    from retain import run_retain
    run_retain(hook_input, force=True)
```


cursor-cli inherited `codex`'s weakness here rather than `claude-code`'s fix. Preferred resolution:

**Mirror `claude-code` with a `sessionEnd` hook.** Cursor [does expose a `sessionEnd` event](https://cursor.com/docs/hooks) — it fires once when the conversation terminates (`reason`: `completed` / `aborted` / `error` / `window_close` / `user_close`), distinct from `stop` which fires per turn. Its payload carries both `transcript_path` and `conversation_id`, so a `session_end.py` wired there can call retain with `force=True` to bypass the turn gate and guarantee short sessions still land on disk — exactly the `claude-code` pattern. (The PR currently wires only `sessionStart` / `beforeSubmitPrompt` / `stop`.)

### Docs describe a `preCompact` hook that isn't in the code

The "align with codex" refactor dropped `preCompact`/`pre_compact.py` and the sessionStart banner, but a few docs still describe them:

- **`hindsight-integrations/README.md`** advertises "compaction preservation on `preCompact`" — there's no such hook; please drop that clause.
- The **PR description** still documents the `preCompact` row and the "Hindsight is active" sessionStart banner. Worth trimming so reviewers/users aren't surprised.
- Minor: the description says **72 tests**; the suite is **68**.

### Smaller things (non-blocking)

- All type annotations from the `codex` source were dropped (`content.py`/`client.py` go from fully annotated to none) — the repo standard is type hints required, so restoring them would keep it consistent with the template it's based on.
- `.gitignore` has some unrelated churn: `blog-post*` → `blog-post` (narrows an existing glob, looks unintended) and `*.worktrees`.
- `daemonIdleTimeout` defaults to `0` in config but `daemon.py` falls back to `300` — the fallback is dead code since the key is always present; worth picking one.

Happy to re-review once the retain flush is sorted. Nice work overall. 🚀

Korayem:
Addressed the latest review in `3722f0f2`.

What changed:
- Added Cursor `sessionEnd` wiring and `session_end.py`.
- Extracted `retain.run_retain(hook_input, force=False)` and made `sessionEnd` call it with `force=True`, so short sessions bypass the per-turn `retainEveryNTurns` gate at final flush.
- Added regression coverage proving `stop` still skips below threshold while `sessionEnd` retains the same short transcript.
- Removed stale `preCompact` / sessionStart banner docs from the integration README and PR body.
- Restored the `.gitignore` `blog-post*` glob and made `.worktrees/` the ignored runtime directory.
- Aligned the daemon idle-timeout fallback with the shipped `daemonIdleTimeout: 0` default.

Verification:
- `uv run --no-project --with pytest python -m pytest tests/ -v` → 70 passed
- `uv run --no-project --with ruff ruff check scripts/ tests/` → clean
- `shellcheck scripts/install.sh scripts/uninstall.sh` → clean

I also ran the repo-wide `./scripts/hooks/lint.sh`; it failed in the unrelated control-plane eslint step because `@eslint/js` was not resolvable from `hindsight-control-plane/eslint.config.mjs` after npm tried to install `eslint@10.4.1`.

DK09876 (commented):
Reviewed against `.claude/skills/code-review/SKILL.md` using the same lens we applied to the existing `feat/cursor-integration` work. Tests pass (70/70), `ruff check .` clean, integration completeness all green (tests dir, `integrations-cursor-cli` + `test-cursor-cli-integration` in `test.yml`, `"cursor-cli"` in `release-integration.sh`). Two **must-fix** items below — the bigger one is a project-wide convention violation, the smaller is type-hint completeness on internal helpers. A handful of should-fix and notes after.

### Must fix

#### 1. Goal-5: integration does not default to Hindsight Cloud

The project's published 6-goal framework (`.claude/skills/code-review/SKILL.md` and `integrationReview/review-log.md`) requires:

> The default URL should point to Hindsight Cloud (`https://api.hindsight.vectorize.io`). API key is not required at construction time — it just fails at call time if not set. Users who self-host can override the URL to point to their instance.

This integration ships with:

- `scripts/lib/config.py:45` — `"hindsightApiUrl": None,`
- No `DEFAULT_HINDSIGHT_API_URL = "https://api.hindsight.vectorize.io"` constant
- `scripts/lib/daemon.py:get_api_url()` resolution: explicit URL → existing local server → opt-in daemon-start → **raises `RuntimeError("No Hindsight server on port 9077. Set hindsightApiUrl for external API, start hindsight-embed manually, or wait for the retain hook to auto-start the daemon.")`**

Every other integration in this batch satisfies the rule:

| Integration | Default URL constant |
|---|---|
| `cursor` (existing) | `scripts/lib/config.py` — `DEFAULT_HINDSIGHT_API_URL = "https://api.hindsight.vectorize.io"` |
| `opencode` | `src/config.ts` — same |
| `autogen` | `hindsight_autogen/config.py` — same |
| `litellm` | `hindsight_litellm/__init__.py` — same |
| `llamaindex` | `hindsight_llamaindex/config.py` — same |
| `superagent` | `hindsight_superagent/config.py` — same |
| `haystack` | `hindsight_haystack/config.py` — same |
| `openai-agents` | `hindsight_openai_agents/config.py` — same |
| `langgraph` | `hindsight_langgraph/config.py` — same |
| `claude-agent-sdk` | `hindsight_claude_agent_sdk/config.py` — same |
| **`cursor-cli` (this PR)** | **none** |

Compare with the existing `cursor` integration's `daemon.py:get_api_url()` (which solved this exact problem):

```python
# Resolution order:
#   1. hindsightApiUrl configured → use it (explicit override wins).
#   2. A local server is already healthy → use it.
#   3. useLocalDaemon: True (opt-in) AND allow_daemon_start → start the
#      embed daemon and use it.
#   4. Otherwise → fall back to DEFAULT_HINDSIGHT_API_URL (hosted backend).
#      API key is not required at this point — a missing key only fails
#      when an actual request is made.
```

For `cursor-cli` to satisfy Goal 5:
1. Add `DEFAULT_HINDSIGHT_API_URL = "https://api.hindsight.vectorize.io"` to `config.py`.
2. Change `"hindsightApiUrl"` default from `None` to `""` (or keep `None` and treat both as "unset").
3. Add `"useLocalDaemon": False` to DEFAULTS so self-hosters opt in.
4. Rewrite `get_api_url()` so the final fallback is `DEFAULT_HINDSIGHT_API_URL`, not a `RuntimeError`. Failures should happen at the actual API call (with a missing-key error), not at URL resolution.
5. README "Installation" section should show Cloud as the Quick Start path, with self-hosted under a "Self-hosting" heading (mirrors what `cursor`/`opencode`/`autogen` READMEs do post-V2 audit).

This is the same gap the V2 audit caught on the existing `cursor` PR (#866) — and now the `cursor-cli` PR has it. Worth fixing here to keep the convention airtight.

#### 2. Type hints largely missing on internal helpers

Per SKILL §5: *"Missing type hints on function parameters and return types"* is listed as a **must fix**.

These functions in `cursor-cli/scripts/lib/` have **no** type hints at all (parameters or return types):

- `config.py:100` `def _cast_env(value, typ):`
- `config.py:112` `def _load_settings_file(path, config):`
- `config.py:124` `def load_config():`
- `config.py:152` `def debug_log(config, *args):`
- `content.py:39` `def strip_memory_tags(content):`
- `content.py:57` `def read_transcript(transcript_path, include_tool_calls=False):`
- `content.py:84` `def _read_transcript_text(transcript_path):`
- `content.py:124` `def _read_transcript_rich(transcript_path):`
- `content.py:246` `def _extract_text_from_blocks(blocks):`
- `content.py:261` `def _coerce_result_text(result):`
- plus `daemon.py:get_api_url(config, debug_fn=None, allow_daemon_start=False)` — params untyped (existing `cursor` typed all three).

These are all internal helpers, but SKILL applies the rule to internal helpers explicitly. Minimum needed: param types (`dict`, `str`, `Path | None`, `list[dict[str, Any]]`, etc.) and return types (`-> None`, `-> list[dict]`, `-> str`). If we're following SKILL §5 strictly (Cline review went through the same gate at https://github.com/vectorize-io/hindsight/pull/1956), this needs to land before merge — or be raised as a cross-cutting "all three Cursor-derived integrations need typed signatures" task and addressed there.

### Should fix

#### 3. Raw `dict` for the structured `config` parameter

Same finding as the Cline review (#1956): `config` has known keys (the `DEFAULTS` dict in `config.py:14-90`), so SKILL §5 says it should be a Pydantic model or dataclass, not raw `dict`. Mitigating context: the existing `cursor` and (now `cline`) integrations follow the same pattern, so this is a cross-cutting refactor target rather than a per-PR blocker.

#### 4. `_read_transcript_text` may miss Cursor 3 transcript shape

The parser at `scripts/lib/content.py:84` handles two transcript shapes:
- Flat: `{role, content}` (with `content` optionally a list of typed blocks)
- Type-nested: `{type: "user", message: {role, content: [blocks]}}`

It does **not** handle a third shape we just fixed in `feat/cursor-integration` commit `df4267ed`:
```
{"role":"user","message":{"content":[{"type":"text","text":"..."}]}}
```
(top-level `role`, no top-level `type`, content nested under `message.content` as typed blocks)

Cursor IDE 3.6.31 writes its stop-hook transcript in this third shape — silently dropping every message under the original two-branch parser. That's how the existing `cursor` integration's auto-retain looked correct under unit tests but produced empty bank stores under real Cursor.

**Action:** verify that `cursor-cli` writes its hook transcript in one of the two formats this parser DOES handle. If it writes the role-nested-with-blocks shape (like Cursor IDE), this PR has the same silent-empty-transcript bug — the fix is the third branch we added to `cursor`'s `retain.py:read_transcript()` and the same regression test (`tests/test_hooks.py::TestRetainHook::test_read_transcript_parses_cursor3_role_nested_with_block_content`). Worth confirming with a live capture from `~/.cursor/projects/.../agent-transcripts/<conv>/<conv>.jsonl`.

### Note

#### 5. Unrelated commit on the branch

`63748bd7 Add .worktrees to .gitignore` — SKILL §1 (Branch Hygiene): *"Only include commits relevant to the PR/branch/feature — no unrelated changes."* This is a 2-line `.gitignore` tweak that has nothing to do with the cursor-cli integration; it should land in a separate "dev convenience" PR. Harmless if left, but flagging per the rule.

#### 6. Naming/positioning vs the existing `cursor` integration

Both `hindsight-integrations/cursor` and `hindsight-integrations/cursor-cli` now exist. The README should make the distinction visible (this targets the Cursor CLI binary at `cursor.com/docs/cli`; the existing `cursor` integration targets the IDE plugin). A line in `hindsight-integrations/README.md` clarifying which is which would save users from picking the wrong one.

#### 7. Architectural notes worth a comment

- `daemonIdleTimeout: 0` (config.py:50) differs from existing `cursor`'s `300`. If this is intentional (CLI hooks are sync + short-lived so the daemon doesn't need to stay around), a one-line comment would be helpful for future readers.
- `retainTags: ["{conversation_id}"]` (config.py:39) — template substitution is non-standard (other integrations use static tags). The substitution must be documented in README "Configuration" so users know what `{conversation_id}` resolves to.
- `recall.py`'s "Exit codes: 0 — always (graceful degradation on any error). Non-zero blocks the prompt, which is a dangerous default for a memory hook." — this is a great comment. Keep this style elsewhere.

### Architecture appreciation

One thing worth calling out: this integration uses Cursor CLI's `beforeSubmitPrompt` hook with `additional_context` to inject memories. That's the **correct** API path. The existing `cursor` integration had to ship a rules-file workaround because Cursor IDE's `sessionStart additionalContext` was broken in 3.6.x (see https://forum.cursor.com/t/sessionstart-hook-additional-context-is-never-injected-into-agents-initial-system-context/158452). Cursor CLI exposes the working surface directly, so this integration skips that workaround entirely. Nice architectural fit.

### Summary

Two must-fix items (Goal 5 default-to-Cloud + missing type hints), three should-fix (raw-dict, transcript parser may miss Cursor 3 shape, plus the cross-cutting refactor framing), and a few notes. Once Goal 5 lands and type hints fill in, this should slot cleanly into the same fold as the other 10 integrations.

benfrank241 (approved):

PR #1986, opened 2026-06-04
feat(llm): opt-in strict json_schema for capable openai-compatible backends

MattJackson:
## Problem

Structured-output calls (retain's fact extraction, consolidation's observation
merge) go through the **soft `json_object`** path on the OpenAI-compatible
provider: the JSON schema is appended to the prompt and `response_format` is set
to `{"type": "json_object"}`. This relies on the model *voluntarily* emitting
valid JSON.

Strong/hosted models comply. But weaker self-hosted instruction-followers behind
an OpenAI-compatible server — small Qwen / Llama / Mistral GGUF models served by
`llama.cpp`, vLLM, etc. — routinely return:

- a prose preamble (`Sure, here is the JSON you requested:`),
- a fenced block ` ```json … ``` `,
- or subtly invalid JSON.

`_strip_code_fences` handles some of these, but not all. The failures surface as
`json.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`, the async
operation retries indefinitely, and **retain / consolidation wedge** — the LLM
runs hot producing output that never parses, and the backlog never drains.

The provider already has a strict `json_schema` code path, but it only triggers
when a caller passes `strict_schema=True` (retain/consolidation don't), and its
docstring marks it *"OpenAI only"*. In practice every `json_schema`-capable
backend — OpenAI, **llama.cpp**, vLLM — will *grammar-enforce* the schema, making
the output valid by construction (no prose, no fences, no malformed JSON).

## Fix

Add an opt-in config flag **`HINDSIGHT_API_LLM_STRICT_SCHEMA`** (default `false`,
so **no behavior change** for existing deployments). When enabled, the
OpenAI-compatible provider uses strict `json_schema` for *all* structured calls,
letting operators on schema-capable self-hosted backends get reliable structured
output instead of the fragile soft path.

- `config.py`: new env `HINDSIGHT_API_LLM_STRICT_SCHEMA` → `Config.llm_strict_schema`
  (default `False`).
- `openai_compatible_llm.py`: `use_strict_schema = strict_schema or get_config().llm_strict_schema`;
  the strict `json_schema` branch is taken when either is set. The per-call
  `strict_schema=True` argument keeps working exactly as before.

## Verification

Reproduced and fixed in a live self-hosted deployment: **qwen2.5-coder-7B-instruct
(Q4_K_M) served by llama.cpp** behind an OpenAI-compatible endpoint.

- **Soft mode (default):** consolidation returned prose + fenced JSON, failed to
  parse, retried forever — backlog frozen, zero drain, GPU pinned at ~99%
  producing nothing.
- **`HINDSIGHT_API_LLM_STRICT_SCHEMA=true`:** the same model returns clean,
  grammar-enforced JSON every call; consolidation parses and drains normally.

A direct probe confirmed the model emits a valid `{"observations": [...]}`
object under `response_format: json_schema` both directly and through a gateway —
the only change needed was letting Hindsight ask for it.

## Tests

`tests/test_llm_strict_schema.py` (mirrors the existing
`test_openai_compatible_response_hardening.py` mock pattern):

- with `llm_strict_schema=True`, a structured call sends
  `response_format.type == "json_schema"` (and `strict: true`);
- with it `False` (default), the call keeps sending
  `response_format.type == "json_object"`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -448,6 +448,7 @@ ENV_LLAMACPP_GPU_LAYERS = "HINDSIGHT_API_LLAMACPP_GPU_LAYERS"
 ENV_LLAMACPP_CONTEXT_SIZE = "HINDSIGHT_API_LLAMACPP_CONTEXT_SIZE"
 ENV_LLAMACPP_CHAT_FORMAT = "HINDSIGHT_API_LLAMACPP_CHAT_FORMAT"
 ENV_LLAMACPP_NO_GRAMMAR = "HINDSIGHT_API_LLAMACPP_NO_GRAMMAR"
+ENV_LLM_STRICT_SCHEMA = "HINDSIGHT_API_LLM_STRICT_SCHEMA"
 ENV_LLAMACPP_EXTRA_ARGS = "HINDSIGHT_API_LLAMACPP_EXTRA_ARGS"
 
 # Optimization flags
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Opened #2003 as a broader version of this: instead of wiring the flag into only the OpenAI-compatible provider, it resolves `HINDSIGHT_API_LLM_STRICT_SCHEMA` once in `LLMProvider.call` and passes it down so **every** provider honors it (LiteLLM `strict: true`, Anthropic forced tool-use, Gemini already native), and it also covers the **batch retain path** (`_build_request_body`), which this PR's `.call()`-level change doesn't reach. Same opt-in flag, same default-off behavior.

nicoloboschi:
superseded by #2003

PR #1987, opened 2026-06-04
feat(metrics): expose async-operation queue + consolidation backlog as gauges

MattJackson:
## Problem

Hindsight's `BankGraphStats` endpoint already computes the operational numbers
that matter for an async-consolidation system — `operations_by_status`,
`pending_consolidation`, `failed_consolidation` — but only as a **point-in-time
HTTP response, per bank**. There's no way to trend them over time or alert on
them from Prometheus. The two questions an operator actually asks —

- *"Is the worker keeping up?"* (queue depth)
- *"Is the knowledge base caught up?"* (consolidation backlog)

— aren't answerable from `/metrics` today. In particular there's **no signal for
an operation stuck in `processing`**, which is exactly the failure mode that
silently pins a worker slot while the backlog sits flat.

## Fix

Three observable gauges, mirroring fields the stats endpoint already returns,
fed by a 30s background-refresh cache — the same pattern as the existing
`hindsight_db_pool_*` gauges, so the scrape path stays synchronous:

| Metric | Mirrors | Meaning |
|---|---|---|
| `hindsight_async_operations{operation_type,status}` | `operations_by_status` | Worker queue depth, non-terminal states. `pending`=backlog, `processing`=in-flight, `failed`=stranded. |
| `hindsight_consolidation_backlog` | `pending_consolidation` | Source memories (experience/world) not yet consolidated. |
| `hindsight_consolidation_failed` | `failed_consolidation` | Source memories whose consolidation permanently failed (recoverable via the recovery endpoint). |

Design notes:
- **Terminal states excluded.** The queue gauge counts only `pending`,
  `processing`, `failed` — `completed`/`cancelled` would grow unbounded and say
  nothing about current load.
- **Cardinality bounded.** Per-bank labels are gated behind the existing
  `metrics_include_bank_id` flag (off by default); when off, counts aggregate
  per tenant/schema — a handful of series total. No new config knob.
- **Index-backed.** SQL is lifted from the stats endpoint and rides
  `idx_async_operations_status` / `idx_memory_units_unconsolidated`.
- **PostgreSQL-specific** (`FILTER`, `information_schema`), consistent with this
  collector already being bound to an asyncpg pool.

## Tests

`tests/test_backlog_metrics.py` (mirrors `test_metrics.py`'s mock-meter/collector
pattern):
- `_refresh_backlog` aggregates the queue and consolidation counts into the
  caches with the expected `(schema, op_type, status, bank)` keys;
- the queue query filters to non-terminal statuses only (no `completed`/`cancelled`);
- the gauges register under the expected names and their callbacks emit cached
  values with correct attributes, omitting `bank_id` when the flag is off.

All 3 pass; the existing 26 `test_metrics.py` tests still pass.


hindsight-api-slim/hindsight_api/metrics.py
```diff
@@ -656,6 +670,165 @@ def get_pool_max_size(_options):
...
+                # bank-stats endpoint. The partial index
+                # idx_memory_units_unconsolidated backs the pending count.
+                try:
+                    rows = await conn.fetch(
+                        f"SELECT {bank_sel}"
+                        "COUNT(*) FILTER (WHERE consolidated_at IS NULL "
+                        "AND fact_type IN ('experience', 'world')) AS pending, "
+                        "COUNT(*) FILTER (WHERE consolidation_failed_at IS NOT NULL "
+                        "AND fact_type IN ('experience', 'world')) AS failed "
+                        f'FROM "{schema}".memory_units' + (" GROUP BY bank_id" if per_bank else "")
```
nicoloboschi: **Perf: this is a recurring full-table scan of `memory_units`, not an index-backed count.**

A single aggregate with two different `FILTER` predicates and no `WHERE` clause (the `per_bank=False` path) has to visit every row of the relation — Postgres can't combine the two partial indexes (`idx_memory_units_unconsolidated` + `idx_memory_units_consolidation_failed`) into one scan, so this resolves to a **seq scan of the whole `memory_units` table** (the largest table). The PR borrowed this SQL from the bank-stats endpoint, but there it's scoped `WHERE bank_id = $1` *and* runs only on-demand. Here it runs **every 30s, across every provisioned schema, in both the API and worker processes** (both call `set_db_pool`). On dense/large banks this is exactly the recurring heavy scan this codebase has had perf trouble with.

Suggested fix: split into two separate `COUNT(*)` queries, each with a `WHERE` matching a partial-index predicate so each becomes an index-only scan over just the unconsolidated/failed subset:

```sql
SELECT COUNT(*) FROM "<schema>".memory_units
WHERE consolidated_at IS NULL AND fact_type IN ('experience','world');
-- and the analogous consolidation_failed_at IS NOT NULL query
```

`GROUP BY bank_id` still composes with the partial indexes when `per_bank` is on. (The `async_operations` query above is fine — `status IN (...)` rides `idx_async_operations_status`.)

hindsight-api-slim/hindsight_api/metrics.py
```diff
@@ -335,6 +341,13 @@ def __init__(self):
         # DB pool metrics holder (set via set_db_pool)
         self._db_pool: "asyncpg.Pool | None" = None
 
+        # Backlog / queue-depth gauge caches, refreshed by a background task
+        # (see _setup_backlog_metrics) so the scrape path stays synchronous.
+        self._async_ops_counts: dict[tuple, int] = {}
```
nicoloboschi: Note: raw `dict[tuple, int]` with untyped `tuple` keys deviates from the project's no-raw-dict/tuple-for-structured-data standard. The composite keys `(schema, op_type, status, bank)` would be clearer and type-checked as a `NamedTuple`. Borderline since these are metric-label cache keys (idiomatic), but flagging given the repo's strict stance.

hindsight-api-slim/hindsight_api/metrics.py
```diff
@@ -656,6 +670,165 @@ def get_pool_max_size(_options):
...
+            callbacks=[get_consolidation_failed],
+            description="Source memories whose consolidation permanently failed "
+            "(recoverable via the consolidation recovery endpoint)",
+            unit="{memories}",
+        )
+
+        # Drive the caches from a background task on the running loop.
+        # set_db_pool runs during async startup, so a loop is normally present;
+        # if not, the gauges simply stay empty rather than crashing collection.
+        import asyncio
```
nicoloboschi: Note: `import asyncio` is done locally here (and again in `_backlog_refresh_loop`). `asyncio` isn't a top-level import in this module yet — prefer adding it at module scope rather than importing inside methods.

hindsight-api-slim/hindsight_api/metrics.py
```diff
@@ -656,6 +670,165 @@ def get_pool_max_size(_options):
...
+
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            logger.warning("No running event loop; backlog metrics disabled")
+            return
+        self._backlog_task = loop.create_task(self._backlog_refresh_loop())
```
nicoloboschi, line 827: Note: `_backlog_task` is spawned but never cancelled on shutdown — it's left to GC. Fine for a process-lifetime collector and there's no existing teardown hook to hang it on, but worth a comment, or a cancel if a shutdown path is ever added.

nicoloboschi (commented):
Reviewed the metrics changes — lint passes, the 3 new tests pass, the SQL matches the engine stats query, and the branch is clean off `main`. One perf concern worth addressing before merge, plus a few smaller notes (left inline).

hindsight-api-slim/tests/test_backlog_metrics.py
```diff
@@ -0,0 +1,140 @@
...
+
+    collector._db_pool = _FakePool(fetch)
+    await collector._refresh_backlog()
+
+    ops_sql = next(s for s in captured if "async_operations" in s and "GROUP BY" in s)
+    assert "status IN ('pending', 'processing', 'failed')" in ops_sql
+    assert "completed" not in ops_sql and "cancelled" not in ops_sql
+
+
+def test_gauges_register_and_emit_cached_values_without_bank_id():
```
nicoloboschi, line 184: Note: tests cover the `per_bank=False` path well, but there's no test for `per_bank=True` — the cardinality-sensitive branch where `bank_id` enters the cache key and the SQL switches to `GROUP BY bank_id`. Worth one test asserting a bank-labelled key lands in the cache and the gauge emits the `bank_id` attribute.

MattJackson:
Thanks for the thorough review — all five points addressed, but the perf one had a twist worth reporting since I verified it against a live 114k-row `memory_units` with `EXPLAIN ANALYZE` rather than assuming.

**Splitting fixed the `failed` count but not the `backlog` count:**

- `consolidation_failed` → ✅ `Index Only Scan using idx_memory_units_consolidation_failed`, ~0.1 ms. Exactly as you described.
- `unconsolidated` backlog → ❌ **still a `Seq Scan` (~92 ms)**, even split out with a predicate that matches `idx_memory_units_unconsolidated` exactly. Root cause: `consolidated_at IS NULL` is true for ~40% of the table (every observation has a null `consolidated_at`), so the planner estimates low selectivity and can't see that `AND fact_type IN ('experience','world')` empties the set. `ANALYZE` doesn't change the plan (structural misestimate, not stale stats); `SET enable_seqscan=off` confirms the index *is* usable and the predicate matches (Bitmap Index Scan, ~0.13 ms).

So the split alone shifted the scan rather than removing it (and the lone backlog seq scan even loses the parallelism the combined scan had).

**Fix applied (`6a6e6156`):** run just the backlog count in a scoped transaction with `SET LOCAL enable_seqscan = off`, which forces the partial-index scan — verified **~0.07 ms**, transaction-scoped, doesn't leak to subsequent queries. The `failed` count is left as-is (no nudge needed). I went with the planner override because it's surgical and needs no schema change, but since it *is* an override I'm happy to switch if you'd prefer an extended-statistics object on `(consolidated_at, fact_type)`, a covering-index tweak, or a different cadence — your call on the perf lever.

EXPLAIN outputs:

```
-- OLD single query, two FILTERs, no WHERE
Parallel Seq Scan on memory_units  (actual time=0.136..43.347 rows=38154 loops=3)
Execution Time: 59.450 ms

-- NEW backlog split, default planner
Seq Scan on memory_units  (Rows Removed by Filter: 114464)
Execution Time: 92.572 ms

-- NEW backlog split, SET LOCAL enable_seqscan=off
Bitmap Index Scan on idx_memory_units_unconsolidated
Execution Time: 0.073 ms

-- NEW failed split, default planner
Index Only Scan using idx_memory_units_consolidation_failed
Execution Time: 0.114 ms
```

Other notes: cache keys are now `NamedTuple`s (`_AsyncOpKey`/`_BacklogKey`), `asyncio` hoisted to module scope, added a comment on `_backlog_task` being process-lifetime, and added `per_bank=True` coverage. 6 backlog tests + the existing 26 in `test_metrics.py` pass; ruff clean.

PR #1988, opened 2026-06-04
fix(opencode): append recall context to existing system message instead of pushing new section

sdrobov:
Not all LLMs support multiple system sections, so recall context is now appended to the first system message rather than added as a separate array element.


nicoloboschi:
Superseded by #2052, which ports this exact approach (fold recall into the first system section) onto current main — your original couldn't merge after the 0.2.4 logging/recall refactor. Credited via Co-authored-by. Thanks @sdrobov! Shipped in 0.2.5.

PR #1989, opened 2026-06-04
feat(agent-framework): Hindsight memory for Microsoft Agent Framework (no MCP)

benfrank241:
## Summary

A first-class **Microsoft Agent Framework** integration that gives MAF agents persistent long-term memory via Hindsight — **without MCP**.

> **Why MAF, not classic Semantic Kernel?** SK's memory API (`SemanticTextMemory`/`MemoryStoreBase`) is deprecated and slated for removal, and its replacement (preview `VectorStore`) is a raw vector-DB contract that fits Hindsight poorly. Microsoft Agent Framework is SK's official successor and exposes a clean **context-provider** memory hook — the right surface for Hindsight.

Lives in `hindsight-integrations/agent-framework/` (package `hindsight-agent-framework`).

> **Status: draft** — opened for review. Not released yet.

## What it does

`HindsightProvider` is an Agent Framework `ContextProvider`:

| Hook | Behavior |
| --- | --- |
| `before_run` | Recall memories relevant to the user's message → inject a `## Memories` block into the agent's instructions. |
| `after_run` | Retain the user input + agent response to the bank. |

So memory happens **automatically** every turn — no MCP, no tools the model must remember to call.

```python
from agent_framework.openai import OpenAIChatClient
from hindsight_agent_framework import HindsightProvider

agent = OpenAIChatClient().as_agent(
    name="assistant",
    context_providers=[HindsightProvider(bank_id="user-123")],
)
```

## Design notes

- **Reuses the LlamaIndex integration's** `_client.py`/`config.py`/`errors.py` pattern and the `hindsight-client` Python SDK (`arecall`/`aretain`/`acreate_bank`).
- **API-churn guard:** Agent Framework's `ContextProvider` API changed across releases (`invoking`/`invoked` → `before_run`/`after_run` + `SessionContext`). I verified the actual contract against the installed `agent-framework-core` 1.7.0 before writing the provider, and the tests subclass the **real** base class so future drift fails loudly. Pinned `agent-framework-core>=1.7,<2`.
- **Resilient:** recall/retain are best-effort — a memory hiccup never blocks the agent.
- **No feedback loop:** recalled memories are injected as *instructions* (not messages), so they're never re-retained; `after_run` also excludes our own source defensively.
- **Standards:** structured config is a `@dataclass`; no raw dicts for structured data (the only `dict`s are MAF's own `state` hook param + a constructor-kwargs builder); no tuple returns.

## Tests / wiring

- **15 unit tests** drive the real `ContextProvider`/`SessionContext` with a mocked Hindsight client: recall injection, retain transcript, feedback-loop avoidance, disabled flags, silent failure, query composition. Plus a gated `requires_real_llm` e2e (live Hindsight round-trip).
- CI job `test-agent-framework-integration` (uv build → `uv sync --frozen` → ruff → pytest `-m "not requires_real_llm"`) + detect-changes filter + aggregate gate; `VALID_INTEGRATIONS`; changelog generator; `integrations.json`; docs page + changelog page + icon.

## Out of scope (v1) / follow-ups

- Plugin / `@ai_function` tools (explicit retain/recall/reflect) — provider surface chosen for v1.
- A classic **Semantic Kernel** adapter — only if there's demand from the existing SK base.
- **.NET / C#** `AIContextProvider` — strategically valuable, but a separate NuGet/dotnet toolchain.


benfrank241:
Synced to `main` and verified review-ready:

- **Merged latest `main`** (was 140 commits behind). Resolved the expected union conflicts in `scripts/release-integration.sh` (VALID_INTEGRATIONS) and `.github/workflows/test.yml` (detect-changes output/filter, the `test-agent-framework-integration` job, and the aggregation `needs:` list). `generate_changelog.py` and `integrations.json` auto-merged with the agent-framework entries intact.
- **API-churn check:** `agent-framework-core` has shipped 1.8.0/1.8.1 since this PR opened. Refreshed the lock to **1.8.1** and confirmed the provider + all tests pass against it (the deterministic tests subclass the real `ContextProvider`/`SessionContext`, so any drift fails loudly). Pin stays `>=1.7,<2`.
- **Tests:** 15 passed, 1 deselected (the `requires_real_llm` e2e roundtrip, gated behind a live server). `ruff check` + `format` clean.
- `check-integrations.mjs` green (gallery entry + doc page + icon present); changelog page is a clean template (no stray "Unreleased").

Note for reviewers: this is a **ContextProvider** integration (no MCP, no tools). It intentionally swallows recall/retain failures and logs them so memory can never block an agent run — hence `HindsightError` is exported for cross-integration API consistency but never raised here.

DK09876:
**Review — looks great, one fix before merge.**

Clean `ContextProvider` integration (no MCP, automatic recall+retain), defaults to Cloud, and fails gracefully so a Hindsight outage never blocks the agent. Test coverage is excellent — especially the feedback-loop case (injected memories aren't re-retained) and the silent-failure paths. Wiring is all there: gated `requires_real_llm` e2e, CI job, `VALID_INTEGRATIONS`, `integrations.json`, docs. Nice work.

**Should fix — unused per-op timeouts (`_client.py`)**
`TIMEOUT_RETAIN`/`TIMEOUT_RECALL`/`TIMEOUT_BANK` are defined but never used — only `TIMEOUT_DEFAULT` is applied (on the client constructor). Since they're named per-operation, I think the intent was to pass them to `arecall`/`aretain`/`acreate_bank`, but those calls don't set a timeout, so every op falls back to the 30s default. Either wire them in or drop them — right now it's dead code + a missed timeout behavior.

**Minor (optional)**
- `HindsightError` is exported but never raised (provider swallows exceptions via `logger.debug`). Fine as API-surface convention, just flagging.
- `_ensure_bank` sets `_bank_initialized = True` even when `acreate_bank` raises — best-effort by design, but a transient creation failure won't retry. Acceptable given recall/retain are guarded; noting only.

benfrank241:
Thanks @DK09876 — good catch. Fixed in c78bc6364.

**Should fix (unused per-op timeouts):** I checked the `hindsight-client` SDK — `arecall`/`aretain`/`acreate_bank` take **no per-call `timeout`** argument; the timeout is only set once on the `Hindsight(...)` constructor. So the per-op constants couldn't be wired in as intended. I dropped `TIMEOUT_RETAIN`/`TIMEOUT_RECALL`/`TIMEOUT_BANK` and kept the single constructor-level `TIMEOUT_DEFAULT` (30s), with a comment explaining it applies to every op and is a graceful-degradation backstop rather than a tight per-op budget.

On the two optional notes:
- **`HindsightError`** — kept as cross-integration API-surface convention (every `hindsight_*` package exports it), as you flagged.
- **`_ensure_bank` marking `_bank_initialized` after a failed `acreate_bank`** — intentional: it's best-effort and recall/retain are guarded, so a transient bank-create blip can't wedge the agent. Left as-is per your note.

`ruff` clean, 15 tests pass.

PR #1998, opened 2026-06-05
ci(obs): enable consolidation dedup (0.97) in the observation benchmark

nicoloboschi:
The observation benchmark exists to track consolidation's near-duplicate rate, but the job ran with dedup **disabled** (default threshold `1.0`), so it measured the baseline rather than the shipped create+update dedup feature (#1977).

Enable it at `0.97` in the obs job env so the benchmark + CPM dashboard track the feature going forward. Safe here because the benchmark runs on embedded pg0 (Postgres); the merge path is Postgres-only, which is why the feature stays opt-in / disabled by default rather than enabled globally (enabling it as the global default would break Oracle deployments).

Dispatched a run on this branch to capture the dedup-on number.


nicoloboschi:
Superseded by #2000, which enables dedup by default (so the benchmark picks it up automatically) and skips it on Oracle.

PR #2005, opened 2026-06-05
fix(reflect): gate mental-model retrieval by policy

### LINKED ISSUE #1971: Reflect Forces Lower-Level Retrieval Even When a Mental Model May Be Sufficient
### Bug Description

The current `reflect` pipeline forces the agent to run the full hierarchical retrieval chain:

```text
search_mental_models -> search_observations -> recall
```

before allowing the agent to answer or decide what to do next.

This reduces the practical value of mental models as reusable synthesized knowledge. A mental model is already a higher-level summary created from lower-level observations and raw memories. If a fresh and relevant mental model can answer the user's question, the agent should not always need to read the lower-level material again before responding.

In the current behavior, even when `search_mental_models` retrieves a fresh, highly relevant mental model that appears sufficient to answer the query, the agent is still forced to continue to `search_observations` and then `recall`.

This has two practical effects:

- It increases latency and cost because each forced retrieval step requires another LLM tool-calling round trip plus retrieval work.
- It can duplicate work because the system may read a mental model, then read observations and memories that contributed to the same mental model, then ask the agent to synthesize the answer again.

Lower-level retrieval is still useful when the mental model is stale, incomplete, ambiguous, or when the user asks for concrete source details. The issue is that the current implementation always forces lower-level retrieval before the agent can answer.

In practice, this behaves more like a fixed extra cost than an explicit validation mechanism.

### Steps to Reproduce

1. Create a bank with memories that can be summarized into a mental model.
2. Create or refresh a mental model for those memories.
3. Ensure the mental model is fresh, for example `is_stale=False`.
4. Call `reflect` with a query that is directly answered by the mental model.
5. Enable trace/tool-call output, or inspect logs/tool traces.
6. Observe that the reflect agent still follows the forced retrieval sequence:

```text
search_mental_models
search_observations
recall
```

The relevant code path is in `run_reflect_agent`:

```python
forced_sequence = []
if has_mental_models:
    forced_sequence.append("search_mental_models")
if include_observations:
    forced_sequence.append("search_observations")
if include_recall:
    forced_sequence.append("recall")

if iteration < len(forced_sequence):
    iter_tool_choice = {"type": "function", "function": {"name": forced_sequence[iteration]}}
else:
    iter_tool_choice = "auto"
```

Because `tool_choice` is set to a specific function, the model cannot answer during these forced iterations. It must produce the specified tool call.

### Expected Behavior

The retrieval chain should be conditional rather than always forced.

If `search_mental_models` returns a mental model that is:

- highly relevant,
- `is_stale=False`,
- non-empty,
- and sufficient to answer the user query,

then the agent should be allowed to answer without being forced to call `search_observations` and `recall`.

Lower-level retrieval should still happen when needed, for example:

- no relevant mental model is found,
- the mental model is stale,
- the query asks for concrete source details,
- the query is time-sensitive,
- the mental model is incomplete or ambiguous,
- or the selected budget prioritizes higher confidence over latency/cost.

When the system decides to continue to lower-level retrieval, the later `search_observations` and `recall` queries should focus on the important conclusions from the mental model that actually need verification. If conflicts are found, the agent should explicitly correct or qualify the mental model using the lower-level evidence.

One possible policy:

- `low` budget: force only `search_mental_models`, then allow `auto`.
- `mid` budget: search mental models first; use observations if the model is stale, incomplete, or insufficient.
- `high` budget: preserve a fuller verification path when reliability is preferred over cost/latency.

### Actual Behavior

The agent is forced through all enabled retrieval layers before it can freely answer:

```text
search_mental_models -> search_observations -> recall
```

This happens even when the mental model is fresh and directly relevant.

With the default `low` reflect budget, `DEFAULT_REFLECT_MAX_ITERATIONS = 10` and the low multiplier is `0.5`, giving `max_iterations = 5`. This is still enough to force all three retrieval layers when mental models, observations, and recall are enabled.

As a result:

- mental models cannot act as an answer boundary,
- `is_stale=False` does not allow short-circuiting,
- `low` budget still performs the full three-level retrieval path,
- later retrieval may duplicate the same knowledge that was already summarized in the mental model,
- and the system pays for additional LLM tool-calling turns plus retrieval work.
- the later `search_observations` and `recall` queries are generated by the LLM, which gives the agent a chance to refine or verify the mental model, but does not guarantee that it checks the important conclusions that actually need verification or resolves conflicts correctly.

### Version

v0.7.2

### LLM Provider

None

Sanderhoff-alt:
## Summary

Fixes #1971.

This PR makes the forced `reflect` retrieval path conditional after the mental-model layer. When `search_mental_models` returns fresh, usable mental models that are sufficient for the user query, low- and mid-budget reflect calls can stop forcing lower-level retrieval and let the agent answer or choose the next action normally.

When lower-level retrieval is still needed, the policy can provide a targeted follow-up query for `search_observations` or `recall`, so deeper retrieval focuses on the mental-model conclusion, gap, or stale point that needs verification instead of blindly repeating the original query.

## Motivation

Before this change, the reflect agent forced the same retrieval sequence whenever all layers were enabled:

```text
search_mental_models -> search_observations -> recall
```

Because each forced iteration uses `tool_choice` with a specific function, the agent could not answer after a sufficient mental-model result. It still had to call `search_observations` and then `recall`.

That weakens the value of mental models as reusable synthesized knowledge:

- fresh and relevant mental models cannot act as an answer boundary;
- low-budget reflect calls still pay for the full three-layer retrieval path;
- later retrieval can duplicate evidence already summarized by the mental model;
- the deeper retrieval query is not guaranteed to verify the specific conclusion or gap that matters.

## Changes

- Adds a structured retrieval-policy decision after `search_mental_models`.
- Keeps `high` budget on the full verification path.
- Allows low- and mid-budget calls to stop forced lower-level retrieval when fresh mental models are sufficient.
- Treats stale, missing-staleness, or empty mental models as unsafe for short-circuiting and falls back to lower-level retrieval.
- Falls back to conservative lower-level retrieval if the policy LLM call or structured output parsing fails.
- Uses retrieval relevance as metadata only, without adding a fixed score threshold.
- Carries a policy-generated `follow_up_query` into forced `search_observations` and `recall` calls when deeper retrieval is needed.
- Records the retrieval-policy decision in the reflect trace.
- Removes the old forced-sequence alignment workaround that inserted an already-executed `search_mental_models` entry into the override sequence.

## Behavior

After this change:

- If no mental models are found, reflect continues to the enabled lower layers.
- If any retrieved mental model is stale or has no usable content, reflect continues to the enabled lower layers.
- If the budget is `high`, reflect preserves the fuller verification path.
- If fresh mental models are sufficient, forced lower-layer retrieval stops and the next agent iteration uses normal `auto` tool choice.
- If fresh mental models are not sufficient, reflect forces only the lower layers requested by the policy and uses the policy's targeted query when available.

## Tests

- `uv run ruff check tests/test_reflect_agent.py`
- `uv run ty check hindsight_api/engine/reflect/agent.py`
- `./scripts/hooks/lint.sh`
- `uv run pytest tests/test_reflect_agent.py::TestMentalModelRetrievalPolicyRealLLM -q`
- `HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 uv run pytest tests/test_reflect_agent.py -q`

The new tests cover:

- stopping forced retrieval when fresh mental models are sufficient;
- preventing stale or empty mental models from short-circuiting;
- preserving full retrieval for `high` budget;
- using a targeted follow-up query for forced lower-layer retrieval;
- real-LLM behavior for both sufficient and verification-needed policy decisions.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -14,6 +14,8 @@ import re
 import time
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
 
+from pydantic import BaseModel, Field
+
 from .models import DirectiveInfo, LLMCall, ReflectAgentResult, TokenUsageSummary, ToolCall
 from .prompts import (
     _extract_directive_rules,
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Opened **#2011** as an alternative approach to the same issue (#1971), for comparison.

The root cause is that the forced iterations set `tool_choice` to a **named function**, which forbids the model from emitting `done` — so a fresh, sufficient mental model can never be an answer boundary. This PR resolves that with a dedicated structured "is the mental model sufficient?" LLM classifier call after the mental-model layer.

#2011 instead makes the decision **deterministically, with no extra LLM call**: after the forced `search_mental_models` result, if the call is low/mid budget and every retrieved model is explicitly fresh (`is_stale is False`) with non-empty content, it stops forcing from the next iteration on. That next iteration *already happens* — it just runs under `auto` — so the agent answers directly, or (having just read the model) issues its own targeted `search_observations`/`recall`. Stale/empty/missing keep the full forced path; `high` budget keeps it too.

The motivating difference is cost on the path the issue is about:

| Scenario | Before | This PR (#2005) | #2011 |
|---|---|---|---|
| model **sufficient** | 3 forced rounds | mm + classifier + done → **−1** | mm + done → **−2** |
| model **not sufficient** | 3 forced rounds | mm + classifier + obs + recall → **+1** | mm + obs/recall under auto → **0** |

#2011 never adds an LLM round-trip, and the targeted follow-up query comes for free (the model writes it after reading the mental model, so the `follow_up_query` plumbing / query-override bookkeeping isn't needed).

The one thing this PR's classifier genuinely buys is a decoupled, conservative sufficiency judgment. While writing #2011's real-LLM e2e tests I actually saw the flip side of *not* having that: when a fresh-but-incomplete model is released to `auto`, gemini-flash-lite sometimes just answers from it rather than digging deeper. So the classifier's conservatism is a real, if costly, benefit — #2011 leans on the deterministic fresh-gate + the `high` budget tier (and, optionally, a one-line prompt nudge) to cover that instead of a per-call LLM tax.

Not trying to block this — just offering the lighter-weight alternative so the maintainers can pick the tradeoff (guaranteed extra call + stronger conservatism vs. zero extra calls + deterministic guard). Happy to converge on whichever direction is preferred.

Sanderhoff-alt:
@nicoloboschi 
Thanks for laying out the tradeoff clearly. 

The main thing I was trying to address with #2005 is the fresh-but-incomplete case: `is_stale=False` and non-empty content make a mental model safe to consider, but they do not necessarily mean it is sufficient  for the query. A separate policy step gives stronger conservatism and explicit traceability, but I agree that it comes with a per-call cost that may be too heavy for the default path.

I’m happy with #2011 as the minimal fix for #1971. If the fresh-but-incomplete behavior shows up in practice, a future follow-up could add a cheaper prompt nudge, config-gated conservative mode, or optional policy classifier for deployments that prefer correctness over latency.

PR #2018, opened 2026-06-05
feat(integrations): add oh-my-openagent (OMO) integration

dcbouius:
## Summary

- Cloud-first Hindsight memory integration for the [oh-my-openagent](https://github.com/code-yeongyu/oh-my-openagent) (OMO) agent harness
- 5 lifecycle hooks: SessionStart, UserPromptSubmit (recall), Stop (retain), SubagentStop (retain), SessionEnd
- Always-apply rule for memory guidance injected into OMO's rules engine
- Config hierarchy: `settings.json` → `~/.hindsight/omo.json` → `HINDSIGHT_*` env vars
- Bearer token auth for Hindsight Cloud mode (`hsk_*` keys), silent skip when no token configured
- Interactive `demo.py` for local dev testing against the Hindsight dev server
- 29 tests (15 config + 14 hook e2e)

OMO has no built-in memory system — this fills that gap with automatic recall on every prompt and retain on session stop, zero-config for cloud users.

## Test plan

- [ ] `cd hindsight-integrations/omo && uv run pytest tests/ -v` passes (29 tests)
- [ ] `python demo.py` against local dev server completes full lifecycle
- [ ] CI job runs successfully


benfrank241 (dismissed):
Two mechanical bits before merge:
 1. Add `"omo"` to `VALID_INTEGRATIONS` in `scripts/release-integration.sh:16` so we can cut releases via the standard tooling — same line as the other 30+ integrations.
2. Rebase on `main` — currently conflicting.

benfrank241:
1. Remove the root pyproject.toml (release-blocking).
  omo is a hooks/plugin integration installed via scripts, not a published package. The other integrations of this kind (codex, claude-code, cursor-cli) have no root
  pyproject.toml. The release workflow (.github/workflows/release-integration.yml) decides package type by "does pyproject.toml exist?" — so this file makes it mis-classify omo as
  a Python package and run uv build, which fails (there's no [project]/[build-system], only [tool.pytest.ini_options]). Please move the pytest config into pytest.ini (or
  tests/conftest.py) and delete the root pyproject.toml. That way omo releases as a plugin (no PyPI publish), which is the intent.

  2. Add omo to the changelog generator map.
  hindsight-dev/hindsight_dev/generate_changelog.py needs an entry, e.g.:
  "omo": IntegrationMeta("hindsight-omo", "OMO"),
  Without it, the release script's changelog-generation step fails.

  3. Add omo to the docs single source of truth.
  The moment a release tag integrations/omo/v* is cut, hindsight-docs/scripts/check-integrations.mjs (run by the build-docs CI job) requires:
  - an entry in hindsight-docs/src/data/integrations.json with an internal "link": "/sdks/integrations/omo", and
  - a matching doc page hindsight-docs/docs-integrations/omo.md (plus an icon under static/img/icons/).

  If these are missing, build-docs goes red on main for every PR, not just this one. (See #2060 for the equivalent cursor-cli fix — easiest to just include it here so it lands
  atomically.)

  Nit: settings.json has no version field — worth adding "version": "0.1.0" so the release script and gallery have a single version to read.

  VALID_INTEGRATIONS and the CI test job are already handled in the PR — thanks for those. Items 1–3 are the blockers; happy to push the doc/changelog scaffolding if you'd rather I
  take that part.

PR #2036, opened 2026-06-06
docs(models): flag Fireworks AI Batch API support in the provider capabilities table

r266-tech:
The provider capabilities table renders Fireworks' Batch API column as `—` (not supported), but it *is* supported:

- `provider == "fireworks"` dispatches to `FireworksLLM` (`llm_wrapper.py:424`)
- `FireworksLLM.supports_batch_api()` overrides to `True` (`fireworks_llm.py:106`)
- the base `OpenAICompatibleLLM` only grants batch to openai/groq (`openai_compatible_llm.py:1236`), so the override is load-bearing

Added `"batchApi": true` to the fireworks entry in the canonical `llmProviders.json` and regenerated the CI-enforced skills mirror (`models.md`). The async Batch API path (cheaper retain extraction) is opt-in — gated behind `HINDSIGHT_API_RETAIN_BATCH_ENABLED=true` with async retain (`DEFAULT_RETAIN_BATCH_ENABLED = False`) — but the capabilities table previously implied Fireworks couldn't use it at all, which is wrong: it's fully supported, users just have to enable it.


nicoloboschi (approved):
Cross-checked all claims against the code: `FireworksLLM` dispatch (llm_wrapper.py:424), `supports_batch_api()` override → True (fireworks_llm.py:106), and the load-bearing base that only grants batch to openai/groq (openai_compatible_llm.py:1236). The capabilities-table fix and regenerated CI-enforced mirror are correct. LGTM.

PR #2040, opened 2026-06-07
fix(trace): preserve RRF source ranks

zwcf5200:
## Summary

- preserve `semantic_rank`, `bm25_rank`, `graph_rank`, and `temporal_rank` in recall traces
- accept both the flattened metadata currently emitted by the recall pipeline and the existing nested `source_ranks` shape
- add a regression test for the flattened metadata path

`SearchTracer.add_rrf_merged()` currently reads `rrf_meta["source_ranks"]`, while the recall pipeline expands `MergedCandidate.source_ranks` into the metadata top level. The trace therefore reports an empty source-rank map even though fusion computed the ranks correctly.

This change only fixes trace serialization; it does not change fusion, reranking, strategy boosts, or result ordering.

## Validation

- `uv run ruff check hindsight-api-slim/hindsight_api/engine/search/tracer.py hindsight-api-slim/tests/test_search_trace.py`
- `uv run pytest hindsight-api-slim/tests/test_search_trace.py::test_rrf_trace_preserves_flattened_source_ranks -q`
- `uv run pytest hindsight-api-slim/tests/test_interleave_fusion.py hindsight-api-slim/tests/test_fusion_cap.py hindsight-api-slim/tests/test_recall_boost.py -q` (29 passed)
- pre-commit hooks: docs generation and lint passed

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/search/tracer.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/tracer.py
@@ -358,12 +358,15 @@ class SearchTracer:
         """
         self.rrf_merged = []
         for rank, (doc_id, data, rrf_meta) in enumerate(merged_results, start=1):
+            source_ranks = rrf_meta.get("source_ranks")
+            if source_ranks is None:
+                source_ranks = {key: value for key, value in rrf_meta.items() if key.endswith("_rank")}
             self.rrf_merged.append(
                 RRFMergeResult(
                     node_id=doc_id,
                     text=data.get("text", ""),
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM. Confirmed the bug: the recall pipeline flattens `source_ranks` into the top level of `rrf_meta` (memory_engine.py:4142/4246 via `**mc.source_ranks`), so the old `rrf_meta.get("source_ranks", {})` always returned empty. The fallback correctly harvests the `*_rank` keys. Fast deterministic regression test is the right call for pure serialization logic. Minimal scope — no fusion/ranking/ordering changes.

PR #2042, opened 2026-06-07
fix(retain): preserve client timezone context

### LINKED ISSUE #2033: [Bug] Extractor time-zone context is lost at retain: replace(tzinfo=UTC) silently drops offset, and no client_timezone reaches the extraction LLM
### Problem

When a client sends an offset-aware ISO-8601 datetime (e.g. `2026-06-05T20:36:00+08:00`) inside the retain content JSON, the server-side parsing hard-codes UTC instead of converting the offset properly, **silently discarding the client’s timezone**. Meanwhile the extraction LLM receives no indication of the user’s local timezone, so it always renders dates/times in UTC terms. The resulting memory prose permanently records the wrong local time.

### Real-world impact (UTC+8 / Asia/Shanghai)

1. Local event at **20:36 CST**.
2. Client retain payload includes a structured JSON message with `timestamp: "2026-06-05T12:36:00Z"` (UTC Zulu, because that is what the API expects).
3. At `orchestrator.py`, the timestamp is parsed and then `parsed.replace(tzinfo=UTC)` is applied.
4. Extraction LLM sees **12:36 UTC** and generates memory text:  
   ```
   2026-06-05 12:36 起, user was attacked.
   ```
5. For a +08:00 user, 12:36 means **noon**, not evening. The memory’s natural-language date/time is semantically wrong for the user’s context.
6. Later, if the client does presentation-layer conversion, the raw text already says "12:36", so the client’s converted "20:36 CST" and the stored "12:36" are treated as **two separate events**.

### Two parallel gaps

| Gap | Where | What happens | What should happen |
|---|---|---|---|
| **A. Timezone parsing bug** | `orchestrator.py` → `parse_datetime_flexible()` → `replace(tzinfo=UTC)` | Offset is overwritten, not converted. A time that was originally `+08:00` becomes UTC by label only, without the 8-hour arithmetic. | `astimezone(UTC)` so offset-aware timestamps normalize to UTC numerically. |
| **B. Missing client-timezone hint** | `fact_extraction.py` prompt | The extraction LLM has zero knowledge of the user’s actual timezone. It uses the server’s UTC wall-clock when writing dates/times in prose. | A `client_timezone` field (e.g. `"Asia/Shanghai"`) is injected into the extraction LLM system prompt so it writes local time. |

Both gaps are needed for correctness:
- Fix only A: Physical timeline is right, but LLM still writes "12:36" because it thinks in UTC.
- Fix only B: LLM knows the timezone, but `replace` has already dropped the offset, so the LLM is reasoning from a wrong UTC anchor.

### Suggested minimal changes

1. **Bug fix** in retain orchestrator  
   Change `parsed.replace(tzinfo=UTC)` → `parsed.astimezone(UTC)` so offset-aware datetimes are numerically normalized to UTC before storage.

2. **Feature request** — new optional retain field `client_timezone`  
   Example:
   ```json
   {
     "bank_id": "gh-user",
     "items": [{
       "content": "[{\"role\":\"user\",\"content\":\"...\",\"timestamp\":\"2026-06-05T12:36:00Z\"}]",
       "client_timezone": "Asia/Shanghai"
     }]
   }
   ```
   When present, inject into the extraction LLM system prompt:
   ```
   The user's local timezone is Asia/Shanghai (+08:00).
   When generating date/time references in extracted memory text, use the user's local time.
   ```

These are additive, backward-compatible, and keep the server-side storage fully UTC (no change to DB or indexing).

### Related

- #1232 — Timezone mismatch in recalled memory dates (the same root cause, but framed around recall-time symptoms)
- #1789 — OpenClaw UTC timestamp labeling (adjacent, but toolchain-specific)

### Would a PR help?

Happy to contribute if the maintainers agree on the approach.

xmh1011:
Summary
- add optional `client_timezone` on retain items and validate it as an IANA timezone
- normalize offset-aware retain timestamps to UTC instead of relabeling the wall-clock time
- pass client timezone through normal retain, Batch API retain, streaming retain, and delta retain paths
- render the extraction Event Date in the client timezone while keeping the UTC instant visible for storage/reference
- regenerate OpenAPI, clients, and the docs skill references

Closes #2033

Tests
- `uv run --project hindsight-api-slim pytest -n 0 hindsight-api-slim/tests/test_retain_timezone.py hindsight-api-slim/tests/test_fact_extraction_metadata.py -q`
- `uv run --project hindsight-api-slim ruff check hindsight-api-slim/hindsight_api/api/http.py hindsight-api-slim/hindsight_api/engine/retain/types.py hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py hindsight-api-slim/tests/test_retain_timezone.py`
- `./scripts/generate-openapi.sh`
- `./scripts/generate-clients.sh`
- `./scripts/generate-docs-skill.sh`
- `./scripts/hooks/lint.sh`

Note: the new `hs_llm_core` regression test is present but skipped in my local run because `HINDSIGHT_API_LLM_API_KEY` is not configured locally.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -13,6 +13,7 @@ import uuid
 from contextlib import asynccontextmanager
 from datetime import datetime, timezone
 from typing import Any, Literal
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 from fastapi import Depends, FastAPI, File, Form, Header, HTTPException, Query, Request, UploadFile
 from fastapi.middleware.gzip import GZipMiddleware
@@ -446,6 +447,7 @@ class MemoryItem(BaseModel):
             "example": {
                 "content": "Alice mentioned she's working on a new ML model",
                 "timestamp": "2024-01-15T10:30:00Z",
+                "client_timezone": "America/Los_Angeles",
                 "context": "team meeting",
                 "metadata": {"source": "slack", "channel": "engineering"},
                 "document_id": "meeting_notes_2024_01_15",
@@ -468,6 +470,14 @@ class MemoryItem(BaseModel):
     context: str | None = None
     metadata: dict[str, str] | None = None
     document_id: str | None = Field(default=None, description="Optional document ID for this memory item.")
+    client_timezone: str | None = Field(
+        default=None,
+        description=(
+            "Optional IANA timezone name for the client/user (e.g. 'Asia/Shanghai'). "
+            "When present, fact extraction renders natural-language time references in this local timezone while "
+            "stored timestamps remain normalized to UTC."
+        ),
+    )
     entities: list[EntityInput] | None = Field(
         default=None,
         description="Optional entities to combine with auto-extracted entities.",
@@ -544,6 +554,22 @@ class MemoryItem(BaseModel):
                     f"Invalid timestamp/event_date format: '{v}'. Expected ISO format like '2024-01-15T10:30:00' or '2024-01-15T10:30:00Z', or the special value 'unset' to store without a timestamp."
                 ) from e
         raise ValueError(f"timestamp must be a string or datetime, got {type(v).__name__}")
+
+    @field_validator("client_timezone", mode="before")
+    @classmethod
+    def validate_client_timezone(cls, v):
+        if v is None or v == "":
+            return None
+        if not isinstance(v, str):
+            raise ValueError(f"client_timezone must be a string, got {type(v).__name__}")
+        timezone_name = v.strip()
+        if not timezone_name:
+            return None
+        try:
+            ZoneInfo(timezone_name)
+        except ZoneInfoNotFoundError as e:
+            raise ValueError(f"Invalid client_timezone: '{v}'. Expected an IANA timezone like 'Asia/Shanghai'.") from e
+        return timezone_name
 
 
 class RetainRequest(BaseModel):
@@ -6060,6 +6086,8 @@ def _register_routes(app: FastAPI):
                     content_dict["event_date"] = None
                 elif item.timestamp:
                     content_dict["event_date"] = item.timestamp
+                if item.client_timezone:
+                    content_dict["client_timezone"] = item.client_timezone
                 if item.context:
                     content_dict["context"] = item.context
                 if item.metadata:
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -9,8 +9,9 @@ import asyncio
 import json
 import logging
 import re
-from datetime import datetime, timedelta
+from datetime import UTC, datetime, timedelta
 from typing import Any, Literal, cast
+from zoneinfo import ZoneInfo
 
 from pydantic import BaseModel, ConfigDict, Field, create_model, field_validator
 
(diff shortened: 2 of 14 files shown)


xmh1011:
Detailed fix notes for #2033

What was broken
- Retain could receive a timestamp that represented the correct event instant, but extraction only had the normalized/reference instant and no caller-local timezone context.
- For users whose local day differs from the UTC day, relative terms like “today”, “tonight”, or time-of-day phrasing could be interpreted against the UTC date instead of the user's local date.
- Offset-aware timestamps also now have an explicit invariant: convert to UTC with `astimezone(UTC)`, not by relabeling the wall-clock time as UTC.

What changed
- Added optional per-item `client_timezone` to retain inputs. It is validated as an IANA timezone in both the HTTP API model and the internal retain content builder.
- Kept stored/reference timestamps normalized to UTC, but changed the extraction user message to render the Event Date in the client timezone when provided, including both the local ISO timestamp and the UTC instant.
- Passed `client_timezone` through normal retain, Batch API retain request generation, streaming retain chunking, and delta retain content construction.
- Left the bank-agnostic system prompt unchanged; the timezone context is in the per-request user message so prompt caching behavior is not fragmented by user-specific timezone values.
- Regenerated OpenAPI, Python/TypeScript/Go client models, and the generated docs skill references.

Compatibility
- `client_timezone` is optional and defaults to current behavior when omitted.
- No DB schema change is needed: persisted timestamps remain UTC; this only changes extraction context and API/client surface.

Regression coverage
- Deterministic tests cover offset-aware timestamp conversion, HTTP validation, internal retain content propagation, and user-message rendering.
- Added an `hs_llm_core` LLM-judge regression that exercises the real extraction path for a cross-day case: `2025-12-31T16:30:00Z` is `2026-01-01T00:30:00+08:00` in `Asia/Shanghai`, so “today” should resolve to Jan 1 locally instead of Dec 31 UTC.

Validation run locally
- `uv run --project hindsight-api-slim pytest -n 0 hindsight-api-slim/tests/test_retain_timezone.py hindsight-api-slim/tests/test_fact_extraction_metadata.py -q` -> 7 passed, 1 skipped. The skipped test is the new `hs_llm_core` case because this local environment does not have `HINDSIGHT_API_LLM_API_KEY` configured.
- `uv run --project hindsight-api-slim ruff check hindsight-api-slim/hindsight_api/api/http.py hindsight-api-slim/hindsight_api/engine/retain/types.py hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py hindsight-api-slim/tests/test_retain_timezone.py` -> passed.
- `./scripts/generate-openapi.sh` -> passed.
- `./scripts/generate-clients.sh` -> passed.
- `./scripts/generate-docs-skill.sh` -> passed.
- `./scripts/hooks/lint.sh` -> passed.

huangjingxun:
老哥神速，头天提的issue，隔天PR就来了，厉害👍

nicoloboschi:
Thanks for the detailed write-up — but I believe this is solving a client-side problem inside the server, and the change isn't needed. Two things:

**1. The "`replace(tzinfo=UTC)` drops the offset" bug doesn't exist.** `parse_datetime_flexible` only applies `replace(tzinfo=UTC)` to **naive** datetimes; offset-aware values are returned untouched (`orchestrator.py:73-84`). The 8-hour arithmetic is never lost.

**2. The `timestamp` field already accepts offset-aware ISO-8601**, and the offset already reaches the extraction LLM as local time. With `timestamp: "2026-06-05T20:36:00+08:00"` the prompt the model sees is:

```
Event Date: Friday, June 05, 2026 (2026-06-05T20:36:00+08:00)
```

i.e. **20:36 local**, already correct. The issue's scenario only breaks because the client chose to send `12:36:00Z` (Zulu). The fix is to send the offset, not to add a parallel `client_timezone` field. A fixed offset is sufficient to render a single event's local time — an IANA zone only adds value when you need DST rules for a *different* instant.

This is doable from every client today:

**HTTP / curl**
```json
{ "items": [ { "content": "...", "timestamp": "2026-06-05T20:36:00+08:00" } ] }
```

**Python** (`hindsight-client`)
```python
from hindsight_client_api.models.memory_item import MemoryItem
from hindsight_client_api.models.timestamp import Timestamp

MemoryItem(content="...", timestamp=Timestamp("2026-06-05T20:36:00+08:00"))
```
(Prefer the string form — passing a `datetime` object trips a `to_json` serialization quirk in the generated client.)

**TypeScript / JS**
```ts
const item = { content: "...", timestamp: "2026-06-05T20:36:00+08:00" };
// NOTE: new Date().toISOString() emits "...Z" (UTC) and loses local time —
// format with the offset instead (e.g. date-fns-tz `format(d, "yyyy-MM-dd'T'HH:mm:ssXXX", { timeZone })`).
```

**Go**
```go
ts := "2026-06-05T20:36:00+08:00"
item := MemoryItem{Timestamp: *NewNullableTimestamp(&Timestamp{String: &ts})}
```

**Rust**
```rust
MemoryItem { timestamp: Some("2026-06-05T20:36:00+08:00".to_string()), ..Default::default() }
```

Given that, I'd rather close this and address #2033 with a docs note ("send an offset-aware timestamp and local time is preserved") than merge the new field plus the circular UTC→local round-trip. Really appreciate the effort here, and sorry to redirect it — the underlying ask is already supported.

Issue #1758, opened 2026-05-26
[BUG] Three integration plugin issues across OpenClaw, OpenCode, and Claude Code


joyous-coder:
## Summary

Hindsight integration plugins have issues on all three platforms (OpenClaw, OpenCode, Claude Code). Each plugin has subtle bugs in edge cases — dispatch routing checks, hook execution order, and platform compatibility — that silently break autoRecall/autoRetain or MCP server startup.

---

#### 1. OpenClaw — recall/retain silently skipped on non-main dispatch surfaces

**Plugin:** `@vectorize-io/hindsight-openclaw`

In `dist/index.js`, the `resolveAndCacheIdentity()` function has this guard:

```js
if (parsedSession.provider &&
    options.dispatchChannel &&
    parsedSession.provider !== options.dispatchChannel) {
    return; // ← skips retain and recall entirely
}
```

When the session provider is `main` (e.g. `agent:main:main`) but the dispatch channel is `webchat` (Control UI / webchat surface), the check evaluates `main !== webchat` → recall and retain are silently skipped. No error is logged, so users have no idea why auto-memory doesn't work.

This affects anyone using hindsight with any non-`main` dispatch surface on OpenClaw.

#### 2. OpenCode — autoRecall silently skipped due to hook execution order

**Plugin:** `@vectorize-io/opencode-hindsight`

The plugin uses two hooks to manage autoRecall:
- `session.created` — marks the sessionId in the `recalledSessions` set
- `system.transform` — checks `recalledSessions.has(sessionId)`, skips if already recalled

The intended hook order is `session.created` → `system.transform`. However, the actual execution order is **`system.transform` fires before `session.created`**. Since `recalledSessions` is always empty when `system.transform` runs, the guard always skips the recall, effectively disabling autoRecall entirely.

#### 3. Claude Code — two platform/encoding issues

**3a. Windows path incompatibility in `run_mcp.sh`**

`scripts/run_mcp.sh` hardcodes the Unix venv path (`venv/bin/python`). On Windows, Python venv uses `venv\Scripts\python.exe`. Additionally, `.mcp.json` uses `bash` to execute the script, and MSYS2 (Git Bash) path conversion fails when `${CLAUDE_PLUGIN_ROOT}` expands to a `C:/Users/...` path under `CreateProcess`. The MCP server never starts with no useful error message.

**3b. Unicode surrogate crash in recall**

[...]

Issue #1564, opened 2026-05-10
run_mcp.sh re-bootstraps venv every session on Windows; `[ -x bin/python ]` never matches `python.exe`


ottopichlhoefer:
## Summary

On Windows (msys2/Git Bash + native Python), `hindsight-integrations/claude-code/scripts/run_mcp.sh` re-creates the venv on every Claude Code session start. Each new session collides with the previously spawned MCP server still holding `python3.exe`/`pip.exe` open, and the bootstrap fails. Claude Code surfaces this as:

```
Failed to reconnect to plugin:hindsight-memory:hindsight.
```

## Environment

- Windows 11
- Plugin version: `0.6.4` (matches `main` at time of filing)
- Bash: msys2 5.2.37 (`/usr/bin/bash`)
- Python: `C:\msys64\mingw64\bin\python.exe` 3.12.11 — produces a POSIX-layout venv (`bin/`, not `Scripts/`), but `bin/` contains only `.exe` files (`python.exe`, `python3.exe`, `pip.exe`, …). No extensionless `python` file exists.
- `~/.hindsight/claude-code.json` has `enableKnowledgeTools: true` and a valid token, so the server is supposed to stay alive.

## Reproduction

`bash -x scripts/run_mcp.sh` shows:

```
+ '[' '!' -x .../venv/bin/python ']'        # TRUE — no extensionless 'python'
+ diff -q REQ_SRC REQ_CACHED                # FAILS first run (cache is missing)
+ python3 -m venv ...                       # silently fails (msys python3 not on PATH)
+ python -m venv ...                        # Error: [Errno 13] Permission denied:
                                            #   '.../venv/bin/python3.exe'
```

After the first failure, orphan \`pip.exe\` / \`python3.exe\` from the prior aborted bootstrap remain alive and lock the venv files, so every subsequent attempt also hits \`ERROR_SHARING_VIOLATION\`.

## Root cause

Two compounding bugs in \`scripts/run_mcp.sh\`:

1. **\`[ -x \"\${VENV}/bin/python\" ]\` never matches on Windows.** Bash's \`-x\` test calls \`access(2)\` on the literal path string and does not honor PATHEXT. The venv only contains \`python.exe\`, so the probe always fails and the bootstrap branch fires every session.
2. **\`REQ_CACHED\` is the last write of the bootstrap branch.** A failed bootstrap leaves it absent, guaranteeing the diff check fails forever — even after fixing (1).

## Suggested fix

[...]

PR #2066, opened 2026-06-08
fix(claude-code-mcp): resolve venv interpreter in the Windows Scripts/ layout (#1758, 3a)

eldar702:
## Problem

`hindsight-integrations/claude-code/scripts/run_mcp.sh` resolves the plugin venv interpreter via `resolve_py()`, which only probes `${VENV}/bin/python` and `${VENV}/bin/python.exe`. A **standard** Windows CPython venv — produced by the python.org installer, the Windows Store Python, and `py -m venv` — puts the interpreter at `${VENV}/Scripts/python.exe`, never under `bin/`.

So on the common Windows setup `resolve_py` returns an empty `$PY`, the launcher falls through to venv **re-creation**, and re-creation then fails whenever `python`/`python3` aren't on the spawning process's PATH (frequent when Claude Code's VS Code extension host spawns the server) → "MCP server never starts, no useful error".

This is sub-item **3a** of #1758. The prior fix (#1564/#1565) only added the msys2/mingw `bin/python.exe` layout; this composes cleanly with it.

## Fix (`scripts/run_mcp.sh`)

- Add one `elif` branch to `resolve_py()` for the `${VENV}/Scripts/python.exe` layout (with the matching `Scripts/pip.exe`).
- Update the now-misleading `venv create failed: no python interpreter at ${VENV}/bin/` message to also mention `${VENV}/Scripts/`.

POSIX behaviour is unchanged — the new branch only matches Windows-layout venvs, and the existing `bin/` branches still take precedence.

## Test (RED-first)

New `tests/test_run_mcp.py` invokes the **real** bash `resolve_py` function (extracted from the script so none of its top-level side effects run) against a fabricated venv tree:

- `test_windows_scripts_layout_is_resolved` — venv has `Scripts/python.exe` and **no** `bin/`. **RED on `main`** (`resolve_py` returns empty `$PY`), **GREEN** after the `elif`.
- `test_posix_bin_layout_is_resolved` — regression guard asserting the POSIX `bin/python` path still wins.

## Verification

```
$ python -m pytest tests/test_run_mcp.py -v   # pre-fix
test_windows_scripts_layout_is_resolved FAILED   # got: '' (the bug)
test_posix_bin_layout_is_resolved        PASSED

$ python -m pytest tests/                      # post-fix
183 passed in 0.61s
$ bash -n scripts/run_mcp.sh                    # syntax OK
```

Scope: 2 files, +105/-1, no new dependencies, scoped to 3a only.

🤖 AI-assistance: drafted with Claude (Opus 4.8), reviewed for correctness

### DIFF
--- a/hindsight-integrations/claude-code/scripts/run_mcp.sh
+++ b/hindsight-integrations/claude-code/scripts/run_mcp.sh
@@ -9,6 +9,8 @@ REQ_CACHED="${CLAUDE_PLUGIN_DATA}/requirements.txt"
 
 # Resolve the venv interpreter. On Windows-built venvs `bin/python` is
 # `python.exe`; bash's `[ -x ]` does not honor PATHEXT, so probe both forms.
+# Standard Windows CPython (python.org installer, Windows Store, `py -m venv`)
+# puts the interpreter under `Scripts/` instead of `bin/` — probe that too.
 resolve_py() {
   if [ -x "${VENV}/bin/python" ]; then
     PY="${VENV}/bin/python"
@@ -16,6 +18,9 @@ resolve_py() {
   elif [ -x "${VENV}/bin/python.exe" ]; then
     PY="${VENV}/bin/python.exe"
     PIP="${VENV}/bin/pip.exe"
+  elif [ -x "${VENV}/Scripts/python.exe" ]; then
+    PY="${VENV}/Scripts/python.exe"
+    PIP="${VENV}/Scripts/pip.exe"
   else
     PY=""
     PIP=""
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Approving. Minimal, well-scoped fix for the Windows `Scripts/` venv layout (#1758, 3a). Branch ordering is correct (new elif appended last, POSIX paths take precedence), the RED-first test genuinely guards the regression, and the existing `test-claude-code-integration` CI job auto-collects the new test. Nits only: $PIP is set but untested, and the exec-bit assumption is pre-existing.

eldar702:
Thanks for the quick merge, @nicoloboschi — happy to have helped harden the Windows `Scripts/` venv-interpreter resolution. 🙏

PR #2068, opened 2026-06-08
Implement Memory Guard Lite for OSS

chrislatimer:
Allow users to prevent token and secret leakage in agent memory. feat(memory-defense): reject quarantine action in policy parser

refactor(retain): drop quarantine branch from orchestrator

test(retain): remove quarantine-path tests

refactor(memory-defense): remove DefenseAction.QUARANTINE enum value

refactor(api): remove include_quarantined query parameter

refactor(recall): drop include_quarantined parameter from memory engine

test(memory-defense): replace stale parser-reject test with full-union accept test

The previous parametrized test asserted parse_policy() should 422 on any detector name other than sensitive_data. That contract was deliberately widened on 2026-06-07 so cloud-style policies pass through api-slim's parser unchanged. The test was stale; the runtime is correct.

Replaced with test_parse_policy_accepts_full_detector_union, which proves the actual contract: all 7 detector names are valid in the parser, with dispatch and entitlement enforcement deferred to the loaded extension.

Memory defense UI


nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/2077

PR #2070, opened 2026-06-08
fix(control-plane): match bank ID from URL regardless of locale prefix

### LINKED ISSUE #2069: Control Plane: bank selection lost on page refresh (URL locale prefix mismatch)
## Bug Description

In the Control Plane web UI, selecting a bank from the dropdown works correctly. However, refreshing the page causes the selected bank to be lost — the UI resets to "Select a bank".

## Root Cause

The URL path follows the pattern `/{locale}/banks/{bankId}` (e.g. `/en/banks/hermes`), but `bank-context.tsx` parses the bank ID from the URL using:

```
const bankMatch = pathname?.match(/^\/banks\/([^/?]+)/);
```

The `^` anchor requires the path to start with `/banks/...`, but due to Next.js i18n routing, the actual path starts with the locale prefix (`/en/`). The regex never matches, so `currentBank` always remains `null` on page load.

When selecting a bank via the dropdown, `setCurrentBank()` is called directly before `router.push()`, which is why the interactive flow works. Only the page-refresh path is broken.

## Suggested Fix

Remove the `^` anchor so it matches `/banks/{bankId}` at any position:

```
const bankMatch = pathname?.match(/\/banks\/([^/?]+)/);
```

## Steps to Reproduce

1. Open Control Plane
2. Select any bank from the dropdown
3. Press F5 / refresh the page
4. The bank selection resets to "Select a bank"

xiaoxinova:
## What does this PR do?

Control Plane loses the selected bank on page refresh.

Root cause: `bank-context.tsx` parses the bank ID from the URL using a regex with a `^/banks/` anchor. However, Next.js i18n routing produces paths like `/{locale}/banks/{bankId}` (e.g. `/en/banks/hermes`). The `^` anchor prevents a match, so `currentBank` stays `null` on page load.

Fix: remove the `^` anchor to match `/banks/` at any path position.

## Related Issue

Fixes #2069

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hindsight-control-plane/src/lib/bank-context.tsx`: remove `^` from URL regex anchor (1 line)

## How to Test

1. Open Control Plane
2. Select a bank
3. Refresh the page
4. The bank should remain selected

## Test Results

- All 65 existing vitest tests passed (5 files, 0 failures, 1.03s)

### DIFF
--- a/hindsight-control-plane/src/lib/bank-context.tsx
+++ b/hindsight-control-plane/src/lib/bank-context.tsx
@@ -58,7 +58,7 @@ export function BankProvider({ children }: { children: React.ReactNode }) {
 
   // Initialize bank from URL on mount
   useEffect(() => {
-    const bankMatch = pathname?.match(/^\/banks\/([^/?]+)/);
+    const bankMatch = pathname?.match(/\/banks\/([^/?]+)/);
     if (bankMatch) {
       setCurrentBank(decodeURIComponent(bankMatch[1]));
     }


nicoloboschi (approved):

nicoloboschi:
Superseded by #2075, which fixes this at the routing layer (`localePrefix: "never"`) so the locale slug never appears in the URL for any language — the unanchored regex then isn't needed. Closing in favor of that approach (the locale slug was unintentional for an authed admin UI).

PR #2071, opened 2026-06-09
feat(api): expose retrieval scores in recall API via include_scores

### LINKED ISSUE #1624: [feature] ~ Expose retrieval scores in recall API response for transparency and threshold-based filtering
### Use Case

I'm building an AI agent system that uses Hindsight's recall API for memory retrieval. My application needs to make intelligent decisions about which retrieved results to act on — not just consume everything returned, but selectively use results based on their confidence/relevance level. Today I have to enable `trace: true` and parse debug output just to get any score signal, which is verbose and not suitable for production use.

### Problem Statement

Currently, the recall API does not return any numeric score alongside results. The documentation acknowledges that scores exist internally but intentionally omits them from the response, citing that "raw retrieval scores are not meaningful on an absolute scale." While I understand the reasoning around relative ordering, this design creates three concrete problems:

1. **Transparency**: Users cannot see why certain results ranked higher than others. The ordering alone gives no insight into the confidence gap between results.

2. **Debugging**: Diagnosing poor retrieval quality requires enabling `trace: true`, which dumps verbose internal debug data. This is not practical in production pipelines and leaks internal implementation details.

3. **Custom filtering**: Advanced users cannot implement threshold-based filtering logic. For example, in an agent system, I may want to discard results below a certain relevance confidence rather than blindly passing all `top_k` results to the LLM context.

Scores ARE already computed internally (cross-encoder score, temporal proximity boost, recency boost, final combined score) and ARE already exposed via `trace: true`. The gap is simply that there's no lightweight way to get the final combined score without opting into full trace mode.

With this feature, I would be able to:

- **Filter low-confidence results** in my agent pipeline before passing context to an LLM, reducing hallucination risk from irrelevant memories
- **Understand retrieval quality** at a glance without needing to enable trace mode or parse verbose debug payloads
- **Implement dynamic `top_k` logic** — e.g., take up to 10 results but stop early if score drops below a threshold
- **Monitor retrieval health** in production by logging score distributions over time, enabling proactive detection of embedding drift or recall degradation
- **Build user-facing explanations** that show why a particular memory was surfaced in an AI application

**Proposed API change (non-breaking):**

Add an optional `include_scores` parameter (default: `false`) to the recall request. When enabled, each result object includes a `score` field with the final normalized combined score (0.0–1.0):

```json
{
  "results": [
    {
      "id": "...",
      "content": "...",
      "score": 0.87
    }
  ]
}
```

This is strictly additive — it does not change existing behavior and preserves the current default experience. Advanced users who want the full breakdown can still use `trace: true`.

### Proposed Solution

It would be great if Hindsight could expose scores in the recall response via an opt-in flag (e.g., `include_scores: true`). The score returned should be the final normalized combined score already computed by the ranking pipeline — no new computation needed.

Additionally, an optional `score_threshold` request parameter would allow server-side filtering, so users don't have to over-fetch and filter client-side:

```json
{
  "query": "...",
  "top_k": 10,
  "include_scores": true,
  "score_threshold": 0.6
}
```

This keeps the API ergonomic for simple use cases (no scores by default) while unlocking powerful patterns for advanced users building production agent systems.

### Alternatives Considered

_No response_

### Priority

Nice to have

### Additional Context

_No response_

### Checklist

xiaoxinova:
## Summary

Add an optional `include_scores` parameter to the recall endpoint. When enabled, each result includes a `score` field with the final combined relevance score (0.0–1.0). Default is `false` — no breaking change.

Closes #1624

## Changes

| File | Change |
|------|--------|
| `hindsight-api-slim/hindsight_api/api/http.py` | Add `include_scores` to `RecallRequest`, `score` to `RecallResult`, pipe through in `_fact_to_result` |
| `hindsight-api-slim/hindsight_api/engine/response_models.py` | Add optional `score` field to `MemoryFact` |
| `hindsight-api-slim/hindsight_api/engine/memory_engine.py` | Pass `combined_score` from `ScoredResult.to_dict()` into `MemoryFact` |

## Design

- Scores are already computed internally in the reranking pipeline (`ScoredResult.combined_score`) and exposed via `trace: true` — this just surfaces the final value without requiring full trace mode.
- The combined score is the output of Hindsight's existing scoring pipeline (cross-encoder score → normalization → recency/temporal boosts → weighted combination), already normalized to 0.0–1.0.
- Non-breaking: `include_scores` defaults to `false`, preserving existing API behavior.

## Testing

- When `include_scores: false` (default): `score` field is omitted from response — existing behavior unchanged.
- When `include_scores: true`: each result object includes `score: 0.87` with the final combined relevance score.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -182,6 +182,10 @@ class RecallRequest(BaseModel):
         description="Compound tag filter using boolean groups. Groups in the list are AND-ed. "
         "Each group is a leaf {tags, match} or compound {and: [...]}, {or: [...]}, {not: ...}.",
     )
+    include_scores: bool = Field(
+        default=False,
+        description="If true, include the relevance score for each result.",
+    )
 
     @field_validator("query")
     @classmethod
@@ -234,6 +238,7 @@ class RecallResult(BaseModel):
     metadata: dict[str, str] | None = None  # User-defined metadata
     chunk_id: str | None = None  # Chunk this fact was extracted from
     tags: list[str] | None = None  # Visibility scope tags
+    score: float | None = None  # Relevance score (only set when include_scores is enabled)
     source_fact_ids: list[str] | None = (
         None  # IDs of source facts (observation type only, when source_facts is enabled)
     )
@@ -3376,6 +3381,7 @@ def _register_routes(app: FastAPI):
                     metadata=fact.metadata,
                     chunk_id=fact.chunk_id,
                     tags=fact.tags,
+                    score=fact.score if request.include_scores else None,
                     source_fact_ids=fact.source_fact_ids,
                 )
 
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -4709,6 +4709,7 @@ class MemoryEngine(MemoryEngineInterface):
                         metadata=result_dict.get("metadata"),
                         chunk_id=result_dict.get("chunk_id"),
                         tags=result_dict.get("tags"),
+                        score=result_dict.get("combined_score"),
                         source_fact_ids=source_fact_ids_by_obs.get(result_id) if include_source_facts else None,
                     )
                 )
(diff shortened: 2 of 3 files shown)


nicoloboschi:
hey, thanks for the pr. scores are meaningless in Hindsight since they are used to sort results but the threshold and absolute numbers are not very useful so we prefer to not expose any score for now. the scores you see in the trace are really for debugging

PR #2097, opened 2026-06-09
feat(providers): add Nous Portal LLM provider with rotating inference…

Spaceman-Spiffy:
…:invoke JWT

Adds a 'nous' provider that speaks the OpenAI-compatible wire format (thin subclass of OpenAICompatibleLLM) but authenticates with a short-lived inference:invoke JWT instead of a static key. Token refresh delegates to Hermes' resolve_nous_runtime_credentials() for a single source of truth against ~/.hermes/auth.json, with proactive pre-expiry refresh and reactive refresh on 401.

- nous_auth.py: NousAuthManager (proactive + reactive refresh, thread-safe)
- nous_llm.py: NousLLM (refresh-before-call, 401 retry-once)
- llm_wrapper.py: register nous in dispatch, validator, no-key set, base-url default


nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/2102

PR #2116, opened 2026-06-10
feat(claude-code): add surface:claude-code to default retainTags

PathKnower:
## What

Adds `surface:claude-code` to the Claude Code integration's default `retainTags` (alongside the existing `{session_id}`), so facts written by this integration carry surface attribution at the tag level out of the box.

## Why

The one-bank-many-surfaces pattern is the documented way to share memory across integrations, and tags are the documented isolation/filter mechanism for it. But today nothing distinguishes at the tag level which surface wrote a fact:

* `retainContext` carries a `"claude-code"` label, but context is not a recall filter.
* Tags are the native filter (`tags=[...]`, `tag_groups`), and the shipped default only carries the session id.

With a default surface tag, recall can include or exclude a surface directly, e.g. `tags=["surface:claude-code"]`, or a `tag_groups` exclusion such as:

```json
{ "tag_groups": [ { "not": { "tags": ["surface:claude-code"], "match": "any_strict" } } ] }
```

This mirrors what multi-surface users already do by hand (e.g. a Telegram-bot surface tagging `surface:hermes` via its own client code) and makes the convention available to every integration user without config.

## Compatibility

* Users who override `retainTags` (in `~/.hindsight/claude-code.json` or env) are unaffected; defaults only apply when not overridden.
* Forward-only: existing facts are not retro-tagged.
* No code paths changed; the tag rides the existing `retainTags` mechanism (literal tag, no new template variable).

## Changes

* `settings.json`: default `retainTags` is now `["{session_id}", "surface:claude-code"]`
* `README.md`: config table updated with the new default and rationale
* `CHANGELOG.md`: entry under Unreleased / Changed
* `tests/test_manifest.py`: regression test asserting the shipped default keeps both session scoping and surface attribution

## Testing

`python -m pytest tests/` in `hindsight-integrations/claude-code`: 182 passed (includes the new regression test).


nicoloboschi (changes requested):
let's not change the default. tags have downstream effect on observation scopes and I'd prefer to not have a breaking change for users unless clearly better UX

PathKnower:
> let's not change the default. tags have downstream effect on observation scopes and I'd prefer to not have a breaking change for users unless clearly better UX

Thanks for the pointer, that's a fair call. You're right that the tag isn't cosmetic here - in combined mode the observation scope is keyed off the full tag set, so changing the shipped default would shift consolidation scoping for everyone, not just add a recall filter. Closing this; I'll keep the surface tag as a local opt-in (~/.hindsight/claude-code.json) rather than a default. Appreciate you taking the time

PR #2127, opened 2026-06-11
fix(api): cancel abandoned recall & reflect via cooperative cancellation token

### LINKED ISSUE #2122: Recall computation is not cancelled on client disconnect — abandoned requests accumulate and starve the instance (v0.7.1)
### Bug Description

Following up on #1887 (sorry, wouldn't let me reopen it).

Recall requests that outlive their client are never cancelled. When a client
times out and drops the connection, the recall task keeps running to completion
server-side (~2 CPUs for 60–95 s on our 11.6K-unit bank). A client with a
45-second timeout that retries on failure creates an unbounded pile of zombie
recalls: we observed process RSS grow to **7.2 GB** and CPU pinned at the
container cap (`cpus: '2'`) **for hours**, triggered by a single user's
auto-recall retry loop. With `HINDSIGHT_API_RECALL_MAX_CONCURRENT` defaulting
to 32, one impatient client can stack 32 × 2-CPU computations whose results
nobody will ever read.

A second, compounding issue: **consolidation's internal recalls contend with
user recalls.** During consolidation we logged internal recall phases of
**195–580 s per memory** (`[CONSOLIDATION] ... recall=582.811s`), which
monopolize the CPU cap and push concurrent user recalls past their client
timeouts — which is exactly what ignites the retry spiral above. One
consolidation op took 899 s to process a single memory.

This reproduced the operator-visible symptoms of our previous incident
(sustained 450–550% CPU, VPS provider resource warnings, recall timeouts) with
a completely different root cause. (Follow-up to the entity-link issue closed
as resolved-by-redesign — happy to confirm that fix is holding in production:
0 entity rows, `memory_links` stable at 116 MB.)

**Environment:**
- Hindsight v0.7.1 (Docker), external PostgreSQL with pgvector
- 6 banks, ~22,000 `memory_units` (largest bank: 11,617)
- 16 GB RAM multi-core VPS; hindsight-api capped at `cpus: '2'`, 12 GB mem
- Reranker: flashrank, `RERANKER_MAX_CANDIDATES=50`

### Steps to Reproduce

1. Run Hindsight 0.7.1 with a bank large enough that recall latency exceeds
   ~45 s under CPU constraint (ours: 11.6K memory units, `cpus: '2'`).

2. Issue a recall and abandon it before completion (any client timeout):
   ```bash
   timeout 45 curl -s -X POST -H "Authorization: Bearer $KEY" \
     -d '{"query":"test","limit":3}' \
     http://localhost:8888/v1/default/banks/$BANK/memories/recall
   ```

3. Observe that after the client disconnects, the server keeps computing:
   ```bash
   docker stats hindsight-api   # CPU stays pinned at the cap for the full
                                # recall duration despite no connected client
   ```
   `pg_stat_activity` shows no active queries during the spin — the work is
   in-process (graph expansion / rerank), so even `DB_COMMAND_TIMEOUT` cannot
   bound it.

4. Repeat step 2 in a loop (i.e., what any auto-recall integration with a
   timeout-and-retry policy does naturally). Each iteration adds another
   zombie task. RSS and CPU grow monotonically; recall latency for new
   requests increases as zombies contend, guaranteeing further timeouts.

5. Optionally, let consolidation run concurrently: its internal recalls
   (195–580 s each in our logs) occupy the CPU cap and push user recalls past
   their client timeouts, igniting the same spiral without any client
   misbehavior.

### Expected Behavior

- When a client disconnects, the associated recall task is cancelled and its
  CPU/memory reclaimed (FastAPI/Starlette exposes `request.is_disconnected()`;
  asyncio task cancellation can propagate to the retrieval stages).
- At minimum: abandoned-request cancellation available as a configuration
  option, and documentation warning that `RECALL_MAX_CONCURRENT` (default 32)
  bounds *concurrent* but not *useful* work.
- Consolidation's internal recalls should be deprioritized relative to
  interactive recalls (QoS/nice), or schedulable into maintenance windows, so
  background work cannot push foreground requests past their timeouts.

### Actual Behavior

- Recall tasks run to completion regardless of client state. Each abandoned
  request burns ~2 CPUs for 60–95 s (our bank size) producing output nobody
  reads.
- Under a routine timeout-and-retry client policy, zombies accumulate toward
  the `RECALL_MAX_CONCURRENT=32` default. Observed steady state: CPU pinned at
  the container cap for hours, RSS 7.2 GB, VPS provider issuing resource
  warnings — from one user.
- Consolidation amplifies the effect: its internal recalls (logged at
  195–580 s/memory) saturate the CPU cap, pushing interactive recalls past
  client timeouts even before any zombies exist.
- Mitigations we applied operationally: raised client timeouts above worst-case
  recall latency, lowered `RECALL_MAX_CONCURRENT` to 4, `VACUUM ANALYZE`
  (95 s → 60 s). These contain the spiral but don't fix the underlying
  no-cancellation behavior.

**Side question:** `"explain": true` on `/memories/recall` returns
`"trace": {}` on 0.7.1 — is a debug flag or log level required to populate the
per-phase timings? That trace would have cut our diagnosis time substantially,
and we'd use it to report properly on why recall latency tripled (~20 s → ~60 s)
while the bank grew only ~10%.

### Version

0.7.1

### LLM Provider

None

nicoloboschi:
### Summary

Cancel abandoned HTTP **recall and reflect** work via a cooperative cancellation token carried on `RequestContext`, instead of relying on asyncio task cancellation. Alternative to #2125 with an architecture that stops the most expensive stage and generalizes across endpoints.

Fixes #2122.

### Problem

A `/memories/recall` (or `/reflect`) that outlives its client kept running server-side — ~2 CPUs for 60–95 s per abandoned recall on an 11.6K-unit bank — and abandoned requests accumulated toward `RECALL_MAX_CONCURRENT` until the instance starved (RSS 7.2 GB, CPU pinned at the container cap for hours, from a single retry-looping client). See #2122.

### Why a token, not asyncio task cancellation

The recall pipeline's heaviest stage — cross-encoder reranking (flashrank) — runs in a `ThreadPoolExecutor` via `run_in_executor`. Cancelling the awaiting asyncio task only unblocks the `await`; **the worker thread keeps burning CPU to completion**. So task cancellation (the approach in #2125) returns `499` and frees the recall semaphore while an uncancellable thread still pins a CPU — `RECALL_MAX_CONCURRENT` stops bounding actual CPU.

Instead, a `CancellationToken` is threaded through `RequestContext` (already passed into every engine operation) and checked at **stage / iteration boundaries**, bailing out *before* dispatching the next expensive stage.

### Changes

- New `hindsight_api/cancellation.py`: `CancellationToken` (one-shot, awaitable, idempotent) + `OperationCancelledError`.
- `RequestContext` gains a `cancellation` field and a `raise_if_cancelled()` helper (no-op when no token attached).
- **Recall** checkpoints in `recall_async` / `_search_with_retries`: pre-retrieval, **pre-rerank**, and pre-enrichment (chunk/entity/source-fact fetches).
- **Reflect** checkpoints: an early check in `reflect_async`, plus a `cancel_check` hook in `run_reflect_agent` invoked between agent iterations (and the nested recall tool already checks at its own boundaries since it shares the `RequestContext`).
- **Shared HTTP wiring**: `run_cancellable_on_disconnect()` attaches the disconnect-driven token (1 s poll, restores the prior token on exit) and maps `OperationCancelledError` → **499**. Both the recall and reflect handlers call it — no duplicated try/except.
- Scoped to HTTP recall + reflect: internal callers (consolidation, MCP, mental-model triggers) pass no token, so every checkpoint is a no-op and behavior is unchanged.

### Honest limitation

Cooperative: it cannot interrupt a computation **already inside** a worker thread. If the client disconnects mid-rerank, we finish that one rerank then abort at the next checkpoint. The win is that an abandoned request never *starts* rerank (if the client left during retrieval), never runs post-rerank enrichment, and a reflect loop stops between iterations instead of spending every remaining LLM round-trip — and `RECALL_MAX_CONCURRENT` keeps bounding real work. Truly bounding mid-rerank would need chunked token checks inside the predict loop or a killable subprocess (out of scope). A deadline-based driver (for the consolidation-contention half of #2122) is a trivial extension since the token lives on `RequestContext`.

### Relationship to #2125

Alternative implementation of the same fix. #2125 wraps recall in a disconnect watcher and cancels the asyncio task; this PR uses a checkpoint token. The token approach stops the threadpool rerank stage from starting (which task cancellation can't reach), avoids early-releasing the concurrency slot while a thread runs on, reuses a carrier every engine operation already receives, and **covers reflect** with the same mechanism. Pick one.

### Tests

```bash
uv run pytest tests/test_recall_cancellation.py -q   # 13 tests, all green
uv run ruff check hindsight_api/ && uv run ty check hindsight_api/api/http.py
```

Covers the token primitive, `RequestContext` integration, the disconnect driver (attach/restore/abort), the shared `run_cancellable_on_disconnect` helper (result-on-connected, 499-on-disconnect), and an ASGI end-to-end (disconnect → aborted staged work → 499). Existing reflect suites (`test_reflect_agent`, `test_per_operation_llm_config`, etc.) remain green — `cancel_check` defaults to `None`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -6,10 +6,12 @@ the FastAPI application with all API endpoints.
 """
 
 import asyncio
+import contextlib
 import json
 import logging
 import re
 import uuid
+from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 from datetime import datetime, timezone
 from typing import Any, Literal
@@ -17,6 +19,7 @@ from typing import Any, Literal
 from fastapi import Depends, FastAPI, File, Form, Header, HTTPException, Query, Request, UploadFile
 from fastapi.middleware.gzip import GZipMiddleware
 
+from hindsight_api.cancellation import CancellationToken, OperationCancelledError
 from hindsight_api.engine.audit import (
     AuditEntry,
     AuditLogger,
@@ -89,6 +92,49 @@ from hindsight_api.metrics import create_metrics_collector, get_metrics_collecto
 from hindsight_api.models import RequestContext
 
 logger = logging.getLogger(__name__)
+
+# Starlette surfaces client disconnects through an async receive check. A
+# one-second poll stops abandoned long-running recalls without busy-waiting.
+_CLIENT_DISCONNECT_POLL_INTERVAL_SECONDS = 1.0
+# 499 is the de facto reverse-proxy status for "client closed request".
+_CLIENT_CLOSED_REQUEST_STATUS_CODE = 499
+_CLIENT_DISCONNECTED_REASON = "client disconnected"
+
+
+@asynccontextmanager
+async def _cancel_on_client_disconnect(
+    http_request: Request,
+    request_context: RequestContext,
+    *,
+    poll_interval: float = _CLIENT_DISCONNECT_POLL_INTERVAL_SECONDS,
+) -> AsyncIterator[CancellationToken]:
+    """Attach a cancellation token that fires when the HTTP client disconnects.
+
+    The engine checks ``request_context`` at each pipeline stage boundary and
+    raises ``OperationCancelledError`` once this token fires, so an abandoned
+    request stops consuming CPU/DB resources instead of running to completion
+    (issue #2122). A background task polls ``request.is_disconnected()``; the
+    token is restored to its previous value on exit so nested scopes compose.
+    """
+    token = CancellationToken()
+    previous_token = request_context.cancellation
+    request_context.cancellation = token
+
+    async def _watch() -> None:
+        while True:
+            if await http_request.is_disconnected():
+                token.cancel(_CLIENT_DISCONNECTED_REASON)
+                return
+            await asyncio.sleep(poll_interval)
+
+    watcher = asyncio.create_task(_watch())
+    try:
+        yield token
+    finally:
+        watcher.cancel()
+        with contextlib.suppress(asyncio.CancelledError):
+            await watcher
+        request_context.cancellation = previous_token
 
 
 class EntityIncludeOptions(BaseModel):
@@ -3465,6 +3511,7 @@ def _register_routes(app: FastAPI):
     async def api_recall(
         bank_id: str,
         request: RecallRequest,
+        http_request: Request,
         request_context: RequestContext = Depends(get_request_context),
         _precheck: None = Depends(precheck_for("recall")),
     ):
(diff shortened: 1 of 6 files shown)


Sanderhoff-alt:
I agree this is the better implementation. The cooperative cancellation token fits the recall pipeline better than my outer asyncio-task cancellation approach, especially around threadpool reranking.

I'll close my PR in favor of this one.

PR #2139, opened 2026-06-11
feat(retain): make structured-unit chunk limit configurable

### LINKED ISSUE #2136: Allow customizable chunk overflow factor or max chunk size
### Use Case

Followup to #2113: I am ingesting jsonl messages, and some of them can be significantly larger but still comfortably within the extraction model's capable context size.

### Problem Statement

Currently, longer messages will still be split, which loses context such as time, role, filename, etc. If I want to prevent this, I can only increase the chunk size.

I'm not opposed to just increasing the chunk size if you think that's a better solution and won't harm extraction quality, though it doesn't seem like it should normally be necessary.

Either a customizable multiplier or something like `HINDSIGHT_API_RETAIN_MAX_CHUNK_SIZE`. This would allow defaulting to a smaller chunk size wherever possible but increasing it for json/jsonl to ensure longer messages stay within a single chunk. I think this is reasonable especially for models like ds4 flash that support a much longer context than the default chunk size.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Priority

Nice to have

### Additional Context

_No response_

### Checklist

Sanderhoff-alt:
## Summary

Adds `retain_structured_chunk_size` as an explicit retain chunking configuration for structured inputs such as JSONL and conversation arrays.

When unset, structured chunks now follow the effective `retain_chunk_size` instead of using the hidden `retain_chunk_size * 1.5` overflow allowance. Operators can tune structured inputs independently by setting `retain_structured_chunk_size` explicitly.

Closes #2136.

## Motivation

The previous structured chunking behavior allowed a single JSONL line or conversation turn to exceed `retain_chunk_size` by an implicit 1.5x factor. That made the effective size limit difficult to reason about and impossible to tune directly.

This PR replaces that hidden heuristic with a named configuration field:

- Default behavior stays simple: if `retain_structured_chunk_size` is unset, the structured chunk size equals the effective `retain_chunk_size`.
- Smaller structured caps are supported: set `retain_structured_chunk_size` below `retain_chunk_size` to split very large JSONL lines or conversation turns sooner.
- Larger structured caps are also supported: set `retain_structured_chunk_size` above `retain_chunk_size` to preserve larger structured records whole.
- Validation rejects non-positive or non-integer chunk sizes while keeping the existing `retain_max_completion_tokens > retain_chunk_size` check unchanged.

## Changes

- Added `HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE` / `retain_structured_chunk_size`.
- Threaded the setting through retain extraction, streaming retain, append/prepend chunking, and retain strategy application.
- Validated resolved global, tenant, bank, and retain strategy configs so chunk size values are checked after hierarchy and strategy overrides are applied.
- Updated bank config API models, bank templates, MCP tool docs, generated OpenAPI artifacts, and generated clients.
- Updated the control-plane retain configuration UI to expose the new field for default retain settings and named strategies.
- Updated the Rust CLI `bank set-config` command to expose `--retain-chunk-size` and `--retain-structured-chunk-size`.
- Updated `.env.example` files and generated docs/skill references.

## Testing

- `git diff --check`
- `./scripts/hooks/lint.sh`
- `cd hindsight-dev && uv run cli-coverage-check`
- Commit hook: `generate-docs-skill.sh`
- Commit hook: `lint.sh`

Additional targeted coverage was added for:

- JSONL and conversation chunking with explicit and default structured chunk sizes.
- Config validation for positive chunk sizes and smaller structured caps.
- Hierarchical config resolution, including bank overrides, null tombstones, and strategy overrides.
- Bank template configurable fields.
- MCP tool configuration documentation.
- Python client payload construction for clearing the structured chunk override.
- Control-plane retain strategy serialization for inherited, explicit-null, and numeric structured chunk sizes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -47,6 +47,9 @@ HINDSIGHT_API_LLM_BASE_URL=https://api.openai.com/v1
 HINDSIGHT_API_HOST=0.0.0.0
 HINDSIGHT_API_PORT=8888
 HINDSIGHT_API_LOG_LEVEL=info
+# Optional retain chunking override for structured logs/transcripts.
+# Unset uses HINDSIGHT_API_RETAIN_CHUNK_SIZE as the structured-chunk limit.
+# HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE=
 
 # Base Path / Reverse Proxy Support (Optional)
 # Set these when deploying behind a reverse proxy with path-based routing
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1149,7 +1149,14 @@ class CreateBankRequest(BaseModel):
     )
     retain_chunk_size: int | None = Field(
         default=None,
-        description="Maximum token size for each content chunk during retain.",
+        description="Target maximum characters for each content chunk during retain.",
+    )
+    retain_structured_chunk_size: int | None = Field(
+        default=None,
+        description=(
+            "Maximum characters for a single JSONL line or conversation turn to keep whole during retain. "
+            "Defaults to retain_chunk_size when unset."
+        ),
     )
     enable_observations: bool | None = Field(
         default=None,
@@ -1189,6 +1196,7 @@ class CreateBankRequest(BaseModel):
             "retain_extraction_mode",
             "retain_custom_instructions",
             "retain_chunk_size",
+            "retain_structured_chunk_size",
             "enable_observations",
             "observations_mission",
         ):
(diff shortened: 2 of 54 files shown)


nicoloboschi:
One naming suggestion: `retain_structured_unit_max_chars` / `HINDSIGHT_API_RETAIN_STRUCTURED_UNIT_MAX_CHARS` is a bit verbose and breaks from the existing `retain_chunk_size` convention.

Consider **`retain_structured_chunk_size`** (`HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE`):
- Keeps the important "structured-only" signal (applies to JSONL lines / conversation turns, not plain text).
- Mirrors the sibling `retain_chunk_size` by using `...chunk_size` rather than `..._unit_max_chars`, so the pair reads naturally (`retain_chunk_size` + `retain_structured_chunk_size`).

Purely a naming nit — the approach and the absolute-cap semantics look good.

Sanderhoff-alt:
@nicoloboschi 
Thanks for the suggestion. I agree the retain_structured_chunk_size naming fits better with the existing retain_chunk_size convention while still making the structured-only behavior clear.

I’ve updated the PR to use retain_structured_chunk_size / HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE and --retain-structured-chunk-size throughout the API, config, CLI, clients, docs, and control-plane UI. I kept the existing description semantics focused on the actual behavior: the value controls the maximum size for a single JSONL line or conversation turn to stay whole, and defaults to retain_chunk_size when unset.

nicoloboschi (approved):
Verified locally: fork PR so CI test-api/Core-LLM were skipped (no secrets). Ran the affected suites locally — test_chunking/test_config_validation/test_hierarchical_config/test_bank_template_configurable_fields (114 passed), test_mcp_tools (179), python client payload (2), control-plane retain-strategy-config/sdk-response (16). Generated artifacts in sync. Clean refactor; LGTM.

PR #2141, opened 2026-06-11
feat(retain): configurable chunk overflow factor (#2136)

### LINKED ISSUE #2136: Allow customizable chunk overflow factor or max chunk size

nicoloboschi:
Closes #2136.

## What

Makes the chunk **overflow factor** configurable. Previously hardcoded at `1.5`, this factor controls how far a single JSONL line or conversation turn may exceed the chunk-size budget while still being kept whole (before falling back to splitting it as text).

New flag: **`HINDSIGHT_API_RETAIN_CHUNK_OVERFLOW_FACTOR`** (default `1.5`, hierarchical / per-bank overridable).

## Why

Per the issue: when ingesting JSONL where some messages are larger but still comfortably within the extraction model's context, the only way to avoid splitting them today is to raise the global chunk size. Raising the overflow factor instead (e.g. to `3.0`) keeps longer structured messages intact in a single chunk — preserving role/time/filename context — while keeping a smaller default chunk size everywhere else.

## How

- `config.py` — env constant, default, dataclass field, `_CONFIGURABLE_FIELDS` entry, `from_env()` wiring, and `_parse_chunk_overflow_factor()` (validates `>= 1.0`, fails fast otherwise).
- `fact_extraction.py` — `chunk_text` / `_chunk_conversation` / `_chunk_jsonl` take an `overflow_factor` param (default = the existing module constant). The three bank-config-driven call sites pass `config.retain_chunk_overflow_factor`.
- `orchestrator.py` — both `chunk_text` call sites thread the resolved factor through.
- `memory_engine.py` — `_resolve_retain_chunk_size` → `_resolve_retain_chunk_params`, returning a `RetainChunkParams` dataclass (chunk size + factor) so the chunk-offset counting matches the orchestrator's actual chunking when an oversized unit is present.
- Docs row added to `configuration.md` (+ regenerated docs skill).

## Tests

- `test_chunking.py` — a raised `overflow_factor` keeps a JSONL line / conversation turn whole that the default would split.
- `test_config_validation.py` — default, env override, and `< 1.0` rejection.
- `test_hierarchical_config.py` — new field is configurable; count assertion 38 → 39.

All affected non-DB tests pass; lint + ty clean.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -396,6 +396,7 @@ ENV_LLM_PROMPT_CACHE_ENABLED = "HINDSIGHT_API_LLM_PROMPT_CACHE_ENABLED"
 # Retain settings
 ENV_RETAIN_MAX_COMPLETION_TOKENS = "HINDSIGHT_API_RETAIN_MAX_COMPLETION_TOKENS"
 ENV_RETAIN_CHUNK_SIZE = "HINDSIGHT_API_RETAIN_CHUNK_SIZE"
+ENV_RETAIN_CHUNK_OVERFLOW_FACTOR = "HINDSIGHT_API_RETAIN_CHUNK_OVERFLOW_FACTOR"
 ENV_RETAIN_EXTRACT_CAUSAL_LINKS = "HINDSIGHT_API_RETAIN_EXTRACT_CAUSAL_LINKS"
 ENV_RETAIN_EXTRACTION_MODE = "HINDSIGHT_API_RETAIN_EXTRACTION_MODE"
 ENV_RETAIN_MISSION = "HINDSIGHT_API_RETAIN_MISSION"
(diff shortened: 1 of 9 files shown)


nicoloboschi:
Closing in favor of #2139, which already closes #2136 with a more complete implementation (typed bank-template API, OpenAPI, generated clients, control-plane UI, CLI). #2139 uses an absolute char cap (`retain_structured_unit_max_chars`) rather than this PR's multiplier — the issue offered both shapes and the absolute cap is easier to reason about.

PR #2163, opened 2026-06-12
fix(consolidation): honor an observation scope limit of 0 (no new observations)

r266-tech:
### What

`observation_scope_limits` / `max_observations_per_scope` document the limit as:

> `-1 = unlimited, 0 = no new observations, >0 = hard cap`

…but a configured limit of **0** currently behaves like **unlimited**.

### Why

In `_process_memory_batch` the slot computation gates on `> 0`:

```python
max_obs = _effective_scope_limit(config, fact_tags)
remaining_observation_slots: int | None = None
if max_obs > 0 and fact_tags:          # 0 is excluded -> stays None
    ...
```

So for `max_obs == 0`, `remaining_observation_slots` stays `None`, and
`_build_response_model(None)` builds the *unconstrained* response model — the
scope is free to create unlimited new observations, the inverse of the
documented "no new observations". The capacity-note guard in
`_consolidate_batch_with_llm` had the same `> 0` defect, so the LLM was never
even told the scope was frozen.

The enforcement machinery already handles 0 correctly: `_build_response_model(0)`
forbids all creates (covered by `test_zero_limit_forbids_creates`) and the
create-truncation guard already uses `>= 0`. Only the two call-site guards were
blocking 0 from reaching it.

### Change

- `> 0` → `>= 0` at both call-site guards, matching the existing `>= 0`
  truncation guard.
- Short-circuit the `_count_observations_for_scope` query when the limit is 0
  (there are no slots regardless of the current count).
- Add `test_max_observations_per_scope_zero_forbids_all_creates`: with a scope
  cap of 0 and a mock LLM that tries to create one observation per fact,
  consolidation must create **zero** observations (this fails before the fix).

Caught while reading #2140.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -1489,8 +1489,14 @@ async def _process_memory_batch(
     # the bank-wide max_observations_per_scope for scopes matching its tag pattern.
     max_obs = _effective_scope_limit(config, fact_tags)
     remaining_observation_slots: int | None = None
-    if max_obs > 0 and fact_tags:
-        current_count = await _count_observations_for_scope(conn, bank_id, fact_tags)
+    if max_obs >= 0 and fact_tags:
+        # max_obs == 0 means "no new observations": there are no slots regardless
+        # of the current count, so skip the count query for that case.
+        current_count = (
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Correct, minimal, and well-tested. Verified the two `> 0` call-site guards were the only ones out of step with the existing `>= 0` truncation guard, and the regression test genuinely fails pre-fix. LGTM.

Issue #2106, opened 2026-06-10
Migration a7b8c9d0e1f2_split_history_into_own_tables fails on PostgreSQL when memory_units.bank_id exceeds 64 chars


vvlang:
### Summary

The PostgreSQL path of migration `a7b8c9d0e1f2_split_history_into_own_tables.py` defines `observation_history.bank_id` as `VARCHAR(64)`, but the source column `memory_units.bank_id` is `TEXT` (unbounded). When any existing row has a `bank_id` longer than 64 characters, the backfill `INSERT` aborts with `StringDataRightTruncation`, the migration transaction rolls back, and container startup fails permanently.

The Oracle path of the same migration uses `VARCHAR2(256)` for `bank_id`, so the two dialects are inconsistent.

### Environment

- Image: `ghcr.io/vectorize-io/hindsight:0.8.1` (revision `4dc149a`)
- Backend: PostgreSQL 17.10
- Alembic head before failure: `c1d2e3f4a5b6`
- Trigger: upgrading to a release that includes `a7b8c9d0e1f2`

### Error

```
sqlalchemy.exc.DataError: (psycopg2.errors.StringDataRightTruncation)
value too long for type character varying(64)

[SQL:
    INSERT INTO "public".observation_history (observation_id, bank_id, content, changed_at)
    SELECT mu.id, mu.bank_id,
           e - 'changed_at',
           COALESCE(NULLIF(e->>'changed_at', '')::timestamptz, now())
    FROM "public".memory_units mu
         CROSS JOIN LATERAL jsonb_array_elements(mu.history) WITH ORDINALITY a(e, ord)
    WHERE mu.fact_type = 'observation' ...
]
RuntimeError: Database migration failed
```

### Root cause

`memory_units.bank_id` is declared `TEXT` in `5a366d414dce_initial_schema.py` (and the same in `documents`, `entities`, `chunks`, `async_operations`, etc.). Only the legacy `mental_models.bank_id` and `directives.bank_id` are `VARCHAR(64)`.

The new history tables in `a7b8c9d0e1f2` follow the wrong precedent — `bank_id VARCHAR(64)` — so the backfill from `memory_units` is silently lossy in the schema sense and explicitly broken when any value exceeds 64 chars.

In our case the offending value was 78 characters and looked like:

```
<scope>::ou_<32hex>::ou_<32hex>
```

[...]

PR #2175, opened 2026-06-12
fix(db): widen remaining live bank_id columns to TEXT on PostgreSQL (#2106 follow-up)

r266-tech:
## Problem

Follow-up to #2110 (issue #2106). #2110 widened the two **history** tables (`observation_history`, `mental_model_history`) `bank_id` from `VARCHAR(64)` to `TEXT` because the narrow column bricked startup. The same `VARCHAR(64)` / `TEXT` inconsistency still affects the **live** tables that store a user-supplied `bank_id`:

| table | `bank_id` | created in |
|---|---|---|
| `directives` | `VARCHAR(64)` | `p1k2l3m4n5o6` |
| `mental_models` | `VARCHAR(64)` | `h3c4d5e6f7g8` (origin `pinned_reflections` in `n9i0j1k2l3m4`) |
| `mental_model_versions` | `VARCHAR(64)` | `o0j1k2l3m4n5` / `j5e6f7g8h9i0` |

`banks.bank_id` is `TEXT` (unbounded), so a deployment can create a bank whose id exceeds 64 chars — the 78-char hierarchical org-unit shape reported in #2106 — and the bank insert succeeds. The next write that propagates that id (`create_directive`, `create_mental_model` / consolidation, or mental-model versioning) then aborts with:

```
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(64)
```

i.e. a 500 on core write endpoints, instead of the startup brick #2110 already repaired.

## Change

A forward migration (`down_revision = c3e5a7b9d1f4`, the current head) that widens the three columns to `TEXT`, mirroring `c3e5a7b9d1f4`:

- `ALTER COLUMN ... TYPE TEXT` is a no-op on columns already `TEXT`, so every upgrade path converges. `varchar(64) → text` is binary-coercible in PostgreSQL (no table rewrite, foreign keys preserved).
- Per-tenant, schema-prefixed, same search-path-aware mechanism as `c3e5a7b9d1f4`.
- PostgreSQL only — these tables are created by PostgreSQL-only migrations; on Oracle they are absent / already `VARCHAR2(256)` (consistent), so the Oracle slot is intentionally absent, exactly as in `c3e5a7b9d1f4`.
- Downgrade is a no-op (narrowing back could truncate real data).

## Test

`tests/test_migration_remaining_bank_id_text.py` (mirrors `test_migration_history_long_bank_id.py`): migrates a dedicated `pg0` instance to head, asserts all three `bank_id` columns are `TEXT`, then round-trips a >64-char `bank_id` through every widened table — including `mental_model_versions`, whose composite FK `(mental_model_id, bank_id) → mental_models(id, bank_id)` exercises the widened columns on both sides.

## Scope notes

- `learnings.bank_id` is also `VARCHAR(64)`, but it has no runtime `INSERT` path that carries a user-supplied `bank_id`, so it is left out to keep the change to the reachable surface (happy to include it if you'd prefer the column types fully uniform).

### DIFF
--- a/hindsight-api-slim/hindsight_api/alembic/versions/a1d3f5b7c9e2_widen_remaining_bank_id_to_text.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/a1d3f5b7c9e2_widen_remaining_bank_id_to_text.py
@@ -1,0 +1,80 @@
+"""Repair: widen the remaining live ``bank_id`` columns from VARCHAR(64) to TEXT on PostgreSQL.
+
+Follow-up to ``c3e5a7b9d1f4`` (issue #2106), which widened the two *history*
+tables (``observation_history``, ``mental_model_history``) to ``TEXT`` after the
+narrow ``VARCHAR(64)`` declaration bricked startup. The same VARCHAR(64) / TEXT
+inconsistency still affects the live tables that store a user-supplied
+``bank_id``:
+
+* ``directives``            -- created VARCHAR(64) in ``p1k2l3m4n5o6``
+* ``mental_models``         -- VARCHAR(64) (origin ``pinned_reflections`` in
+                               ``n9i0j1k2l3m4``; recreated in ``h3c4d5e6f7g8``)
+* ``mental_model_versions`` -- created VARCHAR(64) in ``o0j1k2l3m4n5`` /
+                               ``j5e6f7g8h9i0``
+
+``banks.bank_id`` is ``TEXT`` (unbounded), so a deployment can create a bank
+whose id exceeds 64 chars -- the 78-char hierarchical org-unit shape reported in
+issue #2106 -- and the bank insert succeeds. The next write that propagates that
+id (``create_directive``, ``create_mental_model`` / consolidation, or
+mental-model versioning) then aborts with::
+
+    psycopg2.errors.StringDataRightTruncation: value too long for type
+    character varying(64)
+
+i.e. a 500 on core write endpoints, instead of the startup brick that
+``c3e5a7b9d1f4`` already repaired.
+
+``ALTER COLUMN ... TYPE TEXT`` is a no-op on a column that is already ``TEXT``,
+so every upgrade path converges on ``TEXT``. These tables are per-tenant (they
+live in each tenant schema, not ``public``), so this runs for every migrated
+schema via the search-path-aware prefix -- the same mechanism as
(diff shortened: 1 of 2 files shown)


nicoloboschi:
## Root cause + correction (pushed as `e3d664dab`)

The fix is right for **two** of the three tables, but the third would have **bricked API startup** on every deployment — worse than the bug it repairs. I pushed a corrected commit.

### The problem
The migration issued:
```sql
ALTER TABLE mental_model_versions ALTER COLUMN bank_id TYPE TEXT
```
but `mental_model_versions` **does not exist at head**. It's created in `j5e6f7g8h9i0`, then dropped in `o0j1k2l3m4n5._pg_upgrade()`:
```python
# 4. Drop the mental_model_versions table (no longer used)
op.execute(f"DROP TABLE IF EXISTS {schema}mental_model_versions CASCADE")
```
…and only re-created in that migration's *downgrade*. Nothing recreates it on the upgrade path, and a repo-wide grep finds **zero** runtime/ORM references — it's a fully retired table.

Verified empirically on a clean pg0 migrated to head `c3e5a7b9d1f4`:
```
mental_model_versions present at head?  NO (count=0)
  directives.bank_id     = character varying   ← real bug
  mental_models.bank_id  = character varying   ← real bug
```
Because migrations run inside the lifespan-startup transaction, the `ALTER` raises `UndefinedTable`, the whole migration rolls back, and the API never comes up — exactly the failure class #2106/#2110 were about.

### Where the description went wrong
The PR conflated **`mental_model_versions`** (retired, dropped) with **`mental_model_history`** — the live table created by #2007 (`a7b8c9d0e1f2`) and already widened to `TEXT` by `c3e5a7b9d1f4`. They're different tables with similar names. The claimed composite FK `(mental_model_id, bank_id) → mental_models(id, bank_id)` also doesn't exist — the only definition (the downgrade re-create) is a standalone table with `model_id VARCHAR(128)` and no such FK.

### The correction
- Migration widens only `directives` and `mental_models` (the live `VARCHAR(64)` tables that exist at head); docstring documents why `mental_model_versions` is excluded.
- Test drops the `mental_model_versions` assertion + FK insert, and fixes the `mental_models` insert (it omitted NOT NULL `source_query`/`content` and used an invalid `subtype`).

The test now actually passes — previously it could never pass, since the head migration crashed before any assertion:
```
2 passed in 3.51s
```

PR #2177, opened 2026-06-12
fix(control-plane): add missing /observations/scopes endpoint

### LINKED ISSUE #2176: Control Plane UI polls GET /api/banks/{bank_id}/observations/scopes every ~5s — endpoint does not exist, returns 404
## Summary

The Hindsight Control Plane UI (v0.8.1) sends a periodic GET to an endpoint that does not exist in the open-source Hindsight API. Every ~5 seconds the UI re-fires the request, producing 100+ 404s in the browser console during a normal session and cluttering server logs.

## Environment

- Control Plane: `@vectorize-io/hindsight-control-plane` v0.8.1, port 9999
- Dataplane: `hindsight-api` v0.8.1, port 8888
- Banks in use: `main`, `lapo73`, `claude_code` (single-org deployment)
- Browser console: the same 404 loop is reproducible on any bank

## Steps to reproduce

1. Open the Control Plane UI at `http://<host>:9999`
2. Select any memory bank (e.g. `main`)
3. Open the **Observations** view
4. Open DevTools → Network / Console
5. Observe the failing call:

```
GET http://<host>:9999/api/banks/main/observations/scopes 404 (Not Found)
```

The request fires on mount, on bank switch, and on a setInterval in `1cld2c0a065s-.js`. The UI keeps retrying silently — the page is usable, but the scope filter dropdown in the Observations view never populates and the console is flooded with retries.

## Root cause

The UI client calls `fetchApi("/api/banks/{bank_id}/observations/scopes")` (function `listObservationScopes` in `hindsight-control-plane/src/lib/api/observations.ts` or equivalent), but the corresponding Next.js route handler is missing:

- The path `hindsight-control-plane/src/app/api/banks/[bank_id]/observations/scopes/route.ts` does **not** exist in the open-source repo (404 on the main branch).
- The underlying dataplane also has no equivalent: `GET /openapi.json` on port 8888 has no `/v1/default/banks/{bank_id}/observations/scopes` path. The closest related route is `DELETE /v1/default/banks/{bank_id}/observations` (clears all observations).

So the UI is calling a route that is neither implemented in the Next.js control plane nor proxied to the dataplane.

## Expected behaviour

The `observations/scopes` endpoint should return the list of distinct **observation scopes** (i.e. the tag sets used on observation memories in the bank) so the UI can populate the scope filter dropdown on the Observations page. The internal concept already exists — see issue #1625 ("Targeted consolidation API: trigger consolidation for specific observation scopes only"), where the author notes that consolidation already groups memories by tag sets internally (the `_count_observations_for_scope` helper). Surfacing that distinct list is the missing piece.

## Proposed fix

Two layers need to be in sync:

### 1. Add the route to the dataplane (Hindsight API)

Add to `hindsight-api/openapi.json` and implement in the routers:

```http
GET /v1/default/banks/{bank_id}/observations/scopes
→ 200 [{ "tags": ["user:alice"] }, { "tags": ["user:bob"] }, { "tags": [] }, ...]
```

Implementation can reuse the same scope-grouping logic that consolidation already uses internally — `SELECT DISTINCT tags FROM observations WHERE bank_id = $1 AND fact_type = 'observation'` (or the equivalent ORM query). The response is small (one entry per distinct tag set), no pagination needed.

### 2. Add the proxy route in the Control Plane (Next.js)

Create `hindsight-control-plane/src/app/api/banks/[bank_id]/observations/scopes/route.ts`:

```ts
import { NextRequest, NextResponse } from "next/server";

export async function GET(
  req: NextRequest,
  { params }: { params: { bank_id: string } }
) {
  const dataplane = process.env.HINDSIGHT_API_URL ?? "http://localhost:8888";
  const r = await fetch(
    `${dataplane}/v1/default/banks/${params.bank_id}/observations/scopes`,
    { headers: forwardAuthHeaders(req), cache: "no-store" }
  );
  return new NextResponse(r.body, { status: r.status, headers: r.headers });
}
```

This matches the existing proxy pattern used by the other bank-scoped routes in `hindsight-control-plane/src/app/api/banks/[bank_id]/.../route.ts`.

## Workarounds for users hitting this now

- Until the fix lands, the page is still usable — the failing request is non-fatal and the rest of the UI (memories, recall, reflect, documents) works. Just ignore the 404s in the console, or block the URL in DevTools.
- The scope filter on Observations stays empty; the only way to see all observations is to leave the filter on "All scopes" (which is the default).

## Impact

- Severity: low — feature degradation (broken scope filter), not data loss
- Frequency: continuous — fires every ~5s, every page mount, every bank switch
- Blast radius: every Control Plane user on v0.8.1

Happy to send a PR for the route + proxy if helpful.

xdemocle:
## Summary

The Control Plane UI polls `GET /api/banks/{bank_id}/observations/scopes` every ~5 seconds to populate the scope filter on the Observations view. The route was never implemented in either the Next.js control plane or the underlying Hindsight dataplane, producing a 404 loop in the browser console and clogging server logs.

This PR implements the endpoint end-to-end in two layers.

### Dataplane (`hindsight-api-slim`)
- **New `MemoryEngine.list_observation_scopes()`** — returns the distinct tag sets used on observation memories (the effective scope in all three supported modes: `per_tag`, `all_combinations`, `combined`). The empty `[]` scope is included for observations with no tags.
- **New route** `GET /v1/default/banks/{bank_id}/observations/scopes` with a matching `ListObservationScopesResponse` model in the OpenAPI spec.
- Auth/authz/rate-limit/error handling follow the existing `/tags` endpoint pattern.

### Control plane (`hindsight-control-plane`)
- **New Next.js proxy** at `/api/banks/[bankId]/observations/scopes/route.ts` that forwards the optional `limit` query param to the dataplane and localises any errors.

## Verification

Tested end-to-end with the local Hindsight instance (banks `main`, `lapo73`, `claude_code`):

```bash
$ curl http://localhost:9999/api/banks/main/observations/scopes
{"scopes":[{"tags":[]}]}

$ curl http://localhost:9999/api/banks/claude_code/observations/scopes
{"scopes":[
  {"tags":["5b8bd321-4087-4a51-9dca-8dc82ee01463"]},
  {"tags":["bbecfa20-d8e4-4404-ba42-fed2ddbd4268"]},
  ...
]}
```

Negative test (`?limit=foo`) returns 422 with localised error. Non-existent banks return 200 with empty scopes (consistent with `/observations` and `/tags` behavior).

## Fixes

Closes #2176

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1330,6 +1330,41 @@ class ListTagsResponse(BaseModel):
     offset: int
 
 
+class ObservationScopeItem(BaseModel):
+    """A single observation scope, represented by the tag list that defines it.
+
+    An empty ``tags`` list corresponds to the "global" scope — observations
+    with no tags. Otherwise the scope is the set of tags used on the
+    observation's memory.
+    """
+
+    tags: list[str] = Field(
+        default_factory=list,
+        description=(
+            "The list of tags that define this scope. An empty list represents "
+            "the global (no-tag) scope."
+        ),
+    )
+
+
+class ListObservationScopesResponse(BaseModel):
+    """Response model for the list observation scopes endpoint."""
+
+    model_config = ConfigDict(
+        json_schema_extra={
+            "example": {
+                "scopes": [
+                    {"tags": []},
+                    {"tags": ["user:alice"]},
+                    {"tags": ["user:alice", "team:engineering"]},
+                ],
+            }
+        }
+    )
+
+    scopes: list[ObservationScopeItem]
+
+
 class DocumentResponse(BaseModel):
     """Response model for get document endpoint."""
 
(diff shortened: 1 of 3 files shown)


nicoloboschi:
this was already shipped in 0.8.2 please upgrade

PR #2180, opened 2026-06-12
feat(composio): add Composio integration (Hindsight memory as custom tools)

benfrank241:
Adds a **Composio** integration: Hindsight memory exposed as Composio in-process custom tools.

## What it does

`register_hindsight_tools(composio, ...)` returns Composio custom tools for **retain / recall / reflect**. The Hindsight bank for each call is the Composio session's `user_id`, so a single registered tool set **isolates memory per user automatically**. Also ships `memory_instructions()` to pre-recall memories for system-prompt injection (Composio doesn't auto-inject context).

```python
from composio import Composio
from hindsight_composio import register_hindsight_tools

composio = Composio()
tools = register_hindsight_tools(composio, api_key="hsk_...")
session = composio.create(
    user_id="user-123",                       # becomes the Hindsight bank_id
    experimental={"custom_tools": tools},
)
```

## Contents
- `hindsight_composio/` — `tools.py` (3 tools + `memory_instructions`), `config.py` (dataclass config + `HINDSIGHT_API_KEY` env fallback), `errors.py`, typed Pydantic inputs.
- `tests/` — **50 tests** using a `FakeComposio` (mirrors the real `experimental.tool` decorator + `SessionContext`) and a mocked Hindsight client, exercising the actual framework wiring (per-`user_id` bank resolution, tool registration, error paths).

## Completeness (per CLAUDE.md)
- ✅ Tests exercise the framework (FakeComposio + mocked client) — 50 pass
- ✅ CI job `test-composio-integration` (uv build → sync → ruff → pytest) + `detect-changes` output + path filter
- ✅ `composio` in `VALID_INTEGRATIONS` (`release-integration.sh`)
- ✅ Gallery card + doc page + official Composio icon (`/img/icons/composio.png`)
- ✅ ruff clean, ruff-format clean, `uv lock --check` consistent

> Touches `.github/**`, so the full CI matrix runs; the gating job here is `test-composio-integration`.


DK09876 (commented):
Reviewed and ran the full suite locally (checked out the branch, `uv sync`):

- ✅ 58 tests pass, `ruff check` clean, `ruff format` clean, `uv build` ok, `uv lock --check` consistent — matches CI.
- Verified the decorator usage against the real `composio` source: `experimental.tool` is overloaded to support **both** the bare (`tool(fn)`) and parameterized (`tool()`) forms, so the bare usage here is correct. The `(input, ctx)` signature, `ctx.user_id`, dict returns, and `experimental={"custom_tools": [...]}` registration all match the real API.
- 👍 Nice that `TestRealComposioRegistration` exercises a real `Composio()` instance (asserts the `CustomTool` slugs + `input_params` schema) — a good guard against the experimental API drifting.

A few minor, non-blocking suggestions:

1. **No tool-invocation / real-Hindsight E2E.** The real-Composio tests cover registration + input schema, and the logic tests use a mocked client — but nothing exercises an actual tool *call* through Composio's execution path or against a live Hindsight server (no `requires_real_llm` bucket like autogen/continue/llamaindex have). Optional, but it would close the loop on the end-to-end path.
2. **Tighten config types.** `budget`, `recall_tags_match`, and `memory_instructions(tags_match=...)` are typed `str`; the autogen/continue integrations use `Literal["low", "mid", "high"]` / `Literal["any", "all", "any_strict", "all_strict"]`. Cheap type-safety + consistency win.
3. **Parameterize bare generics.** `register_hindsight_tools(...) -> list` and `tools: list` could be `list[Any]` (or the Composio `CustomTool` type).
4. **`_ensure_bank` swallows all exceptions.** Fine since `create_bank` is idempotent, but a `logger.debug` there would help surface a real auth/network failure that currently only shows up later on `retain`.

Optional: Composio supports `output_params` (a Pydantic output schema). The tools return bare dicts (which Composio accepts), but a typed output model would give the agent a real output schema.

Overall this is clean and well-tested — looks good to me modulo the minor notes above. 🚀

DK09876 (approved):
Re-reviewed at `e63e8fe` — all four points addressed, and I re-ran the suite locally (not just trusting CI):

✅ 58 deterministic tests pass, 3 E2E correctly deselected (`-m "not requires_real_llm"`), `ruff check`/`format` clean, `uv build` ok, `uv lock --check` consistent.

Point-by-point:
1. **Real-Hindsight E2E** — `test_e2e.py` drives retain/recall/reflect through the actual `(input, ctx)` tool-call path against a live server, with a health-check skip, async-aware recall polling, and per-bank cleanup; gated `requires_real_llm` and excluded from PR CI. Together with `TestRealComposioRegistration` (real decorator + input schema) that covers both ends — registration *and* execution. 👍
2. **Literal config types** — `Budget` / `TagsMatch` aliases applied across `config.py`, `tools.py`, and `memory_instructions`. Matches autogen/continue.
3. **Typed generics** — `-> list[Any]` / `tools: list[Any]`.
4. **`_ensure_bank` debug log** — the swallowed `create_bank` failure is now logged at debug, so a real auth/network error is visible rather than only surfacing later on `retain`.

Optional FYI (not a request): the E2E constructs `Hindsight(base_url=...)` without an api_key, so it targets a local/open server — which matches the docstring's intent. Pointing it at an authenticated Cloud endpoint would need the key threaded in, but that's fine as-is.

LGTM — approving. 🚀

PR #2189, opened 2026-06-14
fix(consolidation): make dedup tag matching configurable

### LINKED ISSUE #2188: Dedup probe tag matching hardcoded to all_strict — causes runaway observation duplication when agents use per-session tags
**Use Case**

We run Hindsight as the memory layer for Hermes Agent (by Nous Research) — a single-user, single-bank deployment. Hermes uses per-session tags on `retain()` calls for provenance and session filtering. Every conversation, subagent delegation, and scheduled cron job gets a unique session ID tagged onto the retained facts (e.g. `session:hm-cron_<hex>_<timestamp>`).

**Problem Statement**

The creation-time dedup probe (`_dedup_adjudicate`, consolidator.py:172) and the consolidation recall (`_find_related_observations`, consolidator.py:1892) both hardcode:

```python
tags_match = "all_strict" if tags else "any"
```

When an agent uses tags for session provenance, this prevents the dedup guard from ever finding near-duplicates from different sessions. Each new session's consolidation probes with `all_strict` against its own unique session tag and finds zero existing observations — so it creates a new one every time.

In our deployment this produced **211 near-identical observations** of the same fact (a skill's governance instructions re-injected into context each cron run), each from a different 2-hourly run over ~6 weeks. Cosine similarities between pairs range 0.78–0.96. The LLM adjudicator would merge them instantly, but the probe never surfaces them.

**How This Feature Would Help**

A configurable tag match mode for the dedup probe would let single-user / single-bank deployments avoid runaway observation duplication without losing session-level filtering on recall. We'd set it to `any` and the dedup would see across session boundaries, while multi-tenant deployments keep `all_strict` as the default.

**Proposed Solution**

Make the dedup probe's tag match mode independently configurable, separate from the consolidation recall:

```python
# config.py
DEFAULT_CONSOLIDATION_DEDUP_TAG_MATCH = "all_strict"
ENV_CONSOLIDATION_DEDUP_TAG_MATCH = "HINDSIGHT_API_CONSOLIDATION_DEDUP_TAG_MATCH"
```

```python
# consolidator.py line 172 — dedup probe only
tags_match = config.consolidation_dedup_tag_match if tags else "any"
```

Consolidation recall (line 1892) stays hardcoded `all_strict` — that's the real security boundary where cross-tenant observation text is shown to the LLM. The dedup probe only decides whether to skip a CREATE, so widening it is a quality concern, not a security one.

**Alternatives Considered**

- **Container patch** (change line 172 to `"any"`): works but breaks on every `docker pull`.
- **Drop session tags from the agent side**: session-level filtering is useful for recall scoping and debugging. Tags are the right mechanism for provenance — the issue is that Hindsight treats *any* tag as a tenant boundary.
- **Manual dedup script**: we built one and cleaned up 2,300+ historical duplicates, but this is ongoing maintenance, not a fix.

**Priority**

**Additional Context**

- Hindsight v0.8.x (self-hosted Docker, single bank, single user)
- Hermes Agent `plugins/memory/hindsight` — session tagging at `__init__.py:1492`
- The two `all_strict` call sites serve different security profiles. Consolidation recall (line 1892) exposes observation text cross-tenant if widened — that's a real leak. The dedup probe (line 172) only returns a boolean merge/keep — no cross-tenant data is exposed.

**Checklist**

koriyoshi2041:
## Summary
- add HINDSIGHT_API_CONSOLIDATION_DEDUP_TAG_MATCH for the near-duplicate observation probe
- keep the default at all_strict so current tag isolation behavior is unchanged
- document the single-user/session-tag use case and keep consolidation recall strict

Fixes #2188.

## Tests
- uv run pytest tests/test_consolidation_dedup.py tests/test_config_validation.py
- uv run ruff check hindsight_api/config.py hindsight_api/engine/consolidation/consolidator.py tests/test_consolidation_dedup.py tests/test_config_validation.py
- git diff --check


koriyoshi2041:
Local verification after the latest upstream fetch:

- `cd hindsight-api-slim && uv run pytest tests/test_config_validation.py -q` -> 48 passed
- `cd hindsight-api-slim && uv run pytest tests/test_consolidation_dedup.py -q` -> 19 passed

The branch is also current with `origin/main` in my local worktree. The new setting only changes the semantic dedup probe scope; consolidation recall context remains strict-scoped as before.

koriyoshi2041:
Makes sense to close this in favor of #2202. I agree the `shared` scope keeps recall and the dedup probe aligned instead of widening only the probe path, and it fits the root cause better than an env-level override. I will leave this branch closed and follow that direction.

PR #2190, opened 2026-06-14
config: add consolidation_dedup_tag_match for cross-scope dedup

### LINKED ISSUE #2188: Dedup probe tag matching hardcoded to all_strict — causes runaway observation duplication when agents use per-session tags

kevin-ho:
Closes #2188

Adds HINDSIGHT_API_CONSOLIDATION_DEDUP_TAG_MATCH env var to make the dedup probe's tag matching mode independently configurable.

**What changed:**
- config.py: new env var constant, default all_strict, validation function, config field
- consolidator.py line 173: dedup probe uses config value instead of hardcoded all_strict
- Consolidation recall (line 1981) stays hardcoded all_strict — security boundary intact

**Validation:**
- Invalid values log a warning and fall back to all_strict
- Valid: any, all, any_strict, all_strict, exact

**Default behavior unchanged.** Single-user deployments set HINDSIGHT_API_CONSOLIDATION_DEDUP_TAG_MATCH=any to enable cross-scope dedup.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -442,6 +442,7 @@ ENV_CONSOLIDATION_BATCH_SIZE = "HINDSIGHT_API_CONSOLIDATION_BATCH_SIZE"
 ENV_CONSOLIDATION_MAX_MEMORIES_PER_ROUND = "HINDSIGHT_API_CONSOLIDATION_MAX_MEMORIES_PER_ROUND"
 ENV_CONSOLIDATION_LLM_BATCH_SIZE = "HINDSIGHT_API_CONSOLIDATION_LLM_BATCH_SIZE"
 ENV_CONSOLIDATION_DEDUP_THRESHOLD = "HINDSIGHT_API_CONSOLIDATION_DEDUP_THRESHOLD"
+ENV_CONSOLIDATION_DEDUP_TAG_MATCH = "HINDSIGHT_API_CONSOLIDATION_DEDUP_TAG_MATCH"
 ENV_CONSOLIDATION_LLM_PARALLELISM = "HINDSIGHT_API_CONSOLIDATION_LLM_PARALLELISM"
 ENV_CONSOLIDATION_MAX_TOKENS = "HINDSIGHT_API_CONSOLIDATION_MAX_TOKENS"
 ENV_CONSOLIDATION_MAX_COMPLETION_TOKENS = "HINDSIGHT_API_CONSOLIDATION_MAX_COMPLETION_TOKENS"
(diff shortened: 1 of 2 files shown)


kevin-ho:
Closing in favor of #2189, which covers the same fix with tests and docs. Thanks @koriyoshi2041 for the quick turnaround.

Issue #850, opened 2026-04-02
Memory leak: bank_id as metric label causes unbounded OTel histogram growth (~3 GB after 15 days)


sireika:
### Bug Description

`MetricsCollector.record_operation()` in `hindsight_api/metrics.py` (line ~333) includes `bank_id` as an attribute in OpenTelemetry histogram and counter recordings. Since `bank_id` is a per-user value (e.g., user-123), every unique user creates a permanent, never-evicted time series in the OTel SDK's in-memory aggregation buffers.

Over time this causes unbounded memory growth proportional to unique_users × operations × budgets × statuses.

### Steps to Reproduce

1. Run `hindsight-api` with default configuration (metrics enabled)
2. Issue recall/retain/reflect requests for multiple distinct `bank_id` values
3. Observe memory growth via `vmmap --summary <pid>` (macOS) or `/proc/<pid>/smaps` (Linux)

After 15 days of normal usage on a dev instance with ~50 users, we observed:

-  Physical footprint: 3.1 GB (peak 3.2 GB)
-  MALLOC_SMALL: 1.7 GB virtual, 1.6 GB dirty, 15.2 million allocations
-  RSS reported by ps was only ~15 MB because macOS compressed the allocations, which also caused PM2's `max_memory_restart` (RSS-based) to never trigger

**Root cause**

```
# hindsight_api/metrics.py, MetricsCollector.record_operation()
  attributes = {
      "operation": operation,
      "bank_id": bank_id,      # <-- HIGH CARDINALITY: one series per user
      "source": source,
      "tenant": _get_tenant(),
  }
```
The OTel SDK's ExplicitBucketHistogramAggregation stores a full bucket array per unique attribute set. With default 16 histogram buckets, each unique `{operation, bank_id, source, tenant, budget, max_tokens, success}` tuple allocates ~400 bytes that are never freed. The combinatorial explosion creates millions of allocations.

**Suggested fix**

Remove `bank_id` from metric attributes. It belongs in tracing spans (which are exported and evicted), not in metrics (which accumulate in process for the lifetime of the SDK).

```
  attributes = {
      "operation": operation,
      # bank_id removed — high-cardinality label causes unbounded memory growth
      "source": source,
      "tenant": _get_tenant(),
  }
```

If per bank observability is needed in metrics, perhaps consider a bounded approach like hashing bank_id into a small number of buckets (for example, `bank_id_bucket: str(hash(bank_id) % 64)`).

**Environment**

[...]

PR #2191, opened 2026-06-14
fix: template bank-id path segment in HTTP metric endpoint label (follow-up to #850)

sireika:
## Summary

`http_metrics_middleware` (`hindsight_api/api/http.py`) labels each request with a normalized path, but only templates UUIDs and **pure-numeric** segments:

```python
path = re.sub(r"/[0-9a-f]{8}-...-[0-9a-f]{12}", "/{id}", path)  # UUIDs
path = re.sub(r"/\d+(?=/|$)", "/{id}", path)                    # numeric only
```

Bank-scoped routes are all `/v1/{tenant}/banks/{bank_id}/...`, so a bank id that is **not** a bare integer or UUID (e.g. `user-123`, `tenant-acme`, `org_42`, slug ids) survives in the `endpoint` label. Each distinct bank then creates a never-evicted OTel series — unbounded process-memory growth, on every per-bank request (it fires on every HTTP call, not only recorded operations).

This is the same unbounded-OTel-cardinality issue as #850 (fixed in #898 for the `record_operation` `bank_id` **attribute**), via a second code path #898 didn't touch: the HTTP request **path** label. Unlike the attribute, there is currently no config switch that bounds it.

## Change

- Extract endpoint normalization into a pure, unit-tested `normalize_http_endpoint()` in `metrics.py`, next to `get_token_bucket` (its low-cardinality-label sibling).
- Additionally template the `/banks/<id>` segment to `{bank_id}` for any bank id (the existing UUID/numeric templating moves into the helper unchanged).
- Call it from `http_metrics_middleware`.

## Test

`tests/test_metrics.py::TestNormalizeHttpEndpoint` covers non-numeric bank ids, UUIDs, numeric ids, list/non-bank paths, and unchanged-path cases.

## Alternatives to consider

Should the path label be always-templated (this PR), or gated behind the existing `HINDSIGHT_API_METRICS_INCLUDE_BANK_ID` flag for symmetry with #898?

A more thorough alternative is labeling by the matched route template (`route.matches(request.scope)`, as `unknown_params_middleware` already does), which would also bound other non-UUID/non-numeric id segments (document ids, etc.). Happy to adjust to whichever you prefer.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -88,7 +88,12 @@ from hindsight_api.engine.providers.none_llm import LLMNotAvailableError
 from hindsight_api.engine.response_models import VALID_RECALL_FACT_TYPES, MemoryFact, TokenUsage
 from hindsight_api.engine.search.tags import TagGroup, TagsMatch
 from hindsight_api.extensions import HttpExtension, OperationValidationError, load_extension
-from hindsight_api.metrics import create_metrics_collector, get_metrics_collector, initialize_metrics
+from hindsight_api.metrics import (
+    create_metrics_collector,
+    get_metrics_collector,
+    initialize_metrics,
+    normalize_http_endpoint,
+)
 from hindsight_api.models import RequestContext
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
LGTM. Correct, minimal fix for a real second code path of the #850/#898 unbounded-cardinality issue. Verified the analysis and ran the new test locally (passes).

On the open question: keep it always-templated (don't gate behind HINDSIGHT_API_METRICS_INCLUDE_BANK_ID) — that flag exists to optionally expose bank_id as a deliberate metric *dimension*; the endpoint path label has no such purpose. Tenant and document-id segments are also technically unbounded but fine as a follow-up.

PR #2212, opened 2026-06-15
fix(control-plane): expose "shared" observation scope in the Add Document UI

r266-tech:
## Summary

#2202 added the **"shared"** observation-scopes mode and synced it across the server engine, the HTTP model, retain types, and every client — including the control-plane client type (`api.ts`). But the control-plane **GUI** component (`bank-selector.tsx`) was missed, so a user **cannot select "shared" from the Add-Document form** even though the API and all clients already accept it. ("shared" is the supported root-cause fix for the observation-duplication behavior.)

## Change — single file, `bank-selector.tsx`

- Add `"shared"` to the `docObservationScopes` state union, the inline retain-item type, and the `onValueChange` cast.
- Add `<SelectItem value="shared">Shared</SelectItem>` to the mode dropdown — a plain literal, consistent with the existing "Combined"/"Custom" items, so **no locale-file changes**.
- Send `observation_scopes: "shared"` in the retain request (the client type already accepts it).
- `computeScopes`: `"shared"` returns a single global scope `[[]]`, matching the server's `parsed === "shared" -> [[]]` handling. The preview renders a dedicated "one shared, untagged scope" line rather than routing an empty scope through the tag-based preview helpers (which are malformed for `[]`).

No i18n/locale or generated-file changes; existing modes (per_tag / combined / all_combinations / custom) are unchanged.

If this is already being handled elsewhere, feel free to close.


nicoloboschi (approved):

nicoloboschi (changes requested):
thanks. don't use hardcoded string. this check is failing - Run npm run i18n:check --workspace=hindsight-control-plane

> @vectorize-io/hindsight-control-plane@0.8.2 i18n:check
> tsx scripts/find-untranslated.ts


src/components/bank-selector.tsx
  src/components/bank-selector.tsx:1234:31  text: "All observations are stored in one shared, untagged scope (tags are ignored)."

1 suspicious string across 1 file.
npm error Lifecycle script `i18n:check` failed with error:
npm error code 1
npm error path /home/runner/work/hindsight/hindsight/hindsight-control-plane
npm error workspace @vectorize-io/hindsight-control-plane@0.8.2
npm error location /home/runner/work/hindsight/hindsight/hindsight-control-plane
npm error command failed
npm error command sh -c tsx scripts/find-untranslated.ts

r266-tech:
Fixed in 71e03e5f.

What changed:
- Replaced the new shared-scope dropdown label/help text with `tAddDocument(...)` messages.
- Also moved the neighboring `Custom` option into the same `addDocument` message namespace, since Codex caught that it would otherwise leave the selector mixed translated/English.
- Added the new keys across all 10 locale catalogs.

Checks run:
- `npm run i18n:check --workspace=hindsight-control-plane`
- locale key parity check across all message catalogs
- `git diff --check`
- `npm test --workspace=hindsight-control-plane -- tests/messages/messages.test.ts` (20 passed; Vitest printed the existing config-resolution warning but exited 0)

nicoloboschi (approved):

PR #2219, opened 2026-06-16
Fix archived memory embedding dimension migration

### LINKED ISSUE #2209: invalidated_memory_units not included in embedding dimension migration — invalidating memories fails after switching embedding models
## Description

When switching embedding models (e.g., from 384-dim `bge-small-en-v1.5` to 1536-dim `text-embedding-3-small`), the `_migrate_table_embedding_dimension` function in `migrations.py` only migrates two tables:

```python
_migrate_table_embedding_dimension(conn, schema_name, "memory_units", required_dimension, vector_ext)
_migrate_table_embedding_dimension(conn, schema_name, "mental_models", required_dimension, vector_ext)
```

The `invalidated_memory_units` table (introduced in #1976) is **not included** in this migration. As a result:

1. `memory_units` is correctly updated to the new dimension (e.g., 1536)
2. `invalidated_memory_units` remains at the old dimension (e.g., 384)
3. When a user tries to **invalidate** a memory via `PATCH /v1/default/banks/{bank}/memories/{id}`, the operation fails with:

```
asyncpg.exceptions.DataError: expected 384 dimensions, not 1536
```

The invalidation operation moves the row from `memory_units` → `invalidated_memory_units` via `INSERT … SELECT`, and the 1536-dimensional embedding cannot be inserted into the 384-dimensional column.

## Steps to Reproduce

1. Start Hindsight with a 384-dim embedding model (e.g., `bge-small-en-v1.5`)
2. Let it create memories (these get 384-dim embeddings)
3. Switch to a 1536-dim model (e.g., `text-embedding-3-small`) in `.env`
4. Restart Hindsight — migration updates `memory_units` and `mental_models` to 1536, but `invalidated_memory_units` stays at 384
5. Try to invalidate a memory: `PATCH /v1/default/banks/{bank}/memories/{id}` with `{"state": "invalidated"}`
6. **Actual:** 500 error — `expected 384 dimensions, not 1536`

## Expected Behavior

The embedding dimension migration should also cover `invalidated_memory_units` (and any other tables with an `embedding` column), so that invalidation/revert operations work correctly after switching embedding models.

## Suggested Fix

Add `invalidated_memory_units` to the migration list in `migrations.py`:

```python
_migrate_table_embedding_dimension(conn, schema_name, "memory_units", required_dimension, vector_ext)
_migrate_table_embedding_dimension(conn, schema_name, "mental_models", required_dimension, vector_ext)
_migrate_table_embedding_dimension(conn, schema_name, "invalidated_memory_units", required_dimension, vector_ext)
```

Alternatively, dynamically discover all tables with an `embedding` column rather than hardcoding table names.

## Environment

- Hindsight version: latest (Docker image `registry.cn-hangzhou.aliyuncs.com/renhai/hindsight:latest`)
- Database: embedded pg0 (PostgreSQL 18.1)
- Embedding model switch: 384-dim → 1536-dim (`text-embedding-3-small`)
- Deployment: Docker, single container with embedded pg0

## Workaround

Manually alter the column dimension (safe when the table is empty):

```sql
ALTER TABLE invalidated_memory_units ALTER COLUMN embedding TYPE vector(1536);
-- Recreate HNSW index if needed
CREATE INDEX idx_invalidated_memory_units_embedding 
ON invalidated_memory_units USING hnsw (embedding vector_cosine_ops);
```

koriyoshi2041:
Problem
- Switching embedding dimensions updates memory_units and mental_models, but leaves invalidated_memory_units at the old vector dimension.
- After that, invalidating a memory can fail when the row moves into the archive table.

Fix
- Include invalidated_memory_units in ensure_embedding_dimension.
- Add a preflight over all embedding tables before ALTERs so a populated table blocks the migration before any table is partially changed.
- Add regression coverage for empty archive-table migration and populated archive-table blocking.

Test
- uv run ruff format hindsight_api/migrations.py tests/test_custom_embedding_dimension.py
- uv run ruff check hindsight_api/migrations.py tests/test_custom_embedding_dimension.py
- uv run pytest tests/test_custom_embedding_dimension.py::TestEmbeddingDimension -q -k "not local_embeddings_dimension_detection"

Risk
- Low: this uses the existing single-table migration helper and only adds the archive table to the same dimension sync path.
- Full TestEmbeddingDimension also runs the local embedding model test; locally that unrelated test hit a missing socksio dependency for the configured SOCKS proxy.

Closes #2209

### DIFF
--- a/hindsight-api-slim/hindsight_api/migrations.py
+++ b/hindsight-api-slim/hindsight_api/migrations.py
@@ -565,7 +565,8 @@ def ensure_embedding_dimension(
     """
     Ensure the embedding column dimension matches the model's dimension for all tables.
 
-    Checks and adjusts memory_units.embedding and mental_models.embedding:
+    Checks and adjusts memory_units.embedding, invalidated_memory_units.embedding,
+    and mental_models.embedding:
     - If dimensions match: no action needed
     - If dimensions differ and table is empty: ALTER COLUMN to new dimension
     - If dimensions differ and table has data: raise error with migration guidance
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Closing this in favor of #2210.

I missed that PR when I opened this one. The #2210 approach is cleaner because the archive is cold storage and does not need to keep embeddings at all; recomputing on revert avoids keeping another vector column in the dimension-migration surface.

I did a quick follow-up pass on #2210 while deciding this:
- the new migration shape check passes locally for d4f6a8c2e1b3_null_archive_embeddings.py
- the red test-api shard appears unrelated to this change: it fails while starting pg0 for tests/test_migration_remaining_bank_id_text.py with an existing postgres pid on port 5568, after 996 tests had already passed
- the red CLI doc example is reflect.sh timing out on Gemini tool calls, also not connected to the archive embedding change

Thanks for the better fix.

PR #2221, opened 2026-06-16
docs: fix stale documentation links

Sanderhoff-alt:
## Summary
- update stale docs.hindsight.vectorize.io/docs links to current docs paths
- point open-source integration and installation docs to hindsight.vectorize.io

## Verification
- checked every new URL returns HTTP 200
- confirmed canonical URLs and page titles match expected docs pages
- git diff --check


nicoloboschi (approved):
thanks!

PR #2224, opened 2026-06-16
fix(litellm): hard-cap completions with asyncio.wait_for so a hung call can't block forever

kubaodias:
## Problem

The LiteLLM provider issues completions as a bare `await self._acompletion(...)`. The only timeout is the `timeout=` kwarg handed to `litellm.acompletion()`, which is **not always honored** (e.g. a connection held open with no token progress). When that happens the coroutine awaits indefinitely, holding the worker slot and a concurrency-semaphore permit for the lifetime of the process.

Fact extraction fans these calls out through `asyncio.gather`, so a **single hung straggler stalls the entire operation** even though its sibling calls returned. The observable symptom is confusing: per-call `slow llm call` log lines keep appearing (siblings succeeding) while operation completion throughput collapses to zero and stuck ops sit at `...retain_extract_facts+structured` with a `stage_age` that grows 1:1 with wall-clock and never resets.

## Fix

- Wrap the request in `asyncio.wait_for(timeout=self.timeout)` in both `call` and `call_with_tools` (mirroring the Gemini provider, which already does this).
- Treat the resulting `TimeoutError` as a normal retryable attempt, so the task can retry or fail cleanly and release its slot. The existing `asyncio.gather(..., return_exceptions=True)` callers absorb the timeout with no extra handling.
- **Converge timeout handling:** litellm's own `Timeout` and the outer `wait_for` `TimeoutError` are armed at the same deadline, so catch both in the same except block. Otherwise which one trips is a race that flows through two different paths (dedicated vs generic) with different log lines and backoff. The exception class name is logged so the firing mechanism stays visible.
- **Make the cap configurable** by reading `HINDSIGHT_API_LLM_TIMEOUT` (default `DEFAULT_LLM_TIMEOUT` = **120s**) in `LiteLLMLLM.__init__` — the same one-line pattern `OpenAICompatibleLLM` already uses. A `None` timeout resolves to that env/default, never `None`, so the `wait_for` cap is always bounded.

### Scope

The hard `asyncio.wait_for` backstop added here covers the **LiteLLM family only** (`litellm`, `litellmrouter`, `bedrock`, which all subclass/reuse `LiteLLMLLM`). Gemini already has its own 90s hard cap. Other providers (anthropic, openai-compatible) are unchanged.

## Tests

`tests/test_litellm_timeout.py`:
- a hung `_acompletion` is cancelled per attempt and raises `TimeoutError` with bounded wall-clock and the correct retry count (`call` and `call_with_tools`);
- an unset timeout resolves to the finite `DEFAULT_LLM_TIMEOUT` (120s).

All new tests pass; existing router/provider tests unaffected.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
@@ -19,6 +19,8 @@ import os
 import time
 from typing import Any
 
+from litellm.exceptions import Timeout as LiteLLMTimeout
+
 from hindsight_api.config import DEFAULT_LLM_TIMEOUT, ENV_LLM_TIMEOUT
 from hindsight_api.engine.llm_interface import LLMInterface, OutputTooLongError
 from hindsight_api.engine.response_models import LLMToolCall, LLMToolCallResult, TokenUsage
@@ -207,8 +209,6 @@ class LiteLLMLLM(LLMInterface):
                 },
             }
 
-        from litellm.exceptions import Timeout as LiteLLMTimeout
-
         last_exception = None
 
         for attempt in range(max_retries + 1):
@@ -376,8 +376,6 @@ class LiteLLMLLM(LLMInterface):
         call_kwargs = self._build_common_kwargs(messages, max_completion_tokens, temperature)
         call_kwargs["tools"] = tools
         call_kwargs["tool_choice"] = tool_choice
-
-        from litellm.exceptions import Timeout as LiteLLMTimeout
 
         last_exception = None
         for attempt in range(max_retries + 1):
--- a/hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
@@ -15,9 +15,11 @@ is handled automatically by LiteLLM.
 import asyncio
 import json
 import logging
+import os
 import time
 from typing import Any
 
+from hindsight_api.config import DEFAULT_LLM_TIMEOUT, ENV_LLM_TIMEOUT
 from hindsight_api.engine.llm_interface import LLMInterface, OutputTooLongError
 from hindsight_api.engine.response_models import LLMToolCall, LLMToolCallResult, TokenUsage
 from hindsight_api.metrics import get_metrics_collector
@@ -53,10 +55,9 @@ class LiteLLMLLM(LLMInterface):
         **kwargs: Any,
     ):
         super().__init__(provider, api_key, base_url, model, reasoning_effort, **kwargs)
-        # ``None`` means "use the default"; keep the historical 300s cap so the
-        # hard ``asyncio.wait_for`` backstop in ``call`` is always bounded even
-        # when no timeout is configured by the caller.
-        self.timeout = timeout if timeout is not None else 300.0
+        # ``None`` falls back to HINDSIGHT_API_LLM_TIMEOUT, then DEFAULT_LLM_TIMEOUT — never None,
+        # so the hard ``asyncio.wait_for`` backstop in ``call`` is always bounded.
+        self.timeout = timeout if timeout is not None else float(os.getenv(ENV_LLM_TIMEOUT, str(DEFAULT_LLM_TIMEOUT)))
         self._litellm: Any = None
         # User-configured extra params merged as top-level kwargs into every
         # completion call so LiteLLM normalizes them per-provider (e.g. maps
@@ -206,6 +207,8 @@ class LiteLLMLLM(LLMInterface):
                 },
             }
 
+        from litellm.exceptions import Timeout as LiteLLMTimeout
+
         last_exception = None
 
         for attempt in range(max_retries + 1):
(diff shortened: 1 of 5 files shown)


hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
```diff
@@ -47,13 +49,15 @@ def __init__(
...
+        timeout: float | None = None,
         extra_body: dict[str, Any] | None = None,
         bedrock_service_tier: str | None = None,
         **kwargs: Any,
     ):
         super().__init__(provider, api_key, base_url, model, reasoning_effort, **kwargs)
-        self.timeout = timeout
+        # ``None`` falls back to HINDSIGHT_API_LLM_TIMEOUT, then DEFAULT_LLM_TIMEOUT — never None,
+        # so the hard ``asyncio.wait_for`` backstop in ``call`` is always bounded.
+        self.timeout = timeout if timeout is not None else float(os.getenv(ENV_LLM_TIMEOUT, str(DEFAULT_LLM_TIMEOUT)))
```
nicoloboschi, line 62: **Should fix — provider reads the env var directly, bypassing centralized config.**

`config.py` already resolves this exact value into `HindsightConfig.llm_timeout` (`llm_timeout=float(os.getenv(ENV_LLM_TIMEOUT, str(DEFAULT_LLM_TIMEOUT)))`). Re-reading the raw env var here means the setting is now sourced in two places, and per CLAUDE.md static config should come via `get_config()` / be threaded from `HindsightConfig`, not scattered `os.getenv` calls in engine code.

Concrete consequences: this constructor can't see a CLI/`main.py` override of `llm_timeout`, and it sidesteps the hierarchical (global → tenant → bank) resolution — a config-level override won't reach these providers, and tests that set config without touching the process env won't be honored.

Cleaner: thread the resolved timeout down from `create_llm_provider` / `get_config().llm_timeout` and drop the `os.getenv` fallback. That `OpenAICompatibleLLM` already does the same `os.getenv` read (`openai_compatible_llm.py:351`) isn't a reason to extend the pattern — ideally both converge on the config object.

hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
```diff
@@ -304,6 +313,25 @@ async def call(
                     logger.error(f"LiteLLM returned invalid JSON after {max_retries + 1} attempts")
                     raise
 
+            except (TimeoutError, asyncio.TimeoutError, LiteLLMTimeout) as e:
```
nicoloboschi, line 316: **Nit:** on Python 3.11+ `asyncio.TimeoutError` is an alias for the builtin `TimeoutError`, so `(TimeoutError, asyncio.TimeoutError, LiteLLMTimeout)` lists the same class twice. `(TimeoutError, LiteLLMTimeout)` is enough. Harmless either way (same applies to the matching handler in `call_with_tools` at line 460).

hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
```diff
@@ -203,13 +207,18 @@ async def call(
                 },
             }
 
+        from litellm.exceptions import Timeout as LiteLLMTimeout
```
nicoloboschi: **Nit:** `from litellm.exceptions import Timeout` is re-imported on every `call` / `call_with_tools` invocation (also at line 380). It's cached so cheap, but litellm is a hard dependency already imported in `__init__`, so hoisting this to a module-level import would be cleaner.

nicoloboschi (commented):
Reviewed via `/code-review`. Solid, well-scoped fix — the `asyncio.wait_for` backstop is the right call for litellm not honoring its own `timeout=`, the except-block ordering is correct (specific timeout handler precedes the generic `except Exception`, so it wins and the hung coroutine is actually cancelled), and the tests are deterministic monkeypatch tests (no real LLM, so the absence of `hs_llm_*` markers is correct). Lint passes locally.

No must-fix blockers. One should-fix (env coupling) plus a couple of nits inline.

**Two notes on the framing, not the code:**
- The "300s → 120s" change is really *removing a stale outlier*: `DEFAULT_LLM_TIMEOUT = 120.0` and `ENV_LLM_TIMEOUT` already existed and are what the docs advertise; the two LiteLLM signatures had a hardcoded `timeout=300.0` that ignored both. No caller passes `timeout=` (`create_llm_provider` doesn't accept it), so this is behaviorally safe and *more* consistent, not a new cap.
- The PR implies parity with `OpenAICompatibleLLM`, but that provider only hands `timeout` to the httpx/OpenAI client — it has **no `asyncio.wait_for` hard cap**. The same hang this PR fixes still exists there. Fine to leave out of scope, just worth not claiming full parity in the description.

hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
kubaodias: It was implemented this way in the [1st commit](https://github.com/vectorize-io/hindsight/pull/2224/changes/9ec2ca18eb23f48035732f703d0823fa2c739268) but then I've used Ponytail which suggested this change to follow the `OpenAICompatibleLLM` way. I can revert it if you'd prefer to.

hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
kubaodias: I deliberately left it this way although my PR reviewer flagged this too. This is kept separate in case `asyncio.TimeoutError` alias changes. Fine for me to change.

hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
kubaodias: Or apply the same to OpenAI provider but I didn't want to overgrow this PR.

nicoloboschi (approved):
Re-reviewed: core asyncio.wait_for hard-cap is correct, well-tested (3 new tests + 17 related pass), lint/type-clean. Prior nits resolved or intentionally declined. Approving.

PR #2226, opened 2026-06-16
fix(control-plane): add filter active message

oldschoola:
## Summary
- Add missing `dataView.filterActive` translation to every locale catalog.

## Tests
- `npx vitest run tests/messages/messages.test.ts`
- `npm run i18n:check`


nicoloboschi:
supersded by https://github.com/vectorize-io/hindsight/pull/2232

PR #2233, opened 2026-06-16
Arch Linux systemd packaging with dependency ordering

Raudbjorn:
## Summary

Packaging for Arch Linux with systemd service units, dependency ordering, and Arc A770 XPU support.

### Changes

- **hindsight-api.service**: 
  - Inline all env vars (fixes startup freeze caused by EnvironmentFile parsing issues)
  - , 
  -  for startup ordering
  -  +  — the working combo that allows PostgreSQL's  socket file while still hardening the service

- **hindsight-control-plane.service**:
  -  for dependency ordering
  -  
  -  (via env, Next.js standalone ignores most port env vars)

- **hindsight-worker.service**:
  - New service for async import processing
  - Depends on 

- **embeddings.py**: Added  detection for Arc A770 XPU embedding acceleration

### Deployment

The installed package at  provides:
- : API on   
- : CP on 
- : Async worker

Ports 6699 and 9966 are reachable over Tailscale at  and .

### Known limitations

- Arc A770 XPU for Ollama LLM is fragile when shared with the Wayland compositor (KWin). The GPU can deadlock under heavy concurrent use.
- Embeddings run on CPU (XPU torch causes iGPU initialization spin with  override).
-  on Arc A770 via Ollama Vulkan is slow (~30s+ per response); smaller models like  are more responsive.
- Control plane  endpoint returns 500 (calls  but API has ).
- Reflect operations can time out on large banks (54K+ pending ops).

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/embeddings.py
+++ b/hindsight-api-slim/hindsight_api/engine/embeddings.py
@@ -190,7 +190,7 @@ class LocalSTEmbeddings(Embeddings):
             device = "cpu"
             logger.info("Embeddings: forcing CPU mode")
         else:
-            # Check for GPU (CUDA) or Apple Silicon (MPS)
+            # Check for GPU (CUDA), Apple Silicon (MPS), or Intel XPU
             # Wrap in try-except to gracefully handle any device detection issues
             # (e.g., in CI environments or when PyTorch is built without GPU support)
             device = "cpu"  # Default to CPU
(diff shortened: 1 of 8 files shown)


nicoloboschi (changes requested):
this is a good example but I prefer to not commit it since it contains custom hindsight vars

can you remove the ollama config?

nicoloboschi:
Thanks for this @Raudbjorn — the Intel XPU support is genuinely useful and I've pulled it out into its own PR (#2260) so it can land independently with your authorship credited. I'm going to **close this PR** in favor of that split. Below is the reasoning and what would be needed to revisit the packaging side.

### Why split

This PR bundles two unrelated concerns: a product-code change (`embeddings.py` XPU detection) and Arch-specific distro packaging. The first is broadly useful and low-risk; the second is niche, contentious, and untested. Keeping them together means the good change waits on the debatable one.

### Packaging — concrete issues found

1. **Env var name mismatch.** `env.env.systemd` uses `HINDSIGHT_API_LLM_SKIP_VERIFICATION`, but the real config var is `HINDSIGHT_API_SKIP_LLM_VERIFICATION` (`config.py`). The service file gets it right, so the two disagree.
2. **Installed config file is never read.** PKGBUILD installs the env template to `/etc/hindsight/env` and the README tells users to edit it — but neither unit uses `EnvironmentFile=`; they inline all env. So editing `/etc/hindsight/env` has no effect. That's a confusing operator trap.
3. **`hindsight-worker.service` is added but never installed** — `package()` only installs the api and control-plane units.
4. **Wrong license** — PKGBUILD declares `ISC`; this repo is MIT.
5. **Points at a personal fork** — `url`/`source` reference `Raudbjorn/hindsight` instead of `vectorize-io/hindsight`.
6. **Personal infra baked in** — hardcoded ports 6699/9966 and Tailscale references are deployment-specific.
7. **Redundant version** — `pkgver=0.8.2` is hardcoded alongside a `pkgver()` that derives it from git tags.

### On distro packaging in-tree

Conventionally Arch packaging lives in the **AUR** rather than the upstream repo, since it's distro-specific and needs ongoing maintenance + CI to not bitrot. If the team does want first-party packaging here, the bar would be: upstream URL/source, correct license, config actually wired via `EnvironmentFile=` (or drop the file), all three units installed, and a CI job that at least lints the PKGBUILD.

Happy to look again if you'd like to reopen the packaging piece against that bar. Thanks again for the XPU contribution! 🙏

PR #2238, opened 2026-06-16
log(consolidation): show call count + avg for each timing phase

cdbartholomew:
## Problem

The consolidation summary log currently prints only the total time per phase:

\`\`\`
[4] Timing breakdown: recall=15.425s, llm=43.181s, embedding=0.301s
\`\`\`

This makes \`recall=15s\` easy to misread as a single slow query — when it is usually the sum of many sequential sub-calls (typically ~100 internal recalls at ~150ms each within one consolidation batch). The misreading sends operators down the wrong diagnostic path (\"the recall query is broken\") when the actual question is \"is per-call latency normal? is there a way to parallelize?\".

## Change

Add a \`timing_counts\` dict on \`ConsolidationPerfLog\` that runs in lockstep with \`timings\`, and surface it in the summary line when count > 1:

\`\`\`
[4] Timing breakdown: recall=15.425s (100 calls, avg=154ms),
                      llm=43.181s (12 calls, avg=3598ms),
                      embedding=0.301s (3 calls, avg=100ms),
                      db_write=0.829s
\`\`\`

Single-call timings keep the existing terse format (no \`(1 calls, ...)\` clutter).

## Backward compat

- \`timings\` dict and \`llm_calls\` keep their current semantics.
- \`timing_counts\` is a new attribute; nothing else reads it.
- \`merge_from()\` updated to fold counts from per-batch perf logs into the job-level one.

## Test plan

- [ ] No existing tests reference \`ConsolidationPerfLog\` directly, so behavior is unchanged outside the log line shape.
- [ ] CI green on this PR.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -640,6 +640,7 @@ class ConsolidationPerfLog:
         self.start_time = time.time()
         self.lines: list[str] = []
         self.timings: dict[str, float] = {}
+        self.timing_counts: dict[str, int] = {}
         self.llm_calls: int = 0
         self.total_obs_in_context: int = 0
         self.total_prompt_chars: int = 0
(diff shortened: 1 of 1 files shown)


nicoloboschi (approved):
LGTM — clean, minimal logging change. timing_counts tracks in lockstep with timings and folds correctly in merge_from(). Lint passes.

PR #2240, opened 2026-06-16
fix(consolidation): handle single-value source_fact_ids from LLM

haodonp:
Some LLMs return source_fact_ids as a string instead of a list when there is only one source ID. Add field_validator on both _CreateAction and _UpdateAction to auto-wrap into a single-element list.

Relates-to: https://github.com/vectorize-io/hindsight/issues/1566


nicoloboschi (approved):
LGTM. Sound, low-risk fix for scalar `source_fact_ids` from LLMs. The mode=before coercion prevents a single scalar field from dropping the entire consolidation batch (a failure mode that plausibly feeds #1566's duplicate-sibling symptom).

PR #2242, opened 2026-06-16
fix(mcp): omit reflect tool_trace/llm_trace from responses by default

yunanwg:
## Problem

The MCP `reflect` tool returns the full result of `reflect_async`, which includes `tool_trace` and `llm_trace` — the entire internal agent loop, including full mental-model text. On a real bank, a default reflect response measured:

| field | chars |
|---|---|
| `text` | 5,987 |
| `tool_trace` | 52,711 |
| **total** | **59,657** |

So every MCP `reflect` call silently spends tens of KB of the client's context window on a debug trace the caller didn't ask for. The REST `/reflect` API already omits the trace by default; only the MCP path returns it.

## Fix

Add a symmetric `include_trace: bool = False` flag, mirroring the existing `include_based_on` (which is already defaulted off "because broad reflections can exceed MCP client result limits" — the same rationale, more so). When not set, `tool_trace`/`llm_trace` are popped from the response. The trace stays available opt-in for debugging.

Applied to both the multi-bank and single-bank `reflect` registrations.

## Compatibility

Backwards-compatible: same fields available, just opt-in. No REST/OpenAPI/control-plane surface changes (the flag is MCP-tool-only; REST already omitted the trace).

## Testing

- `tests/test_mcp_tools.py::TestReflectTraceOmission` (new) — 6 cases (both bank modes × default-omitted / included-when-requested / independent-of-`include_based_on`). All green.
- `uvx ruff check` / `format --check` clean on the changed files.

### DIFF
--- a/hindsight-api-slim/hindsight_api/mcp_tools.py
+++ b/hindsight-api-slim/hindsight_api/mcp_tools.py
@@ -983,6 +983,7 @@ def _register_reflect(mcp: FastMCP, memory: MemoryEngine, config: MCPToolsConfig
             tags: list[str] | None = None,
             tags_match: str = "any",
             include_based_on: bool = False,
+            include_trace: bool = False,
             bank_id: str | None = None,
         ) -> str:
             """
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Verified the leak is real (engine unconditionally populates tool_trace/llm_trace on ReflectResult; MCP dumps the whole model) and that REST /reflect already omits the trace unless include.tool_calls is set — so this restores parity and mirrors the existing include_based_on flag. Tests cover both bank modes × default-omitted / opt-in / independent-of-include_based_on. Rebased on main (resolved the test-file append conflict with #2243); full test_mcp_tools.py green (192 passed), lint clean.

yunanwg:
thanks @nicoloboschi !

PR #2244, opened 2026-06-16
fix(backup): add bank backup and restore

oldschoola:
## Summary
- guard the shared pytest pg0 fixture so bare `pg0` aliases are redirected to `hindsight-test` instead of truncating the live local `hindsight` instance
- add whole-bank backup and same-bank restore endpoints, daily backup scheduling, retention settings, docs, and OpenAPI updates
- add control-plane backup actions plus a Backup tab, and regenerate the TypeScript SDK

## Tests
- `uv run pytest tests/test_pg0_fixture_safety.py tests/test_document_transfer.py tests/test_config_validation.py tests/test_maintenance_loop.py tests/test_bank_template_configurable_fields.py tests/test_hierarchical_config.py -q`
- `uv run pytest tests/test_document_transfer.py::test_http_bank_backup_restore_endpoints tests/test_document_transfer.py::test_import_bank_rejects_documents_archive -q`
- `uv run ty check hindsight_api/config.py hindsight_api/api/http.py hindsight_api/engine/memory_engine.py hindsight_api/engine/maintenance.py`
- `uv run pytest tests/test_env_template.py -q`
- `npx eslint src/app/[locale]/banks/[bankId]/page.tsx src/components/bank-backup-settings-view.tsx src/app/api/banks/[bankId]/backup/route.ts src/app/api/banks/[bankId]/restore/route.ts`
- `npx vitest run tests/messages/messages.test.ts`
- `npm run i18n:check`
- `npm run build` (in `hindsight-clients/typescript`)

## Notes
- `npx tsc --noEmit` in `hindsight-control-plane` still fails on two pre-existing unrelated diagnostics in `src/app/api/banks/[bankId]/health/llm/route.ts` and `src/app/api/list/route.ts`.
- `npm --prefix hindsight-docs run build` still fails on the pre-existing `check-integration-seo` frontmatter issues.
- Python/Go SDK regeneration was not completed locally because the Docker Linux engine was unavailable on this workstation; the TypeScript SDK was regenerated from the updated OpenAPI spec.


nicoloboschi:
Thanks so much for this — it's a genuinely thorough PR (endpoints, scheduling, control-plane UI, docs, i18n, SDK regen all wired up), and the care shows. I want to explain our thinking before closing it, because the reasoning matters more than the decision.

**Whole-bank backup/restore already exists via the admin CLI**, and we'd rather not maintain a second backup/restore surface (plus a scheduler) alongside it. Today:

```bash
# Back up a single bank to a portable archive
hindsight-admin export-bank --bank <bank_id> -o <bank_id>.zip
#   --schema <schema>        (tenant schemas)
#   --include-history        (audit_log + llm_requests)

# Restore: delete the bank (API / control plane), then import the archive
hindsight-admin import-bank --archive <bank_id>.zip
#   --target-bank <bank_id>  (restore under a different id)
#   --schema <schema>
```

`import-bank` intentionally refuses to overwrite a live bank, so restore-in-place is the explicit two-step above (delete, then import) — that keeps the destructive step a deliberate operator action rather than an implicit part of "restore."

For **scheduled** backups, a cron job or k8s CronJob calling `export-bank` and piping to durable storage (S3, a mounted volume) gives you the same outcome with retention/destination you control — without the API server writing to its own local (often ephemeral) filesystem. We'd rather document that pattern than ship a built-in scheduler.

So we're going to pass on this one — not for lack of quality, but to avoid a parallel mechanism we'd have to keep in sync.

**One thing worth salvaging:** the pytest pg0 fixture guard (redirecting a bare `pg0` / `pg0://hindsight` alias to the dedicated `hindsight-test` instance so the cleanup never truncates a developer's live local bank) is a real, standalone safety fix. If you're up for splitting that into its own small PR, I'd happily review and merge it on its own merits.

Really appreciate the effort here — thank you. 🙏

PR #2253, opened 2026-06-17
feat(metrics): instrument async worker completion path with operation metrics

kubaodias:
## Problem

`hindsight_operation_operations_total` / `hindsight_operation_duration_seconds` are only emitted from the **synchronous API layer** — `metrics.record_operation()` is called in `api/http.py` (recall, reflect, retain) and nowhere in the worker. In deployments where retain/consolidation (and the other worker task types) run through the **async worker** path (`POST … async=true` → 202 → worker), the result is:

- No `operation="retain"` (or consolidation, etc.) samples with a worker source, so the Operations dashboard shows no async activity.
- No Prometheus signal for async worker throughput or latency — exactly the path most likely to silently stall.

LLM-scoped metrics (`hindsight_llm_*`) keep emitting from the async path, but they don't distinguish "operations completing" from "operations wedged" — LLM sub-calls can keep succeeding while operations never finish.

## Fix

- **`MetricsCollector.record_operation_result()`** — a direct (non-context-manager) recording path with an explicit `success` label, for code that needs success control rather than the exception-based `record_operation()` context manager. The context manager now delegates to it, so there's a single place that builds attributes and writes to `operation_duration` / `operation_total` (cancellation handling from #2122 is preserved). Added to the base/no-op collectors too.
- **`WorkerPoller._execute_task_inner`** records `source="worker"` on terminal outcomes only. Deferrals (`DeferOperation`) and retries (`RetryTaskAt`) are **not** terminal and are deliberately not counted — that's the intake-vs-completion distinction needed to tell a healthy queue from a wedged one.
- Recording happens **outside** the executor's `try`/`except` (via a `terminal_success` flag) and is guarded by its own `try`/`except`, so a metrics-reporting failure can never be mistaken for a task failure and flip terminal state.
- **Label normalisation:** `batch_retain` / `file_convert_retain` collapse onto `operation="retain"` so worker completions share the API path's series. Other types pass through unchanged.

`bank_id` remains gated behind `metrics_include_bank_id` (cf. #850), so this adds no per-bank cardinality unless that flag is already enabled.

## Relationship to #1987

Complementary, not overlapping. #1987 adds queue-depth / backlog **gauges** (is the queue piling up / anything stuck?); this adds completion **counters** (is anything finishing?). Together they answer both halves of "is the worker healthy?". The only shared file is `metrics.py`, in different regions.

## Scope of the `success` label

The `success` label is a **completion-throughput** signal, not a failure-rate one. `success=false` means the task raised out to the poller (an unexpected error, or retry-exhausted). It does **not** capture deterministic failures that `memory_engine.execute_task` handles itself and returns from normally (`file_convert_retain`, non-retryable errors) — those record `success=true`. For authoritative failure visibility use the now-merged `hindsight_async_operations{status="failed"}` gauge (#1987), which reads each operation's final DB status. This PR (counters: is anything finishing?) and #1987 (gauges: is the queue piling up / anything stuck or failed?) are complementary.

## Tests

- `record_operation_result` direct recording (explicit success, `source="worker"`).
- Worker emits on completion and failure, and **not** on `DeferOperation`/`RetryTaskAt`; a self-handled (return-normally) failure records `success=true` by design (no-DB, `TestWorkerOperationMetrics`).
- Retain-variant label normalisation.
- Full `test_metrics.py` (incl. existing cancellation tests) + new worker tests pass (35); `ruff` clean.


hindsight-api-slim/hindsight_api/worker/poller.py
```diff
@@ -717,14 +742,28 @@ async def _execute_task_inner(self, task: ClaimedTask, holder: StageHolder | Non
                 task.task_dict["_schema"] = task.schema
             await self._executor(task.task_dict)
             logger.debug(f"Task {task.operation_id} execution finished")
+            terminal_success = True
```
nicoloboschi, line 752: **`success=true` is recorded for a class of real terminal failures.** Success is inferred here from "the executor didn't raise" — but `memory_engine.execute_task` catches and *swallows* several terminal failures, marking the op failed itself and returning normally:

- `file_convert_retain` failures (`memory_engine.py:1773-1778`) → `_mark_operation_failed`, no re-raise
- non-retryable errors (integrity violations, invalid embedding dims, …) via `_is_non_retryable_task_error` (`:1779-1794`) → `_mark_operation_failed`, no re-raise

Both return normally, so `terminal_success` becomes `True` and we record `success=true` for an op that actually failed. Only retry-exhausted (`:1850 raise`) and unexpected errors reach the poller's `except Exception` → `success=false`.

Net: the worker `success=false` counter systematically under-reports failures — exactly the "is anything failing?" signal this PR adds. An operator alerting on worker success rate would see ~zero failures while conversions / deterministic errors fail constantly (false confidence — arguably worse than no signal).

The executor already *knows* the real outcome (it called `_mark_operation_failed`). Cleanest fix is to thread that status back — have `execute_task` return a terminal status, or read the op's final `status` — rather than inferring from raise/no-raise. If that's out of scope here, please narrow the documented meaning so nobody builds alerts on a metric that can't see deterministic failures.

hindsight-api-slim/tests/test_worker.py
```diff
@@ -70,6 +70,95 @@ async def clean_operations(pool):
...
+
+    assert _metric_operation_label("retain") == "retain"
+    assert _metric_operation_label("batch_retain") == "retain"
+    assert _metric_operation_label("file_convert_retain") == "retain"
+    assert _metric_operation_label("consolidation") == "consolidation"
+    assert _metric_operation_label("reflect") == "reflect"
+    assert _metric_operation_label(None) == "unknown"
+
+
+class TestWorkerOperationMetrics:
```
nicoloboschi, line 87: These tests mock the executor, so they only exercise raise vs. return-normally — they never hit the mark-failed-then-return-normally path that produces the wrong `success` label. Green tests here mask the gap flagged in `poller.py`. Suggest adding a case where the executor stub calls `_mark_operation_failed` and returns normally, asserting the recorded `success` reflects the real outcome (it would currently fail — which is the point).

nicoloboschi (commented):
Reviewed the metrics refactor plus the executor failure semantics this depends on. The `record_operation_result` extraction is clean, #2122 cancellation handling is preserved, recording is correctly outside the executor `try` and guarded, and defer/retry are properly excluded as non-terminal. One material gap on the `success` label (inline), plus a couple of minor notes.

**PR description nit:** it lists `reflect` as running through the worker, but `execute_task` has no `reflect` branch, so `operation="reflect"` / `source="worker"` will never emit. Worth correcting the description.

hindsight-api-slim/hindsight_api/worker/poller.py
kubaodias: Good catch, and agreed it's material — thanks. Rather than re-engineer `execute_task` to thread terminal status back (the more invasive option), I've gone with scoping this metric honestly, since #1987 landed yesterday and already covers the gap authoritatively:

- The `success` label is now documented as a **completion-throughput** signal, not a failure-rate one. `success=false` = the task raised out to the poller (unexpected / retry-exhausted); deterministic self-handled failures (`file_convert_retain`, non-retryable) record `success=true` — explicitly called out in the code comment and PR description so nobody builds a failure-rate alert on it.
- For authoritative failure visibility, operators should use **`hindsight_async_operations{status="failed"}`** (#1987), which reads each operation's final DB `status` — so it sees exactly the swallowed-failure cases this counter can't. The two are complementary: counter (is anything finishing?) + gauge (is anything stuck/failed?).

hindsight-api-slim/tests/test_worker.py
kubaodias: Addressed in the same direction as the `poller.py` thread. Since we're scoping the metric as completion-throughput (not failure-rate), the mark-failed-then-return-normally path is — by design — indistinguishable from a clean completion at the poller boundary and records `success=true`.

`test_executor_returning_normally_records_success` covers this: its docstring documents that return-normally → `success=true` is intentional (including deterministic self-handled failures), and points to the `hindsight_async_operations{status="failed"}` gauge for the real failure signal. A dedicated assertion isn't possible at this layer — with the executor mocked there's no real `_mark_operation_failed` / DB status to observe.

If we instead decide to make the metric self-sufficient on failures, this is the test I'd extend to read the op's final status and assert `success=false`.

nicoloboschi (approved):

PR #2274, opened 2026-06-17
fix(api): resolve pg_trgm `%` operator via runtime search_path (#2270)

### LINKED ISSUE #2270: operator does not exist: text % text
### Bug Description

I keep getting errors where the last saving part crashes

2026-06-17 14:46:39,348 - ERROR - hindsight_api.api.http - Error in /v1/default/banks/user-123/memories (retain): operator does not exist: text % text

HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.

Input (1 items):
  [0] content='user likes the C# programming language', context=None, timestamp=None

```py
  File "/app/api/hindsight_api/engine/entity_resolver.py", line 420, in _resolve_entities_batch_trigram
    await conn.fetch(
  File "/app/api/hindsight_api/engine/db/postgresql.py", line 45, in fetch
    return await self._conn.fetch(query, *args, timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/api/.venv/lib/python3.11/site-packages/asyncpg/connection.py", line 694, in fetch
    return await self._execute(
           ^^^^^^^^^^^^^^^^^^^^
  File "/app/api/.venv/lib/python3.11/site-packages/asyncpg/connection.py", line 1873, in _execute
    result, _ = await self.__execute(
                ^^^^^^^^^^^^^^^^^^^^^
  File "/app/api/.venv/lib/python3.11/site-packages/asyncpg/connection.py", line 1970, in __execute
    result, stmt = await self._do_execute(
                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/api/.venv/lib/python3.11/site-packages/asyncpg/connection.py", line 2021, in _do_execute
    stmt = await self._get_statement(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/api/.venv/lib/python3.11/site-packages/asyncpg/connection.py", line 443, in _get_statement
    statement = await self._protocol.prepare(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "asyncpg/protocol/protocol.pyx", line 165, in prepare
```

I have setup hindsight up with my own docker which has pgterm and pgvector
in the schema where I've put hindsight the % works (also for their user)
```sql
SELECT 'hello' % 'helo';
-> true
```

The llm request does correctly save, just not the document
Image

I have confirmed the extensions are there and WORK 

### Steps to Reproduce

1. Configured hindsight+control plane
2. Set my postgresdb to my own (pg18.2)
3. Ran the migrations
4. Connected Lmstudio server (does work)
5. In the control plane make a new document (small text no file)
6. Type a short sentence (Eg 'user likes C#')
7. document fails to safe

### Expected Behavior

I expect the docker container to retain documents

### Actual Behavior

Crash when saving document, only saves LLM request

### Version

latest-slim (2026-06-12)

### LLM Provider

LM Studio

nicoloboschi:
## Problem

Retain crashes on external Postgres with:

```
operator does not exist: text % text
HINT: No operator matches the given name and argument types.
  ...entity_resolver.py, line 420, in _resolve_entities_batch_trigram
```

Reported in #2270. The reporter confirmed pg_trgm is installed and `SELECT 'hello' % 'helo'` works in their psql session — yet retain still fails.

## Root cause

The trigram entity lookup uses the **bare `%` operator**:

```sql
... AND LOWER(e.canonical_name) % LOWER(q.query_text)
```

Hindsight reaches its **tables** via fully-qualified names (`fq_table` → `"schema".entities`), but the `%` **operator** (and `gin_trgm_ops`) are resolved through the connection's `search_path`. The runtime asyncpg pool only set `search_path` for the **vchord** backend — every other backend used the DB role's default. When that default omits the schema holding pg_trgm, the operator can't resolve.

This explains every symptom:
- **psql works** — the interactive session's search_path includes the pg_trgm schema.
- **Migrations succeed** — `alembic/env.py` sets `search_path TO "<schema>", public` before running them.
- **Retain fails** — the pool connection's role default search_path doesn't include it.

There's also a sharp edge: `CREATE EXTENSION IF NOT EXISTS pg_trgm` (migration `c1a2b3d4e5f6`) has no `SCHEMA` clause, so on a fresh custom-schema install pg_trgm can land in the **tenant schema** rather than `public`.

## Fix

In the per-connection init callback (`_init_connection`):
- Set `search_path` explicitly on **every** runtime connection (not just vchord), generalizing the existing vchord handling of the identical operator-resolution problem (`<&>`).
- **Discover** the schema pg_trgm actually lives in and add it to the path, so `%` resolves even when the extension landed in a custom schema.

Search_path computation is extracted into `build_connection_search_path()` so it's unit-testable without a live PG. This does **not** weaken schema isolation — tenant tables are still reached via `fq_table`; only operator/type resolution relies on search_path.

## Tests

`tests/test_connection_search_path.py` covers pgvector/vchord backends, pg_trgm in public, in a custom schema (quoted + appended), and absent. Existing `test_entity_resolver_pg_trgm.py` still passes.

Closes #2270

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -139,6 +139,37 @@ def fq_table(table_name: str) -> str:
     return _fq_table(table_name)
 
 
+def build_connection_search_path(text_search_extension: str, trgm_schema: str | None) -> str:
+    """Build the ``SET search_path`` statement for a runtime PostgreSQL connection.
+
+    Extension operators resolve their operand types through the session
+    ``search_path``, not through the fully-qualified table names we use for
+    tenant isolation. A connection whose role default ``search_path`` omits the
+    schema that holds the extension fails with errors like
+    ``operator does not exist: text % text`` (pg_trgm's ``%``, used by the
+    retain trigram entity lookup) or ``type "bm25vector" does not exist``
+    (vchord BM25's ``<&>``). External Postgres deployments frequently pin the
+    role ``search_path`` to a custom schema, so we set it explicitly here.
+
+    This does not weaken schema isolation: tenant tables are always accessed via
+    fully-qualified names (:func:`fq_table`); only operator/type resolution
+    relies on ``search_path``.
+
+    ``trgm_schema`` is the schema pg_trgm was actually installed into (``CREATE
+    EXTENSION`` without a ``SCHEMA`` clause lands in the first ``search_path``
+    entry at migration time, which may be a custom tenant schema rather than
+    ``public``). Pass ``None`` when pg_trgm is absent.
+    """
+    schemas = ['"$user"', "public"]
+    if trgm_schema and trgm_schema != "public":
+        schemas.append(f'"{trgm_schema}"')
+    if text_search_extension == "vchord":
+        # VectorChord BM25 registers objects in dedicated schemas
+        # (vchord_bm25 -> bm25_catalog, pg_tokenizer -> tokenizer_catalog).
+        schemas.extend(["bm25_catalog", "tokenizer_catalog"])
+    return "SET search_path TO " + ", ".join(schemas)
+
+
 def _json_default(obj: Any) -> str:
     """JSON serializer for types commonly carried through async task payloads."""
     if isinstance(obj, datetime):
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Closing — this is a rare external-Postgres config case (pg_trgm reachable in psql but not on the API role's runtime search_path). Rather than change search_path behavior for all deployments, we'll give the reporter a config-side workaround on #2270.

Issue #626, opened 2026-03-20
pg_trgm not available on Azure Postgres causes startup crash and silent background task failures


eight-atulya:
### Bug Description

When running Hindsight on Azure PostgreSQL Flexible Server, things break in two steps when the `pg_trgm` database extension is not available.

**Step 1 — Startup crash:** On first boot, the database migration `c1a2b3d4e5f6` fails immediately because it tries to create search indexes that depend on `pg_trgm`. Azure's managed Postgres doesn't always have this extension installed, so the whole app crashes at startup.

**Step 2 — Silent runtime failure after bypass:** If you work around Step 1 by setting `HINDSIGHT_API_DISABLE_PG_TRGM_MIGRATION=true`, the app starts fine — but background "retain" jobs start failing right away with the error: `operator does not exist: text % text`.

This happens because even after skipping the migration, Hindsight still defaults to a trigram-based search strategy (`DEFAULT_RETAIN_ENTITY_LOOKUP`) at runtime, which also requires `pg_trgm`. So the failure just moves from startup to runtime, and it's not obvious why things are broken unless you already know about the extension dependency.

**Database:** Azure PostgreSQL Flexible Server (managed)
**Migration affected:** `c1a2b3d4e5f6`
**First seen:** 2026-03-16

### Steps to Reproduce

1. Set up an Azure PostgreSQL Flexible Server where the `pg_trgm` extension is **not** installed (this is common on Azure managed Postgres — the extension is not always available by default).
2. Point a fresh Hindsight deployment at this database and start the API.
3. Watch the startup logs — you'll see a crash during database migration `c1a2b3d4e5f6` with errors related to a missing operator or index class that requires `pg_trgm`.
4. Add the environment variable `HINDSIGHT_API_DISABLE_PG_TRGM_MIGRATION=true` and restart. The app now starts successfully.
5. Trigger any retain operation (for example, store some content so Hindsight can retrieve it later).
6. Check the background worker logs — you'll see repeated errors like `operator does not exist: text % text`.

**Confirmed workaround (both env vars needed):**
- `HINDSIGHT_API_DISABLE_PG_TRGM_MIGRATION=true` — skips the broken migration at boot
- `HINDSIGHT_API_DEFAULT_RETAIN_ENTITY_LOOKUP=full` — switches runtime search away from trigram, stopping the background errors

### Expected Behavior

If `pg_trgm` is not installed on the database, Hindsight should handle this automatically — no manual env vars should be needed.

[...]

PR #2275, opened 2026-06-17
fix(db): pin pg_trgm to public schema (#2270)

nicoloboschi:
## Problem

Follow-up to #2274 for issue #2270 (`operator does not exist: text % text` on retain).

`CREATE EXTENSION IF NOT EXISTS pg_trgm` (revision `c1a2b3d4e5f6`) has **no `SCHEMA` clause**. On a custom-schema install the extension lands in whatever schema is first on the `search_path` at migration time — typically a **tenant schema** rather than `public`. Its `%` operator then only resolves for connections whose `search_path` includes that schema.

## Fix

New migration `d7e9f1a3b5c2` pins pg_trgm to `public` via `ALTER EXTENSION pg_trgm SET SCHEMA public`, so the operator lives in the standard, always-on location.

- **Idempotent / per-tenant safe** — pg_trgm is database-global, so the first schema relocates it and later schemas find it already in `public`.
- **Degrades gracefully** — skips when pg_trgm is absent (managed Postgres without the contrib module → resolver falls back to the `full` lookup, per #626) or already in `public`; on insufficient privileges it rolls back and leaves the extension in place.
- **No-op downgrade** — we don't track the original schema, and moving it back would risk re-breaking operator resolution.

## Relationship to #2274

This is belt-and-suspenders. PR #2274 makes the runtime discover pg_trgm's actual schema and add it to `search_path`, so `%` resolves regardless of where the extension lives — that's the behavioural fix. This PR standardises the install location so the extension sits where everything expects it. They're complementary; either alone resolves the reporter's crash.

## Tests

`tests/test_migration_shape.py` and `tests/test_alembic_dag.py` pass — dialect dispatcher present, single head preserved.

Refs #2270

### DIFF
--- a/hindsight-api-slim/hindsight_api/alembic/versions/d7e9f1a3b5c2_relocate_pg_trgm_to_public.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/d7e9f1a3b5c2_relocate_pg_trgm_to_public.py
@@ -1,0 +1,69 @@
+"""Relocate pg_trgm into the public schema.
+
+``CREATE EXTENSION IF NOT EXISTS pg_trgm`` (revision ``c1a2b3d4e5f6``) has no
+``SCHEMA`` clause, so on a custom-schema install it lands in whatever schema is
+first on the ``search_path`` at migration time — typically a tenant schema
+rather than ``public``. Its ``%`` operator then only resolves for connections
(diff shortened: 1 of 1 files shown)


nicoloboschi:
Closing — opting for a reporter-side workaround instead of relocating pg_trgm on existing deployments. See #2270.

PR #2283, opened 2026-06-18
fix(tests) S3 testcontainers port mapping race

Sanderhoff-alt:
## Summary

- retry testcontainers Docker port mapping lookups while starting the SeaweedFS S3 test container
- scope the retry patch to the fixture startup path and restore the original testcontainers method afterward
- keep the existing fixture cleanup and Ryuk cleanup behavior intact

## Root Cause

On Docker Desktop, a container can be reported as running before its published port appears in Docker's `NetworkSettings.Ports`. testcontainers reads Ryuk's 8080 mapping during `container.start()`, and the S3 test then reads SeaweedFS's 8333 mapping. If either lookup happens during that short window, the test fails before the SeaweedFS readiness probe can run.

## Validation

- reproduced the failure locally before the fix with `tests/test_file_storage_s3.py::test_s3_storage_store_and_retrieve`
- `cd hindsight-api-slim && uv run pytest tests/test_file_storage_s3.py -q`
- `./scripts/hooks/lint.sh`

### DIFF
--- a/hindsight-api-slim/tests/test_file_storage_s3.py
+++ b/hindsight-api-slim/tests/test_file_storage_s3.py
@@ -12,6 +12,8 @@ import subprocess
 import tempfile
 import time
 import uuid
+from collections.abc import Iterator
+from contextlib import contextmanager
 
 import httpx
 import pytest
@@ -21,6 +23,7 @@ logger = logging.getLogger(__name__)
 
 try:
     from testcontainers.core.container import DockerContainer
+    from testcontainers.core.docker_client import DockerClient as _DockerClient
 
     _has_testcontainers = True
 except ImportError:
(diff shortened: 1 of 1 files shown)


nicoloboschi (approved):

- **Targeted fix, correct exception**: `DockerClient.port` raises `ConnectionError` when the published port isn't yet in `NetworkSettings.Ports` — exactly what the retry catches. Monkeypatch signature `(self, container_id, port)` matches the real method exactly.
- **Properly scoped**: the patch is applied via a contextmanager around just `container.start()` + the port lookups, and restored in `finally`. No global testcontainers state leaks past the fixture.
- **Validated locally**: `uv run pytest tests/test_file_storage_s3.py` — the 4 S3 testcontainers tests pass (incl. `test_s3_storage_store_and_retrieve`). The lone error was `test_s3_file_retain_api_end_to_end` hitting a pg0 port-5556 collision from another local instance — unrelated to this change.
- CI lint/generated-file checks green; test-api is skipped only because this is a fork PR (no secrets).

Clean, minimal, test-only. LGTM.

PR #2288, opened 2026-06-18
fix(mcp): restore custom tool instructions

### LINKED ISSUE #2286: HINDSIGHT_API_MCP_INSTRUCTIONS documented but not wired to MCP tool descriptions
### Bug Description

  Hi, I’m trying to customize the MCP `retain` / `recall` tool descriptions using the documented `HINDSIGHT_API_MCP_INSTRUCTIONS` environment variable.

  The docs say:

  https://hindsight.vectorize.io/developer/configuration#mcp-server

  > `HINDSIGHT_API_MCP_INSTRUCTIONS` | Additional instructions appended to retain/recall tool descriptions

  However, in the current `main` branch, I can’t find this env var wired into runtime config.

  Relevant code:

  - `hindsight-api-slim/hindsight_api/config.py`
    - Defines `ENV_MCP_ENABLED`
    - Defines `ENV_MCP_ENABLED_TOOLS`
    - Defines `ENV_MCP_STATELESS`
    - But does not appear to define or read `HINDSIGHT_API_MCP_INSTRUCTIONS`

  - `hindsight-api-slim/hindsight_api/mcp_tools.py`
    - `MCPToolsConfig` has:
      ```py
      retain_description: str | None = None
      recall_description: str | None = None
      ```
    - `_register_recall()` does use:
      ```py
      description = config.recall_description or DEFAULT_MCP_RECALL_DESCRIPTION
      ```
    - `_register_retain()` has the same pattern for retain

  - `hindsight-api-slim/hindsight_api/api/mcp.py`
    - `create_mcp_server()` constructs `MCPToolsConfig(...)`
    - But it only passes resolvers/auth/tools
    - It does not pass `retain_description` or `recall_description`

  So it looks like the lower-level description override exists, but the documented env var is not currently connected to it.

  Could you confirm whether `HINDSIGHT_API_MCP_INSTRUCTIONS` is expected to work in the API server MCP endpoint (`/mcp/{bank_id}/`)?

  If yes, should it be implemented by:
  1. adding `ENV_MCP_INSTRUCTIONS = "HINDSIGHT_API_MCP_INSTRUCTIONS"` in `config.py`,
  2. adding a `mcp_instructions` field to global config,
  3. reading it from `os.getenv(...)`,
  4. and passing appended descriptions into `MCPToolsConfig(retain_description=..., recall_description=...)` in `api/mcp.py`?

  Thanks.

### Steps to Reproduce

 -     image: ghcr.io/vectorize-io/hindsight:latest-slim
  1. Start Hindsight with MCP enabled and set `HINDSIGHT_API_MCP_INSTRUCTIONS`.

     Example Docker env:

     ```env
     HINDSIGHT_API_MCP_ENABLED=true
     HINDSIGHT_API_MCP_ENABLED_TOOLS=retain,recall,reflect
     HINDSIGHT_API_MCP_INSTRUCTIONS="This memory bank stores reusable agent/team problem-solving experience. Use recall before choosing tools when prior experience may affect the route."

  2. Restart/recreate the container so the environment is loaded.

     docker compose up -d --force-recreate

  3. Verify the env var exists inside the running container.

     docker exec <container> env | grep HINDSIGHT_API_MCP

     Expected to see HINDSIGHT_API_MCP_INSTRUCTIONS.

  4. Call MCP tools/list against a bank endpoint, for example:

     curl -sS -X POST http://localhost:8888/mcp/default/ \
       -H "Accept: application/json, text/event-stream" \
       --data '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}'

     If stateful MCP requires initialization first, initialize the session and then call tools/list.

  5. Inspect the returned recall.description and retain.description.


### Expected Behavior

It would also be useful to support full description overrides, not only appended instructions.

  For example:

  ```env
  HINDSIGHT_API_MCP_RETAIN_DESCRIPTION="Custom full description for retain..."
  HINDSIGHT_API_MCP_RECALL_DESCRIPTION="Custom full description for recall..."
  HINDSIGHT_API_MCP_INSTRUCTIONS="Additional shared guidance appended to both retain and recall..."

  Suggested precedence:

  1. If HINDSIGHT_API_MCP_RECALL_DESCRIPTION is set, use it as the full recall tool description.
  2. Otherwise use DEFAULT_MCP_RECALL_DESCRIPTION.
  3. If HINDSIGHT_API_MCP_INSTRUCTIONS is set, append it to the final description.

  Same behavior for retain.

  This would make it possible to adapt Hindsight MCP for non-personal-memory use cases, such as agent/team engineering experience memory, where the default user-preference-oriented descriptions can cause models to underuse recall.

### Actual Behavior

The returned retain.description and recall.description remain the default descriptions. The custom text from HINDSIGHT_API_MCP_INSTRUCTIONS is not included.

### Version

ghcr.io/vectorize-io/hindsight:latest-slim

### LLM Provider

None

Sanderhoff-alt:
## Summary

Fixes #2286.

This restores `HINDSIGHT_API_MCP_INSTRUCTIONS` for HTTP MCP servers by parsing the environment variable into configuration and appending the extra guidance to the `retain` and `recall` MCP tool descriptions. The behavior matches the original local MCP implementation and does not change `reflect` or other management tools.

## Root cause

This is a regression introduced by `ea8163c56d65f5c18984d2d7435dd21b94454cd3` (`fix(mcp): unify hindsight-mcp-local and server mcp  (#407)`). That commit replaced the local stdio MCP server with an HTTP server wrapper and removed the code path that read `HINDSIGHT_API_MCP_INSTRUCTIONS` and passed custom `retain_description` / `recall_description` values into `MCPToolsConfig`. The shared hooks remained in place, but the HTTP MCP server never supplied them.

## Validation

- `cd hindsight-api-slim && uv run pytest tests/test_mcp_routing.py -q`
- `cd hindsight-api-slim && uv run pytest tests/test_mcp_tool_filtering.py tests/test_mcp_extension.py -q`
- `./scripts/hooks/lint.sh`

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/mcp.py
+++ b/hindsight-api-slim/hindsight_api/api/mcp.py
@@ -9,7 +9,7 @@ from fastmcp import FastMCP
 
 from hindsight_api import MemoryEngine
 from hindsight_api import __version__ as HINDSIGHT_VERSION
-from hindsight_api.config import _get_raw_config
+from hindsight_api.config import DEFAULT_MCP_RECALL_DESCRIPTION, DEFAULT_MCP_RETAIN_DESCRIPTION, _get_raw_config
 from hindsight_api.engine.memory_engine import _current_schema
 from hindsight_api.extensions import MCPExtension, load_extension
 from hindsight_api.extensions.tenant import AuthenticationError
@@ -78,6 +78,19 @@ def get_current_mcp_authenticated() -> bool:
     return _current_mcp_authenticated.get()
 
 
+def _build_mcp_tool_descriptions(extra_instructions: str | None) -> tuple[str | None, str | None]:
+    """Return custom retain/recall descriptions when server-level MCP instructions are set."""
+    if not isinstance(extra_instructions, str):
+        return None, None
+
+    extra_instructions = extra_instructions.strip()
+    if not extra_instructions:
+        return None, None
+
+    suffix = f"\n\nAdditional instructions: {extra_instructions}"
+    return DEFAULT_MCP_RETAIN_DESCRIPTION + suffix, DEFAULT_MCP_RECALL_DESCRIPTION + suffix
+
+
 def create_mcp_server(memory: MemoryEngine, multi_bank: bool = True) -> FastMCP:
     """
     Create and configure the Hindsight MCP server.
@@ -135,6 +148,10 @@ def create_mcp_server(memory: MemoryEngine, multi_bank: bool = True) -> FastMCP:
         allowed = frozenset(global_config.mcp_enabled_tools)
         base_tools = (base_tools if base_tools is not None else _ALL_TOOLS) & allowed
 
+    retain_description, recall_description = _build_mcp_tool_descriptions(
+        getattr(global_config, "mcp_instructions", None)
+    )
+
     # Configure and register tools using shared module
     config = MCPToolsConfig(
         bank_id_resolver=get_current_bank_id,
@@ -144,6 +161,8 @@ def create_mcp_server(memory: MemoryEngine, multi_bank: bool = True) -> FastMCP:
         mcp_authenticated_resolver=get_current_mcp_authenticated,  # Propagate MCP pre-auth flag
         include_bank_id_param=multi_bank,
         tools=base_tools,
+        retain_description=retain_description,
+        recall_description=recall_description,
     )
 
     register_mcp_tools(mcp, memory, config)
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
LGTM. Clean, minimal regression fix that faithfully restores HINDSIGHT_API_MCP_INSTRUCTIONS (dropped in #407), correctly scoped to retain/recall only. Config plumbing complete and treated as static (correct — create_mcp_server runs once at startup). Docs already cover it; tests pass locally.

PR #2297, opened 2026-06-18
feat(aider): add Aider integration (session-bracketing memory wrapper)

DK09876:
## Summary

Adds a first-party **Aider** integration for Hindsight — persistent long-term memory across pair-programming sessions.

Aider has **no MCP client and no per-prompt hook**, so (unlike the Zed/Continue/OpenHands MCP integrations) this is a **drop-in wrapper** around the `aider` CLI that brackets each session:

- **Recall (before):** queries Hindsight, writes results to `.aider.hindsight-memory.md`, and launches `aider --read .aider.hindsight-memory.md …` so the memory is in context. `aider -m "…"` uses the message as the recall query.
- **Retain (after):** when Aider exits, reads the slice of the chat-history file written during the session and retains it to the repo's bank.

Bank is **per git repo**, so a project's memory is shared with the other Hindsight editor integrations.

```bash
pip install hindsight-aider aider-chat
hindsight-aider -m "add retry logic"   # used exactly like aider
```

## What's included
- **`hindsight_aider`** package: `config`, `bank` (per-repo derivation), `content` (recall query / memory file / transcript), `runner` (recall → aider subprocess → retain, with an injectable `run_aider` for testing), `cli` (passthrough).
- **33 deterministic tests** exercising the real flow — recall writes the `--read` file, aider is invoked with `--read <file>` + the passthrough args, and retain captures **only the new chat-history slice** (not prior sessions); plus a gated `requires_real_llm` retain→recall E2E.
- **CI** job `test-aider-integration` + detect-changes + aggregation.
- **Release** registration in `VALID_INTEGRATIONS` **and** the changelog generator registry.
- **Docs** page, registry entry, icon, README row.

## Testing
- 33 deterministic tests pass; `ruff check`/`format` clean (repo `ruff.toml`); `uv build` + `uv lock --check` consistent.
- Gated E2E verified against a live local Hindsight (real retain → recall roundtrip writes the recalled memory file).


benfrank241:
@DK09876 nice work, this one's clean. wrapper pattern matches `cline` / `agent-framework`, and using the git repo name so memory is shared with the other editor integrations on the same project is the right call.

couple of comments before merging:

- `recall_preamble` is on `AiderConfig` (used in `runner.do_recall` via `format_memory`) but missing from both `_FILE_KEYS` and `_ENV_KEYS`, so users can't actually override it — it's effectively a constant. either wire it through (`recallPreamble` / `HINDSIGHT_AIDER_RECALL_PREAMBLE`) or drop the field and inline the default.
- typing nit: `recall_types: list = field(...)` → `list[str]`; same for `argv: Optional[list] = None` in `cli.py`. our standard is typed kwargs everywhere.
- `_cast` in `config.py` is untyped and silently returns `None` on bad input which then gets skipped — fine as a soft-fail, but please add `def _cast(value: object, typ: type) -> object | None:` so ty doesn't ignore it.

once the above lands i'll bundle this with the next release.

DK09876:
Thanks @benfrank241 — all three addressed in dd3584c0:

- **`recall_preamble`:** wired through as `recallPreamble` / `HINDSIGHT_AIDER_RECALL_PREAMBLE` (with a test). Good catch — it was a dead override.
- **Typing:** `recall_types -> list[str]`, `cli.main(argv: Optional[list[str]])`.
- **`_cast`:** now `def _cast(value: object, typ: type) -> object | None`.

34 tests green, ruff clean. Ready when you are.

PR #2304, opened 2026-06-19
fix(retain): make chunk_text idempotent to avoid duplicate chunk_id crash (#2301)

### LINKED ISSUE #2301: Increasing structured chunk size causes retain failures
### Bug Description

Increasing structured chunk size causes retaining documents with large jsonl lines to fail.

### Steps to Reproduce

1. Set e.g. `HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE=4500` (above the default chunk size)
2. Try to retain a jsonl document where a line exceeds the chunk size

Agent provided code showing why this happens:
 ```python                                                                                                                                                                                                                      
   from hindsight_api.engine.retain.fact_extraction import chunk_text                                                                                                                                                           
   R, S = 3000, 5000                                                                                                                                                                                                            
   content = '[{"role":"assistant","content":"' + "x"*6000 + '"}]'                                                                                                                                                              
   pre = chunk_text(content, max_chars=R, structured_chunk_size=S)        # producer pre-chunks once                                                                                                                            
   for p in pre:                                                                                                                                                                                                                
       sub = chunk_text(p, max_chars=R, structured_chunk_size=S)          # extraction RE-chunks each pre-chunk                                                                                                                 
       if len(sub) > 1:                                                                                                                                                                                                         
           ids = [f"bank_doc_0"] * len(sub)                                # consumer flattens all sub-chunks to one index                                                                                                      
           print(f"{len(sub)} sub-chunks -> chunk_ids in one ON CONFLICT upsert = {ids}  -> DUPLICATE")                                                                                                                         
   # 6000-char turn: pre_lens=[20,12,4999,1002,5], pre[3] re-splits into 2 -> crash                                                                                                                                             
 ```                                                                                                                                                                                                                            


### Expected Behavior

Retains successfully

### Actual Behavior

`asyncpg.exceptions.CardinalityViolationError: ON CONFLICT DO UPDATE command cannot affect row a second time`

### Version

0.8.3

### LLM Provider

None

r266-tech:
Fixes #2301.

### Problem
Raising `HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE` above `retain_chunk_size` makes retaining a doc with an oversized conversation turn / JSONL line crash with `asyncpg.CardinalityViolationError: ON CONFLICT DO UPDATE command cannot affect row a second time`.

### Root cause
`chunk_text()` is not a fixed point in this regime, but the pipeline assumes **one pre-chunk → one stored chunk** — the consumer maps every extracted chunk back to a single `global_idx`-derived `chunk_index`. When `structured_chunk_size > max_chars`:

- an oversized unit is split via `_split_oversized_unit(unit, structured_limit)` into fragments up to `structured_limit`, i.e. **larger than `max_chars`** (the comment *"so no chunk runs far over budget — the extractor won't re-chunk"* is violated);
- extraction re-runs `chunk_text` on every pre-chunk and re-splits those fragments into >1 sub-chunks, which all collapse onto one `chunk_index` → duplicate `chunk_id`s in one `ON CONFLICT` upsert → crash.

A second variant: a JSONL line kept whole between `max_chars` and the structured cap is re-chunked on its own, where `_chunk_jsonl` declines (needs ≥2 lines) and it falls back to `_split_oversized_unit` → the same collapse.

### Fix (producer-side; preserves the 1:1 pre-chunk→chunk invariant)
Make `chunk_text` idempotent so the extractor's re-chunk pass is a no-op:

1. Split oversized units to `min(structured_limit, max_chars)` — fragments never exceed `max_chars`, so they are fixed points. This keeps the existing `structured_limit < max_chars` behaviour (still splits down to the smaller cap).
2. Keep a lone single JSON object whole only up to `structured_limit`, mirroring how a single JSONL line / conversation turn is kept whole — so a kept-whole line re-chunked alone comes back unchanged.

### Tests
Added re-chunk-stability tests covering the conversation-oversized-turn (the issue repro), the kept-whole JSONL line, and the oversized JSONL line — each asserts every produced chunk is a fixed point of `chunk_text`. All existing `test_chunking.py` cases still pass (the `min()` leaves the `structured_limit < max_chars` outputs identical). Pure-function tests, no DB required.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -473,6 +473,15 @@ def chunk_text(text: str, max_chars: int, structured_chunk_size: int | None = No
         if isinstance(parsed, list) and all(isinstance(turn, dict) for turn in parsed):
             # This looks like a conversation - chunk at turn boundaries
             return _chunk_conversation(parsed, max_chars, structured_limit)
+        if isinstance(parsed, dict) and len(text) <= structured_limit:
+            # A lone structured unit (a single JSON object) is kept whole up to the
+            # structured-chunk cap, mirroring how a single JSONL line or conversation
+            # turn is kept whole. Keeping it whole here makes chunk_text a fixed point:
+            # extraction re-runs chunk_text on every pre-chunk, so a kept-whole line
+            # re-chunked on its own must come back unchanged. Otherwise it expands into
+            # several sub-chunks that all collapse onto one chunk_index and crash the
+            # ON CONFLICT upsert with duplicate chunk_ids (#2301).
+            return [text]
     except (json.JSONDecodeError, ValueError):
         pass
 
@@ -519,7 +528,7 @@ def _chunk_conversation(turns: list[dict], max_chars: int, structured_limit: int
         # text so no chunk runs far over budget (the extractor won't re-chunk).
         if turn_unit_size > structured_limit:
             _flush()
-            chunks.extend(_split_oversized_unit(turn_json, structured_limit))
+            chunks.extend(_split_oversized_unit(turn_json, min(structured_limit, max_chars)))
             continue
 
         # If adding this turn would exceed limit and we have turns, save current chunk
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Thanks @r266-tech — this overlaps with #2308, which took the same approach (idempotent `chunk_text`: `min(structured_limit, max_chars)` at both split sites + keep a lone JSON object whole up to the structured cap) plus an end-to-end retain regression test, and has now been merged. Closing this as a duplicate. Appreciate the fix!

PR #2306, opened 2026-06-19
[opencode] Add suport for HINDSIGHT_RETAIN_TAGS

mdbenito:
This PR adds support to the opencode integration for a new environment `HINDSIGHT_RETAIN_TAGS`. If this variable is set, tags are added to every retain operation. I also added two missing documentation lines for `HINDSIGHT_RECALL_TAGS`. Incidentally, I noted that there are no tests for `HINDSIGHT_RECALL_TAGS`, but thought those would be better left for another PR. 

(Sorry about the heavy reformatting of the table in the readme, I tried to keep it tidy as it was, but thought that it would be too wide if I adjusted all rows to match the longest one anyway, so I went for the middle ground)


koriyoshi2041:
I reproduced the failing `verify-generated-files` diff locally. The functional opencode test is green; the remaining change is just prettier formatting for `hindsight-integrations/opencode/README.md`.

Command I ran:

```bash
cd hindsight-integrations/opencode
npx --yes prettier --write --config ../../.prettierrc.json --ignore-path ../../.gitignore .
```

It only changed the env-var table alignment in `README.md` (`32 +++++++++++++++----------------`), matching the CI log. Committing that formatted README should clear the generated-files check.

mdbenito:
> Command I ran:
> 
> ```shell
> cd hindsight-integrations/opencode
> npx --yes prettier --write --config ../../.prettierrc.json --ignore-path ../../.gitignore .
> ```

Thanks! I didn't think the README would be run through prettier, although on second thought it was clear it did. Sorry for not checking first.

nicoloboschi (approved):

PR #2315, opened 2026-06-19
fix(cache): prevent stale bank stats after concurrent invalidation

zupengwang:
## Summary

Prevent an in-flight bank stats query from repopulating the cache with stale
data after the cache has been invalidated or cleared.

## Root cause

`BankStatsCache` coalesces concurrent cache misses using an in-flight future
stored by `(schema, bank_id)`.

`invalidate()` previously removed only the cached entry. If a stats query was
already running:

1. A bank mutation invalidated the cache.
2. Requests arriving after the mutation joined the old in-flight query.
3. The old query completed with pre-mutation data.
4. That stale result was stored for the full cache TTL.

The same race existed in `clear()`.

## Fix

- Detach in-flight loaders when invalidating a key or clearing the cache.
- Allow post-invalidation requests to start a fresh loader immediately.
- Require a loader to still own the key's in-flight slot before it may populate
  the cache or remove that slot.
- Let callers that started before invalidation finish with their original
  snapshot without allowing it to overwrite newer data.

## Tests

Added concurrent regression coverage for:

- invalidation while a loader is in flight;
- clearing while a loader is in flight;
- a stale loader completing after the fresh loader;
- preservation of the fresh cached result.

## Verification

Run on Linux x86_64 with Python 3.11, Node.js 22, and PostgreSQL 17 + pgvector:

- `12 passed` - `tests/test_bank_stats_cache.py`
- `17 passed` - `tests/test_bank_stats.py`
- `ty check hindsight_api` passed
- Full `./scripts/hooks/lint.sh` passed

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/bank_stats_cache.py
+++ b/hindsight-api-slim/hindsight_api/engine/bank_stats_cache.py
@@ -96,7 +96,10 @@ class BankStatsCache:
             value = await loader()
         except BaseException as exc:
             async with self._lock:
-                self._in_flight.pop(key, None)
+                # Invalidation may have detached this loader and allowed a new
+                # one to claim the key. Never remove that newer loader's slot.
+                if self._in_flight.get(key) is in_flight:
+                    self._in_flight.pop(key, None)
             if not in_flight.done():
                 in_flight.set_exception(exc)
             # Suppress "Future exception was never retrieved" when no other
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM. Identity-guarded slot ownership is the correct fix for the coalescing-invalidation race; store + exception + invalidate/clear paths are all consistently guarded. Traced the orderings and ran the suite (12 passed). Approving.

PR #2322, opened 2026-06-19
fix(recall): allow exact filtering of untagged observations (#2295)

### LINKED ISSUE #2295: Allow filtering to untagged observations
### Use Case

I want to have both global and more specific observations. I want to be able to switch recall filtering between them.

### Problem Statement

If you use `shared` for observation scopes and switch later or use `[[], ["project"],...]`, there is currently no way to filter to global observations only for recall (duplicate observations will be recalled without maintaining a complicated updated negative match).

You can workaround this currently by doing something like tagging all memories with "global" and filtering to that, though it won't work for old untagged observations, for example. Seems like it makes more sense to handle things the same way `shared`

This would allow to filter to only untagged observations/"shared" observations with multiple observation scopes.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Priority

Nice to have

### Additional Context

_No response_

### Checklist

zupengwang:
## Summary

Fixes #2295.

Allow recall to select only global/untagged observations by using an empty tag
set with `tags_match="exact"`.

This enables users to switch between shared observations and tagged observation
scopes without adding a synthetic "global" tag or maintaining negative filters.

## Root cause

The centralized tag filters treated `None` and `[]` as "no filter" before
checking the matching mode.

As a result:

- SQL retrieval applied no filter and returned every scope.
- Python post-processing returned every result.
- An exact empty compound tag group excluded all untagged results.
- Link-expansion skipped Python filtering when the tag list was empty.

This contradicted the existing scope semantics where the empty tag set
represents the global observation scope.

## Fix

- Interpret an empty tag set as the global/untagged scope in `exact` mode.
- Match both historical `NULL` tags and current empty-array tags.
- Keep parameter indexes unchanged because the empty-scope SQL clause requires
  no bind parameter.
- Apply the same semantics to flat SQL filters, Python post-processing, and
  compound tag groups.
- Run link-expansion post-filtering for empty exact scopes.
- Preserve existing behavior for all other modes: empty tags still mean no
  filtering for `any`, `all`, `any_strict`, and `all_strict`.

## Tests

Added regression coverage for:

- flat SQL filtering with `None` and `[]`;
- bind parameter numbering after an empty exact scope;
- Python-side filtering of NULL, empty, and tagged results;
- compound exact-empty tag groups in SQL and Python;
- recall API filtering that returns an untagged memory while excluding a tagged
  memory.

## Verification

Run on Linux x86_64 with Python 3.11, Node.js 22, and PostgreSQL 17 + pgvector:

- `91 passed` - `tests/test_tags_visibility.py`
- `10 passed` - `tests/test_graph_filtering.py`
- `ty check hindsight_api` passed
- Full `./scripts/hooks/lint.sh` passed

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/search/link_expansion_retrieval.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/link_expansion_retrieval.py
@@ -251,7 +251,7 @@ class LinkExpansionRetriever(GraphRetriever):
             result.activation = row["score"]
             results.append(result)
 
-        if tags:
+        if tags or tags_match == "exact":
             results = filter_results_by_tags(results, tags, match=tags_match)
 
         if tag_groups:
--- a/hindsight-api-slim/hindsight_api/engine/search/tags.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/tags.py
@@ -7,13 +7,14 @@ Supports five matching modes via TagsMatch enum:
 - "all": AND matching, includes untagged memories
 - "any_strict": OR matching, excludes untagged memories
 - "all_strict": AND matching, excludes untagged memories
-- "exact": set-equality matching, excludes untagged memories
+- "exact": set-equality matching; an empty tag set selects untagged memories
 
 OR matching (any/any_strict): Memory matches if ANY of its tags overlap with request tags
 AND matching (all/all_strict): Memory matches if ALL request tags are present in its tags
 EXACT matching: Memory matches only if its tag set EQUALS the request tag set (order-
     independent). Used for observation "scope" filtering, where each observation lives
-    under exactly one scope (its full tag set) and "scope [a]" must not match "[a, b]".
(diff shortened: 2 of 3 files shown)


nicoloboschi:
Thanks for this @wangzupeng12061 — the approach here is solid. I've opened #2364 which implements the same exact-empty-scope fix and additionally updates the `RecallRequest` API descriptions and regenerates the client SDKs (Go/Python/TypeScript), OpenAPI spec, and docs so the user-facing surface is complete. It also keeps your parameter-offset regression test.

Proposing #2364 to supersede this PR. Closing in favor of it (credit to your investigation of the root cause).

nicoloboschi:
Closing in favor of #2364, which supersedes this with the same fix plus the regenerated client SDKs, OpenAPI spec, and docs. Thanks again @wangzupeng12061 for the root-cause work — credited in the superseding PR.

Issue #2312, opened 2026-06-19
Claude Code: directoryBankMap silently ignores a mapped directory reached via a symlink (normpath, not realpath)


andb:
> **Disclosed transparently:** found by an AI agent (Claude Opus 4.8) via a high-effort, adversarial multi-agent verification workflow, and filed as AI-discovered on purpose. It's checked against your actual code — a runnable repro and a failing test are below, and the exact verification steps are in the collapsed section at the end. Please judge it on the repro.

## Bug Description
In the Claude Code integration, `directoryBankMap` entries are matched against the session `cwd` with a comparison that does **not** resolve symlinks, so a directory and a symlink to that same directory are treated as different paths. The result: an explicit user-configured mapping silently fails to apply and session memory is routed to the wrong bank — with no error or log.

`hindsight-integrations/claude-code/scripts/lib/bank.py`, `derive_bank_id()` (~L98–100 on `main`):
```python
normalized_cwd = os.path.normcase(os.path.normpath(cwd))
for dir_path, bank_id in dir_map.items():
    if os.path.normcase(os.path.normpath(dir_path)) == normalized_cwd:
        return ...
```
`os.path.normpath` does not resolve symlinks (only `os.path.realpath` does). When the mapped directory and the live `cwd` are the **same directory reached by two different path strings** — the mapped dir is itself a symlink, sits under a symlinked mount (e.g. macOS `/var`→`/private/var`), or the map and cwd simply disagree on symlink-vs-real form — the entry does not match and the session falls through to the static/dynamic bank.

## Steps to Reproduce
```python
import sys, os, tempfile
sys.path.insert(0, "hindsight-integrations/claude-code/scripts")
from lib.bank import derive_bank_id

base = tempfile.mkdtemp()
real = os.path.realpath(os.path.join(base, "proj")); os.makedirs(real)  # canonical, immune to /var symlinks
link = os.path.join(base, "proj-link"); os.symlink(real, link)          # SAME dir, different path string
assert os.path.realpath(link) == real

cfg = {"directoryBankMap": {real: "myproj"}, "bankId": "fallback"}
print(derive_bank_id({"cwd": real}, cfg))   # -> myproj
print(derive_bank_id({"cwd": link}, cfg))   # -> fallback   (expected: myproj)
```
(`realpath` on the map key matters on macOS, where `mkdtemp` lives under the `/var`→`/private/var` symlink; without it the control case can mask the effect.)

## Expected Behavior
[...]

PR #2324, opened 2026-06-19
fix(claude-code): use realpath for directoryBankMap symlink resolution

365diascollaboration-prog:
## Problem

`directoryBankMap` entries are matched against the session `cwd` using `os.path.normpath`, which does **not** resolve symlinks. When the `cwd` is a symlink to a mapped directory (or when the map key itself is a symlink), the comparison silently fails and memory is routed to the wrong bank with no error.

Reported in #2312.

## Fix

Replace `os.path.normpath` with `os.path.realpath` on both sides of the comparison in `derive_bank_id()`. `realpath` resolves symlinks before normalizing, so a directory and any symlink pointing to it produce the same canonical path string.

```python
# Before
normalized_cwd = os.path.normcase(os.path.normpath(cwd))
if os.path.normcase(os.path.normpath(dir_path)) == normalized_cwd:

# After
normalized_cwd = os.path.normcase(os.path.realpath(cwd))
if os.path.normcase(os.path.realpath(dir_path)) == normalized_cwd:
```

`os.path.realpath` is available on all platforms (POSIX and Windows). `normcase` is preserved for Windows drive-letter case-folding.

## Test

Added `test_directorybankmap_matches_symlinked_cwd` to `TestDirectoryBankMap` — creates a real directory and a symlink to it, verifies both paths resolve to the mapped bank. The test is skipped on platforms that don't support symlinks.

## Affected scope

- `hindsight-integrations/claude-code/scripts/lib/bank.py` — 2-line change
- `hindsight-integrations/claude-code/tests/test_bank.py` — 1 new test

### DIFF
--- a/hindsight-integrations/claude-code/scripts/lib/bank.py
+++ b/hindsight-integrations/claude-code/scripts/lib/bank.py
@@ -95,9 +95,9 @@ def derive_bank_id(hook_input: dict, config: dict) -> str:
         # case-sensitive compare silently misses the map and falls through
         # to the default bank. normcase is a no-op on POSIX, preserving
         # case-sensitive matching there.
-        normalized_cwd = os.path.normcase(os.path.normpath(cwd))
+        normalized_cwd = os.path.normcase(os.path.realpath(cwd))
         for dir_path, bank_id in dir_map.items():
-            if os.path.normcase(os.path.normpath(dir_path)) == normalized_cwd:
+            if os.path.normcase(os.path.realpath(dir_path)) == normalized_cwd:
                 return f"{prefix}-{bank_id}" if prefix else bank_id
 
     if not config.get("dynamicBankId", False):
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):

**Correct, minimal fix for #2312.** `realpath` resolves symlinks before normalizing on both sides of the comparison, so a mapped directory and any symlink to it canonicalize to the same string. `normcase` is preserved for Windows drive-letter folding, and `realpath` is cross-platform.

**Verified:**
- New `test_directorybankmap_matches_symlinked_cwd` is a genuine regression test — fails (routes to `fallback`) on the old `normpath` code, passes with `realpath`.
- Full suite green locally: `40 passed`.
- No JS counterpart in this integration (pure-Python), so no parity gap.

Nit (non-blocking): the inline `import os` / `__import__("os")` in the test is a touch unidiomatic vs. a module-level import, but it works and keeps the diff focused.

PR #2331, opened 2026-06-20
feat(claude-code): Add recall tag filters to memory hook

### LINKED ISSUE #2329: Claude Code auto-recall: expose tag filters (recallTags / tagsMatch / tagGroups) from client config
## Summary

The Claude Code plugin's auto-recall hook (`scripts/recall.py`) doesn't expose tag filtering to the client config (`~/.hindsight/claude-code.json`). When a bank defines `entity_labels` with `tag: true`, the LLM classification cost is paid at retain time but cannot be leveraged at recall time via the hook.

## Motivation

The official `client.recall(tags=..., tag_groups=..., tags_match=...)` API supports rich tag filtering — and the [best-practices §Filtering by Memory Shape](https://docs.hindsight.io/) docs explicitly recommend `entity_labels` with `tag: true` as the canonical pattern for separating semantically similar memories that serve different purposes (e.g. concise rules vs detailed procedures).

However the auto-recall hook in `hindsight-memory/<v>/scripts/recall.py` doesn't read any tag-related keys from client config and doesn't pass `tags`/`tag_groups` to `client.recall()`:

```text
$ grep -nE "tags|recall_tags|recallTags|tag_groups|tagGroups|tagsMatch" scripts/recall.py
(0 matches)
```

while `lib/client.py` does accept tag parameters in `recall()` (`tags: Optional[list] = None` around line 138).

## Use case (the one we hit)

We maintain a single normative methodology bank that all project sessions read cross-bank via `recallAdditionalBanks`. We configured `entity_labels`:

```json
{
  "entity_labels": [
    {"key": "memory_type", "type": "value", "tag": true, "values": [
      {"value": "rule", "description": "concise operating rule"},
      {"value": "procedure", "description": "step-by-step instruction"}
    ]},
    {"key": "tech_stack", "type": "multi-values", "tag": true, "values": [
      {"value": "supabase"}, {"value": "lovable"}, ...
    ]}
  ]
}
```

We wanted auto-recall to:
- Bias toward `memory_type:rule` (durable axioms) over `memory_type:procedure` (runbooks) on planning-style turns
- Filter by the current project's `tech_stack:<name>`

Today we either pay classification cost with zero recall benefit, or remove `entity_labels` entirely (our current workaround).

## Proposal

Add optional config keys to `~/.hindsight/claude-code.json` that map to the existing `client.recall()` parameters:

- `recallTags: list[str]` — tags to pass to `client.recall(tags=...)`
- `tagsMatch: "any" | "all" | "any_strict" | "all_strict"` (default `"any"` per best-practices)
- `tagGroups: <tree>` — optional nested boolean tree per recall.md §tag_groups

If those keys are present, `recall.py` passes them through to `client.recall(tags=..., tag_groups=..., tags_match=...)`. Otherwise behavior unchanged.

Per-bank overrides would be especially useful for our topology — something like `recallTagsByBank: { "<bank_id>": [...] }` so a normative bank read via `recallAdditionalBanks` can be filtered (e.g. `memory_type:rule` only) without affecting the project bank's own recall.

## Version

- plugin: hindsight-memory 0.7.0 (also reviewed 0.7.1 changelog — no mention of tag filter support)
- platform: Windows 11

Happy to send a PR if this direction is welcome. Thanks for the great plugin.

koriyoshi2041:
Summary
- Adds recallTags, tagsMatch, and tagGroups config support for the auto-recall hook.
- Passes tag filters through to primary and additional bank recall requests.
- Adds recallAdditionalBankFilters so shared banks can override tag filters independently from the project bank.
- Documents the settings and env overrides.

Tests
- uv run pytest hindsight-integrations/claude-code/tests/test_config.py hindsight-integrations/claude-code/tests/test_client.py hindsight-integrations/claude-code/tests/test_hooks.py
- uv run ruff check hindsight-integrations/claude-code/scripts/lib/config.py hindsight-integrations/claude-code/scripts/lib/client.py hindsight-integrations/claude-code/scripts/recall.py hindsight-integrations/claude-code/tests/test_config.py hindsight-integrations/claude-code/tests/test_client.py hindsight-integrations/claude-code/tests/test_hooks.py
- git diff --check

Note
- The full commit hook did not complete locally because the control-plane ESLint config could not load @eslint/js in this checkout. The focused hook tests and lint above passed.

Closes #2329


hindsight-integrations/claude-code/scripts/lib/config.py
```diff
@@ -17,6 +17,10 @@
     "recallContextTurns": 1,
     "recallMaxQueryChars": 800,
     "recallRoles": ["user", "assistant"],
+    "recallTags": [],
+    "tagsMatch": "any",
```
nicoloboschi: let's use the recall convention recallTagsMatch and recallTagGroups

nicoloboschi (changes requested):

hindsight-integrations/claude-code/scripts/lib/config.py
koriyoshi2041: Updated in 3c0f9fd9f. I switched the config/settings/docs/tests to `recallTagsMatch` and `recallTagGroups`, including the per-bank override keys.

Checked:
- `uv run pytest hindsight-integrations/claude-code/tests/test_config.py hindsight-integrations/claude-code/tests/test_hooks.py -q`
- `uv run ruff check hindsight-integrations/claude-code/scripts/lib/config.py hindsight-integrations/claude-code/scripts/recall.py hindsight-integrations/claude-code/tests/test_config.py hindsight-integrations/claude-code/tests/test_hooks.py`
- `uv run ruff format --check hindsight-integrations/claude-code/scripts/lib/config.py hindsight-integrations/claude-code/scripts/recall.py hindsight-integrations/claude-code/tests/test_config.py hindsight-integrations/claude-code/tests/test_hooks.py`

koriyoshi2041:
Quick recheck note: the requested naming change should be covered by 3c0f9fd9f now. I switched the config/settings/docs/tests to the recall convention (`recallTagsMatch`, `recallTagGroups`, plus the per-bank override keys) and the focused Claude Code hook tests/lint passed.

I think the remaining blocking review state is just waiting for a re-review, but happy to adjust if you wanted a different shape for the per-bank keys.

nicoloboschi (approved):

PR #2332, opened 2026-06-20
Add a failover LLM in case the primary one becomes unavailable

chrislatimer:
## Summary

Adds an **optional cross-provider failover** for the engine's LLM calls. When the primary provider exhausts its retry budget and raises, the  same call is re-dispatched to a second provider with identical arguments. Disabled by default — when none of the four env vars are set, the hot path is byte-identical to the pre-failover behaviour.

## Motivation

Provider outages, regional rate limits, and 5xx blips currently surface as user-facing 5xx on retain/reflect/consolidation, even though the same prompt would succeed on a different provider. This PR adds a single server-level failover so deployments can ride through provider-side incidents without per-call orchestration.

## What's in the PR
  - **`engine/failover_llm.py`** — new `FailoverLLMProvider` composite (and `ConfiguredFailoverLLMProvider` for the per-bank `with_config()`  path). Wraps a primary + failover, re-dispatches on `Exception`, propagates `OutputTooLongError` and `CancelledError` / `KeyboardInterrupt` /  `SystemExit` unchanged. Attribute access falls through to the primary so existing reads of `.provider` / `.model` / `._client` keep working.

  - **`engine/memory_engine.py`** — `_maybe_wrap()` wraps each of the four per-operation `LLMConfig`s (default, retain, reflect, consolidation) in the composite when failover is configured; returns the bare `LLMConfig` otherwise. A single failover instance is shared across all four composites. `verify_connection` is hard on the primary and soft on the failover so a misconfigured failover cannot block startup. 

  - **`config.py`** — four new env vars (provider / api_key / model / base_url), all `None` by default. Marked as sensitive in the config's redaction sets. Server-level only (not in `_CONFIGURABLE_FIELDS`).

  - **`engine/retain/fact_extraction.py`** — `max(1, llm_max_retries)` so the extraction loop always makes at least one attempt (previously `max_retries=0` skipped the call entirely and raised the fall-through `RuntimeError`). Failed-chunk summary now includes the inner error message and logs full inner tracebacks instead of hiding them behind a wrapping `RuntimeError`.
 
  - **Tests** — `tests/test_failover_llm.py` (composite behaviour, error classification, cancellation semantics, config plumbing) and `tests/test_failover_llm_memory_engine.py` (engine wiring with failover on/off).

  - **Docs** — new "Failover LLM Provider" section in `developer/configuration.md` (and the agent reference copy), entries in `.env.example` and the embed env template.

  ## Behaviour notes

  - **What triggers failover:** rate limits, 5xx, connection errors, schema-validation failures from soft JSON mode — anything inheriting from `Exception`.

  - **What does NOT:** `OutputTooLongError` (failover won't help — the prompt simply exceeds the budget) and `CancelledError` / `KeyboardInterrupt` / `SystemExit` (propagate so cancellation works).

  - **Retry semantics:** `HINDSIGHT_API_LLM_MAX_RETRIES` (and per-op variants) still controls primary retries. Failover only fires once the primary has exhausted them.

  - **Batch APIs:** not in scope — `submit_batch` / `get_batch_status` / `retrieve_batch_results` do not failover.

  - **Backwards compatibility:** with all four vars unset, MemoryEngine constructs bare `LLMConfig`s exactly as before; no composite is wrapped.

## Configuration

  | Variable | Description | Default |
  |---|---|---|
  | `HINDSIGHT_API_LLM_FAILOVER_PROVIDER` | Provider name (`openai`, `anthropic`, `gemini`, `groq`, etc.). Unset disables failover. | unset |
  | `HINDSIGHT_API_LLM_FAILOVER_API_KEY` | API key for the failover provider. Required unless the provider is no-auth (`claude-code`, etc.). |  unset |
  | `HINDSIGHT_API_LLM_FAILOVER_MODEL` | Model name for the failover. | provider default |
  | `HINDSIGHT_API_LLM_FAILOVER_BASE_URL` | Base URL for the failover provider. | provider default |

  ## Test plan

  - [ ] `cd hindsight-api-slim && uv run pytest tests/test_failover_llm.py tests/test_failover_llm_memory_engine.py -v`
  - [ ] `cd hindsight-api-slim && uv run pytest tests/` (full suite, ensure no regressions in retain / reflect / consolidation paths)
  - [ ] `cd hindsight-api-slim && uv run ruff check . && uv run ty check hindsight_api/`
  - [ ] Manual: with `HINDSIGHT_API_LLM_FAILOVER_PROVIDER` unset, confirm logs / behaviour are unchanged.
  - [ ] Manual: set a bogus primary (e.g. fake API key) and a real failover; trigger a retain and confirm the WARNING log fires and the request succeeds via the failover.
  - [ ] Manual: misconfigure failover (bad key) and confirm startup logs a soft warning but does not exit.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,15 @@ HINDSIGHT_API_LLM_BASE_URL=https://api.openai.com/v1
 # Reasoning effort for providers/models that support it. Examples: low, medium, high, xhigh.
 # HINDSIGHT_API_LLM_REASONING_EFFORT=low
 
+# Failover LLM provider (optional). When set, calls that fail after the primary
+# exhausts HINDSIGHT_API_LLM_MAX_RETRIES are re-dispatched to this provider.
+# Backwards compatible: leave all four unset to keep behaviour identical to
+# pre-failover releases. See docs/developer/configuration.md for the full table.
+# HINDSIGHT_API_LLM_FAILOVER_PROVIDER=anthropic
+# HINDSIGHT_API_LLM_FAILOVER_API_KEY=
+# HINDSIGHT_API_LLM_FAILOVER_MODEL=claude-3-5-sonnet-latest
+# HINDSIGHT_API_LLM_FAILOVER_BASE_URL=
+
 # Example: Anthropic Claude configuration
 # HINDSIGHT_API_LLM_PROVIDER=anthropic
 # HINDSIGHT_API_LLM_API_KEY=your-anthropic-api-key
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,9 @@ logs/
 # Generated docs files
 hindsight-docs/static/llms-full.txt
 
+# Local-only superpowers plan/scratch dir
+docs/superpowers/
+
 
 hindsight-dev/benchmarks/locomo/results/
 hindsight-dev/benchmarks/longmemeval/results/
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -156,6 +156,15 @@ ENV_LLM_SEND_BANK_AS_USER = "HINDSIGHT_API_LLM_SEND_BANK_AS_USER"
 # disambiguates from the embeddings/reranker LITELLM_* settings.
 ENV_LLM_LITELLMROUTER_CONFIG = "HINDSIGHT_API_LLM_LITELLMROUTER_CONFIG"
 
+# Failover LLM provider (optional, server-level only). When set, calls that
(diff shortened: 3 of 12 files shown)


nicoloboschi:
Superseding this with #2365, which generalizes the idea: instead of a single `HINDSIGHT_API_LLM_FAILOVER_*` secondary, you configure N LLMs by index (`HINDSIGHT_API_LLM_<n>_*`) and pick a strategy via `HINDSIGHT_API_LLM_STRATEGY` JSON — `failover` or `round-robin` (with optional weights), per-operation overridable. Thanks for the original idea here — closing in favor of #2365.

PR #2337, opened 2026-06-21
fix(stats): invalidate bank stats cache on unit/document deletes and observation clears

r266-tech:
## Summary

`get_bank_stats` is served from a 60s-TTL `BankStatsCache`. `delete_bank` invalidates it after mutating counts (with a comment to that effect), but the other operations that change the same counts did not — so a client polling stats right after a deletion saw pre-mutation node/link/document/observation counts for up to a minute.

Follow-up to #2315, which hardened the cache primitive (`bank_stats_cache.py`) but did not touch the `memory_engine.py` mutation call sites. This wires the same `delete_bank`-style invalidation into every path that changes the counts `get_bank_stats` reports:

- `delete_memory_unit` — on an actual delete (cascades to links/entities)
- `delete_document` — on an actual delete (cascades to memory units/links)
- `clear_observations` — bulk observation delete + consolidation-timestamp reset
- `update_document` — when a tag change deletes observations (`invalidated_obs > 0`)

High-frequency inserts intentionally keep tolerating TTL staleness (only `delete_bank` invalidated before); this matches that intent by invalidating on the explicit user-initiated delete/clear paths only. Invalidation is best-effort (guarded `try/except` + warning, like the other post-commit side-effects in these methods), so a cache-eviction failure can never fail an already-committed operation. `delete_bank` is intentionally left untouched.

## Tests

`tests/test_bank_stats_cache_invalidation.py` pins a long cache TTL, warms `get_bank_stats`, mutates, and asserts the next read reflects the change immediately — without the fix the long-TTL cache would still serve the stale value. Covers `delete_memory_unit`, `delete_document`, and `clear_observations`.

If this has already been addressed via a separate patch, please feel free to close.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -5620,7 +5620,9 @@ class MemoryEngine(MemoryEngineInterface):
             try:
                 await self._bank_stats_cache.invalidate(get_current_schema(), bank_id)
             except Exception as e:
-                logger.warning(f"Failed to invalidate bank stats cache after memory unit deletion for bank {bank_id}: {e}")
+                logger.warning(
+                    f"Failed to invalidate bank stats cache after memory unit deletion for bank {bank_id}: {e}"
+                )
 
         if bank_id_for_consolidation:
             config = await self._config_resolver.resolve_full_config(bank_id_for_consolidation, request_context)
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -5334,6 +5334,17 @@ class MemoryEngine(MemoryEngineInterface):
                     "memory_units_deleted": units_count if deleted else 0,
                 }
 
+        # Drop any cached stats for this bank — deleting the document changed
+        # the document count and (via cascade) the memory-unit/link counts
+        # get_bank_stats reports, which the TTL would otherwise serve at
+        # pre-delete values for up to a minute (mirrors delete_bank). Best-effort:
+        # a cache-eviction failure must not fail an already-committed delete.
+        if deleted:
+            try:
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Quick CI note: `verify-generated-files` is failing after `./scripts/hooks/lint.sh` because ruff rewrites two files, not because a generated API/client artifact changed.

The formatter diff is tiny:
- wraps the long `logger.warning(...)` line in `hindsight_api/engine/memory_engine.py`
- collapses the `test_delete_document_invalidates_stats_cache` signature in `tests/test_bank_stats_cache_invalidation.py`

I reproduced the formatter part locally. After applying that ruff output, these pass:

```bash
cd hindsight-api-slim
uv run ruff check hindsight_api/engine/memory_engine.py tests/test_bank_stats_cache_invalidation.py
uv run ruff format --check hindsight_api/engine/memory_engine.py tests/test_bank_stats_cache_invalidation.py
```

So committing the ruff-format output should clear this particular CI failure. I tried the focused pytest too, but my local run fails during `SentenceTransformer` setup because this machine has a SOCKS proxy configured and `httpx[socks]`/`socksio` is not installed; I would not treat that as a PR signal.

r266-tech:
Thanks @koriyoshi2041 — spot-on diagnosis. Pushed the `ruff format` output as a single commit (f5d22bb): wrapped the long `logger.warning(...)` in `memory_engine.py` and collapsed the `test_delete_document_invalidates_stats_cache` signature. `ruff format --check` is clean on both files now, and the change is format-only (no logic / import-order touched), so `verify-generated-files` should go green on the next run.

r266-tech:
Thanks for the precise diagnosis and the local repro, @koriyoshi2041 — you're right that it was purely `ruff format` rewriting those two files, not a generated-artifact change. I pushed the formatter output in `f5d22bb` and `verify-generated-files` is green now. And agreed: the `SentenceTransformer`/SOCKS `httpx[socks]` failure is a local-env artifact, not a PR signal. Appreciate the help!

nicoloboschi (approved):

PR #2345, opened 2026-06-22
fix(reflect): unwrap JSON answer envelopes

### LINKED ISSUE #2298: reflect: model emits JSON envelope as answer text on 0.8.3
### Symptom

Reflect occasionally returns a JSON envelope as the answer text instead of a plain markdown answer. Example payload (`text` field):

```json
{
  "answer": "...",
  "directive_compliance": "Directive 1: ... Directive 2: ... Directive 3: ...",
  "memory_ids": ["45ab116d-...", "..."],
  "mental_model_ids": [],
  "observation_ids": []
}
```

`directive_compliance` isn't a Hindsight field — the model is inventing a JSON envelope to "prove" it followed instructions. The outer API response is well-formed; the inner content is the model going off-script.

### Repro state

- Hindsight `0.8.3`
- Happens **with and without** reflect directives (directives amplify the rate but aren't the root cause)
- Re-running the same query with a narrower scope sometimes returns clean markdown
- User workaround: a directive forcing single quotes — makes the JSON syntactically wrong and drops the rate measurably, but not to 0
- Reporter's directives (when present) included: a "format" directive opening with `ALWAYS` + bullet list; a no-hallucination rule; a long PlantUML rule list with multiple negative imperatives ("Do NOT…", "Never…", "Every arrow must…"). Heavy compliance-checklist framing.

### Hypothesis

The reflect prompt assembly doesn't anchor output format strongly enough. When the query is structured (asks for tables, separate aspects, etc.) and/or directives read like compliance rules, some models default to emitting a JSON "report" instead of plain markdown. The reporter's experience that single-quote enforcement helps is consistent — single quotes aren't valid JSON, so the model is less likely to commit to the envelope.

### Needed from reporter

- LLM provider + model
- Bank `background` (if set)
- Full query text alongside the directives (already captured above)

### Possible fix angles

- Stronger output-format anchor in the reflect system prompt — explicit "respond in plain markdown, never wrap your response in a JSON object" near the end
- Defensive post-processing: detect a top-level JSON object whose keys look like our shape (`answer` + `memory_ids` + `…_ids`) and unwrap `.answer` into `text` (band-aid, but recovers gracefully)
- Guidance in docs that reflect directives should be phrased as declarative instructions, not compliance rules ("Return …", "Use …") to reduce envelope-mode hallucinations

cc @nicoloboschi

## Steps to Reproduce

1. Have a bank with content that can answer a structured query (e.g. interface inventory split into inbound/outbound).
2. (Optional, amplifies) Set 1–3 reflect directives that read like compliance rules — `ALWAYS …`, `Do NOT …`, `Every X must …`, numbered/bulleted format rules.
3. Call reflect with a query that asks for structured output ("return a markdown doc with separate tables for inbound and outbound interfaces…").
4. Inspect the `text` field on the response.

Repeats intermittently. Re-running with a narrower scope, or with all directives removed, lowers the rate but does not eliminate it.

## Expected Behavior

The `text` field contains a plain markdown answer rendered against the reflect prompt and directives.

## Actual Behavior

The `text` field contains a JSON-shaped string the model invented, e.g.:

```json
{
  "answer": "...",
  "directive_compliance": "Directive 1: ... Directive 2: ...",
  "memory_ids": ["45ab116d-...", "..."],
  "mental_model_ids": [],
  "observation_ids": []
}
```

`directive_compliance` is not a Hindsight field. The outer API response is well-formed; the model has wrapped its real answer inside a self-invented JSON envelope.

## Version

0.8.3

## LLM Provider

Not yet confirmed by reporter — will update once known.

xmh1011:
Summary
- unwrap reflect responses when a model returns the done-tool JSON envelope as visible answer text
- keep normal JSON answers intact by requiring reflect envelope keys and tightening trailing id cleanup
- strengthen the reflect tool prompt against JSON-envelope answers

Closes #2298

Tests
- uv run --no-sync pytest -q tests/test_reflect_agent.py::TestCleanAnswerText tests/test_reflect_agent.py::TestCleanDoneAnswer tests/test_reflect_prompt_builder.py
- uv run --no-sync ruff check hindsight_api/engine/reflect/agent.py hindsight_api/engine/reflect/prompts.py tests/test_reflect_agent.py tests/test_reflect_prompt_builder.py
- uv run --no-sync ruff format --check hindsight_api/engine/reflect/agent.py hindsight_api/engine/reflect/prompts.py tests/test_reflect_agent.py tests/test_reflect_prompt_builder.py
- pre-commit hooks completed during git commit

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -95,7 +95,7 @@ _TRAILING_IDS_PATTERN = re.compile(
 )
 _JSON_CODE_FENCE_PATTERN = re.compile(r"^\s*```(?:json)?\s*(\{.*\})\s*```\s*$", re.DOTALL | re.IGNORECASE)
 
-_REFLECT_ENVELOPE_KEYS = frozenset(
+_DONE_ARGUMENT_KEYS = frozenset(
     {
         "answer",
         "directive_compliance",
@@ -105,16 +105,17 @@ _REFLECT_ENVELOPE_KEYS = frozenset(
         "model_ids",
     }
 )
-_REFLECT_ENVELOPE_MARKER_KEYS = _REFLECT_ENVELOPE_KEYS - {"answer"}
+_DONE_ARGUMENT_MARKER_KEYS = _DONE_ARGUMENT_KEYS - {"answer"}
 _LEAKED_JSON_ID_KEYS = frozenset({"memory_ids", "mental_model_ids", "observation_ids", "model_ids"})
 
 
-def _unwrap_reflect_answer_envelope(text: str) -> str | None:
-    """Return the answer from a model-invented reflect JSON envelope.
+def _unwrap_leaked_done_arguments(text: str) -> str | None:
+    """Return the answer when a done tool call was rendered as JSON text.
 
-    Some models return the done-tool argument shape as user-visible text, e.g.
-    {"answer": "...", "memory_ids": [...]}.  Only unwrap objects that look like
-    that reflect envelope so normal JSON answers stay intact.
+    Some providers leak the done tool's argument object instead of surfacing it
+    as a native tool call, e.g. {"answer": "...", "memory_ids": [...]}. Only
+    unwrap objects that match the done argument shape so normal JSON answers
+    stay intact.
     """
     candidate = text.strip()
     if not candidate:
@@ -136,9 +137,9 @@ def _unwrap_reflect_answer_envelope(text: str) -> str | None:
         return None
 
     keys = set(payload)
-    if not keys.intersection(_REFLECT_ENVELOPE_MARKER_KEYS):
+    if not keys.intersection(_DONE_ARGUMENT_MARKER_KEYS):
         return None
-    if not keys.issubset(_REFLECT_ENVELOPE_KEYS):
+    if not keys.issubset(_DONE_ARGUMENT_KEYS):
         return None
 
     for key in ("memory_ids", "mental_model_ids", "observation_ids", "model_ids"):
@@ -178,7 +179,7 @@ def _clean_answer_text(text: str) -> str:
     Some LLMs output the done() call as text instead of a proper tool call.
     This strips out patterns like: done({"answer": "...", ...})
     """
-    unwrapped = _unwrap_reflect_answer_envelope(text)
+    unwrapped = _unwrap_leaked_done_arguments(text)
     if unwrapped is not None:
         return unwrapped
 
@@ -200,7 +201,7 @@ def _clean_done_answer(text: str) -> str:
     if not text:
         return text
 
-    unwrapped = _unwrap_reflect_answer_envelope(text)
+    unwrapped = _unwrap_leaked_done_arguments(text)
     if unwrapped is not None:
         return unwrapped
 
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -90,12 +90,86 @@ _LEAKED_JSON_SUFFIX = re.compile(
     r'\s*```(?:json)?\s*\{[^}]*(?:"(?:observation_ids|memory_ids|mental_model_ids)"|\})\s*```\s*$',
     re.DOTALL | re.IGNORECASE,
 )
-_LEAKED_JSON_OBJECT = re.compile(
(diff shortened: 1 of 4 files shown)


nicoloboschi (commented):
I tried to reproduce #2298 against the exact model from the issue (`ollama-cloud/gpt-oss:120b-cloud`, per @benfrank241's comment) before reviewing, and the investigation changes what the right fix looks like.

## Reproduction: 35/35 clean runs, no envelope

I replicated the **exact** reflect LLM call — real `build_system_prompt_for_tools` + the real `done` tool schema (including the directive-compliance field from `_build_done_tool_with_directives`) — and ran a minimal tool loop against `gpt-oss:120b-cloud`:

| Effort | Runs | Envelope leaks | Clean native `done()` |
|--------|------|----------------|------------------------|
| low | 6 | 0 | 6 |
| medium | 8 + 15 (heavy "prove per-directive compliance or be REJECTED" directives) | 0 | 23 |
| high | 6 | 0 | 6 |

Every run produced a proper **native** `done` tool call with clean markdown in `answer`. I could not trigger the JSON envelope even with aggressive compliance-checklist directives. (Tested through a current ollama; harmony tool-call parsing surfaces the call natively.)

## The root-cause framing in #2298 is wrong

`directive_compliance` is **not** a field the model invents to "prove" compliance. It's a **real, required Hindsight field** — `tools_schema.py::_build_done_tool_with_directives()` adds it to the `done` tool whenever directives are present. So the envelope the reporter saw:

```json
{"answer": "...", "directive_compliance": "...", "memory_ids": [...], "mental_model_ids": [], "observation_ids": []}
```

is **literally the `done` tool-call arguments**, serialized as text instead of delivered as a native tool call. This is a **tool-call delivery/parsing issue** specific to gpt-oss's harmony format (the repo already band-aids this — see the `done<|channel|>commentary` handling in `_normalize_tool_name` and the `done({...})`-as-text regex `_DONE_CALL_PATTERN`). It is **not** the prompt-anchoring problem the issue hypothesizes. The leak appears to be environment-specific (older ollama / the ollama.com endpoint where the harmony channel isn't parsed into native tool calls).

## What that means for this PR

- **The unwrap post-processing is the genuinely useful part.** It recovers the answer from a leaked done-args envelope at `_clean_answer_text` (the no-native-tool text path) and `_clean_done_answer` — exactly where a leaked call lands. Including `directive_compliance` in the allowlist is **correct** (it's a real schema field, not overfitting), and the subset guard sensibly avoids clobbering legitimate JSON answers. This is the same family of recovery shim the repo already ships for harmony leaks. I'd suggest renaming/recommenting it to reflect the real mechanism — *"recover a `done` tool-call that was rendered as text"* — rather than "model-invented envelope," so the next reader isn't misled.

- **The prompt line is the weak half, and slightly self-contradictory.** It instructs the model to *"never include fields like `directive_compliance` ... in the answer text,"* while the `done` tool schema in the **same** prompt *requires* the model to produce `directive_compliance`. Since the actual failure is at the parsing layer (a tool call that wasn't parsed as one), a generation-side instruction is aimed at the wrong layer — not harmful, but unlikely to be what fixes the reported symptom. I'd drop or soften it.

## Suggestion

Land the unwrap shim (it's a reasonable, correctly-shaped band-aid for a provider leak that's outside this codebase to fix properly), but reframe the PR/issue away from "the model invents a JSON envelope to prove compliance." The durable fix is at the harmony tool-call parsing layer / an ollama version floor; worth a note in the issue since the symptom isn't reproducible on a current parser.

xmh1011:
@nicoloboschi  Addressed in c8cace024.

- Reframed the recovery helper/tests around leaked `done` tool-call arguments rendered as JSON text, including `directive_compliance` as a real done-schema field.
- Removed the new prompt instruction that discouraged `directive_compliance`/done fields in answer text, since that conflicted with the required tool schema and the useful fix is the parser-side recovery shim.
- Re-ran targeted reflect tests and lint.

Validation:
- `uv run --no-sync pytest -q tests/test_reflect_agent.py::TestCleanAnswerText tests/test_reflect_agent.py::TestCleanDoneAnswer tests/test_reflect_prompt_builder.py`
- `uv run --no-sync ruff check hindsight_api/engine/reflect/agent.py hindsight_api/engine/reflect/prompts.py tests/test_reflect_agent.py tests/test_reflect_prompt_builder.py`
- `uv run --no-sync ruff format --check hindsight_api/engine/reflect/agent.py hindsight_api/engine/reflect/prompts.py tests/test_reflect_agent.py tests/test_reflect_prompt_builder.py`
- `./scripts/hooks/lint.sh`

nicoloboschi (approved):

PR #2346, opened 2026-06-22
fix(control-plane): show pending uploaded documents

### LINKED ISSUE #2314: Uploaded documents show in list while being retained
### Use Case

When uploading a document for the first time via the Control Plane -> Add Document -> Submit, The dialog closes and no feedback is given that the document has been uploaded (The list doesn't update). Only after going to bank configuration do you see the file is being retained, and after completion, the file then appears in the document section.

### Problem Statement

Currently I upload a document and go to the bank configuration to see which file I've uploaded (that are being processed). I loose track of which files have been uploaded.

Uploaded documents are shown immediately in the document section with zero facts / observations made. As they are retained in the background, they get facts & observations added.

### Proposed Solution

Document gets added to the documents list first and perhaps shows its being retained via a UI badge.

### Alternatives Considered

_No response_

### Priority

Nice to have

### Additional Context

_No response_

### Checklist

xmh1011:
Summary
- generate document IDs for uploaded files before submitting retain, so the Documents view can show each accepted file immediately
- add a client-side pending-document store backed by sessionStorage
- render Processing/Failed rows in Documents until list_documents returns the real document row
- poll the related file_convert_retain operation to surface conversion failures

Closes #2314

Tests
- npm run --workspace @vectorize-io/hindsight-control-plane i18n:check
- npx prettier --check src/components/bank-selector.tsx src/components/documents-view.tsx src/lib/pending-documents.ts src/messages/en.json src/messages/zh-CN.json src/messages/zh-TW.json src/messages/yue-Hant.json src/messages/de.json src/messages/es.json src/messages/fr.json src/messages/ja.json src/messages/ko.json src/messages/pt.json
- npx eslint src/components/bank-selector.tsx src/components/documents-view.tsx src/lib/pending-documents.ts (0 errors; existing warnings only)
- npm exec --workspace @vectorize-io/hindsight-control-plane -- tsc --noEmit --pretty false (fails on existing generated SDK mismatches: testBankLlm, consolidationState, exclude_parents)
- pre-commit hooks completed during git commit


nicoloboschi:
Thanks a lot for this @xmh1011 — the UX you proposed here (showing accepted files as `Processing`/`Failed` rows until the real document lands) is exactly what we want, and the i18n coverage was a nice touch.

We've decided to land this server-driven instead of client-side. Rather than a `sessionStorage` store with client-generated document ids, the new version derives the pending rows from the existing `file_convert_retain` operations (the server already stores the `document_id` and original filename in the operation), so the status survives reloads/tabs/devices and there's no client-side id generation to keep in sync.

Superseded by #2420 (which closes #2314). Closing this one — really appreciate the contribution and the clear write-up that made the follow-up easy. 🙏

PR #2349, opened 2026-06-22
fix(config): validate disposition_* range on bank-config write so one malformed bank can't 500 the whole bank list (#2348)

### LINKED ISSUE #2348: PATCH bank config accepts out-of-range/wrong-type disposition_* values, then 500s the entire list endpoint
## Summary

A per-bank `PATCH /v1/default/banks/{bank_id}/config` accepts **out-of-contract values** for `disposition_skepticism` / `disposition_literalism` / `disposition_empathy` (e.g. floats, or ints outside 1–5). The write path validates field *names* but not value *types/ranges*, so the bad value is persisted to the `banks.config` JSONB. The read path is strict (`DispositionTraits`, `int`, `ge=1, le=5`), so it then raises a `ValidationError` on serialization — and because it surfaces in the **list** response, a single malformed bank **500s `GET /v1/default/banks` for the entire tenant**, not just that one bank.

Observed in production (v0.8.3): one bank had `{disposition_skepticism: 0.7, disposition_literalism: 0.8, disposition_empathy: 0.3}` in its config override (a 0–1 scale mistakenly used instead of 1–5). `GET /v1/default/banks` and `GET /v1/default/banks/{id}/profile` both returned 500:

```
3 validation errors for BankListResponse
banks.103.disposition.skepticism
  Input should be a valid integer, got a number with a fractional part [type=int_from_float, input_value=0.7, ...]
banks.103.disposition.literalism  ... input_value=0.8
banks.103.disposition.empathy     ... input_value=0.3
```

The `banks.disposition` *column* was a clean `{3,3,3}`; the floats lived only in the `config` override and entered the resolved disposition via the config overlay.

## Root cause (write side is permissive, read side is strict)

1. **Write accepts `Any`.** The PATCH request model `BankConfigUpdate.updates` is typed `dict[str, Any]`, so no per-field validation happens at the HTTP layer.
2. **`ConfigResolver.update_bank_config` validates only field *names*** — credential rejection, the configurable-field allowlist, permissions, and value validators for `entity_labels` / `retain_strategies` / `recall_budget` / chunking. There is **no type or range check for `disposition_*`** (or most other bounded scalar fields). The raw dict is `json.dumps`'d straight into `banks.config`.
3. **The type contract exists but is never enforced on this path.** `HindsightConfig` declares `disposition_skepticism: int | None` etc., but dataclass annotations aren't validated at runtime and the write path never constructs the dataclass. (The env path *does* coerce, via `int(os.getenv(...))` — only the per-bank config API is unguarded.)
4. **Read overlays without coercion.** `get_bank_profile` overlays `config_dict["disposition_*"]` over the DB column and passes the value straight to `DispositionTraits` (`int`, `ge=1, le=5`).
5. **Blast radius amplified.** The failure is raised while building `BankListResponse`, so one bad bank takes down the whole list endpoint for the tenant.

## Reproduction

```bash
# Write an out-of-range / wrong-type disposition override
curl -X PATCH .../v1/default/banks/<bank>/config \
  -H 'Authorization: Bearer <key>' -H 'Content-Type: application/json' \
  -d '{"updates": {"disposition_skepticism": 0.7, "disposition_literalism": 0.8, "disposition_empathy": 0.3}}'
# -> 200 OK (accepted & persisted)

curl .../v1/default/banks   # -> 500, validation error; list endpoint down for ALL banks in the tenant
```

## Suggested fixes (any/all)

- **Validate on write:** in `update_bank_config`, coerce/range-check `disposition_skepticism|literalism|empathy` to `int` in `[1,5]` and `422` on violation (consistent with the existing per-field validators for `entity_labels`, `recall_budget`, etc.).
- **Coerce/clamp on read:** in the `get_bank_profile` disposition overlay, `round()` + clamp to `[1,5]` so legacy bad data degrades gracefully instead of 500ing.
- **Fail soft on list:** building `BankListResponse` should not let one malformed bank 500 the whole page (skip/clamp the offending bank and log, rather than aborting the list).

The first item is the actual fix; the second and third are defense-in-depth so existing bad rows don't keep bricking the list endpoint.

## Environment
- hindsight-api v0.8.3

r266-tech:
## Problem

Closes #2348.

`PATCH /v1/{tenant}/banks/{id}/config` validates field **names** only — it never checks scalar type/range. So an out-of-contract `disposition_skepticism` / `disposition_literalism` / `disposition_empathy` (a float, a 0-1 scale, or an int outside 1-5) is accepted and `json.dumps`-ed verbatim into `banks.config` JSONB. The read overlay then injects it into a strict `DispositionTraits(skepticism/literalism/empathy: int, ge=1, le=5)`. Because that model is built while serializing `BankListResponse`, a **single** malformed bank 500s `GET /v1/{tenant}/banks` for the **entire tenant** (and the per-bank profile) — a self-hoster can't list or manage *any* bank until manual JSONB surgery. This is the reported v0.8.3 case (a 0-1 scale used by mistake).

`update_bank_config` already validates names, credentials, permissions, entity_labels, retain_strategies and recall_budget, but has no disposition value check — the per-bank API is the only unguarded numeric entry point (the env path coerces via `int(os.getenv(...))`).

## Fix

Add a write-side `_validate_disposition_updates` that raises `ValueError` (the route already maps `ValueError → 400`), called from `update_bank_config` right after `_validate_recall_budget_updates`. It mirrors the existing, accepted `_validate_recall_budget_updates` shape exactly: rejects floats (incl. in-range `3.0`), bools, strings, and ints outside 1-5; accepts ints 1-5. Plus a unit test mirroring `tests/test_recall_budget_config.py::TestValidateRecallBudgetUpdates`.

**On `None`:** the validator allows `None` as the "clear this per-bank override" sentinel (the field is `int | None`). This is safe and does **not** reintroduce the 500: the read overlay `_overlay_bank_config_disposition_mission` resolves each trait as `cfg_value if cfg_value is not None else <legacy banks.disposition column>`, so a stored `null` falls back to the bank's legacy disposition and never reaches `DispositionTraits` — it can't poison the list.

## Scope

This closes the issue as stated — the write path is the entry point that lets bad values in. Already-poisoned **legacy** rows (written before this guard) would still need either a one-time cleanup or read-side hardening; I kept a read-side clamp/round out of this PR since "reject-at-write + migrate" vs "clamp-on-read" is a design call. Happy to add a clamp in `_overlay_bank_config_disposition_mission` as a follow-up if you'd prefer the list to self-heal poisoned rows.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config_resolver.py
+++ b/hindsight-api-slim/hindsight_api/config_resolver.py
@@ -402,6 +402,9 @@ class ConfigResolver:
         # Validate recall budget fields
         _validate_recall_budget_updates(normalized_updates)
 
+        # Validate disposition trait fields (1-5 integer scale)
+        _validate_disposition_updates(normalized_updates)
+
         chunking_fields_updated = (
             "retain_chunk_size" in normalized_updates
             or "retain_structured_chunk_size" in normalized_updates
@@ -514,6 +517,31 @@ def _validate_recall_budget_updates(updates: dict[str, Any]) -> None:
                 f"recall_budget_min ({updates['recall_budget_min']}) must be <= "
                 f"recall_budget_max ({updates['recall_budget_max']})"
             )
+
+
+_DISPOSITION_KEYS = (
+    "disposition_skepticism",
+    "disposition_literalism",
+    "disposition_empathy",
+)
+
+
+def _validate_disposition_updates(updates: dict[str, Any]) -> None:
+    """Validate disposition trait config updates. Raises ValueError on invalid input.
+
+    Each trait is an integer on a 1-5 scale (or None to clear the per-bank
+    override). The read overlay injects the stored value verbatim into a strict
+    ``DispositionTraits(int, ge=1, le=5)``; an out-of-contract value (a float, a
+    0-1 scale, or an int outside 1-5) accepted here would later 500 the whole
+    bank list when any bank profile is serialized (issue #2348).
+    """
+    for key in _DISPOSITION_KEYS:
+        if key in updates:
+            value = updates[key]
+            if value is None:
+                continue
+            if not isinstance(value, int) or isinstance(value, bool) or not (1 <= value <= 5):
+                raise ValueError(f"{key} must be an integer between 1 and 5, got {value!r}")
 
 
 def apply_strategy(config: HindsightConfig, strategy_name: str) -> HindsightConfig:
--- a/hindsight-api-slim/tests/test_disposition_config.py
+++ b/hindsight-api-slim/tests/test_disposition_config.py
@@ -1,0 +1,64 @@
+"""Tests for write-side validation of bank disposition config overrides.
+
+Disposition traits (skepticism / literalism / empathy) are integers on a 1-5
+scale. The ``PATCH /v1/{tenant}/banks/{id}/config`` write path must reject
+out-of-contract values (floats, 0-1 scales, ints outside 1-5) at write time;
+otherwise a single malformed bank 500s the entire bank list because the read
+overlay injects the stored value verbatim into a strict
+``DispositionTraits(int, ge=1, le=5)``. See issue #2348.
+"""
+
+import pytest
+
+from hindsight_api.config_resolver import _validate_disposition_updates
+
+
+_DISPOSITION_FIELD_NAMES = (
+    "disposition_skepticism",
+    "disposition_literalism",
+    "disposition_empathy",
+)
+
+
+class TestValidateDispositionUpdates:
+    def test_no_op_passes(self):
+        _validate_disposition_updates({})
+        _validate_disposition_updates({"unrelated_field": 123})
+
+    def test_valid_in_range_integers_pass(self):
+        for key in _DISPOSITION_FIELD_NAMES:
+            for value in (1, 2, 3, 4, 5):
+                _validate_disposition_updates({key: value})
+
+    def test_none_clears_override(self):
+        # None is the "unset this per-bank override" sentinel (field is int | None).
+        for key in _DISPOSITION_FIELD_NAMES:
+            _validate_disposition_updates({key: None})
+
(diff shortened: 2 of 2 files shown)


koriyoshi2041:
I checked the narrow path locally. The new disposition validator tests and the adjacent recall-budget validator tests pass for me:

```bash
cd hindsight-api-slim
uv run pytest tests/test_disposition_config.py tests/test_recall_budget_config.py::TestValidateRecallBudgetUpdates
```

Result: `14 passed`.

One tiny style note from local ruff:

```bash
uv run ruff check --diff tests/test_disposition_config.py
```

It only wants to remove the extra blank line after the import block in `tests/test_disposition_config.py`. I also tried the existing list overlay test, but my local run hit the same environment issue I saw yesterday: `SentenceTransformer` goes through a SOCKS proxy and this venv lacks `socksio`, so I would not treat that as a PR signal.

The write-side boundary here looks right to me: `None` is kept as a clear-override sentinel, while floats/bools/strings/out-of-range ints get rejected before they can poison `BankListResponse`.

nicoloboschi (approved):
Verified the diagnosis and fix against the code:
- `BankConfigUpdate.updates` is a raw `dict[str, Any]` — the one unguarded numeric entry point (deprecated bank/template paths already carry Pydantic ge=1/le=5).
- `DispositionTraits` is strict int 1-5, built per-bank during list serialization → one poisoned bank 500s the whole tenant list, as described.
- `None` is safe: the read overlay falls back to the legacy disposition column, so it never reaches `DispositionTraits`.
- Placement (after recall-budget validation, before persist) and `ValueError → 400` mapping are correct; centralizing in `update_bank_config` covers all callers.
- Logic is sound on all boundary cases (bool/float/out-of-range/string rejected, string doesn't TypeError due to short-circuit). Mirrors the accepted recall-budget pattern.

Follow-up worth tracking: already-poisoned legacy rows still need a read-side clamp or one-time cleanup to self-heal. Approving as the write-side fix closes the issue as stated.

PR #2353, opened 2026-06-22
fix(graph-maintenance): sort unit_ids to eliminate concurrent-insert deadlock

cdbartholomew:
## Summary

`enqueue_graph_maintenance` (in both the PostgreSQL and Oracle data-access
ops impls) issues an `INSERT INTO graph_maintenance_queue (bank_id, unit_id)`
with `ON CONFLICT DO NOTHING` (PG) / `IGNORE_ROW_ON_DUPKEY_INDEX` (Oracle).
The unique-key check takes a short-lived row-level lock per `(bank_id,
unit_id)` being inserted.

When two transactions on the same bank execute concurrently with
overlapping but differently-ordered `unit_ids` arrays, the per-row locks
get acquired in opposite orders. Postgres detects the cycle and aborts
one transaction with `asyncpg.exceptions.DeadlockDetectedError`, which
the FastAPI layer surfaces as an opaque 500.

This is straightforward to hit from two concurrent
`PATCH /v1/default/banks/{bank_id}/memories/{id}` requests where the
respective victim sets (surviving units whose outgoing links pointed at
the patched units) intersect — `enqueue_relink_victims` then enqueues
each transaction's victim list in arbitrary `SELECT DISTINCT` order.

## Fix

Sort `unit_ids` inside both `PostgreSQLOps.enqueue_graph_maintenance` and
`OracleOps.enqueue_graph_maintenance` before issuing the INSERT. With a
total order over the lock set, deadlock is mathematically impossible —
both transactions queue cleanly on the first conflicting row, then
proceed in lockstep.

The only public caller (`enqueue_relink_victims` in
`hindsight_api/engine/graph_maintenance.py`) doesn't rely on insertion
order, and the abstract method docstring already documents "Order is
unspecified". This change picks a deterministic implementation-internal
order without strengthening the public contract.

## Test plan

`hindsight-api-slim/tests/test_enqueue_graph_maintenance_ordered.py`
(new file, 4 tests):

- [x] `test_pg_enqueue_graph_maintenance_inserts_in_sorted_order` —
  captures the array passed to `conn.execute` from a deliberately
  shuffled input and asserts it is sorted.
- [x] `test_oracle_enqueue_graph_maintenance_inserts_in_sorted_order` —
  same assertion against `conn.executemany`'s tuples.
- [x] `test_pg_empty_unit_ids_short_circuits` /
  `test_oracle_empty_unit_ids_short_circuits` — pin the existing
  early-return when `unit_ids == []`.

Regression sweep on the existing graph-maintenance suite —
**14/14 passing** in `tests/test_graph_maintenance.py` (relink semantics
unchanged).

`ruff check` / `ruff format` clean. Pre-commit hooks pass.

## Compatibility

- **Postgres**: behavioural identity except the (intentional) lock-order
  determinism. No schema changes. `ON CONFLICT DO NOTHING` semantics
  unchanged.
- **Oracle**: same. `IGNORE_ROW_ON_DUPKEY_INDEX` hint unchanged.
- **Public API**: no externally-visible change beyond the deadlock
  no longer firing under concurrent load.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
@@ -256,13 +256,21 @@ class OracleOps(DataAccessOps):
         # Oracle doesn't support ON CONFLICT; rely on the PK and the
         # IGNORE_ROW_ON_DUPKEY_INDEX hint to skip duplicates server-side.
         # The hint name must match the PK constraint exactly.
+        #
+        # Sort to enforce a global lock-acquisition order on the
+        # (bank_id, unit_id) PK. Without this, two concurrent
+        # transactions inserting overlapping unit_id sets in different
+        # orders can deadlock on the unique-check row locks. Sorting
+        # gives every concurrent caller the same lock order, so
+        # conflicting inserts queue cleanly instead of cycling.
+        sorted_unit_ids = sorted(unit_ids)
         await conn.executemany(
             f"""
             INSERT /*+ IGNORE_ROW_ON_DUPKEY_INDEX({table}, pk_graph_maintenance_queue) */
             INTO {table} (bank_id, unit_id)
             VALUES ($1, $2)
             """,
-            [(bank_id, uid) for uid in unit_ids],
+            [(bank_id, uid) for uid in sorted_unit_ids],
         )
 
     async def claim_graph_maintenance_batch(
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -348,6 +348,15 @@ class PostgreSQLOps(DataAccessOps):
     ) -> None:
         if not unit_ids:
             return
+        # Sort to enforce a global lock-acquisition order on the
+        # (bank_id, unit_id) unique-key. Without this, two concurrent
+        # transactions inserting overlapping unit_id sets in different
+        # orders can deadlock on the ON CONFLICT row locks — Postgres
+        # acquires a short-lived lock per row being checked, and cycle
+        # detection then aborts one transaction. Sorting gives every
+        # concurrent caller the same lock order, so conflicting inserts
+        # queue cleanly instead of cycling.
+        sorted_unit_ids = sorted(unit_ids)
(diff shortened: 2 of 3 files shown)


nicoloboschi (approved):
Approving. I reproduced the deadlock against real Postgres with two concurrent transactions inserting overlapping keys in opposite order (PG aborts one with DeadlockDetectedError after deadlock_timeout), and confirmed that a shared sort order eliminates the cycle — so the fix is correct for the queue-insert path.

One caveat for the record: the PR description's "deadlock is mathematically impossible" is overstated. The same PATCH transaction also issues an unsorted `entities ... ON CONFLICT ... DO UPDATE` per resolved entity (entity_resolver.py:936) before reaching the queue insert, which can still deadlock on overlapping entity sets. This PR correctly fixes the queue insert; the entity-upsert path is a separate follow-up (a deadlock-retry around the transaction would cover all sites at once, since acquire_with_retry deliberately does not retry in-transaction errors).

nicoloboschi:
Correction to my approval note above: I went and audited the full `ON CONFLICT` / multi-row-write surface of the engine, and my claim that the entity-upsert path "can still deadlock" was **wrong** — so don't go chasing it.

The `entities ... ON CONFLICT ... DO UPDATE` I pointed at (`entity_resolver.py:936`, via `_create_entity`/`resolve_entity`) is **dead code** — zero live callers. The *live* entity path is already deadlock-safe by the same ordering trick this PR applies to the queue:
- entity create (`bulk_insert_entities`) — caller sorts by `lower(name)` + `DO NOTHING`
- entity-stats UPDATE & co-occurrence `DO UPDATE` (`flush_pending_stats`) — both `sorted()` with explicit "prevents deadlocks" comments
- `bulk_insert_links`, `bulk_insert_unit_entities` — sorted at the caller

So `graph_maintenance_queue` was the *last* site that hadn't received the sort-for-lock-order treatment the rest of the write paths already had. This PR closes that gap — it's the complete fix for this deadlock class as it exists in the code, not a partial one. Apologies for the misdirection in the original review.

Follow-up: I'll open a small PR to delete the dead `resolve_entity`/`_create_entity`/`link_unit_to_entity` methods so that misleading `DO UPDATE` stops tripping up future readers (and reviewers).

PR #2355, opened 2026-06-22
fix(claude-code): segment retains from PreCompact

Sanderhoff-alt:
## Why

The Claude Code integration retains from the on-disk transcript JSONL file provided by the hook input as transcript_path. On current Claude Code builds we verified, that file is append-only across compaction: compact does not truncate the file or delete old transcript entries. Instead, Claude Code appends compact metadata, a compact summary, retained tail entries, and later conversation turns to the same JSONL transcript.

The previous full-session retain logic inferred compaction from a transcript size decrease. That does not match the observed Claude Code transcript behavior, so it is not a reliable way to segment retained documents. With an append-only transcript, shrink-based detection never represents the real compact boundary and can leave pre- and post-compact material mixed under the same document lifecycle.

This change wires the Claude Code PreCompact hook into the integration. Before Claude Code compacts, the hook forces one retain of the current pre-compact transcript. If that retain succeeds, it records the current transcript message count as a checkpoint. The next full-session retain reads the same append-only transcript and writes only the appended compact summary, retained tail, and new turns to a fresh session_id-cN document, preserving the pre-compact document instead of overwriting it.

The checkpoint is only marked after the retain API call succeeds. If the daemon, API URL resolution, or retain request fails, the hook does not advance the boundary. The next normal retain will continue using the original session document and full transcript, which may cost more extraction work but avoids dropping the pre-compact window.

Chunked retain mode continues to use its existing timestamped document behavior and does not create full-session compact checkpoints.

## What changed

- Added a PreCompact hook registration and script for the Claude Code plugin.
- Replaced shrink-based compaction inference with explicit PreCompact checkpoints.
- Made full-session retain segment post-compact content into session_id-cN documents after a checkpoint.
- Made run_retain return a success flag so PreCompact can avoid marking failed retains.
- Updated tests around append-only compact behavior, checkpoint failure, chunked mode, and state tracking.

### DIFF
--- a/hindsight-integrations/claude-code/hooks/hooks.json
+++ b/hindsight-integrations/claude-code/hooks/hooks.json
@@ -34,6 +34,18 @@
         ]
       }
     ],
+    "PreCompact": [
+      {
+        "matcher": "*",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"${CLAUDE_PLUGIN_ROOT}/scripts/pre_compact.py\" || python \"${CLAUDE_PLUGIN_ROOT}/scripts/pre_compact.py\"",
+            "timeout": 15
+          }
+        ]
+      }
+    ],
     "SessionEnd": [
       {
         "hooks": [
--- a/hindsight-integrations/claude-code/scripts/lib/state.py
+++ b/hindsight-integrations/claude-code/scripts/lib/state.py
@@ -157,32 +157,25 @@ def _locked_read_modify_write(state_name: str, lock_name: str, modify_fn):
     return result
 
 
-def track_retention(session_id: str, message_count: int) -> tuple:
-    """Track retention state and detect compaction.
+def mark_precompact(session_id: str, message_count: int) -> tuple:
+    """Record the transcript position before Claude Code compacts a session.
 
-    Compares the current message count against the last retained count for this
-    session.  When the transcript shrinks (compaction), increments a chunk counter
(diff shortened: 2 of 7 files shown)


nicoloboschi:
not sure, the compaction metadata are really just summaries of the existing lines of conversation. I don't see value in adding them

Sanderhoff-alt:
@nicoloboschi 
Thanks for taking a look. I want to clarify the intent here because the PR is not trying to retain compact metadata because those metadata have independent memory value.

The issue is boundary detection and document lifecycle. The current integration assumes compaction can be detected when the transcript message count shrinks (`track_retention`: `message_count < last_count`). On current Claude Code behavior, the `transcript_path` JSONL remains append-only across compaction: Claude Code appends the compact boundary/summary/retained tail/new turns rather than truncating the old lines. That means shrink-based detection never advances to `session_id-cN`.

That matters because Hindsight retain for a stable `document_id` uses replace/reprocess semantics by default. The Claude Code client does not send `update_mode=append`, so continuing to retain post-compact data under the original `session_id` keeps the post-compact segment in the same document lifecycle as the pre-compact document.

`PreCompact` gives us the reliable boundary signal: force-retain the full pre-compact transcript first, mark the message count only if that retain succeeds, then the next full-session retain writes only the appended compact summary/tail/new turns to `session_id-c1`. The summary is not being treated as novel source material; it is included because it is the actual context Claude carries forward after compaction, while the original pre-compact transcript remains preserved in the base document.

PR #2363, opened 2026-06-23
feat(extensions): add auth profile surfaces

### LINKED ISSUE #2343: Feature Request: Improve extension surfaces for authn/authz profiles and control-plane integrations
## Summary

Hindsight already has useful dataplane extension points, including `TenantExtension`, `OperationValidatorExtension`, `HttpExtension`, `MCPExtension`, and `MemoryDefenseExtension`.

However, when implementing a complete organization-based authn/authz profile, the current extension surfaces are not enough to keep the change isolated. The core authn/authz logic can mostly live in dataplane extensions, but several supporting concerns still require changes to dataplane glue code. The control plane currently has even fewer provider/extension abstractions, so adding a new auth mode requires direct changes across middleware, pages, API routes, navigation, feature gates, session handling, and dataplane client creation.

This issue proposes discussing extension/provider surfaces that would let advanced authn/authz integrations be implemented with minimal core changes and no impact on existing deployment modes when disabled.

## Motivation

The ideal target is:

- A deployment can add an authn/authz profile through explicit extension/provider configuration.
- The implementation can live mostly in external or isolated extension modules.
- Existing modes continue unchanged:
  - no auth / public schema
  - fixed API key
  - existing `SupabaseTenantExtension`
  - custom tenant extensions
- If the new profile is disabled, no extra UI, routes, headers, middleware behavior, or authorization logic should appear.
- Core changes should be limited to stable extension contracts, not provider-specific business logic.

Today, this is only partially possible.

## Compatibility Expectation

The `profile` concept here is meant as an opt-in deployment contract for advanced authn/authz integrations, not as a replacement for existing deployment modes.

Existing modes should remain compatible and unchanged unless an advanced profile is explicitly configured:

- no auth / public schema deployments
- fixed API key deployments
- the existing `SupabaseTenantExtension` user-to-schema mode
- custom `TenantExtension` implementations
- deployments that use only `HttpExtension`, `MCPExtension`, or `MemoryDefenseExtension`

For example, a `supabase_org` authn/authz deployment profile might require a coordinated set of pieces:

```text
Dataplane:
  SupabaseOrgTenantExtension
  SupabaseAuthorizationExtension

Control Plane:
  supabase_org auth provider
  request-scoped dataplane credentials
  organization/team/API key UI and APIs

Metadata:
  provider-specific organization/member/API key migrations
```

But when that profile is not configured, none of those requirements or UI behaviors should leak into other deployments. Traditional modes should not need provider-specific metadata tables, extra headers, new login flows, or new authorization behavior.

In other words, profiles should provide clear fail-fast validation when enabled, and zero behavioral impact when disabled.

## Dataplane Gaps

The dataplane extension model is close, but advanced authn/authz profiles expose several missing first-class concepts.

### 1. Resolved caller context

`TenantExtension` and `OperationValidatorExtension` often need the same resolved identity/policy:

- user id
- tenant id / organization id
- selected organization
- API key id
- role
- scopes
- allowed operations
- allowed banks
- provider-specific policy data

Today, implementations need an ad hoc way to pass this information through `RequestContext`.

A first-class resolved caller context would avoid repeated token/API key parsing and make authn/authz extensions compose more cleanly.

### 2. Selected tenant / organization context

Some deployments need a caller to select one tenant/organization from many memberships.

There should be a standard way to pass selected tenant context from HTTP/MCP requests into `RequestContext`, instead of each implementation inventing headers and plumbing.

### 3. Shared policy between tenant and validator extensions

A request commonly flows through:

```text
TenantExtension -> OperationValidatorExtension
```

The tenant extension may resolve the caller, tenant, schema, role, and scopes. The validator then needs the same data.

There should be an official way to cache/share resolved policy across extensions for one request.

### 4. Consistent lifecycle and context injection

Extensions that talk to external systems may need startup/shutdown hooks for HTTP clients, JWKS caches, database pools, or other resources.

Lifecycle and `ExtensionContext` injection should be consistent across extension types, especially tenant and operation validator extensions.

### 5. Grouped profile validation

Advanced authn/authz is often not one extension. It may require a coordinated profile:

```text
TenantExtension
OperationValidatorExtension
Control Plane auth provider
Control Plane route/session behavior
Metadata migrations
```

There should be a way to declare and validate grouped profiles so partial deployments fail fast.

For example, a profile requiring both tenant resolution and operation authorization should not silently run with only one side configured.

### 6. Diagnostics and feature discovery

The control plane needs to know whether the dataplane is compatible with the configured profile.

Useful diagnostics could include:

- active auth profile
- loaded tenant extension
- loaded validator extension
- profile readiness
- supported auth capabilities
- supported management capabilities

This could be exposed through `/version`, health, or a dedicated capabilities endpoint.

### 7. Tenant provisioning and migration lifecycle

Some tenant extensions need to provision schemas dynamically when a new tenant or organization appears.

There should be clear extension support for:

- listing known tenants
- running migrations for a tenant schema
- first-access schema initialization
- avoiding startup failures when external metadata exists but tenant schemas are not created yet

## Control Plane Gaps

The control plane currently appears much less extensible than the dataplane. Adding a new auth mode tends to require direct edits in many places.

Possible provider/extension surfaces to discuss:

### 1. Auth provider abstraction

The control plane could expose a stable auth provider interface for:

- disabled/public mode
- fixed access key mode
- Supabase org mode
- future OAuth / OIDC / enterprise SSO modes
- future user-level Supabase mode

The provider would own login/logout/session behavior instead of scattering auth branches across middleware and pages.

### 2. Request-scoped dataplane client factory

API routes should not need to know how auth headers are produced.

A provider should be able to supply dataplane request credentials:

```text
Authorization
selected tenant / organization header
fixed server API key
no auth
```

This would avoid manually modifying many route handlers whenever a new auth mode is added.

### 3. Route wrappers / proxy helpers

Control-plane API routes that proxy dataplane requests should use shared wrappers instead of manually constructing clients and forwarding headers.

This would reduce accidental bypasses and keep new auth modes isolated.

### 4. Navigation and settings registries

Auth/profile-specific UI should be injected through registries or capability-driven panels.

Examples:

- organization switcher
- team management
- API key management
- billing
- usage
- profile-specific settings

If a profile is disabled, these UI entries should not exist.

### 5. Organization / tenant context provider

Pages and components should be able to read a normalized current context:

```text
current organization / tenant
current role
available organizations
capabilities
selected bank scope
```

without knowing the underlying auth provider.

### 6. Management API abstraction

Organization, team, invite, API key, billing, and usage APIs are control-plane concerns in many deployments.

There should be a clean way for an auth provider or profile to contribute management APIs without deeply coupling them to unrelated pages/routes.

### 7. Feature gates from capabilities

The UI should derive feature availability from a normalized capability model, not from provider-specific checks scattered across components.

## Desired End State

If these extension/provider surfaces existed, an organization auth implementation should be able to work roughly like this:

```text
Dataplane:
  configure TenantExtension
  configure OperationValidatorExtension
  optionally configure profile/capabilities

Control Plane:
  configure AuthProvider
  configure management provider/routes
  register optional navigation/settings panels

Metadata:
  apply provider-specific migrations separately

Existing deployments:
  unchanged when the profile is not configured
```

The implementation should not require broad edits to:

- existing dataplane operation internals
- every control-plane API route
- unrelated pages
- unrelated navigation
- existing auth modes
- existing deployment behavior

## Non-goals

This is not asking for:

- a plugin marketplace
- hot-loading extensions
- one required auth provider
- a full enterprise IAM system
- memory-level ACL design in this issue
- billing implementation in this issue

The goal is a cleaner extension/provider surface so these things can be implemented externally or incrementally.

## Open Questions

- Should authn/authz be modeled as a grouped deployment profile?
- What should the standard resolved caller context contain?
- How should selected organization/tenant context be represented across HTTP and MCP?
- Should control-plane providers be formalized similarly to dataplane extensions?
- Should provider-specific management APIs live in control-plane route registries, `HttpExtension`, or both?
- How should provider-specific migrations be declared and checked?
- How can this avoid over-design while still supporting real multi-user and multi-tenant deployments?

## Why this matters

A cleaner extension surface would let the community experiment with organization auth, scoped API keys, team sharing, usage metering, quota enforcement, enterprise SSO, and other deployment profiles while keeping Hindsight core changes small, reviewable, and safe for existing users.

Sanderhoff-alt:
Closes #2343

## Summary

This PR adds reusable auth profile surfaces for Hindsight deployments that need coordinated authentication, tenant resolution, and authorization behavior without hard-coding a concrete provider into the core codebase.

The immediate motivation is to make cloud-like organization auth, scoped API keys, and provider-owned management UI possible as follow-up work while keeping the default open-source deployments unchanged. This PR intentionally stops at the extension surface: it does not add organization tables, team management, invitations, scoped API keys, billing, or any Supabase organization implementation.

## Problem

Hindsight already has extension points such as tenant extensions and operation validators, but a real auth deployment often needs multiple pieces to be configured together. For example, one component may resolve the caller and tenant, while another enforces operation-level policy. Without a profile-level contract, a partial configuration can start successfully and fail later, or only enforce part of the intended policy.

The control plane had a similar issue. Login, logout, dataplane headers, and settings UI were tied to the built-in disabled/access-key behavior. A future auth provider would have to patch shared pages and many dataplane proxy routes just to forward the correct caller credentials or expose its own settings surface.

## Dataplane changes

This PR introduces an auth profile abstraction on top of the existing extension system:

- `HINDSIGHT_API_AUTH_PROFILE` declares the expected auth profile, defaulting to `disabled`.
- Base extensions can declare `auth_profile`, `auth_profile_component`, and `required_auth_profile_components` metadata.
- Startup validates that a non-disabled profile has the required loaded components, so incomplete auth deployments fail early.
- `RequestContext` can carry resolved auth information such as selected tenant, subject, role, allowed operations, and auth policy data.
- `/version` reports auth profile diagnostics so the control plane can gate provider-specific UI safely.
- Tenant and operation validator startup hooks run before API startup migrations, so external tenant resolvers can be initialized before migration code calls `list_tenants()`.

These changes are provider-neutral. They do not add a concrete auth profile implementation.

## Control-plane changes

This PR adds a small auth provider registry and adapter contract for the control plane. Built-in disabled and access-key providers are registered through that contract, preserving existing behavior while giving future providers a stable place to plug in.

The adapter owns:

- config validation;
- request authentication checks;
- login field metadata;
- login/logout handling;
- request-scoped dataplane headers;
- optional settings path and label;
- expected dataplane auth profile.

The login page now asks `/api/auth/provider` for field metadata instead of hard-coding one login shape. Shared dataplane proxy routes now use request-scoped client/header helpers, so a future provider can forward caller-specific credentials without patching every route again. The generic `/settings` route redirects to the provider-owned settings path when one exists.

## Compatibility

The default deployment remains unchanged:

- If no control-plane auth provider is configured and no access key is set, the provider is `disabled` and public control-plane access continues to work.
- If `HINDSIGHT_CP_ACCESS_KEY` is set, the provider defaults to `access_key` and keeps the existing access-key session behavior.
- `HINDSIGHT_CP_DATAPLANE_API_KEY` is still forwarded for disabled/access-key deployments.
- `HINDSIGHT_API_AUTH_PROFILE` defaults to `disabled`, so existing tenant extensions are not required to declare profile metadata.

## Non-goals

This PR does not implement the cloud-like organization auth profile. In particular, it does not add:

- organization or team tables;
- invitation flows;
- scoped API key management;
- Supabase organization auth;
- billing or usage accounting;
- fine-grained memory-level authorization.

Those can be proposed separately on top of these surfaces.

## Validation

- `uv run pytest tests/test_auth_profile.py -q`
- `npm test --workspace=hindsight-control-plane -- --run tests/lib/auth/provider.test.ts tests/lib/hindsight-client-auth.test.ts tests/lib/dataplane-route-auth.test.ts`
- `npx tsc --noEmit -p tsconfig.json`


nicoloboschi (changes requested):
thanks for your work

although this adds complexity both on ui and api 

on the api we already have the tenant as "granularity" for extension and I think the changes you made are not really necessary and can be bound to the extension itself


on the ui, having extensions here is of course harder. the plan is to keep the ui handling simple and this feature would go in the other direction

my suggestion would be to fork the ui or create a brand new as you like

PR #2364, opened 2026-06-23
fix(recall): allow exact filtering of untagged/global observations (#2295)

### LINKED ISSUE #2295: Allow filtering to untagged observations

nicoloboschi:
## Summary

Fixes #2295. Supersedes #2322.

Allow recall to select only untagged/global observations by passing an empty tag
set (`[]` or omitted) together with `tags_match="exact"`. This lets users switch
between shared and tagged observation scopes without a synthetic "global" tag or
a hand-maintained negative filter.

## Root cause

The centralized tag filters treated `None`/`[]` as "no filter" in **every** mode,
returning every scope. There was no way to recall only the global/untagged scope —
the one that `observation_scopes="shared"` consolidation writes to.

## Fix

In `exact` mode, an empty/absent tag set now means the global/untagged scope and
emits an untagged-only clause (`tags IS NULL OR tags = '{}'`, matching both
historical NULLs and current empty-array storage). Applied consistently across:

- the flat SQL builders (`build_tags_where_clause`, `build_tags_where_clause_simple`)
- Python post-processing (`filter_results_by_tags`)
- compound tag-group leaves (`_build_group_clause`, `_match_group`)
- link-expansion post-filtering (run unconditionally so the empty exact scope applies)

No bind parameter is consumed for the empty-scope clause, so following parameter
indexes stay aligned. All other modes (`any`, `all`, `any_strict`, `all_strict`)
still treat empty/absent tags as "no filtering" — unchanged.

`http.py` field descriptions and the regenerated clients/OpenAPI/docs document the
new exact-empty scope.

## Relationship to #2322

This implements the same fix as #2322 but additionally updates the `RecallRequest`
API descriptions and regenerates the client SDKs (Go/Python/TypeScript), OpenAPI
spec, and docs — so the user-facing surface is complete. It also adds the
parameter-offset regression test from #2322. Opening this to supersede #2322.

## Tests

- SQL builders: flat + compound, `None` and `[]`, param-offset preserved
- Python post-filter: NULL / empty / tagged results under exact-empty scope
- Recall API: `tags=[]` + `tags_match="exact"` returns the untagged memory and
  excludes the tagged one

## Verification

- `tests/test_tags_visibility.py` tag-filtering tests pass (34 passed for the
  exact/untagged/empty/global subset incl. the recall API test)
- `ty check` passed; `ruff check` + `ruff format` clean; generated files in sync

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -286,12 +286,16 @@ class RecallRequest(BaseModel):
     )
     tags: list[str] | None = Field(
         default=None,
-        description="Filter memories by tags. If not specified, all memories are returned.",
+        description="Filter memories by tags. If not specified, all memories are returned. "
+        "Omitting tags (or passing []) together with tags_match='exact' filters to "
+        "untagged/global observations only (the scope written by observation_scopes='shared').",
     )
     tags_match: TagsMatch = Field(
         default="any",
         description="How to match tags: 'any' (OR, includes untagged), 'all' (AND, includes untagged), "
-        "'any_strict' (OR, excludes untagged), 'all_strict' (AND, excludes untagged).",
+        "'any_strict' (OR, excludes untagged), 'all_strict' (AND, excludes untagged), "
+        "'exact' (set-equality on the full scope, excludes untagged). With 'exact' and no tags "
+        "(or []), the empty global scope is selected and only untagged memories match.",
     )
     tag_groups: list[TagGroup] | None = Field(
         default=None,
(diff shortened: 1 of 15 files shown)


zupengwang:
Thanks for mentioning #2322 and keeping the regression test. If appropriate, would it be possible to add me as a co-author for the root-cause investigation and the parameter-offset regression test contribution?

Issue #2359, opened 2026-06-22
Consolidation Task Claiming Bug in v0.8.3 - Worker Never Claims Pending Tasks


entin-hun:
## Description
We're experiencing a critical issue where consolidation tasks are never claimed by workers, despite available capacity and valid pending tasks in the queue.

## Environment
- **Hindsight version**: v0.8.3 (ghcr.io/vectorize-io/hindsight:latest)
- **Deployment**: Docker + local embedded PostgreSQL
- **Configuration**:
  - HINDSIGHT_API_WORKER_CONSOLIDATION_MAX_SLOTS=3
  - HINDSIGHT_API_ENABLE_AUTO_CONSOLIDATION=true
  - consolidation_max_memories_per_round=200
  - consolidation_llm_parallelism=1

### Symptom
Consolidation progress stuck at 34/22,510 (0.15%) - not advancing for 24+ hours.

### Worker Behavior
Worker logs consistently show:
```
consolidation: total=1 claimable=1 payload_null=0 assigned=0
consolidation=0/3(avail=3)
claimed 9 tasks (0 consolidation)
```

**Pattern**: This repeats every 30 seconds across ALL worker instances. Consolidation slots are available but worker never claims consolidation tasks.

### Stuck Operation
- One consolidation operation stuck in `processing` state since 2026-06-22 18:22:31
- progress.processed=0/total=27181 (making zero progress)
- Cannot be cancelled via API (returns: "Operation cannot be cancelled: status is 'processing'")
- New consolidation POST /consolidate requests deduplicate to this existing pending op but are never claimed

## Root Cause Analysis
The worker polling loop has a bug in the task claiming logic for consolidation. Despite:
- ✓ 3 slots reserved for consolidation
- ✓ 1+ claimable consolidation tasks with valid payloads  
- ✓ No payload_null errors
- ✓ No error messages in logs

**Result**: Workers skip consolidation task claims entirely and only claim retain tasks.

## Impact
- Consolidation observations not being generated (only 10 created for 1719 documents)
- Retain tasks cannot effectively consolidate
- Workaround: Disabled consolidation to focus on ingest completion

## Questions
1. Is this a known issue in v0.8.3?
2. Is there a fix in a newer version?
3. What's the correct way to recover stuck 'processing' consolidation operations that cannot be cancelled?
4. Are there guard conditions in the consolidation claim logic that cause all tasks to be skipped?

## Reproduction
[...]

PR #2366, opened 2026-06-23
fix(worker): warn when worker_id unset inside a container (#2359)

nicoloboschi:
## Problem

Several users (e.g. #2359) report consolidation/async tasks getting permanently stuck in `processing`, never claimed despite available worker slots.

Root cause: `worker_id` defaults to `socket.gethostname()`, which inside Docker/Kubernetes is the **random container hostname** — it changes every time the container is recreated. `recover_own_tasks()` only reclaims tasks whose `worker_id` matches the *current* worker, so a task left in `processing` under a now-dead container's hostname is never recovered. New requests then dedupe onto the stuck op, and consolidation wedges indefinitely.

The real fix is to set a stable `HINDSIGHT_API_WORKER_ID`, but the failure mode is silent today.

## Change

- Add `detect_container_runtime()` (`docker` / `kubernetes` / `None`) via `KUBERNETES_SERVICE_HOST`, `/.dockerenv`, and `/proc/1/cgroup`.
- At worker start, when `HINDSIGHT_API_WORKER_ID` is unset **and** a container runtime is detected, log a prominent warning explaining the instability and how to fix it.

No behavior change to id resolution; non-containerized runs are unaffected.

## Notes

This is a guardrail, not a recovery mechanism — an already-stuck op (keyed to a dead container id) still won't self-heal. A follow-up orphan-recovery path can address that separately.

## Tests

`tests/test_container_detection.py` covers all three detection paths plus the negative case.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -3070,7 +3070,31 @@ def create_app(
         if config.worker_enabled and memory._backend.supports_worker_poller:
             from ..config import DEFAULT_DATABASE_SCHEMA
 
+            if not config.worker_id:
+                from ..utils import detect_container_runtime
+
+                runtime = detect_container_runtime()
+                if runtime:
+                    logging.warning(
+                        "\n"
+                        "============================================================\n"
+                        "  WARNING: HINDSIGHT_API_WORKER_ID is not set and Hindsight\n"
+                        f"  appears to be running inside {runtime}.\n"
+                        "\n"
+                        "  The worker id is defaulting to the container hostname,\n"
+                        "  which CHANGES every time the container is recreated.\n"
+                        "  When that happens, tasks left in 'processing' under the\n"
+                        "  old hostname are never recovered — consolidation and other\n"
+                        "  async operations can get stuck indefinitely.\n"
+                        "\n"
+                        "  Set HINDSIGHT_API_WORKER_ID to a STABLE value (e.g. the\n"
+                        "  compose service name or StatefulSet pod name) to avoid this.\n"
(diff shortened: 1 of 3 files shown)


koriyoshi2041:
I checked the changed surface locally:

```bash
cd hindsight-api-slim
uv run pytest tests/test_container_detection.py
uv run ruff check hindsight_api/utils.py hindsight_api/api/http.py tests/test_container_detection.py
uv run ruff format --check hindsight_api/utils.py hindsight_api/api/http.py tests/test_container_detection.py
git diff --check origin/main...review/pr-2366-worker-warning
```

Result: `4 passed`; Ruff/format/diff whitespace all clean.

The warning path looks right for the API-embedded worker. One non-blocking coverage note: the standalone `hindsight-worker` CLI still defaults `--worker-id` to `config.worker_id or socket.gethostname()` in `hindsight_api/worker/main.py`, but it does not call `detect_container_runtime()`. The docs recommend the production split of `HINDSIGHT_API_WORKER_ENABLED=false` plus separate `hindsight-worker` processes, so plain Docker / non-Helm deployments could still miss the new startup warning on the path where they are most likely to run dedicated workers. Helm is already fine because `worker-statefulset.yaml` sets `HINDSIGHT_API_WORKER_ID` from the pod name.

I would either mirror the warning in `worker/main.py` too, or track that as the next small follow-up. Not a reason to block the guardrail here.

The current CI failures I looked at appear unrelated to this Python change: Rust CLI build jobs are failing on generated client signatures expecting `Option<u64>` while `hindsight-cli/src/api.rs` passes `Option<i64>`, and shard 2 has a broad `RuntimeError: function '_has_torch_function' already has a docstring` import failure.

PR #2380, opened 2026-06-23
chore(deps): migrate pipecat integration to pipecat-ai 1.x (clears 4 high-severity alerts)

dcbouius:
## What

Migrates the **pipecat** integration from `pipecat-ai>=0.0.100,<1.0` to `>=1.4.0,<2.0`, clearing four **high-severity** Dependabot alerts that have no fix available within the 0.0.x line:

| Alert | Advisory |
|---|---|
| #1006, #1005 | Telephony WebSocket `/ws` unauthenticated call-control abuse |
| #560, #559 | Path traversal in Pipecat runner `/files` endpoint (arbitrary file read) |

The fix versions (`1.2.0` / `1.4.0`) are only on the 1.x line, which the integration previously capped out (`<1.0`, "1.0+ restructured modules; needs migration"). This PR does that migration.

## Why it's a small change

pipecat 1.x replaced the per-provider `OpenAILLMContext` with the universal `LLMContext` and removed the `pipecat.processors.aggregators.openai_llm_context` module. The processor already imported the modern `LLMContextFrame`, and `LLMContext.messages` is still a live list of OpenAI-format dicts — so the recall/retain/inject logic is unchanged. The migration is mostly swapping the removed legacy imports.

## Changes

- **`memory.py`** — drop the now-impossible legacy `OpenAILLMContextFrame` import branch; match on `LLMContextFrame` directly. Runtime injection logic untouched.
- **tests** — build frames from `LLMContextFrame`; add `TestRealLLMContext` that exercises a **real** pipecat `LLMContext` + `LLMContextFrame` (not mocks) to pin the live-list-mutation contract the integration relies on — the migration's highest-risk assumption.
- **examples** — migrate `basic_pipeline.py` to `LLMContext` + `LLMContextAggregatorPair` + `LLMRunFrame` kickoff (`create_context_aggregator` / `get_context_frame` were removed in 1.x); update `interactive_chat.py` mock frame.
- **pyproject** — pipecat 1.x requires Python `>=3.11`, so bump `requires-python` and drop the 3.10 classifier (CI's `.python-version` is already 3.11).
- **README / docstrings** — `OpenAILLMContextFrame` → `LLMContextFrame`.

## Testing

```
19 passed, 1 skipped (live)
```
`uv run ruff check` / `format` clean; `uv lock --check` consistent; wheel builds.

### DIFF
--- a/hindsight-integrations/pipecat/README.md
+++ b/hindsight-integrations/pipecat/README.md
@@ -49,14 +49,14 @@ See the [Hindsight installation guide](https://hindsight.vectorize.io/developer/
 
 ```
 New turn starts
-  └─ OpenAILLMContextFrame arrives
+  └─ LLMContextFrame arrives
        ├─ Retain previous complete turn (user+assistant) — fire-and-forget
        └─ Recall relevant memories for current user query
             └─ Inject as <hindsight_memories> system message
                  └─ Forward enriched context to LLM
 ```
 
-On each `OpenAILLMContextFrame`:
+On each `LLMContextFrame`:
 
 1. **Retain** — any new complete user+assistant turn pairs are sent to Hindsight asynchronously (non-blocking)
 2. **Recall** — the latest user message is used as the search query; results are injected as a system message before the LLM sees the context
--- a/hindsight-integrations/pipecat/examples/basic_pipeline.py
+++ b/hindsight-integrations/pipecat/examples/basic_pipeline.py
@@ -21,10 +21,12 @@ import asyncio
 import os
 
 from pipecat.audio.vad.silero import SileroVADAnalyzer
+from pipecat.frames.frames import LLMRunFrame
 from pipecat.pipeline.pipeline import Pipeline
 from pipecat.pipeline.runner import PipelineRunner
 from pipecat.pipeline.task import PipelineParams, PipelineTask
-from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
+from pipecat.processors.aggregators.llm_context import LLMContext
+from pipecat.processors.aggregators.llm_response_universal import LLMContextAggregatorPair
 from pipecat.services.cartesia.tts import CartesiaTTSService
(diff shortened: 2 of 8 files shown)


dcbouius:
Rebased onto current `main` and fixed the `verify-generated-files` failure that had this blocked.

**Root cause:** ruff `I001` (unsorted import block) in `examples/basic_pipeline.py` — `hindsight_pipecat` was grouped with the third-party imports instead of first-party. CI runs `ruff check --fix`, which applied the fix and left a diff, so "Verify no uncommitted changes" failed. That's exactly the `1 file changed, 2 insertions(+), 1 deletion(-)` in the old log. Now committed.

Worth flagging for anyone who hits this locally: `lint.sh` only lints integration packages that are **changed** locally, but lints **all** of them in CI (`LINT_ALL=1` when `$CI` is set), so a clean tree silently skips this path. Reproduce CI's behaviour with:

```
LINT_ALL_INTEGRATIONS=1 ./scripts/hooks/lint.sh
```

That also requires `ruff` on `PATH` — the integration lint shells out via `uv run --no-project ruff`, which does not provision ruff itself. Without it the task fails with `pyenv: ruff: command not found` rather than linting, which is easy to misread as passing. `uv tool install ruff` resolves it.

**Verification on the rebased branch:** `uv run pytest tests` → 19 passed, 1 skipped (the skip is `test_live_integration.py`, which needs `HINDSIGHT_LIVE_URL`). pipecat-ai 1.4.0 confirmed installed and the 1.x `LLMContextFrame` import path resolves. `LINT_ALL_INTEGRATIONS=1 ./scripts/hooks/lint.sh` exits 0 with no resulting diff.

Also confirmed the `requires-python` bump to `>=3.11` is forced rather than incidental: `pipecat-ai` 1.4.0 declares `requires_python >=3.11` (0.0.100 allowed `>=3.10`). CI resolves Python from `.python-version` (3.11), so it is unaffected — but this does drop 3.10 support for the published `hindsight-pipecat` package.

PR #2384, opened 2026-06-24
feat(config): let indexed multi-LLM members configure Vertex AI project/region

cdbartholomew:
## Summary

Extends the indexed multi-LLM layer (#2365) so a `vertexai` member can actually be used in a failover / round-robin chain.

Today an indexed member (`HINDSIGHT_API_LLM_<n>_*`) carries only `provider / api_key / model / base_url`. The `vertexai` provider additionally needs a **project id** (and a region; it defaults to `us-central1`) — `GeminiLLM._init_vertexai` raises `"VERTEXAI_PROJECT_ID is required"` without one. And `LLMProvider.__init__` previously sourced Vertex project/region **only** from global config via `get_config()`, with no way to set them per instance. Net effect: `vertexai` could not be a member of a chain — only the global/primary provider.

## Change

- `LLMMemberConfig`: add optional `vertexai_project_id` / `vertexai_region`.
- `_parse_llm_members`: read `{prefix}LLM_{n}_VERTEXAI_PROJECT_ID` / `_VERTEXAI_REGION` (global and the `RETAIN_`/`REFLECT_`/`CONSOLIDATION_` per-op prefixes), mirroring the existing per-member env keys.
- `_member_to_llm`: thread the two fields into the member's `LLMConfig`.
- `LLMProvider.__init__`: accept `vertexai_project_id` / `vertexai_region`; an explicit per-instance value wins, otherwise fall back to global config (`llm_vertexai_project_id` / `llm_vertexai_region` → `us-central1`). Single-LLM setups are unchanged.

## Why

Without this, a multi-LLM chain can only fail over between key-based providers; a `vertexai` member can't initialize. This lets operators run, e.g., a key-based primary with a Vertex AI failover member (or vice-versa) targeting a specific project/region.

## Tests

`tests/test_multi_llm_config.py` (+4):
- vertexai member parses project/region (global prefix)
- per-op prefix (`REFLECT_`) parses them
- a non-vertex member leaves the fields `None` (no regression)
- build path: the member's own project/region reach the Vertex client (SDK client patched; no live call), proving per-instance values are used rather than the global fallback

`uv run pytest tests/test_multi_llm_config.py tests/test_multi_llm_provider.py` → 38 passed. `ruff check` / `ruff format --check` / `ty check` clean.

## Notes / non-goals

- Single-LLM / global-config Vertex behavior is unchanged (global fallback preserved).
- Retry/fast-fail behavior is unchanged — that's existing config (`HINDSIGHT_API_LLM_MAX_RETRIES`, per-op overrides) layered on top of the chain.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -1320,6 +1320,8 @@ class LLMMemberConfig:
     default_headers: dict | None
     bedrock_service_tier: str | None
     gemini_service_tier: str | None
+    vertexai_project_id: str | None = None
+    vertexai_region: str | None = None
 
 
 # Valid multi-LLM strategy modes.
@@ -1413,6 +1415,8 @@ def _parse_llm_members(prefix: str) -> list[LLMMemberConfig]:
                 gemini_service_tier=(
                     parse_gemini_service_tier(gemini_service_tier) if provider.lower() == "gemini" else None
                 ),
+                vertexai_project_id=os.getenv(base + "VERTEXAI_PROJECT_ID") or None,
+                vertexai_region=os.getenv(base + "VERTEXAI_REGION") or None,
             )
         )
         index += 1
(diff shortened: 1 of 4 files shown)


nicoloboschi (approved):
LGTM — clean per-member Vertex project/region threading with explicit-wins-over-global fallback, mirrors the existing member env-key pattern, good parse + build-path test coverage.

PR #2390, opened 2026-06-25
fix(codex): use a stable prompt_cache_key instead of a per-request uuid4 to enable Codex prompt-cache reuse

sumleo:
## What

In the Codex provider, `prompt_cache_key` is set to a fresh `uuid.uuid4()` on every request:

```python
# hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
"prompt_cache_key": str(uuid.uuid4()),   # in both call() and call_with_tools()
```

`prompt_cache_key` is OpenAI/Codex's explicit hint for routing a request to a cached-prefix backend, so it needs to stay **constant** across calls that share the same `instructions` + `tools[]` prefix. A new random value per request means that prefix is never matched, giving a ~100% prompt-cache miss on the hottest LLM path (recall/reflect/retain all funnel through these two methods) with no functional upside.

## Fix

Derive the key once per provider instance from `(account_id, model)` so it is stable across requests but still distinct per account/model, and use it at both call sites. Tiny, self-contained change; the now-unused `uuid` import is dropped.

```python
@functools.cached_property
def _prompt_cache_key(self) -> str:
    seed = f"{self.account_id}:{self.model}"
    return hashlib.sha256(seed.encode("utf-8")).hexdigest()[:32]
```

If you'd prefer the key to be scoped more tightly (e.g. per bank/session/reflect-mission) rather than per provider instance, happy to adjust — I kept it to identifiers already in scope to stay minimal.

## How this was found

Spotted via static analysis of prompt-cache anti-patterns (a tool I'm experimenting with, CacheLint) and then confirmed by hand on `main` at `9dafadc` (lines ~391 and ~714). I have not been able to run the full integration suite against a live Codex account, so a maintainer sanity-check on the cache-key semantics would be appreciated.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
@@ -16,10 +16,11 @@ so that future server-side changes affect both clients identically.
 """
 
 import asyncio
+import functools
+import hashlib
 import json
 import logging
 import time
-import uuid
 from pathlib import Path
 from typing import Any
 
@@ -154,6 +155,20 @@ class CodexLLM(LLMInterface):
     @_auth_file.setter
     def _auth_file(self, v: Path) -> None:
         self._auth_manager._auth_file = v
+
+    @functools.cached_property
+    def _prompt_cache_key(self) -> str:
+        """Stable ``prompt_cache_key`` for OpenAI/Codex prompt-cache routing.
+
+        OpenAI uses ``prompt_cache_key`` as an explicit hint for routing
+        requests to a cached-prefix backend, so it must stay constant across
+        calls that share the same system instructions + tools prefix. A fresh
+        ``uuid4`` per request defeats that routing and forces a 100% cache miss.
+        Derive it once per provider instance from (account, model) so it is
+        stable across requests but still distinct per account/model.
+        """
+        seed = f"{self.account_id}:{self.model}"
+        return hashlib.sha256(seed.encode("utf-8")).hexdigest()[:32]
 
     # ------------------------------------------------------------------
     # Forwarding methods (keep surface area for tests / subclasses)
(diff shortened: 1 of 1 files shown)


sumleo:
Hi @nicoloboschi — gentle follow-up on this one (open since mid-June). It swaps the per-request `uuid4` prompt_cache_key for a stable `sha256(account:model)` key so Codex prompt-cache reuse actually kicks in. CI is green and it's a small, self-contained change. Would you have a moment to take a look, or let me know if you'd like anything adjusted? Happy to rebase. Thanks for your time and for the project!

handnewb (approved):
Nice optimization! Stable prompt_cache_key enables Codex prompt-cache reuse instead of busting it with a new uuid4 every request. Clean change, confined to one file.

nicoloboschi:
Thanks for this, and for the writeup — the diagnosis is right: a fresh `uuid4()` per request is a guaranteed cache miss, and it's still there on `main` at `codex_llm.py:460` and `:849`.

We've since landed #3271, which solved the same problem for the OpenAI-compatible provider family and set the convention we'd want Codex to follow. Rather than a per-instance key, it derives the affinity id in `hindsight-api-slim/hindsight_api/engine/cache_affinity.py`:

- primary source is the operation's `trace_id` (one per retain/reflect/consolidation run, bound around every provider call), so the key is constant across a run but distinct across runs;
- falls back to a fingerprint of the first message when no trace context is bound;
- always 32 lowercase hex chars, and fail-open — if no id can be derived the request goes out byte-identical to before;
- user-wins: a `prompt_cache_key` already in the operator's configured `extra_body` suppresses ours.

So the Codex fix we want is to call `cache_affinity_id(...)` at those two sites instead of minting a stable per-`(account_id, model)` key. That keeps one derivation path across providers and inherits the tests already merged with #3271.

Closing this one in favour of that approach — the Codex gap is real and still open, we just want it done through `cache_affinity.py`. If you'd like to take that follow-up, very happy to review it; otherwise we'll pick it up. Appreciate you flagging it.

PR #2392, opened 2026-06-25
fix(api): prevent reflect from creating banks

Sanderhoff-alt:
## Summary

- Treat reflect as a read-only operation when resolving bank identity.
- Use a non-creating bank profile lookup before starting reflect.
- Return `404` when the target bank does not exist instead of creating an empty bank.
- Update empty-bank reflect tests to create the bank explicitly and add a missing-bank regression test.

## Why

`reflect` is documented in the engine as read-only: it synthesizes an answer from stored bank data and does not persist new memories or profile changes. However, the reflect path fetched the bank profile with the default `get_bank_profile()` behavior, which auto-creates the bank if it is missing.

That means a caller could create an empty bank simply by reflecting against a stale or mistyped `bank_id`. This is surprising for a read-only operation and makes lifecycle and authorization semantics harder to reason about, because reflect ends up carrying an implicit create side effect.

This change keeps creation on explicit write/create paths and makes reflect return a clear `404` for missing banks.

## Test

- `uv run pytest tests/test_reflect_empty_based_on.py -q -n 1`
- `uv run ruff check hindsight_api/engine/memory_engine.py tests/test_reflect_empty_based_on.py`
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -8755,8 +8755,17 @@ class MemoryEngine(MemoryEngineInterface):
         tags_info = f", tags={tags} ({tags_match})" if tags else ""
         logger.info(f"[REFLECT {reflect_id}] Starting agentic reflect for query: {query[:50]}...{tags_info}")
 
-        # Get bank profile for agent identity
-        profile = await self.get_bank_profile(bank_id, request_context=request_context)
+        # Get bank profile for agent identity. Reflect is read-only and must
+        # not create a missing bank as a side effect.
+        profile = await self.get_bank_profile(
+            bank_id,
+            request_context=request_context,
+            create_if_missing=False,
+        )
+        if profile is None:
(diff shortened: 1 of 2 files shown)


nicoloboschi:
this is fine to keep this way

Sanderhoff-alt:
@nicoloboschi 
Thanks for taking a look. I understand this preserves the existing first-use behavior.

One thing I wanted to double-check is whether that behavior is still intentional for reflect specifically. For a missing bank, reflect first creates the bank profile, creates the per-bank vector indexes, may apply HINDSIGHT_API_DEFAULT_BANK_TEMPLATE, and then runs the reflect LLM loop against an otherwise empty bank.

Do we expect users to get meaningful value from reflecting on a bank that did not exist before the request? My concern is that a stale or mistyped bank_id turns into persistent bank creation plus LLM cost, while the result is usually just an empty-context answer.

If keeping the current behavior is mainly for compatibility, that makes sense. I just wanted to confirm the tradeoff explicitly, since reflect has a side effect/cost profile.

PR #2393, opened 2026-06-25
fix(mcp): prevent get_bank from creating banks

Sanderhoff-alt:
## Summary

- Treat the core MCP `get_bank` tool as read-only.
- Look up bank profiles with `create_if_missing=False` in both single-bank and multi-bank MCP modes.
- Return a not-found error when the bank is missing instead of creating it.
- Add MCP regression tests for missing banks and for the non-creating lookup parameter.

## Why

`get_bank` is a read tool: it returns bank metadata such as name, disposition, and mission. Before this change, it called `get_bank_profile()` with the default behavior, which can create a missing bank as a side effect.

That makes a read-only MCP call mutate persistent state when a client passes a stale or mistyped bank id. It also makes lifecycle and authorization semantics harder to reason about, because `get_bank` effectively becomes a hidden create path.

Creation should stay on explicit write/create surfaces such as `create_bank`. This change keeps `get_bank` read-only while preserving its existing response shape for successful reads.

## Test

- `uv run pytest tests/test_mcp_tools.py::TestTagsAndBankTools -q`
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/mcp_tools.py
+++ b/hindsight-api-slim/hindsight_api/mcp_tools.py
@@ -3129,7 +3129,10 @@ def _register_get_bank(mcp: FastMCP, memory: MemoryEngine, config: MCPToolsConfi
                 profile = await memory.get_bank_profile(
                     target_bank,
                     request_context=_get_request_context(config),
+                    create_if_missing=False,
                 )
+                if profile is None:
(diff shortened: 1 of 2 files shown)


Sanderhoff-alt:
@benfrank241 Could you share your view on making `get_bank` strictly read-only? In particular, do you see any compatibility concerns with returning not found instead of creating a missing bank?

nicoloboschi (approved):

PR #2395, opened 2026-06-25
feat(auth): add create bank validation hook

### LINKED ISSUE #2488: Add authorization hook for bank creation
## Problem

Bank creation currently needs a dedicated authorization point before a missing bank row is inserted. Existing bank-scoped write validation hooks operate on an existing target bank, so custom validators can reasonably assume the bank record and bank-scoped policy state already exist. Lazy creation paths make this distinction important because retain, imports, MCP `create_bank`, and profile auto-create flows can create banks implicitly.

## Proposed direction

Add a create-bank validation path that runs before inserting a new bank and route all bank creation entry points through it. Deployments without a custom validator should keep the current default-allow behavior. Existing-bank paths should not trigger the create-bank validation hook.

## Scope

- Authorize explicit MCP `create_bank` calls
- Authorize lazy bank creation from retain and async retain
- Authorize bank creation during template/document/bank imports
- Authorize default `get_bank_profile()` auto-create
- Keep existing-bank writes on the existing bank write validation path

Sanderhoff-alt:
## Summary

Add a dedicated `validate_create_bank()` operation-validator hook and `CreateBankContext` for authorizing bank creation before a missing bank row is inserted. The default implementation accepts everything, so deployments without a custom validator keep their existing behavior.

Route lazy bank creation through the hook from retain, async retain, template/document/bank imports, MCP `create_bank`, and the default `get_bank_profile()` auto-create path. Existing-bank paths skip the hook.

## Why

Bank creation is different from ordinary bank-scoped writes. Hooks such as `validate_bank_write()` operate on an existing target bank and many extension implementations can reasonably assume the bank row already exists when checking membership, metadata, or bank-scoped policy. Even `delete_bank` is still a write against an existing bank.

Create-bank authorization has to happen before the bank row exists. Placing this behind a precise `validate_create_bank()` hook avoids forcing `validate_bank_write()` implementations to special-case a missing target bank and keeps the create-bank capability separate from per-bank operation names.

This also closes the authorization gap for lazy creation paths: retain, imports, MCP `create_bank`, and direct `get_bank_profile()` auto-create now share one creation authorization point.

## Compatibility

The hook defaults to `ValidationResult.accept()`, matching the existing no-op behavior for deployments without a custom operation validator. Existing banks do not trigger the create-bank hook.

## Validation

- `uv run pytest hindsight-api-slim/tests/test_extensions.py::TestMemoryEngineValidation hindsight-api-slim/tests/test_mcp_tools.py::TestTagsAndBankTools hindsight-api-slim/tests/test_bank_templates.py::TestImportValidation`

Fixes #2488

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -5642,8 +5642,12 @@ def _register_routes(app: FastAPI):
     ):
         """Create or update an agent with disposition and mission."""
         try:
-            # Ensure bank exists by getting profile (auto-creates with defaults)
-            await app.state.memory.get_bank_profile(bank_id, request_context=request_context)
+            # Ensure bank exists, validating create_bank only when this call
+            # actually creates a missing bank.
+            await app.state.memory._ensure_bank_exists(
+                bank_id,
+                request_context,
+            )
 
             # Update name if provided (stored in DB for display only, deprecated)
             if request.name is not None:
@@ -5835,8 +5839,12 @@ def _register_routes(app: FastAPI):
                     dry_run=True,
                 )
 
-            # Ensure bank exists (auto-creates with defaults if needed)
-            await app.state.memory.get_bank_profile(bank_id, request_context=request_context)
+            # Ensure bank exists, validating create_bank only when this import
+            # actually creates a missing target bank.
+            await app.state.memory._ensure_bank_exists(
+                bank_id,
+                request_context,
+            )
 
             return await apply_bank_template_manifest(
                 memory=app.state.memory,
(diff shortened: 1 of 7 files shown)


Sanderhoff-alt:
One design note for discussion: this PR currently implements create-bank authorization as a new, dedicated operation-validator hook. Another possible direction would be to keep reusing the existing bank-management hook surface and extend the operation enum to represent bank creation explicitly.

I chose the new hook because create-bank is slightly different from ordinary bank-scoped writes: the target bank may not exist yet, so validators that assume an existing bank record or bank-scoped policy state can become awkward. That said, reusing the existing hook with a clearer operation value may be a better fit if we want to keep the extension API smaller.

I think the right abstraction is worth discussing before finalizing this API shape.

nicoloboschi (approved):

PR #2407, opened 2026-06-25
fix(control-plane): correct relative-time unit labels

benfrank241:
## Problem

A user reported that a memory bank's **documents list** showed `created 22 minutes ago` / `updated 13 minutes ago` for a document actually created **yesterday**, while the **document detail** panel showed yesterday's date correctly on both fields.

## Root cause

`formatRelativeTime` in `relative-time.ts` mislabeled the `units` table. Each `[divisor, label]` pair named the unit being *divided from* rather than the unit being *converted into*, so after dividing seconds by 60 the result (minutes) was still labeled `"second"`, and so on up the chain. Every duration rendered **one unit too small**:

| Actual age | Before | After |
|---|---|---|
| 22 minutes | `22 seconds ago` | `22 minutes ago` |
| 13 hours | `13 minutes ago` | `13 hours ago` |
| 22 hours | `22 minutes ago` | `22 hours ago` |
| 25 hours | `1 hour ago` | `yesterday` |
| 2 days | `2 hours ago` | `2 days ago` |

The reporter's document was ~22h old (created) / ~13h old (updated) — genuinely yesterday — so the list rendered them as "22 minutes / 13 minutes ago". The detail view uses `toLocaleString` (absolute date), which was correct, hence the apparent disagreement. The stored timestamps and the API are correct; this is purely a display bug.

It affects every relative timestamp this helper renders (documents list, mental-model and observation UI).

## Fix

Shift the `units` labels so each entry names the unit you convert **into**, and add a comment to prevent regression.

## Tests

Adds `tests/lib/relative-time.test.ts` covering the minute/hour/day/week/month/year boundaries (including the `22h → "22 hours ago"` and `25h → "yesterday"` cases) plus a future-timestamp case. Verified the test **fails** on the old labels and passes on the fix.

## Note

The same helper was ported into `hindsight-deployment`'s `cloud-ui` (the deployed dashboard) and has the identical bug; a sibling PR fixes that copy so the fix reaches production.

### DIFF
--- a/hindsight-control-plane/src/lib/relative-time.ts
+++ b/hindsight-control-plane/src/lib/relative-time.ts
@@ -3,13 +3,18 @@ export function formatRelativeTime(dateStr: string): string {
   const diffSec = Math.round((Date.now() - then) / 1000);
   const abs = Math.abs(diffSec);
   if (abs < 60) return diffSec >= 0 ? "just now" : "in a moment";
+  // Each entry is [divisor, unit-after-dividing]: when the running value
+  // (still in the previous unit) is at least `divisor`, divide by it and the
+  // result is expressed in `nextUnit`. Labels must name the unit you convert
+  // INTO, not the one you came from — otherwise every duration renders one
+  // unit too small (e.g. 22 hours shown as "22 minutes ago").
   const units: [number, Intl.RelativeTimeFormatUnit][] = [
-    [60, "second"],
     [60, "minute"],
-    [24, "hour"],
-    [7, "day"],
-    [4.345, "week"],
-    [12, "month"],
+    [60, "hour"],
+    [24, "day"],
+    [7, "week"],
+    [4.345, "month"],
+    [12, "year"],
     [Number.POSITIVE_INFINITY, "year"],
   ];
   let value = diffSec;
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Approved. Verified the fix by tracing the loop and running the new test both ways: it fails 4/5 on the old labels and passes 5/5 on the fix, so it genuinely guards the regression. Outputs after the fix are correct across the minute/hour/day/week/month/year boundaries plus the future-timestamp path.

Note for future readers: in this repo the documents list (`documents-view.tsx`) uses its own local `formatRelativeTime`, so it's unaffected here — the actual consumers of the fixed helper are the mental-model views, and the documents-list symptom lives in the deployed cloud-ui copy (sibling PR). Not blocking.

nicoloboschi:
Closing as superseded. The unit-label fix already landed on `main` in #1089 (commit 654e4c0c) — `relative-time.ts` there already has the corrected labels. Diffed against current main, this branch only adds a reworded comment plus a redundant `[Number.POSITIVE_INFINITY, "year"]` sentinel (never reached, since the loop breaks on `abs(value) < Infinity`), which is also what causes the merge conflict.

The genuinely useful piece was the regression test; if we want that guard, it's a one-liner to re-add on top of main separately. Thanks @benfrank241 for catching this — the fix itself is correct, just already in.

PR #2410, opened 2026-06-25
feat(devin-desktop): rename Windsurf→Devin Desktop + fix(continue) thread-safe adapter

DK09876:
This PR bundles two things (per request, kept in one PR):

## 1. Rename Windsurf → Devin Desktop
Cognition rebranded Windsurf to Devin Desktop (June 2026); Cascade is EOL July 1. Since `hindsight-windsurf` was merged but never published, renamed it to `devin-desktop` before first release.
- Package `hindsight-windsurf` → `hindsight-devin-desktop` (module, CLI, `DevinDesktopConfig`, bank `devin-desktop`, `HINDSIGHT_DEVIN_DESKTOP_BANK_ID`)
- Rule now writes to `.devin/rules/hindsight.md` (preferred path); MCP config stays `~/.codeium/windsurf/mcp_config.json` (Devin's data dir, unchanged by rebrand)
- Official Devin logo; docs + `integrations.json` + README refreshed; all registries updated
- 26 unit tests + gated E2E pass; real-app smoke verified

## 2. fix(continue): resolve a fresh Hindsight client per request
**Found via a real in-editor VS Code test.** The Continue adapter runs on a `ThreadingHTTPServer` (one thread per request) but shared a single Hindsight client. The client's aiohttp session is thread/loop-bound, so the **first** `@hindsight` recall worked and **every one after** threw `Timeout context manager should be used inside a task` — Continue then showed an error context item and the model answered with no memory.
- Resolve the client **per request** (test-injected clients still used as-is)
- Close per-request clients in a `finally` so the fresh aiohttp session doesn't leak a connector each call
- Bump `hindsight-continue` → **0.1.1** (the published 0.1.0 has this bug)
- Adds a regression test asserting per-request resolution across the threaded server

Verified live: 5 recalls in a row all succeed; 0 timeout errors; 0 unclosed-connector warnings. 17 continue tests pass; ruff clean.


benfrank241:
Reviewed — the code looks good, and **none of the red CI is from this diff** (it touches only integrations + docs + `test.yml`). Breaking the failures down:

- **`test-rust-cli`, `build-rust-cli-arm64`, `test-doc-examples (cli)`, `verify-generated-files`** — pre-existing breakage on `main`, not this PR: `mental_model.rs` is missing the new `refresh_cron` field on `MentalModelTriggerInput`, and two API test files have ruff-format drift. Fixed in #2427 — once that merges and CI re-runs, these clear.
- **Core LLM tests / LLM acceptance** — real-LLM judge flakiness (2 failed / 88 passed: a fact missing its absolute date, and a `reflect` that answered in English instead of Chinese). Not related to the rename.
- **`test-api`, `test-embed-windows`** — flaky/infra; this PR changes no API or embed code. They only ran here because editing `test.yml` triggers the full suite (they were skipped on integration-only PRs).

On the actual changes:
- **fix(continue)** is a genuine bug fix and well-handled — resolving a fresh client per request is the right call given `ThreadingHTTPServer` + the thread/loop-bound aiohttp session, the per-request client is closed in `finally` (injected clients left alone), and the regression test drives multiple requests through the threaded server. 👍
- The **Windsurf→Devin Desktop** rename is large but mechanical and the registries line up (`check-integrations` is green).

No code changes needed here. Recommend merging #2427 first, then a re-run. One sequencing note: #2417 defers a doc note until this lands, so **#2410 should merge before #2417**.

PR #2412, opened 2026-06-25
fix(retain): merge JSON arrays in append mode to preserve conversation-aware chunking

### LINKED ISSUE #2409: Bug: append mode corrupts `original_text`, breaking conversation-aware chunking on subsequent retains
### Bug Description

When `update_mode=append` is used with conversation-format content (flat JSON arrays of `{role, content, timestamp}` dicts), the `original_text` stored in the `documents` table becomes invalid JSON after the second retain cycle. This happens because `combined_content` is built by newline-joining multiple content items:

```python
combined_content = "\n".join([c.get("content", "") for c in contents_dicts])
```

When `contents_dicts` contains the previously stored text (a valid JSON array) and the new delta (another valid JSON array), the join produces two JSON arrays separated by `\n`, which is not valid JSON.

On subsequent append cycles, this corrupted `original_text` is fetched and passed to `chunk_text()`. The JSON parse fails, `_chunk_jsonl()` also rejects it (each line is an array, not a dict), and the text falls through to `RecursiveCharacterTextSplitter`. This splits the raw JSON string on sentence boundaries (`. `) inside message content, producing chunks that begin mid-sentence with no speaker attribution. The extraction LLM then has no basis for determining who said what.

Affected file: `hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py`

### Steps to Reproduce

1. Retain a conversation-format document with `update_mode=append`:
```json
{
  "document_id": "session-abc",
  "update_mode": "append",
  "items": [{"content": "[{\"role\":\"user\",\"content\":\"hello\"},{\"role\":\"assistant\",\"content\":\"hi\"}]"}]
}
```

2. Send a second retain with new turns to the same document_id:
```json
{
  "document_id": "session-abc",
  "update_mode": "append",
  "items": [{"content": "[{\"role\":\"user\",\"content\":\"how are you\"},{\"role\":\"assistant\",\"content\":\"fine\"}]"}]
}
```

3. Send a third retain with any new content to the same document_id.

4. Inspect the chunks produced for the document.

Minimal script demonstrating the storage corruption:

```python
import json

turn1 = json.dumps([
    {"role": "user", "content": "User: hello", "timestamp": "2026-06-23T22:00:00+00:00"},
    {"role": "assistant", "content": "Assistant: hi there", "timestamp": "2026-06-23T22:00:01+00:00"},
])

turn2 = json.dumps([
    {"role": "user", "content": "User: how are you", "timestamp": "2026-06-23T22:01:00+00:00"},
    {"role": "assistant", "content": "Assistant: doing well", "timestamp": "2026-06-23T22:01:01+00:00"},
])

# After 2nd retain, original_text in the DB becomes:
original_text = "\n".join([turn1, turn2])

# 3rd retain fetches this and tries to chunk it:
try:
    parsed = json.loads(original_text)
    print("JSON parse succeeded")
except json.JSONDecodeError as e:
    print(f"JSON parse failed: {e}")
    print("Falls through to RecursiveCharacterTextSplitter -> mid-sentence splits")
```

Output:
```
JSON parse failed: Extra data: line 2 column 1 (char 175)
Falls through to RecursiveCharacterTextSplitter -> mid-sentence splits
```

### Expected Behavior

All chunks produced from conversation-format append documents should go through `_chunk_conversation()`, splitting at turn boundaries. Every chunk should begin with a complete `{"role": ...}` object so the extraction LLM always has speaker attribution.

### Actual Behavior

After the second append cycle, `original_text` becomes newline-joined JSON arrays (not valid JSON). On subsequent retains, `chunk_text()` fails to parse it, `_chunk_jsonl()` rejects it (lines are arrays not dicts), and it falls through to sentence-boundary text splitting.

### Corruption mechanism

On the second retain, the server fetches `original_text` from step 1 and prepends it as a new content item:

```python
contents_dicts = [
    {"content": '[{"role":"user","content":"hello"},{"role":"assistant","content":"hi"}]'},
    {"content": '[{"role":"user","content":"how are you"},{"role":"assistant","content":"fine"}]'}
]
```

Then builds `combined_content`:

```python
combined_content = "\n".join([c.get("content", "") for c in contents_dicts])
```

Producing:

```
[{"role":"user","content":"hello"},{"role":"assistant","content":"hi"}]
[{"role":"user","content":"how are you"},{"role":"assistant","content":"fine"}]
```

This is stored as `original_text`. **Not valid JSON** - it's two JSON arrays separated by `\n`.

### Chunking failure

On the third retain onward, the server fetches the corrupted `original_text` and passes it to `chunk_text()`:

```python
parsed = json.loads(text)  # Fails - Extra data: line 2 column 1
```

Falls through to `RecursiveCharacterTextSplitter` with separators `["\n\n", "\n", ". ", "! ", "? ", "; ", ", ", " ", ""]`. The raw JSON string gets split on `. ` boundaries inside message content, producing chunks like:

```
. The configuration was working fine until...
. In that sense it has a complet...
. I've been trying to debug this...
```

No `role` prefix, no speaker attribution.

### Observable symptoms

In the Hindsight dashboard chunks view, a document retained via append mode shows an alternating pattern:

| Chunk | Starts with | Source |
|---|---|---|
| #0 | `[{"role": "user", "content":` | Recent delta (valid JSON, chunked correctly) |
| #1 | `. The configuration was working fine until...` | Corrupted `original_text` (text-split) |
| #2 | `[{"role": "user", "content":` | Recent delta |
| #3 | `. In that sense it has a complet...` | Corrupted `original_text` |

Chunks from the corrupted path lack speaker attribution entirely.

## Suggested Fix

When building `combined_content` for append-mode documents, the content items need to be merged in a format-aware way rather than blindly newline-joined. For conversation arrays specifically:

**Option A: Merge JSON arrays before storing.** Parse each content item as JSON; if both are arrays of dicts, concatenate the lists and re-serialize as a single flat array. `original_text` stays valid JSON and `_chunk_conversation()` fires on every subsequent fetch.

**Option B: Use JSONL.** Store each content item as one line (already the `\n` join behaviour). Teach `chunk_text()` to detect JSONL-formatted conversation arrays (multiple lines, each a JSON array of role/content dicts) and route them to `_chunk_conversation()` after flattening.

Option A is simpler and preserves the existing `_chunk_conversation` path without modification.

### Version

commit 6a10b62

### LLM Provider

OpenAI

qxxaa:
## Problem

When `update_mode=append` is used with conversation-format content (flat JSON arrays of `{role, content, timestamp}` dicts), `original_text` becomes invalid JSON after the second retain cycle.

In `_streaming_retain_batch`, `combined_content` is built with:

```python
combined_content = "\n".join([c.get("content", "") for c in contents_dicts])
```

For conversation content, this produces `[...]\n[...]` - two JSON arrays joined by a newline. This is neither valid JSON nor valid JSONL (lines are arrays, not dicts).

On subsequent append cycles, `chunk_text()` fails to parse the corrupted `original_text`. `_chunk_jsonl()` also rejects it. The text falls through to `RecursiveCharacterTextSplitter`, which splits on sentence boundaries (`. `) inside message content with no awareness of conversation turn structure. Chunks begin mid-sentence without speaker attribution, and the extraction LLM misattributes statements.

## Fix

In `orchestrator.py`, in `_streaming_retain_batch`, after the append-mode block assembles `contents_dicts = [existing_content, *contents_dicts]`: detect when all content items are JSON arrays of dicts, parse them, merge into a single flat array, and re-serialize.

Non-conversation content (plain text, JSONL, single JSON objects) is unaffected - the `isinstance` check fails and the block is skipped.

## Tests

- New chunking and integration tests in `test_chunking.py` and `test_retain_append_mode.py` verify both the corrupted path (regression guard) and the merged path (correct routing through `_chunk_conversation`).
- `test_newline_joined_json_arrays_bypass_conversation_chunking` is intentionally a regression document (proves the broken path exists), not a behavioural assertion. The second test (`test_merged_json_array_routes_to_conversation_chunking`) validates the correct routing.
- `max_chars=120` in the routing test is budget-sensitive but intentionally tight to force a multi-chunk split with short test messages.

Closes #2409

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -834,6 +834,28 @@ async def retain_batch(
             if first.get("tags"):
                 existing_content["tags"] = first["tags"]
             contents_dicts = [existing_content, *contents_dicts]
+            # Merge JSON arrays to keep original_text valid (#2409).
+            # Without this, combined_content joins items with "\n", producing
+            # "[...]\n[...]" which is not valid JSON. On the next append cycle
+            # chunk_text() fails to parse it and falls through to sentence-
+            # boundary text splitting, breaking speaker attribution.
+            try:
+                _merged = []
+                for _item in contents_dicts:
+                    _parsed = json.loads(_item.get("content", ""))
+                    if isinstance(_parsed, list) and all(isinstance(_e, dict) for _e in _parsed):
+                        _merged.extend(_parsed)
+                    else:
+                        _merged = None
+                        break
+                if _merged is not None:
+                    contents_dicts = [{"content": json.dumps(_merged, ensure_ascii=False)}]
+                    if first.get("context"):
+                        contents_dicts[0]["context"] = first["context"]
+                    if first.get("tags"):
+                        contents_dicts[0]["tags"] = first["tags"]
+            except (json.JSONDecodeError, ValueError, TypeError):
+                pass
             # Rebuild contents list to match
             contents = _build_contents(contents_dicts, document_tags)
             log_buffer.append(
--- a/hindsight-api-slim/tests/test_chunking.py
+++ b/hindsight-api-slim/tests/test_chunking.py
@@ -395,3 +395,64 @@ def test_rechunk_preserves_one_chunk_id_per_pre_chunk():
             chunk_ids.append(f"bank_doc_{global_idx}")
 
     assert len(chunk_ids) == len(set(chunk_ids)), f"duplicate chunk_ids in one batch: {chunk_ids}"
+
+
+# ---------------------------------------------------------------------------
+# Append-mode JSON array merge simulation (issue #2409)
+# ---------------------------------------------------------------------------
+
+
+def test_newline_joined_json_arrays_bypass_conversation_chunking():
+    """Newline-joined JSON arrays (the pre-fix append-mode storage format)
+    fail both the conversation and JSONL detection paths and fall through
+    to sentence-boundary text splitting.
+
+    This test documents the broken state that issue #2409 fixes at the
+    orchestrator level. chunk_text() itself is not changed; the fix
+    merges the arrays before they reach chunk_text().
+    """
+    turn1 = json.dumps([{"role": "user", "content": "Hello"}, {"role": "assistant", "content": "Hi there"}])
+    turn2 = json.dumps([{"role": "user", "content": "How are you"}, {"role": "assistant", "content": "Fine"}])
+    corrupted = turn1 + "\n" + turn2
+
+    chunks = chunk_text(corrupted, max_chars=80)
+
+    # The corrupted format does NOT route through _chunk_conversation.
+    # At least one chunk will not be a valid JSON array of dicts.
+    has_non_json_chunk = False
+    for chunk in chunks:
+        try:
+            parsed = json.loads(chunk)
+            if not (isinstance(parsed, list) and all(isinstance(e, dict) for e in parsed)):
+                has_non_json_chunk = True
+        except json.JSONDecodeError:
+            has_non_json_chunk = True
+    assert has_non_json_chunk, (
+        "Newline-joined JSON arrays should NOT produce valid conversation chunks. "
+        "If this fails, chunk_text() learned to handle the format and the "
+        "orchestrator-level merge in #2409 may be redundant."
+    )
+
+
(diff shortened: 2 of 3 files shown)


nicoloboschi (approved):
Confirmed the bug and traced it through the code. Forward fix is correct: merging the JSON arrays keeps original_text valid so subsequent appends route back through _chunk_conversation. Tests pass locally. Healing of pre-existing corrupted docs is intentionally out of scope.

PR #2417, opened 2026-06-26
fix(aider,openhands): close client on exit + OpenHands Docker MCP docs

DK09876:
Polish from the real-app integration test sweep (separate from the Devin Desktop rename + Continue fix in #2410).

## aider — close the client on exit
The wrapper opened a Hindsight client (aiohttp session) and never closed it, so `aider` printed `Unclosed connector` warnings on exit. Now `run()` closes the client **when it owns it** (a test/caller-injected client is left alone). Regression tests cover both paths. Bump `hindsight-aider` → 0.1.1.

## openhands — document the Docker-app MCP path
Real testing showed the **OpenHands Docker app ignores the project `config.toml`** our `init` writes — it only loads MCP from **UI settings** — and adding the server as **SSE** silently fails (Hindsight's endpoint is **Streamable HTTP**). Added:
- a README section telling Docker users to add the server in **Settings → MCP** as **Streamable HTTP** via `host.docker.internal`, and
- the same hint printed by `init`.

Bump `hindsight-openhands` → 0.1.1.

## Verification
- aider: 36 tests pass (2 new) · openhands: 27 pass · ruff check + format clean
- Both findings came from driving the real apps (aider CLI; OpenHands in OrbStack)

## Not included here (tracked separately)
- **MCP tool-selection dilution** — across Copilot/Zed/Devin/OpenHands the agent only reliably calls `recall` when explicitly nudged (29 tools + a soft rule). This is a design decision (stronger instruction vs. trimmed toolset) and deserves its own discussion/PR.
- **Devin Desktop "click Connect" doc note** — deferred until the rename PR #2410 lands (the `devin-desktop` dir doesn't exist on `main` yet).

### DIFF
--- a/hindsight-integrations/aider/hindsight_aider/runner.py
+++ b/hindsight-integrations/aider/hindsight_aider/runner.py
@@ -97,6 +97,20 @@ def do_retain(client: Any, config: AiderConfig, bank_id: str, transcript: str) -
         logger.warning("Hindsight retain failed: %s", e)
 
 
+def _close_client(client: Any) -> None:
+    """Best-effort close of the Hindsight client's HTTP session.
+
+    The client wraps an aiohttp session; leaving it open prints
+    "Unclosed connector" warnings after aider exits.
+    """
+    close = getattr(client, "close", None)
+    if callable(close):
+        try:
+            close()
+        except Exception:  # pragma: no cover - cleanup best-effort
+            pass
+
+
 def _default_run_aider(cmd: list[str]) -> int:
     """Run aider, inheriting stdio so the session is interactive."""
     try:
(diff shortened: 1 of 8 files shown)


benfrank241:
The two red checks here aren't from this PR — they're pre-existing breakage on `main`, and this branch is level with it. This diff only touches the aider/openhands packages, but both failures are in files it never modifies:

- **`verify-generated-files`** — `hindsight-api-slim/tests/test_retain_append_mode.py:45` has an unused local `v2_units` (F841). The job runs `lint.sh`, which auto-removes it, producing a diff. Fix: drop the assignment (or run `./scripts/hooks/lint.sh`).
- **`test-doc-examples (cli)`** — `hindsight-cli/src/commands/mental_model.rs` (lines 120 and 198) is missing the `refresh_cron` field that was added to the generated `MentalModelTriggerInput`, so the CLI no longer compiles. Fix: set `refresh_cron: None` in both initializers.

Both currently block every open PR. Worth a tiny standalone fix PR against `main`. The actual changes in this PR — closing the aider-owned client in a `finally` (leaving caller-injected clients alone, with tests for both paths) and the OpenHands Docker MCP docs — look good and are appropriately scoped.

PR #2419, opened 2026-06-26
refactor(extensions): centralize validator operation names

Sanderhoff-alt:
## Summary

Centralizes the operation names passed through `precheck`, `validate_bank_read`, and `validate_bank_write` by adding `PrecheckOperation`, `BankReadOperation`, and `BankWriteOperation` `StrEnum` types. The existing `PrecheckContext`, `BankReadContext`, and `BankWriteContext` call sites now use enum members instead of scattered string literals.

## Why this helps

`ctx.operation` is the main signal an `OperationValidatorExtension` can use for operation-specific authorization and quota checks. Before this change, the allowed values only existed as string literals spread across `api/http.py`, `memory_engine.py`, and HTTP route wiring, so extension authors had to search the codebase to discover the full set and could easily miss routes like bank config updates, async maintenance submissions, document/chunk access, file retain, or dry-run extraction.

Exported enums give extension authors a single source of truth, better IDE completion, and a clearer review surface when new operation-scoped hooks are added. They also make call sites self-documenting: `BankWriteOperation.UPDATE_BANK_CONFIG` and `PrecheckOperation.FILES_RETAIN` communicate the hook contract more clearly than raw string literals.

## Compatibility

The new enums inherit from `str`, so existing extensions that compare `ctx.operation` to string values such as `"retain"`, `"get_bank_stats"`, or `"update_bank_config"` continue to work. The compatibility test covers both enum identity checks and legacy string comparison behavior.

## Tests

- `uv run pytest tests/test_extensions.py -q`
- `./scripts/hooks/lint.sh`


nicoloboschi (approved):
Reviewed for backwards compatibility.

The new `PrecheckOperation`/`BankReadOperation`/`BankWriteOperation` `StrEnum`s are fully string-compatible — equality, `str()`, f-strings, `format()`, `json.dumps`, dict-key lookup, set membership, and str methods all behave as the bare string (verified empirically). The context dataclasses do no runtime type enforcement, so existing extensions comparing `ctx.operation == "retain"` and any not-yet-updated string call sites keep working. Exports are additive only; nothing removed.

Since this is a fork PR, `test-api`/Core LLM jobs are skipped in CI, so I ran `tests/test_extensions.py` locally on the PR branch: **42 passed**.

PR #2435, opened 2026-06-27
fix(cli): show filenames in operation list

r266-tech:
## Summary
- carry the optional operation `filename` field through the CLI's operation list model
- show `Filename:` in pretty `hindsight operations list` output when the API provides it
- set `refresh_cron: None` on existing mental-model trigger requests so the CLI continues compiling against the current generated Rust client schema

## Context
#2420 made file-upload operations expose the original filename on the API response so operators can identify pending or failed uploads. The CLI still converted the response into its local `Operation` type without that field, so pretty output only showed opaque operation/document IDs.

While verifying the CLI tests, the current generated Rust client also required the newer nullable `refresh_cron` field on `MentalModelTriggerInput`; adding `None` preserves the existing CLI behavior and keeps the focused tests compiling.

## Tests
- `cargo test --manifest-path hindsight-cli/Cargo.toml --locked operation::tests`
- `cargo test --manifest-path hindsight-cli/Cargo.toml --locked api::tests::test_operation_deserialize`
- `cargo test --manifest-path hindsight-cli/Cargo.toml --locked api::tests::test_operation_deserialize_with_error`
- `cargo test --manifest-path hindsight-cli/Cargo.toml --locked api::tests::test_operations_response_deserialize`

### DIFF
--- a/hindsight-cli/src/api.rs
+++ b/hindsight-cli/src/api.rs
@@ -62,6 +62,7 @@ pub struct Operation {
     pub id: String,
     pub task_type: String,
     pub items_count: i32,
+    pub filename: Option<String>,
     pub document_id: Option<String>,
     pub created_at: String,
     pub status: String,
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
LGTM — small, correct, well-tested. Verified refresh_cron:None compiles against main's generated client (nullable→Option) and missing filename deserializes to None.

PR #2438, opened 2026-06-28
fix(config): thread groq/openai service_tier into constructed LLM providers

r266-tech:
## Problem

`HINDSIGHT_API_LLM_GROQ_SERVICE_TIER` (default `"auto"`) and `HINDSIGHT_API_LLM_OPENAI_SERVICE_TIER` (OpenAI Flex, documented "50% cheaper") are parsed into `HindsightConfig` (`config.py`) and the providers already know how to apply them:

- `OpenAICompatibleLLM` sets `extra_body["service_tier"]` from `groq_service_tier` — gated on `if self.provider == "groq"`
- `fact_extraction.py` sets `request_body["service_tier"]` from `openai_service_tier` — gated on `if llm_config.provider == "openai"`

But the values are never threaded from config into any constructed provider. The four per-operation `LLMConfig` builds in `memory_engine.py` (default / retain / reflect / consolidation) and `LLMProvider.from_env` thread `bedrock_service_tier` and `gemini_service_tier`, but omit `groq`/`openai`. So setting either knob is a **silent no-op** — and since `groq` is the default provider, the cost-tier control is dead on the default deployment path.

## Fix

Thread `groq_service_tier`/`openai_service_tier` from config into the four per-operation builds and into `from_env`, exactly mirroring the existing `bedrock`/`gemini` lines. No new fields, no behavior change for unset knobs (`groq` default `"auto"` and `openai` default `None` match today's effective behavior), and the provider gating already prevents any cross-provider leak.

Scope notes:
- Config fields already exist (`config.py` `llm_groq_service_tier` / `llm_openai_service_tier`); this only adds the missing feed-in.
- The `_member_to_llm` path is intentionally untouched — `LLMMemberConfig` has no groq/openai tier fields (separate follow-up if ever wanted).
- `from_env` stays a lightweight env-only loader (no `HindsightConfig` build); `test_llm_provider_from_env_keeps_lightweight_loader` still passes.

## Tests

`uv run pytest tests/test_per_operation_llm_config.py` passes. Added `test_groq_openai_service_tier_threaded_into_per_operation_configs`, which sets both env knobs and asserts all four per-operation configs (`_llm_config`, `_retain_llm_config`, `_reflect_llm_config`, `_consolidation_llm_config`) receive both values. `ruff check`/`ruff format` clean on the changed source files.


nicoloboschi (approved):
Verified the wiring end-to-end: #2098 wired bedrock/gemini but left groq/openai parsed-but-dead, and since groq is the default provider the knob was a no-op on the default path. The four per-operation builds + from_env now mirror bedrock/gemini exactly; consumption stays gated (groq inside `provider == "groq"`, openai on `provider == "openai"`) so no cross-provider leak despite groq's truthy "auto" default. Test is a real regression guard. LGTM.

PR #2440, opened 2026-06-28
fix(retain): clarify fact type boundary for user rules

### LINKED ISSUE #2437: fact_type misclassified — user preferences/rules/corrections stored as "experience" instead of "world"
### Bug Description

 Title:

   

   
    Bug Description

    When users state preferences, rules, or corrections during assistant interactions (e.g., "优先用 browser_navigate 而非 web_search"), the retain pipeline classifies them as experience ("assistant") type facts instead of world type.

    Root Cause

    The LLM extraction prompt in fact_extraction.py uses a rule that is too broad:

    > fact_type: use "world" unless the content is clearly an interaction with the assistant.

    This causes any fact stated within an interaction to be classified as "assistant" (→ experience), regardless of its nature. User preferences and permanent rules are objective facts about the world — they would exist without this conversation — but are being stored as transient experiences.

    Location

    hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py:

    - Line 769: "fact_type: use \"world\" unless the content is clearly an interaction with the assistant."
    - Lines 1393-1402: raw "assistant" → "experience", "world" → "world" mapping
    - Lines 195-196, 298-299, 348-349: Field descriptions that don't distinguish between "doing something" and "saying something"

    See also retain/prompts.py line 870:
    > world: User's life, other people, events (would exist without this conversation)

    This definition is actually correct — the problem is the line 769 rule overrides it.

    Consequence

    - User preferences stored as experience degrade recall ranking (experience is for one-time events, not persistent rules)
    - Consolidation treats them as episodic rather than durable knowledge
    - A planned world-fact conflict detection feature would miss them entirely

    Proposed Fix

    Update the fact_type boundary guidance to distinguish "action" from "statement":

    diff
    - fact_type: use "world" unless the content is clearly an interaction with the assistant.
    + fact_type 判定：
    + - "world": user preferences, rules, corrections, constraints, and objective facts —
    +   even when stated during conversation.
    +   Examples: "user wants browser_navigate prioritized", "user forbids unapproved code changes"
    + - "assistant" (→ experience): actions the assistant/agent actually performed,
    +   not things discussed.
    +   Examples: "I deployed X", "I debugged Y"

    Also update the Field descriptions in the Pydantic models to emphasize that factual content remains "world" even when communicated through interaction.

    Environment

    - Hindsight v0.8.3 (ee81c65)
    - LLM: deepseek-v4-flash
    - Bank: hermes

### Steps to Reproduce

above

### Expected Behavior

above

### Actual Behavior

above

### Version

v0.8.3

### LLM Provider

OpenAI

koriyoshi2041:
## Problem

User preferences, rules, corrections, and constraints can be stated during assistant interactions, but they are durable facts about the user rather than assistant experiences. The current fact extraction guidance made that boundary easy to blur, which can send durable user facts through the experience path.

Fixes #2437.

## Fix

- Clarify the retain fact_type prompt so user preferences/rules/corrections/constraints remain `world` even when mentioned in an assistant interaction.
- Narrow `assistant` to actions or experiences the assistant/agent actually performed.
- Align the Pydantic schema field descriptions and verbatim/verbose prompt guidance with the same boundary.
- Add a prompt/schema regression test that locks this wording down without calling a live LLM.

## Test

- `uv run pytest tests/test_fact_extraction_fact_type_prompt.py -q`
- `uv run ruff check hindsight_api/engine/retain/fact_extraction.py tests/test_fact_extraction_fact_type_prompt.py`
- `uv run ruff format --check hindsight_api/engine/retain/fact_extraction.py tests/test_fact_extraction_fact_type_prompt.py`
- `git diff --check`

## Risk

Low-to-medium. This is prompt/schema guidance only, not persistence or mapping logic. It should reduce misclassification of user facts, but prompt wording can affect extraction behavior, so the change is intentionally narrow.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -193,7 +193,7 @@ class ExtractedFact(BaseModel):
     occurred_start: str | None = Field(default=None, description="ISO timestamp for events")
     occurred_end: str | None = Field(default=None, description="ISO timestamp for event end")
     fact_type: Literal["world", "assistant"] = Field(
-        description="'world' = objective/external facts. 'assistant' = first-person actions, experiences, or observations by the speaker."
+        description="'world' = objective/external facts, including user preferences, rules, corrections, and constraints even when stated during a conversation. 'assistant' = actions, experiences, or observations the assistant/agent actually performed."
     )
     entities: list[Entity] | None = Field(default=None, description="People, places, concepts")
     causal_relations: list[FactCausalRelation] | None = Field(
@@ -296,7 +296,7 @@ class ExtractedFactVerbose(BaseModel):
     )
 
     fact_type: Literal["world", "assistant"] = Field(
-        description="'world' = objective/external facts about other people, events, general knowledge. 'assistant' = first-person actions, experiences, or observations by the speaker (e.g., 'I changed X', 'I discovered Y')."
+        description="'world' = objective/external facts about the user, other people, events, general knowledge, preferences, rules, corrections, or constraints. 'assistant' = actions, experiences, or observations the assistant/agent actually performed (e.g., 'I changed X', 'I discovered Y')."
     )
 
     entities: list[Entity] | None = Field(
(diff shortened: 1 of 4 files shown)


nicoloboschi (approved):
LGTM

PR #2441, opened 2026-06-28
fix(openclaw): apply configured defaults to dynamic banks

devinkuhn:
## Summary
- Add optional OpenClaw plugin config for full Hindsight bank defaults on dynamic banks.
- Apply configured defaults before first retain/recall for derived dynamic banks.
- Preserve existing behavior when defaults are unset; existing mission-only config keeps working.
- Document per-user dynamic bank defaults in the OpenClaw README and docs site.
- Remove the unsupported `max_results` parameter from the `agent_knowledge_recall` tool.

## New optional config
- retainExtractionMode
- enableObservations
- enableAutoConsolidation
- dispositionSkepticism / dispositionLiteralism / dispositionEmpathy
- entityLabels

## Why
Dynamic per-user OpenClaw banks currently get created with server defaults unless manually patched after creation. That means teams using per-user isolation can accidentally retain first-turn memories under concise extraction/no entity labels even when their intended base/shared bank uses verbose extraction and labels. This patch lets the plugin stamp the desired defaults on each derived bank before retain/recall.

## `max_results` removal (agent-sdk)
The `agent_knowledge_recall` tool previously accepted a `max_results` parameter that was silently aliased to the recall **token budget** (`max_tokens`) — it never actually capped the number of results. Rather than turn it into a real result-count cap, it has been removed entirely: the tool now accepts only `max_tokens`. For a count cap on automatically injected memories, use `recallTopK` on auto-recall. The docs-site guidance that referenced `max_results` has been corrected accordingly.

## Test plan
- cd hindsight-integrations/openclaw && npm run build && npm test -- --run
- cd hindsight-tools/hindsight-agent-sdk && npm run build && npm test -- --run


koriyoshi2041 (commented):
Did a local pass since this touches dynamic bank setup plus the shared knowledge-tool factory.

Verified locally:
- `cd hindsight-integrations/openclaw && npm ci && npm test -- --run` -> 273 passed
- `cd hindsight-integrations/openclaw && npm run build` -> passed
- `cd hindsight-tools/hindsight-agent-sdk && npm ci && npm test -- --run` -> 18 passed
- `cd hindsight-tools/hindsight-agent-sdk && npm run build` -> passed

I also traced the first-use paths: `getClientForContext()` applies configured defaults before auto-recall/retain, and the knowledge-tool execute path applies defaults before dispatching tool calls. I did not find a blocking issue in that pass.

PR #2445, opened 2026-06-28
fix(search): guard CrossEncoderReranker.ensure_initialized with asyncio.Lock

Srujan-rai:
## Summary

- Two concurrent requests that both observed `_initialized=False` would both
  call `initialize()`, double-loading the local model and causing a memory spike
- Added a lazily-created `asyncio.Lock` with double-checked locking pattern so
  only the first caller runs initialization; subsequent callers wait and return
  immediately once `_initialized=True`

## Tests

Added `tests/test_reranker_init_lock.py`:
- `test_concurrent_ensure_initialized_calls_initialize_once` — fires 3 concurrent
  `ensure_initialized()` calls via `asyncio.gather`, asserts `initialize` called once
- `test_second_call_after_init_skips_initialize`  sequential second call is a no-op

## Test plan
- [ ] Concurrent recall requests no longer double-initialize the reranker model
- [ ] Memory usage stable under concurrent load with local cross-encoder

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/search/reranking.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/reranking.py
@@ -214,6 +214,7 @@ class CrossEncoderReranker:
             cross_encoder = create_cross_encoder_from_env()
         self.cross_encoder = cross_encoder
         self._initialized = False
+        self._init_lock = None  # asyncio.Lock — created lazily inside async context
 
     async def ensure_initialized(self):
         """Ensure the cross-encoder model is initialized (for lazy initialization)."""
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
there's no chance we initialize 2 rerankers, can you show me where? thanks

nicoloboschi:
there's no chance we initialize 2 rerankers, can you show me where? thanks

Srujan-rai:
> there's no chance we initialize 2 rerankers, can you show me where? thanks

The race is between reranking.py:220 (the check) and reranking.py:248 (the flag set). The gap contains await asyncio.wait_for(init) at line 240  asyncio yields there, letting a second coroutine enter ensure_initialized, pass the same _initialized=False check, and call run_in_executor again. Both then load the model concurrently in the thread pool. The window is the model load time (~2–10s for the default local model), so any two cold-start recall requests arriving simultaneously trigger it.

nicoloboschi:
You're right about the mechanism — there's a genuine check-then-act race across the `await` between the `_initialized` check and the flag set, and a second coroutine can slip through and load the model twice. Good catch.

One qualifier worth recording: this can only be hit when **lazy reranker init is enabled** (`HINDSIGHT_API_LAZY_RERANKER=true`, which is off by default). In the default path `MemoryEngine.initialize()` runs `init_cross_encoder()` eagerly at startup — single-threaded, before any request is served — which sets `_initialized = True` before the first `recall`. So `ensure_initialized()` short-circuits and the window never opens. The race only exists when eager init is skipped and the first concurrent cold-start recalls race into `ensure_initialized()`.

Given that, rather than guard the lazy path with a lock, we're going to **remove `HINDSIGHT_API_LAZY_RERANKER` entirely**. Eager startup init eliminates the race by construction and removes the first-recall latency cliff; the only thing the flag bought was skipping an ~80MB model load for retain-only deployments, which isn't worth the extra config surface and concurrency footgun. Closing this in favor of that change — thanks for the careful analysis, it's what surfaced the cleaner fix.

Srujan-rai:
> You're right about the mechanism — there's a genuine check-then-act race across the `await` between the `_initialized` check and the flag set, and a second coroutine can slip through and load the model twice. Good catch.
> 
> One qualifier worth recording: this can only be hit when **lazy reranker init is enabled** (`HINDSIGHT_API_LAZY_RERANKER=true`, which is off by default). In the default path `MemoryEngine.initialize()` runs `init_cross_encoder()` eagerly at startup — single-threaded, before any request is served — which sets `_initialized = True` before the first `recall`. So `ensure_initialized()` short-circuits and the window never opens. The race only exists when eager init is skipped and the first concurrent cold-start recalls race into `ensure_initialized()`.
> 
> Given that, rather than guard the lazy path with a lock, we're going to **remove `HINDSIGHT_API_LAZY_RERANKER` entirely**. Eager startup init eliminates the race by construction and removes the first-recall latency cliff; the only thing the flag bought was skipping an ~80MB model load for retain-only deployments, which isn't worth the extra config surface and concurrency footgun. Closing this in favor of that change — thanks for the careful analysis, it's what surfaced the cleaner fix.

That makes perfect sense. Removing the lazy initialization flag entirely is a much cleaner approach it eliminates the config surface and ensures predictable startup behavior without adding unnecessary lock overhead.

Glad the analysis was helpful in simplifying the codebase. Thanks for the quick review and the detailed explanation!

PR #2475, opened 2026-06-30
feat(gemini): opt-in 4xx request-dump for diagnosing rejected calls

cdbartholomew:
## What

Adds an **opt-in** diagnostic to the Gemini/Vertex provider that logs the exact request behind any Gemini `4xx`, gated by the env var `HINDSIGHT_API_LLM_DEBUG_DUMP_4XX` (off by default).

When enabled, on a `4xx` the provider logs `[LLM_4XX_DUMP]` with:
- the serialized `GenerateContentConfig` — the response schema as the SDK renders it on the wire, plus generation params, safety settings, and the cached-content reference;
- per-message `role` + character count and a length-capped text preview.

It fires before the cache-drop retry rebuilds the config, so both the cached and uncached request forms are captured.

## Why

Some `400 INVALID_ARGUMENT` rejections from structured-output calls are not reproducible by reconstructing the request after the fact — the failing factor is in the request as actually assembled at runtime. Reconstructed replays of the same inputs return `200`, so the only reliable way to see what the model rejected is to capture the real request at the moment it fails. This gives an operator a way to do that without standing up bespoke instrumentation.

## Safety / scope

- **Off by default** — the env var is unset in normal operation.
- The serialized config carries no message content; message previews are length-capped, so an enabled dump can't flood logs or spill large bodies.
- The helper **never raises** — diagnostics must not break the request path (falls back to `repr()` if serialization fails).
- Scoped to the Gemini provider's `APIError` handler; no behavior change when the flag is off.

## Tests

`tests/test_gemini_4xx_dump.py` covers:
- disabled → no-op (no log emitted)
- enabled → emits serialized config + capped preview + error code
- preview truncation (long bodies capped)
- unserializable config → falls back, never raises

All pass; ruff clean.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/gemini_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/gemini_llm.py
@@ -11,6 +11,7 @@ import base64
 import io
 import json
 import logging
+import os
 import time
 from contextvars import ContextVar
 from typing import Any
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Superseded by #2865, which generalizes this diagnostic from Gemini-only into a single provider-agnostic helper wired into every remote LLM provider (Gemini/Vertex, OpenAI-compatible + Fireworks/Nous, Anthropic, LiteLLM + router, Codex), same opt-in flag (`HINDSIGHT_API_LLM_DEBUG_DUMP_4XX`) and same behaviour. Suggest closing this in favor of #2865.

Issue #2459, opened 2026-06-29
Fix LiteLLM provider temperature handling for GPT-5.5


aronnoordhoek:
### Bug Description

When using Hindsight with the LiteLLM provider and Azure OpenAI GPT-5.5, internal LLM calls fail because Hindsight forwards explicit `temperature` values.

GPT-5.5 only supports the default temperature, so calls with values such as `0.0` or `0.1` are rejected.

### Steps to Reproduce

1. Configure Hindsight with:

   HINDSIGHT_API_LLM_PROVIDER=litellm
   HINDSIGHT_API_LLM_MODEL=azure/gpt-5.5

2. Run an operation that uses the LiteLLM provider, for example retain/fact extraction.

3. Observe the LiteLLM/Azure OpenAI error about unsupported `temperature`.

### Expected Behavior

Hindsight should work with Azure OpenAI GPT-5.5 through the LiteLLM provider.

For models that do not support configurable temperature, the LiteLLM provider should omit `temperature` rather than forwarding explicit internal values.

### Actual Behavior

The LiteLLM provider forwards explicit temperature values from internal calls.

Examples include:

- connection verification uses `temperature=0.0`
- retain/fact extraction uses `temperature=0.1`
- think uses `temperature=0.9`
- mental model delta operations use `temperature=0.0`

Azure OpenAI GPT-5.5 rejects these calls with an error like:

Unsupported value: 'temperature' does not support 0.1 with this model. Only the default (1) value is supported.

### Version

0.8.3

### LLM Provider

Other

PR #2477, opened 2026-06-30
fix(missions): honor configured LLM temperature for bank mission merge (#2469 follow-up)

r266-tech:
### Problem

#2469 made the internal LLM temperature configurable and its docs state the global `HINDSIGHT_API_LLM_TEMPERATURE` "applies to every operation" and `none` "drops temperature from every operation" (the documented remedy for models that reject explicit temperatures, e.g. Azure `gpt-5.5`, issue #2459).

But bank/folder **mission merging** is a 5th internal LLM call that #2469 didn't reach. `_llm_merge_mission` (`engine/retain/bank_utils.py`, `scope="bank_mission"`, used by `merge_bank_mission`) still hardcodes `temperature=0.3` and never reads `get_config()`:

```python
content = await llm_config.call(
    messages=messages, scope="bank_mission", temperature=0.3, max_completion_tokens=8192
)
```

So setting `HINDSIGHT_API_LLM_TEMPERATURE=none` (your documented fix for Azure gpt-5.5) does **not** cover mission merging — the `0.3` is still sent, the call raises, the `try/except` in `_llm_merge_mission` swallows it, and the merge silently degrades to naive `"{current} {new_info}"` concatenation. With folder missions / mission-sandbox (#2455, #2254) actively shipping, this path is gaining users following the docs and still hitting a silent failure.

### Fix

Add `bank_mission` as a 5th per-operation temperature knob, mirroring the four introduced in #2469 (verification / retain / reflect / consolidation):

- `ENV_LLM_TEMPERATURE_BANK_MISSION` + `DEFAULT_LLM_TEMPERATURE_BANK_MISSION = 0.3` (preserves the historical value)
- `HindsightConfig.llm_temperature_bank_mission`, resolved via the existing `_resolve_operation_temperature` (per-op env → global `HINDSIGHT_API_LLM_TEMPERATURE` → default `0.3`)
- `bank_utils.py` now passes `temperature=get_config().llm_temperature_bank_mission`

The documented global override now genuinely covers mission merging, and `none` correctly omits the parameter there too. Default behavior is unchanged (still `0.3`).

### Docs / tests

- Documented the new knob in `configuration.md` and regenerated the CI-enforced skills mirror via `scripts/generate-docs-skill.sh` (diff is the single new row).
- `test_llm_temperature_env.py`: `bank_mission` added to `_OP_FIELDS`, so the default / global-override / global-`none` resolution cases now assert it too.
- `test_llm_temperature_pipeline.py`: two new tests drive `_llm_merge_mission` through the mock LLM and assert the `bank_mission` call receives `0.3` by default and `None` (omitted) under `HINDSIGHT_API_LLM_TEMPERATURE=none`.

`test_llm_temperature_env.py` passes locally (13 passed); the pipeline tests mirror the existing reflect/retain pipeline tests (they need the Postgres-backed `memory` fixture, which runs in CI).

Follow-up to merged #2469.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -160,6 +160,7 @@ ENV_LLM_TEMPERATURE_VERIFICATION = "HINDSIGHT_API_LLM_TEMPERATURE_VERIFICATION"
 ENV_LLM_TEMPERATURE_RETAIN = "HINDSIGHT_API_LLM_TEMPERATURE_RETAIN"
 ENV_LLM_TEMPERATURE_REFLECT = "HINDSIGHT_API_LLM_TEMPERATURE_REFLECT"
 ENV_LLM_TEMPERATURE_CONSOLIDATION = "HINDSIGHT_API_LLM_TEMPERATURE_CONSOLIDATION"
+ENV_LLM_TEMPERATURE_BANK_MISSION = "HINDSIGHT_API_LLM_TEMPERATURE_BANK_MISSION"
 
 # Multi-LLM strategy. Extra LLMs are configured by index alongside the unindexed
 # primary (e.g. HINDSIGHT_API_LLM_1_PROVIDER, HINDSIGHT_API_LLM_2_PROVIDER, ...),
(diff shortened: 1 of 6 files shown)


nicoloboschi:
this was on purpose because that endpoint is deprecated

PR #2479, opened 2026-06-30
fix(search): include proof_count in temporal spreading SQL SELECT

Srujan-rai:
## Summary

- The LATERAL join query for temporal graph spreading omitted `mu.proof_count` from the SELECT list
- `RetrievalResult.from_db_row()` calls `row.get("proof_count")`, which always returned `None` for spread neighbors
- Spread neighbors always got a neutral 0.5 proof-count boost regardless of actual observation evidence strength

## Tests

Existing `test_reranking_proof_count.py` covers proof_count boost scoring.The SQL omission meant spread neighbors silently bypassed this logic this fix ensures the column is present for all retrieval paths.

## Test plan
- [ ] Temporal graph spread neighbors now carry correct proof_count boost
- [ ] Observations with high proof_count rank higher than unverified ones in temporal queries


nicoloboschi (approved):
nice catch!

PR #2484, opened 2026-06-30
feat(extensions): OIDC + GitHub tenant extensions; fix(worker): drain scan-only schemas (follow-up to #1555)

skymoore:
## Summary

Two related changes that enable multi-user, self-hosted Hindsight deployments with per-user tenant isolation behind a generic identity layer:

1. **New built-in tenant extensions** (`feat`) — OIDC and GitHub, plus a GitHub role-based operation validator.
2. **Worker poller follow-up fix** (`fix`) — drains pending work even when the in-memory tenant list is empty (the scenario self-registering extensions create). Follow-up to #1555.

Both changes are additive and backward-compatible: behavior is unchanged unless the new extensions are explicitly selected via environment variables.

## Tenant extensions

Adds two built-in `TenantExtension` implementations alongside the existing `ApiKeyTenantExtension` / `SupabaseTenantExtension`:

- **`OidcTenantExtension`** — generic OIDC/OAuth bearer-token auth via JWKS, with per-user schema isolation. Works with any standards-compliant IdP.
- **`GitHubTenantExtension`** — GitHub OAuth: validates the token via `GET /user` and `GET /user/teams`, derives a per-user schema and team-based roles.
- **`GitHubRoleOperationValidator`** — gates `retain`/`recall`/`reflect` by the GitHub team-derived role.

Activated only when `HINDSIGHT_API_TENANT_EXTENSION` points at them — no default behavior change.

### Control Plane (BFF) auth made provider-agnostic

The Control Plane now resolves the downstream dataplane `Authorization` header through a new `resolveDataplaneAuthHeader()` cascade:

1. Forward an inbound `Authorization` / `X-Forwarded-Access-Token` header when present (identity-aware proxy or direct caller) — preserves end-to-end per-user identity for tenant isolation.
2. Else fall back to the static `HINDSIGHT_CP_DATAPLANE_API_KEY` (historical single-tenant behavior).
3. Else no auth (local dev / internally-trusted network).

This forwards an opaque `Authorization` header rather than hard-coding any IdP, so the same path works for GitHub OAuth, OIDC, API gateways, or static keys — **no coupling to oauth2-proxy or any specific provider**.

## Worker fix (follow-up to #1555)

#1555 fixed the `'public'` vs `None` normalization mismatch in `_scan_active_schemas()`. While integrating self-registering tenant extensions I hit two remaining blind spots in `claim_batch()` in the same code path:

1. `if not schemas: return []` early-exits whenever `_get_schemas()` (the in-memory tenant list from `list_tenants()`) is empty. A worker process serves no authenticated user requests, so extensions that only learn about a schema during a request leave the worker's tenant list empty — and the worker never even scans for work.
2. Claiming is built from the tenant list rather than the scan results, so a schema discovered **only** by `public.schemas_with_pending_work()` is never claimed.

Fix: drop the early-exit and union scan-discovered schemas with the in-memory tenant list before claiming (preserving `_get_schemas()` ordering for stable rotation). Pending work now drains regardless of whether the tenant list is populated.

## Tests

- New unit tests: `test_oidc_tenant.py`, `test_github_tenant.py`, `test_github_role_validator.py`.
- New worker regression test `test_claim_batch_claims_scan_only_schema_when_list_tenants_empty` covering the empty-tenant-list path.
- `ruff check` and `ruff format --check` pass on all changed Python.

## Notes

- Docs added under `docs/developer/extensions.md`.
- No new dependencies (no `pyproject.toml` / lockfile changes).

### DIFF
--- a/hindsight-api-slim/hindsight_api/extensions/__init__.py
+++ b/hindsight-api-slim/hindsight_api/extensions/__init__.py
@@ -18,7 +18,10 @@ with the system (e.g., running migrations for tenant schemas).
 from hindsight_api.extensions.base import Extension
 from hindsight_api.extensions.builtin import (
     ApiKeyTenantExtension,
+    GitHubRoleOperationValidator,
+    GitHubTenantExtension,
     MemoryDefenseRegexExtension,
+    OidcTenantExtension,
     SupabaseTenantExtension,
 )
 from hindsight_api.extensions.context import DefaultExtensionContext, ExtensionContext
@@ -118,6 +121,9 @@ __all__ = [
     # Tenant/Auth
     "ApiKeyTenantExtension",
     "SupabaseTenantExtension",
+    "OidcTenantExtension",
+    "GitHubTenantExtension",
+    "GitHubRoleOperationValidator",
     "AuthenticationError",
     "RequestContext",
     "Tenant",
--- a/hindsight-api-slim/hindsight_api/extensions/builtin/__init__.py
+++ b/hindsight-api-slim/hindsight_api/extensions/builtin/__init__.py
@@ -7,18 +7,29 @@ They can be used directly or serve as examples for custom implementations.
 Available built-in extensions:
     - ApiKeyTenantExtension: Simple API key validation with public schema
     - SupabaseTenantExtension: Supabase JWT validation with per-user schema isolation
+    - OidcTenantExtension: Generic OIDC/OAuth (JWKS) auth with per-user schema isolation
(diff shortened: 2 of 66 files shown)


nicoloboschi:
while I think the 2 extensions are good ideas, we're having too many builtin extensions at the moment

also this PR has 3-4 different fixes in one PR, so please split them

I think we should create a folder with all community extensions, will do and then I will add yours

nicoloboschi:
closing for inactivity

PR #2485, opened 2026-07-01
fix(retain): honor configured LLM temperature in the batch fact-extraction path (#2469 follow-up)

### LINKED ISSUE #2459: Fix LiteLLM provider temperature handling for GPT-5.5

r266-tech:
## Summary

#2469 (fixes #2459) made per-operation temperature configurable so models like Azure `gpt-5.5` — which reject *any* explicit `temperature` — work with `HINDSIGHT_API_LLM_TEMPERATURE=none`. It de-hardcoded the streaming fact-extraction call (`temperature=config.llm_temperature_retain` at `_extract_facts_from_chunk`), but the **batch** path was missed: `_build_request_body` still hardcodes `"temperature": 0.1`.

So on the batch retain path, `HINDSIGHT_API_LLM_TEMPERATURE=none` is silently ignored and Azure `gpt-5.5` keeps rejecting the request with `Unsupported value: 'temperature' does not support 0.1 with this model. Only the default (1) value is supported.` — the exact failure #2469 set out to fix.

## Change

`_build_request_body` now honours `config.llm_temperature_retain`, and **omits** the field entirely when it is `None`, mirroring `LLMProvider.call` (`if temperature is not None`). The `0.0` case is preserved via `is not None` (not a falsy check).

## Tests

- `test_build_request_body_forwards_configured_temperature` — configured value reaches the request body
- `test_build_request_body_omits_temperature_when_none` — `none` drops the field for temperature-rejecting models

Follow-up to merged #2469.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -1203,8 +1203,16 @@ def _build_request_body(llm_config, config, prompt: str, user_message: str, resp
     request_body = {
         "model": llm_config.model,
         "messages": [{"role": "system", "content": prompt}, {"role": "user", "content": user_message}],
-        "temperature": 0.1,
     }
+
+    # Honour the configured retain temperature. ``None`` omits the parameter
+    # entirely (for models like Azure GPT-5.5 that reject explicit temperatures),
+    # mirroring LLMProvider.call, which drops temperature when it is None. The
+    # batch path builds the request body directly instead of going through
+    # LLMProvider.call (#2469 only de-hardcoded the streaming path), so it must
+    # apply the same rule here.
+    if config.llm_temperature_retain is not None:
+        request_body["temperature"] = config.llm_temperature_retain
 
     # Add max_completion_tokens if configured
     if config.retain_max_completion_tokens:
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Verified against main: batch `_build_request_body` was the one temperature call site #2469 missed. Fix mirrors LLMProvider.call's `is not None` drop-when-None pattern (preserves 0.0), tests cover both branches. LGTM.

PR #2487, opened 2026-07-01
fix(claude-code): don't retain Claude Code injected/synthetic context

hankel-ai:
## Problem

The Claude Code auto-retain hook stores a lot of behind-the-scenes context that the user never actually sent, polluting the memory bank with harness noise (skill docs, settings schemas, slash-command expansions, system reminders, command output). In practice this drowns real project/preference facts under injected-context memories.

Two root causes, plus a Windows packaging papercut:

1. **`read_transcript()` ignores `isMeta`.** Claude Code marks injected/synthetic transcript turns with `isMeta: true` (skill documentation, settings-schema dumps, slash-command expansions). These were being retained as if the user authored them.

2. **Harness context rides inside real turns.** Genuine user/assistant turns can embed harness-injected context in XML wrappers — `<system-reminder>`, `<task-notification>`, `<local-command-stdout>`, and `<command-name>/<command-message>/<command-args>`. None of that was sent by the user, but it was being retained verbatim.

3. **Shell hooks get CRLF on Windows.** With the Git-for-Windows default `core.autocrlf=true`, `scripts/*.sh` check out with CRLF endings, which breaks their `\` line-continuations and shebang under Git Bash.

## Changes

- `scripts/retain.py`: skip transcript entries where `isMeta` is set.
- `scripts/lib/content.py`: add `strip_injected_context()` (removes the Claude Code harness wrappers) and call it from `strip_memory_tags()`, so every existing retention/recall call site benefits.
- `hindsight-integrations/claude-code/.gitattributes`: pin `*.sh text eol=lf`.
- `tests/test_content.py`: unit tests for `strip_injected_context()` and the `strip_memory_tags()` path.

## Testing

`PYTHONPATH=scripts python -m pytest tests/test_content.py` → all pass (63). Full-suite failures on my Windows box are pre-existing and unrelated (path-separator/symlink assertions in `test_run_mcp.py`, `test_config.py`, `test_bank.py`) — they fail identically on a pristine checkout.

### DIFF
--- a/hindsight-integrations/claude-code/.gitattributes
+++ b/hindsight-integrations/claude-code/.gitattributes
@@ -1,0 +1,4 @@
+# Shell scripts must always check out with LF endings. On Windows, a global
+# core.autocrlf=true otherwise rewrites them to CRLF, which breaks the `\`
+# line-continuations and the shebang when the hooks run under Git Bash.
+*.sh text eol=lf
--- a/hindsight-integrations/claude-code/scripts/lib/content.py
+++ b/hindsight-integrations/claude-code/scripts/lib/content.py
@@ -56,9 +56,13 @@ def strip_injected_context(content: str) -> str:
     wrappers (``<system-reminder>``, ``<command-*>``, ...). These were never
     sent by the user, so drop them before retention to keep memories limited to
     the messages actually exchanged.
+
+    The opening-tag pattern (``<tag\b[^>]*>``) tolerates attributes, mirroring
+    ``strip_channel_envelope`` — so a future ``<system-reminder priority="x">``
+    still matches instead of silently slipping through.
     """
     for tag in INJECTED_CONTEXT_TAGS:
-        content = re.sub(rf"<{tag}>[\s\S]*?</{tag}>", "", content)
+        content = re.sub(rf"<{tag}\b[^>]*>[\s\S]*?</{tag}>", "", content)
     return content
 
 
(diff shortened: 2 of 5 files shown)


nicolas-vivot:
Found the exact same issue and was going to push a fix for that. Thank you so much for addressing it :)

nicolas-vivot:
@hankel-ai 
A few things to consider (nice to have):

1. **Add a `read_transcript` test covering the `isMeta` skip.** This is the primary behavior introduced by the change, but it currently isn't covered. The new tests only exercise `strip_injected_context()` and `strip_memory_tags()`, so a regression in `read_transcript()` could go unnoticed.

2. **Keep `strip_injected_context()` separate from `strip_memory_tags()`.** Having one concern embedded inside an unrelated function makes the behavior implicit and easy to overlook at the call sites. It may be clearer to either call `strip_injected_context()` explicitly where needed, or introduce a generic wrapper (for example, `sanitize_content()`) that composes both operations while keeping each function focused on a single responsibility.

3. **Consider making the tag regex more resilient to future changes.** The current pattern only matches tags without attributes. If attributes are added later, the stripping would silently stop working. `strip_channel_envelope()` in the same module already handles this more robustly using `<channel\b[^>]*>`. Following the same approach here would make the implementation less tightly coupled to the current tag format.

hankel-ai:
Thanks @nicolas-vivot — all three addressed in the latest push:

1. **`read_transcript` `isMeta` coverage** — added `TestReadTranscript` in `tests/test_hooks.py`: isMeta turns are dropped, non-meta turns kept, nested `{type, message:{role, content}}` unwrapping, and missing-file handling.
2. **Separated the concerns** — `strip_injected_context()` is no longer embedded inside `strip_memory_tags()`. Each is single-responsibility again, and a new `sanitize_content()` composes the two; the four call sites now go through it, so behavior is unchanged where both strips were wanted.
3. **Attribute-resilient regex** — the injected-context and memory-tag patterns now use `<tag\b[^>]*>` (same style as `strip_channel_envelope()`), so a future `<system-reminder priority="…">` or `<hindsight_memories count="…">` still matches. Added tests for both.

handnewb (commented):
This is a solid fix — Claude Code injects harness-level scaffolding (PreToolUse, system reminders) into the conversation that shouldn't be retained as memories. The filter in content.py is the right place. One suggestion: consider adding a config flag/env var to make this filtering opt-out, in case some users actually want the harness context retained (e.g., for debugging memory quality). Otherwise LGTM!

nicoloboschi:
Closing: superseded by **Coding Agents**, where injected and synthetic context is filtered structurally rather than by content matching.

The reader drops `isMeta` and `isSidechain` records outright, and a tag-structural rule strips the host's own transport wrappers from retained text — Hindsight's own injection blocks, codex's `<hook_prompt>`, Claude Code's `<task-notification>`, and `<system-reminder>`. Measuring real transcripts while fixing that turned up two leaks worth more than the originally reported ones (#3378, #3383).

`npx @vectorize-io/hindsight-coding-agents install claude-code` ([docs](https://hindsight.vectorize.io/sdks/integrations/coding-agents)). Thanks for pushing on this — retaining the harness's own words as the user's is exactly the kind of quiet memory-quality bug worth chasing.

PR #2490, opened 2026-07-01
Show full memory details in explore

### LINKED ISSUE #2482: hindsight explore - how to check details of selected entry?
### Use Case

when I explore memory and select some entry I expect to see ALL attributes which I see in json output format - id, entities, etc.

### Problem Statement

I can check 'text' only 

I can easily find needed observation/world/experience...
and then use its 'id' or other attributes in queries/code

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Priority

### Additional Context

_No response_

### Checklist

koriyoshi2041:
## Summary
- add a Details JSON panel when viewing a memory in `hindsight explore`
- keep the existing metadata and full-text panels while exposing fields such as id, entities, and metadata
- add a focused CLI unit test for non-text attributes in the details output

Fixes #2482

## Tests
- `cargo test --manifest-path hindsight-cli/Cargo.toml memory_details_json -- --nocapture`
- `git diff --check`

## Risk
Low. This only changes the memory detail view rendering and leaves list navigation/data loading untouched.

### DIFF
--- a/hindsight-cli/src/commands/explore.rs
+++ b/hindsight-cli/src/commands/explore.rs
@@ -931,7 +931,8 @@ fn render_memories(f: &mut Frame, app: &mut App, area: Rect) {
             .direction(Direction::Vertical)
             .constraints([
                 Constraint::Length(7),  // Memory metadata
-                Constraint::Min(0),     // Full text content
+                Constraint::Percentage(45), // Full text content
+                Constraint::Percentage(55), // Complete JSON details
             ])
             .split(area);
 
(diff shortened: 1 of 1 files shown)


nicoloboschi:
@koriyoshi2041 merge conflicts

koriyoshi2041:
Rebased on current main and resolved the conflict with #2489 by keeping both explore test groups in the shared test module.

Checked locally: cargo test --manifest-path hindsight-cli/Cargo.toml commands::explore::tests -- --nocapture

Issue #1270, opened 2026-04-26
hindsight-openclaw 0.6.x: default bankMission clobbers per-bank missions on every gateway restart via createBank({reflectMission}) upsert


willfivestar:
## Summary

In `hindsight-openclaw` 0.6.x with `dynamicBankGranularity` configured (multi-agent setup, one bank per agent), the plugin substitutes a hard-coded default `bankMission` text whenever the user sets `bankMission: ""` in plugin config. The substituted default then gets stamped onto every dynamic bank on first touch each gateway lifetime via the new `createBank({reflectMission: ...})` upsert path — **clobbering any per-bank `mission` and `reflect_mission`** that were written out-of-band (via API `PATCH /v1/default/banks/{id}` or MCP `update_bank`).

The intent of `bankMission: ""` is "no shared mission, manage per-bank out-of-band" — but the substitution + upsert defeats that intent.

0.5.x had the same substitution but its older `setBankMission` API didn't clobber existing values, so the bug was latent.

## Environment

- Plugin: `@vectorize-io/hindsight-openclaw@0.6.5` (latest at time of report)
- Daemon: `hindsight-api-slim==0.5.0` + `pg0-embedded==0.12.3` (embedded mode)
- Mode: external API (`hindsightApiUrl: http://127.0.0.1:9077`)
- OpenClaw: 2026.4.10 on macOS (Apple Silicon, M3 Pro)
- Plugin config (relevant fields):
  ```json
  {
    "bankMission": "",
    "dynamicBankId": true,
    "dynamicBankGranularity": ["agent"],
    "bankIdPrefix": "prod",
    "statelessSessionPatterns": ["agent:*:subagent:**"],
    "ignoreSessionPatterns": ["agent:*:cron:**", "agent:*:heartbeat:**"]
  }
  ```
- Setup: 6 gateway agents (`main`, `thinker`, `builder`, `writer`, `research`, `cron-worker`), each with its own bank (`prod-main`, `prod-thinker`, etc.). Per-bank missions reflect each agent's role (Cooper, Stark, Fox, Oswald, Kip, Duncan).

## Reproduction

1. Have multiple banks under `dynamicBankGranularity: ["agent"]`. Set per-bank missions out-of-band:
   ```bash
   curl -X PATCH http://127.0.0.1:9077/v1/default/banks/prod-main \
     -H "Content-Type: application/json" \
     -d '{"mission":"You are Cooper, the main orchestrator..."}'
   ```
[...]

PR #2493, opened 2026-07-01
fix(claude-code): seed bank missions only when unset; never overwrite existing (#2492)

### LINKED ISSUE #2492: hindsight-claude-code: default bankMission/retainMission clobbers per-bank missions on first touch (unported regression of #1270)
## Summary

The Claude Code plugin stamps its own `bankMission` / `retainMission` onto a bank the first time it touches that bank, **overwriting per-bank `reflect_mission` and `retain_mission` that were set out-of-band** (control plane UI / `PATCH /v1/default/banks/{id}/config`).

This is the **same bug class as #1270** (fixed for OpenClaw by #1473), but the fix was never ported to the Claude Code plugin — even though `scripts/lib/bank.py` documents itself as a "Port of Openclaw's `deriveBankId()` and `banksWithMissionSet` logic." The port copied the clobber behavior but not the fix.

The shipped `settings.json` makes this hit every default install, because it ships **non-empty** mission defaults:

```json
// hindsight-integrations/claude-code/settings.json
"bankMission":   "You are a Claude Code AI assistant. Focus on technical discussions, decisions, and context relevant to the user's projects.",
"retainMission": "Extract technical decisions, architectural choices, user preferences, project context, and people/tool relationships. Ignore routine greetings and transient operational details.",
```

## Environment

- Plugin: `hindsight-integrations/claude-code` (current `main`)
- Daemon: `hindsight-api-slim` **0.8.3** (also reproduced on 0.8.1)
- Mode: external API

## Reproduction

1. Create a bank (e.g. `testing`) via the control plane with custom missions:
   - reflect mission: `"You are an artist, all you care about is paintings"`
   - retain mission: a custom string
2. Confirm they persist (`GET /v1/default/banks/testing/config`).
3. Point the Claude Code plugin at that bank and run any session (recall on `UserPromptSubmit`, or retain at session end).
4. Re-check the config — both missions are now the plugin's `settings.json` defaults quoted above.

## Root cause

`ensure_bank_mission()` unconditionally PATCHes `reflect_mission` + `retain_mission` from plugin config, guarded only by a **process-local state file** that tracks "have *I* pushed a mission," never "does the bank already have one server-side":

```python
# scripts/lib/bank.py  (ensure_bank_mission)
missions_set = read_state("bank_missions.json", {})
if bank_id in missions_set:
    return
retain_mission = config.get("retainMission")
client.set_bank_mission(bank_id, mission, retain_mission=retain_mission, timeout=10)
```

```python
# scripts/lib/client.py  (set_bank_mission)
path = f"/v1/default/banks/{...}/config"
updates = {"reflect_mission": mission}
if retain_mission:
    updates["retain_mission"] = retain_mission
return self.request("PATCH", path, {"updates": updates}, timeout=timeout)
```

The server-side PATCH is a blind JSONB merge (`config = COALESCE(config,'{}') || $1`, `config_resolver.py`) — by design — so whatever the plugin sends wins.

Because the guard is a local file, it treats any bank authored elsewhere (control plane, another machine, fresh checkout, deleted state file) as "unset" and stomps it. Both the recall path (`scripts/recall.py:126`) and the retain path (`scripts/retain.py:155`) call it, so it can fire on the very first prompt.

## Impact

- Per-bank missions set via the control plane / API are silently overwritten.
- Degrades retain extraction and reflect output (missions steer both).
- Silent — only logged at debug level.

## Not fixed by any release

The clobber code is present in the current `main` / 0.8.3 tree. Server 0.8.1 → 0.8.3 does not address it (the behavior is client-driven; server merge semantics are unchanged and intentional).

## Suggested fixes (mirror #1473)

1. Treat empty `bankMission`/`retainMission` as explicit opt-out (already partly true via the empty-check — but the shipped `settings.json` overrides it with non-empty defaults).
2. Ship `settings.json` with empty mission defaults (match `config.py` defaults) so the plugin doesn't stamp unless the user opts in.
3. Make `ensure_bank_mission` idempotent against the **server**: `GET` the bank config first and only set mission fields that are currently empty — never overwrite a non-empty value. Removes reliance on the fragile local `bank_missions.json`.
4. (Optional, server-side) support a `set-if-absent` semantic so first-run defaulting can't clobber operator-authored missions.

## References

- #1270 — same bug in the OpenClaw plugin (closed)
- #1473 — the OpenClaw fix (`fix(openclaw): mission semantics + retainQueue config whitelist`), which modified only `hindsight-integrations/openclaw/` and was never ported here
- #1758 — separate open Claude Code plugin bugs (Windows paths, surrogate crash); this mission bug is not among them

nicolas-vivot:
## Summary

Fixes #2492 — the Claude Code plugin overwrites per-bank missions authored out-of-band (control plane UI or `PATCH /banks/{id}/config`) the first time it touches a bank. This is the Claude Code counterpart of the OpenClaw bug #1270 (fixed by #1473), which was never ported here even though `scripts/lib/bank.py` documents itself as a port of OpenClaw's `banksWithMissionSet` logic.

## The bug

`ensure_bank_mission` unconditionally PATCHed `reflect_mission` + `retain_mission` from the plugin's configured `bankMission`/`retainMission`, guarded only by a **local** state file (`bank_missions.json`) tracking "have *I* pushed a mission" — never "does the bank already have one server-side". Consequences:

- A bank configured elsewhere (control plane, API, another machine) is treated as "unset" and gets stomped.
- The clobber re-fires whenever the local state is lost (fresh machine, cleared cache).
- It runs on both recall (`UserPromptSubmit`) and retain (session end), so it can fire on the very first prompt.
- The server-side PATCH is a blind JSONB merge (`config || $1`), so whatever the plugin sends wins.

A straight port of #1473 was insufficient: that fix only stops stamping when the config field is *empty*, but the Claude Code `settings.json` ships **non-empty** mission defaults, so it would still clobber a differing per-bank mission.

## The fix

`ensure_bank_mission` is now **server-aware**: it reads the bank's existing overrides and only fills mission fields that are not already set — user-authored missions always win, the plugin merely seeds gaps.

- **`client.py`**
  - New `get_bank_config(bank_id)` → returns the bank's config; a `404` (bank not created yet — banks are lazily created on first retain) is treated as "empty" so brand-new banks are still seeded.
  - `set_bank_mission(...)` now takes `reflect_mission=` / `retain_mission=` and sends **only the fields provided** (no-ops if none). This also fixes a pre-existing quirk where `retainMission` could not be seeded unless `bankMission` was also configured.
- **`bank.py`** — `ensure_bank_mission` GETs `overrides`, computes per-field gaps, and PATCHes only the missing ones. The local `bank_missions.json` is demoted to a **perf cache** (avoids a GET on every hook); correctness now comes from the server check, so a wiped state file can no longer re-trigger a clobber.
- Transient failures (unreachable server, timeout, non-404 HTTP errors) skip cleanly **without** marking the bank reconciled, so they retry on the next hook rather than seeding or clobbering.

## Behaviour matrix

| Server state (bank overrides) | Plugin configured | Result |
|---|---|---|
| both missions set | both | no PATCH (missions preserved) |
| reflect set, retain empty | both | PATCH retain only |
| both empty / bank 404 | both | seed both |
| both empty | retain only | seed retain only |
| unreachable / timeout / 5xx | any | skip, retry next hook (no write) |

## Edge cases intentionally left as-is

- **Non-default retain strategies.** The "already set?" check inspects only the *root* `overrides.retain_mission`. But the server auto-applies `retain_default_strategy` when a retain call sends no strategy (`memory_engine.py`), and `apply_strategy` overlays that strategy's `retain_mission` over the root (`config_resolver.py`). So if the root retain mission is empty but a default strategy carries its own, the plugin still PATCHes a root retain mission. **Impact is non-functional** — the default strategy's mission still wins at retain time, so this only writes an otherwise-masked root value (semantic pollution, not a behaviour clobber). Fully resolving this would couple the client to server strategy-resolution internals (which also have global defaults), so it was deliberately deferred.
- **404 is overloaded.** The server returns `404` both when a bank doesn't exist *and* when the bank-config API is disabled (`HINDSIGHT_API_ENABLE_BANK_CONFIG_API=false`). In the disabled case the plugin does a harmless GET+PATCH that both 404, never marks reconciled, and retries each hook — no clobber, just mildly chatty at debug level.
- **Server-side hardening not included.** The PATCH endpoint still has no "set-if-absent" semantics for mission fields; this PR fixes the client behaviour only. A server-side guard could be a separate change.

## Scope

This PR is **Patch 1 only** (the behavioural fix). Blanking the shipped `settings.json` mission defaults is a separate, independently-adoptable change and will come as its own PR.

## Tests

`tests/test_bank.py` and `tests/test_client.py` updated/extended: new-bank seeding, existing-mission preservation, partial (gap-only) fill, retain-only config, local fast-path, opt-out, graceful-error-then-retry, plus `get_bank_config` (200 / 404 / non-404) and independent-field `set_bank_mission`. Full plugin suite: 201 passing.

### DIFF
--- a/hindsight-integrations/claude-code/CHANGELOG.md
+++ b/hindsight-integrations/claude-code/CHANGELOG.md
@@ -22,6 +22,18 @@
   `HINDSIGHT_USER_ID` is unset) are now dropped from retain requests. Previously
   such tags were sent as-is. Tags without `:` are unaffected.
 
+### Fixed
+
+- Bank mission seeding no longer overwrites per-bank `reflect_mission` /
+  `retain_mission` that were authored out-of-band (control plane UI or
+  `PATCH /banks/{id}/config`). `ensure_bank_mission` now reads the bank's
+  existing overrides first and only fills mission fields that are not already
+  set, so shared banks keep their configured missions. Correctness no longer
+  depends on the local `bank_missions.json` state file (which reset per machine
+  and re-triggered the clobber). Also fixes an issue where `retainMission`
+  could not be seeded unless `bankMission` was also configured. Fixes #2492
+  (Claude Code counterpart of #1270 / #1473).
+
 ## [0.1.0] - 2025-03-23
 
 ### Added
--- a/hindsight-integrations/claude-code/README.md
+++ b/hindsight-integrations/claude-code/README.md
@@ -205,8 +205,8 @@ A **bank** is an isolated memory store — like a separate "brain." These settin
 | Setting | Env Var | Default | Description |
 |---------|---------|---------|-------------|
 | `bankId` | `HINDSIGHT_BANK_ID` | `"claude_code"` | The bank ID to use when `dynamicBankId` is `false`. All sessions share this single bank. |
-| `bankMission` | `HINDSIGHT_BANK_MISSION` | generic assistant prompt | A short description of the agent's identity and purpose. Sent to Hindsight when creating or updating the bank, and used during recall to contextualize results. |
-| `retainMission` | — | extraction prompt | Instructions for the fact extraction LLM — tells it *what* to extract from conversations (e.g. "Extract technical decisions and user preferences"). |
+| `bankMission` | `HINDSIGHT_BANK_MISSION` | generic assistant prompt | A short description of the agent's identity and purpose (the bank's reflect mission). Seeded onto the bank on first use **only if the bank has no reflect mission yet** — an existing per-bank mission (set via the control plane or API) is never overwritten. |
+| `retainMission` | — | extraction prompt | Instructions for the fact extraction LLM — tells it *what* to extract from conversations (e.g. "Extract technical decisions and user preferences"). Seeded onto the bank on first use only if no retain mission is set yet; never overwrites an existing one. |
 | `dynamicBankId` | `HINDSIGHT_DYNAMIC_BANK_ID` | `false` | When `true`, the plugin derives a unique bank ID from context fields (see `dynamicBankGranularity`), giving each combination its own isolated memory. |
 | `dynamicBankGranularity` | — | `["agent", "project"]` | Which context fields to combine when building a dynamic bank ID. Available fields: `agent` (agent name), `project` (working directory), `session` (session ID), `channel` (channel ID), `user` (user ID). |
 | `bankIdPrefix` | — | `""` | A string prepended to all bank IDs — both static and dynamic. Useful for namespacing (e.g. `"prod"` or `"staging"`). |
--- a/hindsight-integrations/claude-code/scripts/lib/bank.py
+++ b/hindsight-integrations/claude-code/scripts/lib/bank.py
@@ -143,37 +143,83 @@ def derive_bank_id(hook_input: dict, config: dict) -> str:
     return f"{prefix}-{base_bank_id}" if prefix else base_bank_id
 
 
-def ensure_bank_mission(client, bank_id: str, config: dict, debug_fn=None):
-    """Set bank mission on first use, skip if already set.
+def _mark_bank_reconciled(bank_id: str) -> None:
+    """Record that we've reconciled missions for a bank (local perf cache).
 
-    Port of: banksWithMissionSet Set tracking in index.js
-
-    Uses a state file to persist which banks have had their mission set
-    across ephemeral hook invocations.
+    This is only an optimization to avoid a GET on every hook invocation — it
+    is NOT the correctness boundary. Correctness comes from checking the
+    server's per-bank overrides before writing (see ensure_bank_mission).
     """
-    mission = config.get("bankMission", "")
-    if not mission or not mission.strip():
+    missions_set = read_state("bank_missions.json", {})
+    missions_set[bank_id] = True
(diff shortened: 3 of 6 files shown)


nicoloboschi:
Closing: superseded by **Coding Agents**, which now does what this PR argued for — shipped in v0.2.1 (#3352), against your issue #2492.

The rule there is seed-once. Before importing the bank template the client reads the bank-scoped *overrides*; if any of `reflect_mission` / `retain_mission` / `observations_mission` is already set — ours from an earlier pass or the user's own edit — it imports a structure-only manifest that omits them. Omitted fields are untouched server-side, so an edit made in the control plane survives every subsequent seed pass.

Retain strategies and entity labels are still re-applied, because those aren't preferences: the plugin writes documents under `git`/`gitlog`/`conversation`/`document` and a bank missing one would reject the write.

`npx @vectorize-io/hindsight-coding-agents install claude-code` ([docs](https://hindsight.vectorize.io/sdks/integrations/coding-agents)). Thanks for tracing it back to #1270 — that the OpenClaw fix was never ported is exactly what made it worth fixing properly rather than patching the symptom.

PR #2494, opened 2026-07-01
chore(claude-code): drop opinionated bankMission/retainMission defaults (#2492)

nicolas-vivot:
## Summary

Part of #2492. Removes the opinionated `bankMission` / `retainMission` defaults shipped in the Claude Code plugin's `settings.json`, which are what triggered the mission-clobber for default installs.

The bundled `settings.json` shipped:

```json
"bankMission":   "You are a Claude Code AI assistant. Focus on technical discussions, decisions, and context relevant to the user's projects.",
"retainMission": "Extract technical decisions, architectural choices, user preferences, project context, and people/tool relationships. Ignore routine greetings and transient operational details.",
```

So any default install stamped a generic persona onto a bank the first time the plugin touched it — overwriting per-bank missions authored via the control plane / API. This blanks both to match the plugin's own internal defaults (`config.py`: `bankMission=""`, `retainMission=None`), so the plugin only seeds missions the user explicitly configures.

## Changes

- **`settings.json`** — `bankMission: ""`, `retainMission: null`.
- **`README.md`** — config table default column updated to `""` / `null` (unset), documenting opt-in seeding.
- **`CHANGELOG.md`** — `Changed` entry.

## Behaviour

No behavioural change when the fields are empty: `bank.py` already treats an empty/`None` mission as an opt-out and sends nothing. `""` vs `null` is a typing convention only (`bankMission` is a `str`, `retainMission` is `Optional[str]`) — both are falsy and indistinguishable to the code.

## Relationship to the other PR

This is the **defaults-removal half** of #2492, deliberately kept separate from the server-aware seeding fix (#2493) so each can be adopted independently:

- #2493 — makes `ensure_bank_mission` server-aware (never overwrites an existing per-bank mission). Fixes the clobber even for users who keep non-empty missions configured.
- **This PR** — stops default installs from seeding a generic persona at all.

Note: #2493 and this PR touch adjacent regions of `README.md` / `CHANGELOG.md`; depending on merge order a trivial conflict may need resolving.

## Tests

Full Claude Code plugin suite passes (192 tests); `settings.json` validated as JSON.

### DIFF
--- a/hindsight-integrations/claude-code/CHANGELOG.md
+++ b/hindsight-integrations/claude-code/CHANGELOG.md
@@ -21,6 +21,11 @@
 - Tags that resolve to an empty namespace content (e.g. `"user:"` when
   `HINDSIGHT_USER_ID` is unset) are now dropped from retain requests. Previously
   such tags were sent as-is. Tags without `:` are unaffected.
+- The bundled `settings.json` no longer ships opinionated `bankMission` /
+  `retainMission` defaults; both are now empty (matching the internal config
+  defaults). The plugin only seeds missions you explicitly configure, so a
+  default install never stamps a generic persona onto a bank. Set these
+  fields to opt into seeding. See #2492.
 
 ## [0.1.0] - 2025-03-23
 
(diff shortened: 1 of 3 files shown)


nicoloboschi (changes requested):
we should document it, not changing the default, thanks

nicolas-vivot:
@nicoloboschi Got it. Could you have a look at [this one](https://github.com/vectorize-io/hindsight/pull/2493)? (the real fix / non optional fix IMO) (fixes [this reported issue](https://github.com/vectorize-io/hindsight/issues/2492))

PR #2502, opened 2026-07-02
fix(consolidation): stop emitting unsupported maxItems that breaks all Bedrock consolidation (#2500)

### LINKED ISSUE #2500: max_observations_per_scope > 0 breaks ALL consolidation on Bedrock: emits unsupported "maxItems" JSON Schema keyword
## Bug Description

Setting `max_observations_per_scope` to any positive value (e.g. `50`) on a bank running on AWS Bedrock (any Claude model via Bedrock Converse) causes **100% of consolidation batch calls to fail**, not intermittently. Every batch fails after 3/3 retry attempts and is skipped.

## Error (from container logs)

```
litellm.BadRequestError: BedrockException - {"message":"The model returned the following errors: output_config.format.schema: For 'array' type, property 'maxItems' is not supported"}
```

Reproduced consistently, ~15+ consolidation batches failing within a 6-second window, 0 successful calls.

## Root Cause

`hindsight_api/engine/consolidation/consolidator.py`'s `_build_response_model()` dynamically subclasses the consolidation response model when an observation-scope limit is active:

```python
def _build_response_model(max_creates: int | None = None) -> type[_ConsolidationBatchResponse]:
    if max_creates is None or max_creates < 0:
        return _ConsolidationBatchResponse
    clamped = max(max_creates, 0)
    class _ConstrainedConsolidationBatchResponse(_ConsolidationBatchResponse):
        creates: list[_CreateAction] = PydanticField(default=[], max_length=clamped)
    return _ConstrainedConsolidationBatchResponse
```

Pydantic v2 maps `max_length` on a `list[...]`-typed field to the JSON Schema keyword `"maxItems"` (confirmed via `pydantic/json_schema.py`'s `ValidationsMapping.array = {'min_length': 'minItems', 'max_length': 'maxItems'}` — this is intentional, documented Pydantic behavior, not an edge case).

That schema is set as `response_format` and forwarded unmodified through `LiteLLMLLM.call()` (`engine/providers/litellm_llm.py`) into `litellm.acompletion(response_format={'type': 'json_schema', 'json_schema': {'schema': <the maxItems-constrained schema>, ...}})`.

**AWS Bedrock Converse / Anthropic's structured-output validator does not support `maxItems`** (or array-length constraints beyond `minItems` of 0 or 1) — confirmed via Anthropic's own docs ([platform.claude.com/docs — JSON Schema limitations](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#json-schema-limitations): "Not supported: Array constraints beyond minItems of 0 or 1"). LiteLLM's Bedrock Converse transformation path (`litellm/llms/bedrock/chat/converse_transformation.py`) never applies `filter_anthropic_output_schema` (the sanitizer that exists for the Anthropic-direct path) — same architectural gap already documented in this repo's own closed issue #1289 (a different keyword, `minimum`/`maximum` on `number` types, same root cause).

Independent, non-circular corroboration that Bedrock/Claude reject `maxItems` specifically: [BerriAI/litellm#19444](https://github.com/BerriAI/litellm/issues/19444) (Jan 2026, unrelated reporter, hitting `api.anthropic.com` directly) reproduces the identical error string: `"output_format.schema: For 'array' type, property 'maxItems' is not supported"`.

## Repro

1. Deploy Hindsight with `HINDSIGHT_API_LLM_PROVIDER=bedrock` and any Claude model
2. `PATCH /v1/default/banks/{bank_id}/config` with `{"updates": {"max_observations_per_scope": 50}}` (any value `> 0`)
3. Trigger consolidation (e.g. via retain + auto-consolidation, or `POST /v1/default/banks/{bank_id}/consolidate`)
4. Every batch fails 3/3 attempts with the `maxItems` `BedrockException`

## Environment

- Hindsight: v0.8.3 (confirmed unpatched through v0.8.4 — `_build_response_model` and the call chain into `litellm.acompletion` are unchanged between these versions)
- `HINDSIGHT_API_LLM_PROVIDER=bedrock`, model `us.anthropic.claude-haiku-4-5-20251001-v1:0`
- Access via AWS Bedrock Converse API

## Suggested Fix

Same pattern as the fix for #1289: either (a) drop the `max_length` constraint from the dynamically-built Pydantic model and instead enforce the cap via prompt guidance + post-hoc truncation of the LLM's response (the code already has a "defensive truncation... belt-and-suspenders" comment for LLM providers that ignore the schema constraint — this could become the primary enforcement mechanism instead of a backup), or (b) strip `minItems`/`maxItems` from the generated schema specifically on the Bedrock/Anthropic provider path before sending, mirroring what LiteLLM's Anthropic-direct filter already does.

## Workaround

Leave `max_observations_per_scope` at its default (`-1`, unlimited) when running on Bedrock with Claude models.

r266-tech:
Fixes #2500.

## Problem

Setting `max_observations_per_scope > 0` on a Bedrock-backed bank (any Claude model via Bedrock Converse) causes **100% of consolidation batches to fail** — every batch fails all 3 retries and is silently skipped, so memory quality decays permanently:

```
litellm.BadRequestError: BedrockException - {"message":"The model returned the following errors:
output_config.format.schema: For 'array' type, property 'maxItems' is not supported"}
```

## Root cause

`_build_response_model()` dynamically subclasses the consolidation response model when an observation cap is active and puts a Pydantic `max_length` on `creates`. Pydantic v2 serializes list `max_length` to the JSON-schema `maxItems` keyword, and Bedrock Converse rejects `maxItems` on array types outright — so a single `maxItems` in the structured-output schema breaks the whole call.

## Fix

Drop the schema-level constraint: `_build_response_model()` now always returns the base `_ConsolidationBatchResponse`, whose `creates` field carries no length constraint, so the emitted schema is provider-portable.

This is safe because the observation cap does **not** depend on the schema hint — it is enforced by two other layers that are unchanged:

1. **Prompt capacity note** — whenever the cap binds (`remaining_observation_slots <= len(memories)`, or `== 0`), the consolidation prompt already tells the model `"This scope has {k} observation slot(s) remaining … Prefer UPDATE over CREATE"` / `"OBSERVATION LIMIT REACHED … Do NOT create new ones"`. Every provider (Bedrock or not) still receives the cap.
2. **Authoritative truncation** — the batch-call path unconditionally does `creates = creates[:remaining_observation_slots]`, with your own comment *"some LLM providers may not enforce JSON schema max_length"*. This was already the guaranteed enforcement path.

## Known tradeoff (documented deliberately)

Providers that *did* honor `maxItems` lose that redundant schema-level ceiling and now rely on the prompt note + keep-first-N truncation. But keep-first-N was already your designated fallback for non-honoring providers, and consolidation creates are not priority-ordered, so this matches existing semantics. If you'd prefer to preserve the hint for compatible providers, a provider-gated variant (emit `maxItems` only when the provider supports it) is a straightforward follow-up — I kept this PR to the minimal, provably-safe change that unblocks Bedrock.

## Tests

`TestBuildResponseModel` is rewritten to the new contract:
- `test_returns_base_model_regardless_of_cap` — factory always returns the base model.
- `test_schema_omits_max_items` — **#2500 regression**: emitted schema must not contain `maxItems` on `creates`.
- `test_model_does_not_reject_creates_over_cap` — over-cap responses are accepted (truncated downstream), not rejected wholesale.

End-to-end cap enforcement remains covered by the existing `test_max_observations_per_scope_*` integration tests (which exercise the truncation path via a mock LLM, independent of schema validation).


nicoloboschi (changes requested):
@r266-tech can you make this change only true if a new config about the llm "supports max items" is set to false (default = true). the maxItems is actually very important here and removing from any llm is quite concerning to me

r266-tech:
Reworked this around an explicit capability flag as requested.

`HINDSIGHT_API_LLM_SUPPORTS_MAX_ITEMS` defaults to `true`, so existing schema-level `maxItems` enforcement remains unchanged. Operators on Bedrock Converse (or another backend that rejects the keyword) can set it to `false`; that choice is read from the operation's config, and the existing runtime truncation remains authoritative.

I also added zero-slot and positive-slot tests that make the LLM overproduce with schema enforcement disabled and assert the returned creates are still capped, plus environment/default and schema-shape coverage. The focused tests pass (10 tests), changed-source Ruff checks pass, and the reworked diff passed Codex adversarial review.

nicoloboschi (approved):
lgtm

nicoloboschi:
@r266-tech please rebase

r266-tech:
Rebased onto `a23187a4` as requested.

During the post-rebase adversarial review I also tightened the new capability flag: it now defaults to `true` for legacy direct `HindsightConfig` construction, and ambiguous environment values fail fast instead of silently disabling `maxItems`. The focused schema/config/runtime-cap tests pass (15 tests), along with changed-source Ruff and `ty` checks. CI is running on `ea9a6d77`.

PR #2509, opened 2026-07-02
fix(migrations): reconcile invalidated archive search_vector for pgroonga

### LINKED ISSUE #2503: invalidate_memory fails under pgroonga: invalidated_memory_units.search_vector left as tsvector (not reconciled by ensure_text_search_extension)
## Summary

On a self-hosted instance using the **pgroonga** text-search backend, `invalidate_memory` (soft-retire / curation) fails with:

```
column "search_vector" is of type tsvector but expression is of type text
```

Memory curation (invalidate / edit / delete) is completely broken on pgroonga as a result.

## Root cause

`invalidate_memory` copies the memory row from `memory_units` into the archive table `invalidated_memory_units`, including `search_vector`.

- Under pgroonga, `ensure_text_search_extension()` reconciles `search_vector` to a **dummy TEXT** column — but only for the tables in `tables_to_check`:

  ```python
  # hindsight_api/migrations.py  (ensure_text_search_extension)
  tables_to_check = [
      "memory_units",
      "reflections",
  ]
  ```

  `invalidated_memory_units` is **not** in that list, so its `search_vector` stays `tsvector` (as created by migration `c9a1b2d3e4f5_add_invalidated_memory_units`, mirroring the original tsvector schema).

- Result: `memory_units.search_vector` is `text` (pgroonga) while `invalidated_memory_units.search_vector` is `tsvector`. The archive `INSERT ... SELECT` from `memory_units` then tries to put a `text` value into a `tsvector` column → the error above.

This is the same class of bug as #2209 / #2210 (the archive table was missed by the embedding-dimension migration). That fix reasoned explicitly about the *embedding* column only (see the `# NOTE: invalidated_memory_units is deliberately omitted ...` comment in `_migrate_*_embedding_dimension`); the `search_vector` column type was not considered.

Confirmed present in current `main` / **v0.8.4** (checked `migrations.py`: `invalidated_memory_units` is still absent from `tables_to_check`).

## Reproduce

1. Deploy with `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=pgroonga` (fresh, or migrated native→pgroonga per #1973).
2. Retain any memory, then call `invalidate_memory` (MCP tool) / the curation invalidate endpoint on it.
3. → `column "search_vector" is of type tsvector but expression is of type text`.

Column types on the affected instance:

```
memory_units.search_vector             text
invalidated_memory_units.search_vector tsvector   <-- mismatch
mental_models.search_vector            tsvector    (GENERATED + GIN, intended)
```

## Suggested fix

Include `invalidated_memory_units` in `ensure_text_search_extension()`'s `tables_to_check` (i.e. keep the archive table's `search_vector` type in lockstep with `memory_units` across all non-native backends: pgroonga / pg_search / pg_textsearch / vchord). The archive column is only a passive copy, so it should always match the source table's type.

## Workaround

```sql
ALTER TABLE invalidated_memory_units
  ALTER COLUMN search_vector TYPE text USING search_vector::text;
```
(safe: not generated, no index; matches memory_units under pgroonga).

## Environment

- Hindsight 0.8.3-slim (self-hosted, Docker), Postgres 16 + pgroonga + pgvector
- `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=pgroonga`

xmh1011:
Summary
- include invalidated_memory_units in text-search column reconciliation without adding recall-path indexes to the cold archive
- allow archive-only TEXT reconciliation with existing rows, so pgroonga/pg_textsearch/pg_search deployments can convert inherited native tsvector columns safely
- keep hot tables strict: memory_units/reflections still require their configured text-search indexes, and a regression test covers missing-index recreation

Closes #2503.

Tests
- uv run ruff check hindsight_api/migrations.py tests/test_text_search_extension_archive.py
- uv run ruff format --check hindsight_api/migrations.py tests/test_text_search_extension_archive.py
- uv run --frozen ty check hindsight_api/migrations.py
- uv run --extra embedded-db pytest tests/test_text_search_extension_archive.py tests/test_migrations_parallel_schemas.py tests/test_migrations_newer_revision.py tests/test_migrations_thread_safety.py -q
- ./scripts/hooks/lint.sh

### DIFF
--- a/hindsight-api-slim/hindsight_api/migrations.py
+++ b/hindsight-api-slim/hindsight_api/migrations.py
@@ -852,10 +852,13 @@ def ensure_text_search_extension(
 
     engine = create_engine(to_libpq_url(database_url), poolclass=NullPool)
     with engine.connect() as conn:
-        # Tables with search_vector columns to check
+        # Tables with search_vector columns to check. The invalidation archive
+        # is cold storage, so its column must match memory_units for lossless
+        # INSERT ... SELECT moves, but it must not get a recall-path index.
         tables_to_check = [
-            "memory_units",
-            "reflections",  # Renamed from pinned_reflections in p1k2l3m4n5o6 migration
+            ("memory_units", True),
+            ("reflections", True),  # Renamed from pinned_reflections in p1k2l3m4n5o6 migration
+            ("invalidated_memory_units", False),
         ]
 
         # Determine target column type and index type
@@ -883,7 +886,7 @@ def ensure_text_search_extension(
         mismatched_tables = []
         tables_with_data = []
 
-        for table_name in tables_to_check:
+        for table_name, requires_text_search_index in tables_to_check:
             # Check if table exists
             table_exists = conn.execute(
                 text("""
@@ -913,7 +916,7 @@ def ensure_text_search_extension(
 
             if not current_column_info:
                 logger.warning(f"No search_vector column found for {table_name}, will create it")
-                mismatched_tables.append((table_name, None, None, False))
+                mismatched_tables.append((table_name, None, None, False, requires_text_search_index))
                 continue
 
             # Check column type (udt_name contains the actual type: tsvector, bm25vector, etc.)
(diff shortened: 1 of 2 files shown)


nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/2514

PR #2510, opened 2026-07-02
fix(openclaw): gate append on stored document text

### LINKED ISSUE #2505: openclaw: append-capability probe checks only API version, ignores features.store_document_text — every session retain 400s on text-disabled deployments
## Summary

The openclaw integration decides to use a session-scoped document with `update_mode: "append"` based solely on the server **version** (≥ 0.5.0). But `update_mode='append'` also requires `HINDSIGHT_API_STORE_DOCUMENT_TEXT` to be enabled — on deployments that disable it (a documented privacy/storage opt-out), every session-scoped retain is rejected with:

```
400: update_mode='append' is not supported when HINDSIGHT_API_STORE_DOCUMENT_TEXT is disabled:
the prior document text is not stored and cannot be appended to. Use update_mode='replace' instead.
```

(raised in `engine/memory_engine.py` retain-batch validation; covered by `tests/test_store_document_text.py::test_append_mode_rejected_when_text_disabled`).

The integration keeps retrying/accumulating failed retains — session memory is silently lost on such deployments (we hit exactly this in production with a derivative of this integration: dozens of failed `retain`/`batch_retain` ops over a week).

## Root cause

`hindsight-integrations/openclaw/src/index.ts`:

- ~line 2769: `useSessionScopedDoc = options?.appendSupported === true`
- ~line 2806: `updateMode: useSessionScopedDoc ? "append" : undefined`
- probe (~lines 1428-1449): fetches `/version` and compares against `MIN_VERSION_FOR_UPDATE_MODE_APPEND = 0.5.0` — **version only**.

Meanwhile `/version` already exposes exactly the needed flag:

```json
{"api_version": "0.8.4", "features": {"store_document_text": false, ...}}
```

## Suggested fix

In the append-capability probe, also require `features.store_document_text === true` (defaulting to `true` when `features` is absent, for pre-feature servers). When false, fall back to the already-existing per-process unique `document_id` path with no `update_mode` — extraction still works; only cross-process append dedup is lost, which is unavoidable on such servers by design.

We applied exactly this fix to our openclaw-derived client and the failures stopped; happy to PR the openclaw change if useful.

koriyoshi2041:
Fixes #2505.

## Summary
- parse `/version.features.store_document_text` during the OpenClaw append-capability probe
- fall back to per-turn document ids when stored document text is explicitly disabled
- keep the existing version-only behavior when older `/version` payloads omit feature metadata

## Tests
- `npx vitest run src/index.test.ts`
- `npm run build`
- `npm test`
- `git diff --check`

### DIFF
--- a/hindsight-integrations/openclaw/src/index.test.ts
+++ b/hindsight-integrations/openclaw/src/index.test.ts
@@ -13,6 +13,7 @@ import {
   truncateRecallQuery,
   buildRetainRequest,
   meetsMinimumVersion,
+  getAppendCapabilityFromVersionPayload,
   parseSessionKey,
   extractTelegramDirectSenderId,
   resolveSessionIdentity,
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Closing this in favor of #2511. I compared the two patches against #2505 and the current /version feature flag surface; #2511 keeps append gated on stored document text and covers the same OpenClaw path cleanly. Thanks!

PR #2515, opened 2026-07-02
fix(embeddings): cap LiteLLM SDK inputs over the model token limit (#2501)

### LINKED ISSUE #2501: Mental-model content grows unbounded via delta-refresh, permanently overflows embedding model's input token limit (no truncation before embed)
## Bug Description

A pinned mental model in `delta` refresh mode can accumulate `content` that grows monotonically across refresh cycles, eventually exceeding the configured embedding model's input token limit (e.g. AWS Bedrock Titan Text Embeddings V2's hard 8,192-token cap). When this happens, `refresh_mental_model` fails permanently — every retry hits the identical overflow, since the underlying stored content never shrinks.

## Error (from container logs)

```
litellm.exceptions.ContextWindowExceededError: litellm.ContextWindowExceededError: litellm.BadRequestError: BedrockException: Context Window Error - {"message":"400 Bad Request: Too many input tokens. Max input tokens: 8192, request input token count: 8481 "}
```

```
  File "hindsight_api/engine/retain/embedding_utils.py", line 98, in generate_embeddings_batch
    embeddings = await loop.run_in_executor(...)
  File "hindsight_api/engine/embeddings.py", line 128, in encode_documents
    return self.encode(texts)
  File "hindsight_api/engine/embeddings.py", line 1327, in encode
    response = self._litellm.embedding(**embed_kwargs)
```

Two separate mental models hit this independently within the same bank (input token counts 8341 and 8481, both over Titan V2's 8192 limit). A third and fourth model were found via direct measurement at ~7,357 and ~7,598 estimated tokens — 90%+ of the limit, not yet failed but one refresh cycle away.

## Root Cause

Two compounding gaps:

**1. Delta-mode content accumulation is unbounded.** In `memory_engine.py`'s `refresh_mental_model()`, when a structured delta patch applies successfully, the persisted content is:
```python
final_content = render_document(apply_operations(current_doc, ops).document)
```
`apply_operations()` (`reflect/delta_ops.py`) only appends/inserts/replaces blocks or sections — any section untouched by an operation is copied through verbatim, with zero size check anywhere in the function. The per-model `max_tokens` setting only bounds the LLM completion that *produces the ops list* (`delta_max_tokens = max(2048, doc_max_tokens * 1.5)`, passed as `max_completion_tokens`) — it does **not** bound the resulting merged/rendered document, which is a strict superset of all prior content plus new blocks. Over many refresh cycles on a stable `source_query`, this grows without limit. The `mental_models.content` database column is plain `TEXT` (Postgres) / `CLOB` (Oracle) with no length constraint at any layer.

**2. The embedding call has zero truncation.** `update_mental_model()` builds `embedding_text = f"{name} {content}"` and passes it straight to `embedding_utils.generate_embeddings_batch()` → `LiteLLMSDKEmbeddings.encode()` (`hindsight_api/engine/embeddings.py`), which sends it to `litellm.embedding()` with no length check, token count, or truncation anywhere in the call chain. The `batch_size` parameter used for batching is purely count-based (number of texts per call) and provides zero protection against a single oversized text.

The combination means: content grows unbounded (gap 1) until it silently crosses whatever the configured embedding model's real input limit is (gap 2), at which point every future refresh of that model fails permanently with no recovery path except manually clearing the model's content via `POST .../mental-models/{id}/clear`.

Note: `HINDSIGHT_API_REFLECT_SOURCE_FACTS_MAX_TOKENS` and per-model `max_tokens` are both real, correctly-enforced caps on *LLM completion calls* — they were verified working as designed during this investigation. Neither caps the *persisted document size* or the *embedding call*, which is the actual gap.

## Repro

1. Deploy Hindsight with any embedding model that has a fixed input-token limit smaller than what a mental model's content can reach after sustained delta-mode growth (e.g. Bedrock Titan V2, 8,192 tokens)
2. Let a `delta`-mode mental model with `refresh_after_consolidation: true` run through enough refresh cycles (weeks of normal use, in our case) for `content` to exceed the embedding model's limit
3. Next refresh's embedding call fails with `ContextWindowExceededError`; retries indefinitely with the same result

## Environment

- Hindsight: v0.8.3 (confirmed unpatched through v0.8.4 — `apply_operations`, `update_mental_model`, and the embedding call chain are byte-identical between versions for this path)
- `HINDSIGHT_API_EMBEDDINGS_PROVIDER=litellm-sdk`, `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL=bedrock/amazon.titan-embed-text-v2:0`
- Titan V2's 8,192-token input limit is a fixed model-architecture property per [AWS's own documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/titan-embedding-models.html) — not configurable, not an adjustable service quota, applies per individual text (not a combined batch total)

## Suggested Fix

At minimum, add a token-length check + truncation (or chunking) in `LiteLLMSDKEmbeddings.encode()` before the `litellm.embedding()` call, scoped to the target embedding model's actual known input limit — similar to the existing pattern in `HINDSIGHT_API_RERANKER_LITELLM_MAX_TOKENS_PER_DOC` (added for the reranker call site, never extended to embeddings). More completely, `apply_operations()`/`render_document()` in the delta-refresh path should also enforce a size ceiling on the merged document (e.g. relative to the mental model's own `max_tokens`), rather than only bounding the LLM call that produces the incremental patch.

## Workaround

Periodically call `POST /v1/default/banks/{bank_id}/mental-models/{id}/clear` on any model approaching the embedding limit, to reset accumulated content before it overflows. We deployed a size-aware CronJob that checks each model's content size every 10 minutes and clears anything at/above ~73% of the embedding limit.

r266-tech:
## Summary

Fixes #2501. A delta-refreshed mental model can grow past the embedding model's input token limit (e.g. Bedrock `amazon.titan-embed-text-v2:0` = 8192 tokens). Once it does, every scheduled refresh re-sends the same oversized input to `litellm.embedding`, which the provider rejects — so the model is pinned in a permanent refresh-failure loop until manually cleared.

## Change

`LiteLLMSDKEmbeddings`:
- Resolves the model's max input tokens at `initialize()` — a small known-model table (substring match) first, then `litellm.get_model_info(model)["max_input_tokens"]` as a fallback. Unknown ⇒ `None`.
- In `encode()`, truncates any oversized input **before** the embedding call, using `litellm.encode`/`litellm.decode` with a 0.95 safety ratio (leaving headroom for server-side tokenizer differences), and a proportional character-cut fallback when tokenization/decode is unavailable.
- When no limit is known (`None`), inputs are sent unchanged — **no behavior change** for models without a known limit.

This is a focused runtime guard; it does not redesign the delta-refresh document pipeline.

## Tests

Adds regression tests in `test_litellm_sdk_embeddings.py`:
- limit resolution: known-table / `get_model_info` fallback / unknown-is-None
- oversized single input truncated to the token budget
- empty-`decode()` still cuts an oversized input (never sends it whole — the exact #2501 failure mode)
- mixed-size batch preserves 1:1 input→vector alignment and ordering (only oversized inputs change)
- within-budget input is a no-op (no `decode` call)
- unknown-limit input is a no-op (no tokenization)

All existing `LiteLLMSDKEmbeddings` tests continue to pass.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/embeddings.py
+++ b/hindsight-api-slim/hindsight_api/engine/embeddings.py
@@ -1226,6 +1226,10 @@ class LiteLLMSDKEmbeddings(Embeddings):
         self.encoding_format = encoding_format or None
         self._litellm = None  # Will be set during initialization
         self._dimension: int | None = None
+        # Max per-input tokens accepted by the embedding model, resolved at
+        # initialize(). None means "unknown" -> inputs are sent unchanged
+        # (preserves prior behavior for models we have no limit for).
+        self._max_input_tokens: int | None = None
 
     @property
     def provider_name(self) -> str:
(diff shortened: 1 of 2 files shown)


nicoloboschi (changes requested):
I prefer if the user can set the limit truncation value via a config env instead of guessing. then it will work for any embeddings provider

PR #2522, opened 2026-07-02
hindsight-coding-agents: harness-pluggable long-term memory for coding agents

nicoloboschi:
## hindsight-coding-agents — long-term memory for coding agents (reflect + INJECT)

Adds `hindsight-integrations/hindsight-coding-agents/`: reflect-only Hindsight memory for coding
agents, **harness-pluggable**. On a task it phrases the bug's symptom to `reflect`, gets a synthesized
root-cause answer from the ingested history, and injects it into the system prompt (no tools, no
recall). Opt-in live write-back upserts the running session under a stable `document_id`.

A one-shot **backfill** configures the bank (reflect mission, observations on, `git`+`chat` retain
strategies), ingests every commit (full message + full diff), reads the harness's past sessions as raw
JSON transcripts (custom extraction → ≤2 coherent facts), and synthesizes generic codebase knowledge
pages. Every item carries a `REF-ID` tracer.

### Harness-pluggable
A "harness" (coding agent) differs in exactly two places; everything else is shared core:
- `src/core/` — HindsightClient, missions/strategies/pages, git + chat ingest, prompt injection, and a
  `RuntimeCore` (harness-agnostic reflect-once → inject → opt-in write-back state machine).
- `src/core/types.ts` — `HarnessAdapter` + `ChatReader` interfaces.
- `src/harness/` — per-agent adapters + registry. **opencode** implemented; adding an agent = one
  adapter file + one registry line.

Selection: `--harness <name>` (backfill) and `HINDSIGHT_HARNESS` (runtime, default `opencode`). An
unknown name errors with the available list.

### Notes
- Two commits: (1) the initial opencode plugin, (2) the refactor to the shared-core / adapter layout.
- Builds + type-checks clean; smoke-tested all three runtime paths (opencode hooks, unknown-harness
  error, `HINDSIGHT_DISABLED` inert).
- Based on `feat/knowledge-pages-okf` (it uses the knowledge-pages API), so this stacks on that PR;
  retarget to `main` once OKF lands.

**Draft** for review of the harness abstraction before wiring a second agent.


hindsight-integrations/hindsight-coding-agents/src/installer.ts
```diff
@@ -0,0 +1,445 @@
...
+  ctx.log?.(
+    command === "install"
+      ? `\n✅ installed. Configure the server in ~/.hindsight/coding-agent.json (apiUrl/apiToken) and start a session.`
+      : `\n✅ uninstalled.`
+  );
+  return 0;
+}
+
+/* c8 ignore start */
+const isMain = process.argv[1] && import.meta.url === new URL(`file://${process.argv[1]}`).href;
```
benfrank241: **The documented install path is a silent no-op.** `npm install -g` + `hindsight-coding-agents install claude-code` exits 0 and wires nothing — no hooks, no skill, no MCP.

npm's global bin is a symlink, so the two guards both miss:

```
argv[1]         /Users/ben/.hermes/node/bin/hindsight-coding-agents  (symlink)
import.meta.url file:///Users/ben/.../dist/installer.js              (resolved)
```

`isMain` is false (paths differ), and `argv[1].endsWith("installer.js")` is false (argv[1] is the bin *name*, not the target). The block never runs, `process.exit` is never reached, so the shell sees a clean exit and the user believes they're installed. I only got it wired by running `node dist/installer.js install claude-code` directly.

This is exactly the flow the README recommends (and it explicitly steers people away from `npx`, so the global bin is *the* path). Suggest comparing resolved paths:

```ts
import { realpathSync } from "node:fs";
const self = fileURLToPath(import.meta.url);
const invoked = process.argv[1] ? realpathSync(process.argv[1]) : "";
if (invoked === self) { ... }
```

Worth a test that runs the actual bin through a symlink — a unit test importing the module can't catch this.

hindsight-integrations/hindsight-coding-agents/src/system.live.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  }, 200_000);
+
+  it("second prompt in the same session is served from cache (no new reflect)", () => {
+    const before = diagLines().length;
+    const out = runHookBin("claude-hook.js", {
+      session_id: `live-claude-${RUN}`,
+      cwd: repo,
+      prompt: "another question, same session",
+    });
+    expect(JSON.parse(out).hookSpecificOutput.additionalContext).toContain("429");
```
benfrank241: This live test asserts behavior the code deliberately no longer has, so it fails with `SyntaxError: Unexpected end of JSON input`.

`buildHookOutput` now injects the reflect block **once** (the turn reflect ran) and otherwise only on the roster cadence — the comment at `core/hook.ts` says so explicitly ("injected ONCE ... never stacked on every turn"), and `core/hook.test.ts:71` asserts `t2.context` does *not* contain the answer. So on turn 2 there's no context and no notice, `runHook`'s `if (output.context || output.notice)` writes nothing, and `JSON.parse("")` throws here.

The caching invariant the test name describes is still worth covering — I'd keep the `diagLines().length` assert (no new reflect fired) and drop the `toContain("429")` one, or assert empty stdout instead.

hindsight-integrations/coding-agents/README.md
```diff
@@ -0,0 +1,311 @@
...
+   idempotent background **deepen engine**: it ingests any conversations not yet in the bank and the
+   next batch of recent commits **individually with their full diffs, newest first** — so full
+   decision-level precision arrives across sessions without a big-bang ingest. A per-bank lock and
+   document-id dedup make concurrent or repeated runs a no-op. Ask the agent
+   `hindsight_sync_status` to see where ingestion stands (`synced: true` = seeded memory fully
+   queryable).
+1. **Reflect once per session.** On the session's first prompt, the entry point runs Hindsight
+   `reflect` — an agentic synthesis over the bank that returns the past decision explaining the task
+   at hand, with its exact rule and literal values. The answer is cached for the session and
+   re-injected on every turn, wrapped in a visible-attribution directive so the agent surfaces a
```
benfrank241, line 32: Same drift as the live test: "The answer is cached for the session and **re-injected on every turn**" no longer matches the code, which injects the reflect block once and then only on the `pageRefreshEveryTurns` cadence. Reads like the docs and the live test were both written against the earlier policy.

benfrank241 (commented):
Pulled this branch and exercised it end to end against a local API server (`:8899`, real LLM). Short version: the integration genuinely works — but the documented install path silently does nothing.

**What passed**
- `npm test` — 213/213 unit tests
- `tsc --noEmit` and `tsup` build — clean
- Live E2E (`npm run test:live` against a real server + real LLM) — 3 of 4. The claude, codex and cursor hooks all recovered the planted decision literals from the seeded repo, so the semantic retrieval path is real.
- Wired into Claude Code for real: hooks merged, companion skill installed, MCP server registered and connected, `reflect` fires and injects a well-formed `<hindsight_memory>` block.

**Two issues**, inline below. The first one is the one that matters — I couldn't install the package via the README's own instructions.

PR #2529, opened 2026-07-02
fix(graph-maintenance): retry cooccurrence sweep on deadlock

jordigilh:
## Summary

`run_graph_maintenance_job`'s Pass 2/3 sweep (`prune_orphan_entities` + `prune_stale_cooccurrences`) has no retry protection against `DeadlockDetectedError`, unlike other concurrency-sensitive paths in this codebase (e.g. #2353's sorted `graph_maintenance_queue` inserts, or `entity_resolver`'s sorted cooccurrence upserts).

- `prune_stale_cooccurrences` deletes `entity_cooccurrences` rows via a join/`NOT EXISTS` scan with no consistent lock-acquisition order.
- `entity_resolver._flush_pending`'s concurrent cooccurrence upserts lock the same rows in sorted `(entity_id_1, entity_id_2)` order — sorted specifically to avoid deadlocking against *other* concurrent upserts, but that guarantee says nothing about the maintenance sweep's scan order.
- When the sweep and a concurrent upsert touch overlapping rows in opposite orders, Postgres detects a genuine cycle and aborts one side with `DeadlockDetectedError`.
- `acquire_with_retry` explicitly does **not** retry exceptions raised inside the transaction body (only the acquire step — see its docstring / `test_db_utils.py`), so this deadlock previously propagated straight out of `execute_task` and dropped the maintenance pass for that invocation entirely.

Evidence: in a week of self-hosted production logs, this was **39 of 41** `DeadlockDetectedError` occurrences (the remaining 1 was in `claim_graph_maintenance_batch`, not addressed here to keep this PR focused — the queue-claim path is separately protected by submit-time per-bank dedup).

## Steps to reproduce

Both a raw DB-level reproduction and an application-level (pre-fix-fails / post-fix-passes) reproduction are included as tests, so this is fully reproducible without any external setup — the test suite uses an embedded Postgres (`pg0`), no Docker or manual DB needed.

**1. Raw two-connection deadlock** (no application code, exercises Postgres directly — same technique the existing `test_unordered_concurrent_enqueue_deadlocks` in this file already uses for the `graph_maintenance_queue` case, since a deadlock can't be forced deterministically inside a single un-pausable SQL statement):

```bash
cd hindsight-api-slim
uv run pytest tests/test_graph_maintenance_deadlock.py::test_unordered_concurrent_sweep_and_upsert_deadlocks -v
```

Two transactions touch the same two `entity_cooccurrences` rows in opposite orders (one mimicking the sweep's scan order, the other mimicking retain's sorted-upsert order via a barrier-synced interleaving) → Postgres raises `DeadlockDetectedError` on one of them, reproducing the exact exception class seen in production.

**2. Confirming the bug end-to-end (this test fails on `main`, passes with this fix):**

```bash
cd hindsight-api-slim
git stash  # or: check out this branch, then revert just graph_maintenance.py to main's version
uv run pytest tests/test_graph_maintenance_deadlock.py::test_graph_maintenance_sweep_retries_on_deadlock -v
```

On `main` (pre-fix), this fails with:

```
asyncpg.exceptions.DeadlockDetectedError: deadlock detected
```

propagating straight out of `run_graph_maintenance_job` — reproducing the exact "maintenance pass silently dropped" behavior seen in production logs. With this PR's fix applied, the same test passes (the deadlock is retried once and the sweep completes with correct prune counts). I verified this locally by reverting only `graph_maintenance.py` to its pre-PR state and re-running the test suite before submitting.

## Fix

Both `prune_orphan_entities` and `prune_stale_cooccurrences` are idempotent, bank-wide sweeps — rerunning only deletes what's still stale, no partial-completion hazard. Wrap the whole Pass 2/3 transaction in `retry_with_backoff`, which is already deadlock-aware (checks `DeadlockDetectedError` by name, exponential backoff) but was previously only used internally by `acquire_with_retry`'s legacy raw-pool path, not around any actual write transaction.

## Tests

Added to `test_graph_maintenance_deadlock.py` (extends the existing PR #2353 deadlock-test file):

- `test_unordered_concurrent_sweep_and_upsert_deadlocks` — raw two-connection reproduction of the exact production deadlock: two transactions touch the same two `entity_cooccurrences` rows in opposite orders (mimicking the sweep's scan order vs. retain's sorted-upsert order) and Postgres aborts one with `DeadlockDetectedError`. No application code involved — this documents the DB-level phenomenon directly, same technique the file already uses for the queue-insert case.
- `test_graph_maintenance_sweep_retries_on_deadlock` — injects one `DeadlockDetectedError` into `prune_stale_cooccurrences` via a mock and asserts `run_graph_maintenance_job` retries and still returns correct prune counts (`stale_cooccurrences_pruned == 1`, `orphan_entities_pruned == 0`).

```
cd hindsight-api-slim && uv run pytest tests/test_graph_maintenance_deadlock.py tests/test_graph_maintenance.py tests/test_db_utils.py -v
# 19 passed
```

Also ran `ruff check`, `ruff format --check`, and `ty check` on the changed files — all clean.

## Test plan

- [x] New deadlock reproduction test passes (`test_unordered_concurrent_sweep_and_upsert_deadlocks`)
- [x] New retry-behavior test passes (`test_graph_maintenance_sweep_retries_on_deadlock`)
- [x] Existing `test_graph_maintenance.py` (14 tests) and `test_db_utils.py` still pass
- [x] `ruff check` / `ruff format --check` / `ty check` clean on changed files

Made with [Cursor](https://cursor.com)

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/graph_maintenance.py
+++ b/hindsight-api-slim/hindsight_api/engine/graph_maintenance.py
@@ -203,27 +203,43 @@ async def run_graph_maintenance_job(
 
     # --- Pass 2 & 3: entity / cooccurrence sweeps ---
     # Bank-wide single-statement deletes. Cheap when there's nothing to do.
+    #
+    # Unlike Pass 1's queue claim, these DELETEs aren't protected by any
+    # consistent lock-ordering guarantee: prune_stale_cooccurrences scans
+    # entity_cooccurrences via a join/NOT EXISTS plan, while retain's
+    # concurrent cooccurrence upserts (entity_resolver._flush_pending) lock
+    # the same rows in sorted (entity_id_1, entity_id_2) order. When a sweep
+    # and a concurrent upsert touch overlapping rows in opposite orders,
+    # Postgres detects a genuine circular wait and aborts one side with
+    # DeadlockDetectedError. Both prunes are idempotent bank-wide sweeps —
+    # rerunning only deletes what's still stale — so retrying the whole
+    # transaction on deadlock is safe.
+    from .db_utils import retry_with_backoff
     from .memory_engine import acquire_with_retry
 
-    async with acquire_with_retry(backend) as conn:
-        async with conn.transaction():
-            result.orphan_entities_pruned = await ops.prune_orphan_entities(
-                conn,
-                fq_table("entities"),
-                fq_table("unit_entities"),
-                bank_id,
-            )
-            # The orphan prune above cascades cooccurrences via FK. The
-            # explicit cooccurrence pass below catches the *stale-count*
-            # case: both entities still exist but no current unit witnesses
-            # them together.
-            result.stale_cooccurrences_pruned = await ops.prune_stale_cooccurrences(
-                conn,
-                fq_table("entity_cooccurrences"),
-                fq_table("unit_entities"),
-                fq_table("entities"),
-                bank_id,
-            )
+    async def _run_sweep() -> tuple[int, int]:
+        async with acquire_with_retry(backend) as conn:
+            async with conn.transaction():
+                orphan_pruned = await ops.prune_orphan_entities(
+                    conn,
+                    fq_table("entities"),
+                    fq_table("unit_entities"),
+                    bank_id,
+                )
+                # The orphan prune above cascades cooccurrences via FK. The
+                # explicit cooccurrence pass below catches the *stale-count*
+                # case: both entities still exist but no current unit
+                # witnesses them together.
+                stale_pruned = await ops.prune_stale_cooccurrences(
+                    conn,
+                    fq_table("entity_cooccurrences"),
+                    fq_table("unit_entities"),
+                    fq_table("entities"),
+                    bank_id,
+                )
+                return orphan_pruned, stale_pruned
+
+    result.orphan_entities_pruned, result.stale_cooccurrences_pruned = await retry_with_backoff(_run_sweep)
 
     elapsed = time.time() - job_start
     logger.info(
--- a/hindsight-api-slim/tests/test_graph_maintenance_deadlock.py
+++ b/hindsight-api-slim/tests/test_graph_maintenance_deadlock.py
@@ -1,6 +1,12 @@
 """Reproduces the concurrent-insert deadlock on ``graph_maintenance_queue``
 that PR #2353 targets, and demonstrates that a shared insertion order cures it.
 
+Also covers the ``entity_cooccurrences`` deadlock between the Pass 2/3 sweep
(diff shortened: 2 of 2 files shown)


jordigilh:
Ran the full suite locally (\`uv run pytest tests/\`, 4056 tests, 8 xdist workers, ~8 min): **3726 passed, 102 failed, 224 skipped, 3 xfailed, 1 error, 43 rerun**.

Verified none of the failures touch this PR's changed code — \`test_graph_maintenance.py\`, \`test_graph_maintenance_deadlock.py\`, \`test_db_utils.py\`, and \`test_entity_resolver.py\` are 100% green. Every failure has an independent, pre-existing cause unrelated to this diff:

- ~100 failures: tests needing a real LLM (fact extraction / causal relations / consolidation / mental models / multilingual quality suites) — all fail with the same `AuthenticationError: Incorrect API key provided: sk-test-*ummy` since I don't have a real provider key configured locally.
- 1 error: `test_migration_observation_search_vector_backfill` — an embedded-Postgres (`pg0`) instance failed to bind a fixed port (`5568`) under the 8-way parallel run, a port-contention flake unrelated to this change.

Happy to re-run against a real LLM key if that'd help review.

jordigilh:
Follow-up: re-ran with the deterministic (mock-LLM) subset for a cleaner signal:

\`\`\`
uv run pytest tests/ -m "not hs_llm_core and not hs_llm_mat"
# 3725 passed, 10 failed, 216 skipped, 1 xfailed
\`\`\`

The remaining 10 failures are pre-existing real-LLM tests missing the \`hs_llm_core\` marker (same \`sk-test-*ummy\` auth cause as before) — still no overlap with this PR's changed files.

benfrank241:
Thanks @jordigilh — nice catch. The analysis is spot-on: Pass 2/3 of `run_graph_maintenance_job` (`prune_orphan_entities` + `prune_stale_cooccurrences`) had no `DeadlockDetectedError` retry protection, unlike the sorted-order cooccurrence upserts in `entity_resolver._flush_pending` and the #2353 queue inserts, so a concurrent sweep could deadlock and drop the pass.

Heads up that @nicoloboschi's #2534 already **supersedes/completes this** — it includes your original fix commit unchanged, then extends it with (1) continuous-perf coverage that reproduces the deadlock so we don't regress, and (2) driving the ~14% dropped-pass rate under load toward zero. Since this is concurrency-sensitive maintenance-loop code and Nicolo already has your commit carried forward in an open PR, I'll defer the merge decision to him on #2534 rather than landing the two in parallel.

cc @nicoloboschi — #2534 carries this forward; closing #2529 in favor of it (or landing it first) is your call.

nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/2534 - thanks for your precious work!

PR #2530, opened 2026-07-03
fix reranker heap release after local scoring

oldnicke:
## Summary
- Run Python GC before malloc_trim after local reranker scoring
- Apply cleanup to both SentenceTransformers and FlashRank reranker paths
- Avoid importing flashrank for empty input batches

## Tests
- pytest -o addopts='' hindsight-api-slim/tests/test_local_cross_encoder.py -q
- uv run ruff format --check hindsight-api-slim/hindsight_api/engine/cross_encoder.py hindsight-api-slim/tests/test_local_cross_encoder.py
- uv run ruff check hindsight-api-slim/hindsight_api/engine/cross_encoder.py hindsight-api-slim/tests/test_local_cross_encoder.py

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -7,6 +7,7 @@ Configuration via environment variables - see hindsight_api.config for all env v
 """
 
 import asyncio
+import gc
 import logging
 import os
 import warnings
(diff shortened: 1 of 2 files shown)


benfrank241 (approved):
lgtm — the `gc.collect()` before `malloc_trim` is the right fix: without it the transient scoring objects were still referenced, so `malloc_trim` had nothing to return and the intended heap release after local reranking was a no-op. shared `_release_rerank_heap()` applies it consistently to both the SentenceTransformers and FlashRank paths, and the flashrank import is now behind the empty-batch early return. localized to the local-reranker cleanup path, 20/20 tests pass. small per-rerank gc cost only on local-reranker setups, which is the right tradeoff for the memory it frees.

approving + kicked off CI.

PR #2535, opened 2026-07-03
feat: add periodic full refresh interval for delta-mode mental models

nicha16:
## Summary

Adds a native `full_refresh_interval` trigger field for delta-mode mental models. A delta model normally performs surgical delta refreshes, but periodically forces a full re-synthesis to reset accumulated drift — all without an external cron wrapper.

```json
{
  "mode": "delta",
  "refresh_cron": "0 3 * * *",
  "full_refresh_interval": "7d"
}
```

## Key changes

- **Engine** (`memory_engine.py`): New `_parse_full_refresh_interval`, `_parse_refresh_datetime`, and `mental_model_periodic_full_refresh_due()` helpers. Refresh logic records `last_full_refreshed_at`, `refresh_mode_requested`, `refresh_mode_effective`, and `refresh_full_reason` in `reflect_response` metadata.
- **Scheduled refresh loop** (`maintenance.py`): Periodic full sweeps bypass the stale-content skip so drift can be reset even when no new memories arrived.
- **API schema** (`http.py`): `MentalModelTrigger.full_refresh_interval` field with validation (duration strings like "7d", "24h", numeric seconds).
- **Docs**: Updated trigger settings table and examples in both current and v0.8 docs.
- **Generated clients**: Python (`MentalModelTriggerInput`/`Output`) and TypeScript (`types.gen.ts`) include the new field. Go/Rust clients require Docker/Cargo regeneration.
- **OpenAPI**: Spec regenerated.

## Tests

```
✓ test_delta_mode_periodic_full_interval_forces_full_refresh
✓ test_delta_mode_periodic_full_interval_preserves_delta_until_due
✓ test_full_refresh_interval_validates_duration_values
✓ test_due_periodic_full_refresh_bypasses_staleness_skip
```

All existing delta and scheduled-refresh tests continue to pass.

Behavioral notes:
- Invalid/malformed intervals are logged as warnings and ignored — models with `full_refresh_interval` set before API validation migration work without errors.
- The field only takes effect when `mode="delta"`. Full-mode models ignore it.
- `last_full_refreshed_at` is preserved across delta refreshes and updated on full refreshes.
- No database migration needed. All state is in JSONB columns.

## Open items for reviewers

- [ ] Go and Rust generated clients need regeneration with Docker/Cargo (not available in this environment).
- [ ] Full docs site build was verified (Docusaurus, webpack artifacts).


benfrank241:
thanks @nicha16 — this is a nicely-scoped capability (periodic full re-synthesis to reset delta drift, no external cron), with the trigger field, engine helpers, generated clients, docs, and skill mirror all carried through.

routing it to @nicoloboschi to own rather than merging here — it's a `feat` that adds to the mental-model **API surface** and changes **when** models refresh (the maintenance loop now bypasses the stale-content skip for periodic full sweeps), which is a product/design call for the maintainer. worth his eyes on the `full_refresh_interval` shape and that stale-skip bypass specifically, plus a `verify-generated-files` pass given the large generated surface (openapi + py/ts clients + docs + skill mirror).

one concrete cleanup before it lands, @nicha16: the diff includes an **unrelated file** — `hindsight-docs/static/img/blog/devin-desktop-persistent-memory.png` (a blog image that has nothing to do with this feature, likely picked up from a stale base). could you drop it so the PR only carries the mental-model changes? a quick `git rm --cached hindsight-docs/static/img/blog/devin-desktop-persistent-memory.png` (or rebase onto current main) should clear it.

no objection to the direction — just want the API/refresh-behavior call to be Nicolo's, and the stray image gone.

nicoloboschi:
I don't support this change and I don't actually see the value of it. I know we have added a reset command but that should be excpetional, not regular. 

please provide a concrete use case. if there's something wrong with the delta updates, we should fix them and not build a house of cards around it

nicha16:
Hi Nicolò — thanks, that’s fair. The broader use case is users who
operate long-lived mental models over continuously evolving evidence
and choose delta mode to control refresh cost and latency.

Delta refresh is efficient because it processes new evidence
incrementally, but some changes have effects beyond the newly updated
section: retractions can invalidate older conclusions, several small
updates can collectively change a relationship, and a model’s overall
structure may need rebalancing as its evidence base grows. A periodic
full synthesis gives users a bounded, opt-in anti-entropy pass that
re-evaluates the complete evidence set. This is useful for models such
as project portfolios, people and organization maps, operating
policies, and other cumulative summaries where global coherence
matters.

The practical benefit to users would be:

- predictable control over the cost–coherence trade-off, such as
inexpensive delta refreshes during the week plus a monthly or weekly
full reconciliation;
- no external cron that temporarily mutates a model between delta and
full modes;
- atomic scheduling and state tracking inside Hindsight, including
when and why the last full synthesis occurred; and
- different policies per model, rather than requiring users to choose
permanently between full-refresh cost and indefinite incremental
operation.

Our deployment is one example: four scheduled models reduced refresh
input from roughly 100K to 20–30K tokens per run and daily consumption
from about 5.6M to 1–1.7M tokens after moving to delta mode, while
retaining a weekly full synthesis.

I agree that this should not conceal a delta correctness bug. The
intended distinction is between repairing incorrect delta behavior and
allowing users to request periodic global reconciliation as a
deliberate lifecycle policy. If Hindsight’s intended contract is that
delta mode must remain semantically equivalent to repeated full
synthesis indefinitely, then this feature is the wrong abstraction and
I’m happy to leave the PR closed. Is that the intended contract, or
would an opt-in anti-entropy policy be in scope if accompanied by a
reproducible example demonstrating where full reconciliation benefits
a long-lived model?

PR #2539, opened 2026-07-03
fix(cli): send default user agent

### LINKED ISSUE #2538: No user-agent in CLI triggers WAF rules
### Bug Description

The CLI does not send a user agent which triggers WAF rules similar to https://github.com/vectorize-io/hindsight/issues/1041

### Steps to Reproduce

Use the CLI to make a request and see that it has no user agent. Which gets the request blocked by some WAF configs.

### Expected Behavior

Should set a user-agent 

### Actual Behavior

doe not set a user-aget

### Version

0.8.4

### LLM Provider

None

koriyoshi2041:
## Summary
- add a default `User-Agent` header to the CLI reqwest client
- keep the existing bearer token default header path intact
- cover both unauthenticated and authenticated header construction

Fixes #2538.

## Tests
- `cargo test --manifest-path hindsight-cli/Cargo.toml default_headers`
- `git diff --check`
- pre-commit hooks via `git commit`

### DIFF
--- a/hindsight-cli/src/api.rs
+++ b/hindsight-cli/src/api.rs
@@ -10,6 +10,8 @@ use serde::{Deserialize, Serialize};
 use serde_json;
 use std::collections::HashMap;
 
+const USER_AGENT: &str = concat!("hindsight-cli/", env!("CARGO_PKG_VERSION"));
+
 /// Convert a progenitor client error into an anyhow error that includes the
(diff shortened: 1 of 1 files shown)


koriyoshi2041:
Closing this older branch in favor of #2564, which keeps the same fix for #2538 with a cleaner diff and passing CI.

PR #2552, opened 2026-07-04
Refresh search vector on memory curation edits

### LINKED ISSUE #2546: curation: text edit re-embeds but does not refresh `search_vector` under native FTS — BM25 arm keeps matching the pre-edit text
### Summary
`update_memory_unit` (`engine/memory_engine.py:6440`, exposed via the memory-curation PUT endpoint) is the supported way to correct a fact's text. On a text edit it correctly recomputes the **embedding** (`_reembed_memory_text`), resets `consolidated_at` (derived observations regenerate), repairs entity/graph links, and stamps `edited_at` — but the UPDATE statement (~`:6640`) sets `text`, `embedding`, temporal fields, and consolidation state **without touching `search_vector`**.

Under native FTS, `search_vector` is a plain stored `tsvector` column — not generated, no trigger — and the BM25 arm both filters and ranks against it (`search_vector @@ to_tsquery(...)`, `ts_rank_cd(search_vector, ...)`, `search/postgresql.py:231/:233`). Consolidation only rewrites **observation** search_vectors, not source-fact ones.

### Consequence
After a curation text edit, keyword recall behaves as if the edit never happened:
- the fact **keeps matching queries for its OLD (wrong/removed) content** — false positives against the very content curation was used to remove;
- it **fails to match new terms introduced by the correction** — false negatives for the corrected content;
- semantic and keyword arms permanently disagree about what the fact says (embedding = new text, tsvector = old text).

Curation exists to correct facts; the keyword arm silently un-corrects them.

### Reproduction (0.8.4, native FTS)
1. Retain a fact with distinctive long text; note `length(search_vector::text)`.
2. PUT a curation edit replacing the text with something short/different.
3. Observe: `text` and `embedding` updated, `edited_at` stamped — `search_vector` byte-identical (in our repro: text 7,996 → 733 chars; tsvector length stayed 3,788).
4. Recall with a keyword only present in the OLD text → the edited fact still matches via the BM25 arm.

### Proposed fix
Append the same conditional native-FTS clause the consolidator already uses for observation writes (`consolidator.py:282/:342/:1913` pattern) to the curation UPDATE:

```sql
, search_vector = to_tsvector('<native_language>'::regconfig, COALESCE($text, ''))
```

gated on the native FTS backend exactly as in the consolidator (VectorChord/other backends manage their own index). One clause; makes curation atomic across both retrieval arms.

### Env
`ghcr.io/vectorize-io/hindsight:0.8.4`, native tsvector FTS backend. Line numbers from the shipped image.

koriyoshi2041:
Problem
- `update_memory_unit` re-embeds curated memory text but leaves the stored `search_vector` unchanged for native PostgreSQL FTS.
- That lets BM25 keep matching removed terms and miss terms introduced by the correction.

Fix
- Rebuild `search_vector` in the curation edit UPDATE when the configured PostgreSQL search backend stores one.
- Build it from the new text/context parameters, not the pre-update columns.

Test
- `uv run --directory hindsight-api-slim ruff format hindsight_api/engine/memory_engine.py tests/test_memory_curation.py`
- `uv run --directory hindsight-api-slim ruff check hindsight_api/engine/memory_engine.py tests/test_memory_curation.py`
- `env -u ALL_PROXY -u all_proxy -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy uv run --directory hindsight-api-slim pytest tests/test_memory_curation.py::TestEdit::test_edit_changes_text_and_rederives -q`
- `git diff --check`
- repo pre-commit hook on commit: passed

Risk
- Low. The change is scoped to live curation edits and uses the existing PostgreSQL search-vector helper.
- Non-stored search backends keep the existing behavior because the helper returns no expression.

Fixes #2546

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -6639,6 +6639,17 @@ class MemoryEngine(MemoryEngineInterface):
                         mentioned_at=live["mentioned_at"],
                         entities=[r["canonical_name"] for r in ent_rows],
                     )
+                    # Keep the stored text-search vector in sync with curated
+                    # text/context edits. Use the incoming parameters here:
+                    # PostgreSQL evaluates UPDATE RHS expressions before the
+                    # sibling SET assignments take effect, so column references
+                    # would see the pre-edit text/context.
+                    from .db.ops_postgresql import pg_search_vector_expr
+
+                    sv_expr = pg_search_vector_expr(get_config(), text_col="$3", context_col="$4")
+                    search_vector_clause = (
+                        f",\n                            search_vector = {sv_expr}" if sv_expr else ""
+                    )
                     await enqueue_relink_victims(conn, bank_id, [memory_id], ops=backend.ops)
                     await conn.execute(
                         f"""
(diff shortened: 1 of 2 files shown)


benfrank241:
Matches revert path, native test verified. lgtm

PR #2559, opened 2026-07-04
Disable MiniMax M3 thinking by default

### LINKED ISSUE #2555: `reasoning_split` parameter is misnamed — does not disable thinking mode, causing 4-7x output bloat and consolidation deadlocks
## Problem

When using MiniMax models (M2.7/M3) as the LLM backend, Hindsight's `openai_compatible_llm.py` hardcodes `extra_body["reasoning_split"] = True` for `provider == "minimax"` (lines ~609 and ~1018). However, `reasoning_split` only controls **response field splitting** (whether thinking and answer are returned in separate fields) — it does **not** disable the thinking/reasoning mode itself.

This causes two cascading issues:

### 1. Thinking mode always ON → 4-7x output token bloat

Since `reasoning_split` is not a thinking on/off switch, MiniMax models continue generating thinking tokens in every LLM call:

| Metric | With thinking (bug) | Without thinking (fixed) |
|:--|:--|:--|
| output_tokens per call | 2,300–4,900 | 600–800 |
| output/input ratio | 0.64–1.24 | 0.18–0.20 |
| Time per LLM call | 80–138s | 10–20s |

This makes consolidation extremely slow — a single batch of 8 memories takes 5+ minutes.

### 2. Thinking content breaks JSON parsing → consolidation deadlock

MiniMax thinking output gets mixed into the LLM response, causing it to not match `_ConsolidationBatchResponse` Pydantic schema. This leads to:
- `validation errors for _ConsolidationBatchResponse`
- `generator didn't stop after athrow()`
- Permanent consolidation deadlock — worker slots occupied but no progress

As a workaround, we had to add `@field_validator` decorators to normalize LLM output (the "nesting fix"), which should not be necessary if thinking were properly disabled.

## Root Cause

`openai_compatible_llm.py`:
```python
# Line ~609 and ~1018
if self.provider == "minimax":
    extra_body["reasoning_split"] = True  # ← WRONG: this doesn't disable thinking
```

Additionally, `_is_reasoning_model()` returns `False` for non-GPT models (e.g., `glm-4.7-flash`), so `reasoning_effort` is never sent — but thinking remains ON by default.

## Fix

The correct way to disable thinking for MiniMax models is:

```python
extra_body["thinking"] = {"type": "disabled"}
```

Or for models that support it via `reasoning_effort`:
```python
extra_body["reasoning_effort"] = "none"
```

`reasoning_split` should be kept if response field splitting is desired, but it must not be treated as a thinking on/off switch.

## Environment

- Hindsight 0.7.1 → 0.8.2
- LLM: MiniMax M2.7 / M3 via `api.minimaxi.com`
- Symptom: consolidation worker stuck 5+ min per batch, output_tokens 4-7x normal

## Workaround

Set `HINDSIGHT_API_LLM_EXTRA_BODY='{"thinking": {"type": "disabled"}}'` in `.env` (note: single quotes required — bash `source` will execute `{` as command without them).

But this is a workaround. The hardcoded `reasoning_split = True` should be fixed to use the correct thinking-disable parameter.

## Related Issues

- #2195 (closed) — strips reasoning tags from output, but doesn't address the root cause of thinking being enabled
- #1758 (open) — mentions MiniMax thinking issues in integration plugins

koriyoshi2041:
## Summary
- set MiniMax-M3 OpenAI-compatible calls to send `extra_body.thinking={"type": "disabled"}` by default
- preserve explicit `HINDSIGHT_API_LLM_EXTRA_BODY` / provider `extra_body` overrides via `setdefault`
- cover both normal `call()` and `call_with_tools()` paths

## Notes
MiniMax's OpenAI-compatible docs state that MiniMax-M3 enables thinking by default, that `thinking: {"type": "disabled"}` skips thinking, and that `reasoning_split` only controls response field splitting. The same docs also say M2.x cannot disable thinking, so this change is intentionally scoped to MiniMax-M3 rather than claiming to fix M2.x.

Closes #2555 for the MiniMax-M3 default-thinking case.

## Tests
- `uv run pytest tests/test_minimax_thinking_defaults.py`
- `uv run ruff check hindsight_api/engine/providers/openai_compatible_llm.py tests/test_minimax_thinking_defaults.py`
- `uv run ruff format --check hindsight_api/engine/providers/openai_compatible_llm.py tests/test_minimax_thinking_defaults.py`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -644,6 +644,18 @@ class OpenAICompatibleLLM(LLMInterface):
         # use the widely-supported max_tokens
         return "max_tokens"
 
+    def _apply_minimax_defaults(self, extra_body: dict[str, Any]) -> None:
+        """Apply MiniMax request defaults while preserving explicit config."""
+        if self.provider != "minimax":
+            return
+
+        # MiniMax-M3 enables thinking by default. Hindsight's LLM calls expect
+        # direct answers unless callers opt into provider-specific behavior.
+        # MiniMax documents that M2.x accepts this parameter but cannot disable
+        # thinking, so keep the default scoped to M3.
+        if self.model.lower() == "minimax-m3":
+            extra_body.setdefault("thinking", {"type": "disabled"})
+
     async def call(
         self,
         messages: list[dict[str, str]],
@@ -731,6 +743,7 @@ class OpenAICompatibleLLM(LLMInterface):
 
         # Provider-specific parameters
         extra_body: dict[str, Any] = {**self._config_extra_body}
+        self._apply_minimax_defaults(extra_body)
         if self.provider == "groq":
             call_params["seed"] = DEFAULT_LLM_SEED
             # Add service_tier if configured
@@ -1149,6 +1162,7 @@ class OpenAICompatibleLLM(LLMInterface):
 
         # Provider-specific parameters
         extra_body: dict[str, Any] = {**self._config_extra_body}
+        self._apply_minimax_defaults(extra_body)
         if self.provider == "groq":
             call_params["seed"] = DEFAULT_LLM_SEED
         if extra_body:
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Closing this as an accidental duplicate of #2558. I opened it after re-verifying the MiniMax docs, but #2558 already covers the same default-thinking issue and is clean, so keeping one maintainer-visible PR is better.

PR #2562, opened 2026-07-04
fix(trace): skip pre-init LLM trace writes quietly instead of warning

Cybercommanders:
## Problem

Every deployment logs a spurious warning at boot:

```
WARNING - hindsight_api.engine.llm_trace - LLM trace write failed for scope=verification: PostgreSQLBackend is not initialized. Call initialize() first.
```

Root cause: `MemoryEngine.initialize()` runs `verify_llm()` inside the parallel init gather, which by design executes **before** the DB backend/pool is initialized (pg0 startup is in the same gather). The verification LLM call records a trace, and its write can therefore never succeed at boot — the writer's existing `pool is None` guard doesn't catch it because the pool getter returns a backend object that raises `RuntimeError` on acquire instead of returning `None`.

## Fix

Treat the backend's not-initialized `RuntimeError` in `_safe_write` the same way as the existing pool-is-None case: debug-level skip. All other write failures still warn. The not-initialized message shape is shared by both backends (`postgresql.py`, `oracle.py`).

Note: this matches on the exception message (`"is not initialized"`). If you'd prefer a typed exception on the backend interface instead, happy to rework — kept the change minimal here.

## Test

Regression test reproduces the exact production condition (recorder whose pool getter returns an uninitialized backend) and asserts the write skips at debug with no warning. Full `test_llm_trace.py` suite: 29 passed; ruff/format/ty clean.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/llm_trace.py
+++ b/hindsight-api-slim/hindsight_api/engine/llm_trace.py
@@ -517,6 +517,14 @@ class LLMTraceRecorder:
                     _safe_json(record.llm_info, self._max_chars) or "{}",
                     _safe_json(record.metadata, self._max_chars) or "{}",
                 )
+        except RuntimeError as e:
+            if "is not initialized" in str(e):
+                # Expected during startup: LLM calls (e.g. scope=verification) run in
+                # the init gather before the DB backend is initialized, so their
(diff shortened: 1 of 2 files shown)


Cybercommanders:
Closing at maintainer/requester direction.

PR #2563, opened 2026-07-04
Fix _strip_code_fences truncating JSON when content contains inner backticks

poog26:
The current implementation uses `content.split("```json")[1].split("```")[0]` to strip markdown code fences. This finds the **first** ` ``` ` after the opening fence — so when the extracted JSON itself contains literal triple-backtick characters, the split matches those inner backticks and truncates the JSON mid-string.

**When this bites:** Hindsight's fact-extraction LLM produces output wrapped in ` ```json ... ``` ` fences. If one of the extracted facts describes code fence behavior (e.g. "the model wraps output in ```json fences"), the JSON now contains ` ```json ` as a literal string value, and `split("```")` finds that inner occurrence. Result: truncated JSON → `json.loads` fails → retry loop (up to 4 retries per chunk).

**The fix** replaces string-splitting with line-based fence detection. Per the markdown spec, code fences must appear at the start of a line. Inner triple backticks inside JSON string values are preserved because they aren't at line boundaries.

**Before (broken):**
```python
_strip_code_fences('```json\n{"facts": [{"what": "uses ```json fences"}]}\n```')
# → '{"facts": [{"what": "uses '  (truncated!)
```

**After (fixed):**
```python
_strip_code_fences('```json\n{"facts": [{"what": "uses ```json fences"}]}\n```')
# → '{"facts": [{"what": "uses ```json fences"}]}'  (intact)
```

Edge cases: normal fences ✓, inner backticks preserved ✓, no fences no-op ✓, opening-only (truncated output) returned unchanged ✓, text around fence block handled correctly ✓.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -78,12 +78,24 @@ def _strip_code_fences(content: str) -> str:
     """
     if "```" not in content:
         return content
-    try:
-        if "```json" in content:
-            return content.split("```json")[1].split("```")[0].strip()
-        return content.split("```")[1].split("```")[0].strip()
-    except (IndexError, ValueError):
+    lines = content.split("\n")
+    # Find first line that starts a code fence (``` optionally followed by language)
+    fence_start = None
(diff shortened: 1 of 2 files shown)


benfrank241:
thanks for this! just had to add a test. merging

PR #2564, opened 2026-07-05
fix(cli): set default User-Agent header

### LINKED ISSUE #2538: No user-agent in CLI triggers WAF rules

koriyoshi2041:
## Summary
- set a default `User-Agent` for the CLI reqwest client
- keep the existing bearer auth default header behavior
- add unit coverage for CLI default headers with and without an API key

Fixes #2538.

## Tests
- `rustfmt --edition 2021 --check hindsight-cli/src/api.rs`
- `cargo test --manifest-path hindsight-cli/Cargo.toml api::tests::default_headers -- --nocapture`
- `cargo test --manifest-path hindsight-cli/Cargo.toml api::tests -- --nocapture`

Note: full `cargo fmt --manifest-path hindsight-cli/Cargo.toml --check` is currently blocked by pre-existing formatting diffs in other CLI files outside this patch.

### DIFF
--- a/hindsight-cli/src/api.rs
+++ b/hindsight-cli/src/api.rs
@@ -10,6 +10,26 @@ use serde::{Deserialize, Serialize};
 use serde_json;
 use std::collections::HashMap;
 
+const DEFAULT_CLI_USER_AGENT: &str = concat!("hindsight-cli/", env!("CARGO_PKG_VERSION"));
+
+fn default_headers(api_key: Option<&str>) -> Result<reqwest::header::HeaderMap> {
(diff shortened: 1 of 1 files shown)


ScottGuymer:
New version fixed this issue for me

PR #2565, opened 2026-07-05
fix(consolidation): default invalid dedup actions to keep

### LINKED ISSUE #2545: consolidation: dedup verdict rejects near-miss `action` values from lax-JSON models — `_DedupDecision` fails validation instead of degrading to `keep`
### Summary
The consolidation dedup step asks the LLM for a focused merge-or-keep verdict, validated as:

```python
class _DedupDecision(BaseModel):          # consolidator.py:98
    action: Literal["merge", "keep"] = "keep"   # :101
```

Lax-JSON models (observed: `gpt-oss-120b`) intermittently return non-literal `action` values, which fail pydantic `Literal` validation (`literal_error`) at the structured call (`scope="consolidation_dedup"`, :195). The consolidation operation then fails for that round: the LLM call is wasted, and the units are re-queued and re-drained later. No data loss, but recurring wasted work and delayed consolidation on exactly the cheap/fast models users run background consolidation on.

### Observed in production
12 failed operations across 5 banks (7 / 2 / 1 / 1 / 1) in a ~7h heavy-load window (2026-07-02T19:21 → 2026-07-03T02:14), `ghcr.io/vectorize-io/hindsight:0.8.4`. The exact invalid `action` values, from persisted failed ops:

`'error'` (×4), `'need_input'` (×2), `'need_observations'`, `'await'`, `'request'`, `'need_more_info'`, `'need_more_information'`, `'need clarification'`

Consequence path confirmed: the round fails → source facts remain `consolidated_at IS NULL` → re-selected and re-drained later. No data loss; cost = one wasted dedup LLM call per failure + delayed consolidation.

### Why strict validation is the wrong failure mode here
The field already has a safe default (`"keep"`). Semantically, `keep` is the conservative, non-destructive verdict — it stores the new observation separately rather than risking a wrong merge. Any unrecognized `action` therefore has an obviously-correct fallback: treat it as `keep`, log the raw value for observability, and proceed. Failing the whole consolidation round to protect against a value that would at worst have meant "don't merge" inverts the cost/benefit.

The observed values make this stronger than a generic "unknown → default" argument: they are the model trying to signal **"can't decide / need more information"** (`'need_input'`, `'need clarification'`, `'await'`, `'error'`). For an uncertain dedup verdict, *don't merge — store separately* is not merely the safe fallback, it is the semantically **correct** answer to that signal.

### Proposed fix
A `mode="before"` field_validator on `action`: normalize (strip/lower), accept the two literals, coerce anything else to `"keep"` with a warning log carrying the raw value. Merge remains possible only on an explicit `"merge"` — strictness is preserved exactly where it is destructive.

```python
@field_validator("action", mode="before")
@classmethod
def coerce_action(cls, v):
    s = str(v).strip().lower() if v is not None else ""
    if s in ("merge", "keep"):
        return s
    logger.warning(f"dedup verdict: unrecognized action {v!r} — treating as 'keep'")
    return "keep"
```

### Env
`ghcr.io/vectorize-io/hindsight:0.8.4` · consolidation provider `litellmrouter` → `openrouter/openai/gpt-oss-120b` · line numbers from the shipped 0.8.4 image.

r266-tech:
## Summary

Fixes #2545.

Consolidation dedup already defaulted a missing structured `action` to `keep`, but near-miss or malformed action values from a lax structured-output model could still fail Pydantic validation before the conservative no-merge path ran.

This adds a `mode="before"` validator on `_DedupDecision.action` that:

- preserves exact `"merge"` / `"keep"` decisions
- logs invalid action values
- defaults every other scalar or non-scalar shape to `"keep"`

That keeps malformed dedup verdicts from accidentally merging observations or requeueing the consolidation job.

## Tests

- `ruff check hindsight_api/engine/consolidation/consolidator.py tests/test_consolidation_dedup.py`
- `ruff format --check hindsight_api/engine/consolidation/consolidator.py tests/test_consolidation_dedup.py`
- `python3 -m py_compile hindsight_api/engine/consolidation/consolidator.py tests/test_consolidation_dedup.py`
- local validator smoke check for exact valid actions, unknown string actions, near-miss `Merge`, list/dict values, and missing action

I could not run `pytest tests/test_consolidation_dedup.py` in this macOS arm64 environment: `uv run` resolves `onnxruntime==1.27.0`, which has no wheel for the current Python/platform combo, and direct `PYTHONPATH=. pytest ... -o addopts=''` cannot collect without the project runtime deps such as `asyncpg` installed.

Codex review noted a low-confidence import concern about `field_validator`; the module already imports it from Pydantic and `py_compile` passed.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -102,6 +102,15 @@ class _DedupDecision(BaseModel):
     text: str = ""  # the synthesized merged observation (when action == "merge")
     reason: str = ""
 
+    @field_validator("action", mode="before")
+    @classmethod
+    def _normalize_action(cls, value: object) -> str:
+        if isinstance(value, str) and value in {"merge", "keep"}:
+            return value
+
+        logger.warning("Invalid consolidation dedup action %r; defaulting to keep", value)
+        return "keep"
+
 
 _DEDUP_PROMPT = """You reconcile long-term memory observations. A NEW observation is about to be \
 stored, and it is highly similar to an EXISTING one:
--- a/hindsight-api-slim/tests/test_consolidation_dedup.py
+++ b/hindsight-api-slim/tests/test_consolidation_dedup.py
@@ -5,6 +5,7 @@ guard the fix in CI — unlike the real-LLM integration test, which only trigger
 the path stochastically.
 """
 
+import logging
 import types
 import uuid
 from dataclasses import dataclass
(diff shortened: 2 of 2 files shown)


benfrank241:
lgtm. defaulting anything that isn't exactly merge/keep to keep is the right conservative call. Merging. Thanks!

PR #2574, opened 2026-07-06
fix(opencode): install plugin SDK at runtime

### LINKED ISSUE #2573: OpenCode plugin should install plugin SDK as a runtime dependency
## Summary

The OpenCode integration imports `@opencode-ai/plugin/tool` from its built server entrypoint, but `@opencode-ai/plugin` is currently declared only as a peer dependency. In OpenCode's plugin cache installation flow, that peer can be omitted, causing the plugin entrypoint to fail during module loading.

## Reproduction

1. Install the published plugin in a fresh OpenCode project:

```bash
opencode plugin @vectorize-io/opencode-hindsight
```

2. Inspect the OpenCode plugin cache. The package installs under:

```text
~/.cache/opencode/packages/@vectorize-io/opencode-hindsight@latest
```

3. `@vectorize-io/hindsight-client` is present because it is a dependency, but `@opencode-ai/plugin` may be absent because it is only a peer dependency.

4. Importing the built plugin entrypoint can then fail with:

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@opencode-ai/plugin' imported from .../opencode-hindsight/dist/index.js
```

## Expected behavior

Runtime imports needed by the built OpenCode plugin should be installed with the plugin package inside OpenCode's isolated plugin cache.

## Proposed fix

Move `@opencode-ai/plugin` from `peerDependencies` to `dependencies`, keeping `@vectorize-io/hindsight-client` in `dependencies`. This makes the generated plugin cache self-contained for the runtime imports used by `dist/index.js`.

## Verification

A local package with this dependency layout was installed with:

```bash
opencode plugin @vectorize-io/opencode-hindsight@file:/path/to/package.tgz
```

The resulting OpenCode cache contained both runtime dependencies, and direct import of the plugin entrypoint succeeded.

Sanderhoff-alt:
## Summary

Move `@opencode-ai/plugin` into the OpenCode integration's runtime dependencies and remove the peer-only declaration. The built plugin entrypoint imports `@opencode-ai/plugin/tool`, so OpenCode needs that package available in its isolated plugin cache when loading the integration.

`@vectorize-io/hindsight-client` remains a runtime dependency, and the lockfile is updated to reflect the new dependency classification.

Fixes #2573

## Verification

- `npm run build`
- `npm pack --json`
- Installed the generated tarball with `opencode plugin ...@file:` in a fresh temporary project
- Confirmed the OpenCode cache contained both runtime dependencies
- Confirmed direct import of the built plugin entrypoint succeeded

### DIFF
--- a/hindsight-integrations/opencode/package-lock.json
+++ b/hindsight-integrations/opencode/package-lock.json
@@ -9,10 +9,10 @@
       "version": "0.2.6",
       "license": "MIT",
       "dependencies": {
+        "@opencode-ai/plugin": "^1.3.13",
         "@vectorize-io/hindsight-client": "^0.4.19"
       },
       "devDependencies": {
-        "@opencode-ai/plugin": "^1.3.13",
         "@types/node": "^22.0.0",
         "tsup": "^8.5.1",
         "typescript": "^5.7.0",
@@ -20,9 +20,6 @@
       },
       "engines": {
         "node": ">=22"
-      },
-      "peerDependencies": {
-        "@opencode-ai/plugin": ">=1.0.0"
       }
     },
     "node_modules/@emnapi/core": {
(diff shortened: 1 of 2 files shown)


benfrank241:
clean, correct fix. thanks!

PR #2575, opened 2026-07-06
feat(integrations): add recall score floors

ishanmalik:
## Problem

Recall always returns top-k results, so users who err toward not injecting memory can still get marginal matches injected on every prompt. On a small bank, junk prompts often produce reranker scores around 0.00-0.03, while relevant prompts land around 0.4-1.0; without a floor, both can be injected.

## Solution

Add an optional `recallMinScores` config object to the Claude Code and Codex integrations. After results are collected, each configured score field is treated as a floor; numeric scores below the floor are dropped. Missing or null scores pass through so BM25-only hits and passthrough rerankers are not accidentally suppressed.

## Testing

- `python -m pytest hindsight-integrations/claude-code/tests/test_hooks.py`
- `python -m pytest hindsight-integrations/codex/tests/test_hooks.py`
- `python -m py_compile hindsight-integrations/claude-code/scripts/recall.py hindsight-integrations/codex/scripts/recall.py`


benfrank241:
thanks for this!

PR #2576, opened 2026-07-06
fix(api): accept verbatim extraction mode in bank manifests

ishanmalik:
## Problem

The engine accepts `retain_extraction_mode=verbatim` through environment/config resolution, but bank-template manifest import rejected the same mode because the validator had a re-typed mode tuple that omitted it. This surfaced during a verbatim-mode data migration.

## Solution

Use the canonical `RETAIN_EXTRACTION_MODES` tuple for bank-template validation so future extraction modes cannot drift, and update the related API schema descriptions to include `verbatim` and `chunks`.

## Testing

- `uv run --directory hindsight-api-slim pytest tests/test_bank_templates.py::TestImportValidation::test_verbatim_extraction_mode_is_valid`
- `uv run --directory hindsight-api-slim python -m py_compile hindsight_api/api/http.py`
- `scripts/generate-bank-template-schema.sh`
- `scripts/generate-docs-skill.sh`

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -52,6 +52,7 @@ from fastapi.routing import APIRoute
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 from hindsight_api import MemoryEngine
+from hindsight_api.config import RETAIN_EXTRACTION_MODES
 
 
 def _annotation_is_nullable(annotation: Any) -> bool:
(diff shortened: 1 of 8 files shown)


ishanmalik:
Follow-up: re-ran `scripts/generate-openapi.sh` with full docs dependencies installed — build validates and the regenerated artifacts match the committed ones exactly (no diff). The client-description strings updated manually in this PR are likewise identical to generator output.

benfrank241:
thanks @ishanmalik good catch.

PR #2583, opened 2026-07-06
fix(memory): return metadata from memory browse endpoints

### LINKED ISSUE #2582: Memory browse endpoints omit stored metadata
### Summary

`GET /v1/default/banks/{bank_id}/memories/list` and `GET /v1/default/banks/{bank_id}/memories/{memory_id}` do not return memory-unit `metadata`, even though metadata is stored on `memory_units` and recall responses already expose it.

### Impact

Users trying to audit or troubleshoot retained metadata through the browse/detail endpoints cannot see it. This makes it look like metadata was not persisted, while `recall` can still return the same metadata correctly.

### Expected behavior

Both memory browse endpoints should include the `metadata` object for live memory units and invalidated memory units.

### Actual behavior

The query paths select fields such as `id`, `text`, `context`, dates, tags, and curation state, but omit the `metadata` column, so the response cannot include stored metadata.

Sanderhoff-alt:
## Summary

Fixes #2582.

This updates the memory browse/detail read paths so stored memory-unit metadata is returned by `GET /memories/list` and `GET /memories/{id}`. Both live memory units and invalidated archive rows now select and serialize the `metadata` column.

## Root cause

Retain stores caller-provided metadata on each memory unit and recall already selects and returns it, but the list/get memory-unit queries omitted the `metadata` column. That made metadata invisible when users inspected memories directly, even when the database row had the expected value.

## Changes

- Select and return `metadata` from `list_memory_units()`.
- Select and return `metadata` from `get_memory_unit()` for live and invalidated memory units.
- Add a curation regression test covering list/get for both live and invalidated rows.
- Update API docs, examples, OpenAPI examples, and docs-skill references to mention metadata.

## Validation

- `cd hindsight-dev && uv run pytest tests/test_generate_openapi.py -q`
- `cd hindsight-api-slim && uv run pytest tests/test_base_path.py tests/test_memory_curation.py::TestGuardsAndListing::test_list_and_get_memory_units_include_metadata tests/test_http_api_integration.py::test_http_recall_preserves_metadata -q`
- `cd hindsight-docs && npm run build`
- `./scripts/hooks/lint.sh`

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1433,6 +1433,7 @@ class ListMemoryUnitsResponse(BaseModel):
                         "date": "2024-01-15T10:30:00Z",
                         "type": "world",
                         "entities": "Alice (PERSON), Google (ORGANIZATION)",
+                        "metadata": {"source": "slack", "channel": "engineering"},
                     }
                 ],
                 "total": 150,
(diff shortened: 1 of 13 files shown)


benfrank241:
thanks for finding this a opening a fix!

PR #2586, opened 2026-07-06
fix(parsers): coerce non-bytes buffers before the UTF-8 charset probe in markitdown

cdbartholomew:
## Problem

`MarkitdownParser._utf8_stream_info()` is type-hinted `file_data: bytes` and calls `file_data.decode("utf-8")` to check whether a text file is valid UTF-8 before handing markitdown an explicit charset hint (the hint added to work around markitdown's sample-based charset mis-detection on UTF-8 text).

But the value that flows into the parser is not always a concrete `bytes`. Callers can pass any buffer-protocol object — a `memoryview`, or a native/Rust-backed buffer (e.g. a PyO3 `Bytes`). Those support the buffer protocol but have **no `.decode` method**, so the probe raises:

```
AttributeError: '...Bytes' object has no attribute 'decode'
```

and **every text-file parse fails** (`.txt`, `.json`, `.jsonl`, `.ipynb`, `.md`, `.csv`, `.html`, ...) with `Failed to parse '<file>'`.

This wasn't caught by the existing parser tests because they pass `bytes` literals (`b"..."`, `"...".encode()`), for which `.decode()` works — the failure only appears when the upstream read path yields a non-`bytes` buffer.

## Fix

Coerce through the buffer protocol before the probe:

```python
bytes(file_data).decode("utf-8")
```

`bytes()` accepts any buffer-protocol object (the same protocol `tempfile.write(file_data)` a few lines up already relies on), and is a no-op copy for a real `bytes`. No behavior change for the `bytes` path; the non-`bytes` path now works instead of raising.

## Test

Adds `test_utf8_stream_info_accepts_non_bytes_buffer`, which passes a buffer-only object (implements the buffer protocol, has no `.decode`) and asserts the UTF-8 hint is still produced. Fails before the fix, passes after. Full `test_markitdown_parser.py` is green (6/6).

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/parsers/markitdown.py
+++ b/hindsight-api-slim/hindsight_api/engine/parsers/markitdown.py
@@ -186,7 +186,11 @@ class MarkitdownParser(FileParser):
         if Path(filename).suffix.lower() not in _TEXT_EXTENSIONS:
             return None
         try:
-            file_data.decode("utf-8")
+            # file_data may arrive as a non-``bytes`` buffer (e.g. a memoryview or
+            # a native/Rust-backed buffer object) that has no ``.decode``; coerce
+            # through the buffer protocol before the UTF-8 probe. The ``tmp.write``
+            # in the caller already relies only on the same buffer protocol.
+            bytes(file_data).decode("utf-8")
         except UnicodeDecodeError:
             return None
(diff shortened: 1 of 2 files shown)


benfrank241:
lgtm, flake on pg0 startup, re-running, then merging

benfrank241 (approved):

PR #2588, opened 2026-07-07
docs(eve): sync assistant reply default

r266-tech:
## Summary

Sync the Eve integration guide with the current auto-memory default from #2585: `includeAssistantReply` now defaults to `true`.

The package README and integration changelog already describe the default-on behavior, but the public guide still said the default was `false` / user-message-only.

## Checks

- Source-checked `hindsight-integrations/eve/src/config.ts`
- Compared the package README and Eve changelog
- Confirmed there is no `skills/hindsight-docs/references/docs-integrations/eve.md` mirror to regenerate


benfrank241:
good catch, thanks!

PR #2589, opened 2026-07-07
fix(ollama): make native num_ctx opt-in

### LINKED ISSUE #2572: Ollama native provider hardcodes num_ctx=16384, overriding Modelfile context and causing reload/eviction loop
### Bug Description

Ran into this while debugging a full-model reload/eviction loop on every conversational turn in a self-hosted deployment (Hermes Agent + Hindsight, both talking to a local Ollama instance) - tracked it down to the source, so filing with the exact location and a confirmed reproduction rather than just a symptom description. Apologies for the length below - wanted to include enough detail to reproduce and fix this without back-and-forth.

`_call_ollama_native()` in `hindsight_api/engine/providers/openai_compatible_llm.py` (~line 1340) hardcodes `num_ctx: 16384` in the `options` dict sent on every request to Ollama's native API, with no environment variable, bank config, or other override path available:

```python
options: dict[str, Any] = {
    "num_ctx": 16384,  # 16k context window for larger prompts
    "num_batch": 512,  # Optimal batch size for prompt processing
}
```

This silently overrides any custom context window configured via an Ollama Modelfile. Ollama's own documentation confirms a per-request `num_ctx` always takes precedence over a model's Modelfile default, so there is currently no way to configure this from outside the source.

**Impact:** on single-GPU hosts running a large local model, Ollama treats requests with different `num_ctx` values against the same model name as distinct runner configurations. Where VRAM can't hold two full copies of the model simultaneously (the common case for consumer/prosumer single-GPU setups), Ollama evicts and reloads the entire model every time the caller alternates between, e.g., a normal agent chat completion (configured with a different context size) and a Hindsight retain/reflect call. Each reload appears to be a full model load from disk - at least that's what's observed via `ollama ps` and GPU memory monitoring, showing VRAM drop to zero and climb back up alongside a 10-20+ second stall - added to every conversational turn that touches both paths. Have not isolated the exact reload duration in logs separate from total turn latency, so treat the specific timing as an observed correlation rather than a precisely measured breakdown.

This makes self-hosted Ollama impractical for any deployment where the primary agent's context window differs from 16384, which will be the common case for agentic workloads - most agent frameworks recommend 32k-64k+ context, well above Ollama's own default.

Confirmed via direct source inspection: grepping the entire file for `num_ctx` returns only this one occurrence, with no conditional or config lookup feeding it - it is a bare literal, not a fallback that yields to something else.

**Suggested fix:** only set `num_ctx` in the `options` dict when explicitly configured via a new env var (e.g. `HINDSIGHT_API_LLM_OLLAMA_NUM_CTX`), and omit the key entirely otherwise - letting Ollama fall through to the model's own Modelfile/server default, per Ollama's documented precedence order. This exact bug class - a client tool unconditionally injecting `num_ctx` and silently overriding a Modelfile default - has been reported and fixed in two comparable tools already: RooCodeInc/Roo-Code#7797 and devoxx/DevoxxGenieIDEAPlugin#979. Both landed on the same fix shape: stop injecting an inferred/default value, only pass it through when the user explicitly configures one.

**Workaround currently in use:** running this as a bind-mounted patched copy of `openai_compatible_llm.py` with the hardcoded line removed. Confirmed working in production - `ollama ps` now holds a stable context value across retain and chat calls with no reload cycling. Happy to share the exact patch/diff if useful for a PR.

**Environment:** Hindsight `ghcr.io/vectorize-io/hindsight:latest`, self-hosted via docker-compose, external PostgreSQL/pgvector (not embedded pg0); `HINDSIGHT_API_LLM_PROVIDER=ollama`; Ollama co-located on the same Docker host, `OLLAMA_CONTEXT_LENGTH=65536`; single NVIDIA GPU, 24GB VRAM; client agent is Hermes Agent (Nous Research) with the Hindsight memory plugin in `local_external` mode.

### Steps to Reproduce

1. Create a custom Ollama model pinning a context window other than 16384:
```bash
   cat > Modelfile << 'EOF'
   FROM 
   PARAMETER num_ctx 65536
   EOF
   ollama create mymodel-65k -f Modelfile
```
2. Configure Hindsight:
```bash
   HINDSIGHT_API_LLM_PROVIDER=ollama
   HINDSIGHT_API_LLM_MODEL=mymodel-65k
   HINDSIGHT_API_LLM_BASE_URL=http://:11434/v1
```
3. Trigger any retain call (send a message through an agent using Hindsight, or call the `/memories` endpoint directly).
4. Check what actually loaded:
```bash
   ollama ps
```

### Expected Behavior

`CONTEXT` shows `65536`, matching the Modelfile's `PARAMETER num_ctx 65536` - Hindsight's request should either omit `num_ctx` entirely (letting Ollama fall back to the model's own Modelfile/server default) or respect an explicit user-configured value.

### Actual Behavior

`CONTEXT` shows `16384` regardless of what the Modelfile specifies. Hindsight's native Ollama call unconditionally sends `num\_ctx: 16384` in the request `options`, overriding the Modelfile every time. On repeated calls that alternate between this forced 16384 and any other caller's different context setting (e.g. the same LLM also serving normal agent chat completions at a larger context), Ollama treats them as distinct runner configurations and reloads the entire model from disk on every switch - observable as the model disappearing and reappearing in `watch -n 1 ollama ps`, and as a 10-20+ second stall before each affected response.

### Version

0.8.4

### LLM Provider

Ollama

r266-tech:
## Summary

- stop sending native Ollama `num_ctx` unless it is explicitly configured
- add `HINDSIGHT_API_LLM_OLLAMA_NUM_CTX` and thread it through config, `LLMProvider`, and `MemoryEngine`
- document the migration path for operators who want to keep the old `num_ctx=16384` request behavior
- add regression coverage for env parsing, direct-constructor validation, positional timeout compatibility, and native Ollama request options

Fixes #2572.

## Tests

- `python3 -m py_compile hindsight-api-slim/hindsight_api/config.py hindsight-api-slim/hindsight_api/engine/llm_wrapper.py hindsight-api-slim/hindsight_api/engine/memory_engine.py hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py hindsight-api-slim/tests/test_config_validation.py hindsight-api-slim/tests/test_llm_wrapper.py`
- `ruff check hindsight-api-slim/hindsight_api/config.py hindsight-api-slim/hindsight_api/engine/llm_wrapper.py hindsight-api-slim/hindsight_api/engine/memory_engine.py hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py hindsight-api-slim/tests/test_config_validation.py hindsight-api-slim/tests/test_llm_wrapper.py`
- `ruff format --check hindsight-api-slim/hindsight_api/config.py hindsight-api-slim/hindsight_api/engine/llm_wrapper.py hindsight-api-slim/hindsight_api/engine/memory_engine.py hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py hindsight-api-slim/tests/test_config_validation.py hindsight-api-slim/tests/test_llm_wrapper.py`

Targeted pytest did not run in this local environment: `uv run` cannot install `onnxruntime==1.27.0` for macOS arm64, and the system Python environment is missing project deps such as `asyncpg`.


benfrank241:
really nice bug report and fix — the repro and the `ollama ps` eviction trace made this trivial to verify. removing the hardcoded `num_ctx=16384` and making it opt-in is exactly right, and the validation + regression tests are solid.

i pushed one small commit (`da12fd93`) adding `HINDSIGHT_API_LLM_OLLAMA_NUM_CTX` to `.env.example` (+ the byte-identical `hindsight-embed` copy) — new config flags need to land there per our checklist, and the env-template sync test stays green. hope you don't mind me pushing directly.

one thing to confirm before i merge: you're threading it from `config.llm_ollama_num_ctx` rather than per-member/per-bank, so it's static-global. i think that's the right call — context window is a property of the ollama host, not a per-bank preference — but it does diverge from the `member.*` pattern in that same ensemble block, so just confirm that's intentional and i'll merge.

r266-tech:
Yes, that scope is intentional.

For the native Ollama path this PR changes, the override is a process-level config value: `config.llm_ollama_num_ctx` is threaded into `LLMProvider` / `MemoryEngine` and then used by `_call_ollama_native()`. It is not a member or bank field.

The behavior I intended and covered here is:

- unset env: do not send `num_ctx`, so Ollama's Modelfile/server/model default wins
- `HINDSIGHT_API_LLM_OLLAMA_NUM_CTX=16384`: preserve the old request-forced value for operators who want that migration path

I chose that scope because the bug was caused by Hindsight forcing a request-level Ollama runner option even when the Ollama deployment already had its own context configuration. If a deployment needs different context sizes per Ollama host/model, I think the safer path is server-side Modelfile settings or separate Hindsight processes rather than one shared process env var.

I also pushed `cc207113` to keep the new config field optional for direct `HindsightConfig` construction (`kw_only` default `None`) and added a small regression for that compatibility case. Local `py_compile` passed for the touched config/test files; CI has started on the new head.

Thanks for adding the env-template entries and keeping the generated-files check green.

PR #2595, opened 2026-07-07
fix(recall): decouple temporal seed threshold

### LINKED ISSUE #2594: Temporal seed threshold is unexpectedly coupled to semantic min_scores
## Summary

Temporal retrieval currently uses `min_scores.semantic` as the embedding threshold for temporal entry-point selection when that request-level value is provided.

That makes temporal seed selection stricter when a caller only intended to tighten the semantic retrieval arm.

At the same time, graph retrieval seed selection is not affected by `min_scores.semantic`; it keeps its own fixed seed threshold (`0.3`). This makes seed behavior inconsistent across retrieval arms.

## Current behavior

In the main recall path:

- Semantic retrieval uses:

```python
min_semantic if min_semantic is not None else config.semantic_min_similarity
```

- Temporal retrieval is called with:

```python
semantic_threshold=min_semantic if min_semantic is not None else 0.1
```

- Graph retrieval finds its own semantic seeds with:

```python
threshold=0.3
limit=20
```

So if the caller passes `min_scores.semantic = 0.5`:

- semantic retrieval becomes stricter: expected
- temporal entry-point selection also becomes stricter: surprising
- graph seed selection is unchanged: inconsistent

## Why this matters

`min_scores.semantic` sounds like a filter for the semantic retrieval arm, not a global seed threshold for every retrieval strategy that happens to use embeddings internally.

Temporal retrieval has a different job from semantic retrieval. It needs to find relevant entry points inside a time window and preserve time coverage. A stable, temporal-specific seed threshold may be more appropriate than coupling it to the semantic retrieval result threshold.

Otherwise a caller trying to reduce noisy semantic matches can accidentally make temporal recall much narrower, especially for time-bound queries where semantic similarity is moderate but the time constraint is strong.

## Questions

1. Should temporal entry-point selection have its own stable threshold, independent of `min_scores.semantic`?
2. Should graph seed selection and temporal seed selection both use explicit seed-specific config values?
3. Should `min_scores.semantic` apply only to the semantic retrieval arm?
4. If the current coupling is intentional, should it be documented as applying to temporal seed selection too?

## Possible direction

Introduce explicit seed thresholds, for example:

```text
graph_seed_min_similarity = 0.3
temporal_seed_min_similarity = 0.1
```

Then keep `min_scores.semantic` scoped to semantic retrieval results, unless a caller explicitly provides a temporal-specific threshold.

Sanderhoff-alt:
## Summary

Fixes #2594.

This keeps `min_scores.semantic` scoped to the semantic retrieval arm. Temporal retrieval uses embeddings only to choose time-window entry points, so reusing the request-level semantic floor there made temporal recall unexpectedly narrower when callers only wanted to prune weak semantic matches.

The change restores the temporal entry-point default threshold of `0.1` while leaving semantic and BM25 request floors unchanged.

## Tests

`.venv/bin/python -m pytest hindsight-api-slim/tests/test_temporal_recall_selection.py`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/search/retrieval.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/retrieval.py
@@ -793,7 +793,7 @@ async def retrieve_all_fact_types_parallel(
                 tc_start,
                 tc_end,
                 budget=thinking_budget,
-                semantic_threshold=min_semantic if min_semantic is not None else 0.1,
+                semantic_threshold=0.1,
                 tags=tags,
                 tags_match=tags_match,
                 tag_groups=tag_groups,
--- a/hindsight-api-slim/tests/test_temporal_recall_selection.py
+++ b/hindsight-api-slim/tests/test_temporal_recall_selection.py
@@ -10,10 +10,12 @@ to a full scan + disk-spilling sort while dropping the most relevant in-window m
 These are pure mechanics (no LLM), so they assert directly.
 """
 
+from contextlib import asynccontextmanager
 from datetime import UTC, datetime, timedelta
 
 import pytest
 
+import hindsight_api.engine.search.retrieval as retrieval_module
 from hindsight_api.engine.search.retrieval import _select_with_temporal_coverage, retrieve_temporal_combined
 from hindsight_api.engine.task_backend import fq_table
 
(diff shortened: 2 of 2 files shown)


benfrank241:
lgtm — right call. min_scores.semantic should scope to the semantic arm, not silently tighten temporal seeds, and 0.1 was already the default so non-min_scores callers see no change. also lines it up with graph, which keeps its own fixed 0.3 seed.

one ask before merge: add a test that pins the decoupling. test_temporal_recall_selection.py doesn't currently touch min_semantic, so nothing stops this getting re-coupled later. a small one — recall with min_scores.semantic=0.5 and assert the temporal call still gets semantic_threshold=0.1 — is enough add that and i'll merge.

Sanderhoff-alt:
@benfrank241 thanks, added the regression test you suggested. It exercises recall with `min_semantic=0.5` and verifies the temporal path still gets `semantic_threshold=0.1`, so the semantic min score stays scoped to the semantic arm and won’t tighten temporal seeds again.

PR #2599, opened 2026-07-07
refactor(zed): port setup CLI to Node, drop the Python dependency

benfrank241:
## Why

The Zed integration currently needs **two runtimes**: `pip install hindsight-zed` (Python) *and* Node.js (for the `npx mcp-remote` bridge Zed uses to reach the HTTP MCP server). But the Python package is **configuration-only** — it just writes the `context_servers` entry into `settings.json` and a recall/retain rule into `AGENTS.md`. Since Node is already a hard requirement, requiring Python on top of it is pure overhead for a Zed user.

This ports the `hindsight-zed` setup CLI to a **zero-dependency Node CLI**, so the integration needs only Node.

## What

- Node CLI under `src/` + `bin/hindsight-zed.js`, shipped via `package.json` — consistent with the other TypeScript integrations (ai-sdk, chat, openclaw). `release-integration.yml` already detects `package.json` and publishes to npm.
- **Behavior-preserving** — a faithful port of the four Python modules:
  - same commands: `init` / `status` / `uninstall`, plus `--print-only`
  - same config resolution: defaults → `~/.hindsight/zed.json` → env vars → CLI flags
  - same JSONC-safe settings handling (never rewrites a commented `settings.json`; prints the snippet instead)
  - same fenced `<!-- HINDSIGHT -->` rule block in `AGENTS.md`, byte-identical `context_servers` output
- Tests ported to Node's built-in runner (`node --test`) — **21 tests**, no test deps.
- CI (`test.yml`) updated to run `npm test` on Node 22 instead of pytest.
- Removes the Python package (`hindsight_zed/`, `pyproject.toml`, `uv.lock`, Python `tests/`).

## Notes

- @DK09876 — you wrote the original integration (#2153), so flagging you as the owner. This is a straight Python→Node port with no behavior change; happy to adjust if you'd rather keep Python or take a different shape.
- Zed users now need only Node.js (which `mcp-remote` already required).


DK09876:
Hey @benfrank241 — pushed a small fixup commit (41fdaa0) to the branch to unblock the **release** path. Heads up since it's your PR.

**The problem:** `release-integration.yml` buckets any `package.json` integration as `type=typescript` and unconditionally runs `npm ci` + `npm run build` in the integration dir. That bucket was built for the compiled TS integrations (ai-sdk/chat/openclaw) — this zero-dep, no-build JS package has neither a `build` script nor a committed lockfile, so `integrations/zed/v0.2.0` would've failed at release time. It's **invisible in the green test CI** here, because `test.yml` only runs `npm test` (no `npm ci`, no build).

**The fix (3 changes):**
- add a no-op `build` script so `npm run build` succeeds
- commit `package-lock.json` so `npm ci` succeeds (npm refuses to run `ci` without one, even at zero deps); the lockfile has no `node_modules/` entries so `check-integration-lockfiles.sh` passes trivially
- drop the stray `settings.json` (a local `init` scaffold that got committed) + gitignore it

**Verified locally**, replaying the exact release steps in order: `node --test` → 21/21, `npm ci` ✓, `npm run build` ✓, `npm publish --dry-run` ships only `bin/`+`src/`+`README.md` (10 files, no settings.json).

The ported CLI logic itself looked clean — nice port. One optional follow-up worth considering separately: the release workflow could skip `npm run build` when a package has no `build` script, so the next plain-JS integration doesn't hit this same trap.

PR #2600, opened 2026-07-07
feat(config): optional cap on planner parallelism for pool connections

cdbartholomew:
## Summary

Adds `HINDSIGHT_API_DB_MAX_PARALLEL_WORKERS_PER_GATHER` — an optional, static (process-level) config. When set, every pool connection of the process runs `SET max_parallel_workers_per_gather = <n>` at init time, alongside the existing `statement_timeout` / ANN-tuning session setup. Unset (the default) leaves the server setting untouched, so existing deployments see no behavior change.

## Motivation

In multi-tenant deployments where background workers share a database with latency-sensitive foreground traffic, bulk maintenance queries (consolidation, graph upkeep) can fan out across Postgres parallel workers and occupy several cores each. Parallelism buys *latency* — which invisible background work doesn't need — at the cost of *concurrent CPU footprint*, which a shared primary very much cares about. Under load, with a multi-tenant workload where one tenant has a much larger dataset than the rest, a single tenant's background maintenance can monopolize the primary's cores and starve foreground queries.

Setting the cap to `0` on worker processes makes those queries run serially. Measured on a representative multi-million-row aggregate: serial execution cost ~29% more wall-clock but used **67% fewer concurrent cores** — and less total CPU, since parallel coordination isn't free.

## Design notes

- `0` is a meaningful value (disable parallelism), so the env parse distinguishes *unset* (None, no opinion) from *0* via a new `_parse_optional_non_negative_int` helper; negative/non-integer values fail fast at startup.
- Static field — deliberately not in `_CONFIGURABLE_FIELDS` (infrastructure tuning, not per-bank behavior).
- `.env.example` + bundled embed template + configuration docs updated per the config-flag checklist.

## Test plan

- 10 new unit tests (`test_db_parallel_workers_config.py`) pinning the parse semantics, incl. the 0-vs-unset distinction and fail-fast on invalid values
- `./scripts/hooks/lint.sh` clean; `ty check` clean
- Behavior validated against a real PostgreSQL: connections initialized by a worker process (through PgBouncer) observe the configured value, and heavy aggregates plan serially (`EXPLAIN ANALYZE` shows no Gather node)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -95,6 +95,7 @@ HINDSIGHT_API_LOG_LEVEL=info
 # HINDSIGHT_API_READ_DATABASE_URL=  # Optional read-replica URL. When set, recall queries (semantic, BM25, graph, temporal) flow through a separate pool against this URL, offloading the primary. Typically points to a read-only endpoint (CNPG's <cluster>-ro service or Aurora reader endpoint).
 # HINDSIGHT_API_MIGRATION_DATABASE_URL=  # Direct PostgreSQL URL for migrations (bypasses PgBouncer). Falls back to DATABASE_URL.
 # HINDSIGHT_API_DATABASE_SCHEMA=public  # PostgreSQL schema name (default: public)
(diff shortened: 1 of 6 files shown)


benfrank241:
lgtm other than failing test https://github.com/vectorize-io/hindsight/actions/runs/28878324529/job/85664627860?pr=2600

nicoloboschi (approved):
LGTM

PR #2616, opened 2026-07-08
fix(claude-code): dedupe additional bank recall calls

### LINKED ISSUE #2604: claude-code plugin: recallAdditionalBanks re-recalls the primary bank when it appears in the list
## Environment
claude-code plugin (hindsight-integrations/claude-code), observed at 0.7.2 and confirmed unchanged at repo tip.

## Problem
`recallAdditionalBanks` is a single global list, and the recall loop does not skip an entry equal to the session's resolved primary bank. Bidirectional cross-bank visibility between two banks (A-primary sessions see B, B-primary sessions see A) requires listing BOTH banks — so every session re-recalls its own primary redundantly: one wasted recall call (~0.4–0.65s) per user prompt, plus duplicate context.

## Repro
`settings/claude-code.json`: `"recallAdditionalBanks": ["econ-research", "dev-env"]` with primary resolving to either bank (static bankId or directoryBankMap) → the recall hook queries the primary twice per prompt.

## Suggestion
One-line guard in the additional-banks loop: `if extra_bank_id == resolved_primary: continue`. (Per-bank filters at tip make the loop richer, but the primary-dedup is still absent.)

r266-tech:
## Summary
- dedupe Claude Code additional-bank recall calls by bank plus effective filter state
- preserve distinct same-bank recalls when filters differ
- keep malformed or failed recall attempts retryable while treating missing `results` as an empty compatible response

Fixes #2604.

## Tests
- `python3 -m py_compile hindsight-integrations/claude-code/scripts/recall.py hindsight-integrations/claude-code/tests/test_hooks.py`
- `python3 -m pytest hindsight-integrations/claude-code/tests/test_hooks.py -q` (46 passed)
- `git diff --check`
- Codex adversarial review: approve (`/tmp/codex-review-hindsight-2604-v12.final.txt`)

### DIFF
--- a/hindsight-integrations/claude-code/scripts/recall.py
+++ b/hindsight-integrations/claude-code/scripts/recall.py
@@ -41,6 +41,50 @@ from lib.state import write_state
 LAST_RECALL_STATE = "last_recall.json"
 
 
+def recall_filter_key(tags, tag_groups, tags_match):
+    def canonical_tags(value):
+        if not value:
+            return None
+        if not isinstance(value, list):
+            return value
+        unique = {json.dumps(item, sort_keys=True): item for item in value}
+        return [unique[key] for key in sorted(unique)]
+
+    def canonical_tag_groups(value):
+        if not value:
+            return None
+        if isinstance(value, dict):
+            return {
+                key: canonical_tags(value[key]) if key == "tags" else canonical_tag_groups(value[key])
+                for key in sorted(value)
+            }
+        if isinstance(value, list):
+            return [canonical_tag_groups(item) for item in value]
+        return value
+
+    canonical_filter_tags = canonical_tags(tags)
+    canonical_filter_groups = canonical_tag_groups(tag_groups)
+    return json.dumps(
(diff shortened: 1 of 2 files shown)


benfrank241:
thanks for this — closing as superseded. #2625 landed the recallAdditionalBanks dedup (skip the primary + repeated bank ids) and it's already shipped in claude-code v0.7.4, so the core of this is covered.

one thing your version does that #2625 doesn't: you key on (bank + tags + tag_groups + tags_match) rather than bank id alone, so the *same* bank listed twice with *different* filters would still fire both calls. that's a real (if narrow) distinction — but recallAdditionalBanks doesn't attach per-entry filters that way today (per-bank filters come from recallAdditionalBankFilters keyed by bank id, one per bank), so in practice the two collapse to the same behavior. if you think there's a config shape where the finer key matters, happy to look at a focused follow-up.

appreciate the malformed-results guard too — if you want that hardening on its own, that'd be a clean standalone PR.

PR #2618, opened 2026-07-08
fix(trace): skip LLM trace writes during daemon shutdown/pre-init races

yingliang-zhang:
## Problem

`LLMTraceRecorder._safe_write` and `_attach_memory_ids` produce spurious WARNING logs during two race windows in the daemon lifecycle:

**1. Pre-init race:** `MemoryEngine.initialize()` runs `verify_llm()` inside the parallel init gather, which executes **before** the DB backend pool is ready. The `pool_getter` returns a backend object whose `_pool` is still `None`, so `acquire()` raises `RuntimeError("PostgreSQLBackend is not initialized")`.

**2. Shutdown race:** `MemoryEngine.close()` calls `backend.shutdown()` (sets `_pool=None`) before setting `self._backend=None`. Fire-and-forget trace tasks that are still in-flight see a non-None backend object, but its internal asyncpg pool is already closed — hitting either `RuntimeError("not initialized")` or `InterfaceError("pool is closing")`.

Both are expected lifecycle states, not actionable errors. Yet they produce repeated WARNING logs on every daemon restart:

```
WARNING - hindsight_api.engine.llm_trace - LLM trace write failed for scope=verification: PostgreSQLBackend is not initialized. Call initialize() first.
WARNING - hindsight_api.engine.llm_trace - LLM trace write failed for scope=verification: pool is closing
```

## Fix

Three changes in `llm_trace.py`:

1. **Pre-acquire guard** in `_safe_write`: check `getattr(pool, "_pool", "missing") is None` before attempting to acquire — catches the post-shutdown case where the backend object exists but its pool is gone.

2. **Error downgrade** in `_safe_write` except block: `"pool is closing"` and `"not initialized"` errors are downgraded to DEBUG (expected shutdown races). All other errors still WARN.

3. **Same downgrade** in `_attach_memory_ids` except block — covers the same race at the second write site.

## Relation to #2562

PR #2562 covered only the pre-init `RuntimeError("is not initialized")` path. This PR additionally covers:
- The shutdown `"pool is closing"` race (asyncpg `InterfaceError` — different exception type, not caught by #2562's `except RuntimeError`)
- The `_attach_memory_ids` write path (not touched by #2562 at all)

## Test

5 regression tests in `test_llm_trace.py`:

| Test | Scenario | Asserts |
|------|----------|---------|
| `test_safe_write_pool_none_skips_quietly` | `pool_getter` returns `None` | No WARNING |
| `test_safe_write_backend_pool_none_skips_quietly` | Backend exists, `_pool=None` | No WARNING, DEBUG msg present |
| `test_safe_write_pool_closing_downgrades_to_debug` | `acquire()` raises `"pool is closing"` | No WARNING, "shutdown race" in DEBUG |
| `test_safe_write_unexpected_error_still_warns` | `acquire()` raises unrelated error | WARNING preserved |
| `test_attach_memory_ids_pool_none_skips_quietly` | `_attach_memory_ids` with `_pool=None` | No WARNING |

Full `test_llm_trace.py` suite: **32 passed, 1 deselected** (real-LLM test needing API key).

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/llm_trace.py
+++ b/hindsight-api-slim/hindsight_api/engine/llm_trace.py
@@ -477,6 +477,12 @@ class LLMTraceRecorder:
         if pool is None:
             logger.debug("LLM trace skipped: pool not available")
             return
+        # Guard against the backend existing but its internal asyncpg pool being
+        # None (during/after shutdown — close() calls backend.shutdown() which
+        # sets _pool=None before the backend object itself is dereferenced).
+        if getattr(pool, "_pool", "missing") is None:
+            logger.debug("LLM trace skipped: backend pool not initialized (shutdown in progress?)")
+            return
         try:
             schema = self._schema_getter()
             table = f"{schema}.llm_requests"
@@ -518,7 +524,13 @@ class LLMTraceRecorder:
                     _safe_json(record.metadata, self._max_chars) or "{}",
                 )
         except Exception as e:
-            logger.warning(f"LLM trace write failed for scope={record.scope}: {e}")
+            err_str = str(e)
+            # Downgrade known shutdown-related errors to DEBUG — these are
+            # expected races during daemon close()/restart and are not actionable.
+            if "pool is closing" in err_str or "not initialized" in err_str:
+                logger.debug("LLM trace write skipped (shutdown race) for scope=%s: %s", record.scope, e)
+            else:
+                logger.warning(f"LLM trace write failed for scope={record.scope}: {e}")
 
     async def _flush_pending(self, trace_id: str) -> None:
         """Await this trace's in-flight writes so its rows exist before an UPDATE."""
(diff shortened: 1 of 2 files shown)


benfrank241 (approved):
lgtm — reproduced the bug on main: with a backend whose `_pool` is None (post-shutdown / pre-init), `_safe_write` emits exactly `WARNING ... LLM trace write failed for scope=verification: PostgreSQLBackend is not initialized`. running these 5 tests against pristine main fails the three shutdown-race cases (warning emitted); on this branch all 5 pass. clean fails-before/passes-after.

the fix is well-targeted — verified the `getattr(pool, "_pool", ...)` guard matches the real `PostgreSQLBackend` (`db/postgresql.py` sets `self._pool=None` on shutdown and raises that exact message), and the string-downgrade covers both the `"not initialized"` RuntimeError and the asyncpg `"pool is closing"` InterfaceError that #2562's `except RuntimeError` missed — at both write sites (`_safe_write` and `_attach_memory_ids`). unexpected errors still WARN (explicitly tested). 32 pass locally (the lone error is the api-key real-llm test).

approving + kicked off CI.

PR #2619, opened 2026-07-08
fix: lower default reranker_max_candidates from 300 to 50

### LINKED ISSUE #2554: [Bug] reranker_max_candidates default=300 causes 30s+ recall latency — should default to 50
## Bug Description

Hindsight 0.8.2 defaults `HINDSIGHT_API_RERANKER_MAX_CANDIDATES` to 300. This means every recall request reranks up to 300 candidates through the cross-encoder, which takes 30+ seconds on a local jina-reranker-v3.

This is not a performance optimization issue — it's a default misconfiguration. 300 candidates is excessively high for cross-encoder reranking. Most production use cases need 20-50 candidates for accurate top-K results.

## Impact

In a production deployment with local jina-reranker-v3 (Q6_K GGUF, GPU):

| Setting | Recall Latency | Quality |
|---------|---------------|---------|
| `reranker_max_candidates=300` (default) | **30-45 seconds** | Excellent |
| `reranker_max_candidates=50` (our fix) | **13-19 seconds** | No noticeable degradation |
| `reranker_max_candidates=20` | **8-12 seconds** | Slight degradation |

Users experiencing 30s+ recall latency may assume Hindsight is broken or unusably slow, without realizing a single environment variable fixes it.

Issue #1081 reported Reranker HTTP ReadTimeout under contention — the root cause may partially be this excessive default.

## Workaround Applied

Set `HINDSIGHT_API_RERANKER_MAX_CANDIDATES=50` in hindsight config. This immediately reduced recall latency from 30-45s to 13-19s with no measurable quality degradation.

Also set `HINDSIGHT_API_RECALL_MAX_CANDIDATES_PER_SOURCE=20` to further reduce the candidate pool.

## Expected Behavior

The default should be 50 (or configurable with a documented recommended range of 20-100). 300 is excessive for most use cases and creates a poor first-run experience.

Alternatively, add a startup warning when `reranker_max_candidates > 100`:
```
⚠️ reranker_max_candidates=300 may cause high recall latency (>20s). Consider setting HINDSIGHT_API_RERANKER_MAX_CANDIDATES=50.
```

## Environment

- Hindsight: 0.8.2
- Reranker: jina-reranker-v3 Q6_K GGUF via llama-server (local, GPU)
- Recall latency before fix: 30-45s
- Recall latency after fix: 13-19s

handnewb:
## Summary

`DEFAULT_RERANKER_MAX_CANDIDATES` was 300, causing 30-45s recall latency with local reranker (e.g. jina-reranker-v3). Changed to 50, reducing latency to 13-19s with no measurable quality degradation.

## Impact

| Setting | Recall Latency | Quality |
|---------|---------------|---------|
| 300 (old default) | 30-45s | Excellent |
| **50 (new default)** | **13-19s** | No degradation |
| 20 | 8-12s | Slight degradation |

Users can still override via `HINDSIGHT_API_RERANKER_MAX_CANDIDATES`.

## Change

1 line in `config.py`:
```python
-DEFAULT_RERANKER_MAX_CANDIDATES = 300
+DEFAULT_RERANKER_MAX_CANDIDATES = 50
```

Closes #2554


benfrank241:
thanks @handnewb — the latency point is real and worth addressing, but i want to hold this one for @nicoloboschi to call rather than merge it directly. it's a change to a global recall-quality default, not a bug fix, so it's a product tradeoff rather than a correctness question.

two things worth surfacing for that decision:

- **the quality claim needs evidence.** `reranker_max_candidates` is exactly how many fused (RRF) candidates the cross-encoder gets to re-score, so 300→50 means any relevant memory that fusion ranks 51–300 can no longer be rescued by the reranker. "no measurable degradation" is very plausible, but on an accuracy-focused product this is the kind of default we'd want backed by a `longmemeval`/`locomo` run at 300 vs 50 rather than a subjective read. happy to run that.

- **it's tuned for one profile.** the 30–45s is a local jina-reranker-v3; TEI/GPU and hosted rerankers (cohere/voyage) chew through 300 fast and would just lose headroom. so a global flip optimizes the local-cpu case at some cost to everyone else — which is partly why a middle value (say 100) or the startup-warning-over-100 alternative from #2554 might land better.

also, if we do change it, the documented default (`300` in `configuration.md` + the generated skill mirror) needs the same bump or the docs drift.

net: not blocking the direction — 300 is arguably high — just don't want to move a core-quality default on anecdote. nicolo, your call on the value (50 vs 100 vs warn-only), and i can get benchmark numbers first if useful.

nicoloboschi:
just change your server value. 300 is for production environments and it's fine as default, thanks.

PR #2620, opened 2026-07-08
fix: handle FK violation in observation_history during parallel consolidation

### LINKED ISSUE #2506: Consolidation: observation_history FK violation (observation_history_observation_id_fkey) under parallel/batched observation updates
## Summary

With default (parallel/batched) consolidation settings, observation reconsolidation can intermittently fail with:

```
insert or update on table "observation_history" violates foreign key constraint "observation_history_observation_id_fkey"
```

The consolidation task fails and is retried; subsequent runs usually succeed (race-dependent).

## Evidence

Self-hosted 0.8.3-slim (ARM64, Postgres 16 + pgvector + pgroonga), seen repeatedly while reconsolidating a bank's observation corpus. Latest occurrence 2026-07-01 on a bank running global default consolidation settings; banks pinned to sequential consolidation via per-bank config:

```json
{"consolidation_llm_batch_size": 1, "consolidation_llm_parallelism": 1}
```

have never reproduced it. That workaround (sequential consolidation) makes the failure disappear, which points at a race between parallel observation update batches: one batch deletes/replaces an observation row while another inserts an `observation_history` row referencing the old `observation_id` (history moved to dedicated tables in #2007).

## Notes

- 0.8.4 contains a related-but-different concurrency fix (#2353 — graph-maintenance concurrent-insert deadlock, sorted unit_ids); it does not obviously cover the observation_history FK path.
- Not found in existing issues (searched observation_history / FK).

## Suggested direction

Make observation replacement + history insertion transactional per observation (or take a consistent lock ordering across parallel batches), so a parallel batch can't reference a just-deleted observation id. Alternatively validate/skip history rows whose observation vanished mid-batch.

## Environment

- Hindsight 0.8.3-slim → 0.8.4-slim, self-hosted Docker, ARM64
- Postgres 16, pgvector, pgroonga text search
- Workaround in production: per-bank `consolidation_llm_batch_size=1`, `consolidation_llm_parallelism=1`

### LINKED ISSUE #2597: Consolidation fails with foreign key violation on observation_history.observation_id
While diagnosing a stuck consolidation queue (via `GET /v1/default/banks/{bank_id}/operations?status=failed`), I found one `consolidation` task failing with a database-level error rather than the usual timeout/provider errors:

```
insert or update on table "observation_history" violates foreign key constraint "observation_history_observation_id_fkey"
DETAIL:  Key (observation_id)=(b782b079-097e-4ad2-8d43-b125cda5b4d3) is not present in table "observations".
```

- **API version:** 0.8.4
- **Operation type:** `consolidation`
- **Occurred:** 2026-07-06T19:45:48 UTC
- **Bank:** single-user bank, 4957 nodes / 85,184 links at the time

This happened shortly before a burst of retain requests (~20:20-20:27 UTC the same evening) that overloaded our LLM provider and triggered circuit-breaker errors on other operations -- possibly related, if consolidation and retain share any write path under load.

It looks like consolidation attempted to write a history row referencing an `observation_id` that either hadn't been committed yet or had already been removed by the time the history insert ran -- possibly a race between observation creation and the history write itself under concurrent load (we did not call any observation-deletion endpoint around that time, so it's likely internal to consolidation's own write ordering rather than an external actor).

I don't have a minimal repro yet (this surfaced organically during normal usage, not a synthetic test), but wanted to flag it since it suggests a possible ordering/transaction issue between observation writes and `observation_history` inserts under concurrent/high load. Happy to share more operation logs if useful.

handnewb:
## Summary

Under parallel/batched consolidation, `_append_observation_history` can fail with:

```
insert or update on table "observation_history" violates foreign key constraint "observation_history_observation_id_fkey"
```

One batch deletes/replaces an observation row while another writes its history — a race condition. Instead of failing the entire consolidation task, catch the `ForeignKeyViolationError` and skip the history entry.

## Change

- Add `import asyncpg` to consolidation module
- Wrap the `INSERT` in `_append_observation_history` with `try/except asyncpg.exceptions.ForeignKeyViolationError`
- Log a warning when the FK violation occurs (observation was removed before history)

## Impact

- Consolidation tasks no longer fail due to this race condition
- Observation history entries are best-effort (safe to skip — the current observation state is the source of truth)
- No schema changes required

## Reproduction

Triggered by parallel consolidation (`consolidation_llm_parallelism > 1`). Sequential consolidation (`consolidation_llm_parallelism = 1`) never hits this, confirming the race condition.

Closes #2597
Closes #2506

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -28,6 +28,7 @@ from fnmatch import fnmatchcase
 from itertools import combinations
 from typing import TYPE_CHECKING, Any, Literal
 
+import asyncpg
 from pydantic import BaseModel, field_validator
 
 from ...config import get_config
@@ -1775,15 +1776,22 @@ async def _append_observation_history(
     history from growing without bound.
     """
     obs_uuid = uuid.UUID(observation_id)
-    await conn.execute(
-        f"""
+    try:
+        await conn.execute(
+            f"""
         INSERT INTO {fq_table("observation_history")} (observation_id, bank_id, content, changed_at)
         VALUES ($1, $2, $3::jsonb, now())
         """,
-        obs_uuid,
-        bank_id,
-        json.dumps(asdict(snapshot)),
-    )
+            obs_uuid,
+            bank_id,
+            json.dumps(asdict(snapshot)),
+        )
+    except asyncpg.exceptions.ForeignKeyViolationError:
+        logger.warning(
+            f"FK violation writing observation_history for {observation_id}: "
+            "observation was removed before history could be written (race with parallel consolidation). Skipping."
+        )
+        return
     if max_entries and max_entries > 0:
         await conn.execute(
             f"""
--- a/hindsight-api-slim/tests/test_observation_history_fk_race.py
+++ b/hindsight-api-slim/tests/test_observation_history_fk_race.py
@@ -1,0 +1,48 @@
+"""Regression: observation_history write for a since-deleted observation is skipped.
+
+Under parallel (or same-batch delete-then-update) consolidation, one write path
+can remove an observation from ``memory_units`` before another writes its
+``observation_history`` snapshot. The INSERT then trips
+``observation_history_observation_id_fkey``. Because consolidation runs in
+autocommit (no enclosing transaction), catching the FK violation and skipping
+the best-effort history row is safe — the connection stays usable and the
+current observation state remains the source of truth.
+
+Regression for #2597 / #2506: before the fix this raised
+``asyncpg.ForeignKeyViolationError`` and failed the whole consolidation task.
+"""
+
+import uuid
+
+import pytest
+
+from hindsight_api.engine.consolidation.consolidator import (
+    _append_observation_history,
+    _ObservationHistorySnapshot,
+)
+from hindsight_api.engine.db_utils import acquire_with_retry
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


yingliang-zhang:
Thanks for this fix — it matches exactly what I was about to PR myself.

I hit the same `ForeignKeyViolationError` on `observation_history_observation_id_fkey` and traced the root cause. One additional finding worth noting:

**This race is not limited to parallel consolidation (`consolidation_llm_parallelism > 1`).** It can also occur within a single batch in sequential mode:

Looking at `_process_memory_batch` (around L1566-1625 in `consolidator.py`), deletes execute first (L1588-1598), then updates (L1600-1625). If the LLM returns both a `delete` and an `update` action for the **same observation_id** within one batch:

1. The delete succeeds — observation row is removed from `memory_units`
2. The update proceeds to `_execute_update_action` → `_append_observation_history` (L1895)
3. The `INSERT INTO observation_history` fails because the observation no longer exists in `memory_units`

The security check at L1592 (`observation must be present in the unioned recall`) only guards the delete path, and the check at L1605 (`observation must be in per_fact_obs_ids`) only guards the update path — neither cross-references the other action list within the same batch.

So the `try/except ForeignKeyViolationError` approach in this PR is the correct and sufficient fix — it handles both the parallel race and the same-batch delete-then-update sequence. The observation_history entry is best-effort by nature (the current observation state is the source of truth), so skipping it on FK violation is safe.

Reproduced on v0.8.4 with sequential consolidation (`consolidation_llm_parallelism=1`). After applying this patch locally, the 2 failed consolidation tasks stopped recurring.

PR #2623, opened 2026-07-08
fix: coerce non-string metadata values to strings in MemoryFact.parse_metadata

### LINKED ISSUE #2622: Bug: Consolidation blocked when metadata values are non-string types (e.g., integer original_id)
### Version
hindsight-api 0.8.4

### Summary
The consolidation worker fails with a Pydantic validation error when `metadata.original_id` is stored as a JSON integer (number) instead of a string. This blocks ALL consolidation until the affected `memory_units` are fixed.

### Symptoms
```
Failed to search memories (ValidationError): 1 validation error for MemoryFact
metadata.original_id
  Input should be a valid string [type=string_type]
```
- Consolidation operations fail repeatedly (71 consecutive failures observed)
- Worker enters retry backoff; retry_count reaches 29 before giving up
- `pending_consolidation` accumulates indefinitely (6,321 unconsolidated MUs)
- No data loss — raw recall still works, but consolidated observations are never created

### Root Cause
In `engine/response_models.py:253`, `MemoryFact.metadata` is typed as `dict[str, str]`. The `parse_metadata` validator handles `str → dict` coercion (JSONB from asyncpg) but does NOT handle `int → str` coercion for individual dict values.

### Reproduction
1. Have a Hindsight bank with observation-type memory_units containing numeric `original_id` in metadata JSONB
2. Run consolidation
3. Observe the ValidationError

Query to check affected rows:
```sql
SELECT COUNT(*) FROM memory_units
WHERE metadata->'original_id' IS NOT NULL
  AND jsonb_typeof(metadata->'original_id') != 'string';
```

### Impact
- 394 out of 10,741 memory_units affected (3.7%) in our production bank
- Consolidation completely blocked until all affected rows are fixed
- Affected MUs are mostly `observation` type with empty context (internal bookmarks)

### Workaround
```sql
UPDATE memory_units 
SET metadata = jsonb_set(metadata, '{original_id}', 
    to_jsonb((metadata->>'original_id')::text))
WHERE metadata->'original_id' IS NOT NULL 
  AND jsonb_typeof(metadata->'original_id') = 'number';
```

### Proposed Fix
Add value coercion in `parse_metadata`:
```python
if isinstance(v, dict):
    return {str(k): str(val) for k, val in v.items()}
```

See linked PR for implementation.

handnewb:
## Summary
Fix consolidation blocker caused by non-string metadata values in MemoryFact.

## Problem
When `memory_units` contain metadata with non-string values (e.g., `{"original_id": 348}` from observation bookmarks), the consolidation worker fails with:
```
ValidationError: metadata.original_id — Input should be a valid string [type=string_type]
```

This happens because `MemoryFact.metadata` is typed as `dict[str, str]` but the `parse_metadata` validator doesn't coerce non-string dict values.

## Root Cause
In `engine/response_models.py:253`, `metadata: dict[str, str]`. The `@field_validator("metadata")` method only handles `str → dict` coercion (for asyncpg JSONB serialization), not `int/other → str` coercion for individual values.

## Fix
Added value coercion in `parse_metadata`: after JSON parsing (or on raw dict input), iterate all key-value pairs and convert both to strings with `{str(k): str(val) for k, val in v.items()}`. This safely handles integer, float, bool, and None values.

## Impact
- 394 out of 10,741 memory_units affected (3.7%) in a production bank
- Consolidation was completely blocked with 6,321 pending items
- 71 consecutive consolidation failures before workaround applied

## Testing
- Existing tests pass (no behavior change for already-string metadata)
- Verified on bank with 10,741 MUs — consolidation resumed immediately after fix + workaround cleanup
- Backward compatible: string values pass through unchanged

Closes #2622

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/response_models.py
+++ b/hindsight-api-slim/hindsight_api/engine/response_models.py
@@ -255,13 +255,20 @@ class MemoryFact(BaseModel):
     @field_validator("metadata", mode="before")
     @classmethod
     def parse_metadata(cls, v: Any) -> dict[str, str] | None:
-        """Parse metadata from JSON string if needed (asyncpg may return JSONB as str)."""
+        """Parse metadata from JSON string if needed (asyncpg may return JSONB as str).
+
+        Also coerces non-string dict values (e.g., integer IDs stored in JSONB)
+        to strings, preventing ValidationError when consolidation encounters
+        metadata like {"original_id": 348} instead of {"original_id": "348"}.
+        """
         if v is None:
             return None
         if isinstance(v, str):
             import json
 
-            return json.loads(v)
+            v = json.loads(v)
+        if isinstance(v, dict):
+            return {str(k): str(val) for k, val in v.items()}
         return v
 
     chunk_id: str | None = Field(
--- a/hindsight-api-slim/tests/test_response_models.py
+++ b/hindsight-api-slim/tests/test_response_models.py
@@ -1,0 +1,31 @@
+"""Tests for MemoryFact response-model validation."""
+
+from hindsight_api.engine.response_models import MemoryFact
+
+_BASE = {"id": "u1", "text": "hi", "fact_type": "observation"}
+
+
+class TestMemoryFactMetadataCoercion:
+    """metadata values may arrive as non-strings from JSONB; they must coerce to str.
+
+    Regression for #2622: an integer ``original_id`` stored in ``metadata`` raised
+    ``ValidationError`` (``string_type``) and blocked consolidation, because
+    ``metadata`` is typed ``dict[str, str]`` and the raw dict passed straight
+    through to Pydantic.
+    """
+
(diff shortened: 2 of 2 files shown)


benfrank241:
verified this fixes #2622 — reproduced the `metadata.original_id` `string_type` error on main and confirmed the coercion resolves it (int → "12345", also handles the asyncpg JSONB-string path, strings unchanged). pushed a regression test (`tests/test_response_models.py`, 4 cases) since there wasn't one.

one follow-up (not for this PR): the same `dict[str, str]` typing is on the HTTP response models too — `RecallResult.metadata` and the memory-unit list/detail models in `api/http.py` (lines ~376 / ~612). those don't run through `MemoryFact.parse_metadata`, so a numeric metadata value could still trip a `ValidationError` on the direct HTTP recall/browse path. worth a separate pass to apply the same coercion there. this PR correctly unblocks the consolidation path.

PR #2624, opened 2026-07-08
fix(claude-code): skip duplicate recall banks

### LINKED ISSUE #2604: claude-code plugin: recallAdditionalBanks re-recalls the primary bank when it appears in the list

koriyoshi2041:
## Summary
- skip recallAdditionalBanks entries that repeat the primary Claude Code memory bank
- de-duplicate repeated additional bank IDs while preserving order
- add a regression test for primary-bank and repeated extra-bank entries

Closes #2604

## Test
- uv run pytest hindsight-integrations/claude-code/tests/test_hooks.py::TestRecallHook::test_additional_banks_skip_primary_and_duplicates -q
- uv run pytest hindsight-integrations/claude-code/tests/test_hooks.py -q
- uv run ruff format --check hindsight-integrations/claude-code/scripts/recall.py hindsight-integrations/claude-code/tests/test_hooks.py
- uv run ruff check hindsight-integrations/claude-code/scripts/recall.py hindsight-integrations/claude-code/tests/test_hooks.py
- git diff --check

### DIFF
--- a/hindsight-integrations/claude-code/scripts/recall.py
+++ b/hindsight-integrations/claude-code/scripts/recall.py
@@ -72,6 +72,18 @@ def filter_by_min_scores(results: list[dict], min_scores: dict, config: dict) ->
     return filtered
 
 
+def unique_additional_banks(additional_banks: list, primary_bank_id: str) -> list[str]:
+    """Return configured extra banks without repeating the primary bank."""
+    seen = {primary_bank_id}
+    unique = []
+    for extra_bank_id in additional_banks:
+        if not isinstance(extra_bank_id, str) or not extra_bank_id or extra_bank_id in seen:
+            continue
+        seen.add(extra_bank_id)
+        unique.append(extra_bank_id)
+    return unique
+
+
 def read_transcript_messages(transcript_path: str) -> list:
     """Read messages from a JSONL transcript file for multi-turn context.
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Closing this in favor of #2625, which covers the same recallAdditionalBanks duplicate/primary-bank case and is already green. Thanks for picking it up.

PR #2621, opened 2026-07-08
Record refresh mental model outcome metadata

koriyoshi2041:
## Summary
- add outcome metadata for completed async mental-model refresh operations
- record content length, populated-content signal, and based_on counts by fact type
- keep submit-time metadata such as mental_model_id/name intact

Closes #2605

## Test
- uv run --with socksio pytest tests/test_mental_model_scheduled_refresh.py::test_async_refresh_writes_outcome_metadata -q -n0
- uv run ruff format --check hindsight_api/engine/memory_engine.py hindsight_api/engine/operation_metadata.py tests/test_mental_model_scheduled_refresh.py
- uv run ruff check hindsight_api/engine/memory_engine.py hindsight_api/engine/operation_metadata.py tests/test_mental_model_scheduled_refresh.py
- git diff --check

PR #2627, opened 2026-07-08
feat(api): enrich refresh_mental_model result_metadata with semantic outcome

### LINKED ISSUE #2605: refresh_mental_model result_metadata lacks semantic outcome fields (retain ops have them since 0.8.x)
## Environment
API v0.8.4, PostgreSQL backend.

## Problem
`refresh_mental_model` operations complete with `result_metadata` carrying only `{"name", "mental_model_id"}` — no semantic outcome. Historically (v0.5.x era) refreshes completed `status=completed` while writing a 19-char stub ("No answer provided.") to content: wire-level success masking semantic failure. That failure class appears fixed at 0.8.4, but the observability gap remains — a monitoring layer still cannot distinguish "refreshed with real content" from "refreshed empty" without a follow-up content fetch.

## What 0.8.4 already did right
Retain operations now expose rich semantic outcome (`facts_committed`, `unit_ids_count`, progress stages) — this request is parity for refresh ops.

## Suggestion
Add to refresh completion metadata: `content_len` (and/or `populated_content: bool`), ideally `based_on` counts. Cheap to emit at write time; converts every downstream health check from fetch-and-measure to metadata-read.

hb-cam:
Fixes #2605.

## What

`refresh_mental_model` operations completed with `result_metadata` carrying only the submit-time `{mental_model_id, name}` stub (set in `submit_async_refresh_mental_model` before the op runs, never enriched). This PR mirrors the retain enrichment pattern (`operation_metadata.py` + completion-time metadata write) as suggested in the issue thread: the worker handler now merges the semantic outcome into `result_metadata` at completion.

New fields on completed refresh ops:

- `content_len` — length of the final stored content
- `populated_content` — `true` only for real synthesis: the reflect agent's `"No answer provided."` fallback and the `"Generating content..."` placeholder complete wire-successful but read as `false` (a bare length check would report the 19-char stub as populated)
- `based_on_counts` — per-fact-type grounding counts from the reflect response

## How

- `RefreshMentalModelOutcomeMetadata` dataclass alongside `RetainOutcomeMetadata` in `operation_metadata.py`
- `_write_refresh_outcome_metadata` in the engine, mirroring `_write_retain_outcome_metadata`: jsonb merge (`COALESCE(result_metadata, '{}'::jsonb) || …`) so the submit-time keys consumers join on are **preserved, not replaced**; best-effort with a loud warning on failure
- Called from `_handle_refresh_mental_model` after a successful refresh, before the worker marks the operation completed (same ordering as retain)
- The reflect agent's fallback literal is promoted to a `NO_ANSWER_TEXT` constant so the populated judgment compares against the constant rather than a copied string

No endpoint or request model changes — `result_metadata` is already a free-form dict in the operation status response, so no OpenAPI/client regeneration is needed.

## Testing

- New `tests/test_refresh_outcome_metadata.py` (2 tests, written first and watched fail): a completed refresh exposes all three fields with the submit-time keys intact; the no-answer stub completes but reads `populated_content: false`
- Tests exercise the real path (submit → task backend → handler → metadata write → `get_operation_status`) against the embedded pg0 database with only the LLM-bound `refresh_mental_model` stubbed
- Neighboring suites pass: `test_mental_model_scheduled_refresh.py`, `test_mental_model_delta.py`, `test_mental_model_hooks.py` (24 passed, 4 skipped), `test_async_batch_retain.py` (22 passed)
- `ruff check`, `ruff format --check`, `ty check` clean on changed files

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -60,6 +60,7 @@ from .llm_trace import (
 from .operation_metadata import (
     BatchRetainChildMetadata,
     BatchRetainParentMetadata,
+    RefreshMentalModelOutcomeMetadata,
     RetainExtractionErrors,
     RetainOutcomeAggregate,
     RetainOutcomeMetadata,
@@ -1828,6 +1829,10 @@ class MemoryEngine(MemoryEngineInterface):
         if refreshed is None:
             raise ValueError(f"Mental model {mental_model_id} not found in bank {bank_id}")
 
+        # Enrich the submit-time result_metadata with the semantic outcome
+        # before the worker marks the operation completed (#2605).
+        await self._write_refresh_outcome_metadata(task_dict.get("operation_id"), refreshed)
+
         # Compute facts/mental_models counts for the post-op validator hook.
         # refresh_mental_model already persisted everything; the hook only needs
         # tallies that derive from the stored reflect_response payload.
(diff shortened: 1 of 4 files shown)


koriyoshi2041:
I checked this locally on macOS with Python 3.11.14 and the behavior looks good to me.

What I ran:
- `env -u ALL_PROXY -u all_proxy -u HTTPS_PROXY -u https_proxy -u HTTP_PROXY -u http_proxy uv run pytest tests/test_refresh_outcome_metadata.py tests/test_mental_model_scheduled_refresh.py -q -n0`

Result: `8 passed, 2 warnings in 21.83s`. The first run inherited my local SOCKS proxy env and failed before exercising the patch because `socksio` is not installed; rerunning without proxy env cleared that and the focused refresh tests passed.

The extra no-answer stub coverage here is useful, so I am going to close my overlapping #2621 and leave this as the stronger path for #2605.

hb-cam:
Thanks @koriyoshi2041 for taking the time to verify this locally, and for closing #2621 in favor of this branch — much appreciated. Good catch on the SOCKS-proxy/socksio env interaction as well. If any no-answer stub cases from your #2621 work aren't covered by the tests here, happy to fold them in.

koriyoshi2041:
Thanks for checking. I compared my closed #2621 against this branch, and I do not see any extra no-answer case that still needs to be folded in.

The only substantive coverage in #2621 was the happy-path outcome metadata shape (`content_len`, `populated_content`, and `based_on_counts` while preserving submit-time `mental_model_id` / `name`). This PR covers that in `test_completed_refresh_enriches_result_metadata`, and it adds the stronger boundary I was missing: `test_no_answer_stub_reads_as_unpopulated` for the `No answer provided.` fallback.

So I think this branch already subsumes mine; no additional case from #2621 is needed from my side.

benfrank241:
verified locally — flagging for @nicoloboschi to make the merge call since this is a feature (observability parity), not a bug fix.

context on the classification: #2605 is an observability gap, not a behavior bug. the underlying "wire-success masks empty content" failure was already fixed at 0.8.4; what's missing is the metadata to *see* it. concretely, on current main a completed refresh op's `result_metadata` carries only `{mental_model_id, name}` — running this PR's `test_completed_refresh_enriches_result_metadata` against pristine main fails with `KeyError: 'content_len'`, and passes on the branch. clean fails-before / passes-after.

on the implementation itself, it's solid:
- mirrors the retain enrichment exactly — same `RetainOutcomeMetadata`-style dataclass and the same `COALESCE(result_metadata,'{}') || $2::jsonb` merge used in 4 other result_metadata writes here, so submit-time keys are preserved and there's no new dialect gap
- `populated_content` correctly reads both the `NO_ANSWER_TEXT` stub and the `MENTAL_MODEL_PENDING_CONTENT` placeholder as unpopulated (a bare length check would call the 19-char stub populated) — and it promotes the `"No answer provided."` literal to a shared constant so the judgment and the reflect agent stay in sync
- no OpenAPI/client regen needed (result_metadata is already free-form)
- 8/8 local (2 new outcome tests on the real submit→handler→write→get_operation_status path + 6 scheduled-refresh), and @koriyoshi2041 independently verified and closed the overlapping #2621 in favor of this branch

no blockers from my side — just don't want to self-merge a feature. nicolo, your call on whether the added metadata surface is worth it.

nicoloboschi (approved):

Issue #1002, opened 2026-04-12
[Feature]: Use tool-use structured output for Anthropic JSON responses to eliminate invalid JSON retries


hb-cam:
### Use Case

A 10K-node memory bank running consolidation and retain extraction against the Anthropic API (claude-opus-4-6). Consolidation runs continuously in the background synthesizing observations from ~5,000 unconsolidated memory units.

### Problem Statement

The Anthropic LLM provider requests JSON via prompt instruction but does not use Anthropic's tool-use mechanism for constrained decoding. This causes consolidation and retain extraction to produce invalid JSON at approximately a 1:1 ratio with successful calls.
 
Over a 3-hour window:
 
| Metric | Count |
|--------|-------|
| Successful LLM calls | 338 |
| Invalid JSON retry attempts | 331 |
| Fully exhausted retries (all 11 failed) | 11 |
| Consolidation batch failures | 9 |
 
The invalid JSON is not caused by markdown code fences (v0.5.0 PR #646 strips those). The model produces genuinely malformed JSON. A single 42-character retain call triggered 11 consecutive failures and took 759 seconds. The retry storm also contributed to a container OOM kill (exit 137).

### How This Feature Would Help

Eliminates invalid JSON retries entirely by using constrained decoding on Anthropic's side. The response is guaranteed to be valid JSON matching the schema. Specific impact:
 
- Consolidation throughput doubles (~200 units/hr to ~400+)
- Retry amplification eliminated (currently 11 retries x 3 batch attempts = up to 33 API calls per failed unit)
- Cost savings from ~331 wasted API calls per 3-hour window
- Latency improvement from 759s worst-case retain to 3-8s
- Reduced memory pressure (retry storm contributed to OOM kill)

### Proposed Solution

Use Anthropic's tool-use structured output for operations that expect JSON responses:
 
```python
msg = client.messages.create(
    model="claude-sonnet-4-20250514",
    tools=[{
        "name": "consolidate_observation",
        "description": "Process a new fact against existing observations",
        "input_schema": { ... }  # JSON Schema matching consolidation response format
    }],
    tool_choice={"type": "tool", "name": "consolidate_observation"},
    messages=[...]
)
result = next(b for b in msg.content if b.type == "tool_use").input
```
 
Apply to:
1. Consolidation calls (`_consolidate_with_llm`) -- define tool schema for action/observation/reason response format
2. Retain extraction calls -- define tool schema for fact extraction format
[...]

PR #2628, opened 2026-07-08
feat(anthropic): implement the batch API interface via Message Batches (50% token discount)

hb-cam:
## What

Implements the `LLMInterface` batch methods (`supports_batch_api` / `submit_batch` / `get_batch_status` / `retrieve_batch_results`) in `AnthropicLLM` via Anthropic's [Message Batches API](https://platform.claude.com/docs/en/build-with-claude/batch-processing), which bills all token usage at 50% of standard price.

The engine's batch path (retain fact extraction, gated on `retain_batch_enabled`) has been available to the OpenAI-compatible and Gemini providers, but the gate hard-fails on Anthropic purely because the provider never implemented the interface. This fills that gap — no engine changes.

## How

The engine speaks the OpenAI batch wire shape (JSONL entries with `custom_id`/`body` in, `response.body.choices[0].message.content` + OpenAI-keyed `usage` out), so the provider translates both directions, mirroring the existing conversion rules of its own `call()` path:

- **submit**: system messages fold into `system`; `max_completion_tokens` → `max_tokens`; `temperature` is dropped (the sync path never sends it either — current Claude models reject non-default sampling params); `response_format` json_schema becomes a single forced `tool_use` tool when `strict` (native constrained decoding, same approach as #1002) or a system-prompt schema injection otherwise; operator `extra_body` params merge directly since batch params are the raw Messages body. Anthropic batches take requests inline — no file upload step.
- **status**: `processing_status` maps onto the vocabulary the engine's poll loop speaks — `"ended"` → `"completed"` (per-request failures surface in the results, matching OpenAI's completed-with-errors semantics); non-terminal states pass through; `request_counts` aggregate to `total`/`completed`/`failed`.
- **results**: succeeded messages render as `choices[0].message.content` (forced-tool JSON re-serialized as the content string) with `prompt_tokens`/`completion_tokens`/`total_tokens` usage; errored/canceled/expired entries become per-result `error` entries so the engine's existing error handling applies unchanged.

## Testing

- 8 new tests in `tests/test_anthropic_batch.py` (written first, watched fail): request translation (strict and non-strict), temperature dropping, status mapping both ways, result translation for tool-use and text content, and the not-ended guard.
- Existing suites pass unchanged: `test_batch_api.py`, `test_batch_api_validation.py`, `test_anthropic_structured_tool_use.py`, `test_llm_extra_body.py` (41 tests).
- `ruff check`, `ruff format --check`, `ty check` clean.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
@@ -543,6 +543,211 @@ class AnthropicLLM(LLMInterface):
             raise last_exception
         raise RuntimeError("Anthropic tool call failed")
 
+    # ── Message Batches API (50% token discount) ─────────────────────────────
+
+    _BATCH_TOOL_NAME = "structured_response"
+
+    async def supports_batch_api(self) -> bool:
+        """Anthropic supports batch operations via the Message Batches API."""
+        return True
+
+    @staticmethod
+    def _map_batch_status(processing_status: str) -> str:
+        """Map Anthropic ``processing_status`` onto the OpenAI vocabulary.
+
+        The engine's poll loop breaks on "completed" and hard-fails on
+        "failed"/"expired"/"cancelled"; anything else keeps polling. Anthropic
+        batches only end as "ended" (per-request failures surface in the
+        results, mirroring OpenAI's "completed"-with-errors semantics), so
+        "ended" maps to "completed" and the non-terminal states pass through.
+        """
+        return "completed" if processing_status == "ended" else processing_status
+
+    def _translate_batch_body(self, body: dict[str, Any]) -> dict[str, Any]:
+        """Translate one OpenAI-shaped request body into Messages API params.
+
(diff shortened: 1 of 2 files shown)


benfrank241 (approved):
lgtm — clean fill of the batch interface for anthropic, no engine changes. the wire-shape translation both directions matches the openai-provider contract the poll loop is built around (`request_counts{total,completed,failed}`, `{custom_id, response.body}` / `{custom_id, error}`), and the conversion rules mirror `call()` exactly — system folding, `max_completion_tokens`→`max_tokens`, temperature dropped, strict→forced tool_use / non-strict→schema-in-system, extra_body merge. the status mapping is the right call: anthropic only ends as "ended"→"completed" with per-request failures surfacing as error entries, so the engine's existing per-result error handling applies unchanged and it never sees a spurious whole-batch fail. ran the 8 new tests + the existing test_batch_api/validation suites locally, 18 pass.

approving + kicked off CI. thanks again for the TDD + the thorough translation docstrings.

PR #2629, opened 2026-07-08
feat(anthropic): prompt caching via inline cache_control markers

hb-cam:
## What

Adds Anthropic prompt caching to `AnthropicLLM` via inline `cache_control` markers — the "inline-marker provider" strategy that `LLMInterface.get_or_create_cached_prefix`'s docstring already describes for Anthropic. No engine changes, no new config.

Today the provider sends no `cache_control` at all, so every call pays full input price on content the engine resends verbatim:

- fact extraction reuses the same system prompt across every chunk;
- the reflect agent loop resends the entire growing conversation on each of its iterations (up to `HINDSIGHT_API_REFLECT_MAX_ITERATIONS`), re-paying the full prefix each time.

Anthropic cache reads bill at ~10% of the base input price, so on reflect-heavy workloads (mental-model refresh) this is a large input-cost reduction with no latency penalty.

## How

Two breakpoints (of the four Anthropic allows):

1. **System prompt, both entry points** — rendered as a block list with `cache_control: {type: "ephemeral"}`. Caching is a prefix match, so this caches tools + system together. Schema text-injection (the non-strict `response_format` path) happens before marking and lands inside the cached block.
2. **Final message content block, `call_with_tools()` only** — each agent-loop request's end-marker becomes the next iteration's cache read point. String content converts to a text block (skipped for empty strings, which the API rejects); block-list content (tool results) gets the marker on its last block.

Marking is safe unconditionally: below the model's minimum cacheable prefix the marker is silently ignored — no write premium, no error. `cache_read_input_tokens` already flows through `_usage_from_anthropic_response` into metrics/tracing, so cache effectiveness is observable immediately via `cached_tokens`.

## Testing

- 5 new tests in `tests/test_anthropic_prompt_caching.py` (written first, watched fail): marker placement on both entry points, schema-injection landing inside the cached block, tool-result block-list handling, and the no-system passthrough.
- One existing assertion updated for the representation change: `test_non_strict_keeps_text_injection_fallback` checked a substring on `system` as a plain string; the schema-in-prompt behavior itself is unchanged and still covered.
- `test_anthropic_structured_tool_use.py`, `test_llm_extra_body.py`, and `test_mental_model_delta.py` pass; `ruff` and `ty` clean.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
@@ -34,6 +34,43 @@ def _usage_from_anthropic_response(response: Any) -> LLMResponseUsage:
     )
 
 
+_EPHEMERAL_CACHE = {"type": "ephemeral"}
+
+
+def _cached_system_blocks(system_prompt: str) -> list[dict[str, Any]]:
+    """Render the system prompt as a block list with a cache_control marker.
+
+    Anthropic prompt caching is a prefix match: marking the (single) system
+    block caches tools + system together. The system prompt is stable per
+    scope — fact extraction reuses it across every chunk, reflect and
+    consolidation keep their stable instructions there — so repeat calls read
+    it at ~10% of the base input price. Markers below the model's minimum
+    cacheable prefix are silently ignored (no write premium), so marking is
+    safe unconditionally. This is the "inline-marker provider" strategy that
+    ``LLMInterface.get_or_create_cached_prefix`` documents for Anthropic.
+    """
+    return [{"type": "text", "text": system_prompt, "cache_control": _EPHEMERAL_CACHE}]
+
+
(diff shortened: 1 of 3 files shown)


benfrank241 (approved):
lgtm — clean implementation of the inline-marker strategy the `get_or_create_cached_prefix` docstring already lays out for anthropic. two breakpoints (system, and the last message on the tool path), no engine or config changes, and the important detail is right: markers don't accumulate across the reflect loop — each call rebuilds the message list from plain engine content and marks only system + the final block, so it stays at 2 of 4 breakpoints while the read point walks forward each iteration. empty-string guard, tool-result block-list handling, and schema-injection-inside-the-cached-block are all covered. ran the 5 new tests + the structured-tool-use/extra_body suites locally, 36 pass. `cache_read_input_tokens` already flows into metrics so we'll see effectiveness immediately.

approving + kicked off CI. thanks for the TDD and the thorough docstrings.

PR #2632, opened 2026-07-09
feat(audit): optional per-bank gate for audit logging

cdbartholomew:
## Summary

Lets an embedder enable audit logging for some banks and not others, without changing the deployment-wide master switch. Additive and off by default — with no gate wired, behavior is unchanged.

### AuditLogger per-bank gate
- `AuditLogger.set_bank_gate(gate)` attaches an optional async `Callable[[bank_id], bool]`.
- `should_log(action, bank_id)` applies the global env/action gate first (cheap, sync `is_enabled`) and only then awaits the predicate.
- Default is `None` → `should_log` is equivalent to `is_enabled(action)`, so deployments that don't use per-bank gating pay nothing and behave exactly as before.
- Global-off short-circuits before the predicate; a raising predicate **fails closed** (not logged) so a broken gate can't silently enable auditing for a bank meant to be off.
- All four audit-write paths honor it: the HTTP `@audited` decorator, both MCP tool wrappers, and `audit_context` (the retain engine path).

### Configurable per-bank field
- New `audit_enabled` bank config field (hierarchical, overridable per bank), added to `_CONFIGURABLE_FIELDS` so the bank-config API accepts it. **Opt-in — defaults to `false`.**
- Distinct from the static `audit_log_enabled` master switch. Inert unless a per-bank gate consults it.

## Test plan
- New `tests/test_audit_bank_gate.py` (12 cases): no-gate backward-compat contract, gated selection, global-off short-circuit, fail-closed, `audit_context` honors the gate, gate clearing, and the `audit_enabled` field is configurable + defaults false.
- `ruff` + `ty` clean; configurable-field round-trip suite passes.
- Docs updated (`configuration.md`).

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2991,10 +2991,14 @@ def _make_audited_http(audit_logger_getter: Callable[[], AuditLogger | None]):
             @wraps(func)
             async def wrapper(*args, **kwargs):
                 al = audit_logger_getter()
-                if al is None or not al.is_enabled(action):
+                bank_id = kwargs.get("bank_id")
+                # should_log applies the global env/action gate AND the
+                # optional per-bank predicate; when no predicate is set it is
+                # equivalent to is_enabled(action), so default deployments are
+                # unaffected.
+                if al is None or not await al.should_log(action, bank_id):
                     return await func(*args, **kwargs)
 
-                bank_id = kwargs.get("bank_id")
                 started_at = _dt.now(_tz.utc)
 
                 req_data = None
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -1104,11 +1104,13 @@ DEFAULT_METRICS_BACKLOG_ENABLED = False  # Disabled by default: runs periodic pe
 DEFAULT_AUDIT_LOG_ENABLED = False  # Disabled by default
 DEFAULT_AUDIT_LOG_ACTIONS = ""  # Empty = audit all eligible actions
 DEFAULT_AUDIT_LOG_RETENTION_DAYS = -1  # -1 = keep forever
-# Per-bank audit toggle default. True so that when a per-bank gate IS wired,
-# a bank with no explicit setting participates in auditing (opt-out), which
-# preserves "audit everything" once the deployment-wide switch is on. Inert
-# unless a gate consults it.
-DEFAULT_AUDIT_ENABLED = True
+# Per-bank audit toggle default. FALSE — a bank with no explicit setting is
+# NOT audited (opt-in): auditing must be turned on deliberately per bank. This
+# default surfaces through the resolved bank config, so UIs reading it show a
+# new bank as off. Inert in deployments that don't wire a per-bank gate (the
+# master audit_log_enabled switch governs there), so OSS behavior is unchanged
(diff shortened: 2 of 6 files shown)


benfrank241 (changes requested):
the gate design here is genuinely nice — `should_log` layering the async per-bank predicate on top of the cheap sync `is_enabled`, global-off short-circuiting before the await, fail-closed on a raising gate, and all four write paths (http `@audited`, both mcp wrappers, `audit_context`) honoring it. i ran the 12 new tests locally and they all pass; the backward-compat contract (no gate → exactly `is_enabled`) is well pinned.

it can't merge yet though — CI is red on two things, one of which is this PR:

1. **`test-api` (1/3): `test_hierarchical_config.py::test_hierarchical_fields_categorization` — `assert 41 == 40`.** adding `audit_enabled` to `_CONFIGURABLE_FIELDS` bumped the count; that assert (line ~142) is a deliberate guardrail that trips whenever a configurable field is added. bump it to 41 and slot `audit_enabled` into the categorization lists that test checks. (the `test-api` 2/3 red is unrelated — the known pg0 embedded-postgres boot flake, ignore it.)
2. **`verify-generated-files`.** you edited `configuration.md` but the generated skill mirror wasn't regenerated. run `./scripts/generate-docs-skill.sh` and commit `skills/hindsight-docs/references/developer/configuration.md` alongside.

one design question before i'm comfortable merging, separate from CI: as shipped, `audit_enabled` + `set_bank_gate` are **inert in the OSS product** — nothing wires a gate by default, so the new configurable field has no effect here on its own. `set_bank_gate` is a clean extension point and i'm fine with that living in core; the part i want to think about is adding a per-bank *config field* (with bank-config API surface) whose only consumer is an out-of-band embedder. does the field itself belong in core, or should it live in the embed layer where the gate is actually wired, leaving just `set_bank_gate` here? lmk your thinking — happy to take it in core if there's a reason to, just want it deliberate rather than dead surface.

fix the two CI items and weigh in on the field question and i'll re-review.

nicoloboschi:
Superseded by #2827 — closing this one.

Same goal (audit some banks and not others), different mechanism. Rather than adding `audit_enabled` alongside the existing `audit_log_enabled`, #2827 makes `audit_log_enabled` itself hierarchical (env → tenant → bank).

Two reasons for the change of approach:

1. **Two near-identically-named switches are hard to tell apart.** `audit_log_enabled` vs `audit_enabled` is a coin flip at a glance, and the split meant reasoning about both to answer "is this bank audited?"

2. **The new field was inert.** Nothing read `audit_enabled` unless an embedder called `set_bank_gate` — no caller existed outside tests. So on a default deployment the field, and any UI built on it, would have saved successfully and changed nothing.

Reusing the existing switch forced three call sites into the open that the gate approach left untouched — the `StaticConfigProxy` guard flagged each one by refusing a global read of a now-configurable field:

- the sync `is_enabled` pre-filter, which would have short-circuited a bank opting *in* while the global default was off
- the retention sweep, which gated the purge on the global flag — a bank opting in under a default-off deployment would have accumulated audit rows that were never purged
- `_audit_memory_defense`, which was sync and bypassed per-bank gating entirely

The `should_log` shape and the fail-safe reasoning in this PR carried over — thanks, that part was the right design. The one inversion: #2827 falls back to the deployment default on resolution failure rather than failing closed, so a transient DB blip can't silently create an audit gap for a bank that's meant to be audited.

Happy to reopen if you'd rather go the gate route for an out-of-band store.

PR #2633, opened 2026-07-09
fix(llm): raise OpenRouter Qwen3 verification budget

### LINKED ISSUE #2630: OpenAI-compatible verification can fail on reasoning models because max_tokens=100 is exhausted by reasoning
### Summary

`OpenAICompatibleLLM.verify_connection()` currently performs the startup/provider connectivity probe with a hard-coded `max_completion_tokens=100` for the prompt `Say 'ok'`.

For OpenAI-compatible reasoning/thinking providers, that 100-token cap can be consumed entirely by reasoning tokens before the model emits any final `message.content`. Hindsight then reports the provider as failing verification even though the provider and model work normally with a slightly larger output budget.

This is easy to reproduce with OpenRouter + `qwen/qwen3.6-35b-a3b`, because the model has reasoning enabled by default on OpenRouter and reasoning is not mandatory.

### Environment

- Hindsight installed package observed locally: `0.8.1`
- Upstream `main` still has the same hard-coded verification call:
  - `verify_connection()` calls `self.call(..., max_completion_tokens=100, scope="verification")`
- Provider: OpenRouter via OpenAI-compatible API
- Model: `qwen/qwen3.6-35b-a3b`

OpenRouter model metadata for `qwen/qwen3.6-35b-a3b` reports:

```json
{
  "supported_parameters": ["max_tokens", "reasoning", "include_reasoning", "temperature", "top_p", "top_k", "tools", "response_format"],
  "reasoning": {
    "mandatory": false,
    "default_enabled": true
  }
}
```

### Current behavior

The startup verification request can fail with empty final content / length finish because all available output tokens are spent on reasoning.

Observed pattern from direct OpenRouter probes using the same verification prompt:

```text
model=qwen/qwen3.6-35b-a3b
messages=[{"role":"user","content":"Say 'ok'"}]
```

With a 100-token cap, the model may return:

```text
finish_reason=length
content=null or empty
reasoning_tokens ~= 95-107
completion_tokens=100
```

Hindsight then logs a verification warning/error similar to:

```text
Provider returned empty message content (... scope=verification, finish_reason=length, has_tool_calls=False, refusal=False)
Connection verification failed ...
```

The provider is not actually unreachable, and later LLM calls with sufficient output budget can succeed.

### Reproduction data

Repeated live probes against OpenRouter + `qwen/qwen3.6-35b-a3b` with `Say 'ok'`:

```text
max_tokens=128: 1/5 successful, 4/5 finish_reason=length with empty content
max_tokens=256: 5/5 successful
max_tokens=512: 5/5 successful
max_tokens=1024: 5/5 successful
```

The 256-token runs succeeded in this sample, but reasoning tokens reached ~188, so 512 looks like a safer default for this class of model.

A separate probe showed that disabling reasoning with `reasoning: {"effort":"none"}` also makes the verification succeed reliably, but globally disabling reasoning is not always desirable for Hindsight memory quality. The problem here is specifically that the connection test is too token-constrained for reasoning-enabled providers.

### Expected behavior

The provider verification check should not fail merely because a reasoning model spends the hard-coded 100-token budget before emitting the trivial final answer.

### Proposed fixes

Any of these would solve the operator-facing problem:

1. Increase the verification output cap from `100` to something safer, e.g. `512`.
2. Make the verification cap configurable, e.g. `HINDSIGHT_API_LLM_VERIFY_MAX_TOKENS=512`.
3. For verification only, optionally suppress/disable reasoning when the provider supports it, while leaving normal retain/consolidation/reflect calls unchanged.
4. Treat `scope="verification"` specially in `HINDSIGHT_API_LLM_EXTRA_BODY`, so operators can tune verification without changing all real LLM calls.

My preference would be option 2: a configurable verification max-token budget with a safer default. It preserves reasoning for real memory tasks and avoids provider/model-specific hacks.

### Related issues

- #1567 mentions the same `Say 'ok'` verification request shape and `max_tokens=100`, but the issue is about duplicate retry requests and request IDs, not the token budget being exhausted by reasoning tokens.
- #1227 documents that `HINDSIGHT_API_LLM_EXTRA_BODY` exists for normal provider request tuning, but using it to change max tokens or reasoning globally affects real memory calls too, not just verification.
- #2034 discusses a Control Plane connectivity test, but not the current startup verification token budget.
- #2431 is another `empty message content` / `finish_reason=length` case, but in structured reflect rather than startup provider verification.

r266-tech:
## Summary

- keep the normal OpenAI-compatible startup verification budget at 100 tokens
- raise it to 512 only for OpenRouter `qwen/qwen3...` model IDs, where hidden reasoning can consume the small budget before visible `ok` content is emitted
- add boundary coverage for Qwen3 positives plus Qwen2.5, near-miss aliases, and ordinary custom gateways staying on the default budget

Fixes #2630.

## Tests

- `python3 -m py_compile hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py hindsight-api-slim/tests/test_openai_compatible_response_hardening.py`

Focused pytest could not run in this local sparse checkout because `tests/conftest.py` imports project dependencies not installed in the cron environment (`asyncpg`). Ruff is also not installed in the local Python environment.


benfrank241 (changes requested):
nice catch and the diagnosis is spot on — reproduced the failure mode: openrouter has reasoning `default_enabled: true` for `qwen/qwen3*`, so the 100-token probe gets eaten by hidden reasoning before any visible `ok`. the matcher is careful too (qwen2.5, the `qwen30` near-miss, unnamespaced `my-qwen3-proxy`, and wrong-namespace `other/qwen3-proxy` all correctly stay at 100) and the 12 boundary cases pass locally.

one thing before merge though — i'd rather not special-case this by model id. #2630's own title is the general problem ("verification can fail on reasoning models"), and a `provider==openrouter && qwen/qwen3*` regex only covers today's model. the next reasoning-default model (deepseek-r1, gpt-oss with reasoning, glm, a future qwen tag) hits the exact same failure and needs another special-case. that's whack-a-mole.

the probe is a one-shot `Say 'ok'` at startup, and you're billed on actual output not the cap — so a bigger ceiling costs nothing for non-reasoning models and just lets reasoning models finish. could we drop the regex/provider gate entirely and raise the single default verification budget 100 → 512 for all openai-compatible providers? one constant, covers every reasoning model, and the test collapses to "verification uses the larger budget."

if you'd rather keep it scoped for a reason i'm missing, lmk — but the general bump feels both simpler and more robust here. happy to merge once it's generalized.

r266-tech:
Thanks — generalized as requested.

I pushed `0e4a6fe3`, which removes the OpenRouter/Qwen-specific matcher and raises the single OpenAI-compatible verification budget to 512 for startup probes. The tests now assert the larger budget for both OpenRouter reasoning-model cases and ordinary compatible gateways.

Checks run locally:
- `python3 -m py_compile hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py hindsight-api-slim/tests/test_openai_compatible_response_hardening.py`
- `uv run --frozen pytest hindsight-api-slim/tests/test_openai_compatible_response_hardening.py -q` (7 passed)
- `uv run --frozen ruff format --check hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py hindsight-api-slim/tests/test_openai_compatible_response_hardening.py`
- `uv run --frozen ruff check hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py hindsight-api-slim/tests/test_openai_compatible_response_hardening.py`
- `git diff --check`

benfrank241 (approved):
ack — generalized exactly as hoped. dropping the per-model matcher for a single `DEFAULT_VERIFICATION_MAX_COMPLETION_TOKENS = 512` covers every reasoning-default model, not just qwen3, and the one-shot "Say ok" probe bills on actual output so the higher ceiling costs nothing for non-reasoning gateways. tests now pin 512 for both a reasoning model and an ordinary compatible gateway. 7 pass locally. superseding my earlier changes-requested.

approving + kicked off CI.

PR #2635, opened 2026-07-09
fix(claude-code): isolate MCP server cwd

### LINKED ISSUE #2634: Claude Code plugin MCP server crashes at startup when cwd contains an unreadable `.env`
## Symptom

In any Claude Code project whose working directory contains a `.env` file the session user cannot read (e.g. a root- or deploy-user-owned mode-600 secrets file), the hindsight-memory plugin's MCP server dies at startup. `claude mcp list` from the project dir shows:

```
plugin:hindsight-memory:hindsight: bash .../scripts/run_mcp.sh - ✘ Failed to connect
```

The same command from `$HOME` (or any dir without a `.env`) connects fine. The session loses all knowledge tools (`agent_knowledge_recall`, page CRUD, ingest). The hook path (autoRecall/autoRetain) keeps working — the hooks run `recall.py`/`retain.py` directly and never import FastMCP.

This is a realistic layout for shared deploy hosts: the project `.env` holds secrets owned by a deploy user, agent sessions run as a different user.

## Root cause

Claude Code launches the plugin's MCP stdio server with **cwd = the session's project dir**, and `scripts/run_mcp.sh` execs `scripts/mcp_server.py` without changing directory. Then:

1. `mcp_server.py:31` runs `mcp = FastMCP("hindsight")` — before `load_config()` at `:34`, so no plugin config setting can avoid it.
2. `FastMCP.__init__` instantiates the MCP SDK's pydantic-settings `Settings`, whose `model_config` hardcodes `env_file=".env"` (`mcp/server/fastmcp/server.py`, mcp 1.28.0).
3. pydantic-settings resolves that path relative to the **process cwd** (the project dir), and python-dotenv `open()`s it unconditionally (`dotenv/main.py:63`, python-dotenv 1.2.2).
4. Existing-but-unreadable `.env` ⇒ crash:

```
  File ".../scripts/mcp_server.py", line 31, in <module>
    mcp = FastMCP("hindsight")
  File ".../mcp/server/fastmcp/server.py", line 185, in __init__
    self.settings = Settings(
  ...
  File ".../pydantic_settings/sources/providers/dotenv.py", line 80, in _static_read_env_file
    file_vars = dotenv_values(file_path, encoding=encoding or 'utf8')
  ...
  File ".../dotenv/main.py", line 63, in _get_stream
    with open(self.dotenv_path, encoding=self.encoding) as stream:
PermissionError: [Errno 13] Permission denied: '.env'
```

A *missing* `.env` is fine (pydantic-settings tolerates it); only existing-unreadable crashes. A readable `.env` also works.

## Repro (no root-owned file needed)

```bash
mkdir /tmp/repro && cd /tmp/repro
install -m 000 /dev/null .env
CLAUDE_PLUGIN_DATA=$HOME/.claude/plugins/data/hindsight-memory-hindsight \
CLAUDE_PLUGIN_ROOT=$HOME/.claude/plugins/cache/hindsight/hindsight-memory/0.7.1 \
bash "$CLAUDE_PLUGIN_ROOT/scripts/run_mcp.sh"
# exit 1 with the PermissionError above
```

Controls: the identical command succeeds (answers an MCP `initialize` over stdio) from a `.env`-free dir, and from the same dir after `chmod 600 .env`.

## Versions

- hindsight-memory plugin 0.7.1 (also observed on 0.7.3)
- Claude Code 2.1.204
- plugin venv: Python 3.14.5, mcp 1.28.0, pydantic 2.13.4, pydantic-settings 2.14.2, python-dotenv 1.2.2
- Linux (Fedora 43)

## Suggested fixes (any one suffices)

1. `cd "$CLAUDE_PLUGIN_DATA"` in `scripts/run_mcp.sh` before the exec — the data dir never contains a `.env`, and the server has no legitimate reason to read the *project's* dotenv.
2. `os.chdir(os.environ.get("CLAUDE_PLUGIN_DATA", ...))` at the top of `scripts/mcp_server.py`, before the FastMCP import/construction.
3. Catch `PermissionError` (e.g. `OSError`) around `FastMCP("hindsight")` and retry from a safe cwd, so an unreadable dotenv degrades instead of killing the server.
4. Construct `FastMCP` with explicit settings that don't read a cwd-relative `env_file` (the SDK's `Settings` accepts overrides), so cwd contents can't affect server init at all.

Happy to provide more detail. Workaround we're running meanwhile: a user-scope wrapper that cds to `$CLAUDE_PLUGIN_DATA` and execs the unmodified `run_mcp.sh`.

oldnicke:
Summary

Fixes #2634.

Claude Code starts the hindsight MCP stdio server from the active project directory. If that project contains an unreadable `.env`, FastMCP's settings initialization can fail before the plugin config loads.

This changes the launcher to:
- preserve the original Claude project cwd in `HINDSIGHT_MCP_PROJECT_CWD`
- switch to `CLAUDE_PLUGIN_DATA` before executing `mcp_server.py`
- derive the default bank from the preserved project cwd so `directoryBankMap` and project-based dynamic bank IDs keep working

Tests

- `uv run pytest tests/test_run_mcp.py tests/test_mcp_server.py -v`
- `uv run ruff check hindsight-integrations/claude-code/scripts/mcp_server.py hindsight-integrations/claude-code/tests/test_run_mcp.py hindsight-integrations/claude-code/tests/test_mcp_server.py`

### DIFF
--- a/hindsight-integrations/claude-code/scripts/mcp_server.py
+++ b/hindsight-integrations/claude-code/scripts/mcp_server.py
@@ -19,12 +19,11 @@ sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
 # Launched via scripts/run_mcp.sh which execs the venv's interpreter, so
 # `mcp` and friends resolve from ${CLAUDE_PLUGIN_DATA}/venv/site-packages.
-from mcp.server.fastmcp import FastMCP
-
+from lib.bank import derive_bank_id
 from lib.client import HindsightClient
 from lib.config import debug_log, load_config
 from lib.daemon import get_api_url
-from lib.bank import derive_bank_id
+from mcp.server.fastmcp import FastMCP
 
 # ── Server setup ────────────────────────────────────────
 
@@ -32,7 +31,11 @@ mcp = FastMCP("hindsight")
 
 # Resolve config at startup
 _config = load_config()
-_dbg = lambda *a: debug_log(_config, *a)
+
+
+def _dbg(*a):
+    debug_log(_config, *a)
+
 
 if not _config.get("enableKnowledgeTools"):
     # Knowledge tools are opt-out. When disabled we must NOT exit: the plugin
@@ -51,7 +54,8 @@ except Exception as e:
     print(f"[Hindsight MCP] Failed to resolve API URL: {e}", file=sys.stderr)
     sys.exit(1)
 
-_hook_input = {"cwd": os.getcwd(), "session_id": ""}
+_project_cwd = os.environ.get("HINDSIGHT_MCP_PROJECT_CWD", os.getcwd())
+_hook_input = {"cwd": _project_cwd, "session_id": ""}
 _default_bank_id = derive_bank_id(_hook_input, _config)
 _client = HindsightClient(
     _api_url,
--- a/hindsight-integrations/claude-code/scripts/run_mcp.sh
+++ b/hindsight-integrations/claude-code/scripts/run_mcp.sh
@@ -51,4 +51,12 @@ if [ ! -f "${REQ_CACHED}" ] \
   cp "${REQ_SRC}" "${REQ_CACHED}"
 fi
 
+# Claude Code launches MCP servers from the session project directory. If that
+# directory contains an unreadable .env, FastMCP's default settings loader can
+# fail before our config loads. Run from plugin-owned data instead, while
+# preserving the session project cwd for bank derivation.
+export HINDSIGHT_MCP_PROJECT_CWD="${PWD}"
+mkdir -p "${CLAUDE_PLUGIN_DATA}"
+cd "${CLAUDE_PLUGIN_DATA}"
+
 exec "${PY}" "${CLAUDE_PLUGIN_ROOT}/scripts/mcp_server.py"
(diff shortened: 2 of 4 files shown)


koriyoshi2041:
I checked this locally on macOS (bash 3.2.57, Python 3.11.14).

What passed:
- `uv run ruff check hindsight-integrations/claude-code/scripts/mcp_server.py hindsight-integrations/claude-code/tests/test_run_mcp.py hindsight-integrations/claude-code/tests/test_mcp_server.py`
- a direct launcher-tail check with a fake `exec`: it preserved the original project cwd in `HINDSIGHT_MCP_PROJECT_CWD` and changed the server cwd to `CLAUDE_PLUGIN_DATA`, so the intended runtime behavior looks right.

One local note: `uv run pytest tests/test_run_mcp.py tests/test_mcp_server.py -v` produced `6 passed, 1 failed` here. The failing case was `TestMcpServerWorkingDirectory::test_exec_runs_from_plugin_data_dir_when_project_env_is_unreadable`, with the helper returning empty stdout before the split. Since the direct tail check above passes and GitHub CI is green, this looks more like a local harness/subprocess quirk than a blocker in the patch itself.

PR #2636, opened 2026-07-09
fix(search): avoid year-0 crashes in Chinese rolling-window temporal extraction

### LINKED ISSUE #2533: Consolidation task not working because ValueError('year 0 is out of range')
### Bug Description

The consolidation task is not able to execute successfully. It appears to be stuck or failing during execution without completing the process.

During the consolidation process, the following error is shown:
```
Failed to search memories (ValueError): ValueError('year 0 is out of range')
```

I cancelled the old consolidation task and the new consolidation task auto created, and also not working.

### Steps to Reproduce

no idea.
but the timestamp in retain maybe is the issue.

### Expected Behavior

The consolidation task should:

Execute successfully without runtime errors
Complete memory search and aggregation
Return consolidated results as expected

### Actual Behavior

Consolidation task fails during execution
Memory search step crashes with a ValueError
Process does not complete

### Version

0.8.3 and 0.8.4

### LLM Provider

Other

r266-tech:
Fixes #2533.

## Summary
- avoid constructing year 0 when Chinese rolling-window expressions shift a low reference year backward
- return the existing no-temporal-constraint sentinel for impossible relative year shifts
- keep valid low-year rolling windows and unrelated parser `ValueError`s covered by regression tests

## Tests
- `uv run --frozen pytest tests/test_query_analyzer.py -q`
- `uv run --frozen ruff format --check hindsight_api/engine/chinese_temporal_periods.py tests/test_query_analyzer.py`
- `uv run --frozen ruff check hindsight_api/engine/chinese_temporal_periods.py tests/test_query_analyzer.py`
- `python3 -m py_compile hindsight_api/engine/chinese_temporal_periods.py`
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/chinese_temporal_periods.py
+++ b/hindsight-api-slim/hindsight_api/engine/chinese_temporal_periods.py
@@ -109,6 +109,11 @@ def extract_chinese_period(query: str, reference_date: datetime) -> DateRange | 
             end.replace(hour=23, minute=59, second=59, microsecond=999999),
         )
 
+    def safe_constraint(start: datetime | None, end: datetime | None) -> DateRange | NoTemporalConstraintSentinel:
+        if start is None or end is None:
+            return NO_TEMPORAL_CONSTRAINT
+        return constraint(start, end)
+
     def subtract_months(months: int) -> datetime:
         month_index = reference_date.month - months - 1
         year = reference_date.year + month_index // 12
@@ -126,10 +131,17 @@ def extract_chinese_period(query: str, reference_date: datetime) -> DateRange | 
         day = min(base_date.day, calendar.monthrange(year, month)[1])
         return base_date.replace(year=year, month=month, day=day)
 
-    def add_years(base_date: datetime, years: int) -> datetime:
+    def add_years(base_date: datetime, years: int) -> datetime | None:
         year = base_date.year + years
+        if year < datetime.min.year or year > datetime.max.year:
+            return None
         day = min(base_date.day, calendar.monthrange(year, base_date.month)[1])
         return base_date.replace(year=year, day=day)
+
+    def add_days(base_date: datetime | None, days: int) -> datetime | None:
+        if base_date is None:
+            return None
+        return base_date + timedelta(days=days)
 
     def has_chinese_temporal_context(match: re.Match[str]) -> bool:
         if match.end() >= len(query):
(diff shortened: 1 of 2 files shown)


benfrank241 (changes requested):
thanks for this — the root cause is right and i reproduced the original #2533 crash: `过去一年做了什么` with a reference date in year 1 hits `add_years(ref, -1)` → year 0 → `ValueError('year 0 is out of range')`, exactly the consolidation failure. and `过去一年` at year 1 now correctly degrades to no-constraint, year 2 still extracts. nice.

one thing blocks merge though: the guard isn't threaded all the way through. at line 962 (`relative_year_fixed_day_match`) you converted the `d` computation to `add_days(...)` (which returns `None` on underflow) but left the call as bare `constraint(d, d)` instead of `safe_constraint(d, d)` — so `None` still reaches `constraint`. it just trades one crash for another:

```
去年今天    ref year=1:     AttributeError: 'NoneType' object has no attribute 'replace'
大前年今天  ref year=1,2,3: AttributeError
去年昨天    ref year=1:     AttributeError
```

same failure mode (impossible year shift during consolidation search), so consolidation still breaks on these. the sibling `relative_year_daypart_match` block right above it already does the right thing with `safe_constraint` — line 962 just needs the same swap.

could you:
- swap `constraint(d, d)` → `safe_constraint(d, d)` at line 962
- add a regression covering `去年今天` (and ideally `大前年今天`) at a low reference year, asserting no-constraint rather than a crash
- while you're in there, quick audit of the remaining bare `constraint(d, d)` calls fed by a year-shifted base so we don't leave another one — the `前晚|前夜` one can also underflow at the absolute datetime.min boundary

once that's in i'll re-verify and merge.

r266-tech:
Thanks — pushed the sibling coverage update in `cb6b4468`.

Changes made:
- `relative_year_fixed_day_match` now returns through `safe_constraint(d, d)`.
- `add_days()` now returns `None` on datetime underflow/overflow, so callers can degrade to no temporal constraint instead of raising.
- `昨晚` / `前晚` daypart shortcuts now use `add_days()` + `safe_constraint()` too.
- Added low-year regressions for `去年今天`, `大前年今天`, `去年昨天`, `昨晚`, and `前晚`, plus a valid low-year `去年今天` extraction check.

Checks run locally:
- `uv run --frozen pytest hindsight-api-slim/tests/test_query_analyzer.py -q` (382 passed)
- `uv run --frozen ruff format --check hindsight-api-slim/hindsight_api/engine/chinese_temporal_periods.py hindsight-api-slim/tests/test_query_analyzer.py`
- `uv run --frozen ruff check hindsight-api-slim/hindsight_api/engine/chinese_temporal_periods.py hindsight-api-slim/tests/test_query_analyzer.py`
- `python3 -m py_compile hindsight-api-slim/hindsight_api/engine/chinese_temporal_periods.py hindsight-api-slim/tests/test_query_analyzer.py`
- `git diff --check`

benfrank241 (approved):
ack — sibling coverage complete, verified. the three cases that still crashed on the first pass (`去年今天`, `大前年今天`, `去年昨天` at ref year 1) now cleanly return NO_CONSTRAINT instead of AttributeError, `add_days()` degrades on underflow, and the `昨晚`/`前晚` daypart paths are on `safe_constraint` too. valid low-year windows still extract (去年今天 ref=2 → range), so no over-degrading. 382 pass locally. superseding my earlier changes-requested.

approving + kicked off CI.

PR #2646, opened 2026-07-10
fix(auth): authorize bank writes before provisioning

### LINKED ISSUE #2650: Rejected bank config updates can provision banks or leave partial state
## Bug Description

Rejected bank configuration updates can still create a missing bank or leave an existing bank partially updated.

Bank config writes currently reach `ConfigResolver` from multiple HTTP and MCP paths. Because the resolver also provisions missing bank rows, some callers perform lifecycle work before all requested operations have been authenticated, authorized, and validated.

This produces three related failure modes:

- A rejected PUT or template import can leave a newly provisioned bank behind.
- A PATCH containing both a profile field and config fields can persist the profile change before the config operation is rejected.
- MCP or HTTP config paths can bypass the engine-owned tenant authentication and operation-validation flow.

### Missing bank is created by a rejected update

1. Configure an operation validator that rejects `UPDATE_BANK_CONFIG`.
2. Choose a bank ID that does not exist.
3. Send one of the following requests with a config field such as `reflect_mission`:
   - `PUT /v1/default/banks/{bank_id}`
   - `POST /v1/default/banks/{bank_id}/import`
4. Observe that the update returns `403`.
5. Request `GET /v1/default/banks/{bank_id}/profile` or list banks.

The rejected request can leave the bank row and provisioning side effects behind.

### PATCH leaves a partial profile update

1. Create an existing bank with name `Original`.
2. Configure an operation validator that permits `UPDATE_BANK` but rejects `UPDATE_BANK_CONFIG`.
3. Send:

```http
PATCH /v1/default/banks/{bank_id}
Content-Type: application/json

{
  "name": "Changed",
  "reflect_mission": "blocked"
}
```

4. Observe the `403` response.
5. Fetch the bank profile.

The name can already be `Changed` even though the combined request was rejected.

## Expected Behavior

- Every external bank config read/write authenticates the tenant and enforces the corresponding operation inside `MemoryEngine`.
- All requested authorization and validation completes before provisioning or persistence.
- A rejected request has no bank, config, profile, default-template, or import side effects.
- Valid config-only and name-only requests continue to provision missing banks as documented.

## Actual Behavior

- Provisioning may happen before config authorization or validation.
- Related fields may be written in separate engine/resolver calls, allowing partial state.
- Config lifecycle and authorization behavior differs between HTTP, MCP, and template-import callers.

## Impact

- Unauthorized or invalid requests can create ghost banks and trigger default provisioning work.
- Clients receive an error while the requested resource has still changed.
- Multi-tenant extensions cannot rely on one consistent engine boundary for bank config operations.
- Template imports can leave partially applied state when a later operation is rejected.

## Root Cause

`ConfigResolver.update_bank_config()` combines validation, persistence, and implicit bank creation, while external callers invoke it directly. Provisioning therefore happens outside the engine layer that owns tenant authentication, operation validation, default-template hooks, and combined update ordering.

## Version

`main` at `7bab4db28` (before the associated fix)

## LLM Provider

Not applicable

Sanderhoff-alt:
## Summary

- Authorize client-owned bank writes before provisioning a missing bank.
- Preflight template imports and reuse each authorization decision exactly once.
- Route HTTP and MCP bank operations through the public `MemoryEngine` lifecycle APIs.

Fixes #2650

## Root cause

`ConfigResolver.update_bank_config()` previously combined validation, persistence, and implicit bank creation. Callers could therefore provision a bank before authorization completed, and multi-field requests could persist an earlier change before a later validation failure.

Template import had the same ordering problem across configuration, mental-model, directive, refresh, and read operations. Moving provisioning later also required accounting for resources supplied by the server-owned default template before selecting create or update semantics.

MCP bank creation additionally bypassed the public engine lifecycle by calling `_ensure_bank_exists()` directly.

### Bank configuration and profile updates

- Add typed engine APIs for reading, updating, and resetting bank configuration.
- Split config validation and normalization from persistence.
- Validate profile and config changes together before creating a bank or persisting either change.
- Route PUT, PATCH, config PATCH/DELETE, and MCP config updates through engine-owned authentication and authorization.
- Preserve name-only creation and the existing `422` response for invalid memory-defense policies.
- Return HTTP `400` for invalid PUT/PATCH profile values and `404` if an update-only request loses the bank before its final read.

### Template import

- Project the server-owned default-template state before provisioning so import collisions use update semantics.
- Preauthorize the create fallback required when best-effort default resource creation fails.
- Preauthorize every import write, mental-model refresh, and required read before creating the bank.
- Re-read actual resources after provisioning and consume one-shot grants keyed by operation and resource target.
- Bind authorization state to the engine, request context, bank, and originating asyncio task.
- Reject duplicate mental-model IDs and directive names before provisioning.
- Reject mental-model imports without an available LLM before dependent writes.
- Keep default-template configuration server-owned, without applying client `UPDATE_BANK_CONFIG` field restrictions.

### MCP bank creation

- Delegate `create_bank` to public `MemoryEngine` profile and update APIs.
- Preserve default-template application while keeping lifecycle checks and authorization ordering inside the engine.
- Avoid the duplicate profile read previously performed by the MCP path.

## Impact

- Rejected config updates, template imports, and MCP operations no longer create empty banks or persist partial profile/config state before authorization completes.
- Template imports handle default-template resource collisions and fallback creation without duplicate hook calls.
- HTTP and MCP paths consistently use the engine-owned bank lifecycle.

## Validation

- `./scripts/hooks/lint.sh`
- `uv run ty check hindsight_api/engine/interface.py hindsight_api/engine/memory_engine.py hindsight_api/api/http.py`
- `uv run pytest tests/test_mcp_tools.py -q -n 0` — 192 passed
- `uv run pytest tests/test_bank_templates.py -q` — 36 passed
- Focused authorization, concurrency, and HTTP regression selection — 11 passed

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2488,25 +2488,43 @@ def validate_bank_template(manifest: "BankTemplateManifest") -> list[str]:
     return errors
 
 
+def load_default_bank_template_manifest() -> "BankTemplateManifest | None":
+    """Parse and semantically validate the configured default bank template."""
+    template_dict = get_config().default_bank_template
+    if not template_dict:
+        return None
+    manifest = BankTemplateManifest.model_validate(template_dict)
+    semantic_errors = validate_bank_template(manifest)
+    if semantic_errors:
+        raise ValueError("; ".join(semantic_errors))
+    return manifest
+
+
 async def apply_bank_template_manifest(
-    memory,
+    memory: MemoryEngine,
     bank_id: str,
     manifest: "BankTemplateManifest",
     request_context: "RequestContext",
+    *,
+    authorize_config_update: bool = True,
 ) -> "BankTemplateImportResponse":
     """Apply a validated BankTemplateManifest to an existing bank.
 
     Shared by the /import endpoint and the default-template-on-create hook
     driven by HINDSIGHT_API_DEFAULT_BANK_TEMPLATE. The bank MUST already
     exist; caller is responsible for validation (Pydantic + validate_bank_template).
+    ``authorize_config_update`` is false only for the server-owned default
+    template, which is not a client configuration update.
     """
     config_applied = False
     if manifest.bank:
         config_updates = manifest.bank.get_config_updates()
         if config_updates:
-            await memory._config_resolver.update_bank_config(bank_id, config_updates, request_context)
+            if authorize_config_update:
+                await memory.update_bank_config(bank_id, config_updates, request_context=request_context)
+            else:
+                await memory._config_resolver.update_bank_config(bank_id, config_updates, request_context)
             config_applied = True
-
     created_ids: list[str] = []
     updated_ids: list[str] = []
     operation_ids: list[str] = []
@@ -5781,28 +5799,15 @@ def _register_routes(app: FastAPI):
     ):
         """Create or update an agent with disposition and mission."""
         try:
-            # Ensure bank exists, validating create_bank only when this call
-            # actually creates a missing bank.
-            await app.state.memory._ensure_bank_exists(
-                bank_id,
-                request_context,
-            )
-
-            # Update name if provided (stored in DB for display only, deprecated)
-            if request.name is not None:
-                await app.state.memory.update_bank(
-                    bank_id,
-                    name=request.name,
-                    request_context=request_context,
-                )
-
-            # Apply all config overrides (includes reflect_mission, disposition, retain settings)
+            # The engine validates and authorizes all requested changes before
+            # creating a missing bank.
             config_updates = request.get_config_updates()
-            if config_updates:
-                await app.state.memory._config_resolver.update_bank_config(bank_id, config_updates, request_context)
-
-            # Get final profile
-            final_profile = await app.state.memory.get_bank_profile(bank_id, request_context=request_context)
+            final_profile = await app.state.memory.update_bank(
+                bank_id,
+                name=request.name,
+                config_updates=config_updates or None,
+                request_context=request_context,
+            )
             disposition_dict = (
                 final_profile["disposition"].model_dump()
                 if hasattr(final_profile["disposition"], "model_dump")
@@ -5818,6 +5823,8 @@ def _register_routes(app: FastAPI):
             )
         except OperationValidationError as e:
             raise HTTPException(status_code=e.status_code, detail=e.reason)
+        except ValueError as e:
+            raise HTTPException(status_code=400, detail=str(e))
         except (AuthenticationError, HTTPException):
             raise
         except Exception as e:
@@ -5841,33 +5848,16 @@ def _register_routes(app: FastAPI):
     ):
         """Partially update an agent's profile (name, mission, disposition)."""
         try:
-            # PATCH is update-only; missing banks must not be created as a
-            # side effect of reading the profile.
-            existing_profile = await app.state.memory.get_bank_profile(
-                bank_id, request_context=request_context, create_if_missing=False
-            )
-            if existing_profile is None:
-                raise HTTPException(status_code=404, detail=f"Bank '{bank_id}' not found")
-
-            # Update name if provided (stored in DB for display only, deprecated)
-            if request.name is not None:
-                await app.state.memory.update_bank(
-                    bank_id,
-                    name=request.name,
-                    request_context=request_context,
-                )
-
-            # Apply all config overrides (includes reflect_mission, disposition, retain settings)
+            # Update every requested field through one engine call so all
+            # authorization and validation completes before either write.
             config_updates = request.get_config_updates()
-            if config_updates:
-                await app.state.memory._config_resolver.update_bank_config(bank_id, config_updates, request_context)
-
-            # Get final profile
-            final_profile = await app.state.memory.get_bank_profile(
-                bank_id, request_context=request_context, create_if_missing=False
+            final_profile = await app.state.memory.update_bank(
+                bank_id,
+                name=request.name,
+                config_updates=config_updates or None,
+                create_if_missing=False,
+                request_context=request_context,
             )
-            if final_profile is None:
-                raise HTTPException(status_code=404, detail=f"Bank '{bank_id}' not found")
             disposition_dict = (
                 final_profile["disposition"].model_dump()
                 if hasattr(final_profile["disposition"], "model_dump")
@@ -5883,6 +5873,8 @@ def _register_routes(app: FastAPI):
             )
         except OperationValidationError as e:
             raise HTTPException(status_code=e.status_code, detail=e.reason)
+        except ValueError as e:
+            raise HTTPException(status_code=400, detail=str(e))
         except (AuthenticationError, HTTPException):
             raise
         except Exception as e:
@@ -5978,12 +5970,12 @@ def _register_routes(app: FastAPI):
                     dry_run=True,
                 )
 
-            # Ensure bank exists, validating create_bank only when this import
-            # actually creates a missing target bank.
-            await app.state.memory._ensure_bank_exists(
-                bank_id,
-                request_context,
-            )
+            config_updates = body.bank.get_config_updates() if body.bank else {}
+            if not config_updates:
+                # Config-bearing imports create the bank through update_bank_config
+                # only after validation. Other imports retain the existing
+                # eager-create behavior needed by directive-only manifests.
+                await app.state.memory._ensure_bank_exists(bank_id, request_context)
 
             return await apply_bank_template_manifest(
                 memory=app.state.memory,
@@ -6359,25 +6351,8 @@ def _register_routes(app: FastAPI):
                 detail="Bank configuration API is disabled. Set HINDSIGHT_API_ENABLE_BANK_CONFIG_API=true to re-enable.",
             )
         try:
-            # Authenticate and set schema context for multi-tenant DB queries
-            await app.state.memory._authenticate_tenant(request_context)
-            if app.state.memory._operation_validator:
-                from hindsight_api.extensions import BankReadContext, BankReadOperation
-
-                ctx = BankReadContext(
-                    bank_id=bank_id, operation=BankReadOperation.GET_BANK_CONFIG, request_context=request_context
-                )
-                await app.state.memory._validate_operation(
-                    app.state.memory._operation_validator.validate_bank_read(ctx)
-                )
-
-            # Get resolved config from config resolver
-            config_dict = await app.state.memory._config_resolver.get_bank_config(bank_id, request_context)
-
-            # Get bank-specific overrides only
-            bank_overrides = await app.state.memory._config_resolver._load_bank_config(bank_id)
-
-            return BankConfigResponse(bank_id=bank_id, config=config_dict, overrides=bank_overrides)
+            state = await app.state.memory.get_bank_config(bank_id, request_context=request_context)
+            return BankConfigResponse(bank_id=bank_id, config=state.config, overrides=state.overrides)
         except OperationValidationError as e:
             raise HTTPException(status_code=e.status_code, detail=e.reason)
         except (AuthenticationError, HTTPException):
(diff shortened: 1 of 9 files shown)


benfrank241:
verified this locally and confirming it's a real bug — flagging for @nicoloboschi to own the merge, since the fix is a ~600-line refactor of the bank-config authorization lifecycle across http/mcp/engine rather than a narrow patch, and that's exactly the kind of authz-ordering / multi-tenant-boundary change where the maintainer should have eyes before it lands.

what i confirmed:

- **bug reproduces on main, and it's a touch worse than "partial update".** the combined `PATCH {"name":"Changed","reflect_mission":"blocked"}` with a validator that rejects `UPDATE_BANK_CONFIG` returns **200** on main with *both* fields applied — the config-op authorization is bypassed on that path, not just written out of order. `test_patch_rejection_does_not_partially_update_name` fails on main (expects 403, gets 200) and passes on this branch. that's #2650's third failure mode (http/mcp paths bypassing engine-owned operation validation).
- **the fix is well-tested and looks right** — routes config read/update/reset through typed engine apis, validates profile+config together before persisting, provisions inside the engine after authz, and preauthorizes import writes. 8/8 of the new provisioning/rejection integration tests pass on the branch.

no red flags in what i reviewed — my only reason for not self-merging is blast radius: this re-plumbs authorization ordering, so a subtle regression could open a new bypass or break provisioning for valid requests. nicolo, worth your review pass on the authz ordering (esp. the import preauthorization-reuse scoping and the name-vs-config validation ordering) before merge. happy to trigger CI whenever.

hindsight-api-slim/hindsight_api/api/http.py
```diff
@@ -2442,50 +2442,128 @@ def validate_bank_template(manifest: "BankTemplateManifest") -> list[str]:
...
-    config_applied = False
-    if manifest.bank:
-        config_updates = manifest.bank.get_config_updates()
-        if config_updates:
-            await memory._config_resolver.update_bank_config(bank_id, config_updates, request_context)
-            config_applied = True
+    config_updates = manifest.bank.get_config_updates() if manifest.bank else {}
+    existing_by_id: dict[str, dict[str, Any]] = {}
+    if manifest.mental_models:
+        existing = await memory.list_mental_models(bank_id=bank_id, request_context=request_context)
```
nicoloboschi, line 2557: **Must fix — behavioral regression with `HINDSIGHT_API_DEFAULT_BANK_TEMPLATE`.**

The old flow in `api_import_bank_template` was:

```
get_bank_profile()          # auto-creates -> applies default template
apply_bank_template_manifest()
  └─ list_mental_models()   # sees default-template models -> chooses UPDATE
```

This PR hoists the existence lookups **above** provisioning — `_ensure_bank_exists` now happens inside `bank_template_import_authorization`, which is entered further down. So these reads run against a bank that doesn't exist yet and always return `[]`.

When a server has a default template **and** a client imports a manifest sharing a mental-model `id` or directive `name`, two things break:

- **Wrong permission checked.** `write_operations` gets `CREATE_MENTAL_MODEL` / `CREATE_DIRECTIVE` where it should be `UPDATE_*`. A validator granting update-but-not-create (or the reverse) is now asked the wrong question.
- **500 on import.** `create_mental_model` does a bare `INSERT ... (id, bank_id, ...)` with no `ON CONFLICT` (`memory_engine.py:10798`), so the second insert of the same id raises a unique violation, which the handler surfaces as a 500.

Fix: provision first, then read existing state — likely splitting the context manager into an explicit `preauthorize → provision` step, then computing `existing_by_id` / `existing_by_name` inside the scope before applying.

No test covers the default-template + import intersection today; worth adding one alongside the fix.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -8715,6 +8754,291 @@ async def _ensure_bank_exists(
...
+        self,
+        bank_id: str,
+        request_context: "RequestContext",
+    ) -> _BankTemplateImportAuthorizationState | None:
+        """Return the current scope only when engine, tenant context, and bank match."""
+        state = _bank_template_import_authorization.get()
+        if (
+            state is None
+            or state.engine_id != id(self)
+            or state.request_context_id != id(request_context)
```
nicoloboschi: **Must fix — `id()` comparison without a strong reference is unsound.**

`_BankTemplateImportAuthorizationState` stores `id(request_context)` and `id(self)` but holds no reference to either object. CPython reuses addresses after GC, so a freed `RequestContext` and a newly allocated one can share an id. This is safe today only because the FastAPI handler happens to keep the context alive for the request's duration — that's an accident of the current call graph, not an invariant the type enforces.

The guard is also largely redundant: contextvars are already task-local, so the scope can't leak to a sibling request in the first place.

Either store the objects themselves (`state.request_context is request_context`) or drop the id checks and rely on the contextvar plus the `bank_id` match.

hindsight-api-slim/hindsight_api/api/http.py
```diff
@@ -2442,50 +2442,128 @@ def validate_bank_template(manifest: "BankTemplateManifest") -> list[str]:
...
 
 
 async def apply_bank_template_manifest(
-    memory,
+    memory: MemoryEngine,
     bank_id: str,
     manifest: "BankTemplateManifest",
     request_context: "RequestContext",
+    *,
+    authorize_config_update: bool = True,
```
nicoloboschi: **Should fix — the flag name doesn't describe what it does.**

`authorize_config_update` reads as "check permissions," but it actually means "also persist the config here." With `False`, `_apply_default_bank_template` persists the config itself *and* `_apply_authorized_bank_template_manifest` still reports `config_applied = bool(config_updates)` for a write it did not perform. One responsibility split across two call sites, with a name that hides the split.

`persist_config: bool` would make the contract obvious — or have the default-template path not pass the manifest's `bank` section at all, so there's nothing to skip.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -8715,6 +8754,291 @@ async def _ensure_bank_exists(
...
+
+    def _consume_preauthorized_config_update(
+        self,
+        bank_id: str,
+        updates: dict[str, Any],
+        request_context: "RequestContext",
+    ) -> dict[str, Any] | None:
+        """Return prevalidated config when this is the authorized import write."""
+        state = self._get_bank_template_import_authorization_state(bank_id, request_context)
+        if state is None or state.requested_config_updates != updates:
```
nicoloboschi: **Should fix — silent fallback defeats the purpose of the scope.**

If the dicts ever drift (a caller mutating `config_updates` between preauthorization and application), this quietly returns `None` and falls through to a *second* validator invocation — the exact double-charge `bank_template_import_authorization` exists to prevent. The docstring notes validators may reserve quota, so a silent second call is a real cost, not just noise.

The only in-tree caller passes the identical object, so a mismatch is a programming error. Raising here would surface it instead of masking it.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -12017,9 +12411,14 @@ async def update_bank(
...
-
-        # Return updated profile
-        return await self.get_bank_profile(bank_id, request_context=request_context)
+        profile = await self._get_bank_profile_authenticated(
+            bank_id,
+            request_context=request_context,
+            create_if_missing=False,
+        )
+        if profile is None:
+            raise RuntimeError(f"Bank '{bank_id}' was not found after updating it")
```
nicoloboschi, line 13216: **Should fix — 500 where PATCH previously returned 404.**

A concurrent `delete_bank` landing between the update and this read now produces a 500 with a traceback. The old `api_update_bank` handled exactly this case by re-fetching with `create_if_missing=False` and raising a 404. Minor, but it's a reachable race that got noisier.

nicoloboschi (changes requested):
## Summary

The architecture here is right, and the tests are unusually thorough — the parametrized "rejection does not create bank" cases and the "authenticate once, validate each operation" assertion pin exactly the right invariants. Splitting `validate_bank_config_updates` from `_persist_bank_config` and keeping lifecycle in `MemoryEngine` is the correct shape, and routing MCP through the same engine method closes a real gap (MCP previously skipped the `memory_defense` parse and the `UPDATE_BANK_CONFIG` check that HTTP enforced).

Requesting changes for one behavioral regression plus a soundness issue in the preauthorization scope.

### Must fix

1. **Import + `HINDSIGHT_API_DEFAULT_BANK_TEMPLATE` ordering regression** — hoisting the existence lookups above provisioning means a default template that shares a mental-model `id` or directive `name` with the imported manifest now produces a 500, and checks `CREATE_*` where it should check `UPDATE_*`. Details inline.
2. **`id()`-based identity matching without a strong reference** in the import authorization scope. Details inline.

### Should fix

3. `authorize_config_update` is misnamed and splits one responsibility across two call sites.
4. `_consume_preauthorized_config_update` silently falls back to a second validator call on mismatch — the exact double-charge the scope exists to prevent.
5. `RuntimeError` → 500 on a concurrent delete where PATCH previously returned 404.

### Nits

- `_consume_preauthorized_bank_write` matches by operation *type* only, not by target id/name. Correct today because the counts derive from the same manifest, but a grant for directive *A* can be spent on directive *B* — worth a comment noting the invariant.
- The contextvar propagates into background tasks. `submit_async_refresh_mental_model` drains its `get` token inline so the counters are empty before any task spawns, but a `create_task` inside the scope inherits a reference to the same *mutable* state object. Fine now; fragile if anything moves.
- `except ValueError` asymmetry: the config PATCH route maps `ValueError` → 400, but `PUT /banks/{bank_id}` doesn't, so an invalid field name there is a 500. Pre-existing, but this PR routes more validation through that path.
- `BankConfigState` could be `@dataclass(frozen=True)` — it's a pure return value.

### Security & performance

Security is a net positive: the "rejected request leaves no bank behind" fix closes a resource-creation-via-unauthorized-request vector, and MCP now enforces what HTTP did. Performance is neutral-to-better — PUT/PATCH drop a redundant `get_bank_profile` round-trip, at the cost of one extra `validate_bank_read` hook call on import.

hindsight-api-slim/hindsight_api/api/http.py
Sanderhoff-alt: Fixed in b4baf631c. Import authorization now projects the default-template state before provisioning, preauthorizes UPDATE for collisions, and also preauthorizes the CREATE fallback needed if best-effort default resource creation fails. After provisioning it re-reads the actual resources and safely selects update or create. Regression tests cover mental-model and directive collisions, fallback after default creation failure, and denial before the bank is created.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
Sanderhoff-alt: Fixed in b4baf631c. The authorization state now retains strong references to the engine and request context and matches them with `is`. It also binds the state to the originating asyncio task so inherited ContextVar state cannot be consumed by a child task.

hindsight-api-slim/hindsight_api/api/http.py
Sanderhoff-alt: Fixed in b4baf631c. `authorize_config_update` was removed. Client-owned config application, server-owned default config application, and template resource application now have separate paths, and resource result reporting only reports writes actually performed by that path.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
Sanderhoff-alt: Fixed in b4baf631c. The preauthorized config is stored as a deep-copied snapshot. A mismatch or repeated consumption now raises a RuntimeError instead of silently invoking the validator again. Tests verify that the hook is called only once.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
Sanderhoff-alt: Fixed in bb199fffb. The update-only PATCH path now converts a missing final profile into OperationValidationError(status_code=404). A regression test deletes the bank between the update and response read and verifies the 404.

Sanderhoff-alt:
I force-pushed the branch into three focused commits based on the latest `main`:

- `bb199fffb` validates bank config updates before provisioning
- `b4baf631c` preflights template import operations before provisioning
- `02571ff1d` routes MCP bank creation through the public engine API

All five inline review threads have been addressed and resolved. The four review-body nits are also covered:

- preauthorization is keyed by operation and resource target
- authorization state is bound to the originating asyncio task
- invalid PUT/PATCH profile values map to HTTP 400
- `BankConfigState` is a frozen dataclass

The default-template collision path now projects the expected default state before provisioning and authorizes both the expected UPDATE and the CREATE fallback required when best-effort default resource creation fails. The actual state is re-read after provisioning before applying the import.

Validation completed:

- `tests/test_bank_templates.py`: 36 passed
- focused authorization/concurrency/HTTP regressions: 11 passed
- full lint: passed
- `ty check` on changed sources: passed
- `git diff --check origin/main...HEAD`: passed

Could you please take another look?

Sanderhoff-alt:
@nicoloboschi All requested changes have now been addressed, and the five inline threads are resolved. The branch was force-pushed into three focused commits and the PR description has been updated to match the final scope. Could you please re-review when you have a chance?

PR #2648, opened 2026-07-10
fix(claude-code): retain session deltas

### LINKED ISSUE #2644: Stop hook resends entire session transcript every retain call (O(N²) growth) — still present in 0.7.4
## Summary

The `hindsight-memory` Claude Code plugin's Stop hook (`retain.py`), in the default `retainMode: "full-session"`, resends the **entire cumulative session transcript** on every single retain call instead of only the new messages since the last retain. This causes O(N²) cumulative bytes sent over a session's lifetime, and in real usage it produced multi-megabyte single retain payloads that wedge server-side workers indefinitely.

Confirmed still present in the latest release (**0.7.4**) — `track_retention()` was added in this version, but it only handles advancing past a `/compact` event; the full-session branch still unconditionally does `messages_to_retain = all_messages` (no delta tracking) on every non-compacted Stop event.

## Real-world impact (confirmed via server DB inspection)

- One Claude Code session was retained 76 separate times over about a month. Its payload grew from ~44KB at 55 messages to **~7.6MB (~1.1M tokens) at 7,678 messages** — each retain call resending the whole thing.
- On the server side, this forced our self-hosted Hindsight instance to split a single retain call into **139 sub-batches**, saturating the shared LLM worker concurrency pool.
- Direct DB inspection (`async_operations` table) turned up **8 zombie `processing` operations** in one bank, 7 of them `retain` tasks with **1.7–2MB `task_payload`** each, created 2026-06-15 through 2026-06-17 — wedged in `processing` for **three weeks**, with `updated_at` continuing to advance (worker kept re-claiming them without ever completing), and no way to cancel them via the API or CLI (`cancel` only supports `pending` status, not `processing` — had to hand-patch the Postgres row to `failed` directly).
- These giant wedged retains appear to have also starved an unrelated `consolidation` operation in the same bank, which failed repeatedly with `"connection was closed in the middle of operation"` (retry_count 5) while trying to process a backlog of 23,055 items.

## Root cause

`HindsightClient.retain()` posts to `/v1/default/banks/{id}/memories` with a `document_id`, and re-using the same `document_id` **overwrites** the previous document rather than merging (confirmed via `hindsight-api-slim/tests/test_large_document_replacement.py`). The plugin's full-session mode relies on this by re-posting the complete transcript under `document_id = session_id` every time, so the one document always reflects the full state. That means naively "just sending the delta" isn't safe unless each delta is written to its own document — which is why the bug wasn't a one-line fix.

## Suggested fix (client-side, implemented and tested locally)

1. Track a per-session checkpoint (`message_count`, `chunk`) in local plugin state, committed **only after a successful retain call** (so a failed call safely retries the same delta next time, and a successful one never double-sends).
2. Slice `messages_to_retain = all_messages[last_committed_count:]` instead of resending everything.
3. Give each delta its own document (`chunk 0` → `session_id`, `chunk N>0` → `session_id-cN`), since the API overwrites by `document_id`.
4. On `/compact` (transcript shrinks), reset the checkpoint to 0 and treat the new, shorter transcript as the next chunk's full payload once, then resume delta-only tracking.

Happy to share the patch (~80 lines across `retain.py` + `lib/state.py`) if useful — it passes the existing test suite except for one test that explicitly asserts the old overwrite-driven behavior (which the fix intentionally changes).

## Suggested fix (server-side, defense in depth)

Independent of the client fix: it'd be worth having Hindsight reject or cap oversized single `retain` items server-side (e.g. return 413/422 above some configurable size) rather than accepting arbitrarily large payloads and splitting them into 100+ sub-batches internally. That would keep a single misbehaving client from being able to saturate the shared worker pool the way this did.

Also: today's `operation cancel` (both CLI and `DELETE /v1/default/banks/{id}/operations/{id}`) only supports operations in `pending` status — once something is `processing`, there's no API-level way to kill a wedged job short of hand-editing the Postgres row. An admin-level "force cancel" for `processing` operations would have made this a lot easier to clean up.

## Environment

- Hindsight server: self-hosted, version 0.8.4 (Docker Swarm, Postgres backend)
- `hindsight-memory` Claude Code plugin: 0.7.4 (marketplace `vectorize-io/hindsight`)

oldnicke:
## Summary

Closes #2644

Fixes the Claude Code integration Stop hook so full-session retain no longer resends the entire cumulative transcript on every retain call. The first retain keeps the existing `session_id` document_id behavior, while later normal growth sends only new messages under `session_id-cN` chunk documents. If Claude Code compacts the transcript, the compacted transcript starts a fresh chunk instead of overwriting the pre-compaction document.

## Changes

- Track full-session retain progress with a typed `RetentionProgress` dataclass
- Retain only messages after the last committed transcript count for normal growth
- Skip retain when the transcript has no new messages
- Preserve compaction protection by starting a fresh chunk after transcript shrink
- Update Claude Code README to describe delta chunk behavior
- Add/update regression tests for growth, duplicate Stop hooks, and compaction

## Validation

```bash
uv run pytest hindsight-integrations/claude-code/tests/test_state.py hindsight-integrations/claude-code/tests/test_hooks.py -q
uv run ruff check hindsight-integrations/claude-code/scripts/retain.py hindsight-integrations/claude-code/scripts/lib/state.py hindsight-integrations/claude-code/tests/test_state.py hindsight-integrations/claude-code/tests/test_hooks.py
./scripts/hooks/lint.sh

### DIFF
--- a/hindsight-integrations/claude-code/README.md
+++ b/hindsight-integrations/claude-code/README.md
@@ -243,7 +243,7 @@ Auto-retain runs after Claude responds. It extracts the conversation transcript 
 | Setting | Env Var | Default | Description |
 |---------|---------|---------|-------------|
 | `autoRetain` | `HINDSIGHT_AUTO_RETAIN` | `true` | Master switch for auto-retain. Set to `false` to disable memory storage entirely. |
-| `retainMode` | `HINDSIGHT_RETAIN_MODE` | `"full-session"` | Retention strategy. `"full-session"` sends the full conversation transcript (with chunking). |
+| `retainMode` | `HINDSIGHT_RETAIN_MODE` | `"full-session"` | Retention strategy. `"full-session"` sends the initial transcript once, then only new messages in chunk documents; compaction starts a fresh chunk. |
 | `retainEveryNTurns` | — | `10` | How often to retain. `1` = every turn; `10` = every 10th turn. Higher values reduce API calls but delay memory capture. Values > 1 enable **chunked retention** with a sliding window. |
 | `retainOverlapTurns` | — | `2` | When chunked retention fires, this many extra turns from the previous chunk are included for continuity. Total window size = `retainEveryNTurns + retainOverlapTurns`. |
 | `retainRoles` | — | `["user", "assistant"]` | Which message roles to include in the retained transcript. |
--- a/hindsight-integrations/claude-code/scripts/lib/state.py
+++ b/hindsight-integrations/claude-code/scripts/lib/state.py
@@ -167,6 +167,55 @@ def _locked_read_modify_write(state_name: str, lock_name: str, modify_fn):
     return result
 
 
+def plan_retention(session_id: str, message_count: int) -> RetentionProgress:
+    """Plan retention state without writing the checkpoint.
+
+    The caller must only commit the returned progress after the retain request
+    succeeds; otherwise a transient API failure would skip unsent messages on the
+    next hook run.
+    """
+    data = read_state("retention_tracking.json", {})
+    entry = data.get(session_id, {"message_count": 0, "chunk": 0})
+    last_count = entry["message_count"]
+    chunk = entry["chunk"]
+    compacted = False
+    start_index = last_count
+
+    if message_count < last_count:
+        # Transcript shrank — compaction happened. The compacted transcript is
+        # new canonical content, so send it as a fresh chunk rather than trying
+        # to diff it against the pre-compaction transcript.
+        chunk += 1
+        compacted = True
+        start_index = 0
+    elif message_count > last_count and last_count > 0:
+        # The transcript grew normally. Store only the new suffix in its own
+        # document so repeated Stop hooks grow linearly instead of resending
+        # the whole session under a replacement document_id.
+        chunk += 1
+    elif message_count == last_count:
+        start_index = message_count
+
+    return RetentionProgress(chunk_index=chunk, compacted=compacted, start_index=start_index)
+
+
+def commit_retention(session_id: str, message_count: int, chunk_index: int) -> None:
+    """Persist the checkpoint for a successful retain."""
+
+    def _update(data):
+        data[session_id] = {"message_count": message_count, "chunk": chunk_index}
+
+        # Cap tracked sessions
+        if len(data) > 10000:
+            sorted_keys = sorted(data.keys())
+            for k in sorted_keys[: len(sorted_keys) // 2]:
+                del data[k]
+
+        return data, None
+
+    _locked_read_modify_write("retention_tracking.json", "retention_tracking.lock", _update)
+
+
 def track_retention(session_id: str, message_count: int) -> RetentionProgress:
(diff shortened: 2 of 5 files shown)


benfrank241 (changes requested):
really well-diagnosed report and the delta direction is exactly right — the O(N²) full-transcript resend is a genuine problem and the `all_messages[start_index:]` + per-chunk `document_id` approach is the correct shape. growth and compaction are both covered by the new tests. one correctness hole before this can go in though.

**the checkpoint commits before the retain succeeds, so a failed retain silently drops that delta.** in `retain.py`, `track_retention()` (line ~127) writes the new `message_count`/`chunk` to `retention_tracking.json` immediately — but `client.retain()` doesn't run until line ~230, and its failure path (line ~240) just prints to stderr with no rollback. so on any transient failure (network blip, server 5xx, busy worker) the checkpoint has already advanced, and the next Stop hook slices from the advanced count — the messages in the failed call never get re-sent. that content is lost from memory permanently.

the old full-resend behavior was self-healing here (a failed retain just re-sent everything next time); the delta version gives that up. and #2644's own suggested fix called this exact case out — point 1: *"committed only after a successful retain call (so a failed call safely retries the same delta next time)."*

could you reorder so the checkpoint only advances after a successful `client.retain()`? something like: compute the delta read-only (peek), do the retain, then commit the new count/chunk on success (and leave it untouched on the except path). worth a regression test too — a failed retain followed by another Stop hook should re-send the same delta, not skip it.

everything else looks good — the `RetentionProgress` dataclass, the compaction-resets-to-chunk behavior, and the backwards-compatible chunk-0 = plain `session_id`. just want the failure path to not lose data before we ship it (and it'll need a claude-code release once it lands).

oldnicke:
@benfrank241 I pushed a follow-up fix for the checkpoint issue you flagged.

Full-session retain now plans the delta without mutating state, and only commits the retention checkpoint after `client.retain()` succeeds. If `client.retain()` fails, the checkpoint remains unchanged so the next Stop hook retries the same delta.

I also added regression coverage for the failed-retain retry path.

Thanks @koriyoshi2041 for #2661 as well. I used the same plan-then-commit approach here so we can keep review focused on this PR.

koriyoshi2041 (approved):
Thanks for the follow-up. The plan/commit split fixes the blocker I was worried about: a failed retain no longer advances the checkpoint, so the next Stop hook retries the same delta.

I checked the updated branch locally:

- `uv run pytest hindsight-integrations/claude-code/tests/test_hooks.py hindsight-integrations/claude-code/tests/test_state.py -q`

Both pass on my side.

benfrank241 (approved):
ack — the checkpoint data-loss hole is closed: plan_retention is read-only, commit_retention only runs after client.retain() succeeds, and on failure the checkpoint is untouched so the next Stop hook retries the same delta. the failed-retain regression (test_full_session_retry_keeps_checkpoint_on_retain_failure) locks it in. 47/47 local. superseding my changes-requested. approving + running CI — will need a claude-code release once merged.

PR #2651, opened 2026-07-10
deps(security): require LiteLLM 1.84.0

r266-tech:
## Summary

- raise the LiteLLM lower bound to 1.84.0 for the API-slim package and the LiteLLM integration
- sync the root and integration lockfile `requires-dist` metadata with that floor
- keep the resolved lockfile package at LiteLLM 1.87.0, which is already above the patched 1.84.0 floor

Refs GHSA-4xpc-pv4p-pm3w.

## Tests

- `git diff --check`
- `uv lock --check`
- `(cd hindsight-integrations/litellm && uv lock --check)`

### DIFF
--- a/hindsight-api-slim/pyproject.toml
+++ b/hindsight-api-slim/pyproject.toml
@@ -51,7 +51,7 @@ dependencies = [
     "anthropic>=0.40.0",
     "typer>=0.9.0",
     "cohere>=5.0.0",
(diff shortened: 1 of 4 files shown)


benfrank241 (approved):
lgtm — verified GHSA-4xpc-pv4p-pm3w is CVE-2026-49468 (litellm auth bypass via host header injection, CRITICAL, `< 1.84.0`, patched in 1.84.0), so the floor is exactly the first patched version. `uv lock --check` passes on both the root and the integration, and the resolved package stays 1.87.0 — no behavior change, just raising the floor so nobody resolves below the patch. clean.

approving + kicked off CI.

PR #2652, opened 2026-07-10
feat(anthropic): carry the prompt-cache marker on batch system prompts

hb-cam:
Follow-up to #2628 + #2629 (thanks for the quick reviews and merges on both).

## What

`_translate_batch_body` sent `system` as a plain string, so batch requests never participated in prompt caching. This renders it through the same `_cached_system_blocks` helper the sync path uses — batch items are one-shots, so it applies `call()`'s one-shot rule exactly: **system is the sole cache breakpoint**.

Deliberately **no end-marker on batch messages**: that second breakpoint only pays off on the sync tool loop, where the next iteration reads the growing prefix back. Batch items are independent one-shots — marking each item's final message would be pure cache-write premium with nothing ever reading it.

## Why it pays

Every request in a retain batch shares the fact-extraction system prompt, so the first item's cache write serves the remaining items as best-effort reads within the batch — and the cache-read discount stacks with the 50% batch discount. Marking below the model's minimum cacheable prefix is silently ignored (no write premium), so this is safe unconditionally — same reasoning as the `_cached_system_blocks` docstring.

## Tests

TDD, mirroring the existing batch-suite style:
- cached-block wire shape: marker present on system, `messages` unmarked (`cache_control` nowhere in them)
- non-strict schema injection lands **inside** the cached block (part of the cached prefix)
- requests without a system message are unchanged (no `system` param, no markers)
- existing shape assertions updated from string to block list

`test_anthropic_batch.py` + `test_anthropic_prompt_caching.py` + `test_anthropic_structured_tool_use.py` + `test_llm_extra_body.py`: **46 pass** locally. (`test_llm_provider.py`'s 2 failures reproduce on clean `main` in my env — pre-existing/environmental, not from this diff.) `ruff check` + `ruff format --check` clean on both touched files.

Since `cache_read_input_tokens` already flows into metrics (#2629), batch cache effectiveness is observable immediately.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
@@ -617,6 +617,12 @@ class AnthropicLLM(LLMInterface):
         an OpenAI ``response_format`` json_schema becomes a single forced
         tool_use tool when strict (native constrained decoding, issue #1002),
         else the schema is injected into the system prompt.
+
+        The system prompt carries the same cache_control marker as the sync
+        one-shot path (its sole breakpoint): every request in a retain batch
+        shares the fact-extraction system prompt, so the first item's cache
+        write serves the remaining items as best-effort reads — and the
+        cache-read discount stacks with the 50% batch discount.
         """
         system_prompt: str | None = None
         messages: list[dict[str, Any]] = []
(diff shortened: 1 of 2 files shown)


benfrank241:
verified this locally — clean and correct — but flagging for @nicoloboschi to make the merge call since it's a `feat` (a caching optimization), not a bug fix. nothing misbehaves without it; batch just doesn't get the cache discount yet.

for the record, it's about the lowest-risk feature there is: it reuses the already-merged `_cached_system_blocks` helper (#2629) and applies `call()`'s one-shot rule to the batch path — system as the sole breakpoint, and deliberately no end-marker on batch messages (right call: batch items are independent one-shots, so an end-marker would be pure cache-write premium with nothing reading it back). it completes the gap #2628 left purely because it predated the caching merge. schema-injection still lands inside the cached block; no-system requests unchanged. 15/15 pass locally across the batch + caching suites.

no concerns from my side — nicolo, your call on whether to take the batch caching now.

PR #2659, opened 2026-07-10
feat(engine): add MemoryEngine.delete_memory_units bulk primitive

chrislatimer:
## Summary

Adds `MemoryEngine.delete_memory_units(unit_ids, *, request_context)` — a bulk
variant of `delete_memory_unit` that removes a list of unit ids with the same
referential-integrity lifecycle, batched by bank:

- `enqueue_relink_victims` before the cascade
- chunked cascade `DELETE` (FK `CASCADE` handles `unit_entities` / `memory_links` / observation history)
- `_delete_stale_observations_for_memories` racing-insert sweep
- bank-stats cache invalidation
- deduped async consolidation + graph_maintenance submission per bank

Gives retention loops, LRU eviction, and bulk-maintenance tools a single entry
point that keeps the cascade contract instead of open-coding `DELETE`s outside
the engine and drifting from it.

## Scope note

An earlier version of this PR also added a `last_recalled_at` column to
`memory_units` / `documents`. **Dropped**: it has no OSS consumer, and a
high-frequency write of an *indexed* column doesn't belong on the hot
`memory_units` table (non-HOT updates → bloat/vacuum/lock pressure on the
busiest table). Recency is better as an extension-owned side table, provisioned
through the extension bank-table seam. This PR is now just the bulk-delete
primitive.

## Test plan

- [x] `test_delete_memory_units_validation.py` — empty-list fast path, invalid-UUID rejection, valid-UUID normalization
- [x] `test_migration_shape.py` (172 pass) + existing cascade coverage (`test_bank_stats_cache_invalidation`, `test_graph_maintenance`, `test_delete_document`)

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -6274,6 +6274,185 @@ class MemoryEngine(MemoryEngineInterface):
 
         return result
 
+    async def delete_memory_units(
+        self,
+        unit_ids: list[str],
+        *,
+        request_context: "RequestContext",
+    ) -> dict[str, Any]:
+        """Bulk delete memory units, keeping the same lifecycle as the single-id path.
+
+        Callers pass a list of ``unit_ids`` and this method runs the same steps
+        :meth:`delete_memory_unit` runs for one id, batched by bank:
+
+        1. Look up ``(bank_id, fact_type)`` for every id in one round-trip.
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Closing per direction change: GDPR is moving to data-minimization (per-bank HINDSIGHT_API_STORE_DOCUMENT_TEXT) rather than a bulk-erasure primitive. Reopen if delete_memory_units is still wanted upstream.

PR #2661, opened 2026-07-11
Fix Claude Code full-session retain growth

koriyoshi2041:
## Summary

Fixes the Claude Code plugin full-session retain path so repeated Stop hooks stop resending the entire cumulative transcript.

- first full-session retain keeps the existing plain `session_id` document id
- later full-session retains send only messages after the last successful checkpoint to `session_id-cN`
- checkpoints advance only after the retain API call succeeds, so failed calls retry the same slice
- transcript compaction resets the slice to the compacted transcript while still using the next chunk document id
- old compaction-only state files are treated as already having used the plain `session_id` document, so they do not overwrite it on upgrade

This targets #2644.

## Tests

- `uv run pytest tests/test_state.py tests/test_hooks.py -q` -> `49 passed`
- `uv run pytest -q -k 'not test_exec_runs_from_plugin_data_dir_when_project_env_is_unreadable'` -> `202 passed, 1 deselected`
- `python3 -m py_compile hindsight-integrations/claude-code/scripts/retain.py hindsight-integrations/claude-code/scripts/lib/state.py`
- `git diff --check`
- pre-commit hook on commit passed, including `lint.sh`

One unrelated local test still fails for me unchanged: `tests/test_run_mcp.py::TestMcpServerWorkingDirectory::test_exec_runs_from_plugin_data_dir_when_project_env_is_unreadable` returns no output from the fake `exec` shell probe. I left that path untouched.

### DIFF
--- a/hindsight-integrations/claude-code/scripts/lib/state.py
+++ b/hindsight-integrations/claude-code/scripts/lib/state.py
@@ -194,3 +194,52 @@ def track_retention(session_id: str, message_count: int) -> tuple:
         return data, (chunk, compacted)
 
     return _locked_read_modify_write("retention_tracking.json", "retention_tracking.lock", _update)
+
+
+def plan_delta_retention(session_id: str, message_count: int) -> tuple:
+    """Plan the next full-session retain without mutating state.
+
+    Returns:
+        (start_index, document_index, compacted)
+
+    ``start_index`` is the first transcript message that has not been
+    successfully retained yet. ``document_index`` is used by the caller to build
+    the document_id: 0 maps to the plain session_id, then 1 maps to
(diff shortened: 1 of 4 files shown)


koriyoshi2041:
I missed that #2648 is already the main branch for #2644, so I am closing this to keep review focused there.

The only extra bit this branch was trying to cover is the changes-requested case from #2648: plan the delta without mutating state, commit the checkpoint only after `client.retain()` succeeds, and add a retry regression proving a failed retain resends the same slice on the next Stop hook.

If that branch stalls and maintainers want this version instead, I can reopen or adapt it, but for now #2648 is the right review thread.

PR #2663, opened 2026-07-11
fix(consolidation): request JSON dedup decisions

### LINKED ISSUE #2658: Consolidation dedup fails: `_DEDUP_PROMPT` produces key=value output instead of JSON
### Bug Description

The `_DEDUP_PROMPT` in `hindsight_api/engine/consolidation/consolidator.py` (around line 106) instructs the LLM to `respond action="merge" and provide \`text\``. The model takes this literally and returns key=value format, but the code expects JSON and calls `json.loads()`, causing a `JSONDecodeError` on every dedup-eligible consolidation.

The prompt:

```python
_DEDUP_PROMPT = """You reconcile long-term memory observations. A NEW observation is about to be \
stored, and it is highly similar to an EXISTING one:

[NEW] {new}
[EXISTING] {existing}

If they assert the SAME fact (wording aside), respond action="merge" and provide `text`: a single \
observation that preserves EVERY detail from both. If they differ in ANY important detail — a \
number/quantity, a named entity or language, a negation, or a condition — respond action="keep"."""
```

### Steps to Reproduce

1. Run Hindsight v0.8.4 with `consolidation_dedup_threshold` set to default (0.97) — dedup is active for all banks.
2. Use an LLM provider that follows prompt instructions literally (e.g. Ollama with `deepseek-v4-flash:cloud`).
3. Ingest a batch of similar/repetitive memories into a bank (e.g. agent heartbeat polls, recurring status checks).
4. Wait for the consolidation worker to process them. When two observations are semantically similar (above the 0.97 threshold), `_dedup_adjudicate()` fires and sends `_DEDUP_PROMPT` to the LLM.
5. The LLM returns `action="merge"\ntext="..."` (key=value format).
6. `_call_ollama_native()` / `json.loads()` fails with `JSONDecodeError: Expecting value: line 1 column 1 (char 0)`.
7. The task retries 4x with exponential backoff, then schedules another retry — creating an infinite loop of failed tasks.

### Expected Behavior

The LLM should return a valid JSON object matching the `_DedupDecision` Pydantic model:

```json
{"action": "merge", "text": "...", "reason": "..."}
```

Consolidation should complete successfully, merging near-duplicate observations.

### Actual Behavior

The LLM returns key=value format following the prompt's `respond action="merge"` instruction:

```
action="merge"
text="On July 7, 2026 at 14:07 UTC, 14:37 UTC, ... OpenClaw heartbeat polls occurred. The assistant checked host health and tasks..."
```

`json.loads()` crashes:

```
WARNING - Ollama JSON parse error (attempt 1/4): Expecting value: line 1 column 1 (char 0)
  Model: ollama/deepseek-v4-flash:cloud
  Content length: 650 chars
  Content preview: 'action="merge"\ntext="On July 7, 2026 at 14:07 UTC..."'

ERROR - Unexpected error during Ollama call: JSONDecodeError: Expecting value: line 1 column 1 (char 0)
ERROR - Task execution failed: consolidation, error: Expecting value: line 1 column 1 (char 0)
```

### Impact

- All banks are affected — `consolidation_dedup_threshold` defaults to 0.97, so dedup is active everywhere.
- Banks with repetitive memories (agent heartbeats, status polls) trigger dedup constantly and fail every time.
- Each failure retries 4x, then reschedules — 67 failed tasks accumulated in one bank over several days.
- Banks with diverse observations (e.g. `hermes`) mostly avoid the bug because dedup rarely triggers.

### Proposed fix

Change the prompt to explicitly request JSON output:

```python
_DEDUP_PROMPT = """You reconcile long-term memory observations. A NEW observation is about to be \
stored, and it is highly similar to an EXISTING one:

[NEW] {new}
[EXISTING] {existing}

Respond with ONLY a valid JSON object. Do NOT use key=value format. Do NOT use markdown fences.
Output ONLY JSON like: {{"action": "merge", "text": "...", "reason": "..."}}

If they assert the SAME fact (wording aside), set action to "merge" and provide text: a single \
observation that preserves EVERY detail from both. If they differ in ANY important detail — a \
number/quantity, a named entity or language, a negation, or a condition — set action to "keep"."""
```

Note: Curly braces in the JSON example must be escaped as `{{ }}` because the prompt is processed with Python's `.format(new=..., existing=...)`. Unescaped braces cause `KeyError: '"action"'`.

### Version

0.8.4

### LLM Provider

Ollama

r266-tech:
Fixes #2658.

## Summary

- change the consolidation dedup prompt from the old `action="merge"` key/value wording to explicit JSON shapes for both merge and keep decisions
- tolerate parsed Pydantic decisions, raw JSON strings, and raw dicts from the LLM layer before applying the dedup decision
- default malformed or legacy raw text responses to `keep` so a non-compliant model does not abort consolidation
- add regression coverage for JSON prompt formatting, raw JSON/dict response coercion, legacy raw text fallback, and the create-path merge/keep behavior with raw JSON responses

## Verification

- `ruff format --check hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py hindsight-api-slim/tests/test_consolidation_dedup.py`
- `ruff check hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py hindsight-api-slim/tests/test_consolidation_dedup.py`
- `git diff --check`
- `python3 -m compileall -q hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py hindsight-api-slim/tests/test_consolidation_dedup.py`
- isolated prompt-contract harness
- isolated raw decision coercion harness

`PYTHONPATH=hindsight-api-slim python3 -m pytest -o addopts='' hindsight-api-slim/tests/test_consolidation_dedup.py` could not run in this sparse cron environment because `asyncpg` is not installed. Earlier `uv run --frozen pytest ...` attempts were also blocked by unavailable locked platform wheels for this Python/platform combination.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -115,15 +115,38 @@ class _DedupDecision(BaseModel):
         return "keep"
 
 
+def _dedup_decision_from_response(raw: Any) -> _DedupDecision:
+    try:
+        if isinstance(raw, _DedupDecision):
+            return raw
+        if isinstance(raw, str):
+            return _DedupDecision.model_validate_json(raw)
+        return _DedupDecision.model_validate(raw)
+    except ValueError as exc:
+        logger.warning("Invalid consolidation dedup response %r; defaulting to keep: %s", raw, exc)
+        return _DedupDecision(action="keep", reason="invalid structured response")
+
+
 _DEDUP_PROMPT = """You reconcile long-term memory observations. A NEW observation is about to be \
 stored, and it is highly similar to an EXISTING one:
 
 [NEW] {new}
 [EXISTING] {existing}
 
-If they assert the SAME fact (wording aside), respond action="merge" and provide `text`: a single \
-observation that preserves EVERY detail from both. If they differ in ANY important detail — a \
-number/quantity, a named entity or language, a negation, or a condition — respond action="keep"."""
+Respond with ONLY one valid JSON object matching one of these shapes:
+
+For duplicate facts:
+{{"action": "merge", "text": "...", "reason": "..."}}
+
+For distinct facts:
+{{"action": "keep", "text": "", "reason": "..."}}
+
+Do NOT use key=value lines, markdown fences, or any text outside the JSON object.
+
+If they assert the SAME fact (wording aside), set "action" to "merge" and provide "text": a \
+single observation that preserves EVERY detail from both. If they differ in ANY important detail \
+— a number/quantity, a named entity or language, a negation, or a condition — set "action" to \
+"keep" and "text" to an empty string."""
 
 
 def _dedup_active(config: Any) -> bool:
@@ -201,10 +224,12 @@ async def _dedup_adjudicate(
     if best_id is None:
         return _DedupOutcome(best_id=None, merged_text="", should_merge=False)
 
-    decision: _DedupDecision = await dedup_llm_config.call(
-        messages=[{"role": "user", "content": _DEDUP_PROMPT.format(new=anchor_text, existing=best_text)}],
-        response_format=_DedupDecision,
-        scope="consolidation_dedup",
+    decision = _dedup_decision_from_response(
+        await dedup_llm_config.call(
+            messages=[{"role": "user", "content": _DEDUP_PROMPT.format(new=anchor_text, existing=best_text)}],
+            response_format=_DedupDecision,
+            scope="consolidation_dedup",
+        )
     )
     if decision.action != "merge":
         return _DedupOutcome(best_id=best_id, merged_text="", should_merge=False)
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (approved):
I checked this locally against the reported #2658 path. The prompt now avoids the key=value instruction, the `.format()` JSON examples are escaped correctly, and the new parser path keeps the existing Pydantic-object behavior while accepting raw JSON/dict returns.

Verified with:

```
uv run pytest hindsight-api-slim/tests/test_consolidation_dedup.py -q
git diff --check origin/main...HEAD
```

Both pass on my side.

benfrank241 (approved):
lgtm — the root cause is exactly the prompt: `respond action="merge"` reads as literal key=value to instruction-following models (ollama/deepseek), so they emit `action="merge"\ntext=...` and `json.loads` blows up → 4x retry + reschedule → infinite failed-dedup loop. rewriting the prompt to explicit JSON shapes ("Do NOT use key=value lines / fences") kills it, and `_dedup_decision_from_response` hardens the parse (pydantic / raw json str / dict, malformed → keep). complements #2611's `_normalize_action` rather than colliding — that validator is preserved, `_DedupDecision.reason` exists for the fallback. 32 pass locally including the new prompt-contract + raw-json/dict + legacy-raw-text tests.

approving + kicked off CI.

Issue #1888, opened 2026-06-01
Retain only extracts the first sub-batch of an oversized document (chunks != document body) — follow-up to #1838 / #1855


tobibodamer:
### Bug Description

#1855 (fixing #1838) made same-`document_id` large retains store the full **body** in `documents.original_text`. But the underlying async sub-batch defect still affects **fact extraction**: for a document large enough to be split, only the **first** sub-batch is chunked and turned into memories. The rest of the body is stored but never extracted.

This is now harder to notice than #1838 — the document looks complete (`original_text` is full) while most of its memories are silently missing. The document **chunks** expose it cleanly: their `chunk_text` sums to a single sub-batch, not the document.

Reproduces on the published `:latest` (0.7.1) and on `main` @ `3e99a3f` (includes #1855, merge `dee9a7b9`).

### Steps to Reproduce

1. Retain a **single** document under one `document_id`, `async: true`, with content larger than `RETAIN_BATCH_TOKENS` (default) — e.g. ~83,000 chars.
2. `GET /v1/default/banks/{bank}/documents/{id}` → `original_text` is the full ~83,000 chars. ✅ (post-#1855)
3. `GET /v1/default/banks/{bank}/documents/{id}/chunks` → sum the `chunk_text` lengths.

Result on a default install (83,400-char document):

| Field | Value |
|---|---|
| `original_text` length | **83,400** (full body) |
| chunk count | 11 |
| Σ `chunk_text` length | **29,883** (≈ one sub-batch / `RETAIN_BATCH_TOKENS`) |
| chunk coverage | the **first** sub-batch only — `chunk_index` 0…n over the document start |

The ~53k chars after the first sub-batch are never chunked, so they produce no memories.

### Expected Behavior

Retaining a document as one item chunks and extracts facts across the **entire** body — `Σ chunk_text ≈ original_text length` — regardless of internal sub-batch splitting. (Per docs guidance to pass whole conversations as one document.)

### Actual Behavior

Only the first sub-batch is chunked/extracted. `Σ chunk_text` equals one sub-batch (~`RETAIN_BATCH_TOKENS`) while `original_text` equals the full document — the two no longer agree. Setting `HINDSIGHT_API_RETAIN_BATCH_TOKENS` very high (so the document stays one batch) or using `async: false` makes `Σ chunk_text` equal the full body again and restores full extraction.

### Version

`0.7.1` (image `e82b2c05`) and source build of `main` @ `3e99a3f`. Both reproduce.

[...]

PR #2666, opened 2026-07-11
fix(retain): preserve memories when appending

### LINKED ISSUE #2664: [BUG] update_mode=append silently deletes existing memory_units on re-retain (v0.8.4) — breaks long-term memory in OpenClaw plugin v0.9.0
### Bug Description

When a retain request is sent with `update_mode="append"` and a stable `document_id`, the expected behavior (per the [API docs](https://hindsight.vectorize.io/api-reference) and the [OpenClaw plugin README](https://github.com/vectorize-io/hindsight/blob/main/hindsight-integrations/openclaw/README.md)) is that new content is concatenated to the existing document and memory accumulates. However, the server-side implementation performs a full replace on every retain with new content, silently deleting all existing memory_units for that document.

### Impact on OpenClaw plugin

The [hindsight-openclaw plugin](https://github.com/vectorize-io/hindsight/tree/main/hindsight-integrations/openclaw) uses **session-scoped document IDs** with `update_mode="append"` by default (PR #2259 made this unconditional). Each OpenClaw conversation session shares one document ID like `openclaw:agent:default:telegram:direct:123456`. Every retain within that session sends the accumulated transcript with `update_mode="append"`.

The plugin README says:

> *"all turns of the conversation accumulate under a single Hindsight document"*

But in practice, each autoRetain cycle **deletes all existing memory_units and re-extracts from scratch**. Over time, this causes:

- **Silent memory loss**: Old memory_units are deleted and the re-extraction produces fewer or different facts
- **OpenClaw daily session reset compounds the issue**: OpenClaw rotates sessions daily at 4:00 AM (default behavior, see [openclaw/openclaw#96751](https://github.com/openclaw/openclaw/issues/96751)). After reset, the next autoRetain re-retains the same document_id, triggering another full-replace. Old memories are permanently lost.
- **No long-term memory accumulation**: Despite being marketed as "biomimetic long-term memory," the plugin cannot accumulate memories across sessions

The combination of session-scoped document IDs + `update_mode=append` + `handle_document_tracking` full-replace + daily session reset makes the plugin behave as session-only memory rather than long-term memory.

### Reproduction with OpenClaw

1. Install `@vectorize-io/hindsight-openclaw` with `autoRetain: true`
2. Chat with OpenClaw across multiple sessions (spanning a daily reset)
3. Check `memory_units` table in PostgreSQL — only the last session's memories survive
4. Older memories have been silently deleted by `handle_document_tracking` cascade

## Root cause

In `hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py`, the `_try_delta_retain` path (line ~1170) sets `is_recovery = True` **only** when the `content_hash` matches the existing document's hash:

```python
    existing_chunk_hashes = ...
    if existing_chunk_hashes:
        is_recovery = True
```

When `update_mode="append"` is used, new content is prepended to the existing text before processing, so the `content_hash` is **always different** from the stored document. Therefore `is_recovery` is always `False`.

In the consumer batch TXN (lines ~1425):

```python
if is_recovery:
    await fact_storage.upsert_document_metadata(...)  # preserves old data
else:
    await fact_storage.handle_document_tracking(...)   # DELETE + CASCADE to all chunks & memory_units
```

`handle_document_tracking` deletes the existing document row (CASCADE to chunks, then to memory_units) and re-inserts everything. The `update_mode="append"` flag is only used to prepend existing text before chunking, but does not change the document tracking behavior.

## Expected behavior

When `update_mode="append"` is used with new content:
- The document text should be updated (old + new content concatenated)
- Old chunks whose content_hash hasn't changed should be preserved along with their memory_units
- Only new chunks should go through fact extraction
- `upsert_document_metadata` should be used instead of `handle_document_tracking` for append mode

Alternatively, if the current implementation is intentional, the API docs and the OpenClaw plugin README should be updated to clearly state that `update_mode="append"` only accumulates text but **replaces memory_units on every retain**.

## Database evidence

After repeated retains with `update_mode="append"` on the same `document_id`:

- `documents` table: text accumulates correctly (old + new content)
- `chunks` table: only chunks from the **last** retain exist (all from the latest `created_at` date)
- `memory_units` table: only memory_units from the **last** retain exist (all from the latest `created_at` date)
- `entities` table: accumulates orphaned entities (entities created but no longer referenced by unit_entities)

## Environment

- Hindsight v0.8.4
- PostgreSQL backend
- `@vectorize-io/hindsight-openclaw` v0.9.0
- `update_mode="append"` with stable session-scoped `document_id`
- OpenClaw daily session reset (default-on)

### Steps to Reproduce

1. Install `@vectorize-io/hindsight-openclaw` with `autoRetain: true`
2. Chat with OpenClaw across multiple sessions (spanning a daily reset)
3. Check `memory_units` table in PostgreSQL — only the last session's memories survive
4. Older memories have been silently deleted by `handle_document_tracking` cascade

### Expected Behavior

When `update_mode="append"` is used with new content:
- The document text should be updated (old + new content concatenated)
- Old chunks whose content_hash hasn't changed should be preserved along with their memory_units
- Only new chunks should go through fact extraction
- `upsert_document_metadata` should be used instead of `handle_document_tracking` for append mode

### Actual Behavior

After repeated retains with `update_mode="append"` on the same `document_id`:

- `documents` table: text accumulates correctly (old + new content)
- `chunks` table: only chunks from the **last** retain exist (all from the latest `created_at` date)
- `memory_units` table: only memory_units from the **last** retain exist (all from the latest `created_at` date)
- `entities` table: accumulates orphaned entities (entities created but no longer referenced by unit_entities)

### Version

0.8.4

### LLM Provider

Other

oldnicke:
## Summary

Closes #2664

Fixes `update_mode="append"` so a retain with a stable `document_id` extends the existing document without deleting its existing chunks and memory units. The document's stored `original_text` is updated to the appended body, while only the new request content is extracted and stored as additional chunks.

## Changes

- Keep append mode out of the delta/full-replace path that treats old chunks as removed
- Update document metadata via `upsert_document_metadata` during append instead of `handle_document_tracking`
- Preserve existing chunks and memory units, and offset new chunk indexes after existing chunks
- Preserve valid JSON-array `original_text` when appending transcript-style JSON arrays
- Add a regression test proving append keeps existing memory unit IDs and adds new units/chunks

## Validation

```bash
uv run ruff check hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py hindsight-api-slim/tests/test_delta_retain_duplicates.py
./scripts/hooks/lint.sh

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -3448,19 +3448,23 @@ class MemoryEngine(MemoryEngineInterface):
             # worth of chunks/memories (issue #1888). Counting uses the same
             # bank-resolved, strategy-applied chunk size the orchestrator chunks
             # with, so the offsets match the chunk_index values it assigns.
-            from .retain import fact_extraction, fact_storage
+            from .retain import chunk_storage, fact_extraction
 
             chunking_config = await self._resolve_retain_chunking_config(bank_id, request_context, strategy)
             chunk_offsets: dict[str, int] = {}
 
-            # In update_mode="append", retain_batch prepends the existing document
-            # body to the FIRST sub-batch as an extra content item before chunking
-            # (see orchestrator.retain_batch), consuming chunks(existing_body)
-            # additional chunk_index slots ahead of that sub-batch's own content.
-            # Capture that chunk count per document up front — the first sub-batch
-            # overwrites documents.original_text when it commits, so it can't be
-            # read back afterwards — and fold it into the offset so later
-            # sub-batches continue past the prepended chunks instead of colliding.
+            # In update_mode="append" with the new preserve-units path
+            # (orchestrator.retain_batch >= PR #2666), the first sub-batch
+            # internally sets chunk_index_offset = existing_chunk_count (the
+            # orchestrator reads the DB and overrides whatever offset the caller
+            # passed).  memory_engine's chunk_offsets dict starts at 0, so after
+            # sub-batch 1 it would store chunk_offsets[doc] = 0+1 = 1 instead of
+            # existing_count+1.  Sub-batch 2 would then receive offset 1 and
+            # collide with existing chunk index 1.
+            #
+            # Fix: seed chunk_offsets with the current existing chunk count for
+            # every document that will be appended to, so the cursor starts in the
+            # right place and memory_engine's accounting tracks correctly.
             append_prepend_chunks: dict[str, int] = {}
             backend = await self._get_backend()
             append_doc_ids: set[str] = set()
@@ -3470,15 +3474,10 @@ class MemoryEngine(MemoryEngineInterface):
                     append_doc_ids.add(item_doc_id)
             for append_doc_id in append_doc_ids:
                 async with acquire_with_retry(backend) as conn:
-                    existing_text = await fact_storage.get_document_content(conn, bank_id, append_doc_id)
-                if existing_text:
-                    append_prepend_chunks[append_doc_id] = len(
-                        fact_extraction.chunk_text(
-                            existing_text,
-                            chunking_config.chunk_size,
-                            structured_chunk_size=chunking_config.structured_chunk_size,
-                        )
-                    )
+                    existing_chunks = await chunk_storage.load_existing_chunks(conn, bank_id, append_doc_id)
+                if existing_chunks:
+                    existing_count = max((c.chunk_index for c in existing_chunks), default=-1) + 1
+                    chunk_offsets[append_doc_id] = existing_count
 
             for i, (sub_batch, sub_origins) in enumerate(zip(sub_batches, origin_indices), 1):
                 # Checkpoint: abort if the operation was deleted (bank was deleted) between sub-batches.
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -824,54 +824,37 @@ async def retain_batch(
     if update_mode == "append" and effective_doc_id and is_first_batch:
         async with acquire_with_retry(pool) as conn:
             existing_text = await fact_storage.get_document_content(conn, bank_id, effective_doc_id)
+            existing_chunks = await chunk_storage.load_existing_chunks(conn, bank_id, effective_doc_id)
         if existing_text:
-            # Prepend existing text as a new content item at the beginning
-            existing_content: RetainContentDict = {"content": existing_text}
-            # Copy context/tags from first item for consistency
-            first = contents_dicts[0]
-            if first.get("context"):
-                existing_content["context"] = first["context"]
-            if first.get("event_date"):
-                existing_content["event_date"] = first["event_date"]
-            if first.get("metadata"):
-                existing_content["metadata"] = first["metadata"]
-            if first.get("observation_scopes") is not None:
-                existing_content["observation_scopes"] = first["observation_scopes"]
-            if first.get("tags"):
-                existing_content["tags"] = first["tags"]
-            contents_dicts = [existing_content, *contents_dicts]
-            # Merge JSON arrays to keep original_text valid (#2409).
-            # Without this, combined_content joins items with "\n", producing
-            # "[...]\n[...]" which is not valid JSON. On the next append cycle
-            # chunk_text() fails to parse it and falls through to sentence-
-            # boundary text splitting, breaking speaker attribution.
+            incoming_text = (
+                document_body_override
+                if document_body_override is not None
+                else "\n".join([c.get("content", "") for c in contents_dicts])
+            )
+            appended_body = f"{existing_text}\n{incoming_text}"
+
+            # Keep stored original_text valid for JSON-array transcripts (#2409),
+            # but do not reprocess the old text. Append should accumulate memory:
+            # old chunks/memory_units remain attached to the document, and only
+            # this request's new content is extracted into additional chunks.
             try:
-                _merged = []
-                for _item in contents_dicts:
-                    _parsed = json.loads(_item.get("content", ""))
-                    if isinstance(_parsed, list) and all(isinstance(_e, dict) for _e in _parsed):
-                        _merged.extend(_parsed)
-                    else:
-                        _merged = None
-                        break
-                if _merged is not None:
-                    contents_dicts = [{"content": json.dumps(_merged, ensure_ascii=False)}]
-                    if first.get("context"):
-                        contents_dicts[0]["context"] = first["context"]
-                    if first.get("event_date"):
-                        contents_dicts[0]["event_date"] = first["event_date"]
-                    if first.get("metadata"):
-                        contents_dicts[0]["metadata"] = first["metadata"]
-                    if first.get("observation_scopes") is not None:
-                        contents_dicts[0]["observation_scopes"] = first["observation_scopes"]
-                    if first.get("tags"):
-                        contents_dicts[0]["tags"] = first["tags"]
+                existing_json = json.loads(existing_text)
+                incoming_json = json.loads(incoming_text)
+                if (
+                    isinstance(existing_json, list)
+                    and isinstance(incoming_json, list)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by koriyoshi2041 — CHANGES_REQUESTED** [review]
Thanks for taking this on. The short append regression passes locally, but this branch seems to regress the existing oversized append coverage.

I ran:

```bash
env -u ALL_PROXY -u all_proxy -u HTTPS_PROXY -u https_proxy -u HTTP_PROXY -u http_proxy uv run pytest tests/test_delta_retain_duplicates.py::test_append_mode_preserves_existing_memory_units tests/test_subbatch_chunk_coverage.py::test_oversized_append_chunks_cover_existing_plus_new -q -n0 --tb=short -o log_cli=false
```

The new test passes, but `test_oversized_append_chunks_cover_existing_plus_new` fails with:

```
append chunks cover only 14338/18538 chars (existing+new)
```

From the logs, after the first large retain reaches 64 committed chunks, the oversized append sub-batches start hitting the recovery skip path (`Producer: skipped 1/1 already-committed chunks`) instead of storing the rest of the appended body. So the preserve-existing path looks good for short appends, but it still needs to keep the existing #1888 append-sub-batch coverage green before merging.

benfrank241:
this is a real and serious bug — append silently full-replacing (deleting existing memory_units) breaks long-term accumulation for openclaw's session-scoped-id + append pattern — and the fix direction is right: skip the delta/full-replace path for append, preserve existing chunks/units, and use `upsert_document_metadata` instead of the delete-cascading `handle_document_tracking`. flagging for @nicoloboschi rather than self-merging though, for two reasons:

1. **i can't validate the fix through a failing-on-main test.** `test_append_mode_preserves_existing_memory_units` passes on `main` too — it uses `memory_no_llm`, so extraction is deterministic, the original fact re-extracts identically, and delta-recovery preserves the unit id even on the buggy code. so it doesn't reproduce #2664's deletion (which appears to need real-LLM re-extraction changing the facts). the test proves append *accumulates* structurally, but not that it *fixes* the deletion. a regression that fails on main would make this a clean merge.

2. **it's a ~110-line change in the highest-churn retain path**, overlapping a stack of append PRs from this week — it rebases on and rewrites the exact code #2684 added, and sits alongside #2648 / #2682 / #2505. that's a lot of concurrent surface for one person to self-merge on.

what i did confirm: #2684's metadata projection test still passes on this branch (rewrite doesn't regress it), the new test passes on the branch, and the one red (`test_repeated_upsert_chunks_not_scrambled`) is xdist pg0 pollution — passes in isolation and serially, not a regression.

nicolo — worth your eyes on the append lifecycle here, and ideally a real-LLM regression that fails on main so we lock the deletion out for good.

benfrank241:
re-reviewed after your push, @oldnicke — thanks for adding the real-LLM regression, that was exactly the right move, and the oversized-append offset fix is a good catch too.

heads up though: **the new `test_append_mode_preserves_memory_units_with_real_llm` passes on current `main` too.** i ran it both ways with a real openai LLM:

- this branch: `1 passed`
- current `main`: also `1 passed` (i expected it to fail, reproducing the deletion)

so it still isn't a fails-before repro of #2664's delete-cascade. and there's a wrinkle that i think is the real story: since this PR was opened, **#2684 and #2648 both merged and reworked the same append path**, so "current main" is a long way from the v0.8.4 the bug was filed against. the test passing on today's main suggests the recent append merges may already cover this scenario.

so for @nicoloboschi — the open question isn't really the fix quality (direction's still right: skip full-replace, preserve chunks/units, upsert instead of handle_document_tracking). it's **whether the delete-cascade is still reproducible on today's main at all**, or whether #2684/#2648 already shifted it — because if we can't get a test that fails on current main, we can't lock the regression out. worth a real-LLM repro against current main (or confirming #2664 is still live) before this lands. keeping it in your queue rather than self-merging.

oldnicke:
@benfrank241 Thanks for re-testing this against current main. You're right that the real-LLM test does not currently provide a fails-before regression.

I will rerun the original #2664 scenario against current main and this branch, comparing the persisted document, chunks, and memory_unit IDs. If current main no longer reproduces the deletion after #2684/#2648, I'll narrow or close this PR rather than merge an overlapping retain-path rewrite without a current failing case.

oldnicke:
@benfrank241 Thanks for the re-test. I reran the scenario against current origin/main(395823f7b) and this PR using the same real LLM provider.

I strengthened the assertion to compare both the existing memory_unit IDs and their created_at values, so a delete-and-reinsert with equivalent content cannot pass accidentally. I also ran the existing oversized append sub-batch coverage test.

Results:

- current main: 2 passed
- PR branch: 2 passed

Current main enters delta retain for the append:
1 unchanged, 1 new, 0 removed, 0 deleted chunks. The original memory unit row and created_at are preserved.

I therefore cannot reproduce #2664's delete cascade on current main after the recent append changes. Unless there is another current-main reproducer, I think this PR should be closed as superseded rather than merging an overlapping retain-path rewrite.

nicoloboschi:
Thanks for the detailed writeup and the PR — the trace was clear enough to reproduce against, which made this quick to check.

I ran this against current `main` (`913b99c`) with `provider=none`, so every chunk maps 1:1 to a memory unit and unit survival is directly observable. Four append scenarios:

| Case | Existing units | Survived the append |
|---|---|---|
| plain text, single chunk | 1 | 1/1 |
| plain text, multi-chunk | 3 | 3/3 |
| JSON transcript, multi-chunk | 10 | 10/10 |
| JSON transcript, single chunk | 1 | 0/1 |

So `update_mode="append"` does not full-replace in general — three of the four cases preserve every existing memory unit.

## The mechanism is not `is_recovery`

The root-cause trace points at the `is_recovery` gate in `_streaming_retain_batch`, but that isn't the preservation mechanism. `_try_delta_retain` runs earlier (`orchestrator.py:892`) and returns before streaming is ever reached on the healthy path — it's what preserves unchanged chunks and their memory units, and it does run for append.

What actually happens in the failing case is chunk-hash drift on a not-yet-full chunk:

```
retain #1:  [turn1, turn2]
            -> chunk 0 = "[turn1,turn2]"              hash A

retain #2:  [turn3, turn4], update_mode=append
   existing text is prepended, then the #2409 JSON-array merge
   (orchestrator.py:834) collapses both items into ONE content item
   and re-chunks it -- still fits in a single chunk:
            -> chunk 0 = "[turn1,turn2,turn3,turn4]"  hash B
```

Delta compares A against B, correctly classifies chunk 0 as *changed*, and re-extracts it. When the document is a single not-yet-full chunk, "the changed chunk" and "all chunks" are the same set, so every unit is replaced.

Plain text avoids this only because the merge doesn't apply: old and new stay as separate content items and chunk independently, so the old text re-hashes byte-identically.

## Why we're treating this as working as intended

A chunk whose content changed gets re-extracted — that's the design, and it's what makes delta correct. The important part is that **no information is dropped**: the replacement extraction runs over the *superset* text (old turns + new turns), so the prior turns are still fully in scope. What churns is the memory unit IDs, not the coverage.

That's the gap between the report and what I can measure. The issue claims silent memory *loss* — "the re-extraction produces fewer or different facts". "Different" is expected here. "Fewer" would be a real problem, but it would be an extraction-quality issue on longer accumulated transcripts, not the document-tracking bug described. I wasn't able to reproduce actual loss of coverage.

@oldnicke — could you confirm what you actually observed? Specifically: for a document where old units disappeared, did the *facts* they encoded also disappear, or were they re-created under new IDs from the re-extracted superset? If facts genuinely vanish as the transcript grows, that's worth its own issue with the before/after `memory_units.text` values, and it'd point somewhere quite different from this patch.

## On the patch itself

Beyond the above, taking append out of the delta path (`is_first_batch and update_mode != "append"`) and blind-appending new chunks past the existing ones removes the idempotency guard. Delta is what makes a resent or overlapping payload safe. The OpenClaw `retainFullWindow` path deliberately resends `retainOverlapTurns` turns, and #2644 describes a full-transcript resend — under this change those would land as brand-new chunks and duplicate memories. That risks trading a unit-ID churn complaint for genuine duplication in the same integration.

Closing this for now. Genuinely appreciate the effort here, and happy to reopen if the coverage-loss question above turns up something concrete — that'd be a real bug and we'd want it fixed.

Issue #2668, opened 2026-07-11
Anthropic consolidation batch calls truncate at default max_tokens=4096 → unrecorded invalid-JSON retry storms (path not covered by #1002/#2339 tool-use structured output)


hb-cam:
### Summary

The consolidation batch path still requests structured output as **prose JSON** on the Anthropic provider (it never got the tool-use structured output from #1002 / #2339), and the provider's **default `max_tokens=4096`** truncates batch responses mid-string. Each truncated response triggers an invalid-JSON retry of the *identical* request — which truncates again — burning up to 12 full-priced API calls per batch before the batch is skipped. Because token usage is only recorded after a successful parse, **none of this burn appears in the `hindsight_llm_tokens_*` metrics** — it is visible only on the Anthropic bill.

### Environment

- Hindsight v0.8.4 (`ghcr.io/vectorize-io/hindsight:0.8.4`), self-hosted, pg0 mode
- `HINDSIGHT_API_LLM_PROVIDER=anthropic`, consolidation model `claude-sonnet-5`
- `HINDSIGHT_API_CONSOLIDATION_MAX_COMPLETION_TOKENS` unset (so the provider default applies)

### Evidence

Daily count of `"Anthropic returned invalid JSON"` log lines (each ≈ one failed billed attempt), before/after switching the consolidation model to `claude-sonnet-5` on 2026-07-06:

| Day | invalid-JSON lines |
|---|---|
| 2026-07-06 (switch day) | 86 |
| 2026-07-07 | 1,755 |
| 2026-07-08 | 1,803 |
| 2026-07-09 | 1,730 |
| 2026-07-10 | 654 |
| 2026-07-11 (partial) | 760 |

Typical failure (consolidation batch, 3 outer attempts, each wrapping 4 provider attempts):

```
ERROR - hindsight_api.engine.providers.anthropic_llm - Anthropic returned invalid JSON after 4 attempts
WARNING - [CONSOLIDATION] LLM batch call failed (attempt 2/3) for 8 memories [...]:
  Unterminated string starting at: line 9 column 15 (char 594)
ERROR - [CONSOLIDATION] LLM batch call failed after 3 attempts for 8 memories [...], skipping batch.
```

[...]

PR #2669, opened 2026-07-12
Use strict schema for consolidation output

koriyoshi2041:
## Summary
- request provider-native structured output for consolidation batch responses
- keeps the existing consolidation output budget behavior unchanged
- adds a regression check that `strict_schema` is threaded into the LLM call

This narrows one raw-JSON failure mode related to #2668, but it is not a complete fix for truncation at the output budget. The remaining design question is whether consolidation should be strict-by-default, matching extraction/reflect, given the retry tradeoff for smaller local models.

## Tests
- `uv run pytest tests/test_consolidation_retry_budget.py tests/test_anthropic_structured_tool_use.py -q`
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -2298,6 +2298,7 @@ async def _consolidate_batch_with_llm(
                 ],
                 "response_format": response_model,
                 "scope": "consolidation",
+                "strict_schema": True,
             }
             # Only request an explicit output budget when configured. Left unset by default the key is
             # omitted, so each provider keeps its implicit default (backwards compatible). Operators on
--- a/hindsight-api-slim/tests/test_consolidation_retry_budget.py
+++ b/hindsight-api-slim/tests/test_consolidation_retry_budget.py
@@ -10,6 +10,7 @@ from hindsight_api.engine.consolidation.consolidator import _consolidate_batch_w
 @pytest.fixture
 def mock_llm_config():
     llm = AsyncMock()
+    llm._provider_impl = None
     response = MagicMock()
     response.creates = []
     response.updates = []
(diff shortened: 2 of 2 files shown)


benfrank241:
verified this — mechanically clean and low-risk (8/8 pass incl. the new strict_schema-threading assertion; `strict_schema` is provider-aware everywhere, so the flip routes each provider to its native structured-output path with the existing fallbacks). flagging for @nicoloboschi rather than self-merging though, since it's a provider-behavior design call in the LLM layer you own, not a reproduced bug fix:

- it **addresses but doesn't close #2668** — that issue is truncation at `max_tokens=4096`, and a grammar doesn't help a response that gets cut off; the test asserts the param is threaded, there's no fails-before behavioral repro.
- it flips consolidation to **strict-by-default for all providers/banks**, and there's a documented edge in `llm_wrapper.py:834` — "a small model that can't satisfy strict_schema" can spin retry loops — so for some small local models this could add retries rather than remove them.

so whether consolidation *should* be strict-by-default (matching extraction/reflect) is really a design decision, and it sits right next to your own #2540 strict-schema/caching work. no objection from me either way — your call.

koriyoshi2041:
Thanks, that framing makes sense. I updated the PR body so it no longer reads as a full fix/closure for #2668; it only narrows the raw-JSON path, while the truncation-at-budget part remains separate.

I am also fine with closing this if strict-by-default consolidation is not the direction you want for smaller/local models.

nicoloboschi:
Superseded by #2825, which lands this capability as a per-operation flag (`HINDSIGHT_API_LLM_STRICT_SCHEMA_CONSOLIDATION`) instead of a hardcoded `strict_schema=True`.

Two reasons for the change of approach:

1. The premise here doesn't quite hold — extraction and reflect aren't strict-by-default either. Extraction passes no `strict_schema` and so inherits the global flag, exactly as consolidation already did, and reflect has no `strict_schema` reference at all. The scopes already matched.

2. Because `LLMConfig.call` resolved with `strict_schema or get_config().llm_strict_schema`, hardcoding `True` made consolidation the one scope where `HINDSIGHT_API_LLM_STRICT_SCHEMA=false` no longer turns strict schema off — removing the lever for exactly the case flagged in the description here, the retry tradeoff on smaller local models.

#2825 keeps the escape hatch, and fixes that `or` (a per-call `False` was indistinguishable from "unset", silently ignoring any caller opting out) with a regression test. The `strict_schema` threading test from this PR is carried over, plus an assertion that a disabled flag is passed explicitly rather than omitted.

PR #2676, opened 2026-07-13
Avoid slim embedded daemon startup without local ML deps

### LINKED ISSUE #2671: hindsight-embed: sentence-transformers missing from install_requires — daemon silently fails on fresh venv
`sentence-transformers` isn't in `hindsight-embed`'s deps:

```
$ pip show hindsight-embed
Requires: httpx, rich
```

But `hindsight-embed daemon start` in local embedded mode imports it immediately:

```
ImportError: sentence-transformers is required for LocalSTEmbeddings.
Install it with: pip install sentence-transformers
```

On a fresh venv (e.g. after `hermes update` nukes the site-packages), the daemon just exits. No log, no warning — port 9177 stays dead. Took me two days to notice because memory tools just time out instead of erroring.

Happened twice on macOS 26.5 / Python 3.11 / hindsight-embed 0.7.2.

koriyoshi2041:
## Summary
- skip installed/sibling `hindsight-api` binaries for the embedded daemon when default local embeddings or reranker need `sentence_transformers` but it is not importable
- fall back to `uvx hindsight-api@<version>` in that case, preserving the existing slim binary path for external embedding/reranker providers
- add regression coverage for both the default local and external-provider paths

Fixes #2671

## Tests
- `uv run pytest tests/test_embed_manager.py -v`
- `uv run ruff check hindsight_embed`
- `uv run ty check hindsight_embed`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy NO_PROXY="*" uv run pytest tests -v`

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -14,6 +14,8 @@ import subprocess
 import sys
 import sysconfig
 import time
+from collections.abc import Mapping
+from importlib.util import find_spec
 from pathlib import Path
 from typing import IO, Optional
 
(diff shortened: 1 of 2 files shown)


benfrank241 (changes requested):
nice fix — and the right call not to bolt `sentence-transformers` onto `hindsight-embed`'s install_requires (that'd force the whole torch/ST stack on external-provider users). detecting local-ML-needed + not-importable and falling back to `uvx hindsight-api@<version>`, plus the warning log, fixes both halves of #2671 (the silent exit and the missing signal). the two new tests cover the local+missing and external-provider paths well.

one thing blocks it though: **it breaks 5 pre-existing tests in a slim environment** (i.e. exactly the `test-embed` CI env, where `sentence_transformers` isn't installed). the new `_can_use_installed_api_binary(env)` gate makes `_find_api_command` depend on ST being importable, but the old `test_find_api_command_*` tests don't mock that:

```
FAILED test_find_api_command_prefers_installed_binary_over_uvx
FAILED test_find_api_command_target_install_uses_file_relative_fallback
FAILED test_find_api_command_windows_uses_exe_suffix
FAILED test_find_api_command_windows_prefers_gui_interpreter
FAILED test_find_api_command_windows_prefers_scripts_dir_pythonw_for_wrappers
```

e.g. `test_find_api_command_prefers_installed_binary_over_uvx` (L148) asserts the installed binary, but with providers defaulting to `local` and ST absent, `_can_use_installed_api_binary` returns False and it now falls back to `['uvx', 'hindsight-api@0.0.0']`. it passed for you locally because your dev env has `sentence_transformers` installed; CI didn't catch it because `test-embed` is currently *skipping* on this fork.

fix is mechanical — make those 5 tests deterministic w.r.t. the new gate, same way your new tests do: either `monkeypatch.setattr("hindsight_embed.daemon_embed_manager.find_spec", lambda name: object())` (simulate ST present), or set `HINDSIGHT_API_EMBEDDINGS_PROVIDER`/`HINDSIGHT_API_RERANKER_PROVIDER` to a non-local value so `_needs_local_sentence_transformers` is False. once `uv run pytest tests/test_embed_manager.py` is green in a venv *without* sentence-transformers, this is good to go.

koriyoshi2041:
Thanks, that was the right catch. I pushed `8663c31d6` to make those existing `_find_api_command` tests explicit about the local-ML precondition, so they no longer depend on whether `sentence_transformers` happens to be installed locally.

Local verification now passes:

```
uv run pytest hindsight-embed/tests/test_embed_manager.py
# 20 passed

git diff --check origin/main...HEAD
# no output
```

benfrank241 (approved):
ack — verified: in a slim venv (sentence_transformers not importable) the full test_embed_manager.py is 20/20, including the 5 test_find_api_command_* cases that broke before. pinning the local-ML precondition makes them deterministic in the test-embed CI env. superseding my changes-requested. approving + running CI.

PR #2679, opened 2026-07-13
fix(recall): preserve combined graph activation scores

Sanderhoff-alt:
## Summary

Link Expansion gathers graph candidates from three signals: shared entities, semantic links, and causal links. Within each fact type it computes an additive final score:

- entity score: tanh(shared entity count x 0.5)
- semantic score: the strongest semantic-link weight
- causal score: the strongest causal-link weight

Candidates are correctly ranked and limited by that combined score within a fact type. The bug was in the result conversion step: it stored the original score from one database row in activation instead of the final combined score.

For candidates found through entity expansion, activation was normally the untransformed shared-entity count. For semantic-only or causal-only candidates, it was that individual link weight. As a result, activation could omit the other signals and did not necessarily match the score that selected the candidate.

## Ranking reversal example

| Candidate | Raw shared entities | Entity score | Semantic score | Causal score | Correct combined score | Old activation | Correct rank | Old merged rank |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| A | 1 | 0.462 | 0.900 | 0.300 | 1.662 | 1 | 1 | 2 |
| B | 2 | 0.762 | 0.700 | 0.000 | 1.462 | 2 | 2 | 1 |

```mermaid
flowchart LR
    A[Candidate A: entity 1 plus semantic 0.9 plus causal 0.3] --> AS[combined score: 1.662]
    B[Candidate B: entity 2 plus semantic 0.7] --> BS[combined score: 1.462]
    AS --> IR[Per-fact-type ranking: A before B]
    BS --> IR
    A --> AO[old activation: 1]
    B --> BO[old activation: 2]
    AO --> MR[Old cross-type merge: B before A]
    BO --> MR
    IR --> FIX[Fixed activation uses combined score]
    FIX --> CR[Cross-type merge: A before B]
```

Recall combines graph results from all requested fact types and sorts them by activation before applying the per-source cap and Reciprocal Rank Fusion. The old merge could therefore reverse the graph ranking above, allowing B to receive a better graph rank even though Link Expansion had selected A as more relevant.

## Fix

Store the computed additive score map value in activation. The cross-fact-type graph ordering, source cap, RRF input rank, and trace score now use the same final graph score as the per-fact-type candidate selection.

## Regression coverage

Add a small deterministic unit test with the ranking-reversal example above. It verifies that the candidate with the higher combined score remains ahead after results from two fact types are merged. Reverting activation to a raw row score makes this test fail.

## Validation

- uv run pytest tests/test_link_expansion_scoring.py -q
- ./scripts/hooks/lint.sh
- uv run ty check hindsight_api/
- git diff --check

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/search/link_expansion_retrieval.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/link_expansion_retrieval.py
@@ -243,12 +243,16 @@ class LinkExpansionRetriever(GraphRetriever):
         }
 
         sorted_ids = sorted(score_map.keys(), key=lambda x: score_map[x], reverse=True)[:budget]
-        rows = [row_map[fact_id] for fact_id in sorted_ids]
 
         results = []
-        for row in rows:
+        for fact_id in sorted_ids:
+            row = row_map[fact_id]
             result = RetrievalResult.from_db_row(dict(row))
-            result.activation = row["score"]
+            # ``activation`` is used to re-sort graph results after fact types are
+            # combined. It must retain the final additive score rather than the
+            # raw score from one signal, which would otherwise discard the other
+            # signals and make the cross-fact-type order disagree with this order.
+            result.activation = score_map[fact_id]
             results.append(result)
 
         # filter_results_by_tags is a no-op when no filter applies (tags falsy and not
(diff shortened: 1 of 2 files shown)


benfrank241 (approved):
lgtm — reproduced the reversal: link-expansion selects within a fact type by the combined additive score (entity+semantic+causal), but stored `activation = row["score"]` (a single-signal raw value), so when recall merges graph results across fact types and re-sorts by activation, a candidate with a higher combined score could land behind one that just had a bigger single-signal score. the A/B example fails on main (B before A) and passes here. `activation = score_map[fact_id]` is the right value — no KeyError risk since sorted_ids comes from score_map.keys(). deterministic regression test pins it.

approving + kicked off CI.

PR #2681, opened 2026-07-13
fix(recall): scope entity fanout cap by fact type

Sanderhoff-alt:
## Summary

Fix entity expansion so its per-entity fanout cap is scoped to the requested fact type.

## Root cause

The previous query selected the newest N unit_ids for an entity before applying the recall fact-type filter. For example, a world recall could select 200 newer experience facts, filter all of them out afterward, and never consider older linked world facts. Entity expansion could therefore return no world candidates even when matching world facts existed.

## Fix

- Apply the fact-type predicate inside the LATERAL candidate query.
- Apply the per-entity cap only after target-type candidates are identified.
- Keep the existing ordering, scoring, timeout behavior, and fanout cap.
- Apply the same query semantics to PostgreSQL and Oracle.

## Validation

- Added a regression test for both generated CTEs.
- `uv run pytest tests/test_db_abstraction.py -q` (104 passed)
- `./scripts/hooks/lint.sh`
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
@@ -447,6 +447,14 @@ class OracleOps(DataAccessOps):
                     FROM {ue_table} ue_target
                     WHERE ue_target.entity_id = se.entity_id
                       AND ue_target.unit_id != ALL($1::uuid[])
+                      -- Filter before applying the cap: candidates from other fact
+                      -- types must not consume this entity's bounded fan-out.
+                      AND EXISTS (
+                          SELECT 1
+                          FROM {mu_table} mu_target
+                          WHERE mu_target.id = ue_target.unit_id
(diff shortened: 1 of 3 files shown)


benfrank241 (approved):
lgtm — reproduced the bug at the query level: on main the `fact_type = $2` predicate sits in the outer WHERE, *after* the per-entity `LIMIT/FETCH FIRST N` cap, so an entity with N+ newer facts of another type starves the target type — a world recall grabs the newest N experience facts, filters them all out, and never sees older linked world facts. moving the predicate inside the LATERAL (before the cap) fixes it, and it's applied to both PG and Oracle. the parametrized test pins "fact_type filter appears before the limit clause" for both CTEs; fails on main, passes here. 104 pass locally, no regressions.

minor (non-blocking): the EXISTS-in-LATERAL means the index scan can't early-stop at exactly N rows for entities dominated by the wrong type — bounded by the entity's fact count, and correctness wins.

approving + kicked off CI.

PR #2683, opened 2026-07-13
fix(recall): remove unused graph seed inputs

### LINKED ISSUE #2593: Clarify whether graph retrieval should reuse semantic and temporal seeds from recall
## Summary

`LinkExpansionRetriever.retrieve()` accepts optional `semantic_seeds` and `temporal_seeds`, and its comments describe expanding from semantic/temporal seeds. However, the current recall orchestrator passes both as `None`, so graph retrieval finds its own semantic seeds again and does not reuse temporal entry points.

This may be intentional, but the interface and current call path communicate different things.

## Current behavior

In `retrieve_all_fact_types_parallel()`, graph retrieval is called like this:

```python
results, graph_timing = await retriever.retrieve(
    pool=pool,
    query_embedding_str=query_embedding_str,
    bank_id=bank_id,
    fact_type=ft,
    budget=thinking_budget,
    query_text=query_text,
    semantic_seeds=None,
    temporal_seeds=None,
    ...
)
```

Inside `LinkExpansionRetriever.retrieve()`:

- If `semantic_seeds` are provided, it uses them.
- Otherwise, it runs its own embedding search for up to 20 semantic seeds with threshold `0.3`.
- If `temporal_seeds` are provided, it appends them.
- In the current orchestrator path, temporal seeds are never passed.

## Why this matters

This affects how graph retrieval behaves:

- Graph retrieval may duplicate semantic seed work already done by the semantic retrieval arm.
- When a query has a temporal constraint, temporal retrieval may find good time-window entry points, but graph retrieval does not expand from them.
- The public shape of the retriever suggests semantic/temporal seed reuse is supported, but the active call path does not use it.

This is not necessarily a bug. There may be good reasons to keep graph seed selection independent, such as different thresholds, isolation between retrieval arms, or avoiding graph expansion from weaker temporal hits.

## Questions

1. Is it intentional that graph retrieval always finds its own semantic seeds in the main recall path?
2. Should graph retrieval reuse the semantic results already fetched by `retrieve_semantic_bm25_combined()`?
3. Should graph retrieval expand from temporal entry points when the query has a temporal constraint?
4. If the answer is no, should the interface/comments be updated to make that explicit?
5. If the answer is yes, should we pass selected semantic/temporal seeds into `LinkExpansionRetriever.retrieve()` and avoid duplicate seed search?

## Possible direction

A conservative option is just to document the current design: graph retrieval has its own seed selection, and `semantic_seeds` / `temporal_seeds` are currently test/alternate-path hooks.

If reuse is desired, the orchestrator could pass a bounded set of already computed semantic and temporal results into graph retrieval, with separate thresholds to avoid noisy expansion.

Sanderhoff-alt:
## Summary

Closes #2593.

Graph retrieval does not reuse semantic or temporal results from the other recall arms. It deliberately selects its own bounded semantic seed set before expanding memory links. This PR removes an unused internal interface that incorrectly implied callers could provide those seed sets.

## Root cause

`GraphRetriever.retrieve()` and `LinkExpansionRetriever.retrieve()` accepted `semantic_seeds` and `temporal_seeds`, but the only production caller always supplied `None`. `LinkExpansionRetriever` consequently always ran its own semantic seed query. The stale parameters and comments made the implementation look incomplete and raised the question in #2593 of whether recall was failing to reuse already-fetched semantic and temporal candidates.

## Changes

- Remove `semantic_seeds` and `temporal_seeds` from the internal graph-retriever interface and link-expansion implementation.
- Stop passing placeholder `None` values from the recall orchestrator.
- Document the intentional independent-seed design: the semantic and temporal arms use different candidate limits and thresholds, so reusing their outputs would silently change graph-retrieval behavior.
- Update timing and module comments to match the implementation.
- Add a regression assertion that graph retrieval is not called with the removed inputs.

## Validation

- `uv run pytest tests/test_temporal_recall_selection.py -v`
- `uv run ruff check ...`
- `uv run ruff format --check ...`
- `uv run ty check ...`
- `./scripts/hooks/lint.sh`
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/search/graph_retrieval.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/graph_retrieval.py
@@ -40,8 +40,6 @@ class GraphRetriever(ABC):
         fact_type: str,
         budget: int,
         query_text: str | None = None,
-        semantic_seeds: list[RetrievalResult] | None = None,
-        temporal_seeds: list[RetrievalResult] | None = None,
         adjacency=None,  # TypedAdjacency, optional pre-loaded graph
         tags: list[str] | None = None,  # Visibility scope tags for filtering
         tags_match: TagsMatch = "any",  # How to match tags: 'any' (OR) or 'all' (AND)
@@ -59,8 +57,6 @@ class GraphRetriever(ABC):
             fact_type: Fact type to filter ('world', 'experience', 'observation')
             budget: Maximum number of nodes to explore/return
             query_text: Original query text (optional, for some strategies)
-            semantic_seeds: Pre-computed semantic entry points (from semantic retrieval)
-            temporal_seeds: Pre-computed temporal entry points (from temporal retrieval)
             adjacency: Pre-loaded typed adjacency graph (optional)
             tags: Optional list of tags for visibility filtering (OR matching)
 
--- a/hindsight-api-slim/hindsight_api/engine/search/link_expansion_retrieval.py
+++ b/hindsight-api-slim/hindsight_api/engine/search/link_expansion_retrieval.py
@@ -1,8 +1,8 @@
 """
 Link Expansion graph retrieval.
 
-Expands from semantic/temporal seeds through three parallel, first-class signals
-stored in memory_links:
+Selects bounded semantic seeds, then expands through three parallel,
+first-class signals stored in memory_links:
 
 1. Entity links  — query-time self-join through unit_entities. Score = number of distinct
                    shared entities between the seed set and each candidate, computed via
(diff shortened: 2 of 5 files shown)


benfrank241:
verified locally — flagging for @nicoloboschi since this is a cleanup, not a behavior fix. #2593 was really "is recall failing to reuse the semantic/temporal candidates?" and the answer is no: i confirmed the only production caller (`retrieval.py:808`) always passed `semantic_seeds=None, temporal_seeds=None`, so the `if semantic_seeds:` / `if temporal_seeds:` branches were dead and link-expansion already always ran its own `_find_semantic_seeds`. removing the params + branches is behavior-preserving, and the doc note ("graph intentionally picks its own bounded seeds; reusing the other arms' outputs would silently change behavior since they use different limits/thresholds") is a nice clarification.

no correctness concern — it's aligned with our dead-code standards and 5/5 pass. the only reason i'm not self-merging is that it trims an internal API surface based on an interpretation of #2593 rather than fixing a reproduced bug, so it's your call on whether the independent-seed design is what we want documented-and-locked-in vs. left as a future option. lmk and i'll merge.

nicoloboschi (approved):

PR #2684, opened 2026-07-13
Preserve document metadata during append retains

### LINKED ISSUE #2682: [BUG] update_mode=append drops document_metadata / retain_params.metadata on appended documents
## Bug Description

`POST /v1/default/banks/{bank_id}/memories` with a stable `document_id` and item-level `update_mode: "append"` accepts item-level `metadata`. The same metadata is exposed as `document_metadata` / `retain_params.metadata` for one-shot documents, matching the behavior added in #798.

For appended session documents, the document list/detail response can lose that document-level metadata projection. The extracted memory units still keep tags, and the source facts are created, but the document row returns `document_metadata: null` and `retain_params` with only `context`.

This affects long-running agent/session integrations that rely on `document_metadata` in the document list/control plane for provenance such as `source`, `platform`, `session_id`, and `turn_index`.

## Environment

- Hindsight HTTP API: `0.8.4` (`/openapi.json`)
- Deployment: self-hosted/local external HTTP API
- Python client: `hindsight-client 0.6.1`
- Retain mode: async retain with stable session-scoped document IDs and `update_mode: "append"`

## Minimal Reproduction

Send two retain requests to the same document ID, with metadata present on every item:

```json
{
  "items": [
    {
      "content": "first turn",
      "context": "agent conversation",
      "document_id": "metadata-append-repro",
      "metadata": {
        "source": "hermes",
        "platform": "weixin",
        "session_id": "metadata-append-repro",
        "turn_index": "1"
      },
      "tags": ["source:hermes", "scope:local-agent", "session:metadata-append-repro"],
      "update_mode": "append"
    }
  ],
  "async": true
}
```

Then send another append to the same `document_id` with updated `turn_index`:

```json
{
  "items": [
    {
      "content": "second turn",
      "context": "agent conversation",
      "document_id": "metadata-append-repro",
      "metadata": {
        "source": "hermes",
        "platform": "weixin",
        "session_id": "metadata-append-repro",
        "turn_index": "2"
      },
      "tags": ["source:hermes", "scope:local-agent", "session:metadata-append-repro"],
      "update_mode": "append"
    }
  ],
  "async": true
}
```

After processing completes, inspect:

```bash
GET /v1/default/banks/<bank_id>/documents/metadata-append-repro
GET /v1/default/banks/<bank_id>/documents?tags=source:hermes
```

## Expected Behavior

The document response should retain a useful metadata projection for the appended document, for example:

```json
{
  "document_metadata": {
    "source": "hermes",
    "platform": "weixin",
    "session_id": "metadata-append-repro",
    "turn_index": "2"
  },
  "retain_params": {
    "context": "agent conversation",
    "metadata": {
      "source": "hermes",
      "platform": "weixin",
      "session_id": "metadata-append-repro",
      "turn_index": "2"
    }
  }
}
```

Exact merge policy can be latest-wins, first-wins for stable fields, or documented merge semantics. The key point is that item-level metadata sent during append should remain visible through the document API/control plane metadata surface.

## Actual Behavior

For affected append-updated session documents, the document API response looks like:

```json
{
  "retain_params": {
    "context": "Hermes Agent durable memory extraction; keep curated stable facts only and omit secrets/raw logs/transcripts."
  },
  "document_metadata": null,
  "tags": [
    "session:20260713_174351_d77ce18a",
    "scope:local-agent",
    "source:hermes"
  ]
}
```

The corresponding memory units still exist and retain tags, but the document-level metadata projection is blank.

## Additional Evidence

On one self-hosted 0.8.4 bank, read-only sampling of `source:hermes` documents showed a strong split by document update pattern:

```text
total source:hermes docs: 80
created_at == updated_at and document_metadata present: 53
updated_at > created_at and document_metadata blank: 24
updated_at > created_at and document_metadata present: 3
```

For the affected document above:

```json
{
  "id": "20260713_174351_d77ce18a",
  "created_at": "2026-07-13T09:53:54.307781+00:00",
  "updated_at": "2026-07-13T10:04:41.396283+00:00",
  "memory_unit_count": 2,
  "retain_params": {"context": "..."},
  "document_metadata": null,
  "tags": ["session:20260713_174351_d77ce18a", "scope:local-agent", "source:hermes"]
}
```

The client-side batch item did contain metadata before sending:

```json
{
  "items": [
    {
      "content": "...",
      "metadata": {
        "retained_at": "...",
        "message_count": "2",
        "turn_index": "1",
        "source": "hermes",
        "session_id": "diagnose-real-sync-turn",
        "platform": "weixin",
        "chat_type": "dm",
        "agent_identity": "default"
      },
      "context": "...",
      "tags": ["source:hermes", "scope:local-agent", "session:diagnose-real-sync-turn"],
      "update_mode": "append"
    }
  ],
  "document_id": "diagnose-real-sync-turn",
  "retain_async": true
}
```

## Related Issues / PRs

- #798 exposed `document_metadata` from `retain_params.metadata` in the document API/control plane.
- #803 and #680 fixed metadata being dropped from recall responses.
- #1839 discusses document update semantics for body/tags/metadata/retain params.
- #2409 fixed an append-mode conversation document corruption path.
- #2505 and #2664 track other append-mode correctness issues around stable session-scoped document IDs.

## Suggested Fix Direction

In the retain append/document tracking path, preserve or merge item-level `metadata` into the document's `retain_params.metadata` / `document_metadata` projection when appending to an existing document. Add a regression test that:

1. Retains a document with `document_id`, item `metadata`, tags, and `update_mode: "append"`.
2. Appends a second item with the same document ID and metadata.
3. Asserts `GET /documents/{document_id}` and `/documents` list preserve `document_metadata` / `retain_params.metadata` while keeping tags.

koriyoshi2041:
## Summary

Closes #2682.

When `update_mode=append` prepends the existing document body, the synthetic existing-content item kept context/tags but dropped item-level retain params such as `metadata`. Since document APIs project `document_metadata` from `documents.retain_params.metadata`, appended session documents could lose their document-level metadata even though the new retain item supplied it.

This carries the current append item's metadata/event_date/observation_scopes onto the synthetic/merged append document body so the document row keeps the latest retain params.

## Validation

```bash
uv run ruff check hindsight_api/engine/retain/orchestrator.py tests/test_delta_retain_duplicates.py
uv run python -m py_compile hindsight_api/engine/retain/orchestrator.py tests/test_delta_retain_duplicates.py
git diff --check
```

I also added a DB-backed regression test for the `GET /documents/{document_id}` and list-documents metadata projection. Local execution did not reach the test body because the shared embeddings fixture timed out while initializing/downloading `BAAI/bge-small-en-v1.5`; the failure was in setup, before the new assertions ran.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -831,6 +831,12 @@ async def retain_batch(
             first = contents_dicts[0]
             if first.get("context"):
                 existing_content["context"] = first["context"]
+            if first.get("event_date"):
+                existing_content["event_date"] = first["event_date"]
+            if first.get("metadata"):
+                existing_content["metadata"] = first["metadata"]
+            if first.get("observation_scopes") is not None:
+                existing_content["observation_scopes"] = first["observation_scopes"]
             if first.get("tags"):
                 existing_content["tags"] = first["tags"]
             contents_dicts = [existing_content, *contents_dicts]
@@ -852,6 +858,12 @@ async def retain_batch(
                     contents_dicts = [{"content": json.dumps(_merged, ensure_ascii=False)}]
                     if first.get("context"):
                         contents_dicts[0]["context"] = first["context"]
+                    if first.get("event_date"):
+                        contents_dicts[0]["event_date"] = first["event_date"]
+                    if first.get("metadata"):
+                        contents_dicts[0]["metadata"] = first["metadata"]
+                    if first.get("observation_scopes") is not None:
+                        contents_dicts[0]["observation_scopes"] = first["observation_scopes"]
                     if first.get("tags"):
                         contents_dicts[0]["tags"] = first["tags"]
             except (json.JSONDecodeError, ValueError, TypeError):
--- a/hindsight-api-slim/tests/test_delta_retain_duplicates.py
+++ b/hindsight-api-slim/tests/test_delta_retain_duplicates.py
@@ -10,13 +10,11 @@ Verifies that:
 
 import asyncio
 import logging
-import os
 from datetime import datetime, timezone
 
 import pytest
 import pytest_asyncio
 
-from hindsight_api import RequestContext
 from hindsight_api.engine.task_backend import SyncTaskBackend
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


benfrank241 (approved):
lgtm — reproduced the drop: the append path's synthetic first-content item carried only `context`+`tags`, so item `metadata` (and event_date / observation_scopes) never reached the document projection — append-updated docs return `document_metadata: null` while one-shot docs keep it. copying those three into both append sub-paths fixes it, latest-wins as #2682 expected. the regression test runs the real append×2 → get_document path and asserts the latest metadata survives; fails on main, passes here. sensible guards too (truthy for metadata/tags, `is not None` for observation_scopes so an explicit empty list is preserved).

approving + kicked off CI.

PR #2685, opened 2026-07-13
fix(engine): clarify causal link compatibility

### LINKED ISSUE #2590: Clarify causal link taxonomy: retain only writes caused_by while schema/retrieval support other causal types
## Summary

Current retain/consolidation behavior appears to have converged to writing only `caused_by` causal links, while the database schema, retrieval code, and transfer/export compatibility paths still recognize `causes`, `enables`, and `prevents`.

This may be intentional for historical compatibility, but the current split is not obvious from the code and can be confusing when reasoning about graph/temporal retrieval.

## Current behavior observed

- `memory_links.link_type` still allows:
  - `temporal`
  - `semantic`
  - `entity`
  - `causes`
  - `caused_by`
  - `enables`
  - `prevents`
- The current retain extraction schema only permits `relation_type = "caused_by"`.
- `create_causal_links_batch` validates causal relations against `{"caused_by"}` only.
- Consolidation-created `observation` memory units do not create direct `memory_links`; they rely on `source_memory_ids` / `observation_sources` for derived graph traversal.
- Graph and temporal retrieval still read `causes`, `caused_by`, `enables`, and `prevents` if such rows exist.
- Transfer/export still treats `caused_by`, `causes`, `enables`, and `prevents` as causal archive link types.

## Historical context from git history

It looks like earlier code did support broader causal types:

- `2a00df0bc` (`2026-01-07`): LLM schema and causal link creation supported `causes`, `caused_by`, `enables`, `prevents`.
- `49e233cdb` (`2026-01-08`): changed causal extraction to backward-looking relationships to reduce duplicate/invalid causal relations and token use.
- `9db64ecda` (`2026-01-22`): narrowed the retain schema and link validation to `caused_by` only.
- `15ea23d5d` (`2026-03-13`): `hindsight-api-slim` was introduced with the `caused_by`-only retain path.

## Questions

1. Is `caused_by`-only causal writing the intended long-term design?
2. Are `causes`, `enables`, and `prevents` now intended only for historical/import compatibility?
3. Should comments/docs/schema names be updated to make this explicit?
4. Should retrieval boost logic continue distinguishing `causes/caused_by` from `enables/prevents` if new writes only produce `caused_by`?
5. Should the DB check constraint continue allowing all historical causal types, or should that be documented as a compatibility choice?

## Why this matters

This is not necessarily a bug. The current design may be right, especially if historical data must continue to be searchable. The issue is that the write path, read path, schema, and comments currently communicate slightly different causal taxonomies, which makes future changes and debugging harder.

Sanderhoff-alt:
## Summary

Fixes #2590.

This PR makes the causal-link taxonomy explicit: normal retain writes only the canonical backward-looking `caused_by` relationship. Existing `causes`, `enables`, and `prevents` links remain supported as historical data: the database accepts them, recall continues to traverse them, and transfer archives preserve them.

## Root cause

The write path, schema, retrieval code, and transfer code described different causal taxonomies. In particular, export correctly serialized all four causal link types, but import replayed every archived relationship through the normal retain writer. That writer intentionally accepts only `caused_by`, so valid legacy links were logged as invalid and omitted from the imported graph while the import operation still succeeded.

## What changed

- Document the canonical-write versus historical-compatibility boundary in retain, schema, and transfer code.
- Keep normal retain restricted to `caused_by`; it cannot create legacy causal link types.
- Split the link helpers into a canonical retain writer and a transfer-only legacy restore writer.
- Restore `causes`, `enables`, and `prevents` only while importing a transfer archive, preserving each edge type and its source and target facts.
- Keep recall and other existing read paths compatible with all historical causal types.
- Update stale causal taxonomy assertions and add deterministic coverage for the retain schema boundary.

## Compatibility

This does not change the database constraint, migration behavior, or recall scoring. It only restores transfer compatibility for valid historical causal links already supported by those read paths. Cross-document causal links remain outside the existing archive format and are unchanged by this PR.

## Validation

- `uv run pytest -n 0 -q -p no:logging tests/test_causal_link_taxonomy.py tests/test_document_transfer.py::test_transfer_preserves_legacy_causal_links` — 5 passed.
- The new transfer round-trip test inserts all three legacy link types, exports the document, imports it into a new bank, and verifies the imported type, source text, and target text for every edge.
- `uv run ruff format --check` and `uv run ruff check` on changed files.
- `uv run ty check hindsight_api`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/link_creation.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/link_creation.py
@@ -74,7 +74,9 @@ async def create_causal_links_batch(
     """
     Create causal links between facts.
 
-    Links facts that have causal relationships (causes, enables, prevents).
+    Retain writes the canonical ``caused_by`` relationship only. The database and
+    retrieval paths also recognize historical causal types so imported and
+    pre-existing memories remain traversable.
 
     Args:
         conn: Database connection
@@ -90,22 +92,7 @@ async def create_causal_links_batch(
     if len(unit_ids) != len(facts):
         raise ValueError(f"Mismatch between unit_ids ({len(unit_ids)}) and facts ({len(facts)})")
 
-    # Extract causal relations in the format expected by link_utils
-    # Format: List of lists, where each inner list is the causal relations for that fact
-    causal_relations_per_fact = []
-    for fact in facts:
-        if fact.causal_relations:
-            # Convert CausalRelation objects to dicts
-            relations_dicts = [
-                {
-                    "relation_type": rel.relation_type,
-                    "target_fact_index": rel.target_fact_index,
-                }
-                for rel in fact.causal_relations
-            ]
-            causal_relations_per_fact.append(relations_dicts)
-        else:
-            causal_relations_per_fact.append([])
+    causal_relations_per_fact = [fact.causal_relations or [] for fact in facts]
 
     link_count = await link_utils.create_causal_links_batch(conn, bank_id, unit_ids, causal_relations_per_fact, ops=ops)
 
--- a/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
@@ -7,12 +7,17 @@ import time
 from datetime import UTC, datetime, timedelta
 
 from ..._vector_index import ann_search_tuning_settings, configured_vector_extension
+from ..db.base import DatabaseConnection
+from ..db.ops import DataAccessOps
 from ..memory_engine import fq_table
+from .types import CausalRelation
 
 logger = logging.getLogger(__name__)
 
 # Sentinel UUID used in the unique index to represent NULL entity_id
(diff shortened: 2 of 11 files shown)


benfrank241 (approved):
lgtm — reproduced the drop: on main a transfer round-trip imports legacy `causes`/`enables`/`prevents` edges as an empty set (they route through the retain writer, which only accepts `caused_by`), so valid archived links silently vanish on import while the op still succeeds. splitting into a canonical retain writer (caused_by only) and a transfer-only `restore_legacy_causal_links_batch` fixes it cleanly — no new hole in retain, each type created exactly once on import (`_to_extracted_fact` keeps caused_by, `_legacy_causal_relations` handles the rest). the round-trip test inserts all three legacy types, exports, imports into a fresh bank, and verifies every edge's type/source/target — fails on main, passes here. nice typed-CausalRelation cleanup too, and no DB constraint / migration / recall-scoring changes.

nit (non-blocking): `_LEGACY_CAUSAL_LINK_TYPES` is defined in both link_utils.py and importer.py — could dedupe to one.

approving + kicked off CI.

PR #2691, opened 2026-07-13
fix(engine): consolidate duplicated engine definitions

Sanderhoff-alt:
## Summary

- Centralize causal link taxonomy used by retain and transfer.
- Share transfer history table classifications between export and import.
- Use the shared canonical type in importer filtering.
- Remove the unused wrapper-level Groq seed constant.

## Root cause

Several engine modules independently defined the same causal and transfer
metadata. Import also hard-coded the canonical causal type, so a taxonomy
change could silently cause canonical links to be discarded.

## Validation

- `ruff format --check hindsight_api/engine/transfer/importer.py`
- `ruff check hindsight_api/engine/transfer/importer.py`
- `pytest -q -n0 tests/test_causal_link_taxonomy.py`
- `pytest -q -n0 tests/test_document_transfer.py::test_export_bank_covers_schema`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/causal_links.py
+++ b/hindsight-api-slim/hindsight_api/engine/causal_links.py
@@ -1,0 +1,13 @@
+"""Shared causal-link taxonomy.
+
+Retain writes only the canonical relationship.  Transfer import/export also
+preserves historical relationship types so existing banks keep their graph
+semantics without allowing new retain output to create those types.
+"""
+
+CANONICAL_CAUSAL_LINK_TYPE = "caused_by"
+LEGACY_CAUSAL_LINK_TYPE_NAMES = ("causes", "enables", "prevents")
+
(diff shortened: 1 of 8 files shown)


benfrank241:
verified this — clean and provably behavior-preserving (24/24 across test_causal_link_taxonomy + test_document_transfer, rebased on #2685, shared constants carry the same values). and it nicely closes the exact `_LEGACY_CAUSAL_LINK_TYPES`-duplicated-in-two-files nit i raised on #2685. flagging for @nicoloboschi rather than self-merging though, since it's a refactor, not a reproduced bug fix:

- nothing misbehaves today — the "root cause" is a *could-drift* maintainability concern; the duplicated definitions currently agree, so behavior is unchanged.
- it introduces a new shared `engine/causal_links.py` and rewires the importer/export/schema onto it — that's a home + API-shape call in the causal/transfer path you just had merged (#2685), worth your eyes.

no correctness objection from me — the dedup is the right direction and lines up with our DRY/dead-code standards. your call on the module placement / whether to take it now.

nicoloboschi (approved):

PR #2697, opened 2026-07-13
fix(retain): reassert resolved entities before linking units

### LINKED ISSUE #2662: FK race: prune_orphan_entities deletes entities between retain Phase-1 resolve and Phase-2 unit_entities insert — silent memory loss on re-ingest (v0.8.4)
## Summary

In v0.8.4 (commit `92f433c9`, Postgres backend / asyncpg), the retain pipeline resolves entities (Phase 1) and inserts the `unit_entities` child rows (Phase 2) in **separate transactions on separate connections**. In the window between the two phases, the graph-maintenance task's `prune_orphan_entities` can delete an `entities` row that Phase 1 already resolved but Phase 2 has not yet referenced — at that instant the entity legitimately has no `unit_entities` row, so it matches the pruner's orphan predicate. When Phase 2 then inserts the `unit_entities` row, it violates the FK:

```
insert or update on table "unit_entities" violates foreign key constraint "fk_unit_entities_entity_id_entities"
DETAIL: Key (entity_id)=(<uuid>) is not present in table "entities".
```

The error is classified as a **deterministic (non-retryable) failure**, so the entire `batch_retain` is dropped rather than retried — **the memories in that batch are silently lost**.

## Interleave (reproduced deterministically, 15/15, on the real orchestrator path)

1. Retain Phase 1 (`_pre_resolve_phase1` → `entity_processing.resolve_entities`, on its own `resolve_conn`) resolves/creates entity `E`, returns its id in `resolved_entity_ids`, commits and closes.
2. Before Phase 2 runs, `prune_orphan_entities` executes for the same bank. `E` has no `unit_entities` row yet, so the `NOT EXISTS` predicate matches and `E` is `DELETE`d.
3. Retain Phase 2 (`_insert_facts_and_links`, transactional `conn`) → `link_units_to_entities_batch` → `bulk_insert_unit_entities` inserts `(unit_id, E.id)` and raises `ForeignKeyViolationError`.
4. Caught as deterministic/non-retryable → whole `batch_retain` discarded → silent memory loss.

Most visible on the **document re-ingest / re-consolidation path**: re-ingesting a stable `document_id` invalidates the prior `memory_units`, orphaning their entities; the next maintenance sweep is then eligible to prune exactly the rows a concurrent retain is about to re-reference.

## Root-cause locations (verified against running v0.8.4)

- `engine/retain/orchestrator.py` — Phase-1 `_pre_resolve_phase1` acquires its own `resolve_conn`; Phase-2 `_insert_facts_and_links` builds `unit_entity_pairs` and calls `link_units_to_entities_batch(..., conn=conn)`.
- `engine/entity_resolver.py:785` — `link_units_to_entities_batch` → `_link_units_to_entities_batch_impl` → `bulk_insert_unit_entities`.
- `engine/db/ops_postgresql.py:325` — `bulk_insert_unit_entities` runs `INSERT ... ON CONFLICT DO NOTHING` (which does not suppress FK violations).
- `engine/db/ops_postgresql.py:393` — `prune_orphan_entities`: `DELETE FROM entities e WHERE e.bank_id=$1 AND NOT EXISTS (SELECT 1 FROM unit_entities ue WHERE ue.entity_id=e.id)`; caller `graph_maintenance.py:153` `run_graph_maintenance_job` (prune ~L210).
- FK `fk_unit_entities_entity_id_entities` (`entity_id → entities(id)`, `ON DELETE CASCADE`) from `alembic/versions/5a366d414dce_initial_schema.py`.

Because Phase-1 resolve and Phase-2 insert are on different connections/transactions, no row lock or snapshot spans them, so nothing prevents the pruner from deleting `E` in between.

## Suggested fix (proven end-to-end on the real retain orchestrator path)

Same-transaction re-assert of the referenced entities immediately before the child insert: run an idempotent `INSERT INTO entities (id, bank_id, canonical_name) ... ON CONFLICT DO NOTHING` **on the Phase-2 connection/transaction** for the entities about to be linked, threading the minimal entity metadata (`entity_id`, `bank_id`, `canonical_name` — the name is NOT NULL and unrecoverable once the row is pruned, so it must be carried through) from the Phase-1 result. An entity referenced by a live unit is by definition not an orphan, so resurrecting it is semantically correct, and doing it in the same transaction as `bulk_insert_unit_entities` closes the window regardless of pruner timing. With this in place the 15/15-red repro goes 0/15 with all memories persisted.

An alternative — serializing `prune_orphan_entities` against in-flight retains — also works but adds contention.

## Related prior art

Sibling FK races under concurrent retain were fixed before (#1795, #1882 via #1930; cascade-deadlock FK deferral in #1398), but those cover `memory_links`/`memory_units`; the `prune_orphan_entities` vs `unit_entities` window is distinct and still present in v0.8.4.

## Environment

- Hindsight v0.8.4 (`92f433c9`), Postgres backend (`ops_postgresql`), asyncpg
- Concurrent workload: retain traffic alongside `run_graph_maintenance_job` → `prune_orphan_entities`

r266-tech:
## Summary

- carry each resolved entity's ID and stored canonical name across the retain phase boundary
- in the Phase-2 write transaction, lock parents that still exist and idempotently recreate parents pruned after Phase 1
- implement the reassert operation for PostgreSQL and Oracle before inserting `unit_entities`

`ON CONFLICT DO NOTHING` alone does not lock an existing PostgreSQL row, so the PostgreSQL path first takes `FOR KEY SHARE` locks in stable ID order. The Oracle path uses its `FOR UPDATE` equivalent. This keeps orphan pruning from deleting a parent in the small interval before the child FK insert, while still restoring a parent that was already pruned.

## Tests

- added a real PostgreSQL concurrency regression proving prune blocks until the child link is inserted
- added an end-to-end Phase-1 → prune → Phase-2 regression using a fuzzy alias, verifying the original ID and canonical name are restored
- added Oracle adapter coverage for stable lock order and idempotent reinsert data
- 165 related resolver, DB abstraction, retain, curation, and transfer tests pass
- entity-label resolution regression passes
- Ruff and `ty check hindsight_api` pass

Fixes #2662

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops.py
@@ -173,6 +173,22 @@ class DataAccessOps(ABC):
         ...
 
     @abstractmethod
+    async def bulk_reassert_entities(
+        self,
+        conn: DatabaseConnection,
+        table: str,
+        bank_id: str,
+        entity_ids: list[str],
+        canonical_names: list[str],
+    ) -> None:
+        """Protect resolved parents and reinsert any that disappeared before linking.
+
+        Existing rows are locked in deterministic ID order; missing rows are
+        inserted idempotently in the same transaction as unit_entities.
+        """
+        ...
+
+    @abstractmethod
     async def bulk_insert_unit_entities(
         self,
         conn: DatabaseConnection,
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
@@ -216,7 +216,7 @@ class OracleOps(DataAccessOps):
         for orig_name in missing_names:
             row = await conn.fetchrow(
                 f"""
-                SELECT id, LOWER(canonical_name) AS name_lower
+                SELECT id, canonical_name, LOWER(canonical_name) AS name_lower, $2 AS input_name
                 FROM {table}
                 WHERE bank_id = $1 AND LOWER(canonical_name) = LOWER($2)
                 """,
@@ -224,9 +224,32 @@ class OracleOps(DataAccessOps):
                 orig_name,
             )
             if row:
-                # Wrap in a dict-like to include input_name for downstream compat
                 results.append(row)
         return results
+
+    async def bulk_reassert_entities(
+        self,
+        conn: DatabaseConnection,
+        table: str,
+        bank_id: str,
+        entity_ids: list[str],
+        canonical_names: list[str],
+    ) -> None:
+        # Oracle has no FOR KEY SHARE. Lock existing parents in the same stable
+        # order; IDs already pruned are absent here and recreated below.
+        for entity_id in entity_ids:
+            await conn.fetchrow(
+                f"SELECT id FROM {table} WHERE id = $1 FOR UPDATE",
+                entity_id,
+            )
+        await conn.executemany(
+            f"""
+            INSERT INTO {table} (id, bank_id, canonical_name)
+            VALUES ($1, $2, $3)
+            ON CONFLICT DO NOTHING
+            """,
+            [(entity_id, bank_id, canonical_name) for entity_id, canonical_name in zip(entity_ids, canonical_names)],
+        )
 
     async def bulk_insert_unit_entities(
         self,
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -275,7 +275,7 @@ class PostgreSQLOps(DataAccessOps):
     ) -> list[ResultRow]:
         return await conn.fetch(
             f"""
-            SELECT e.id, LOWER(e.canonical_name) AS name_lower, inputs.input_name
+            SELECT e.id, e.canonical_name, LOWER(e.canonical_name) AS name_lower, inputs.input_name
             FROM {table} e
             JOIN (
                 SELECT LOWER(n) AS input_name_lower, n AS input_name
@@ -285,6 +285,39 @@ class PostgreSQLOps(DataAccessOps):
             """,
             bank_id,
             missing_names,
+        )
+
+    async def bulk_reassert_entities(
+        self,
+        conn: DatabaseConnection,
+        table: str,
+        bank_id: str,
+        entity_ids: list[str],
+        canonical_names: list[str],
+    ) -> None:
+        # ON CONFLICT DO NOTHING does not lock an existing row. Lock first so
+        # orphan pruning cannot delete a parent between this method and the FK
+        # insert; IDs absent at this snapshot are recreated below.
+        await conn.fetch(
+            f"""
+            SELECT id
+            FROM {table}
+            WHERE id = ANY($1::uuid[])
+            ORDER BY id
+            FOR KEY SHARE
+            """,
+            entity_ids,
+        )
+        await conn.execute(
+            f"""
+            INSERT INTO {table} (id, bank_id, canonical_name)
+            SELECT entity_id, $1, canonical_name
+            FROM unnest($2::uuid[], $3::text[]) AS t(entity_id, canonical_name)
+            ON CONFLICT DO NOTHING
+            """,
+            bank_id,
+            entity_ids,
+            canonical_names,
         )
 
     async def bulk_insert_unit_entities(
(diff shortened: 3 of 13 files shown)


benfrank241:
verified this locally and confirming it's a real, serious bug — the Phase-1/Phase-2 entity FK race is genuine silent memory loss on re-ingest, and the fix direction is right (carry the resolved id+canonical name across the phase boundary, lock existing parents `FOR KEY SHARE` in stable id order, idempotently recreate a parent pruned after Phase-1). the new regressions pass on the branch: `test_reassert_locks_existing_parent_until_child_insert` (real PG concurrency — prune blocks until the child link inserts) and the end-to-end Phase1→prune→Phase2 both green (3/3); the only local errors were missing-LLM-key env, not regressions.

flagging for @nicoloboschi to own the merge rather than self-merging, for blast-radius reasons — this is the same call we made on #2646 (authz) and #2666 (append):

- it's ~400 lines of **concurrency/lock-ordering in the core retain/entity hot path** — `FOR KEY SHARE` acquisition + stable lock order + idempotent parent recreation. a subtle miss (lock order, incomplete coverage, or a **deadlock** against the pruner / consolidation sweep that also touches `entities`/`unit_entities`) could be worse than the bug it fixes.
- it's **dual-dialect** (PG `FOR KEY SHARE` + Oracle `FOR UPDATE` reassert) and threads a new `ResolvedEntity` type through ~10 files including the db-ops layer.

worth your read specifically on the lock-ordering vs. the graph-maintenance prune path (`prune_orphan_entities` / `prune_stale_cooccurrences`) to confirm there's no lock-cycle, and on the reassert's idempotent-recreate semantics matching what create-time does. i couldn't get a clean fails-before on `main` since the test imports the new `ResolvedEntity` type — the reporter's 15/15 deterministic repro is the reference there.

benfrank241:
Thanks @r266-tech — this is a well-targeted fix for a real, nasty bug (#2662: silent memory loss on re-ingest when `prune_orphan_entities` deletes a resolved entity in the Phase-1→Phase-2 window and the `unit_entities` FK insert then fails as a non-retryable error).

**Verified locally** (embedded pg0 + local-ml):

- `tests/test_retain_entity_prune_race.py` **3/3 pass**, incl. the real-Postgres contention test (`test_reassert_locks_existing_parent_until_child_insert` — a concurrent parent `DELETE` is genuinely blocked by the `FOR KEY SHARE` lock until the child link commits) and the e2e Phase1→prune→Phase2 regression (entity restored with original ID + stored canonical name).
- `tests/test_entity_resolver.py` + `tests/test_entity_labels.py` — all pass (99 passed total).
- `ruff check hindsight_api/` clean; no new lint.

**Correctness read** (for reviewer head start):
- Lock ordering is stable — entities deduped and sorted by `str(entity_id)`; PG does `... WHERE id = ANY($1) ORDER BY id FOR KEY SHARE`, Oracle loops `FOR UPDATE` over the pre-sorted list. Consistent across dialects, so it doesn't introduce a new deadlock.
- Reinsert is idempotent (`ON CONFLICT DO NOTHING`; Oracle rewrite layer maps this to PK + ORA-00001 suppression, consistent with the rest of `OracleOps`).
- Reassert + link run on the same Phase-2 `conn`, so lock/reinsert and the FK insert are atomic.

**Minor notes worth a look, not blockers:**
1. Freshly *reinserted* (previously-pruned) parents aren't covered by the reassert's lock scan (they were absent from the `id = ANY / FOR UPDATE` snapshot) — the window is closed instead by the `FOR KEY SHARE` the `unit_entities` FK insert itself takes within the same tx. Correct, just subtler than the docstring implies.
2. Oracle uses `FOR UPDATE` (no `FOR KEY SHARE` equivalent) — a stronger lock that also blocks concurrent child inserts on the same parent. Intentional and documented; small concurrency cost, not a correctness issue.

Because this touches the core retain concurrency path across both PG and Oracle DB ops with new lock semantics, I'm deferring the merge decision to @nicoloboschi rather than self-merging — the reproduction is solid and the logic looks sound, so this should be a fast review for him.

nicoloboschi:
Superseded by #2859, which consolidates the best of this PR and the sibling into one fix: it carries the **stored canonical name** across the phase boundary (as here) *and* locks surviving parents `FOR KEY SHARE` so the pruner blocks (closing the residual lock-free window), folds lock+reassert into a single statement matching the existing `bulk_insert_links` convention, and also covers the curation edit path. Closing in favor of #2859 — thanks for the diagnosis and repro, they carried straight over. Reopen if you'd like to keep iterating here instead.

PR #2703, opened 2026-07-14
fix(claude-code): surface the CLI's real error text on is_error results

### LINKED ISSUE #2702: claude-code provider: quota/429 failures surface as "Claude Code returned an error result: success", hiding the real error
## Summary

When the Claude Code CLI reports a usage-limit failure (HTTP 429), every retain/consolidation LLM call fails with the message:

```
Claude Code returned an error result: success
```

"success" is not an error. The real failure is completely hidden, which makes a stalled extraction pipeline very hard to diagnose — in our case ~1,100 `retain` operations burned through their retries and were marked `failed` while the logs said "success" over and over.

## Root cause

The CLI emits a result message with `is_error: true` but keeps `subtype: "success"`, and puts the actual failure detail in `result`:

```json
{"type":"result","subtype":"success","is_error":true,"api_error_status":429,
 "result":"You've hit your weekly limit · resets Jul 18, 12pm (UTC)", ...}
```

The Python Agent SDK builds its fallback error text from `errors` (empty here) and then falls back to `subtype` (`claude_agent_sdk/_internal/query.py`, `_last_error_result_text`), so the exception that reaches `claude_code_llm.py` reads "…: success".

`ClaudeCodeLLM.call()` / `call_with_tools()` only inspect `AssistantMessage` in their streaming loops and ignore `ResultMessage` entirely, so hindsight never sees the CLI's actual error text (`message.result`) even though the SDK yields it before the stream errors out.

## Impact

- Worker logs and `async_operations.error_message` show `Fact extraction failed: ... Claude Code returned an error result: success` for what is actually quota exhaustion.
- Operators can't tell rate-limit exhaustion from a genuine provider bug without exec'ing into the pod and running the CLI by hand.

## Fix

In `ClaudeCodeLLM`'s streaming loops, check `ResultMessage.is_error` and raise with `message.result` (falling back to `subtype`) so the actionable text — e.g. `You've hit your weekly limit · resets Jul 18, 12pm (UTC)` — reaches logs and operation error messages. PR incoming.

Observed on hindsight 0.7.x, `claude-agent-sdk` 0.2.x, provider `claude-code`, models claude-sonnet-5/claude-haiku-4-5.

🤖 Generated with Claude Code - (fd886cc8-b9ce-4333-8e0b-8bda96598269)

ElanHasson:
Fixes #2702

## Problem

When the Claude Code CLI fails with a usage-limit 429, it emits a result message with `is_error: true` but `subtype: "success"`, carrying the real detail in `result`:

```json
{"type":"result","subtype":"success","is_error":true,"api_error_status":429,
 "result":"You've hit your weekly limit · resets Jul 18, 12pm (UTC)"}
```

The Agent SDK's fallback exception is built from the (empty) `errors` array or the `subtype`, so every retain/consolidation failure logs the useless:

```
Claude Code returned an error result: success
```

In our deployment ~1,100 `retain` ops burned their retries and were marked `failed` with that message while the actual cause (quota exhaustion) was invisible without exec'ing into the worker pod and running the CLI by hand.

## Fix

`ClaudeCodeLLM.call()` and `call_with_tools()` now check `ResultMessage.is_error` in their streaming loops and raise with the CLI's actual `result` text (falling back to `subtype`), before the SDK's opaque wrapper exception fires:

```
Claude Code reported an error: You've hit your weekly limit · resets Jul 18, 12pm (UTC)
```

Retry/backoff behavior is unchanged — only the exception text is fixed, so the actionable message now reaches worker logs and `async_operations.error_message`.

## Testing

- 4 new tests in `tests/test_claude_code_llm_error_result.py` (both call paths raise with the result text; subtype fallback when `result` is empty; `is_error=false` result messages remain ignored), following the mock style of `test_claude_code_llm_isolation.py`
- `uv run pytest tests/test_claude_code_llm_error_result.py tests/test_claude_code_llm_isolation.py` — 8 passed
- `uv run ruff check` / `ruff format --check` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code) - (fd886cc8-b9ce-4333-8e0b-8bda96598269)

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/claude_code_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/claude_code_llm.py
@@ -49,6 +49,20 @@ def _get_isolated_claude_env() -> dict[str, str]:
     return _isolated_claude_env
 
 
+def _result_error_detail(message: Any) -> str:
+    """Build an actionable error string from an ``is_error`` ResultMessage.
+
+    The CLI can report a failure with ``is_error=True`` while ``subtype``
+    still reads ``"success"``, putting the real detail in ``result`` (e.g.
+    quota exhaustion: ``You've hit your weekly limit · resets ...`` with
+    ``api_error_status: 429``). The SDK's own fallback exception surfaces
+    only the subtype, producing the misleading "Claude Code returned an
+    error result: success" (issue #2702) — so prefer ``result``.
+    """
+    detail = (message.result or "").strip() or message.subtype or "unknown error"
+    return f"Claude Code reported an error: {detail}"
+
+
 class ClaudeCodeLLM(LLMInterface):
     """
     LLM provider using Claude Code authentication.
@@ -176,6 +190,7 @@ class ClaudeCodeLLM(LLMInterface):
         from claude_agent_sdk import (  # type: ignore[unresolved-import]
             AssistantMessage,
             ClaudeAgentOptions,
+            ResultMessage,
             TextBlock,
             query,
         )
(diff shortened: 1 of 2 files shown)


benfrank241 (approved):
Root cause confirmed: `ClaudeCodeLLM`'s streaming loops only checked `AssistantMessage`; they never inspected `ResultMessage` at all. When the CLI emits `{is_error:true, subtype:"success", result:"...",}`, the SDK's own fallback builds the exception text from `subtype` — "error result: success" — which is actively misleading.

The fix is minimal and correct: a `_result_error_detail()` helper preferring `message.result` over `subtype`, added at both streaming-loop `ResultMessage` branches. Verified locally: 4/4 tests pass — quota-exhaustion scenario, empty-result subtype fallback, and non-error ResultMessage pass-through. All correct.

PR #2706, opened 2026-07-14
Fix Codex OAuth refresh sharing across managers

### LINKED ISSUE #2704: openai-codex OAuth refresh is not single-flight across LLM instances
## Bug description

A long-running Hindsight process can create several `CodexLLM` instances for the default, retain, reflect, and consolidation LLM configurations even when they all resolve to the same `openai-codex` provider and the same `CODEX_HOME`.

Each `CodexLLM` constructs its own `CodexAuthManager`, caches the access/refresh tokens in memory, and owns an instance-local `threading.Lock` / `asyncio.Lock`. That makes refresh single-flight only inside one manager, not across all managers sharing the same `auth.json`.

Because Codex refresh tokens rotate, one manager can refresh successfully and atomically persist the new credentials while sibling managers keep the old in-memory refresh token. When a sibling later refreshes, the OAuth endpoint returns `refresh_token_reused`. The sibling does not re-read the newer credentials already on disk before raising `CodexRefreshExpiredError`, so that LLM path stays broken until its provider/process is recreated.

## Observed behavior

Observed on Hindsight v0.8.4 in a long-running API/worker process:

- four Codex OAuth credential loads during process initialization;
- successful token rotation persisted by one or more managers;
- later `refresh_token_reused` failures from another manager;
- the reflect endpoint returned HTTP 500 with:

```text
Codex refresh_token is permanently invalid (error.code=refresh_token_reused).
```

At that point the shared `auth.json` already contained a newer access token with a future JWT expiry, but the failing manager continued using its stale in-memory credential set. Other paths could remain operational, which makes this a partial/deceptive failure rather than a clean provider-wide outage.

## Root cause

`MemoryEngine` builds separate LLM configurations for default, retain, reflect, and consolidation. Each resulting `CodexLLM` creates a separate `CodexAuthManager`.

`CodexAuthManager.refresh_tokens()` serializes only on `self._lock`. After acquiring that lock it checks only its own cached access token and refresh token. It re-reads `auth.json` when persisting a successful rotation, but not before attempting refresh and not after a terminal `refresh_token_reused` / `refresh_token_invalidated` response.

The same class is also shared with Codex OAuth embeddings, so the ownership contract should be defined per canonical auth-file path rather than per provider object.

## Reproduction

A hermetic regression can use two `CodexAuthManager` instances initialized from one temporary `auth.json` and a mock refresh endpoint with rotating, single-use refresh tokens:

1. Both managers load access token A1 and refresh token R1.
2. Manager A refreshes R1, receives A2/R2, and persists them.
3. Manager B still holds R1 in memory.
4. Manager B attempts refresh and receives `refresh_token_reused`.
5. The file contains A2/R2, but manager B raises instead of adopting them.

The same test should be repeated with concurrent refresh attempts.

## Expected behavior

All provider instances that share one canonical Codex auth file should behave as one refresh domain:

- refresh is serialized across instances (and across processes if separate workers can share the same `CODEX_HOME`);
- after acquiring the refresh lock, the manager re-reads `auth.json` and adopts a newer valid credential set before making a network refresh call;
- on a terminal refresh-token response, it re-reads the file once and retries only when the on-disk credentials differ from the failed in-memory credentials;
- if the file has not changed, the existing clear re-login error is preserved;
- token values are never logged, and atomic `0600` persistence remains intact.

## Suggested acceptance tests

- Two managers sharing one auth file produce at most one refresh network call for the same expired credential generation.
- A sibling manager adopts credentials rotated by another manager without using the old refresh token.
- An external `codex auth login` / atomic auth-file replacement is adopted by a long-running manager.
- A genuinely invalid token with unchanged on-disk credentials still fails once with the existing remediation message and does not loop.
- `call()`, `call_with_tools()`, and Codex OAuth embeddings use the same path-scoped refresh contract.
- A multi-process test covers the worker topology if multiple processes can share one `CODEX_HOME`.

## Version

- Reproduced on v0.8.4.
- Current `main` still uses instance-local cached credentials and locks for this path.

koriyoshi2041:
Fixes #2704.

This makes Codex OAuth refresh coordinate per canonical auth file instead of per `CodexAuthManager` instance. A refresh now:

- takes an in-process path-scoped lock plus an advisory auth-file lock;
- re-reads `auth.json` under the lock and adopts a newer token set before using an old refresh token;
- re-checks `auth.json` once before surfacing terminal refresh-token errors;
- keeps the existing atomic persistence path and token-redaction behavior.

Tests added for stale sibling managers adopting rotated credentials and parallel managers sharing one refresh for the same auth file.

Validation:

```
env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy -u CODEX_HOME uv run pytest hindsight-api-slim/tests/test_codex_oauth_refresh.py
# 27 passed

env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy -u CODEX_HOME uv run pytest hindsight-api-slim/tests/test_codex_home_env.py
# 6 passed

git diff --check
# no output
```

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/codex_auth.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/codex_auth.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import base64
 import binascii
+import contextlib
 import json
 import logging
 import os
@@ -30,6 +31,11 @@ from pathlib import Path
 from typing import Any
 
 import httpx
+
+try:
+    import fcntl
+except ImportError:  # pragma: no cover - Windows
+    fcntl = None  # type: ignore[assignment]
 
 logger = logging.getLogger(__name__)
 
@@ -58,6 +64,9 @@ _CODEX_TOKEN_REFRESH_SKEW_SECONDS = 60
 _CODEX_TERMINAL_REFRESH_ERROR_CODES = frozenset(
     {"refresh_token_expired", "refresh_token_reused", "refresh_token_invalidated"}
 )
+_CODEX_AUTH_LOCK_TIMEOUT_SECONDS = 20.0
+_CODEX_AUTH_LOCKS_GUARD = threading.Lock()
+_CODEX_AUTH_LOCKS: dict[Path, threading.Lock] = {}
 
 
 def default_codex_auth_file() -> Path:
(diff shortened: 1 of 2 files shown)


benfrank241 (approved):
Root cause confirmed: each `CodexLLM` creates its own `CodexAuthManager` with an instance-local `asyncio.Lock`, so refresh is only single-flight within one manager. Multiple LLM configs resolving to the same `openai-codex` provider can race to rotate the same refresh token — first succeeds, siblings get `refresh_token_reused`.

The fix is well-scoped: path-scoped in-process lock + `fcntl` advisory file lock for cross-process coordination + a pre-read-under-lock (`_adopt_tokens`) that adopts newer on-disk credentials before attempting a network refresh. Terminal error path does one re-read before surfacing the existing remediation message. Windows/no-fcntl degrades gracefully.

All 27 `test_codex_oauth_refresh` tests pass including the new sibling-adoption and parallel-managers regression tests. Ruff clean. CI green.

PR #2707, opened 2026-07-14
docs(embed): expose local CPU workarounds

yingliang-zhang:
## Summary

- document the existing `HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU` setting in both self-documenting env templates
- document the existing `HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU` setting in both templates
- add regression coverage that keeps the root and bundled templates synchronized and preserves the commented/default-false contract

## Why

These settings are already supported by the embedded runtime and documented in the configuration reference, but they are absent from the env templates users discover during profile setup. Exposing them there makes the existing macOS MPS/XPC workaround discoverable without changing defaults for other platforms.

Both settings remain commented out and default to `false`.

## Verification

- `uv run --project hindsight-embed pytest hindsight-embed/tests/test_env_template.py hindsight-embed/tests/test_profile_manager.py hindsight-embed/tests/test_embed_manager.py hindsight-embed/tests/test_profile_daemon_config.py -q` — 70 passed
- `uv run --project hindsight-embed ruff check hindsight-embed/tests/test_env_template.py`
- `uv run --project hindsight-embed ruff format --check hindsight-embed/tests/test_env_template.py`
- root and bundled env templates are byte-identical

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -147,6 +147,8 @@ HINDSIGHT_API_LOG_LEVEL=info
 # HINDSIGHT_API_EMBEDDINGS_PROVIDER=local
 # For local provider:
 # HINDSIGHT_API_EMBEDDINGS_LOCAL_MODEL=BAAI/bge-small-en-v1.5
+# Force CPU if local embeddings hit MPS/XPC instability on macOS:
+# HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU=false
 # For ONNX provider (local CPU embeddings without an Ollama/TEI sidecar):
 # HINDSIGHT_API_EMBEDDINGS_PROVIDER=onnx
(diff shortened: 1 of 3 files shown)


benfrank241 (approved):
lgtm — verified both settings are real (`ENV_EMBEDDINGS_LOCAL_FORCE_CPU` / `ENV_RERANKER_LOCAL_FORCE_CPU` in config.py, defaults false) and already in the config reference (configuration.md:522,818); this just surfaces them in the env templates where users actually find settings during profile setup. both templates edited identically (sync contract intact), commented + default false so no behavior change, and the regression test pins their presence. clean.

approving + kicked off CI.

PR #2708, opened 2026-07-14
fix(worker): bound terminal operation history

### LINKED ISSUE #2705: Terminal async operations retain full task payloads indefinitely
## Bug description

Hindsight retains terminal `async_operations` rows indefinitely, including the serialized `task_payload` that can contain the original retain content.

This conflicts with the operation-status API description, which currently says:

> Completed operations are removed from storage.

On current `main`, async submission inserts the complete payload into `async_operations`. Successful and failed workers update the row to a terminal status, but do not remove the row or scrub its payload. I found no general TTL, bounded history policy, or terminal-operation pruning job. The internal delete helper is used for exceptional cleanup paths, and bank deletion removes rows via cascade, but ordinary completed operations remain.

## Observed behavior

On a long-lived v0.8.4 deployment:

- one bank had 11,150 operation rows with no pending or processing work;
- completed and failed operation records survived service restarts;
- fetching terminal operations with `include_payload=true` still returned their stored task payloads;
- in a sample of 100 recent terminal operations, 65 had non-empty payloads totaling 365,055 serialized bytes, with a maximum of 77,909 bytes for one row.

Current list latency is still healthy, so this is not yet a demonstrated latency incident. It is, however, monotonic unbounded growth and retains duplicate raw input after derived memories have already been created.

## Reproduction

1. Submit an asynchronous retain and wait for it to reach `completed`.
2. Query the bank's operations endpoint and record `total`.
3. Fetch the completed operation with `include_payload=true`; observe the original task payload is still present.
4. Repeat with additional retains; observe `total` grows monotonically.
5. Restart the service and query again; the terminal rows and payloads remain.

## Impact

- Unbounded database and backup growth for long-lived banks.
- Raw retain input can outlive the operational need for it, weakening data-minimization expectations.
- `COUNT(*)` plus newest-first operation listing must scan an ever-growing journal; current indexes do not appear to include a bank/time ordering tailored to that query.
- Operators are told completed rows are removed when they are actually queryable indefinitely.

## Relationship to #2677

#2677 asks for manual deletion of terminal operations from the API/control plane. That is useful but does not define automatic retention, payload scrubbing, bounded pruning, or the behavior of completed operation history. This issue is about the storage lifecycle rather than a per-row delete button.

## Expected behavior

Hindsight should define and enforce an explicit terminal-operation retention contract. For example:

- scrub `task_payload` from completed operations after any required result/webhook processing;
- retain failed-operation payloads only for a bounded retry/debug window, since retry needs the payload;
- prune terminal rows by configurable TTL and/or per-bank maximum count in bounded batches;
- keep compact result/error metadata separately when audit history is desired;
- add an appropriate `(bank_id, created_at)` or equivalent index if long operation history remains queryable;
- update API documentation to match the implemented lifecycle.

Defaults should preserve a useful troubleshooting window without retaining complete task inputs forever.

## Suggested acceptance tests

- A completed operation no longer exposes its original retain payload after the configured scrub point.
- A failed operation remains retryable during its configured retention window.
- Expired terminal rows are pruned in bounded batches without affecting pending/processing work.
- Per-bank maximum-count pruning is deterministic when timestamps tie.
- Operation listing remains indexed and stable with a large terminal history.
- Service restart does not resurrect scrubbed/pruned payloads.
- API documentation accurately describes terminal-operation retention.

## Version

- Reproduced on v0.8.4.
- Current `main` still persists terminal operation rows and payloads without a general retention policy.

Tosko4:
## Summary

- add a configurable retention window for completed, failed, and cancelled async operations
- prune expired terminal rows in deterministic, bounded batches without touching pending or processing work
- preserve batch child outcomes until their parent row is pruned and make retry-vs-cleanup transitions atomic
- run cleanup outside the task-claiming loop and bind/reset tenant schema context safely for PostgreSQL and Oracle
- add cleanup/listing indexes and document the retention behavior and configuration

## Why

Terminal `async_operations` rows currently persist indefinitely, including the serialized task payload. On long-running banks this duplicates raw retained input in the database and backups, while operation history keeps growing without a bound.

This keeps the full row available for retry and debugging during the retention window, then removes it as one coherent unit. The default window is 30 days; setting `HINDSIGHT_API_OPERATION_RETENTION_DAYS=0` preserves the current keep-forever behavior.

Fixes #2705.

## Behavior

- terminal statuses: `completed`, `failed`, `cancelled`
- default retention: 30 days
- cleanup batch: 1000 rows per tenant schema and cleanup cycle
- pending and processing rows are never eligible
- failed and cancelled operations remain retryable while retained
- PostgreSQL and Oracle workers select, lock, re-check, and delete bounded candidate sets in one transaction
- completed/failed batch children remain protected while their parent exists; before an expired cancelled child is deleted, a pending same-bank parent is atomically marked cancelled and retained for a fresh window
- Oracle parent lookups validate canonical UUID metadata and explicitly convert it to `RAW(16)`, so malformed metadata cannot abort cleanup
- retries use one conditional state transition, so cleanup cannot race into a false success response
- retention cleanup runs as one non-overlapping background task per poller instead of blocking task claiming
- Oracle pooled sessions explicitly bind each tenant and reset the default schema before cleanup
- one schema failure is logged without blocking cleanup for other schemas

## Validation

- combined worker/config/migration/operation/batch/progress suite: 371 passed, 87 skipped
- batch-parent preservation/cancellation reconciliation, retry-vs-prune, background scheduling, and Oracle schema-reset regressions: passed
- migration-shape and PostgreSQL migration execution: passed
- concurrent PostgreSQL pruning regression: passed
- Oracle adapter rewrite smoke with a 1000-operation batch: passed
- Ruff check and format check: passed
- `ty` on all changed production modules: passed
- Python compilation and `git diff --check`: passed
- env templates and published OpenAPI copies: verified identical

## Risk / compatibility

This intentionally changes the default from unbounded terminal-operation history to 30-day retention. Deployments that need indefinite history can set `HINDSIGHT_API_OPERATION_RETENTION_DAYS=0`. Cleanup is rate-limited and batched; candidate status and cutoff are re-checked under row locks before deletion.


koriyoshi2041:
I did a focused local pass on the retention cleanup pieces here. These passed for me on the branch:\n\n```bash\nuv run pytest hindsight-api-slim/tests/test_config_validation.py::test_operation_retention_defaults hindsight-api-slim/tests/test_config_validation.py::test_operation_retention_env_overrides hindsight-api-slim/tests/test_config_validation.py::test_operation_retention_rejects_invalid_values hindsight-api-slim/tests/test_config_validation.py::test_operation_cleanup_batch_size_requires_positive_integer hindsight-api-slim/tests/test_operation_retention_migration.py -q\n# 9 passed\n\nuv run pytest hindsight-api-slim/tests/test_worker.py::TestWorkerOperationCleanupScheduling -q -n0\n# 6 passed\n\nenv -u ALL_PROXY -u all_proxy -u HTTPS_PROXY -u https_proxy -u HTTP_PROXY -u http_proxy uv run pytest hindsight-api-slim/tests/test_worker.py::TestTerminalOperationRetention::test_prune_terminal_operations_is_bounded_oldest_first_and_preserves_live_rows hindsight-api-slim/tests/test_worker.py::TestTerminalOperationRetention::test_concurrent_pruning_is_safe_and_idempotent hindsight-api-slim/tests/test_worker.py::TestTerminalOperationRetention::test_pruning_preserves_terminal_children_until_batch_parent_row_is_pruned -q -n0 --tb=short\n# 3 passed\n```\n\nI also tried the retry/status behavior tests, but my local run got stuck initializing the BAAI/bge-small-en-v1.5 embedding model after I removed proxy env vars, so I stopped that one rather than treating it as a code signal. No issue found in the config, migration-shape, background cleanup scheduling, or PostgreSQL pruning paths above.

benfrank241 (approved):
Solid work on a real storage-lifecycle gap — thanks for the detailed write-up and thorough test coverage.

**What I verified:**
- Read the core PG and Oracle concurrency logic. The bounded `FOR UPDATE … SKIP LOCKED` candidate lock, regex-validated UUID before `::uuid`/`HEXTORAW` (so malformed metadata can't abort cleanup), parent-guard protecting batch children while their parent row exists, atomic pending-parent cancellation before deleting a cancelled child, and the final DELETE re-checking terminal-status + cutoff under the held lock — all correct. A row retried into `pending` between the select and the delete is excluded by design.
- Ran the full retention + operation-status + worker suites locally: **184 tests pass** (`test_operation_status.py`, `test_operation_retention_migration.py`, `test_config_validation.py`, `test_worker.py`), including `test_retry_does_not_acknowledge_operation_deleted_by_cleanup`, `test_pruning_cancelled_child_cancels_parent_before_deletion_and_releases_siblings_later`, all background-scheduling and per-schema-failure-isolation tests.
- CI fully green including `verify-generated-files` (generated go/python/ts clients + openapi + skill mirrors consistent) and the Oracle jobs. `mergeState: CLEAN`.

**Two intentional design choices to call out — not blockers, but worth noting:**

1. **Full-row retention for the whole window vs early payload scrub.** The issue mentioned scrubbing `task_payload` immediately after result processing (strict data minimization) as one option; this PR instead keeps the complete row for the full 30-day window and removes it atomically. That's simpler, keeps retry viable throughout the window, and bounds growth to 30× what a single-day scrub would retain — a reasonable tradeoff. I'd call the early-scrub a potential follow-up if stricter minimization is needed later.

2. **Default behavior change (unbounded → 30-day).** This is a deliberate breaking change from the keep-forever status quo. Documented in the PR, flagged in the risk section, and `HINDSIGHT_API_OPERATION_RETENTION_DAYS=0` restores old behavior. Operators on long-running deployments with intentionally deep operation history should set that before upgrading.

These are product decisions I'm explicitly signing off on. The implementation is clean and the test surface covers what matters. Merging.

Issue #1240, opened 2026-04-23
daemon_embed_manager: prefer locally-installed hindsight-api over uvx


hachi-gallop:
## Problem

`_find_api_command()` in `daemon_embed_manager.py` falls back to `uvx hindsight-api@{version}` when not in dev mode. On NixOS (and likely other environments where hindsight-all is installed via `uv pip install --target`), this causes a Python version mismatch:

1. `uv pip install hindsight-all` installs everything into a site-packages directory (compiled for the system Python, e.g. 3.11 or 3.13)
2. `uvx` downloads its own standalone Python (currently 3.14) plus all deps
3. The uvx-installed entrypoint resolves imports from the local site-packages due to Python path resolution
4. C extensions like `asyncpg.protocol.protocol` are ABI-incompatible between Python versions → `ModuleNotFoundError`

Error:
```
ModuleNotFoundError: No module named 'asyncpg.protocol.protocol'
```

This causes the daemon to fail to start, and the 180s startup timeout makes the hang severe (each retry blocks for 3 minutes).

## Suggested Fix

Add a check in `_find_api_command()` for an already-installed `hindsight-api` binary in the site-packages bin directory before falling back to uvx:

```python
def _find_api_command(self) -> list[str]:
    # Check if we're in development mode
    dev_api_path = Path(__file__).parent.parent.parent / "hindsight-api-slim"
    if dev_api_path.exists() and (dev_api_path / "pyproject.toml").exists():
        return ["uv", "run", "--project", str(dev_api_path), "hindsight-api"]

    # Check for hindsight-api installed in our site-packages bin
    site_bin = Path(__file__).parent.parent / "bin" / "hindsight-api"
    if site_bin.exists():
        return [str(site_bin)]

    # Fall back to uvx for installed version
    from . import __version__
    api_version = os.getenv("HINDSIGHT_EMBED_API_VERSION", __version__)
    return ["uvx", f"hindsight-api@{api_version}"]
```

This preserves the uvx fallback for users who don't have hindsight-api installed locally, while avoiding the version mismatch for those who do.

## Environment

- NixOS (but affects any setup using `uv pip install --target`)
- hindsight v0.5.4
- uvx downloads Python 3.14, site-packages compiled for Python 3.11

PR #2713, opened 2026-07-14
feat(retain): opt-in minified JSON output to cut LLM output tokens

cdbartholomew:
## Summary

Models pretty-print structured JSON output by default (line breaks + indentation), which is **~40% of the output tokens**. Fact extraction, consolidation, and mental-model delta operations all emit JSON that is machine-parsed (`json.loads`) and never shown to a user — so that whitespace is pure decode cost.

This adds an **opt-in** `HINDSIGHT_API_MINIFY_LLM_JSON_OUTPUT` flag (default **off**) that appends a formatting-only directive to those prompts asking the model to emit compact/minified JSON.

**Measured ~40–47% fewer output tokens** on the same extracted content across two providers — a latency win when self-hosting and a cost win on all providers, with no information loss.

## What changed

- **`config.py`** — new global flag (deliberately *not* per-bank, so the cached consolidation system-prompt prefix stays identical across banks).
- **`engine/prompt_utils.py`** — shared `minified_json_directive()` helper. It's brace-free so it survives prompts later run through `str.format()`, and returns `""` when the flag is off.
- Wired into **retain fact extraction** (all LLM extraction modes), **both consolidation prompt builders**, and the **mental-model structured-delta** call. Worded strictly as a *formatting* instruction so it does not change which facts / observations / operations the model produces.
- **`.env.example`**, **`configuration.md`**, and the bundled embed env template documented.

## Not touched

Prose outputs (the reflect final answer, recall/think synthesis, mission merge) are deliberately left alone — those are user-facing text, not machine-parsed JSON.

## Tests

- `test_minified_json_output.py` — deterministic: off by default, present when enabled (all retain modes + both consolidation builders), flag reversibility, and brace-safety through `str.format()`.
- `test_minified_json_quality.py` — `hs_llm_core` real-LLM judge test that runs extraction with the flag **off vs on** and asserts (via the LLM judge) the **same facts** are extracted — proving the directive is formatting-only.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -89,6 +89,13 @@ HINDSIGHT_API_LOG_LEVEL=info
 # Unset uses HINDSIGHT_API_RETAIN_CHUNK_SIZE as the structured-chunk limit.
 # HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE=
 
+# Emit minified (compact) JSON for machine-parsed structured LLM outputs (fact
+# extraction, consolidation, mental-model deltas). Models pretty-print by default
+# (~40% of output tokens); since these outputs are parsed, never displayed,
+# minifying is a lossless ~40% output-token cut — lower latency on self-hosted
+# models, lower cost on all providers. Opt-in; default off.
+# HINDSIGHT_API_MINIFY_LLM_JSON_OUTPUT=false
+
 # Dry-run extraction preview endpoint (POST /memories/dry-run-extract). Enabled by default; it makes
 # a real LLM call but stores nothing. Set to false to remove the endpoint (returns 404).
 # HINDSIGHT_API_ENABLE_DRY_RUN_EXTRACT=true
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -491,6 +491,7 @@ ENV_RETAIN_MAX_COMPLETION_TOKENS = "HINDSIGHT_API_RETAIN_MAX_COMPLETION_TOKENS"
 ENV_RETAIN_CHUNK_SIZE = "HINDSIGHT_API_RETAIN_CHUNK_SIZE"
 ENV_RETAIN_STRUCTURED_CHUNK_SIZE = "HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE"
 ENV_RETAIN_EXTRACT_CAUSAL_LINKS = "HINDSIGHT_API_RETAIN_EXTRACT_CAUSAL_LINKS"
+ENV_MINIFY_LLM_JSON_OUTPUT = "HINDSIGHT_API_MINIFY_LLM_JSON_OUTPUT"
 ENV_RETAIN_EXTRACTION_MODE = "HINDSIGHT_API_RETAIN_EXTRACTION_MODE"
 ENV_RETAIN_MISSION = "HINDSIGHT_API_RETAIN_MISSION"
 ENV_RETAIN_CUSTOM_INSTRUCTIONS = "HINDSIGHT_API_RETAIN_CUSTOM_INSTRUCTIONS"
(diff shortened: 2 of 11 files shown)


koriyoshi2041:
I did a quick third-party triage pass on the current red jobs.

What I checked locally:

```bash
cd hindsight-api-slim
env -u ALL_PROXY -u all_proxy -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy \
  uv run pytest tests/test_minified_json_output.py -q
```

Result: `11 passed`. The deterministic prompt/config coverage for the minify flag looks clean.

The hosted failures do not look tied to the minify diff itself:

- `test-api (1/3)` fails in `tests/test_db_abstraction.py::TestOracleSetSessionSchema` (`_FakeAsyncCursor` missing `fetchone`; public schema noop assertion). That is outside the touched retain/consolidation/prompt files.
- `test-api (2/3)` fails while starting embedded Postgres for `test_backfill_populates_null_observation_search_vector`; the log shows port `5568` already in use / could not bind sockets, so this looks like pg0 port collision rather than a code assertion.
- `test-embed-windows` fails at the target-install guard: `Expected sibling hindsight-api.exe, got ['uvx', 'hindsight-api@0.0.0']`. That path is the embed API binary fallback / #1240 guard, not a minify prompt path.

I could not run `tests/test_minified_json_quality.py` locally because it needs `HINDSIGHT_API_LLM_API_KEY`; hosted `Core LLM tests` was green on this PR.

benfrank241:
verified the OSS surface — clean and well-built. flagging for @nicoloboschi to own the merge since it's a `feat` (opt-in optimization) and it edits the prompt-engineering surface across retain extraction / both consolidation builders / mental-model delta, which is your area to sign off on (the directive is worded formatting-only and the `hs_llm_core` off-vs-on judge test backs that, but worth your eyes on the wording).

what i confirmed:
- config-flag checklist complete: `HINDSIGHT_API_MINIFY_LLM_JSON_OUTPUT` wired in config.py (default **off**), documented in `.env.example` + configuration.md, embed `env.example` is **byte-identical** to root, and the skill mirror was **regenerated** (flag present) — no verify-generated-files / embed-sync drift.
- deterministic coverage passes: `test_minified_json_output.py` **11/11** (off-by-default, present-when-enabled across all retain modes + both consolidation builders, brace-safety through `str.format()`). @koriyoshi2041 also independently got 11 passed and triaged the hosted reds as unrelated to the diff.
- design is sensible: global (not per-bank) so the cached consolidation prefix stays stable; prose outputs deliberately left alone.

one nit: `engine/prompt_utils.py:48` docstring says "flag (default on)" but the flag is opt-in / default **off** (matches everywhere else) — worth correcting so it doesn't mislead.

no objection to the change itself — just want the "should we ship this optimization flag + is the directive wording right" call to be yours.

cdbartholomew:
Closing this. Testing against a real self-hosted model showed the prompt-directive approach doesn't hold up:

1. **The headline number was wrong for the structured-output path.** When the output is produced via json_schema guided decoding, the model's JSON is already single-line (the grammar suppresses newlines/indentation), so the pure-whitespace savings ceiling is ~16%, not the ~40% measured on free-form generation. Fact-extraction/consolidation always use guided decoding, so they never had the 40% of bloat to reclaim.

2. **At realistic input sizes the directive is not formatting-only.** On a large chunk it caused the model to merge atomic facts into compound ones (34 atomic facts → 26 compound) — same information, coarser granularity. Instructing the model to "be compact" bleeds into extraction behavior, and atomic granularity matters for downstream retrieval/dedup. The "formatting-only, lossless" guarantee doesn't survive.

The reliable way to strip whitespace from guided-decoding output is at the decoder level (e.g. disabling grammar whitespace) rather than via a prompt, which keeps extraction behavior identical — but that's engine/provider-specific and out of scope for a portable prompt change. Not worth carrying this as-is.

Issue #2714, opened 2026-07-15
Control plane bank selector renders bank_id instead of the bank's name (display label already fetched but unused)


fc-friday:
## Summary

The control-plane bank selector shows the raw `bank_id` as the visible label even when a bank has a human-readable `name`. The `name` is already fetched into client state — the selector simply doesn't render it. A one-line change would let banks with opaque IDs display a friendly label while keeping `bank_id` as the tooltip/copy value.

## Context / motivation

We key banks on a stable, opaque identifier by design (in our case a Linear project's hex id, e.g. `project-9d6f1542fbee`), because `bank_id` is immutable and rename-safe. That makes the control plane hard to read: every bank shows as `project-<hex>`. We set each bank's `name` via `PATCH /v1/default/banks/{bank_id}` (e.g. `"Tarn (project)"`), but the UI never displays it.

## Current behavior

`hindsight-control-plane/src/components/bank-selector.tsx` (main) renders `bank.bank_id` for the label, the tooltip, and the copy action (~lines 558–575):

```tsx
const isSelected = currentBank === bank.bank_id;
// ...
<span className="truncate flex-1 font-medium" title={bank.bank_id}>
  {bank.bank_id}
</span>
// ...
onClick={() => navigator.clipboard.writeText(bank.bank_id)}
```

The label shows the id even when a `name` exists.

## Why this is nearly free: the name is already in client state

`hindsight-control-plane/src/lib/bank-context.tsx` keeps `name` on the bank model and populates it from the API response:

```tsx
export interface BankInfo {
  bank_id: string;
  name: string | null;   // already carried
  mission: string | null;
  // ...
}
// ...
name: bank.name ?? null,  // populated from the API
```

So no new fetch or type change is needed — only the render line hard-codes the id.

## Proposed change

Render `name` when present, fall back to `bank_id`:

```tsx
<span className="truncate flex-1 font-medium" title={bank.bank_id}>
  {bank.name || bank.bank_id}
</span>
```

Keep `title={bank.bank_id}` and the copy action on `bank_id` so the canonical identifier stays one hover/click away. Banks without a `name` are unchanged (they already fall back to the id).

## Notes / open question for maintainers

[...]

PR #2715, opened 2026-07-15
fix(control-plane): show bank name in selector, fall back to bank_id

fc-friday:
## What

Render a bank's `name` as the selector label when it's set, falling back to `bank_id`:

```diff
-                          <span className="truncate flex-1 font-medium" title={bank.bank_id}>
-                            {bank.bank_id}
+                          <span className="truncate flex-1 font-medium" title={bank.bank_id}>
+                            {bank.name || bank.bank_id}
                           </span>
```

`title={bank.bank_id}` and the copy button are unchanged, so the canonical identifier stays one hover/click away.

## Why

When banks are keyed on stable, opaque identifiers (immutable, rename-safe ids rather than human-friendly slugs), the selector is hard to read — every bank shows as its raw id (e.g. `project-9d6f1542fbee`). Banks can already carry a human-readable `name` (settable via `PATCH /v1/default/banks/{bank_id}`), and the control plane **already fetches it into client state** — `BankInfo` in `src/lib/bank-context.tsx` declares `name: string | null` and populates it from the API response (`name: bank.name ?? null`). Only the selector's render line hard-codes the id.

## Scope / safety

- One-line render change; no new fetch, no type change, no API change.
- Banks without a `name` are unchanged — they already fall back to the id via `||`.
- Canonical `bank_id` remains visible (tooltip) and copyable (copy button).

## Note on the deprecated `name` field

`hindsight-api` marks `name` deprecated server-side (the `PATCH` handler stores it "for display only, deprecated"; `CreateBankRequest.name` is "display label only, not advertised"). This PR assumes a human-readable label is still worth showing where one exists. If the plan is to fully retire `name`, I'd love to know the supported way to attach a display label to a bank whose `bank_id` must stay opaque — happy to adjust this PR to whatever surface you prefer.

Linked issue: #2714

### DIFF
--- a/hindsight-control-plane/src/components/bank-selector.tsx
+++ b/hindsight-control-plane/src/components/bank-selector.tsx
@@ -556,7 +556,7 @@ function BankSelectorInner() {
                             )}
                           />
                           <span className="truncate flex-1 font-medium" title={bank.bank_id}>
-                            {bank.bank_id}
+                            {bank.name || bank.bank_id}
                           </span>
                           <button
                             type="button"


benfrank241:
thanks @fc-friday — good catch on this, the selector really was showing raw `bank_id` instead of the friendly name. closing as superseded though: #2693 (open since 07-13, `Fixes #2686`) already makes this change and covers a bit more — it also renders the name on the **selected-bank trigger label** (the one you see when the dropdown is closed) and updates the `title` tooltip, whereas this one touches just the dropdown item span. so #2693 is the fuller fix for the same issue.

appreciate you jumping on it — follow #2693 for the merge.

PR #2718, opened 2026-07-15
Show bank names in control-plane selector

### LINKED ISSUE #2714: Control plane bank selector renders bank_id instead of the bank's name (display label already fetched but unused)

koriyoshi2041:
## Summary
- render a bank's display name in the control-plane selector when one is present
- keep the canonical `bank_id` as the hover title and copy value
- add focused coverage for name, null-name, and blank-name fallback behavior

Closes #2714

## Tests
- `npm run test -- tests/lib/bank-label.test.ts`
- `npm run test`
- `npm run build`

### DIFF
--- a/hindsight-control-plane/src/components/bank-selector.tsx
+++ b/hindsight-control-plane/src/components/bank-selector.tsx
@@ -60,6 +60,7 @@ import {
 import { Switch } from "@/components/ui/switch";
 import { cn } from "@/lib/utils";
 import type { BankInfo } from "@/lib/bank-context";
+import { bankDisplayLabel } from "@/lib/bank-label";
 
 function formatCompact(n: number): string {
   if (n >= 1_000_000) return `${(n / 1_000_000).toFixed(n >= 10_000_000 ? 0 : 1)}M`;
@@ -527,6 +528,7 @@ function BankSelectorInner() {
                   {sortedBanks.map((bank) => {
                     const barPct = (bank.fact_count / maxFactCount) * 100;
                     const isSelected = currentBank === bank.bank_id;
+                    const displayLabel = bankDisplayLabel(bank);
                     return (
                       <CommandItem
                         key={bank.bank_id}
@@ -556,7 +558,7 @@ function BankSelectorInner() {
                             )}
                           />
                           <span className="truncate flex-1 font-medium" title={bank.bank_id}>
-                            {bank.bank_id}
+                            {displayLabel}
                           </span>
                           <button
                             type="button"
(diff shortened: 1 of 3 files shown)


koriyoshi2041:
Closing this because I found the same selector-label fix is already covered by #2693 and #2715. Sorry for the duplicate noise.

PR #2724, opened 2026-07-15
feat(paperclip): per-agent enable/disable for pilot rollouts

### LINKED ISSUE #1116: Paperclip plugin: per-agent enable/disable for pilot rollouts
## Use case

Paperclip users running production multi-agent fleets (in our case, 21 agents across multiple specialties) want to pilot `@vectorize-io/hindsight-paperclip` on a small subset of agents (e.g. 2-3 high-signal agents) before enabling fleet-wide. The docs explicitly frame this as a benefit ("Install once, every agent gets memory"), but for fleets where recall/retain has meaningful LLM cost and adds latency to every agent-run lifecycle, all-or-nothing enable is a blocker to adoption.

## Proposed API

Add an optional `enabledAgentIds: string[]` (or equivalent) to `instanceConfigSchema`. When set, the plugin's `agent.run.started` / `agent.run.finished` event handlers early-return unless the incoming event's `agentId` is in the allowlist. When unset, behavior is unchanged (all agents).

Alternative: `disabledAgentIds` (default-on, opt-out) — probably less commonly needed.

## Current workaround

Forking / patching `worker.js` to add an allowlist gate in the event handler dispatch (~20 LOC). Works but creates re-apply burden on every plugin version bump.

## Why this matters now

Hindsight's `retain` and `recall` each make an LLM call against the configured `HINDSIGHT_API_LLM_PROVIDER`. On a 21-agent fleet with frequent wakes (cron routines, issue assignments), blanket enable adds N*2 provider calls per cycle. Pilot-first is the lower-risk adoption path.

Happy to submit a PR if you're open to the shape — point me at the handler file(s).

parties:
*Here's my attempt at adding this functionality; first PR to the repo, my apologies if I missed something. I am working on testing this locally myself, but I'm curious to see what tests/validations exist in the PR pipeline, ~~so I'm submitting this as a draft for now~~.*

Add optional `enabledAgentIds` config field to restrict Hindsight recall/retain to a subset of agents.

Behavior:
- When set, only listed agent IDs trigger memory operations
- Unset or empty array (`[]`) = unchanged behavior (all agents)
  - Enables pilot rollouts on high-signal agents before fleet-wide enable, reducing LLM cost/latency risk.
- Add `enabledAgentIds: string[]` to `instanceConfigSchema` (`manifest.ts`)
- Add `isAgentEnabled()` gate function to `worker.ts`
- Gate `agent.run.started` recall, `agent.run.finished`, and `issue.comment.created` retain handlers (the actual LLM-cost operations)
- Add 6 test cases covering `allowlist` pass/fail, empty array, and unset behavior
- Update `README.md` config table

Resolves #1116

### DIFF
--- a/hindsight-integrations/paperclip/README.md
+++ b/hindsight-integrations/paperclip/README.md
@@ -41,6 +41,7 @@ hindsight-api
 | `bankGranularity`    | `["company", "agent"]`               | Memory isolation when `dynamicBankId` is `true`: per company+agent, per company, or per agent. Add `"user"` for per-user memory isolation (useful for GDPR compliance) |
 | `recallBudget`       | `mid`                                | `low` = fastest, `mid` = balanced, `high` = most thorough                                                                                                              |
 | `autoRetain`         | `true`                               | Automatically retain run output after every run                                                                                                                        |
+| `enabledAgentIds`    | —                                    | Restrict recall/retain to these agent IDs only. Leave empty to enable for all agents (default)                                                                         |
 
 ## Bank ID Format
 
--- a/hindsight-integrations/paperclip/src/manifest.ts
+++ b/hindsight-integrations/paperclip/src/manifest.ts
@@ -74,6 +74,13 @@ const manifest: PaperclipPluginManifestV1 = {
         description: "Automatically retain agent run output to Hindsight when a run completes.",
         default: true,
       },
+      enabledAgentIds: {
+        type: "array",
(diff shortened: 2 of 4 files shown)


parties:
Might as well ask here, anyone have suggestions on how to test this? I'm sure I could spin up a new company in Paperclip and install it, but I'm curious to know if there's a more official/automated process, or if I just need to write some more tests. 🤔

Issue #2725, opened 2026-07-15
Add first-class multimodal image memory support


Sanderhoff-alt:
## Motivation

Hindsight should support images as a first-class memory input, not merely as files to be transcribed with OCR. Modern agents increasingly receive screenshots, product images, marketing creatives, charts, UI states, and mixed text-and-image messages. They need to remember the visual meaning of those inputs and retrieve it later through the same memory experience used for text.

This deserves a focused issue even though there are related multimodal discussions, because image input has a clear user-facing scope and can be evaluated independently from audio, video, and other modalities.

## Requested capability

- Retain one or more images, either by themselves or together with user-provided text and per-image context.
- Use a configured vision-capable model to understand scenes, objects, actions, states, colors, spatial relationships, and obvious chart or UI meaning without treating the feature as an OCR pipeline.
- Make visual semantics available through normal memory recall while preserving an explicit association with the source image.
- Let authorized clients list, read, and delete retained images, and ensure image-backed memories remain understandable when a source image is unavailable or intentionally deleted.
- Preserve image-backed memories and their source associations across supported export and import workflows.
- Provide consistent support through the Data Plane API and official SDKs, including self-hosted and private-network deployments.
- Preserve tenant and bank isolation, clear failure behavior, retry safety, and protection against leaking image data or credentials through logs and operation metadata.

## Example use cases

- Online marketing agents remembering campaign creatives, product placement, visual themes, and changes between revisions.
- Commerce and catalog agents remembering product appearance and relationships between items.
- Support and automation agents remembering screenshots, UI states, and visible error conditions.
- Analytics agents remembering the visible meaning of charts and dashboards without requiring a full text transcription.
- Conversational agents retaining mixed text-and-image interactions as one coherent memory.

## Expected outcome

A caller should be able to submit an image-backed memory, retrieve the resulting visual knowledge through ordinary recall, identify the associated source image when requested, manage that image within the same authorization boundary, and move the memory between supported deployments without silently losing its image provenance.

Sanderhoff-alt:
[...]

PR #2726, opened 2026-07-15
feat(images): add managed multimodal image support

Sanderhoff-alt:
## Summary

This change adds opt-in, first-class multimodal image memory support to the Hindsight backend, Data Plane API, OpenAPI contract, and all four official SDKs. It intentionally does not change the Control Plane. Images are interpreted for visual meaning rather than routed through OCR: the configured vision-capable model produces a strict structured analysis, that analysis is rendered as readable Markdown, and the resulting document continues through the existing Retain pipeline so image-backed knowledge participates in the normal fact, entity, embedding, graph, consolidation, and text Recall behavior.

Part of #2725.

## Scope and boundaries

- Support JPEG, PNG, and WebP uploads, including `.jpg` and `.jpeg`, while preserving the exact uploaded bytes without EXIF correction, resizing, recompression, or format conversion.
- Support image-only, text-plus-image, and ordered multi-image Retain requests.
- Keep the Data Plane complete without requiring the Control Plane, MCP, chat integrations, image embeddings, image-to-image search, or OCR.
- Keep existing text Retain and Recall behavior compatible. The new Recall image map is returned only when explicitly requested.
- Avoid external image URLs and signed URLs. The Worker reads authorized bytes from FileStorage and sends them directly to the configured model adapter.

## Processing model

The write path is `multipart upload -> input validation -> managed asset persistence -> async image analysis -> strict structured result -> readable Markdown document -> existing Retain pipeline`. The visual model output has exactly three fields: `summary`, `visual_details`, and `visual_relations`. Provider-native JSON schema or tool use constrains the response, and Pydantic validates the same closed schema before publication.

The generated document is intentionally human-readable. A new or replaced image document contains one `# Image Memory` root heading, while each image Retain request produces one `## Image Entry`. Optional user-provided content and request-local `Image N` sections are siblings within that entry; append adds a separated entry without repeating the root heading. Per-image context, Summary, Visual Details, and Visual Relationships remain nested under the corresponding image. Image headings and local numbering are presentation-only. Recall returns the associated image descriptors from `document_image_links`, and callers can fetch the original image later by `asset_id`. Structured analysis and model provenance are transient after validation, with the retained Markdown serving as the only persisted semantic representation.

Hindsight does not infer visual capability from a model name. Operators are responsible for configuring a vision-capable model; an incompatible model fails explicitly. When no image-specific model setting is present, the effective Retain failover chain is reused and filtered to providers with an implemented image request adapter. Setting any image-specific field creates one explicit image member and avoids mixing credentials or routing from the Retain chain.

## Data Plane API

- `POST /v1/default/banks/{bank_id}/memories/image-retain` accepts repeated ordered `files` parts and one JSON `request` part. The request carries ordinary Retain fields plus optional per-image `asset_id` and context. The operation is always asynchronous and supports replace and append semantics.
- `GET /v1/default/banks/{bank_id}/image-assets` lists managed image descriptors with document, lifecycle status, and pagination filters.
- `GET /v1/default/banks/{bank_id}/image-assets/{asset_id}` combines authorized binary content and descriptor metadata in one response. It does not expose a separate content URL.
- `DELETE /v1/default/banks/{bank_id}/image-assets/{asset_id}` rejects referenced assets with `409 Conflict`. After the Document is deleted or replaced and the association is removed, it synchronously deletes the unreferenced blob and metadata; the same DELETE can resume an interrupted cleanup.
- `POST /v1/default/banks/{bank_id}/memories/recall` accepts `include.image_assets=true` and returns a top-level `document_id -> ordered image descriptors` map. It does not return base64 data or automatically download images.

Client-supplied asset IDs are bank-scoped and optional; omitted IDs are generated by the service. Each request accepts up to ten ordered images, and each document accepts up to one hundred associated images. The only new admission-size setting is `HINDSIGHT_API_IMAGE_MAX_FILE_SIZE_MB`; model-side automatic downsampling remains the model provider's responsibility.

## Persistence and lifecycle

`image_assets` stores the bank-scoped asset ID, opaque FileStorage key, exact byte length and SHA256, detected MIME type, dimensions, lifecycle status, and timestamps. `document_image_links` stores only the bank-scoped document/asset association, ordinal, optional image context, and creation time. Structured model output and provenance are transient after strict validation; the retained Markdown is the only persisted semantic representation.

The implementation deliberately avoids a batch snapshot table, analysis-version table, source segment table, and multi-version document publication layer. Request grouping is not persisted on image links; `async_operations.result_metadata.published_batch_id` is the durable publication and retry marker. The existing Document row lock and Retain transaction callback atomically publish the retained document, image links, and operation marker.

Text append and ordinary document reprocessing preserve existing image associations. Text replace removes them consistently with replacing the document source. No image-specific reprocess API is introduced: replacing an image-backed document can request new visual analysis, while the existing document reprocess operation rebuilds facts and embeddings from the already saved Markdown without paying for another vision call.

Asset states are limited to `ready`, `failed`, and `deleting`. Failed or unreferenced assets remain visible through the management API so an operator can remove them; no speculative automatic asset-retention policy is added. Deleted rows are removed rather than retained as tombstones.

## Idempotency, concurrency, and failure behavior

Image admission stores the complete managed-asset set and async operation in one database transaction after FileStorage writes succeed. An idempotency-key hash and canonical request fingerprint make concurrent identical submissions converge on one operation and one asset set; reusing a key with different content returns a conflict. Operation payloads contain only ordered logical asset references, Retain parameters, per-image options, and opaque execution identifiers—never image bytes, base64 data, data URIs, external URLs, internal storage keys, or credentials. Replaying a key whose original assets were explicitly deleted also returns a conflict instead of recreating deleted data.

The Worker rechecks asset state before analysis and again while holding the Document lock in the final Retain callback. Append ordinal allocation and the document image limit are enforced inside that lock, which prevents concurrent append and append-versus-replace races. A delete that wins during analysis prevents final link publication. Retried Workers check the operation publication marker before making another provider call or appending semantic text.

FileStorage and the database are not presented as a distributed transaction. Synchronous admission and import paths compensate uploaded blobs when a later database step fails. Terminal analysis failures mark unlinked assets as failed. Delete failures retain a minimal `deleting` marker and storage key so the same synchronous DELETE can safely resume cleanup.

## Model adapters and configuration

The adapters cover OpenAI-compatible providers, Anthropic, Gemini, and Vertex AI. OpenAI-compatible requests use strict JSON schema and inline data URLs; Anthropic uses a forced tool call with a base64 image source; Gemini and Vertex use native byte parts and a response schema. A deterministic mock adapter supports tests without external credentials.

The feature is disabled by default with `HINDSIGHT_API_ENABLE_IMAGE_RETAIN_API=false`. Four optional static overrides follow existing repository naming and inheritance conventions: `HINDSIGHT_API_IMAGE_LLM_PROVIDER`, `HINDSIGHT_API_IMAGE_LLM_API_KEY`, `HINDSIGHT_API_IMAGE_LLM_MODEL`, and `HINDSIGHT_API_IMAGE_LLM_BASE_URL`. If all four are absent, the effective Retain configuration is reused. The API advertises one capability flag, `multimodal_image_input`, only when the endpoint is enabled and at least one adapter configuration is valid.

## FileStorage and transfer

FileStorage gains `store_stream`, `stat`, and `iter_bytes` capabilities. PostgreSQL uses chunked objects, while S3, GCS, and Azure use disk-backed streaming or multipart upload and chunked reads. The image path never calls `get_download_url()`, which keeps self-hosted and private-network deployments independent of model-side callbacks.

Document Transfer schema v2 carries exact image bytes, asset metadata, and typed document links. Export writes a disk-backed streaming ZIP and pages database reads; import spools and validates the archive locally before uploading it once to FileStorage. Entry count, individual and total uncompressed size, compression ratio, path safety, MIME, size, and SHA256 are checked. The existing v1 text format remains importable. Whole-bank export and import preserve images as well as document-scoped transfer. Expiring transfer staging records and blobs are cleaned by the maintenance loop after interrupted work.

## Extension hooks and authorization

Image admission and execution use dedicated `validate_image_retain` hooks because validators may need image count, byte size, MIME, asset ID, update mode, model routing, and idempotency context that cannot be represented accurately as an ordinary text Retain operation. After the configured provider fallback chain resolves each image, `on_image_analyze_complete` runs once for that image's final success or failure with identifiers, selected provider and model, size, status, and a redacted error, but not the image or semantic body.

Derived Markdown still passes through the existing Retain validation and completion hooks. Asset list and binary read use bank-read validation; deletion uses bank-write validation. Transfer export and import use their existing bank access hooks plus body prechecks. New extension methods have compatible defaults so existing validators continue to work and still protect derived text through the original Retain hook.

All asset and link queries are bank-scoped. Cross-bank reads, links, deletes, and transfer reuse are rejected. Uploaded bytes, base64 payloads, image URLs, credentials, and full model request bodies are excluded from operation payloads, audit metadata, LLM traces, and post-hook data.

## SDK and compatibility work

OpenAPI and the Python, TypeScript, Rust, and Go clients are regenerated. Each SDK can submit ordered multipart image batches, parse the Recall image map, list assets, fetch a descriptor and binary content together, delete an asset, and stream or write document transfer archives without requiring the entire archive in memory. Rust retains a tested multipart helper because its generated client path does not provide the required request shape directly.

The default Recall response omits `image_assets`, preserving older strict decoders. New clients connecting to an older service receive an explicit unsupported-endpoint failure instead of silently dropping an image. The Control Plane is unchanged and can later consume the same authenticated Data Plane contracts through server-side multipart and binary proxies.

## Validation

- Backend image tests cover exact-byte validation, JPEG/PNG/WebP handling, provider request contracts, strict analysis schema, operation payload redaction, idempotency, concurrent append and replace, deletion races, failure recovery, hooks, Recall associations, transfer integrity, constant-memory behavior, and normal text append or reprocess behavior.
- A single Alembic revision provides PostgreSQL and Oracle upgrade/downgrade paths. PostgreSQL lifecycle tests run on a fresh database, and no-external-model lifecycle and transfer coverage are included in the Oracle test matrix.
- Native PostgreSQL, S3, GCS, and Azure FileStorage streaming contracts are covered.
- Python, TypeScript, Rust, and Go image behavior tests pass, and TypeScript builds successfully.
- OpenAPI and client generation complete successfully, repository lint and type checking pass, and the generated documentation skill is synchronized.
- A real end-to-end smoke test with `qwen3.5-27b` correctly identified a blue mug, a red book, two red apples, one green apple, and their spatial relationships in a non-OCR scene. The async operation, strict structured analysis, Markdown document, existing Retain pipeline, Recall image map, idempotent replay, and exact download SHA256 all passed.

## Deliberately deferred work

This change does not add Control Plane UI or proxies, MCP tools, chat integration adapters, OCR, image embeddings, image-to-image search, visual query ranking, automatic source cleanup policy, or a document multi-version publication subsystem. Those capabilities can be discussed independently without weakening the complete Data Plane contract delivered here.


nicoloboschi:
closing, we will rethink about it in the future

Issue #2055, opened 2026-06-08
0.8.0 installs tokenizers 0.23.1, breaking local embeddings with current transformers constraint


ht1072:
## Summary

Upgrading to `hindsight-api-slim==0.8.0` can install `tokenizers==0.23.1`, which breaks startup when using local embeddings / reranker with a `transformers` build that requires `tokenizers<=0.23.0`.

In this state, Hindsight fails during local model initialization and startup does not complete.

## Reproduction

1. Start from a working `0.7.2` environment using local embeddings and local reranker.
2. Upgrade to `hindsight-api-slim==0.8.0`.
3. Launch the API with local embeddings/reranker still enabled.

## Observed behavior

Startup fails during local model import/init. The underlying import error is:

```text
ImportError: tokenizers>=0.22.0,<=0.23.0 is required for a normal functioning of this module, but found tokenizers==0.23.1
```

In the failing environment, this prevents local embeddings from initializing, so the API never reaches a usable started state.

## Expected behavior

Installing `hindsight-api-slim==0.8.0` should not resolve to a `tokenizers` version that is incompatible with the currently installed / required `transformers` version for the supported local-ML path.

## Temporary workaround

Downgrading tokenizers to a compatible version restores startup:

```bash
pip install 'tokenizers==0.22.2'
```

## Suggestion

Please tighten or correct the dependency constraints so a normal `0.8.0` install does not land on an incompatible `tokenizers` version for the local embeddings/reranker path.

PR #2727, opened 2026-07-15
chore(deps): bump pydantic-settings, transformers, soupsieve (security)

dcbouius:
Closes the `pydantic-settings` Dependabot alert across every affected manifest, plus the three high-severity alerts in the root lock.

| Package | From | To | Advisory |
|---|---|---|---|
| `pydantic-settings` | 2.12.0 / 2.14.0 / 2.14.1 | 2.14.2 | GHSA-4xgf-cpjx-pc3j |
| `transformers` | 5.3.0 | 5.12.1 | GHSA-fgcw-684q-jj6r |
| `soupsieve` | 2.8 | 2.8.4 | GHSA-2wc2-fm75-p42x, GHSA-836r-79rf-4m37 |

`pydantic-settings` is transitive everywhere (no direct declaration), so the locks are the only lever. It's bumped in all 7 flagged manifests: root, `hindsight-integration-tests`, `strands`, `openai-agents`, `dify`, `claude-agent-sdk`, `agno`.

## Reviewer notes

**1. The transformers floor bump is a behavioral change for downstream consumers.**

`transformers` is a direct dep and `hindsight-api` is published to PyPI, so the *declared floor* — not our lock — is what protects installers of the `local-ml` / `local-onnx` extras. The old `>=4.53.0` floor resolves to 4.57.6 (vulnerable) under any downstream cap of `transformers<5`, so a user could get a vulnerable install while our lock looked clean. Raised to the advisory's first patched version, `>=5.5.0`.

The trade: consumers pinned below transformers 5 now get a **resolution error instead of a silent vulnerable install**. That's intended, but it's the notable call in this PR.

Worth knowing the old floor was already unreachable in practice — 4.53.0 requires `tokenizers<0.22`, which our own cap excludes.

**2. The `tokenizers<=0.23.0` cap is deliberately kept.**

It looks redundant (every transformers >=5.5.0 declares the same cap itself), but it is the guard for #2055: transformers shipped metadata declaring a *wider* tokenizers range than its import-time check enforces, which let 0.23.1 in and broke local embeddings at startup. Dropping the cap re-opens that door and trusts the metadata that already failed once. The comment now records this so it doesn't read as removable next time.

**3. `crewai` is intentionally not bumped.**

Its lock has `pydantic-settings` 2.10.1, which looks stale next to this sweep. The advisory's range is `>=2.12.0, <2.14.2` — `NestedSecretsSettingsSource` didn't exist in 2.10.x, so it's genuinely unaffected. Dependabot correctly did not flag it.

**4. Root `uv.lock` diff is 3353 lines; only 3 are real.**

uv rewrites in its own current format whenever it writes, so the upgrade dragged the root lock from revision 1 to 3 and added `upload-time` metadata throughout. This isn't drift — the other 32 locks in the repo are already revision 3 and CI's `setup-uv@v7` is unpinned, so this brings root in line. It couldn't be split into a separate reformat commit because a no-op `uv lock` doesn't write at all. Verified the change set contains exactly 9 version changes and nothing else.

## Verification

#2055 failed at *import*, not resolution, so this was checked past the lock diff:

- `uv sync --extra local-ml` resolves tokenizers **0.22.2** under transformers 5.12.1
- `LocalSTEmbeddings` initializes and returns 384-dim vectors matching its declared dimension
- `LocalSTCrossEncoder` initializes and scores sanely (relevant 9.55 vs irrelevant -11.17)
- `./scripts/hooks/lint.sh` passes
- Local-ML test files: 30 passed, 10 skipped (9 errors are `connection refused` on port 5432 — no local Postgres, unrelated)


dcbouius:
Rebased onto current `main` (`154998701`, includes the v0.8.4 release). The CI failures were all from the stale base, not this change:

- **`test-api` — `test_db_abstraction.py` Oracle session-schema tests** (`_FakeAsyncCursor has no attribute 'fetchone'`, `cursor() must not be called for the public schema`): fixed by #2722 (`fix(test): update Oracle session-schema tests for #2708 reset behavior`), which landed after this PR's original base. The tests were renamed/updated there; both pass locally on the rebased tree.
- **`test-embed-windows` — issue #1240 uvx-fallback regression**: fixed by #2723 (`fix(embed): use --target sibling binary unconditionally`), also post-base.
- **`Core LLM tests` — `test_date_field_calculation_yesterday`**: an LLM-as-judge temporal test, non-deterministic by design (see CLAUDE.md testing conventions). Unrelated to a dependency bump; should clear on re-run.

Re-verified after rebase: the root `uv.lock` re-locks as a no-op with exactly the 3 intended version changes (pydantic-settings 2.14.2, transformers 5.12.1, soupsieve 2.8.4), workspace versions correctly reflect v0.8.4, and the local-ML import/embedding/reranker path still resolves tokenizers 0.22.2 under transformers 5.12.1.

dcbouius:
The remaining CI reds are environmental, not caused by this dependency change:

- **`test-api (2/3)` — 46× `ImportError: sentence-transformers is required`** + `Failed to start embedded PostgreSQL`. `sentence-transformers` is `5.2.0` in this branch's `uv.lock`, **byte-identical to main** — this bump does not touch it, and `transformers 5.12.1` + `sentence-transformers 5.2.0` was verified locally (LocalSTEmbeddings returns 384-dim vectors; LocalSTCrossEncoder scores correctly). The job's venv cache key hashes `uv.lock` + `pyproject.toml`, so this PR is a cache **miss**; shards restore a stale venv via `restore-keys` and reconcile with `uv sync --frozen --all-extras`. Shards 1 and 3 pass; only shard 2 ends up without `sentence-transformers`, alongside an embedded-Postgres start failure — a sharded-venv-cache/infra flake, not a dependency defect.
- **`Core LLM tests`** — LLM-as-judge non-determinism (`test_date_field_calculation_yesterday`, `test_cognitive_epistemic_dimension`, `test_reflect_chinese_content`) plus a Gemini `400 INVALID_ARGUMENT` ("Multiple tools are supported only when they are all search tools") provider error. Unrelated to a lockfile bump.

The dependency change itself is verified: root `uv.lock` re-locks to a no-op with exactly the 3 intended version changes, `npm ci`/local-ML import path all clean. These jobs may need a maintainer re-run on a warm cache, or a look at the sharded-venv caching if shard 2 keeps missing `--all-extras` deps on cache-miss PRs.

PR #2734, opened 2026-07-16
fix(reflect): preserve required tools for custom OpenAI endpoints

Jordan-Jarvis:
## Summary

- preserve `tool_choice="required"` for generic OpenAI-compatible endpoints that use a custom base URL
- continue downgrading only the explicitly identified LM Studio and Ollama providers
- reject named tool choices unless they reference exactly one declared tool

## Why

A custom URL does not identify an endpoint's tool-choice capabilities. The current blanket classification changes a named forced choice into an effectively automatic request. In a multi-turn Reflect loop, that can make the endpoint select a transcript-dependent tool that was removed when Hindsight narrowed the tool list, causing a valid request sequence to fail.

OpenAI-compatible endpoints that implement the contract should receive the caller's required-tool semantics. Nonconforming implementations should use an explicitly identified provider path rather than being inferred from URL presence.

## Validation

- 17 focused OpenAI-compatible/LM Studio tool-choice tests passed
- Ruff check passed
- Ruff format check passed


Jordan-Jarvis:
Merged current upstream `main` into the branch and refreshed formatting under the current Ruff configuration.

- focused Ruff check: passed
- focused Ruff format check: passed
- `tests/test_tool_choice_required_downgrade.py`: 7 passed
- repository pre-commit lint hook: passed

Jordan-Jarvis:
@benfrank241 @dcbouius could one of you review this custom-endpoint tool-choice fix? The branch is current, clean, mergeable, and green, with focused coverage confirming required tools are preserved instead of being silently downgraded. The dependent typed cutover remains draft until this lands.

nicoloboschi (approved):

PR #2737, opened 2026-07-16
refactor(llm): type provider tool choices

Jordan-Jarvis:
## Summary
- replace the internal `str | dict` tool-choice union with a canonical typed value
- serialize provider-specific tool-choice wire shapes only at provider boundaries
- make named-tool validation explicit and remove runtime payload-shape guessing
- preserve provider retry configuration and response-usage tracing from current upstream main

## Dependency
This is intentionally stacked on #2734 because it types the named-tool behavior introduced there. Review the final refactor commit `7a71d8c11`; after #2734 lands, the base will contain the prerequisite behavior and this PR will show only the typed cutover.

## Validation
- Ruff check passed on all 17 changed files
- Ruff format check passed on all 17 changed files
- 41 focused provider/Reflect tests passed sequentially
- engine compileall passed

A broader local file-level run reached 82 passing tests; four unrelated integration/setup cases failed because the local environment had no `.env` API key and two parallel workers timed out waiting on the shared embeddings fixture lock. The focused deterministic suite above is the authoritative change validation.


Jordan-Jarvis:
Updated the stack to current upstream `main` and the refreshed #2734 head. The typed cutover remains draft until #2734 merges so its final review diff stays focused.

- Ruff check and format check across all 17 changed Python files: passed
- focused provider/Reflect suite: 96 passed
- the three real-LLM fixtures requiring `HINDSIGHT_API_LLM_API_KEY` were excluded from this deterministic run

Jordan-Jarvis:
@nicoloboschi Conflict resolution is pushed at `54f14f3d`. Local deterministic provider/reflect suite: 87 passed. Repository lint hook passed. All upstream checks are green and GitHub reports MERGEABLE. Please squash-merge this before #2766; I will immediately re-sync #2766 onto the resulting main.

PR #2738, opened 2026-07-16
fix(api): report effective MarkItDown image OCR capability

Sanderhoff-alt:
## Summary

- Expose `features.markitdown_image_ocr` from `GET /version` so clients can distinguish general file upload support from usable MarkItDown image OCR.
- Report the capability only when file uploads and OCR are enabled, MarkItDown registered successfully, and the parser allowlist permits it.
- Regenerate the OpenAPI specification and Python, TypeScript, and Go client models, and update control-plane feature types.

## Why

The optional MarkItDown OCR integration was controlled by dedicated server configuration, but `/version` only exposed `file_upload_api`. Clients could determine that the upload endpoint existed but could not determine whether image OCR was actually available. The missing capability flag made integrations rely on trial requests and conversion failures.

## Impact

Clients can now check `version.features.markitdown_image_ocr` before offering or invoking MarkItDown image OCR. Existing deployments keep the default value `false` unless the complete runtime capability is available.

## Validation

- API integration tests cover enabled, upload-disabled, allowlist-excluded, parser-unavailable, and OCR-disabled states.
- Python SDK version tests pass.
- TypeScript SDK version tests pass.
- Rust and Go clients build successfully during client generation.
- OpenAPI generation, documentation build, and repository lint pass.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2819,6 +2819,7 @@ class FeaturesInfo(BaseModel):
     bank_config_api: bool = Field(description="Whether per-bank configuration API is enabled")
     bank_llm_health: bool = Field(description="Whether the per-bank LLM connectivity probe is enabled")
     file_upload_api: bool = Field(description="Whether file upload/conversion API is enabled")
+    markitdown_image_ocr: bool = Field(description="Whether MarkItDown image OCR is enabled and available")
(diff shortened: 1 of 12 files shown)


nicoloboschi:
I'd prefer to not expose the markitdown flag. maybe a better alternative is to expose if image (or some format) is supported or not as flag

for now I'll close until we have clear use case

PR #2740, opened 2026-07-16
feat(api): attribute remote reranker calls by bank

### LINKED ISSUE #2729: feat(api): opt-in per-bank attribution for remote reranker requests
## Problem

Hindsight can now attribute OpenAI-compatible LLM and embedding spend per bank through `HINDSIGHT_API_LLM_SEND_BANK_AS_USER` (#1965), but remote reranker requests still carry no bank attribution. Operators routing Cohere-compatible rerank traffic through a metering/provider gateway therefore cannot resolve the originating bank without maintaining a downstream request-correlation patch.

The Cohere-compatible rerank wire format does not expose the OpenAI `user` field, so this needs an explicit HTTP attribution contract rather than a body-shape extension.

## Proposed upstream behavior

Add an opt-in, disabled-by-default setting for remote reranker transports:

- `HINDSIGHT_API_RERANKER_SEND_BANK_AS_HEADER=false`
- fixed request header: `X-Hindsight-Bank-Id: <current-bank-id>`

When enabled, Hindsight would read the existing current-bank `ContextVar` introduced by #1965 and attach the header to outbound remote rerank requests. The setting would be a strict no-op when disabled or when no bank is bound. Documentation would explicitly state that enabling it transmits the bank ID to the configured reranker endpoint and is intended for trusted gateways/proxies.

This should cover the HTTP-backed reranker implementations that Hindsight owns directly (at minimum the Cohere-compatible client and configurable OpenRouter endpoint). Local rerankers remain unchanged.

## Questions for maintainers

1. Is `X-Hindsight-Bank-Id` the preferred stable header name, or should this reuse the existing `X-Bank-Id` spelling used by the MCP surface?
2. Should the switch remain reranker-specific, or should the existing attribution setting be renamed/generalized in a future release?
3. Which remote reranker implementations should be included in the first PR?

I can submit the implementation, focused tests, and configuration documentation once the preferred public contract is confirmed.

Jordan-Jarvis:
## Summary

- add opt-in `HINDSIGHT_API_RERANKER_SEND_BANK_AS_HEADER` configuration
- emit the fixed `X-Hindsight-Bank-Id` header from Hindsight's existing current-bank context
- cover TEI, Cohere-compatible HTTP, LiteLLM proxy, and LiteLLM SDK reranker transports
- keep local rerankers unchanged and omit attribution when disabled or no non-empty bank is bound
- require LiteLLM's documented typed `RerankResponse` instead of accepting guessed legacy response shapes

This resolves #2729. The setting remains disabled by default, and the documentation explicitly limits bank attribution to trusted endpoints because enabling it transmits the current bank ID.

The generic Reflect context correction discovered during qualification is submitted independently as #2764 so it can merge without waiting for the reranker feature. This branch currently contains the same correction; once #2764 lands, GitHub will exclude the identical change from this PR's effective diff.

## Validation

- Ruff format/check passed for all changed Python files
- 66 focused attribution/Cohere/LiteLLM/TEI tests passed; 6 network benchmarks skipped
- 35 hierarchical-config and reranker-timeout tests passed
- 7 embed env-template tests passed
- `ty check --ignore unresolved-import` passed for the changed runtime modules (the repository's optional MLX imports are unavailable on this Linux environment)


Jordan-Jarvis:
Updated the branch to current upstream `main` while preserving the complete reranker bank-attribution stack, including the Reflect bank-binding commit.

- focused Ruff check and format check: passed
- focused bank attribution and reranker suite: 65 passed, 6 environment-dependent tests skipped

Jordan-Jarvis:
@benfrank241 @dcbouius could one of you review this bank-attribution fix? The current head is clean, mergeable, and green, and focused bank/reranker coverage passes. This is required so remote reranker provider usage is attributed through the canonical bank boundary rather than losing tenant ownership.

nicoloboschi (approved):
lgtm

PR #2741, opened 2026-07-16
fix(retain): preserve initial call with zero retry budget

### LINKED ISSUE #2731: fix(retain): a zero retry budget must still perform the initial fact-extraction request
## Problem

`OpenAICompatibleLLM.call(max_retries=N)` correctly interprets `N` as retries after the initial request and loops over `range(N + 1)`.

`_extract_facts_from_chunk`, however, resolves the same `retain_llm_max_retries` / `llm_max_retries` value and uses it as an attempt count:

```python
for attempt in range(llm_max_retries):
```

When an operator deliberately configures `retain_llm_max_retries=0` because an upstream gateway owns transport retries, Hindsight performs **zero fact-extraction requests** and raises `Fact extraction failed after 0 attempts`.

The same setting is also passed to `llm_config.call(max_retries=llm_max_retries)`, so its meaning differs between the two nested layers.

## Expected behavior

A retry budget of zero must mean one initial request and no retry. Hindsight should preserve a single retry owner without skipping the operation.

## Proposed focused fix

Use at least one outer attempt while preserving the configured transport retry value passed to the provider, and add regression coverage for `retain_llm_max_retries=0`.

A broader follow-up may want to separate content-quality attempts from transport retries explicitly, but the zero-budget behavior is a correctness bug on its own and blocks trusted gateway deployments.

I can submit the focused patch and test.

koriyoshi2041:
## Summary

- make fact extraction perform one initial request when `retain_llm_max_retries=0`
- keep forwarding the configured zero transport retry budget to the provider
- report the actual outer attempt count in the exhausted-attempt fallback

Closes #2731.

## Validation

- `uv run --package hindsight-api-slim pytest hindsight-api-slim/tests/test_fact_extraction_retry.py -q` (13 passed)
- `uv run --package hindsight-api-slim ruff check hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py hindsight-api-slim/tests/test_fact_extraction_retry.py`
- `uv run --package hindsight-api-slim ruff format --check hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py hindsight-api-slim/tests/test_fact_extraction_retry.py`
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -1362,10 +1362,11 @@ async def _extract_facts_from_chunk(
     llm_max_retries = (
         config.retain_llm_max_retries if config.retain_llm_max_retries is not None else config.llm_max_retries
     )
+    extraction_attempts = max(1, llm_max_retries)
     last_error: Exception | None = None
 
     usage = TokenUsage()  # Track cumulative usage across retries
-    for attempt in range(llm_max_retries):
+    for attempt in range(extraction_attempts):
         try:
             initial_backoff = (
                 config.retain_llm_initial_backoff
@@ -1672,7 +1673,7 @@ async def _extract_facts_from_chunk(
     # If we exhausted all retries, raise the last error or a descriptive fallback
     if last_error is not None:
         raise last_error
-    raise RuntimeError(f"Fact extraction failed after {llm_max_retries} attempts: LLM did not return valid JSON")
+    raise RuntimeError(f"Fact extraction failed after {extraction_attempts} attempts: LLM did not return valid JSON")
 
 
 async def _extract_facts_with_auto_split(
--- a/hindsight-api-slim/tests/test_fact_extraction_retry.py
+++ b/hindsight-api-slim/tests/test_fact_extraction_retry.py
@@ -264,6 +264,33 @@ async def test_retain_llm_max_retries_overrides_global():
 
 
 @pytest.mark.asyncio
+async def test_zero_retain_retry_budget_still_makes_initial_call():
+    """A zero transport retry budget must not skip fact extraction."""
+    from hindsight_api.engine.retain.fact_extraction import _extract_facts_from_chunk
+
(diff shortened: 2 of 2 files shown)


ebarkhordar (commented):
The retry change itself is two files and lines up with what #2731 asks for. One note on the branch, not on the code.

This branch is stacked on your #2669 but targets `main`. The head's only parent is #2669's head, which is not on main yet:

```
$ gh api repos/vectorize-io/hindsight/commits/435874aa3 --jq ".parents[].sha"
ab9a9546db395b780a4997688336578000004fb7      # head of #2669, unmerged

$ gh api repos/vectorize-io/hindsight/compare/main...ab9a9546d --jq ".status"
diverged
```

That is where the `"strict_schema": True` hunk in `consolidation/consolidator.py` and its test come from. They are #2669's change, and a zero retry budget has nothing to do with them. So merging this into `main` as it stands would also land #2669, reviewed as a side effect of a retry fix rather than on its own.

Retargeting this PR's base to `rios/consolidation-strict-schema` would show only the retry commit here and keep the two landing in order. Rebasing onto `main` after #2669 merges does the same. Either way #2669 keeps its own review.

Flagging it because the two PRs are easy to merge out of order, not because anything in the retry fix looks wrong to me.

For transparency: I used an AI agent to check this. The two SHAs above are from the API calls shown, and I have not run the test suite here.

koriyoshi2041:
Good catch. I rebased this branch directly onto current `main`, so #2741 now contains only the two-file zero-retry fix; the #2669 strict-schema change is no longer in its history or diff.

New head: `c1cf2357`

Revalidated:
- `uv run --package hindsight-api-slim pytest hindsight-api-slim/tests/test_fact_extraction_retry.py -q` — 13 passed
- focused Ruff check and format check

nicoloboschi (changes requested):
uhm this fix is wrong. with llm_max_retries=1, we should do 1 retry still. so I think we should just do llm_max_retries+1 attempts - please also check the impl in llm_config.call follow the same pattern

nicoloboschi:
closing for https://github.com/vectorize-io/hindsight/pull/2779 which is slightly better

PR #2753, opened 2026-07-16
fix(retain): enforce entity extraction contract

### LINKED ISSUE #2749: [BUG] Auto-extracted entities are never persisted (v0.8.4): prompt teaches string arrays while Entity requires objects, and the field is optional
### Summary

On v0.8.4 with the default `concise` extraction mode, auto-extracted entities are silently never persisted: `entities`, `unit_entities` and `entity_cooccurrences` all stay at 0 while retain reports success and recall keeps working. The knowledge graph is simply empty, with no error anywhere.

### Impact

Graph expansion and entity-based retrieval never engage. Recall silently degrades to semantic + BM25 only. Nothing fails loudly, so this is easy to run in production for a long time without noticing — the retain docs promise automatic entity/relationship discovery, which makes this a behavior-vs-docs defect rather than a config mistake.

### Root cause

The extraction contract is internally inconsistent in `engine/retain/fact_extraction.py`:

1. `entities` is `list[Entity] | None = Field(default=None)` and is absent from `json_schema_extra["required"]` in **all four** extraction models (`ExtractedFact`, `ExtractedFactVerbose`, `ExtractedFactNoCausal`, `VerbatimExtractedFact`) — so omitting it is legal.
2. The prompt's few-shot examples teach a **flat string array**:
   ```
   entities=["Alice", "Kubernetes", "CKA"]
   ```
   while the `Entity` model requires **objects**:
   ```
   entities=[{"text": "Alice"}]
   ```
3. The parser reads only `llm_fact.entities`, with no fallback to the already-required `who`/`where`, so when the field is omitted the resolver receives nothing.

In our provider path (`litellm` → Gemini) the retain request carries no response schema at all — extraction is prompt-driven, so those examples are the **only** contract the model sees. It follows them, returns strings or omits the field, and the entities are dropped.

### Why this isn't universally reported

Stronger models appear to infer the object form despite the examples. An older instance of ours on `gemini-3.5-flash` produced 214 entities / 372 unit_entities / 255 cooccurrences from a comparable corpus. The same corpus on `gemini-3.1-flash-lite` (ranked #2/21 on your own retain leaderboard, 100% schema conformance) produces **0 / 0 / 0**. The defect surfaces exactly on the models that follow the prompt most literally.

### Environment

- Hindsight `v0.8.4` (self-hosted, digest-pinned), pgvector/pg18
- `HINDSIGHT_API_LLM_PROVIDER=litellm`, model `gemini/gemini-3.1-flash-lite`
- `retain_extraction_mode=concise` (default)
- Reproduces with `HINDSIGHT_API_LLM_STRICT_SCHEMA` both `true` and `false`

### Ruled out by measurement

Bank mission (a bank without one also yields 0), extraction mode, strict-schema on/off, and the proxy (verified it does not strip the schema — the request never contains one). Called directly through the same proxy, the same model **does** populate the optional `entities` field when asked with an object-shaped contract.

### Suggested fix

Make the contract consistent — one file, `fact_extraction.py`:

- `entities: list[Entity] = Field(default_factory=list, ...)` in all four models, and add `"entities"` to each `json_schema_extra["required"]` (empty `[]` stays valid, so nothing hard-fails);
- switch the prompt examples to object form and state it explicitly: *"Always return the `entities` field as an array of objects with a `text` field, never as an array of strings."*

### Verification

With exactly that change applied over the official v0.8.4 image (same corpus, same model, same provider): **entities 30, unit_entities 72, entity_cooccurrences 40**, attached to 3/3 memory units in a focused probe; counts identical across a restart, and a backup/restore drill reproduces them. Recall quality is unchanged.

Happy to open a PR with the patch and the accompanying regression test if useful.

oldnicke:
## Summary

- require object-shaped `entities` arrays in every fact-extraction response schema while retaining an empty-list fallback
- update concise examples and shared/verbose instructions to match the `Entity` model contract
- add deterministic schema/prompt regression coverage and a real-LLM entity extraction judge test

Closes #2749

## Testing

- `UV_CACHE_DIR=/tmp/hindsight-uv-cache ./scripts/hooks/lint.sh`
- `uv run pytest tests/test_fact_extraction_entity_contract.py tests/test_fact_extraction_fact_type_prompt.py -m "not hs_llm_core" -q` (7 passed)
- `uv run pytest tests/test_entity_labels.py -m "not hs_llm_core" -k "not entity_resolution_does_not_merge_distinct_label_values" -q` (76 passed)
- real extraction produced `Alice`, `Kubernetes`, `Acme Corporation`, and `Berlin`; the local Gemini judge endpoint timed out, so the committed `hs_llm_core` test is left for the standard quality CI environment

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -180,7 +180,7 @@ class ExtractedFact(BaseModel):
 
     model_config = ConfigDict(
         json_schema_mode="validation",
-        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type"]},
+        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type", "entities"]},
     )
 
     what: str = Field(description="Core fact - concise but complete (1-2 sentences)")
@@ -195,7 +195,7 @@ class ExtractedFact(BaseModel):
     fact_type: Literal["world", "assistant"] = Field(
         description="'world' = objective/external facts, including user preferences, rules, corrections, and constraints even when stated during a conversation. 'assistant' = actions, experiences, or observations the assistant/agent actually performed."
     )
-    entities: list[Entity] | None = Field(default=None, description="People, places, concepts")
+    entities: list[Entity] = Field(default_factory=list, description="People, places, concepts")
     causal_relations: list[FactCausalRelation] | None = Field(
         default=None, description="Links to previous facts (target_index < this fact's index)"
     )
@@ -286,7 +286,7 @@ class ExtractedFactVerbose(BaseModel):
 
     model_config = ConfigDict(
         json_schema_mode="validation",
-        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type"]},
+        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type", "entities"]},
     )
 
     what: str = Field(
@@ -348,8 +348,8 @@ class ExtractedFactVerbose(BaseModel):
         description="'world' = objective/external facts about the user, other people, events, general knowledge, preferences, rules, corrections, or constraints. 'assistant' = actions, experiences, or observations the assistant/agent actually performed (e.g., 'I changed X', 'I discovered Y')."
     )
 
-    entities: list[Entity] | None = Field(
-        default=None,
+    entities: list[Entity] = Field(
+        default_factory=list,
         description="Named entities, objects, AND abstract concepts from the fact. Include: people names, organizations, places, significant objects (e.g., 'coffee maker', 'car'), AND abstract concepts/themes (e.g., 'friendship', 'career growth', 'loss', 'celebration'). Extract anything that could help link related facts together.",
     )
 
@@ -378,7 +378,7 @@ class ExtractedFactNoCausal(BaseModel):
 
     model_config = ConfigDict(
         json_schema_mode="validation",
-        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type"]},
+        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type", "entities"]},
     )
 
     # Same fields as ExtractedFact but without causal_relations
@@ -397,8 +397,8 @@ class ExtractedFactNoCausal(BaseModel):
     fact_type: Literal["world", "assistant"] = Field(
         description="'world' = about the user/others, including user preferences, rules, corrections, and constraints. 'assistant' = actions or experiences the assistant/agent actually performed."
     )
-    entities: list[Entity] | None = Field(
-        default=None,
+    entities: list[Entity] = Field(
+        default_factory=list,
         description="Named entities, objects, and concepts from the fact.",
     )
 
@@ -426,7 +426,7 @@ class VerbatimExtractedFact(BaseModel):
 
     model_config = ConfigDict(
         json_schema_mode="validation",
-        json_schema_extra={"required": ["when", "where", "who", "fact_type"]},
+        json_schema_extra={"required": ["when", "where", "who", "fact_type", "entities"]},
     )
 
     when: str = Field(description="When it happened. 'N/A' if unknown.")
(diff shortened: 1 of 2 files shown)


benfrank241:
Verified the deterministic surface locally on the PR head:

- `test_fact_extraction_entity_contract.py` (non-LLM): **5/5 pass** — `entities` is now in `required` for all four extraction models, `default_factory=list` keeps empty `[]` valid (nothing hard-fails), and the concise prompt teaches the object form (`entities=[{"text": "Alice"}, ...]`) with the explicit "never as an array of strings" instruction.
- `test_fact_extraction_fact_type_prompt.py` + `test_entity_labels.py` (non-LLM): **78/78 pass** — no regression from making `entities` required.

Confirmed the root-cause mechanism too: `ensure_entities_list` only maps `None → []` (no string→object coercion), so on `main` a literal-following model either omits the field (optional → `[]`) or returns bare strings that fail `list[Entity]` validation — both silently drop entities, exactly as #2749 measured.

Routing the merge to @nicoloboschi since this changes the **core fact-extraction prompt** (accuracy-sensitive — retain leaderboard, longmemeval/locomo). Two things want your call that I could not settle locally:
1. The behavioral guarantee lives in the committed `hs_llm_core` judge test (`test_fact_extraction_returns_relevant_entities`); the author noted the local Gemini judge timed out, so it needs the standard quality CI run to confirm the model actually persists entities now.
2. Making `entities` **required** interacts with strict-schema provider mode, so a provider-matrix (`hs_llm_mat`) pass would be worth it before merge.

Fix is minimal, well-targeted, and low-risk (empty-list fallback), so this should be a fast approve once the LLM jobs are green. cc @nicoloboschi

ebarkhordar (commented):
Confirmed the deterministic core on the PR head (0056a04f) in a clean container: `entities` lands in `required` for the four static models, `default_factory=list` keeps the empty case valid, and `ensure_entities_list` still absorbs a `null`. Given `Entity` has no string coercion, the prompt-example change looks like the part that actually moves the needle.

One mode is not covered by the tests here, and it is the strict-schema interaction @benfrank241 flagged above: labels-only mode (`entity_labels` set, `entities_allow_free_form=false`).

`_build_extraction_prompt_and_schema` inherits the base `required` list into the dynamic `LabelsFact` (fact_extraction.py:1157-1164), but when free-form entities are off it re-declares the field back to optional (fact_extraction.py:1150-1154):

```python
dynamic_fields["entities"] = (
    list[Entity] | None,
    Field(default=None, description="Leave empty — labels-only mode"),
)
```

So in that one mode the two halves of this PR disagree: `required` gains `entities`, but the type stays nullable and never picks up the `default_factory`.

What I ran, in a clean python:3.12 container, calling `_build_extraction_prompt_and_schema` with one `value` label group and `entities_allow_free_form=False`, on `main` (9676fc1) and on this branch:

```
main      required: ['what', 'when', 'where', 'who', 'why', 'fact_type', 'labels']
this PR   required: ['what', 'when', 'where', 'who', 'why', 'fact_type', 'entities', 'labels']
```

The `entities` property itself is identical on both sides: `anyOf [array, null]`, `default: null`, description "Leave empty".

The prompt built from that same config now carries both of these lines:

```
Always return the "entities" field as an array of objects with a "text" field, never as an array of strings.
Do NOT add regular named entities — labels-only mode.
```

I do not think this corrupts stored data: non-label entities are filtered out downstream at fact_extraction.py:1576-1582 (batch path at 2335-2340), so whatever a model returns there is dropped. It is the schema and the prompt asking for something that mode explicitly does not want, on a prompt you have already called accuracy-sensitive.

So this is a question, not a change request: is requiring `entities` in labels-only mode intended? If it is not, keeping it out of the inherited list looks sufficient:

```python
json_schema_extra={"required": [*(r for r in base_required if free_form_entities or r != "entities"), "labels"]},
```

I ran that variant in the same container: labels-only goes back to `[..., 'fact_type', 'labels']`, and the free-form path still gets `entities` required.

What I did not verify: any live LLM call in labels-only mode, so I have not measured whether a model behaves differently with the flag set. I only inspected the generated schema and the prompt string. I also did not run the full suite, just the schema and prompt build on both sides.

AI assistance: I investigated and verified this with AI tooling, and I have checked every claim above myself.

oldnicke:
Addressed the labels-only schema concern from review in commit `47ab3e40b`.

When `entities_allow_free_form=false`, the generated `LabelsFact` schema now removes `entities` from `required` while retaining the existing nullable/empty behavior. Free-form entity mode still requires the object-shaped `entities` array. Added a parameterized regression test for both modes.

Validation: 9 entity-contract/prompt tests, 76 non-LLM entity-label tests, Ruff, ty, repository lint, and pre-commit all pass.

nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/2830

PR #2757, opened 2026-07-16
fix(reflect): cap done tool answers

### LINKED ISSUE #2756: [BUG] mental model max_tokens not enforced when reflect agent finishes via the done tool (the common path)
## Summary

`max_tokens` on a mental model is forwarded to the refresh reflect call (#1076), and the reflect agent does cap its *forced-final* and *direct-text short-circuit* paths — but the **`done`-tool path**, which is how the agent normally finishes, bypasses the cap entirely. In practice mental model content routinely exceeds `max_tokens` by 4-6x.

## Environment

- `hindsight-api` 0.8.4 (pip), embedded Postgres, LLM provider `openai` / `gpt-4.1-nano` (no custom base_url)

## Reproduction

1. Create a mental model with the minimum allowed cap:

```bash
curl -X POST "$BASE/v1/default/banks/$BANK/mental-models" \
  -d '{"id":"maxtok-test","name":"maxtok test",
       "source_query":"<any query over a reasonably populated bank>",
       "max_tokens":256,
       "trigger":{"refresh_after_consolidation":false,"exclude_mental_models":true}}'
```

2. Wait for the initial refresh to complete, then `GET .../mental-models/maxtok-test?detail=content`.

**Expected:** content ≤ ~256 tokens.
**Actual:** content of ~4,600 chars (~1,150+ tokens with cl100k), well-formed (not truncated) — i.e. the cap was never applied. Reproduced consistently; also observed with `max_tokens=800` producing ~1,360 tokens.

## Root cause (0.8.4 sources)

The chain is intact up to the last step:

- `MemoryEngine.refresh_mental_model` forwards the stored `max_tokens` to `reflect_async` (`memory_engine.py`, "Forward the per-model max_tokens so the final synthesis is capped..." — the #1076 fix).
- `reflect_async` forwards it to `run_reflect_agent`, whose forced-final synthesis calls pass `max_completion_tokens=max_tokens`, and whose direct-text short-circuit even rewrites the answer when `count_cl100k_tokens(answer) > max_tokens` (`engine/reflect/agent.py`, comment: "must respect max_tokens like the forced-final paths do").

But when the agent finishes via the **`done` tool** — the normal completion path — the answer is taken verbatim from the tool-call arguments:

```python
# engine/reflect/agent.py — _process_done_tool
raw_answer = args.get("answer", "").strip()
```

Those arguments come from `call_with_tools`, which is *intentionally uncapped* (per the in-code comment, to leave headroom for tool-call JSON). `_process_done_tool` performs **no token check and no rewrite**, unlike the short-circuit path. So the only enforcement points are the two paths the agent rarely takes.

| Agent completion path | max_tokens enforced? |
|---|---|
| Forced final (max iterations) | yes (`max_completion_tokens`) |
| Direct-text short-circuit | yes (rewrite if overshoot) |
| **`done` tool (normal path)** | **no** |

## Suggested fix

In `_process_done_tool`, mirror the short-circuit safeguard: if `max_tokens` is set and `count_cl100k_tokens(answer) > max_tokens`, run the same single capped rewrite call before returning the result.

## Impact

- Mental model content cannot be reliably bounded, which matters for agent-boot injection flows with hard context budgets (e.g. Claude Code hooks cap output at 10,000 chars).
- Possibly related in spirit to #2501 (unbounded mental-model growth), though that one is about delta-refresh embedding overflow — this is about the reflect answer cap itself.

Sanjays2402:
Fixes #2756.

The normal `done` tool path now applies the same capped rewrite used by direct-text completions when an answer exceeds `max_tokens`. The rewrite's token usage and trace entry are included in the final result.

Regression coverage confirms the done-tool answer is uncapped before this change and rewritten with the configured budget afterward.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -1035,6 +1035,7 @@ async def run_reflect_agent(
                     directives_applied=directives_applied,
                     llm_config=llm_config,
                     response_schema=response_schema,
+                    max_tokens=max_tokens,
                 )
 
         # Execute other tools in parallel (exclude done tool in all its format variants)
@@ -1244,6 +1245,7 @@ async def _process_done_tool(
     directives_applied: list[DirectiveInfo],
     llm_config: "LLMProvider | None" = None,
     response_schema: dict | None = None,
+    max_tokens: int | None = None,
 ) -> ReflectAgentResult:
     """Process the done tool call and return the result."""
     args = done_call.arguments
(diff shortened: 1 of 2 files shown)


benfrank241:
Thanks @Sanjays2402 — and nice root-cause report on #2756.

I verified this locally: the fix faithfully mirrors the existing short-circuit rewrite path (same prompt, `scope`, and `max_completion_tokens=max_tokens`), plumbs `max_tokens` into `_process_done_tool` correctly, and threads `final_usage` + the `final_rewrite` trace entry through cleanly. `test_reflect_agent.py` passes 57/57 deterministic tests including the new `test_done_tool_answer_respects_max_tokens` (the 3 errors are just `LLM API key required` on the real-LLM classes, which run in the core-LLM CI job).

Routing to @nicoloboschi for the merge call, because this is a behavior decision as much as a bug fix. Today two of the three final-answer paths (forced-final, direct-text short-circuit) enforce the cap; the done path — the common one — does not. Applying the cap here means an **extra LLM round-trip + a lossy summarization rewrite on the hot path**, turning `max_tokens` into a hard, rewrite-enforced contract rather than a soft hint. It is scoped to mental-model refresh (the only place `max_tokens` is populated, per #1076), so it only fires where a user opted into a budget — but whether that budget should be hard-enforced via a second call vs. best-effort is a product call for you.

Fix looks correct and ready if you want it; flagging the design tradeoff so the decision is yours. cc @nicoloboschi

nicoloboschi (approved):

PR #2760, opened 2026-07-17
fix(retain): preserve structured envelope fields across oversized chunks

### LINKED ISSUE #2548: Smarter json chunking to prevent loss of critical context
### Use Case

I'm using hindsight to ingest agent session files as jsonl.

### Problem Statement

I think that #2113 and #2409 are notable improvements for extraction. There are still some cases where single messages are so long they will be above HINDSIGHT_API_RETAIN_STRUCTURED_CHUNK_SIZE and split. Take this example jsonl:

```json
{"timestamp":"2026-06-15T17:22:29.716Z","message":{"role":"user","content": [<very long...>]}}
```

If it’s split as text, some useful context like the role and timestamp are lost for extraction for all later chunks.

In cases where the timestamp is not included, you can get a very incorrect occurred at time for sessions that span days.

Smarter chunking could keep the short fields with each chunk. 

The role in particular can be useful context to have in every chunk to prevent misattribution, and timestamp is needed for accurate occurred at information.

### Proposed Solution

I don’t know how complicated is too complicated. You could try to split within content keeping full sub-messages of the json, e.g. chunk 1:
```json
{"timestamp":"2026-06-15T17:22:29.716Z","message":{"role":"user","content": [{message 1}...{message 3}]}}
```

Chunk 2:
```json
{"timestamp":"2026-06-15T17:22:29.716Z","message":{"role":"user","content": [{message 4}...{message 6}]}}
```

#### More edge cases
How would it know where to split? I guess it could expect some standard format, or it could be somehow specified in the API request. 

That may be too complex, especially since at some point you will have to split earlier if it’s too long, but even just this might be a beneficial improvement:
```json
{"timestamp":"2026-06-15T17:22:29.716Z","message":{"role":"user","content": "<split arbitrarily to limit>"}}
```

However, there is still some useful information that would be lost doing that. For example:

```json
{"type":"message","id":"1d908012","parentId":"237c3b90","timestamp":"2026-06-20T01:54:38.248Z","message":{"role":"assistant","content":[...,{"type":"toolCall","id":"functions.write:21","name":"write","arguments":{"content":"<very long content...>","path":"/path/to/file"}}]}}
```

In this case I would like to keep the path available for every chunk that has content from that file write because I have different rules/caveats in my retain mission for certain files or directories.

### Alternatives Considered

I could split the jsonl on the client side, but then only my client could benefit from this improvement, and it doesn't seem like a good idea to try to pre-chunk. Maybe some hybrid solution would work.

### Priority

### Additional Context

_No response_

### Checklist

r266-tech:
## Summary

- split oversized structured records inside one unambiguous payload path and re-wrap every fragment as valid JSON
- preserve message, tool-call, and path context such as `role`, `timestamp`, `id`, and `path`, while retaining the plain-text fallback for ambiguous shapes
- bound recursion, fragment count, output expansion, and serialized wrapper overhead so chunks remain lossless and idempotent

## Testing

- `uv run pytest tests/test_chunking.py -q` (44 passed)
- `uv run pytest tests/test_structured_chunk_size_retain.py -q -n 0`
- `uv run --frozen ty check hindsight_api`
- `./scripts/hooks/lint.sh`

Fixes #2548

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -9,6 +9,8 @@ import asyncio
 import json
 import logging
 import re
+from collections import Counter
+from collections.abc import Callable
 from datetime import datetime, timedelta
 from typing import Any, Literal, cast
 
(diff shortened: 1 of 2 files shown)


benfrank241:
Thanks @r266-tech — thorough work, and the test coverage (losslessness, idempotency, plain-text fallback for ambiguous shapes, envelope preservation) is exactly right for this kind of change.

Verified locally on the PR head:
- `tests/test_chunking.py`: **44/44 pass**
- `tests/test_structured_chunk_size_retain.py`: **1/1 pass** (`test_jsonl_line_over_chunk_size_retains_without_collision`)

Routing the merge decision to @nicoloboschi rather than merging directly, for two reasons:
1. **This is a feature, not a bug fix.** #2548 is filed as a feature request ("Smarter json chunking"), and the PR implements new structured-splitting behavior (~500 new lines: envelope-key heuristics, recursive JSON-aware splitting with re-wrapping, and the depth/fragment/output-expansion bounds).
2. **It changes the retain chunking path**, which is accuracy-sensitive (feeds extraction/attribution quality on the retain leaderboard + longmemeval/locomo). The envelope-key sets and shape-detection (`_MESSAGE_*_ENVELOPE_KEYS`, `_TOOL_CALL_TYPES`, etc.) are effectively contract decisions worth an owner’s eye, especially how they behave on session formats other than the Claude/OpenAI shapes in the tests.

No blocking concerns from me — the diff is well-bounded (recursion/fragment/expansion caps), lint + `ty` are clean per the PR description and my run, and it degrades to the existing plain-text split for anything ambiguous. Flagging it as a design-review item so the heuristics and forward-compat are your call. cc @nicoloboschi

PR #2763, opened 2026-07-17
fix(reflect): preserve bank attribution on provider calls

Jordan-Jarvis:
## Summary

- bind the existing bank-attribution context across `MemoryEngine.reflect_async`
- preserve `user=<bank_id>` on every OpenAI-compatible Reflect provider call
- add a behavioral regression for binding and reset on early failure

## Root cause

Retain and recall already enter the shared `_bind_bank_id` boundary, but Reflect did not. When `HINDSIGHT_API_LLM_SEND_BANK_AS_USER=true`, outbound Reflect calls therefore omitted `user` even though they were bank-scoped. Gateways that require canonical bank attribution rejected those calls.

This reuses the existing async-safe ContextVar boundary rather than adding a Reflect-specific provider path.

## Validation

- `cd hindsight-api-slim && uv run pytest tests/test_bank_attribution.py -q` (15 passed)
- repository pre-commit lint suite passed
- `uv run ruff check hindsight_api/engine/memory_engine.py tests/test_bank_attribution.py`
- `uv run ruff format --check hindsight_api/engine/memory_engine.py tests/test_bank_attribution.py`
- `uv run ty check hindsight_api`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -194,6 +194,9 @@ HINDSIGHT_API_LOG_LEVEL=info
 # Reranker Configuration (Optional - uses local by default)
 # Provider: "local" (default) or "tei" (HuggingFace Text Embeddings Inference)
 # HINDSIGHT_API_RERANKER_PROVIDER=local
+# Trusted gateway attribution (disabled by default). When enabled, remote
+# reranker requests include X-Hindsight-Bank-Id with the current bank ID.
+# HINDSIGHT_API_RERANKER_SEND_BANK_AS_HEADER=false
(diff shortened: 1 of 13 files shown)


Jordan-Jarvis:
Superseded by the focused standalone fix in #2764. This PR shared the broader reranker-attribution branch with #2740, so keeping both open duplicated the same 13-file head and obscured the review boundary. #2764 contains only the Reflect bank-context correction; #2740 remains the remote-reranker attribution change.

PR #2764, opened 2026-07-17
fix(reflect): preserve bank attribution in provider calls

Jordan-Jarvis:
## Summary

- bind `MemoryEngine.reflect_async` to the existing per-bank provider-attribution context
- preserve the bank for Reflect tool calls and final synthesis
- verify the bank is bound during Reflect and reset after failure
- replace two lower-value direct `ContextVar` implementation tests with one integrated Reflect regression

Without this binding, retain and recall attribute provider usage correctly but Reflect is the only bank-scoped engine operation that loses the bank before its LLM/tool loop.

## Validation

- `uv run pytest tests/test_bank_attribution.py -q` — 15 passed
- `uv run ruff check hindsight_api/engine/memory_engine.py tests/test_bank_attribution.py`
- `uv run ruff format --check hindsight_api/engine/memory_engine.py tests/test_bank_attribution.py`
- patch applies cleanly to current upstream `main` (`9676fc1699c199175809285ea0e1095ab71115d0`)

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -71,7 +71,7 @@ from .sql import SQLDialect, create_sql_dialect
 _current_schema: contextvars.ContextVar[str | None] = contextvars.ContextVar("current_schema", default=None)
 
 # Context variable for the bank an operation runs for (async-safe, per-task isolation).
-# Set by the engine wherever it learns the bank (recall/retain/batch/task execution) so
+# Set by the engine wherever it learns the bank (recall/retain/batch/reflect/task execution) so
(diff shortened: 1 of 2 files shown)


Jordan-Jarvis:
Additional integration evidence from the corresponding Will qualification branch:

- the same bank-binding correction merged in `Will-Augmented-Thought/hindsight` as `d693efd0f644ce75c5be93f26e73e19d318d9cc7`;
- exact-main image run `29551436887` published, scanned, signed, and verified both API and control-plane artifacts;
- GitOps promotion `29554113973` passed dev retain, recall, and canonical per-bank provider attribution through native Reflect. The later failure was the independent noncanonical OpenAI tool-result message shape addressed by #2766, not bank attribution.

This validates that binding `reflect_async` through the existing bank ContextVar boundary preserves one bank-owned routing contract without adding provider-specific inference downstream.

PR #2765, opened 2026-07-17
fix(reflect): emit canonical tool result messages

Jordan-Jarvis:
## Summary
- remove the non-standard `name` field from Reflect role=`tool` messages
- preserve the canonical OpenAI tool-result shape at the shared Reflect producer boundary
- derive Gemini `FunctionResponse.name` from the immediately preceding assistant tool calls by `tool_call_id`
- reject unknown, duplicate-within-turn, and missing tool results while allowing providers that reuse call IDs across turns

## Root cause
Reflect currently adds `name` to tool-result messages for Gemini. Strict OpenAI-compatible gateways reject that extra field because tool-result messages permit `content` and `tool_call_id`; provider-specific requirements should be handled in the provider adapter instead of changing the canonical message shape.

## Verification
- targeted Ruff on all touched files
- `pytest -q tests/test_reflect_agent.py tests/test_gemini_safety_settings.py -k 'not RealLLM and not ContextOverflowIntegration'` (`69 passed`)
- repository pre-commit hooks and lint passed
- `git diff --check`

No second Reflect path or compatibility fallback is introduced.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/gemini_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/gemini_llm.py
@@ -578,12 +578,17 @@ class GeminiLLM(LLMInterface):
         # Convert messages
         system_instruction = None
         gemini_contents = []
+        pending_tool_names_by_call_id: dict[str, str] = {}
         msg_list = list(messages)
         i = 0
         while i < len(msg_list):
             msg = msg_list[i]
             role = msg.get("role", "user")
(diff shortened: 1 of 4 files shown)


Jordan-Jarvis:
Superseded by #2766, which carries the same canonical producer/adapter fix with the stricter unknown-ID and duplicate-within-turn validation coverage. Keeping one upstream contribution avoids split implementations and duplicate review.

Issue #2675, opened 2026-07-13
Crash-interrupted operations are re-claimed indefinitely — claim attempts never count toward max_attempts (poison-pill regrind loop)


kubaodias:
## Summary

When processing an `async_operations` row crashes the worker (e.g. OOM), no failure bookkeeping runs — `retry_count` is only incremented by in-process failure handling. On restart, `recover_own_tasks` resets the worker's own `processing` rows back to `pending` with `retry_count` untouched, and the row is re-claimed as if brand new. An operation that can never complete within one attempt therefore loops forever: claim → grind → crash → recover → re-claim → grind…

Notably, the fix for #991 (stable worker IDs making self-recovery reliable, plus the warnings from #2366/#2383) makes this loop *more* reliable: recovery without an attempt budget converts a poison operation from "stuck once" into "reground indefinitely". Recovery and quarantine need to come as a pair.

## Observed (production, hindsight-api-slim 0.8.4, PostgreSQL, K8s, multi-worker)

A tenant submitted a batch retain that fans out to ~6,624 fact-extraction LLM calls per attempt (~2.9h of work per attempt at full concurrency). It never completed, and the system reground it **54 times over ~22 hours**:

- ~394K LLM calls, ~863M input + ~211M output tokens spent; only ~176 memory units ever written (~0% useful yield).
- The flood saturated the shared LLM backend (avg call latency 12s → 47s), starving retains for **all other banks** — fleet-wide backlog collapse (stuck-pending fraction went 0% → 99% within 3h of onset).
- Worker pods OOM-looped against their 2Gi limit; each OOM restart triggered self-recovery and another full regrind (the recurring ~75–79min gaps between attempt waves match the crash cycles).
- After 54 attempts, the rows still showed `retry_count=0`.
- At least 21 of the 54 attempts happened *after* the client had stopped submitting anything — pure re-claim churn, unattended, overnight.
- Recovery required manual intervention: `UPDATE async_operations SET status='cancelled' WHERE status IN ('pending','processing') ...` + rolling restart. The instant the rows were cancelled, call volume collapsed ~20K/hr → ~100/hr.

Diagnosis was made via the (excellent) `llm_requests` table: grouping by `bank_id` found the source; grouping by `trace_id` found 54 traces with exactly 6,624–6,625 calls each.

Related: #1571 is fixed — the oversized item *was* properly chunked into ~2.2K-token sub-calls this time. Chunking worked; the unbounded re-claim is what remained.

[...]

PR #2777, opened 2026-07-17
feat(api): allow deleting terminal bank operations with control-plane support

### LINKED ISSUE #2677: (Control Plane) Add route to delete Failed operation
### Use Case

I'd like to Cancel (or delete) failed operations

### Problem Statement

I don't have an option to clean failed list

Easily remove not needed failed operations

### Proposed Solution

Add link to related Document form where it could be deleted
Delete related failed or pending operations

### Priority

Nice to have

### Additional Context

_No response_

### Checklist

chethanuk:
## Problem

There is no way to remove terminal (`failed` / `cancelled` / `completed`) async operation rows. `DELETE …/operations/{id}` is bound to **cancel**, which only flips `pending → cancelled` and returns 409 for anything else — the row never leaves the table, so failed operations pile up in the control plane with no cleanup path and no navigation to the document they produced.

Fixes vectorize-io/hindsight#2677

## Fix

- New dataplane route `DELETE /v1/default/banks/{bank_id}/operations/{operation_id}/delete`: status-guarded hard delete for terminal states only, implemented as a single bank-scoped `DELETE … RETURNING` (no TOCTOU window with a concurrent retry). 404 when missing, 409 for non-terminal states. The existing cancel route is untouched.
- Engine: `delete_operation` + `BankWriteOperation.DELETE_OPERATION` (operation-validator aware).
- Control plane: delete buttons in the operations table and detail dialog (terminal rows only) + a **View Document** action whenever `result_metadata.document_id` is present — the navigation path the issue reporter asked for.
- OpenAPI spec and Python/TypeScript/Go client surfaces regenerated.

```mermaid
flowchart LR
    subgraph cp [Control plane]
        U["Operations table / dialog<br/>(Delete on terminal rows)"] --> C["client.deleteOperation()"]
        C --> B["BFF DELETE /api/banks/{bankId}/operations/{id}"]
    end
    B --> S["SDK deleteOperation"]
    S --> R["DELETE …/operations/{id}/delete"]
    R --> G{"status ∈ failed |<br/>cancelled | completed?"}
    G -->|yes| D["single bank-scoped<br/>DELETE … RETURNING"]
    G -->|"no (pending/processing)"| X["409 — use cancel<br/>(unchanged route)"]
    D --> P["row gone on next poll"]
```

## Semantics: cancel vs delete

| Verb | Route | Effect |
|---|---|---|
| Cancel | `DELETE …/operations/{id}` | `pending → cancelled` (soft, unchanged) |
| Delete record | `DELETE …/operations/{id}/delete` | hard-delete of a terminal row |

A review bot suggested the control plane was calling the cancel verb — that's a mix-up of the BFF namespace with the dataplane path: the BFF's `DELETE /api/banks/{bankId}/operations/{id}` handler calls the generated SDK, which requests `…/operations/{id}/delete`; cancel lives on a separate BFF route (`/api/operations/{bankId}`).

## Review fixes folded in

- Python client surface properly regenerated — the previous hand-patch had a doubled `@validate_call` on `delete_operation` and had dropped the decorator from `retry_operation` (silently losing runtime arg validation there).
- Route renamed `…/operations/{operation_id}/record` → `…/operations/{operation_id}/delete` per maintainer review; OpenAPI spec and Go/Python/TypeScript clients regenerated (no client symbol names changed — they derive from the explicit `operation_id="delete_operation"`, not the path).
- `"Failed to delete operation"` registered in `failureErrorKeys` so error payloads localize like every sibling operation endpoint (keys already existed in all 10 locales).

## How to test

```bash
cd hindsight-api-slim
uv run pytest tests/test_operation_status.py tests/test_operation_progress.py -v -n0

cd ../hindsight-control-plane
npm test && npm run i18n:check
```

Manual E2E: retain a doc so an operation ends `failed` → Operations → detail → **View Document** (when `document_id` set) → **Delete** → row disappears on next poll; cancel of a `pending` op still works via the old route.

## Test evidence

Run post-rebase on upstream `main` (`52b893b93`), squashed branch:

| Check | Result |
|---|---|
| `pytest tests/test_operation_status.py tests/test_operation_progress.py -n0` | **20 passed** |
| control-plane `npm test` | **76 passed** (9 files) |
| `npm run i18n:check` | clean |
| `ruff check` / `format --check` / `ty check` on touched modules | clean |
| OpenAPI spec + Python/TS/Go clients | regenerated via project scripts (openapi-generator 7.10.0 / @hey-api), no drift vs spec |

## Out of scope

- Bulk "clear all failed" — the issue reporter confirmed per-row delete is enough.
- Deleting stuck `processing` operations — tracked upstream in #2644 / #2675; the status guard deliberately excludes `processing`.
- `delete_document` does not cascade to operations (separate issue).
- **Known edge (documented in code, guard deliberately omitted):** deleting a terminal *child* of a still-running batch removes it from the parent's roll-up, so the parent can finalize `completed` despite a deleted failed child. Parent linkage lives in JSON `result_metadata` (no FK). If maintainers prefer a hard guard, the clean shape is a `NOT EXISTS (non-terminal parent)` predicate folded into the same single-statement DELETE — happy to add it in this PR on request.


nicoloboschi (changes requested):
let's call the endpoint 
/delete


    @app.delete(
        "/v1/default/banks/{bank_id}/operations/{operation_id}/delete",

nicoloboschi (approved):
Approving. Clean, well-scoped fix for #2677 that follows existing conventions closely.

Highlights:
- `delete_operation` uses a single status-guarded `DELETE … RETURNING`, closing the TOCTOU window that the two-statement `cancel_operation` still has. Bank scoping and 404-vs-409 distinction are correct.
- HTTP handler, `@audited` decorator, and `BankWriteOperation.DELETE_OPERATION` all mirror the cancel/retry patterns.
- Cross-bank isolation is tested; generated clients + both openapi.json copies + all 10 locales are consistent.

Non-blocking notes for follow-up if desired: the documented parent batch roll-up edge case (deleting a terminal child of a running batch), an optional confirm dialog to match other CP destructive deletes, and adding `cancelled`/`completed` to the success test.

PR #2779, opened 2026-07-17
fix(retain): a zero retry budget must still perform the initial fact-extraction request

### LINKED ISSUE #2731: fix(retain): a zero retry budget must still perform the initial fact-extraction request

chethanuk:
## Problem

A retry budget of zero must mean **one initial request and no retries**. Today `retain_llm_max_retries=0` performs **zero** fact-extraction requests and raises `Fact extraction failed after 0 attempts` — blocking trusted-gateway deployments where the operator deliberately sets the budget to 0 because the upstream gateway owns transport retries.

Fixes vectorize-io/hindsight#2731

## Root cause

Two nested layers read the same knob with different meanings:

- `OpenAICompatibleLLM.call(max_retries=N)` correctly loops `range(N + 1)` — N retries *after* the initial request (same in the Anthropic, Gemini, LiteLLM and Claude Code providers).
- `_extract_facts_from_chunk` used the same value as an **outer attempt count**: `for attempt in range(llm_max_retries)` → `range(0)` → zero requests, and `range(1)` → no retry at all for budget 1.

```mermaid
flowchart LR
    A["retain_llm_max_retries = N"] --> B["outer loop (content-quality re-prompts)<br/>OLD: range(N) → N=0 ⇒ 0 attempts ❌<br/>NEW: range(N + 1) → N retries after the initial request ✅"]
    B --> C["llm_config.call(max_retries=N)<br/>provider transport retries: range(N + 1)<br/>(unchanged — provider owns transport)"]
    C --> D["LLM"]
```

## Fix

Make the outer loop follow the same `N + 1` convention every provider already uses, while forwarding the raw budget to the provider unchanged:

```python
outer_attempts = llm_max_retries + 1        # N=0 → 1 attempt; N=1 → initial + 1 retry
llm_config.call(..., max_retries=llm_max_retries)  # raw transport budget, provider does N+1
```

One knob, one meaning: **N retries after the initial request**, at both layers. Budget 0 → exactly one extraction request; budget 1 → one retry (2 outer attempts), which is what the flag has always claimed to do.

Every attempt-count reference is also corrected — retry-decision conditions, warning logs, and **both** error messages — so a malformed response with budget 0 reports `after 1 attempts`, never `after 0 attempts`.

## How this differs from #2741

#2741 floors the outer count at `max(1, N)`, which fixes only the N=0 case and still performs no retry at N=1. This PR aligns the outer loop with the provider convention (`N + 1`) as requested in review, and additionally fixes the error path that actually fires on a malformed response with budget 0: under #2741 the primary raise still reports "after 0 attempts" (only the rarely-reached final fallback message was updated). Coverage is also broader: table-driven tests across N = 0/1/3/unset — including the global-fallback path and the malformed-response contract.

## How to test

```bash
cd hindsight-api-slim
uv run pytest tests/test_fact_extraction_retry.py -v
uv run ruff check hindsight_api/engine/retain/fact_extraction.py tests/test_fact_extraction_retry.py
uv run ruff format --check hindsight_api/engine/retain/fact_extraction.py tests/test_fact_extraction_retry.py
uv run ty check hindsight_api/engine/retain/fact_extraction.py
```

## Test evidence

| Check | Result |
|---|---|
| `pytest tests/test_fact_extraction_retry.py -n0` | **20 passed** |
| `ruff check` / `ruff format --check` on both touched files | clean |
| `ty check hindsight_api/engine/retain/fact_extraction.py` | clean |

## Test plan

- [x] Budget 0 → exactly one extraction request, `max_retries=0` forwarded to provider
- [x] Budget 0 + malformed response → one attempt then raise (never "after 0 attempts")
- [x] Budget 1 → initial request + 1 retry (2 outer attempts)
- [x] Budget 3 / unset (falls back to global `llm_max_retries`) → `N + 1` outer attempts
- [x] Pre-existing non-dict / retain-override / none-event_date tests still green


koriyoshi2041:
I checked this against the branch head `4b7c3690` because it overlaps #2741.

Local verification on macOS / Python 3.11:

- `uv run --package hindsight-api-slim pytest hindsight-api-slim/tests/test_fact_extraction_retry.py -q` — 20 passed
- focused Ruff check and format check — passed

The zero-budget path makes one outer extraction call, forwards `max_retries=0` unchanged to the provider, and malformed output still fails on the content error rather than the old “after 0 attempts” fallback. I did not find a behavioral regression in the focused path.

nicoloboschi (changes requested):
uhm this fix is wrong. with llm_max_retries=1, we should do 1 retry still (so 2 total outer attempts). so I think we should just do llm_max_retries+1 attempts - please also check the impl in llm_config.call follow the same pattern

nicoloboschi (approved):
great, lgtm

PR #2786, opened 2026-07-17
fix: count crash-recovery attempts toward max-retry budget

### LINKED ISSUE #2675: Crash-interrupted operations are re-claimed indefinitely — claim attempts never count toward max_attempts (poison-pill regrind loop)

handnewb:
## Summary

When processing an `async_operations` row crashes the worker (OOM, etc.), no failure bookkeeping runs — `retry_count` is only incremented by in-process failure handling. On restart, `recover_own_tasks` resets `processing` rows back to `pending` with `retry_count` untouched, and the row is re-claimed as if brand new.

An operation that can never complete therefore loops forever: claim → grind → crash → recover → re-claim…

## Production Evidence

Reported in #2675: a batch retain that fans out to ~6,624 LLM calls per attempt was re-ground **54 times over 22 hours**, consuming ~394K LLM calls (863M input + 211M output tokens) with zero useful yield. `retry_count` was still 0 after all 54 cycles.

## Fix

In `recover_own_tasks`:
- Tasks under `max_recovery_attempts` (default 5): increment `retry_count` and reset to `pending`
- Tasks at or above the limit: move to `status='failed'` with error `"exceeded max recovery attempts (likely crash-interrupted)"`
- Logs a warning count when tasks are quarantined

With a cap of 5, the incident above would have been bounded to ≤5 grinds per operation and zero unattended overnight churn.

**31 lines added, 5 removed.**

Closes #2675

### DIFF
--- a/hindsight-api-slim/hindsight_api/worker/main.py
+++ b/hindsight-api-slim/hindsight_api/worker/main.py
@@ -269,6 +269,7 @@ def main():
             max_slots=config.worker_max_slots,
             slot_reservations=config.worker_slot_reservations,
             consolidation_bank_priority=config.worker_consolidation_bank_priority or None,
+            max_retries=config.worker_max_retries,
         )
 
         # Create the HTTP app for metrics/health
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -161,6 +161,7 @@ class WorkerPoller:
         max_slots: int = 10,
         slot_reservations: dict[str, int] | None = None,
         consolidation_bank_priority: dict[str, int] | None = None,
+        max_retries: int = 3,
     ):
         """
         Initialize the worker poller.
@@ -183,6 +184,8 @@ class WorkerPoller:
                 Patterns support ``*`` as wildcard. A bare ``*`` key is the catch-all default.
                 When set, consolidation tasks are claimed in priority tiers rather than
                 pure created_at order. None or empty dict preserves current behavior.
+            max_retries: Maximum retry attempts before a task is marked failed.
+                Must be >= 0. Default 3 (matches DEFAULT_WORKER_MAX_RETRIES).
         """
         self._backend = backend
         self._worker_id = worker_id
@@ -204,6 +207,7 @@ class WorkerPoller:
         self._consolidation_bank_priority: dict[str, int] | None = (
             consolidation_bank_priority if consolidation_bank_priority else None
         )
+        self._max_retries = max(0, max_retries)  # Never negative
         # Cache of which optional PG routines are installed on the server
         # (probed once, memoised for the life of the poller).
         from ..engine.db.optional_routines import OptionalRoutines
@@ -797,14 +801,18 @@ class WorkerPoller:
 
         This handles the case where a worker crashes while processing tasks.
         On startup, we reset any tasks stuck in 'processing' for this worker_id
-        back to 'pending' so they can be picked up again.
+        back to 'pending' so they can be picked up again — provided their
+        ``retry_count`` has not reached ``max_retries``. Tasks at/over the
+        limit are moved to 'failed' with an explanatory error message,
+        breaking the infinite loop where a task that kills the worker
+        (OOM, infinite loop) is re-claimed forever.
 
         Also recovers batch API operations that were in-flight.
 
         If tenant_extension is configured, recovers across all tenant schemas.
 
         Returns:
-            Number of tasks recovered
+            Number of tasks recovered (reset to pending, not including failed)
         """
         schemas = await self._get_schemas()
         total_count = 0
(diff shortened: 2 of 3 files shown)


nicoloboschi (changes requested):
nice catch, but we must reuse HINDSIGHT_API_WORKER_MAX_RETRIES instead of adding a new config field

handnewb:
Superseded by a new version of this fix.

The original PR (#2786) introduced a **new** config field (`max_recovery_attempts`) and conflicted with the upstream refactoring in #2819. After upstream merged #2819, the crash-recovery infinite-loop bug (#2675) still remained — #2819 simplified `recover_own_tasks` and removed the retry-counting logic as a side effect.

A fresh PR is being opened that:
- Reuses the **existing** `HINDSIGHT_API_WORKER_MAX_RETRIES` config (per review feedback)
- Is adapted to the post-#2819 codebase (zero conflicts)
- Includes tests for retry_count increment, exceeded→failed, and NULL handling

Closing this in favor of the new version.

PR #2787, opened 2026-07-17
fix: re-assert entities before Phase-2 unit_entities insert to close FK race

### LINKED ISSUE #2662: FK race: prune_orphan_entities deletes entities between retain Phase-1 resolve and Phase-2 unit_entities insert — silent memory loss on re-ingest (v0.8.4)

handnewb:
## Summary

`prune_orphan_entities` can delete an entity row between Phase-1 entity resolution (separate connection, already committed) and Phase-2 `unit_entities` insert (new transaction). When the pruner strikes in that window, the FK on `unit_entities.entity_id` fails with `ForeignKeyViolationError`, and the entire `batch_retain` is dropped as a non-retryable failure — **silent memory loss**.

Reproduced deterministically 15/15 on the real orchestrator path (#2662).

## Root Cause

- Phase 1: `resolve_entities` creates entities on `resolve_conn`, commits, closes
- Pruner: `prune_orphan_entities` runs, deletes entity E (no `unit_entities` row yet)
- Phase 2: `bulk_insert_unit_entities` inserts `(unit_id, E.id)` → FK violation

No row lock or snapshot spans the two phases, so nothing prevents the race.

## Fix

Re-assert entity rows with `INSERT INTO entities ... ON CONFLICT (id) DO NOTHING` on the Phase-2 connection/transaction immediately before `link_units_to_entities_batch`. 

- If entity still exists (normal case): `ON CONFLICT DO NOTHING` → no-op
- If entity was pruned: row is re-created with entity UUID as fallback `canonical_name` so the FK passes; real name recovered on next retain cycle

**20 lines added**, zero existing behavior changed when the race doesn't fire.

## Trade-off

Pruned entities are resurrected with UUID-as-name instead of their original `canonical_name`. The next retain that encounters the real name will create a new (correct) entity row via normal resolution; the UUID-named row becomes orphaned and is re-pruned. This is a safe, bounded trade-off vs. silent data loss.

Closes #2662

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -6784,7 +6784,7 @@ class MemoryEngine(MemoryEngineInterface):
                     # the graph-maintenance run this edit submits.
                     if new_entities is not None:
                         entity_date = new_occ_start or live["mentioned_at"]
-                        _resolved_ids, _e2u, unit_to_entity_ids = await resolve_entities_only(
+                        _resolved_ids, _e2u, unit_to_entity_ids, _names = await resolve_entities_only(
                             self.entity_resolver,
                             conn,
                             bank_id,
--- a/hindsight-api-slim/hindsight_api/engine/retain/entity_processing.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/entity_processing.py
@@ -76,10 +76,10 @@ async def resolve_entities(
         entity_labels: Optional entity label taxonomy
 
     Returns:
-        Tuple of (resolved_entity_ids, entity_to_unit, unit_to_entity_ids).
+        Tuple of (resolved_entity_ids, entity_to_unit, unit_to_entity_ids, entity_names).
     """
     if not unit_ids or not facts:
-        return [], [], {}
+        return [], [], {}, []
 
     if len(unit_ids) != len(facts):
         raise ValueError(f"Mismatch between unit_ids ({len(unit_ids)}) and facts ({len(facts)})")
--- a/hindsight-api-slim/hindsight_api/engine/retain/entity_processing.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/entity_processing.py
@@ -58,7 +58,7 @@ async def resolve_entities(
     log_buffer: list[str] = None,
     user_entities_per_content: dict[int, list[dict]] = None,
     entity_labels: list | None = None,
-) -> tuple[list[str], list[tuple], dict[str, list[str]]]:
+) -> tuple[list[str], list[tuple], dict[str, list[str]], list[str]]:
     """
     Phase 1: Resolve entity names to canonical IDs (read-heavy).
 
--- a/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
@@ -325,10 +325,11 @@ async def resolve_entities_only(
         entity_labels: Optional entity label taxonomy
 
     Returns:
-        Tuple of (resolved_entity_ids, entity_to_unit, unit_to_entity_ids) where:
+        Tuple of (resolved_entity_ids, entity_to_unit, unit_to_entity_ids, entity_names) where:
         - resolved_entity_ids: list of entity IDs in same order as flattened entities
         - entity_to_unit: maps flat index to (unit_id, local_index, fact_date)
         - unit_to_entity_ids: maps unit_id to list of resolved entity IDs
+        - entity_names: list of entity name strings in same flattened order
     """
     all_entities_flat, _all_entities, entity_to_unit = _prepare_entities_for_resolution(
         unit_ids, sentences, fact_dates, llm_entities, log_buffer
@@ -336,7 +337,7 @@ async def resolve_entities_only(
 
     if not all_entities_flat:
         _log(log_buffer, "  [6.2] Entity resolution (batched): 0 entities", level="debug")
-        return [], [], {}
+        return [], [], {}, []
 
     step_start = time.time()
     resolved_entity_ids = await entity_resolver.resolve_entities_batch(
@@ -366,7 +367,11 @@ async def resolve_entities_only(
         level="debug",
     )
 
-    return resolved_entity_ids, entity_to_unit, unit_to_entity_ids
+    # Extract entity names in the same flattened order for callers that
+    # need to re-assert pruned entities (FK race recovery).
+    all_entity_names = [e["text"] for e in all_entities_flat]
+
+    return resolved_entity_ids, entity_to_unit, unit_to_entity_ids, all_entity_names
 
 
 async def create_temporal_links_batch_per_fact(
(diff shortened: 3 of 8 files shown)


ebarkhordar (commented):
Two notes, one on scope and one a question about the fallback name. Neither is blocking.

Scope: alongside the orchestrator FK-race fix, this PR also rewrites `recover_own_tasks` in `worker/poller.py` (the crash-recovery retry-budget change). That looks like the same change as your #2786, which additionally makes the limit configurable via `HINDSIGHT_API_WORKER_MAX_RECOVERY_ATTEMPTS`; the hardcoded `max_recovery_attempts = 5` here is a subset of it. With both open they will collide on that function. It may be cleaner to keep this PR to the `orchestrator.py` FK fix and let #2786 own the poller change, so each lands on its own merits.

Question on the fallback name: the re-assert inserts the resurrected row with `canonical_name = id::text`, and the comment says the real name is recovered on the next retain cycle. I don't think that holds for the link created during the race. Resolution fetches candidates by name similarity (`LOWER(e.canonical_name) % LOWER(q.query_text)`, `entity_resolver.py` around line 445) and reuses an id only on a name match (exact at line 661, SequenceMatcher score otherwise). A UUID string is not trigram-similar to the real name, so on the next cycle the name resolves to a different id and the `unit_entities` row from the race stays pointed at the UUID-named entity. That reads as a permanent mislabel rather than a transient one, and listing that unit's entities would return the UUID.

Phase 1 has already resolved the name at that point. Would it be feasible to carry it into the re-assert (`canonical_name = <resolved name>`) instead of the UUID, so the resurrected row keeps the correct name? That keeps the FK fix without the naming side effect.

handnewb:
@ebarkhordar thank you for the thorough review! 🙏 Both points were spot-on and both are now addressed:

**1. Scope split** ✅ — Removed the poller.py crash-recovery change. That fix now lives in its own PR (#2834) with proper config reuse (HINDSIGHT_API_WORKER_MAX_RETRIES) and tests.

**2. UUID fallback naming** ✅ — You were completely right about the permanent mislabel. Phase-1 resolution already has the correct entity names, so the re-assert now carries them through via `resolved_entity_names` (threaded through `EntityResolutionResult` → `_insert_facts_and_links`). The `ON CONFLICT DO NOTHING` re-assert uses the actual resolved name instead of `id::text`, so resurrected entities keep their correct `canonical_name` and trigram matching works on the next retain cycle.

Changes: +32 / -139 lines across 7 files (most of the - is from removing the crash-recovery change).

nicoloboschi:
Superseded by #2859, which consolidates the best of this PR and the sibling into one fix: it carries the **stored canonical name** across the phase boundary (as here) *and* locks surviving parents `FOR KEY SHARE` so the pruner blocks (closing the residual lock-free window), folds lock+reassert into a single statement matching the existing `bulk_insert_links` convention, and also covers the curation edit path. Closing in favor of #2859 — thanks for the diagnosis and repro, they carried straight over. Reopen if you'd like to keep iterating here instead.

handnewb:
Thanks for the consolidation @nicoloboschi — the `FOR KEY SHARE` approach and carrying the stored `canonical_name` are clear improvements over my reassert-only approach. Glad the diagnosis and repro were useful. Closing in favor of #2859 makes total sense.

PR #2792, opened 2026-07-17
Make entities required with an empty-list default so literal models stop dropping them

### LINKED ISSUE #2749: [BUG] Auto-extracted entities are never persisted (v0.8.4): prompt teaches string arrays while Entity requires objects, and the field is optional

dimonnld:
Fixes #2749.

## Problem

The four extraction models (`ExtractedFact`, `ExtractedFactVerbose`, `ExtractedFactNoCausal`, `VerbatimExtractedFact`) declared `entities` as an optional field that was **absent from `json_schema_extra["required"]`**. Under a strict JSON schema a literal model resolves the conflict — the field is optional and typed as objects, while the few-shot examples taught bare strings — by **omitting it entirely**: zero entities, no error. It bites precisely the most schema-conformant models and reproduces with strict-schema both on and off.

The retain parser is lenient (it coerces string entities into `Entity(text=...)`), so the string-array shape itself survives — the load-bearing defect is the omitted optional field, not the string shape.

## Fix

- **`entities: list[Entity] = Field(default_factory=list, ...)`** on all four models, and `"entities"` added to each model's `required` array. An empty list stays valid, so lenient parsing never hard-fails, but the field is no longer silently omitted by literal models.
- **Prompts teach the object form.** The concise few-shot examples now use `[{"text": ...}]` objects, and an explicit *"array of objects with a `text` field, never bare strings"* instruction is added to the concise/verbatim (base) prompt and the verbose prompt — so non-strict / no-schema paths stop emitting the ambiguous shape.

## Tests

`test_fact_extraction_entities_schema.py` — deterministic, no LLM:

- each of the four models' generated JSON schema lists `entities` as **required**, typed as an **array of objects with a `text` field** (the CI regression that would have caught this);
- omitting `entities` still parses, defaulting to `[]` (lenient path intact);
- every extraction prompt (concise / verbose / verbatim / custom) teaches the object form and shows no bare-string entity examples.

Verified locally: the schema/prompt suite is green, all prompt modes assemble, and the existing `test_fact_extraction_fact_type_prompt.py` still passes. Note the base prompt is `.format()`-ed twice (module assembly, then per-request), so the object examples in the few-shot are escaped accordingly and the standalone format instruction is brace-free.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -180,7 +180,7 @@ class ExtractedFact(BaseModel):
 
     model_config = ConfigDict(
         json_schema_mode="validation",
-        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type"]},
+        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type", "entities"]},
     )
 
     what: str = Field(description="Core fact - concise but complete (1-2 sentences)")
@@ -195,7 +195,7 @@ class ExtractedFact(BaseModel):
     fact_type: Literal["world", "assistant"] = Field(
         description="'world' = objective/external facts, including user preferences, rules, corrections, and constraints even when stated during a conversation. 'assistant' = actions, experiences, or observations the assistant/agent actually performed."
     )
-    entities: list[Entity] | None = Field(default=None, description="People, places, concepts")
+    entities: list[Entity] = Field(default_factory=list, description="People, places, concepts")
     causal_relations: list[FactCausalRelation] | None = Field(
         default=None, description="Links to previous facts (target_index < this fact's index)"
     )
@@ -286,7 +286,7 @@ class ExtractedFactVerbose(BaseModel):
 
     model_config = ConfigDict(
         json_schema_mode="validation",
-        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type"]},
+        json_schema_extra={"required": ["what", "when", "where", "who", "why", "fact_type", "entities"]},
     )
 
     what: str = Field(
@@ -348,8 +348,8 @@ class ExtractedFactVerbose(BaseModel):
         description="'world' = objective/external facts about the user, other people, events, general knowledge, preferences, rules, corrections, or constraints. 'assistant' = actions, experiences, or observations the assistant/agent actually performed (e.g., 'I changed X', 'I discovered Y')."
     )
 
-    entities: list[Entity] | None = Field(
-        default=None,
+    entities: list[Entity] = Field(
+        default_factory=list,
         description="Named entities, objects, AND abstract concepts from the fact. Include: people names, organizations, places, significant objects (e.g., 'coffee maker', 'car'), AND abstract concepts/themes (e.g., 'friendship', 'career growth', 'loss', 'celebration'). Extract anything that could help link related facts together.",
     )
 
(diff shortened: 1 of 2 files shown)


dimonnld:
Closing as superseded. #2749 was already fixed in main by #2830, which took the `list[str]` (flat strings) approach — main now has `entities: list[str] = Field(default_factory=list)`. This PR used `list[Entity]` (required + empty-list default), which conflicts with that design choice, and its regression test duplicates the one #2830 added (`test_fact_extraction_entities_schema.py`). Nothing left to land here.

PR #2794, opened 2026-07-17
fix(retain): reject degenerate fact text before storage (#2520)

### LINKED ISSUE #2520: Degenerate fact text ("...", punctuation-only) passes both extraction parsing and consolidation into recallable memory — no content guard
## Observation

Fact text with zero information content — the literal string `...` (also `-`, `…`, whitespace-only variants) — can be stored, embedded, BM25-indexed, and served by recall. Observed on a production v0.7.2 deployment: 14 stored rows whose entire text is `...`, all `fact_type=observation`, repeatedly surfacing in recall results (their embeddings/rerank scores match a wide range of queries, so a handful of degenerate rows contaminate many recalls).

Two stages lack a guard:

1. **Extraction parse paths** — the only content gate on the parsed `what` is falsy / literal `"N/A"` rejection; `"..."` passes. Affected: `_extract_facts_from_chunk` (~L1380-1398) and `extract_facts_from_contents_batch_api` (~L2155-2166) in `hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py`; `sanitize_text` strips control chars only. Checked against main @ 6a479dd (2026-07-02): unchanged; same in v0.7.2 and v0.8.2.
2. **Consolidation observation output** — `_create_observation_directly` (consolidator.py) inserts whatever observation text the consolidation LLM emitted, with no degenerate-content rejection. In our production case the degenerate rows are observation-typed, i.e. they reached the layer that default Claude Code plugin recall (`recallTypes: ["observation"]`) reads exclusively — so the noise lands in every consumer.

## Repro pointer

Retain low-signal conversational content where a small extraction model (e.g. gpt-oss-20b under `retain_extraction_mode: concise`) emits `"what": "..."`, and/or let consolidation summarize sparse source facts. Degenerate rows land in `memory_units` and surface in recall.

## Suggested guard

After the `what` fallback in both extraction parse paths, and on consolidation observation text before insert:

```python
if not text or not re.search(r"\w", str(text)):
    logger.warning("dropping degenerate fact text")
    continue  # count in RetainExtractionErrors / consolidation metrics
```

Verbatim mode is unaffected (`what` intentionally absent there). Optionally also validate final `combined_text`.

Versions checked: main @ 6a479dd (2026-07-02), v0.8.2, v0.7.2.

handnewb:
## Summary

Facts with zero information content — empty strings, punctuation-only, and common LLM hallucination patterns (`...`, `-`, `--`) — were being stored, embedded, BM25-indexed, and surfaced in recall results, polluting the memory bank.

## Changes

- Added `_is_degenerate_text()` guard in `ProcessedFact` that rejects:
  - Empty/whitespace-only strings
  - Single/repeated punctuation marks (..., —, -, --, ., •, etc.)
  - Strings composed entirely of non-alphanumeric characters
  - Very short (<=2 char) punctuation-only strings
- `from_extracted_fact()` now returns `None` for degenerate facts with a warning log
- Updated callers in `orchestrator.py` and `importer.py` to filter out `None` results

## Verification

- Degenerate text is caught before storage pipeline (embedding generation, DB insert, BM25 indexing)
- Existing valid facts pass through unchanged
- Callers gracefully skip rejected facts

Closes #2520

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -549,7 +549,11 @@ async def _extract_and_embed(
     embeddings = await embedding_processing.generate_embeddings_batch(embeddings_model, augmented_texts)
     log_buffer.append(f"  Generate embeddings: {len(embeddings)} embeddings in {time.time() - step_start:.3f}s")
 
-    processed_facts = [ProcessedFact.from_extracted_fact(ef, emb) for ef, emb in zip(extracted_facts, embeddings)]
+    processed_facts = [
+        pf
+        for ef, emb in zip(extracted_facts, embeddings)
+        if (pf := ProcessedFact.from_extracted_fact(ef, emb)) is not None
+    ]
 
     return extracted_facts, processed_facts, chunks, usage
 
--- a/hindsight-api-slim/hindsight_api/engine/retain/types.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/types.py
@@ -10,6 +10,10 @@ from datetime import datetime
 from typing import Literal, TypedDict
 from uuid import UUID
 
+import logging
+
+logger = logging.getLogger(__name__)
+
 
 class RetainContentDict(TypedDict, total=False):
     """Type definition for content items in retain_batch_async.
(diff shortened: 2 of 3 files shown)


ebarkhordar (commented):
Good catch on the degenerate-fact pollution. One thing to flag before this lands: making `from_extracted_fact` return `None` shortens `processed_facts` relative to `extracted_facts`, and two spots in `orchestrator.py` still pair those two lists positionally.

`_extract_and_embed` returns the full `extracted_facts` next to the now-filtered `processed_facts`, and the consumers that map chunk ids back onto facts zip them together:

- streaming/full retain: `orchestrator.py:1591` `for fact, processed_fact in zip(batch_extracted, batch_processed)`
- delta re-retain: `orchestrator.py:2185` `for ef, pf in zip(extracted_facts, processed_facts)`

`ProcessedFact` has no `chunk_index`, so these loops read `chunk_index` off the extracted fact to look up the chunk id. Once a degenerate fact is dropped, the lists are off by one from that point, so every surviving fact after the dropped one gets the previous fact's chunk id, and `zip` silently drops the tail. The result mapping already guards this exact subset hazard (the comment at `orchestrator.py:499-503`, "unit_ids has 1:1 alignment with processed_facts ... any upstream drop would otherwise cause an IndexError, see #1037"), but these two zips still assume 1:1.

Reproduced against this PR's head (9c0fc5b) using your own `ProcessedFact.from_extracted_fact` plus the exact loop from :1591/:2185, three facts with a `"..."` in the middle:

```
len(extracted_facts) = 3
len(processed_facts)  = 2   (degenerate dropped)
'Alice joined Acme in 2021' -> CID_for_chunk10
'Bob leads the ML team'     -> CID_for_chunk11   (expected CID_for_chunk12)
```

I only exercised those two functions directly, not a full DB-backed retain, so I have not watched a wrong `chunk_id` get persisted. But the trigger is exactly the case this PR targets: the LLM emitting a `"..."`-style fact ahead of a real one in the same batch.

One option that keeps it simple is to filter `extracted_facts` in lockstep so the pair stays 1:1, or carry `chunk_index` onto `ProcessedFact` and iterate over `processed_facts` alone. A small regression test with a degenerate fact between two real ones would pin whichever shape you pick.

nicoloboschi (approved):

PR #2795, opened 2026-07-17
fix(consolidation): release large objects between retry attempts to prevent RSS leak (#2273)

### LINKED ISSUE #2273: consolidation failure path leaks LLM context — RSS grows unboundedly on retry storms
## Summary

When the consolidation LLM call fails repeatedly, the `consolidator._consolidate_batch_with_llm` function in
`hindsight_api/engine/consolidation/consolidator.py` keeps the LLM call kwargs, response object, and prompt
in scope across the retry loop. With `consolidation_max_attempts=5` (or `max_retries=10-11` via the LLM
client), a single failing batch accumulates hundreds of MB of unreleased objects. In our production
deployment this caused the worker RSS to grow from a 1.2 GB steady state to 14 GB over ~36 hours,
triggering 200,000+ OOM kills (counted from `systemctl show hindsight -p NRestarts`).

## Environment

- `hindsight-api==0.7.1` (still reproducible on `main` as of 2026-06-17, fix not yet shipped)
- Python 3.12
- LLM provider: MiniMax / OpenAI-compatible
- Deployment: long-lived systemd unit (LXC container)
- Affects: any deployment where consolidation encounters transient LLM errors (rate limits, network
  blips, model overloaded) and is allowed to retry.

## Reproduction

1. Point the consolidation LLM at an unreachable URL via
   `HINDSIGHT_API_CONSOLIDATION_LLM_BASE_URL=http://127.0.0.1:9999`.
2. Restart Hindsight and trigger `POST /v1/default/banks/<bank>/consolidate`.
3. Watch `/proc/<pid>/status` `VmRSS` over 2-10 minutes.

**Without the patch**: RSS grows monotonically with each retry attempt (call kwargs + response + prompt
copies held in the except-block scope).

**With the patch below**: RSS stays flat at the 1.2 GB steady state across 7-11 retry attempts
(verified over a 120 s observation window with 12 consecutive `VmRSS` samples within ±5 MB).

## Root cause

`hindsight_api/engine/consolidation/consolidator.py`, `_consolidate_batch_with_llm`, ~line 1380-1395:

handnewb:
## Summary

When a consolidation LLM call fails repeatedly, the `call_kwargs` dict (containing the full system_prompt and user_content strings — potentially hundreds of KB of text) and the `response` object were kept in scope across the retry loop. With `consolidation_max_attempts=5` and retry storms, repeated failures could accumulate hundreds of MB of unreclaimable memory, eventually causing OOM.

## Changes

- Initialize `call_kwargs` and `response` as `None` before the retry loop
- Add `finally` block with explicit `del` to release references between retry attempts
- Use `try/except NameError` guards for safety (the variables may not be assigned if the exception occurs before their assignment line, though in practice this never happens)

## Verification

- Successful consolidation path is unchanged (objects are naturally cleaned up on return)
- Failed retry path now releases references before the next attempt
- No functional change to consolidation logic

Closes #2273

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -2289,6 +2289,8 @@ async def _consolidate_batch_with_llm(
     else:
         ids_label = f"{', '.join(memory_ids[:3])}, ... +{len(memory_ids) - 3} more"
     batch_label = f"{len(memory_ids)} memories [{ids_label}]"
+    call_kwargs = None  # Initialize for finally block cleanup
+    response = None
     for attempt in range(1, max_attempts + 1):
         try:
             call_kwargs: dict[str, Any] = {
@@ -2309,7 +2311,7 @@ async def _consolidate_batch_with_llm(
                 call_kwargs["max_retries"] = inner_max_retries
             if cached_prefix_name is not None:
                 call_kwargs["cached_prefix"] = cached_prefix_name
-            response: _ConsolidationBatchResponse = await llm_config.call(**call_kwargs)
+            response = await llm_config.call(**call_kwargs)
             # Defensive truncation: some LLM providers may not enforce JSON schema max_length
             creates = response.creates
             if remaining_observation_slots is not None and remaining_observation_slots >= 0:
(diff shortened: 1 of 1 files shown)


nicoloboschi:
this is non sense

PR #2797, opened 2026-07-17
perf(postgresql): replace correlated NOT EXISTS with set-based EXCEPT in prune_stale_cooccurrences (#2660)

### LINKED ISSUE #2660: Pass-3 correlated anti-join is excessively slow; equivalent set-based DELETE is substantially faster
## Version, scope, and status

This is an upstream performance observation and proposed direction for Hindsight 0.8.4, not a claim that a production fix has been deployed or fully validated. All SQL timing/equivalence work described below was performed against isolated clone/synthetic environments; the standalone DELETE measurements were rollback-only. No production Hindsight API, database, source, configuration, or cron state was changed.

## Problem

Pass-3 prunes graph cooccurrences through a correlated `NOT EXISTS` witness check over a `unit_entities` self-join. On an anonymized clone with `entities` = 50,839, `unit_entities` = 1,047,047, and `entity_cooccurrences` = 663,967, the original DELETE (O) exceeded a 300-second statement timeout (observed runs: 300.152s, 300.166s, and 300.157s).

The plan retained the adverse shape before and after `VACUUM (ANALYZE)`: a sequential scan of `entity_cooccurrences`, a nested-loop anti-join, and a correlated `unit_entities` self hash join with witness index scans. The necessary indexes already existed; this does not appear to be an index-missing problem. The likely bottleneck is repeated evaluation of the Pass-3 correlated anti-join/witness test.

## Proposed direction and evidence

A set-based rewrite (R) materializes the stale cooccurrence target set and deletes it set-wise rather than re-evaluating the correlated witness check per candidate.

On the large snapshot, O and R selected the same target set:

- O = R = 5,746 rows
- `O EXCEPT R = 0`; `R EXCEPT O = 0`
- ordered digest matched:
`7399ce8bf0c10a8124ade98f8b4df7c8`

The R DELETE alone completed in 3.368s for 5,746 rows in a rollback-only transaction. **This is the standalone DELETE timing, not the runtime of the complete graph-maintenance worker.**

For context, complete worker runs with the canonical patch completed in approximately 14.148s on the synthetic canonical fixture, and in 23.840s on a previously restored logical archive containing 6,575,312 rows across 19 tables. The latter worker reported 2,685 stale-cooccurrence prunes; that count is from a different snapshot/stage and should not be equated with the 5,746-row standalone target set.

The canonical artifact provenance is:

- pristine 0.8.4 full-file SHA-256: `b6e73c042d0f5b29d8d44bcfe32f1b28dbcaa1d9371410d481af53b2a25a06f5`
- canonical patched full-file SHA-256: `8faf437a63dd6dad64ee6f2ab307da3824b45f9a1efa39d93bd08253e9c5107c`
- raw R SQL SHA-256: `db8cc37f4a06cc043aeac16432254fb0067e13200fc0be031f5a59b1a8068412`

## Validation boundaries

The synthetic canonical actual-worker run completed with pre-run O/R = 2/2 and bidirectional `EXCEPT` = 0/0; after execution, two expected rows were removed with bidirectional post-oracle `EXCEPT` = 0/0. It also retained the valid pair and preserved the checked memory-unit/document and control-bank invariants.

However, this is not yet a complete production-readiness validation. The following are not complete: a unified real-scale same-run post-oracle, an exact rollback failpoint, full concurrency retain/recall coverage, edge/idempotency coverage, and physical A/B validation. The real-scale canonical path did not complete its later restore/pre-oracle stage, so no unified real-scale post-oracle result is claimed.

## Request
Could you review whether Pass-3 should replace or avoid this correlated anti-join with a tested set-based formulation, while preserving duplicate/cooccurrence and transaction semantics? It would also be useful to add regression/performance coverage for this plan shape. I can provide the sanitized SQL shape and validation artifacts if maintainers would like them.

handnewb:
## Summary

The original Pass-3 `prune_stale_cooccurrences` uses a correlated `NOT EXISTS` self-join on `unit_entities` that evaluates per-row. On large graphs:
- 50K+ entities
- 1M+ `unit_entities` rows
- 660K+ `entity_cooccurrences` rows

This hits a 300s `statement_timeout` with 0 rows pruned because Postgres can't complete the per-row evaluation in time.

## Changes

Replaced the correlated `NOT EXISTS` with a set-based `EXCEPT`:

1. Materialize the stale cooccurrence target set: `cooccurrences EXCEPT valid_witness_pairs`
2. DELETE joining against the materialized stale set

This produces the **identical result set** (verified via bidirectional `EXCEPT` and ordered digest match) while completing in ~3.4s vs >300s for the correlated form — ~**90x faster**.

## Verification

- Same target set: `OLD EXCEPT NEW = 0`, `NEW EXCEPT OLD = 0`
- Ordered digest matches
- Only targets cooccurrences belonging to the bank (via entity → bank_id join)
- Preserves duplicate/cooccurrence semantics

Closes #2660

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -389,22 +389,26 @@ class PostgreSQLOps(DataAccessOps):
         entities_table: str,
         bank_id: str,
     ) -> int:
-        # Scope by joining through entities.bank_id (entity_cooccurrences itself
-        # has no bank_id column — entities don't span banks, so scoping via
-        # entity_id_1 is sufficient).
+        # Materialize the stale cooccurrence target set via EXCEPT and delete
+        # set-wise. The correlated NOT EXISTS self-join (original Pass-3) evaluates
+        # per-row and hits a 300s statement_timeout on large graphs (50K+ entities,
+        # 1M+ unit_entities). The set-based EXCEPT form evaluates once and the
+        # DELETE joins against the materialized stale set — same result, ~90x faster.
         result = await conn.execute(
             f"""
             DELETE FROM {ec_table} c
-            USING {entities_table} e
-            WHERE e.id = c.entity_id_1
-              AND e.bank_id = $1
-              AND NOT EXISTS (
-                  SELECT 1
-                  FROM {ue_table} u1
-                  JOIN {ue_table} u2 ON u1.unit_id = u2.unit_id
-                  WHERE u1.entity_id = c.entity_id_1
-                    AND u2.entity_id = c.entity_id_2
-              )
+            USING (
+                SELECT c2.entity_id_1, c2.entity_id_2
+                FROM {ec_table} c2
+                JOIN {entities_table} e ON e.id = c2.entity_id_1
+                WHERE e.bank_id = $1
+                EXCEPT
+                SELECT DISTINCT u1.entity_id, u2.entity_id
+                FROM {ue_table} u1
+                JOIN {ue_table} u2 ON u1.unit_id = u2.unit_id
+            ) AS stale
+            WHERE c.entity_id_1 = stale.entity_id_1
+              AND c.entity_id_2 = stale.entity_id_2
             """,
             bank_id,
         )


handnewb:
Superseded by upstream changes that address the same problem more comprehensively.

The upstream already resolved this with two complementary fixes:
- **#2473** — Replaced the correlated self-join with an INTERSECT-based staleness predicate (same ~90x performance gain as the EXCEPT approach proposed here)
- **#2529** — Added ordered locking (CTE + FOR UPDATE) for deadlock avoidance

The upstream solution is strictly better: it achieves the same performance improvement while also preventing deadlocks. Closing as superseded.

PR #2799, opened 2026-07-17
security(pg0): add HINDSIGHT_API_PG0_PASSWORD env var and log warning on default

handnewb:
## Summary

The pg0 embedded PostgreSQL database used hardcoded credentials (`user='hindsight'`, `password='hindsight'`) with no way to override except by passing explicit kwargs to `EmbeddedPostgres()`.

While pg0 only listens on localhost by default, containerized deployments where the pg0 port is accidentally mapped to the host create a risk — an attacker on the same network could connect with known credentials.

## Changes

- Added `HINDSIGHT_API_PG0_PASSWORD` env var to override the default password
- Added `_resolve_default_password()` function that checks the env var first
- `EmbeddedPostgres.__init__()` now uses `None` as default for `password`, resolving via `_resolve_default_password()`
- Log `WARNING` when the hardcoded default is active
- **Backward compatible**: all existing callers continue to work unchanged

## Verification

- Explicit password kwarg still takes precedence (e.g., tests)
- Env var `HINDSIGHT_API_PG0_PASSWORD` overrides the hardcoded default
- Default behavior unchanged (same password, just with a warning log)

Found during cybersecurity audit.

### DIFF
--- a/hindsight-api-slim/hindsight_api/pg0.py
+++ b/hindsight-api-slim/hindsight_api/pg0.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


nicoloboschi:
Superseded by #2832, which addresses the same pg0 hardcoded-credentials concern more generally: rather than a password-only `HINDSIGHT_API_PG0_PASSWORD` env var, it carries both user and password in the existing `HINDSIGHT_API_DATABASE_URL` via `pg0://user:pwd@instance:port`. Closing in favor of that approach — thanks for surfacing this from the security audit.

Issue #2800, opened 2026-07-18
[BUG] Mental model generator returns '0 results' even when direct recall finds data


lightcloud00:
## Bug: Mental model generator returns "0 results" even when direct recall finds data

**Hindsight version:** v0.7.2 (`ghcr.io/vectorize-io/hindsight:latest`)
**HINDSIGHT_API_LLM_MODEL:** `MiniMax-M3` (via LiteLLM proxy)

### Summary

The mental model LLM generator (`POST /v1/default/banks/{bank_id}/mental-models`) returns `"I don't have information"` for a topic even when the bank contains 100+ matching facts and the same query via the direct recall endpoint returns 21-43 results.

This is reproducible: 4 out of 5 mental model attempts for an "AOS architecture" topic all failed with "0 results from semantic search" while the same terms via `/memories/recall` returned 21+ facts.

### Reproduction

**Setup:** 18 AOS canon documents ingested into the `claude_code` bank via `POST /v1/default/banks/claude_code/memories` with `tags: ["aos", "aos:<slug>", "path:aos/..."]`. Documents include AOS-INVENTORY, aos-master-plan (53K, 955 lines), AOS Session Summary, AOS Local-Model Guardrail, etc.

After ingestion, 18 aos-canon documents appear in `GET /v1/default/banks/claude_code/documents`. The bank has 22,560 nodes.

**Direct recall works:**

```bash
curl -X POST http://localhost:8888/v1/default/banks/claude_code/memories/recall \
  -H "Content-Type: application/json" \
  -d '{"query": "AOS architecture council slot Mavis Hermes OpenCode Codex", "max_tokens": 2000, "budget": "high"}'
# Returns 21-43 facts about AOS, council, slots, m3-supervisor, hs27-executor, etc.
```

**Mental model fails (same terms):**

```bash
curl -X POST http://localhost:8888/v1/default/banks/claude_code/mental-models \
  -H "Content-Type: application/json" \
  -d '{
    "id": "aos-architecture-test",
    "name": "AOS Architecture",
    "source_query": "AOS canon documents ingested with aos tag. Look for aos-canon documents. Search for: aos-inventory aos-master-plan AOS-INVENTORY Fleet Intelligence AOS Completion Summary AOS Gaps Analysis AI-INFRA-GAP-HANDOFF CREDVAULT provider audit state awareness audit AOS Final Day Lessons.",
    "tags": ["aos", "architecture", "mental-model"],
    "max_tokens": 4096
[...]

PR #2804, opened 2026-07-18
fix(mental-models): align stale checks with refresh tag scope

r266-tech:
## Summary

- make mental-model staleness use the same resolved tag filtering as refresh
- preserve tagged-model `all_strict` defaults and explicit `tags_match` behavior
- support compound `trigger.tag_groups` when deciding whether a scheduled refresh is stale
- keep `fact_types` as an additional scope constraint

## Why

The scheduled-refresh gate and the refresh itself currently resolve tag scope differently. Refresh uses `_resolve_refresh_tag_filtering()`, including `tag_groups` and the tagged-model `all_strict` default, while `compute_mental_model_is_stale()` ignores `tag_groups` and defaults tagged models to `any`.

That can mark a model stale because of an out-of-scope memory, spend an unnecessary refresh LLM call, and then refresh against a different (or empty) corpus. This was surfaced while tracing #2800.

## Testing

- `uv run pytest tests/test_mental_models.py::TestMentalModelStaleness -q -n 0` (12 passed)
- `uv run ruff format hindsight_api/engine/memory_engine.py tests/test_mental_models.py`
- `uv run ruff check hindsight_api/engine/memory_engine.py`

The regression coverage includes default `all_strict`, explicit `any`, compound tag groups, untagged/global behavior, and flat/group tag filters composed with `fact_types`, including multi-group near misses.

Related to #2800.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -11215,17 +11215,16 @@ class MemoryEngine(MemoryEngineInterface):
         """Check whether a mental model is out of date.
 
         A mental model is stale when a memory in its **scope** has been ingested after
-        ``last_refreshed_at``. The scope is defined by the model's ``tags`` +
-        ``trigger.tags_match`` semantics (``any`` / ``all`` / ``any_strict`` /
-        ``all_strict``, matching recall semantics) and its ``trigger.fact_types`` filter
-        when set. Memories still pending consolidation are included because they are
-        already rows in ``memory_units``; no separate ``pending_consolidation`` signal is
-        needed — it would bypass the tag scope and falsely flag unrelated MMs.
+        ``last_refreshed_at``. The scope uses the same resolved flat-tag or compound
+        ``trigger.tag_groups`` filtering as the refresh it gates, plus the
+        ``trigger.fact_types`` filter when set. Memories still pending consolidation are
+        included because they are already rows in ``memory_units``; no separate
+        ``pending_consolidation`` signal is needed — it would bypass the tag scope and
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Verified this at `d0fbf450` on macOS / Python 3.11 with the embedded Postgres and local embedding/reranker fixtures:

- `uv run pytest tests/test_mental_models.py::TestMentalModelStaleness -q -n 0` — 12 passed
- `uv run ruff check hindsight_api/engine/memory_engine.py` — passed
- `uv run ruff format --check hindsight_api/engine/memory_engine.py tests/test_mental_models.py` — passed

The added cases cover the important scope boundaries: default `all_strict`, explicit `any`, compound `tag_groups`, and composition with `fact_types`. The stale gate now uses the same resolved filtering helpers as refresh, so an out-of-scope memory no longer schedules a refresh against a different corpus.

One scope note for #2800: this fixes stale-check/refresh disagreement, but it intentionally does not change the refresh recall semantics that produced the reporter's zero-result case; the documented `tags_match: any` or narrower model tags remain the relevant configuration path there.

nicoloboschi (approved):
Approve. Correct root-cause fix: routes the scheduled-refresh staleness gate through the same `_resolve_refresh_tag_filtering` + tag/tag_groups where-clause builders the refresh corpus uses, so the gate and the refresh it triggers no longer disagree (old gate defaulted tagged MMs to `any` and ignored `tag_groups`, flipping is_stale on out-of-scope memories and burning no-op refresh calls). Param-offset threading, clause parenthesization, and untagged/global behavior all verified; ran TestMentalModelStaleness (12 passed) with the patch applied.

PR #2810, opened 2026-07-19
fix(opencode): queue explicit retains asynchronously

RemiKalbe:
## Summary
- submit explicit `hindsight_retain` calls with `async: true` and return the operation ID immediately
- add `hindsight_operation_status` so agents can poll pending, processing, completed, failed, cancelled, and not-found operations
- update unit/live integration coverage and synchronize the package, integration, and generated skill documentation

## Motivation
Explicit retains currently wait for fact extraction and indexing to finish. Behind a reverse proxy such as Cloudflare, that can exceed the request timeout and return a 524 even though Hindsight supports asynchronous retain operations already. Auto-retain already uses the asynchronous path; this makes the explicit tool consistent with it.

## Verification
- `npm test` in `hindsight-integrations/opencode`: 121 passed, 3 skipped
- `npm run build` in `hindsight-integrations/opencode`
- `./scripts/hooks/lint.sh`
- generated Hindsight docs skill link validation

### DIFF
--- a/hindsight-docs/docs-integrations/opencode.md
+++ b/hindsight-docs/docs-integrations/opencode.md
@@ -6,7 +6,7 @@ description: "Add long-term memory to OpenCode with Hindsight. Automatically cap
 
 # OpenCode
 
-Persistent long-term memory plugin for [OpenCode](https://opencode.ai) using [Hindsight](https://vectorize.io/hindsight). Automatically captures conversations, recalls relevant context on session start, and provides retain/recall/reflect tools the agent can call directly.
+Persistent long-term memory plugin for [OpenCode](https://opencode.ai) using [Hindsight](https://vectorize.io/hindsight). Automatically captures conversations, recalls relevant context on session start, and provides asynchronous retain, operation status, recall, and reflect tools the agent can call directly.
 
 ## Quick Start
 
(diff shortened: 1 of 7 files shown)


koriyoshi2041 (approved):
Verified at a5a879b locally: the explicit retain now queues with async=true, preserves auth/bank URL handling for operation polling, and covers missing operation IDs plus pending/error status paths. The focused OpenCode suite passed 121 tests (3 skipped), and the package build completed successfully. I did not find a blocker.

RemiKalbe:
Anything I need to do to get this merged @koriyoshi2041 ?

handnewb (commented):
Great improvement! Making explicit retains async is a real-world need — reverse proxies timeout before extraction finishes. Auto-retain already uses the async path, so this aligns the two code paths. The operation_status polling tool is a nice addition for agent UX. One suggestion: consider documenting the async behavior change in the opencode integration README as a breaking change (callers relying on sync behavior for their own timeout handling need to know). Otherwise looks solid!

nicoloboschi:
Closing: superseded by **Coding Agents**, where this asymmetry can't exist — retains are *always* async.

The option to submit synchronously was removed outright in v0.2.0: nothing in the plugin can afford to block a coding agent's hook on the extraction pipeline, so the explicit `hindsight_retain` tool and auto-retain now take the same durable server-queued path. Issue #3206 is closed against it.

`npx @vectorize-io/hindsight-coding-agents install opencode` ([docs](https://hindsight.vectorize.io/sdks/integrations/coding-agents)). Thanks — the observation that the two paths disagreed is what made "always async" the obvious call rather than a toggle.

Issue #2519, opened 2026-07-02
Claude Code / Codex: directoryBankMap exact-match fails when agent navigates to subdirectories (cwd is dynamic)


mmahao:
## Bug Description

`directoryBankMap` in `hindsight-integrations/claude-code/scripts/lib/bank.py` (`derive_bank_id()`) uses exact `==` comparison to match paths:

```
normalized_cwd = os.path.normcase(os.path.normpath(cwd))
for dir_path, bank_id in dir_map.items():
    if os.path.normcase(os.path.normpath(dir_path)) == normalized_cwd:
        return ...
```

**The problem: `cwd` is dynamic. When the agent executes `cd` into a subdirectory, exact matching inevitably fails.**

Claude Code's hook docs explicitly define `cwd` = "Current working directory when the hook is invoked", and there's a dedicated `CwdChanged` event (fires when agent runs `cd`). This means `cwd` changes in real-time as the agent navigates directories.

## Concrete Scenario

```json
{
  "directoryBankMap": {
    "/Users/mahao/project": "my-bank"
  }
}
```

1. Agent starts in `/Users/mahao/project` → `cwd == "/Users/mahao/project"` → matches ✅
2. Agent runs `cd src/` → `cwd == "/Users/mahao/project/src"` → **NO match** ❌ → falls back to default bank
3. Agent runs `cd ../` back to root → `cwd == "/Users/mahao/project"` → matches ✅
4. Agent runs `cd src/utils/` → `cwd == "/Users/mahao/project/src/utils"` → **NO match** ❌

**Result: memories silently written to the wrong bank. No error, no log.**

## User's Real Scenario (from issue reporter — @mmahao)

I use both Claude Code and Codex daily across multiple project directories, with `directoryBankMap` configured to route each project to its dedicated Hindsight bank for memory isolation:

```json
{
  "directoryBankMap": {
    "/Users/mahao/development/project-a": "project-a-bank",
    "/Users/mahao/development/project-b": "project-b-bank",
    "/wandian/ecommerce": "wandian-yinhe-coder"
  }
}
```

**What happens in a typical work session:**

1. Start Claude Code in `/wandian/ecommerce` → bank = `wandian-yinhe-coder` ✅
2. Claude navigates to `src/main/java/com/wandian/controller/` to fix a bug → cwd changes → **bank silently switches to `mahao` (default)** ❌
3. All memories from this turn (decisions, patterns, bug context) go to the wrong bank
[...]

PR #2812, opened 2026-07-20
fix(claude-code): resolve directoryBankMap by nearest ancestor

r266-tech:
## Summary

- make absolute `directoryBankMap` entries apply to descendant working directories
- choose the nearest configured ancestor when mappings are nested
- preserve exact matching for relative keys and existing symlink equivalence
- fail closed for ambiguous canonical aliases and prevent symlinked mappings from widening into unrelated descendant trees

This addresses the Claude Code routing failure reported in #2519: after the agent changes into a project subdirectory, retain/recall operations continue using the configured project bank instead of silently falling back.

The issue also notes Codex dynamic project IDs. I intentionally left that separate: changing Codex's existing directory-derived IDs requires an explicit migration/compatibility design for memories already stored under legacy subdirectory banks.

## Verification

- `PYTHONPATH=scripts .venv/bin/python -m pytest tests/test_bank.py -q` — 58 passed
- focused Ruff check and format check
- generated docs-skill mirror regenerated and link validation passed
- `git diff --check`

## Review notes

Adversarial review drove explicit coverage for nested roots, sibling-prefix boundaries, Windows drive/UNC behavior, symlink containment, canonical alias conflicts, order independence, and mutable-realpath snapshot consistency. Exact symlink equality and POSIX case semantics remain compatible with current behavior; the new lexical guard applies to descendant expansion.

### DIFF
--- a/hindsight-docs/docs-integrations/claude-code.md
+++ b/hindsight-docs/docs-integrations/claude-code.md
@@ -152,7 +152,7 @@ A **bank** is an isolated memory store — like a separate "brain." These settin
 | `bankIdPrefix` | — | `""` | A string prepended to all bank IDs — both static and dynamic. Useful for namespacing (e.g. `"prod"` or `"staging"`). |
 | `agentName` | `HINDSIGHT_AGENT_NAME` | `"claude-code"` | Name used for the `agent` field in dynamic bank ID derivation. |
 | `resolveWorktrees` | — | `true` | When deriving the `project` field, resolve git worktrees to the **main repository's basename** so that all worktrees of the same repo share one bank. Set to `false` to use the literal working directory basename instead (each worktree gets its own bank). |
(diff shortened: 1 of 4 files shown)


nicoloboschi:
Closing: superseded by **Coding Agents**, which resolves paths exactly the way you proposed.

Its `mapPathToBank` matches by longest path *prefix* — `cwd === p || cwd.startsWith(p + sep)`, longest match wins — so a session that `cd`s into a subdirectory keeps its mapped bank instead of silently falling through to the default. Same fix, same reasoning; the corresponding issues (#3133, #2519) are closed against it.

`npx @vectorize-io/hindsight-coding-agents install claude-code` ([docs](https://hindsight.vectorize.io/sdks/integrations/coding-agents)). Thanks for working the ancestor-resolution logic out.

PR #2813, opened 2026-07-20
fix(mental-models): default refresh tag matching to any

### LINKED ISSUE #2808: Mental model refresh returns empty content when model has multiple tags (all_strict default filters out all memories)
## Bug Description

Mental model refresh (async path via `submit_async_refresh_mental_model`) consistently returns empty content ("no relevant memories found") when the mental model has multiple tags, while the exact same `source_query` run through the `reflect` MCP tool synchronously returns all expected hits.

The root cause is the default tag-matching mode in `_resolve_refresh_tag_filtering`.

## Steps to Reproduce

1. Store several `world`/`experience` memories, each with a **single** tag (e.g. `["project:status"]`, `["user:confirmation"]`, `["tooling"]`). No memory carries more than one tag.
2. Create a mental model **with multiple tags**, e.g. `["projects", "mental-model"]`, and a `source_query` that semantically matches the stored memories:
   ```jsonc
   {
     "name": "Current projects and tasks",
     "source_query": "Which projects is the user currently working on? Status, focus, and priority of each.",
     "tags": ["projects", "mental-model"]
   }
   ```
3. Trigger a refresh (`refresh_mental_model`).
4. Read the generated content.

## Expected Behavior

Refreshed content synthesizes the stored memories (same as calling `reflect` with the `source_query` directly).

## Actual Behavior

Refresh produces content like:

> *"No records found for the user's current projects in the memory bank."*
> *"I don't have information regarding the user's current projects… The retrieved memory bank and observation logs are empty for these queries."*

Meanwhile, the **same** query via the `reflect` MCP tool returns 18 observation hits and synthesizes them correctly. Verified across 4 separate refresh attempts (create + 2x `refresh_mental_model` + `clear_mental_model` + refresh), all empty.

`list_memories` / `recall` both surface 34 stored facts; only the async refresh path sees an empty set.

## Root Cause

`hindsight-api-slim/hindsight_api/engine/memory_engine.py:843`:

```python
trigger_tags_match = trigger_data.get("tags_match")
tags_match: TagsMatch = trigger_tags_match if trigger_tags_match else ("all_strict" if model_tags else "any")
```

When a mental model has tags but the persisted `trigger` JSON does not store an explicit `tags_match`, the refresh defaults to **`all_strict`** (require ALL tags to match). This value propagates to `tool_search_observations` and `tool_recall` during the reflect agent run, so any memory not carrying *every* one of the model's tags is filtered out.

With a freshly created mental model there is normally no prior refresh to write `tags_match` into the trigger JSON, so `all_strict` is effectively the default for **any tagged mental model** — which is the common case (the `create_mental_model` API and docs both encourage tagging).

Because users naturally tag mental models broadly (e.g. `["projects", "mental-model"]`) while tagging memories narrowly (single-topic tags), `all_strict` filters out essentially everything, making the feature appear broken out of the box.

## Suggested Fix Options

- **Change the default** from `all_strict` to `any` (consistent with `recall`'s default), OR
- **Inherit `tags_match` from the model config** at creation time (let the caller pick `any`/`all`/`all_strict` explicitly via `create_mental_model` params), OR
- **At minimum, document** that multi-tag mental models must set `tags_match` explicitly, and surface the resolved `tags_match` in `get_mental_model` output so users can diagnose silent empty refreshes.

## Environment

- Version: `0.8.4` (hindsight-api-slim)
- Commit: `327aa05e80c89e2f02e9122123469f8b0bd91d0c`
- LLM Provider: reproduced independent of provider
- Storage: local

## Workaround

For anyone hitting this: create mental models **without tags**, or with a single tag, or ensure every relevant memory carries all the model's tags.

koriyoshi2041:
## Problem

Tagged mental models without an explicit `trigger.tags_match` refresh with `all_strict`, while recall, reflect, and staleness checks default to `any`. With multiple model tags this can mark a model stale and then refresh it from an empty corpus. Fixes #2808.

## Fix

- default refresh filtering to `any`
- preserve explicit modes such as `all_strict`
- update the refresh integration expectation and add focused resolver regressions

## Test

- `uv run --directory hindsight-api-slim pytest tests/test_mental_models.py -k "TestRefreshTagFiltering" -q` (2 passed)
- pre-commit lint hooks passed
- `uv run --directory hindsight-api-slim ruff format --check hindsight_api/engine/memory_engine.py tests/test_mental_models.py`

The database-backed staleness test could not complete locally because the test database fixture did not become ready; the existing overlap test already asserts the `any` staleness behavior, and hosted CI should cover the integration path.

## Risk

The implicit scope becomes broader for tagged models and includes untagged memories. Deployments that require strict isolation can retain it by setting `trigger.tags_match` to `all_strict` explicitly.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2010,9 +2010,9 @@ class MentalModelTrigger(BaseModel):
         default=None,
         description=(
             "Override how the model's tags filter memories during refresh. "
-            "If not set, defaults to 'all_strict' when the model has tags (security isolation) "
-            "or 'any' when the model has no tags. "
-            "Set to 'any' to include untagged memories alongside tagged ones during refresh."
+            "If not set, defaults to 'any', matching recall and reflect: memories with any model tag "
+            "and untagged memories are included. Set an explicit strict mode such as 'all_strict' "
+            "when refresh must exclude untagged or partially matching memories."
         ),
     )
     tag_groups: list[TagGroup] | None = Field(
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -829,7 +829,7 @@ def _resolve_refresh_tag_filtering(
     Priority:
     - If trigger has tag_groups, use those (overrides flat tags entirely)
     - If trigger has tags_match, use model's tags with that match mode
-    - Otherwise default to all_strict when tags present (security isolation)
+    - Otherwise default to any, matching recall, reflect, and staleness checks
     """
     trigger_tag_groups = trigger_data.get("tag_groups")
     if trigger_tag_groups is not None:
@@ -840,7 +840,7 @@ def _resolve_refresh_tag_filtering(
         return RefreshTagFiltering(tags=None, tags_match="any", tag_groups=parsed)
 
     trigger_tags_match = trigger_data.get("tags_match")
-    tags_match: TagsMatch = trigger_tags_match if trigger_tags_match else ("all_strict" if model_tags else "any")
+    tags_match: TagsMatch = trigger_tags_match if trigger_tags_match else "any"
     return RefreshTagFiltering(tags=model_tags, tags_match=tags_match, tag_groups=None)
 
 
--- a/hindsight-api-slim/tests/test_mental_models.py
+++ b/hindsight-api-slim/tests/test_mental_models.py
@@ -8,7 +8,7 @@ import uuid
 
 import pytest
 
-from hindsight_api.engine.memory_engine import MemoryEngine, fq_table
+from hindsight_api.engine.memory_engine import MemoryEngine, _resolve_refresh_tag_filtering, fq_table
 from hindsight_api.engine.retain import embedding_utils
 from tests.llm_judge import assert_meets_criteria, evaluate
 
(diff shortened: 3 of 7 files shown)


ebarkhordar (commented):
The default change looks right for the staleness problem in #2808, and the Risk note is a fair statement of the tradeoff. One thing the diff misses: the API description for the field being changed still advertises the old default.

`MentalModelTrigger.tags_match` at `hindsight-api-slim/hindsight_api/api/http.py:2013` reads:

> If not set, defaults to 'all_strict' when the model has tags (security isolation) or 'any' when the model has no tags.

That is the public description for exactly the branch this PR rewrites at `memory_engine.py:843`, and it ships in the OpenAPI schema, so anyone reading the API reference or a generated client is told the default still isolates. The resolver docstring at `memory_engine.py:832` was updated in the same hunk; this one appears to have been missed.

Worth correcting in this PR rather than later, because the stale text errs toward promising isolation that the code no longer provides, which is the direction that costs someone something. The phrase "security isolation" is also load-bearing elsewhere in the repo: `consolidation/consolidator.py:2056` and `:2074` use `all_strict` with a `SECURITY:` comment about preventing cross-tenant and cross-scope reads. That path is untouched here, so this is not a claim that your change breaks it. It is why I would not leave a doc string promising the old default in place.

One question while you are in here: does the `tag_groups` description just below (`http.py:2020`) need the same pass, or is its wording still accurate under the new default?

koriyoshi2041:
Thanks — fixed the public field description in 5076c6d07 so it now states the implicit any behavior and points strict-scope users to an explicit strict mode. The tag_groups description remains accurate: that branch overrides flat model tags entirely and passes the compound expression through unchanged.

nicoloboschi:
we're not going to change the default and break compatibility here. I'm working on a pr to improve configurability and documentation, thanks

koriyoshi2041:
Understood — preserving the existing default and improving the explicit configuration/docs is the safer compatibility path. Thanks for the direction.

PR #2835, opened 2026-07-20
fix(lint): ruff-format test_query_analyzer.py to unblock verify-generated-files

dcbouius:
`hindsight-api-slim/tests/test_query_analyzer.py` landed on `main` with formatting that `ruff format` doesn't consider stable (multi-line asserts/calls that fit within the 120-col limit collapse to one line).

The `verify-generated-files` CI job runs `ruff format` and fails on the resulting diff, so it has been **red on every open PR that runs it** — #2826 and #2828 both fail `verify-generated-files` for this reason, independent of their own changes. That ambient red masks real failures.

Format-only, no behavioural change. `ruff format --check .` is now clean across `hindsight-api-slim` (535 files already formatted; this was the lone drift).

### DIFF
--- a/hindsight-api-slim/tests/test_query_analyzer.py
+++ b/hindsight-api-slim/tests/test_query_analyzer.py
@@ -875,9 +875,7 @@ def test_query_analyzer_rejects_bare_word_false_positive(query_analyzer):
 
     analysis = query_analyzer.analyze("what did we discuss", reference_date)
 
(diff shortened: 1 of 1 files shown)


dcbouius:
Redundant — the format drift in test_query_analyzer.py was already fixed on main by another commit. Diff vs current main is empty.

Issue #2809, opened 2026-07-19
Add Hindsight to the Umbrel App Store


salmonumbrella:
### Request

Hindsight would be a great fit for the self-hosted [Umbrel App Store](https://apps.umbrel.com/).  
Could you add an official Hindsight app with persistent storage and ARM64 support?

### Why

The existing Docker image should make packaging fairly straightforward.  
An official app would make Hindsight much easier to install and update on Umbrel devices.

### Resources

Umbrel’s app repository and submission instructions are available at [getumbrel/umbrel-apps](https://github.com/getumbrel/umbrel-apps).

benfrank241:
Thanks for the suggestion @salmonumbrella, this is a great fit and we would like to do it.

The good news is that the infrastructure is already in place:

- **ARM64:** our images are built multi-arch (`linux/amd64` + `linux/arm64`), so Umbrel Home and Pi-class devices are covered.
- **Persistent storage:** the standalone image keeps its embedded Postgres data under a single directory that is pre-created for UID 1000, so it maps cleanly onto an Umbrel data volume and survives updates and restarts.
- **Browser UI:** the standalone image bundles the API and the control-plane web UI together and opens straight to a dashboard, which lines up with Umbrel's "understandable from the browser after install" requirement.

**Plan:** we will package Hindsight for the official [Umbrel App Store](https://github.com/getumbrel/umbrel-apps) using the standalone image pinned by digest.

One thing to flag up front on the setup UX: the initial release will ask for an LLM API key (OpenAI, Anthropic, etc.) during setup, since retain / recall / reflect need an LLM. Fully local operation via Ollama is something we can look at as a follow-up, but the first cut will expect a key.

I will post back here with the PR once it is up. Thanks for the nudge and the pointer to the submission repo.

benfrank241:
Quick update and a question on the LLM setup, since it shapes the packaging.

Umbrel apps are pure docker-compose bundles, so the app cannot prompt you for an API key at install and editing env means SSH (which Umbrel discourages). Rather than require a cloud key, the cleanest fit is to have Hindsight use your **local Ollama app** as its LLM backend: it is declared as a dependency, runs fully offline, and needs no key. You would install the Ollama app and pull a model (e.g. `ollama pull llama3.1:8b`).

Before we finalize the package, two questions:

1. Are you good with a **local-Ollama-first** setup (fully offline, no API key)? That is what most Umbrel users prefer, and it is what we have packaged.
[...]

PR #2840, opened 2026-07-20
feat: server-level LLM config settable from the control plane (Umbrel enablement)

benfrank241:
## Summary

Adds a **server-level LLM configuration** that a self-hosted operator can set at runtime from the control plane, instead of only via environment variables. This unblocks packaging Hindsight for the Umbrel App Store (#2809), whose core rule is "configure from the browser, no CLI / no file edits."

Today a fresh install with the default `openai` provider **won't boot** without a key (`MemoryEngine.__init__` raises), and the backend deliberately blocks LLM credentials from bank config. This PR lets the instance boot keyless (`provider=none`) and have the user supply their key in the browser afterward.

## What's included

**Backend**
- New `server_settings` table (dual-dialect migration) + `ServerSettingsStore` persisting the base LLM config (provider/model/api_key/base_url). API key is plaintext by default, or encrypted at rest when `HINDSIGHT_API_SETTINGS_ENC_KEY` is set (Fernet).
- `MemoryEngine.reconfigure_llm(...)` rebuilds the four LLM provider objects in place, so a saved key takes effect **without a restart** (the provider client binds at construction, so config-resolution alone was insufficient). Extracted the shared `_build_operation_llm_configs` so `__init__` and reconfigure can't drift.
- `ConfigResolver` layers the saved base LLM above env (below tenant/bank) for internal resolution; the bank-config **read** path still strips all LLM credentials.
- New endpoints `GET/PUT/DELETE /v1/default/server/llm-config` + `POST .../health`, behind `HINDSIGHT_API_ENABLE_SERVER_LLM_CONFIG_API` (default on) and **single-tenant only** (auto-disabled when a tenant extension is loaded). Persistence/orchestration lives in `MemoryEngine` methods; handlers only validate + shape responses. The API key is never returned (only `api_key_is_set`) and is never written to audit logs.
- `server_settings` added to `BACKUP_TABLES`.

**Control plane**
- New non-bank-scoped **Settings** page (provider/model/write-only API key/base URL + Test connection), proxy routes, `api.ts` client methods, and a first-run banner on the dashboard prompting configuration when the instance LLM is unset. `settings` i18n namespace added to all locales.

**Docs / generated**
- Configuration docs + regenerated OpenAPI spec, skill mirror, and Python/TS/Go/Rust SDK clients.

## Testing

- New `tests/test_server_settings.py` (11 tests): store round-trip + encryption, `reconfigure_llm` none→enabled, resolver layering + precedence, credential redaction on read, bank-PATCH still rejects LLM credentials, and endpoint GET/PUT/DELETE/health incl. write-only-preserve.
- Backup guard test (`test_backup_tables_covers_entire_schema`), `test_bank_attribution`, `test_hierarchical_config`, `test_migration_shape`, retain/reflect suites all green. `ruff` + `ty` clean; `./scripts/hooks/lint.sh` passes.

## End-to-end (Umbrel path)

Boot the standalone image with `HINDSIGHT_API_LLM_PROVIDER=none` → API boots → open the control plane → first-run banner → Settings → save provider/key → memory ops work **without a restart**. The Umbrel package's `docker-compose.yml` is updated to `provider=none` on the packaging branch.

## Notes / open questions for review

- **API key at rest:** plaintext by default (single-tenant DB is the trust boundary), with optional envelope encryption via `HINDSIGHT_API_SETTINGS_ENC_KEY`. Open to requiring encryption.
- **Multi-replica:** a runtime reconfigure only affects the process serving the request; others re-apply on their next `initialize()`. Fine for the single-container Umbrel target; no HA broadcast in v1.
- **Scope:** base LLM only in v1 (retain/reflect/consolidation inherit); no per-op overrides in the UI.
- **Cloud/multi-tenant:** the endpoint is hard-disabled when a tenant extension is loaded.


benfrank241:
Closing this in favor of a much simpler approach that needs no core changes.

The goal was to let the Umbrel-packaged app be configured from the browser, since Umbrel apps are pure docker-compose bundles with no host CLI and no user-facing env/file config passthrough. But making the server LLM config settable via API is a broader surface than we want to own right now.

Instead, we'll package Hindsight for Umbrel to use the official **Ollama** Umbrel app as its LLM backend (`dependencies: [ollama]`, `provider=ollama` pointed at the local Ollama container). That's keyless, boots out of the box, and matches how other Umbrel apps (open-webui, blinko) already work — no changes to Hindsight core. Cloud-key support can be a later follow-up if there's demand.

The branch `feat/server-llm-config` stays around if we want to revisit the in-app config UI down the line.

PR #2844, opened 2026-07-21
fix(reflect): emit canonical OpenAI tool result messages

Jordan-Jarvis:
## Summary
- remove the noncanonical `name` field from Reflect role=`tool` messages
- derive Gemini function-response names from the preceding assistant tool calls by `tool_call_id`
- reject unknown tool-result IDs, duplicate IDs within a turn, and missing results
- preserve call-ID reuse across completed turns

## Why
OpenAI-compatible provider boundaries reject role=`tool` messages that include a `name` field. Gemini still requires a function name in its native `FunctionResponse`, so the provider adapter must derive that provider-specific field from the canonical assistant tool call instead of requiring Reflect to emit a noncanonical message shape.

## Validation
- Ruff formatting and lint passed for changed provider/Reflect files
- 34 focused Gemini/Reflect tests passed
- full two-file run produced 69 passing tests; three existing real-LLM fixtures could not initialize because no `HINDSIGHT_API_LLM_API_KEY` was configured locally
- `git diff --check` passed

---

Replacement for #2766. The original PR became unreopenable after its submitting fork was deleted. This PR restores the exact reviewed head commit and preserves the original implementation without modification.


Jordan-Jarvis:
Ready after #2843. This is the exact restored replacement for unreopenable #2766 and is already merged with current upstream main e07ca1bd. All upstream checks are green. It should follow #2843 because canonical tool-result emission depends on the typed tool-choice/provider contract; I will resync immediately if GitHub reports drift after #2843 lands.

Jordan-Jarvis:
cc @nicoloboschi — all checks are green; please merge after #2843.

nicoloboschi (approved):
Reviewed for Gemini/Vertex backwards compatibility. Both use the same GeminiLLM adapter; deriving the FunctionResponse name from tool_call_id is safe — reflect (the sole builder) always emits id/tool_call_id, LLMToolCall.id is required, and the incremental-cache boundary is captured at start-of-turn so cached prefixes/deltas never split an assistant+results pair. Cross-turn gemini_N id reuse is handled (ids popped before the next turn). Removing the noncanonical `name` also fixes OpenAI-compatible endpoints; Anthropic/Codex/Claude-Code never read it; LiteLLM-routed Vertex derives the name the same way. LGTM.

PR #2845, opened 2026-07-21
fix(llm): emit OpenAI strict JSON schemas

Jordan-Jarvis:
## Summary
- serialize typed Pydantic response models with a dedicated OpenAI strict JSON Schema generator
- close every model object with `additionalProperties: false`
- require every declared property while preserving nullable unions and removing null defaults
- apply the serializer to OpenAI-compatible, LiteLLM, Codex, and direct batch-retain paths

## Root cause
`HINDSIGHT_API_LLM_STRICT_SCHEMA=true` currently sends raw `model_json_schema()` output. OpenAI-compatible strict structured-output endpoints reject those schemas because objects are not closed and defaulted fields are omitted from `required`. This was previously identified in the review of #1954 and is reproducible during native retain fact extraction.

The fix does not coerce incoming data or add a fallback. Typed response models are serialized directly through Pydantic into the provider-required strict wire schema.

## Validation
- targeted strict-schema tests: 13 passed
- Ruff format/check passed
- repository pre-commit hooks passed
- generated docs skill refreshed by the repository hook
- `git diff --check` passed

---

Replacement for #2807. The original PR became unreopenable after its submitting fork was deleted. This PR restores the exact reviewed head commit and preserves the original implementation without modification.


Jordan-Jarvis:
Ready for maintainer merge. This is the exact restored replacement for unreopenable #2807, previously approved by @koriyoshi2041. I synced it with current upstream main to include the ZCode docs inventory, then revalidated the strict-schema/provider suite (33/33), Ruff, formatting, and docs inventory. New head: 8eaaab67. All upstream checks are green.

Jordan-Jarvis:
cc @nicoloboschi @koriyoshi2041 — all checks are green and the prior approval/context is linked above.

koriyoshi2041 (approved):
Revalidated the restored change at 8eaaab67 on current main. The strict-schema regression passed 2/2; adjacent fact-extraction retry coverage passed 12/12; Ruff check, Ruff format, and diff check passed. The two live-provider tests require local LLM configuration and failed before exercising this patch; hosted applicable checks are green. The schema and focused tests are unchanged from the previously approved implementation, and I do not see a blocker.

PR #2847, opened 2026-07-21
fix(memory): preserve bank attribution during curation

Jordan-Jarvis:
## Summary

- bind `update_memory_unit` to the same per-bank context used by retain, recall, reflect, and worker operations
- preserve the canonical OpenAI `user=<bank_id>` attribution on embeddings generated while editing or restoring a memory
- use exact canonical value types in the bank-context decorator
- strengthen the existing bank-attribution regression test without adding a new test

## Why

Editing a memory re-embeds its text. `update_memory_unit` did not bind `_current_bank_id`, so an OpenAI-compatible embeddings gateway that requires Hindsight bank attribution received no `user` field and rejected the update. The engine already owns the bank ID at this boundary; binding it here keeps update/revert provider calls consistent with the other bank-scoped engine operations.

## Validation

- `uv run pytest tests/test_bank_attribution.py -q` (15 passed)
- `./scripts/hooks/lint.sh` (passed)

The broader local curation suite was not used as a signal because its persistent development database is at migration `d7b2f8a1c934`, newer than the checked-out code, and fails before exercising this change (`invalidated_memory_units.search_vector` schema mismatch). The deterministic regression suite above covers the changed context boundary.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -113,9 +113,9 @@ def _bind_bank_id(
         @functools.wraps(func)
         async def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
             value = sig.bind(*args, **kwargs).arguments.get(arg)
-            if key is not None and isinstance(value, dict):
+            if key is not None and type(value) is dict:
                 value = value.get(key)
-            token = _current_bank_id.set(value if isinstance(value, str) else None)
+            token = _current_bank_id.set(value if type(value) is str else None)
             try:
                 return await func(*args, **kwargs)
             finally:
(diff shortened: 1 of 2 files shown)


Jordan-Jarvis:
Live production-like reproduction from a strict OpenAI-compatible embeddings gateway:

- `PATCH /v1/default/banks/{bank_id}/memories/{memory_id}` reaches `_reembed_memory_text`.
- The outbound embeddings request omits `user` because `update_memory_unit` never binds `_current_bank_id`.
- The gateway rejects the request with `invalid OpenAI embeddings request: missing field user`; Hindsight returns HTTP 500.
- Retain/recall attribution works for the same bank and provider.

The deterministic regression in this PR proves both bank visibility during the update path and reset after the call. All upstream CI checks are green.

nicoloboschi (approved):

PR #2853, opened 2026-07-21
fix(consolidation): separate tags from observation scopes

### LINKED ISSUE #2852: Separate source-tag filtering from exact observation-scope targeting in consolidation
## Problem

`POST /v1/default/banks/{bank_id}/consolidate` currently accepts an `observation_scopes` parameter, but uses it to filter the ordinary `tags` column of source facts:

```sql
tags @> requested_scope
```

This conflates two independent source-fact fields:

- `tags`: ordinary tags attached to the source fact
- `observation_scopes`: the scopes to which the source fact writes observations; the resolved scope becomes the derived observation's `tags`

These fields can be completely different. For example:

```json
{
  "text": "Alice exercises three times per week.",
  "tags": ["source:fitness-app"],
  "observation_scopes": [
    ["user:alice", "topic:exercise"]
  ]
}
```

This source fact is tagged `source:fitness-app`, while the observation it produces is tagged `user:alice, topic:exercise`.

Under the current API, a request named `observation_scopes` actually selects this fact only when it contains `source:fitness-app` in its ordinary `tags`. It does not select the fact by its configured observation scope.

## Origin of this issue

This semantic mismatch was introduced by the original targeted-consolidation work:

- [#1625](https://github.com/vectorize-io/hindsight/issues/1625), opened on May 14, 2026, requested an API for consolidating specific observation scopes.
- [PR #1746](https://github.com/vectorize-io/hindsight/pull/1746), merged on May 26, 2026 in commit [`ac3ab2b`](https://github.com/vectorize-io/hindsight/commit/ac3ab2b54ced3780bf8f64e83437a96d6d5bbe6a), implemented that request.
- The implementation PR explicitly defined a requested `observation_scopes` entry as selecting source memories whose ordinary tags contained the scope, and implemented the predicate as `tags @> scope`.

Therefore the current behavior is not a later regression in otherwise-correct observation-scope targeting. The mismatch was present in the first implementation: the public field was named `observation_scopes`, but the selector was wired to the source fact's ordinary `tags` column. This issue corrects that original field-selection error while preserving the old tag-query capability under the explicit `tags` / `tag_groups` API.

### 1. Expose source-fact tag filtering as tag filtering

The existing behavior should be represented by `tags`, not `observation_scopes`.

It should support the same matching modes already available to other tag-filtering APIs:

- `any`
- `all`
- `any_strict`
- `all_strict`
- `exact`

Compound filters should also be expressible through the existing `tag_groups` model where appropriate.

Example:

```json
{
  "tags": ["source:fitness-app"],
  "tags_match": "exact"
}
```

This filter operates only on the source fact's ordinary `tags`.

### 2. Add true targeting by source-fact observation scopes

`observation_scopes` should select source facts according to the scopes they are configured to write observations into.

The implementation should resolve each source fact's concrete write scopes from its `observation_scopes` configuration, including:

- `combined` / default
- `per_tag`
- `all_combinations`
- `shared`
- explicit `list[list[str]]`

A source fact is eligible when at least one of its resolved write scopes exactly matches at least one requested observation scope:

```python
bool(resolve_write_scopes(source_fact) & requested_observation_scopes)
```

Scope equality must be set equality:

- tag order is irrelevant;
- extra or missing tags do not match;
- `[]` matches only the resolved empty/shared scope.

Example request:

```json
{
  "observation_scopes": [
    ["user:alice", "topic:exercise"]
  ]
}
```

This must select the example source fact above because one of its resolved observation scopes is exactly `{"user:alice", "topic:exercise"}`, even though its ordinary tags are only `{"source:fitness-app"}`.

It must not match a source fact whose only resolved scope is:

```json
["user:alice", "topic:exercise", "tier:summary"]
```

## Why these filters must be separate

Source tags answer:

> Which source facts have these ordinary tags?

Observation scopes answer:

> Which source facts are configured to write observations into this exact scope?

They are different dimensions and should be independently targetable by the consolidation API.

## Compatibility framing

The current behavior should be treated as using the wrong source-fact field, rather than as an alternative `observation_scopes` semantic that needs to be preserved. Compatibility, migration, and release-version handling are outside the scope of this request.

Sanderhoff-alt:
## Context

Targeted consolidation accepts `observation_scopes`, but the previous implementation applied each requested scope to the source fact ordinary `tags` column with array containment. The API name therefore described observation write scopes while the runtime behavior performed an ordinary source-tag query.

This distinction is observable because the fields can intentionally be unrelated. A source fact may use `tags: ["source:fitness-app"]` for provenance while using `observation_scopes: [["user:alice", "topic:exercise"]]` to control where derived observations are written. Filtering the first field through an API named after the second made the request ambiguous and provided no way to target the actual write-scope configuration.

## Problem statement

| Source-fact field | Meaning | Example |
| --- | --- | --- |
| `tags` | Ordinary metadata attached to the source fact and used by tag-filtering APIs | `["source:fitness-app"]` |
| `observation_scopes` | Configuration that resolves into one or more scopes used as tags on derived observations | `[["user:alice", "topic:exercise"]]` |

The old implementation effectively evaluated `source.tags @> requested_observation_scope`. It did not inspect or resolve `source.observation_scopes`. A request could therefore select a source merely because the requested strings happened to be ordinary tags, while missing a source that was actually configured to write to the requested observation scope.

## Goals

- Expose the existing source-fact tag query as an explicitly named `tags` filter.
- Reuse the complete tag matching vocabulary already implemented elsewhere.
- Give `observation_scopes` its literal meaning: select source facts by their resolved observation write scopes.
- Define observation-scope equality as exact, order-independent set equality.
- Preserve both filter dimensions through async submission, worker deserialization, round-limit requeues, progress accounting, and generated API clients.
- Keep PostgreSQL and Oracle behavior aligned.

## Non-goals

- Preserve the previous misuse of `observation_scopes` as an alternate source-tag filter.
- Introduce a compatibility alias or silently infer which field a caller intended.
- Change how a selected source fact generates observations after it enters consolidation.
- Change recall or reflect tag semantics.

## API design

`POST /v1/default/banks/{bank_id}/consolidate` now accepts two independent source-selection dimensions:

| Field | Type | Semantics |
| --- | --- | --- |
| `tags` | `list[str] | null` | Flat filter over the source fact ordinary `tags` field |
| `tags_match` | `any | all | any_strict | all_strict | exact` | Match mode for the flat `tags` filter; defaults to `any` |
| `tag_groups` | `list[TagGroup] | null` | Recursive boolean filter over ordinary source tags; mutually exclusive with `tags` |
| `observation_scopes` | `list[list[str]] | null` | Exact filter over the source fact resolved observation write scopes |

`tags` and `tag_groups` are mutually exclusive because `tag_groups` is the compound form of the same ordinary-tag dimension. `observation_scopes` is separate and may be combined with either ordinary-tag form. When dimensions are combined, a source fact must satisfy both.

### Flat tag matching

The endpoint delegates to the existing tag filter builder, so consolidation uses the same matching vocabulary as the other tag-aware APIs:

| Mode | Source-fact match rule | Untagged sources |
| --- | --- | --- |
| `any` | At least one requested tag is present | Included |
| `all` | Every requested tag is present; extra tags are allowed | Included |
| `any_strict` | At least one requested tag is present | Excluded |
| `all_strict` | Every requested tag is present; extra tags are allowed | Excluded |
| `exact` | The complete source tag set equals the requested tag set | Excluded, except omitted `tags` or `[]` selects only the untagged set |

`tag_groups` supports `{tags, match}` leaf nodes and recursive `{and: [...]}`, `{or: [...]}`, and `{not: ...}` nodes. Each leaf owns its match mode, and top-level groups are ANDed. A flat `tags_match` value does not add an implicit empty-tag filter when `tag_groups` is present.

### Exact observation-scope matching

Each source fact retain-time configuration is resolved through the same helper used by consolidation dispatch:

| Source configuration | Resolved write scopes |
| --- | --- |
| omitted or `combined` | One scope containing the complete ordinary source tag set |
| `per_tag` | One singleton scope per source tag |
| `all_combinations` | Every non-empty combination of source tags |
| `shared` | One empty/global scope |
| explicit `list[list[str]]` | The explicitly declared scopes |

Requested and resolved scopes are normalized to sets. A source is eligible when their intersection is non-empty:

```python
bool(resolve_write_scopes(source_fact) & requested_observation_scopes)
```

Scope equality is exact set equality. Tag order and duplicate entries do not affect identity, but missing or additional tags do. `["user:alice"]` does not match `["user:alice", "topic:exercise"]`, and the reverse does not match either. `[[]]` targets only sources whose resolved scopes include the empty/shared scope. An explicit empty outer list `[]` returns `no_new_memories` before any candidate boundary or page scan, while omitting `observation_scopes` disables this filter.

### Example

Given this source fact:

```json
{
  "text": "Alice exercises three times per week.",
  "tags": ["source:fitness-app"],
  "observation_scopes": [["user:alice", "topic:exercise"]]
}
```

Ordinary source metadata can be targeted explicitly:

```json
{
  "tags": ["source:fitness-app"],
  "tags_match": "exact"
}
```

The actual observation destination can be targeted independently, even though those values are absent from the ordinary source tags:

```json
{
  "observation_scopes": [["topic:exercise", "user:alice"]]
}
```

Both dimensions can be required:

```json
{
  "tags": ["source:fitness-app"],
  "tags_match": "all_strict",
  "observation_scopes": [["user:alice", "topic:exercise"]]
}
```

A source whose only resolved scope is `["user:alice", "topic:exercise", "tier:summary"]` is not selected by the requested two-tag scope.

### HTTP validation and task submission

`ConsolidationRequest` validates the new fields, rejects simultaneous `tags` and `tag_groups`, and forwards both dimensions to `submit_async_consolidation`. Omitting the body retains the existing full-bank behavior.

The submission path serializes `observation_scopes`, `tags`, `tags_match`, and aliased recursive `tag_groups` into the async payload. `tag_groups` is serialized whenever it is not `null`, including an explicit empty list, so the HTTP-to-worker path preserves caller input exactly. Filter assembly then applies effective-filter semantics: an empty `tag_groups` list contributes no compound predicate and is omission-equivalent, so `tags_match: "exact"` with no flat tags still selects only untagged source facts.

Unfiltered full-bank jobs retain bank-level pending-operation deduplication. Targeted requests bypass that deduplication so they cannot be merged into a pending full-bank sweep and lose their requested source set. Submission and pending-row inspection share one effective-filter predicate: non-empty `tags`, non-empty `tag_groups`, any explicit `observation_scopes`, or `tags_match: "exact"`. Empty ordinary-tag arrays therefore deduplicate as full-bank work, while exact-empty remains targeted because it selects only untagged sources. A targeted pending task cannot swallow a later full-bank sweep. Round-limit continuations carry all filters unchanged.

### Candidate selection

Ordinary tag predicates are pushed into SQL through the shared parameterized tag builders. The base candidate set remains unconsolidated, non-failed `world` and `experience` source facts in the requested bank.

When `observation_scopes` is present, the job first captures the last candidate key ordered by `(created_at, id)` using only the `id` and `created_at` projection. That key is the upper boundary of the targeted run: sources retained after the boundary are intentionally left for a later request. Candidate projection rows are then read with bounded keyset pagination ordered by the same pair. The implementation never materializes the full candidate row set or the complete eligible UUID set, so worker memory is bounded by one candidate page plus the current consolidation batch rather than growing with bank size.

The bounded candidate set is scanned once. Each page resolves scope configuration, queues only eligible IDs for the next consolidation batch, and is then discarded. Exact totals are intentionally omitted from operation progress until the keyset boundary is exhausted; this avoids doubling database transfer and scope-resolution work merely to pre-count a large or sparse bank. Every candidate projection row is converted at the database boundary into an internal Pydantic model containing `id`, `created_at`, `tags`, and `observation_scopes`. Full source rows are separately parsed into a typed `_ConsolidationMemory` model at the same boundary.

Scope resolution stays in Python because the configuration may be symbolic or explicit and must behave identically across PostgreSQL and Oracle. Reusing the dispatch resolver avoids maintaining a second interpretation of retain-time scope configuration.

### Fetching, progress, and round limits

Scope-targeted jobs fetch full source rows in bounded consolidation batches and recheck pending and failure state before processing. While the keyset scan is still discovering eligible sources, progress reports the processed count without inventing a total. Once the boundary is exhausted, the exact total is the processed count plus the bounded queue of remaining eligible IDs. If a selected candidate disappears before its full row is fetched, the job continues scanning later pages rather than terminating early.

Tag-only jobs retain the dynamic count and fetch path, including progress recounting when new facts arrive during a run. Both paths respect the configured per-round memory limit and preserve filters on continuation.

### Batch isolation

Consolidation previously grouped sources only by ordinary tag set and derived the pass plan from the first source in a group. Once ordinary tags and write scopes are independent, two facts can share identical source tags but require different scope passes.

Batches are now keyed by both the exact ordinary source tag set and the resolved write-scope set. This preserves the existing rule that different source tags never share an LLM call and prevents one source scope configuration from being applied to another fact with the same ordinary tags.

### Database portability

All source-tag predicates remain parameterized and use the shared query builders. Exact matching is represented as both containment directions. PostgreSQL supports `@>` and `<@`; the Oracle rewriter now translates the contained-by direction so exact tag set equality behaves consistently on both backends.

## Selected-source behavior

`observation_scopes` is a source-fact eligibility filter, not a per-pass suppression filter. If a source has resolved scopes A and B and the request matches A, the source is selected and runs its normal consolidation plan for A and B. This implements the requested source-oriented rule: at least one exact resolved scope admits the source, after which consolidation behaves normally.

## Compatibility boundary

The previous behavior is treated as a field-selection bug, not as a second valid meaning of `observation_scopes`. No fallback preserves tag containment under that name. A caller relying on the old behavior should move those values to `tags` and choose the appropriate `tags_match` mode or express the condition with `tag_groups`.

### Rename only the existing behavior

Adding `tags` without real observation-scope targeting would make the old behavior honest but leave no way to select sources by their actual scope configuration.

### Keep containment semantics for observation scopes

Containment would make `["user:alice"]` select a source that writes only to `["user:alice", "topic:exercise"]`. Observation scopes are isolation addresses, so subset matching can cross boundaries and conflicts with exact scope handling elsewhere.

### Resolve scopes entirely in SQL

A SQL-only predicate would require separate equivalent implementations for symbolic and explicit configurations on PostgreSQL and Oracle. Reusing the Python resolver provides one definition of scope identity and avoids cross-database semantic drift.

## Generated surfaces and documentation

The recursive tag-group union now has an explicit shared OpenAPI schema (`TagGroup-Input` / `TagGroup-Output`). Generator model-name mappings deliberately preserve the existing public Python and Go names, including `MentalModelTriggerInputTagGroupsInner`, `MentalModelTriggerOutputTagGroupsInner`, Python `ModelNot` / `Not1`, and Go `Not` / `Not1`. Recall, reflect, mental-model triggers, and the new consolidation field all continue using the legacy input name, avoiding an SDK source compatibility break while keeping the OpenAPI schema stable for future consumers.

The OpenAPI schema, bank-template JSON Schema, Go client, Python client, TypeScript types, developer documentation, operations documentation, and bundled `hindsight-docs` skill are regenerated or synchronized. The public documentation now presents ordinary source tags and observation write scopes as separate, composable selectors.

## Validation

- Unit coverage for exact scope identity, order independence, subset and superset rejection, omitted filters, empty outer lists, `[[]]`, and every symbolic scope mode.
- PostgreSQL integration coverage for all five flat tag modes, compound tag groups, differing ordinary tags and scope configurations, exact resolved-scope selection, and sources with identical ordinary tags but different scope plans.
- HTTP coverage for forwarding independent filters, rejecting simultaneous `tags` and `tag_groups`, preserving explicit empty tag groups through actual async execution, and keeping all tag-group consumers on the shared schema.
- Async operation coverage for payload serialization, worker deserialization, `null` versus `[]`, targeted deduplication behavior, and filter-preserving round-limit requeues.
- Keyset-pagination integration coverage with a two-row page size and identical timestamps, proving that the UUID tie-breaker neither skips nor repeats candidates and that each page is scanned only once.
- Effective-filter deduplication coverage proving empty `tags` and `tag_groups` share full-bank pending operations while exact-empty remains targeted.
- Oracle rewrite coverage for both directions required by exact tag equality.
- `./scripts/hooks/lint.sh`.
- `uv run ty check hindsight_api/api/http.py hindsight_api/engine/memory_engine.py hindsight_api/engine/consolidation/consolidator.py hindsight_api/engine/db/oracle.py`.
- OpenAPI, bank-template schema, generated client, documentation build, and bundled documentation skill generation.
- Generated Python and Go compatibility tests pin the legacy public tag-group imports and recursive type names.
- Consolidation suite: 124 tests passed; 13 real-LLM tests could not initialize because no local LLM API key was configured.

Closes #2852

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2650,11 +2650,35 @@ class ConsolidationRequest(BaseModel):
     observation_scopes: list[list[str]] | None = Field(
         default=None,
         description=(
-            "Optional list of tag scopes to consolidate. Each scope is a list of tags. "
-            "Only unconsolidated memories whose tags contain all tags in at least one scope "
-            "will be processed. If omitted, all unconsolidated memories are processed."
+            "Optional exact observation write scopes to consolidate. A source fact is processed when at least "
+            "one of its resolved observation scopes exactly equals at least one requested scope, regardless of "
+            "tag order. Subset and superset scopes do not match; [[]] targets the shared/untagged scope."
         ),
     )
+    tags: list[str] | None = Field(
+        default=None,
+        description="Filter source facts by their ordinary tags. Mutually exclusive with tag_groups.",
+    )
+    tags_match: TagsMatch = Field(
+        default="any",
+        description=(
+            "How to match source-fact tags: 'any', 'all', 'any_strict', 'all_strict', or 'exact'. "
+            "With 'exact' and no tags (or []), only untagged source facts match."
+        ),
+    )
+    tag_groups: list[TagGroup] | None = Field(
+        default=None,
+        description=(
+            "Compound source-fact tag filter. Each entry is a leaf {tags, match} or compound "
+            "{and: [...]}, {or: [...]}, {not: ...}. Mutually exclusive with tags."
+        ),
+    )
+
+    @model_validator(mode="after")
+    def validate_tags_exclusive(self) -> "ConsolidationRequest":
+        if self.tags is not None and self.tag_groups is not None:
+            raise ValueError("'tags' and 'tag_groups' are mutually exclusive. Use 'tag_groups' for compound filtering.")
+        return self
 
 
 class ConsolidationResponse(BaseModel):
@@ -6406,6 +6430,9 @@ def _register_routes(app: FastAPI):
                 bank_id=bank_id,
                 request_context=request_context,
                 observation_scopes=observation_scopes,
+                tags=request.tags if request else None,
+                tags_match=request.tags_match if request else "any",
+                tag_groups=request.tag_groups if request else None,
             )
             return ConsolidationResponse(
                 operation_id=result["operation_id"],
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -20,7 +20,7 @@ import json
 import logging
 import time
 import uuid
-from collections import defaultdict
+from collections import defaultdict, deque
 from contextlib import AsyncExitStack
 from dataclasses import asdict, dataclass, field
 from datetime import datetime, timezone
@@ -29,10 +29,12 @@ from itertools import combinations
 from typing import TYPE_CHECKING, Any, Literal
 
 import asyncpg
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, Field, field_validator
 
 from ...config import get_config
 from ...worker.stage import set_stage
+from ..db.base import DatabaseConnection
+from ..db.result import ResultRow
 from ..db_utils import acquire_with_retry
 from ..llm_trace import (
     record_created_memory_ids,
@@ -44,6 +46,7 @@ from ..llm_trace import (
 from ..llm_wrapper import sanitize_llm_output
 from ..memory_engine import Budget, fq_table
 from ..retain import embedding_utils
+from ..search.tags import TagGroup, TagsMatch, build_tag_groups_where_clause, build_tags_where_clause
 from .prompts import (
     build_consolidation_input,
     build_consolidation_system_prompt,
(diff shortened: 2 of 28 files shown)


nicoloboschi:
Closing this. Two concerns:

1. **Backward compatibility** — this changes what `observation_scopes` matches versus main. Today `observation_scopes` filters source facts by ordinary-tag containment (`requested ⊆ source.tags`); this PR redefines it to exact set-equality against each fact's *resolved write scopes*. For any source with extra tags or a non-default retain-time scope config, the same request would select a different (often empty) set. We don't want to silently break existing callers relying on the current behavior.

2. **Unclear demand** — the new exact-write-scope targeting is a sizeable surface (new fields, keyset scan, batch re-keying, Oracle rewrite, regenerated clients) and it's not clear anyone actually needs to target sources by resolved observation scope. The current `observation_scopes` field already covers the practical filtering use case.

If a concrete need for exact write-scope targeting comes up, we can revisit — ideally additively (a new, clearly-named field) rather than repurposing `observation_scopes`. Thanks for the thorough writeup on #2852.

Issue #1679, opened 2026-05-21
Bug: delta-mode initial refresh skips all pre-existing facts when mental model is created after consolidation


yugandhar-maram:
### Bug Description

When a mental model is created with `trigger.mode = "delta"` **after** its scoped documents have already been retained and consolidated, the
  initial async refresh finds zero facts and permanently gets stuck at "Generating content...". No subsequent refresh is ever triggered because the
   staleness check also fails.
  
  The root cause is two-fold:

  1. **`last_refreshed_at` defaults to `now()` at INSERT time.** The delta refresh sets `created_after = last_refreshed_at`, so recall only looks
  for `memory_units` with `updated_at > creation_time`. If documents were consolidated *before* the mental model was created, all their
  observations have `updated_at` earlier than the mental model's `last_refreshed_at` — recall returns nothing.
  
  2. **The placeholder "Generating content..." is treated as valid existing content.** The delta-path condition (`memory_engine.py:7599`) is `if
  refresh_mode == "delta" and current_content:` — the placeholder is truthy, so the first refresh enters delta mode instead of performing a full
  generation.
  
  Once the initial refresh completes with `no_new_facts` and preserves the placeholder, the model is permanently stuck:
  - `compute_mental_model_is_stale` returns `false` (no `memory_units` with `updated_at > last_refreshed_at` in scope)
  - Post-consolidation triggers only fire for models whose tags overlap with `consolidated_tags` from that specific run — if the relevant scope was
   consolidated in an *earlier* run, the model is never selected

### Steps to Reproduce

```bash
  # 1. Create a bank
  curl -X POST "http://127.0.0.1:9999/v1/default/banks" \
    -H "Content-Type: application/json" \
    -d '{"bank_id": "repro-test"}'

  # 2. Retain a document (triggers consolidation)
  curl -X POST "http://127.0.0.1:9999/v1/default/banks/repro-test/files/retain" \
    -H "Content-Type: application/json" \
    -d '{"content": "# API Design\nThe service uses REST with JSON payloads.\nRate limit: 1000 req/s.", "tags": ["scope:backend"]}'

  # 3. Wait for consolidation to complete (~10-20s)
  sleep 20

  # 4. Create a mental model scoped to the same tags, with delta mode
  curl -X POST "http://127.0.0.1:9999/v1/default/banks/repro-test/mental-models" \
    -H "Content-Type: application/json" \
    -d '{
      "name": "backend_overview",
[...]

PR #2866, opened 2026-07-21
fix: advance watermark after no-op delta refresh

SunneeYang:
## Summary

Advance a delta mental model's refresh watermark when reflection completes successfully but finds no topic-relevant facts, while preserving facts that arrive during the refresh for a later run.

Without this, the scheduled-refresh maintenance loop sees the same model as due and stale on every tick, repeatedly invoking the Reflect LLM against the same memory window.

## Root cause

The `no_new_facts` path persists `reflect_response` and `last_refreshed_source_query` without passing `content`. `update_mental_model()` previously advanced `last_refreshed_at` only when content was rewritten, so a successful no-op refresh left the scheduler watermark unchanged.

This can happen when the coarse staleness query sees a new memory row but Reflect correctly determines that the row is unrelated to the mental model's source query.

A naive post-reflect `NOW()` update would stop the loop but introduce a data-loss race: a fact arriving after recall but before the watermark write could fall behind the new watermark without ever being considered.

## Fix

- Capture a database-time `refresh_cutoff` before Reflect starts.
- Pass that cutoff as Reflect's `created_before` upper bound.
- Persist the exact cutoff as `last_refreshed_at` after a successful refresh, including the `no_new_facts` path.
- Preserve the existing content and embedding, and do not create a history snapshot for a no-op refresh.
- Keep facts arriving after the cutoff newer than the watermark so a later refresh can still process them.

## Regression tests

The no-op regression test reproduces the scheduler loop:

1. Create an established cron-scheduled delta mental model.
2. Make its refresh watermark old and add a fresh but unrelated memory.
3. Verify the model is stale before refresh.
4. Mock Reflect to return no supporting facts.
5. Verify the refresh advances the watermark, preserves content, performs no embedding generation, creates no history entry, clears staleness, and prevents the next maintenance pass from submitting another refresh.

A second deterministic race test inserts a fact after Reflect has returned but before `update_mental_model()` writes the watermark. It verifies that:

- Reflect received `created_before=refresh_cutoff`;
- `last_refreshed_at` remains earlier than the late fact's `updated_at`;
- the mental model remains stale so the late fact will be reconsidered.

Both tests fail before the fix.

## Impact observed

On Hindsight 0.8.4, one affected scheduled delta model produced 1,408 refresh operations and 1,398 actual LLM requests (2,110,981 tokens) in 24 hours before its cron was temporarily disabled. The configured cron itself was daily; the repeated requests came from the maintenance retry loop.

## Testing

- `uv run pytest tests/test_mental_model_delta.py tests/test_mental_model_scheduled_refresh.py -q -n 0`
  - 17 passed, 4 skipped (real-provider evaluation tests)
- `./scripts/hooks/lint.sh`
  - all lints passed
- Targeted Ruff formatting/checks and `ty` type checking
  - all checks passed
- `uv run pytest tests/test_recall_time_range.py tests/test_temporal_recall_selection.py -q -n 0`
  - 13 passed

## Related work

- #1192 introduced the delta `no_new_facts` short-circuit.
- #2377 introduced scheduled mental model refreshes.
- #1679 fixed a related first-refresh placeholder case, but does not cover established models repeatedly scheduled after a successful no-op delta refresh.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -10837,6 +10837,21 @@ class MemoryEngine(MemoryEngineInterface):
 
             tag_filtering = _resolve_refresh_tag_filtering(mental_model.get("tags"), trigger_data)
 
+            # Bound this refresh to a database-time snapshot. Facts arriving while
+            # reflect is running must remain newer than the persisted watermark so
+            # a later refresh can still see them.
+            backend = await self._get_backend()
+            assert self._dialect is not None
+            async with acquire_with_retry(backend) as conn:
+                refresh_cutoff = await conn.fetchval(
+                    f"SELECT {self._dialect.current_timestamp()} "
+                    f"FROM {fq_table('mental_models')} WHERE bank_id = $1 AND id = $2",
+                    bank_id,
+                    mental_model_id,
+                )
+            if refresh_cutoff is None:
+                return None
+
             # Run reflect with the source query, excluding the mental model being refreshed
             # Skip creating a nested "hindsight.reflect" span since we already have "hindsight.mental_model_refresh"
             # Build context to guide the reflect agent: tell it what this mental
@@ -10873,6 +10888,7 @@ class MemoryEngine(MemoryEngineInterface):
                 # Attribute these LLM calls to the mental-model refresh, not a
                 # plain reflect, so traces group under the right operation.
                 _operation_label="refresh_mental_model",
+                created_before=refresh_cutoff,
             )
             # Forward the per-model max_tokens so the final synthesis is capped at the
             # user-configured limit rather than the reflect_async default.
@@ -11005,6 +11021,7 @@ class MemoryEngine(MemoryEngineInterface):
                             mental_model_id,
                             reflect_response=reflect_response_payload,
                             last_refreshed_source_query=current_source_query,
+                            refresh_watermark=refresh_cutoff,
                             request_context=request_context,
                         )
 
@@ -11112,6 +11129,7 @@ class MemoryEngine(MemoryEngineInterface):
                 content=final_content,
                 reflect_response=reflect_response_payload,
                 last_refreshed_source_query=current_source_query,
+                refresh_watermark=refresh_cutoff,
                 structured_content=(final_structured.model_dump() if final_structured is not None else None),
                 request_context=request_context,
             )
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Reviewed the full diff against the surrounding code. Accurate diagnosis and a minimal, correct fix.

- Confirmed staleness is `updated_at > last_refreshed_at` and the scheduler gates resubmission on both cron due-ness and staleness, both keyed off `last_refreshed_at`.
- The delta `no_new_facts` short-circuit previously left the watermark unadvanced, causing the observed refresh loop. Adding `mark_refreshed` to advance `last_refreshed_at = NOW()` clears both gates.
- Explicit flag (vs. advancing on any `reflect_response`) is the right call since `update_mental_model` also handles user edits that must not bump the watermark.
- Regression test reproduces the exact state transition and locks it in.

Non-blocking nit: the `mark_refreshed` docstring mentions structured-content but the `elif` only guards on `content`; harmless given the sole caller. LGTM.

SunneeYang:
Post-merge follow-up: an independent concurrency review found a blocking transaction-visibility race in the merged cutoff design. Database wall-clock time does not order transaction visibility or commit order.

I reproduced this deterministically on PostgreSQL: a retain transaction starts and assigns `updated_at` before `refresh_cutoff`, remains uncommitted while Reflect runs (so recall cannot see it), then commits before the model update. The merged code stores `last_refreshed_at = refresh_cutoff`; the newly visible row now has `updated_at < last_refreshed_at` and is permanently excluded by the later strict `updated_at > last_refreshed_at` stale/recall checks. A strict upper/lower boundary also leaves equality uncovered.

I am preparing a follow-up fix that does not advance the fact watermark on `no_new_facts`. Instead it stores a fixed-size token of the exact pre-Reflect visible scoped row versions; unchanged state suppresses repeat LLM refreshes, while any later-visible transaction remains stale. The regression test starts an uncommitted transaction before refresh and commits it after Reflect.

Please avoid releasing/cherry-picking #2866 as-is until the follow-up lands (or revert it in the meantime). Sorry this arrived just after merge; the earlier review text appears to describe the pre-force-push `NOW()` implementation rather than the final cutoff diff.

SunneeYang:
Follow-up fix is now available as draft PR #2868: https://github.com/vectorize-io/hindsight/pull/2868

It is based on current `main` after #2866 and includes the deterministic in-flight transaction regression test.

PR #2868, opened 2026-07-21
Fix unsafe delta refresh visibility watermark

SunneeYang:
## Summary

Follow-up to #2866, which was merged with a wall-clock cutoff that does not safely order transaction visibility.

A transaction may receive an `updated_at` before the refresh snapshot but remain uncommitted until after Reflect finishes. Advancing a delta model to either a pre-Reflect cutoff or a completion-time `NOW()` can then permanently hide that fact.

This PR replaces delta-refresh timestamp advancement with an exact visible-state acknowledgement:

1. before Reflect, hash the complete currently visible scoped row state newer than the existing semantic watermark;
2. run Reflect and apply the delta normally;
3. for every successful delta result — contentful or no-op — preserve `last_refreshed_at` and store the pre-Reflect visible-state token;
4. recompute that token for staleness checks; equal means fresh, any later-visible insert/update/delete/effective-input change means stale.

This deliberately favors safe reprocessing over permanent omission. A future database-native commit-order revision could compact/advance the semantic cursor, but a wall-clock timestamp cannot provide that guarantee across PostgreSQL and Oracle.

## What changed

- Removed the `refresh_cutoff` / `created_before` design from delta refreshes.
- Removed `last_refreshed_at` advancement from every successful delta path, including content-changing updates.
- Preserved normal content, embedding, structured-content, provenance, and history updates for contentful deltas.
- Stored `delta_scope_state` in `reflect_response` for both contentful and no-op delta refreshes.
- Added an internal update override so full refreshes and direct user content updates retain existing timestamp behavior.
- Added token-aware staleness to every caller:
  - consolidation
  - search/recall mental-model inclusion
  - scheduled maintenance
- Lightweight callers self-fetch the token/config columns before deciding staleness.
- Propagated `RequestContext` through all staleness callers so tenant-resolved config and permission filtering match actual Reflect behavior.
- Reused an already-held database connection when resolving bank config, avoiding a nested-acquire deadlock when the pool maximum is one.
- Added a per-cron-fire in-process checkpoint and bounded it by pruning deleted/inactive model keys before tenant discovery, including empty and no-due cycles.

## Token v3

The token uses frozen, fixed-schema dataclasses internally and covers:

- mental-model name, source query, mode, and max tokens;
- effective hierarchical `recall_include_chunks`;
- effective `recall_max_tokens` and `recall_chunks_max_tokens`;
- effective `reflect_source_facts_max_tokens`;
- effective `store_document_text`;
- mental-model exclusion options;
- resolved tags, tag matching, tag groups, and fact types;
- canonical visible row ID, UTC-microsecond timestamp, text, fact type, and normalized tags.

Legacy v2 or malformed tokens are conservatively considered stale once, causing a safe refresh and migration to v3 rather than trusting an incomplete acknowledgement. The v3 boundary requires exactly the fixed four-field schema, exact integer types, a non-negative count, canonical UTC-microsecond timestamp, and a 64-character lowercase SHA-256 digest; values are never coerced.

## Why it is safe

The token is computed from committed rows visible to the refresh connection before Reflect.

- Same visible state after refresh: token matches, model is fresh.
- A transaction commits after the snapshot: current token differs, model remains stale.
- A contentful refresh finishes after such a commit: its timestamp is not advanced past the unseen row.
- Two contentful refreshes finish in reverse order: an older worker may temporarily overwrite newer content, but its older token no longer matches current state, so a retry signal remains.
- Rows are inserted, updated, deleted, or move into/out of scope: token differs.
- Effective hierarchical refresh configuration or model prompt inputs change: token differs.

The scoped hash query runs only for due scheduled models, and the per-fire checkpoint prevents re-running it every maintenance tick.

## Deterministic concurrency coverage

Two PostgreSQL transaction tests exercise the invisible-row interleaving:

1. A transaction inserts a relevant memory without committing.
2. Refresh captures its visible-state token and completes Reflect.
3. The transaction commits before the mental-model update.
4. Both no-op and contentful refresh paths preserve the old watermark.
5. Stored token count is `0`, current token count is `1`, and the model remains stale.

A third test runs two contentful refreshes concurrently and forces reverse completion:

1. older worker captures state S0 and blocks;
2. a new fact commits;
3. newer worker captures S1 and completes first;
4. older worker completes last with S0;
5. final state is still stale because S0 != S1.

Additional tests cover content/fact-type/tag changes without timestamp changes, deletion, effective-config invalidation, v2 and malformed-v3 migration, lightweight callers, bounded same-fire maintenance across empty/no-due/tenant-discovery-failure paths, and proving scope-state config resolution does not perform a nested database acquire.

## Validation

- delta suite: `13 passed, 4 skipped`
- scheduled-refresh suite: `6 passed`
- hierarchical ConfigResolver regression: `27 passed`
- wider mental-model suite: `60 passed`; 8 real-LLM fixtures stop during setup because no Groq API key is configured and do not reach product assertions
- Ruff full package: passed
- Ruff format: passed
- `ty check hindsight_api`: passed
- package sdist/wheel build: passed
- `git diff --check`: passed
- repository commit hooks (`check-unused`, generated docs skill, lint): passed
- independent read-only review: no remaining code blocker; malformed-v3 and checkpoint coverage findings addressed

## Operational note

Affected scheduled models remain disabled with `refresh_cron: null` until this follow-up is merged and deployed.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config_resolver.py
+++ b/hindsight-api-slim/hindsight_api/config_resolver.py
@@ -98,7 +98,12 @@ class ConfigResolver:
 
         return config_dict
 
-    async def resolve_full_config(self, bank_id: str, context: RequestContext | None = None) -> HindsightConfig:
+    async def resolve_full_config(
+        self,
+        bank_id: str,
+        context: RequestContext | None = None,
+        conn: Any | None = None,
+    ) -> HindsightConfig:
         """
         Resolve full HindsightConfig for a bank with hierarchical overrides applied.
 
@@ -113,6 +118,7 @@ class ConfigResolver:
         Args:
             bank_id: Bank identifier
             context: Request context for tenant config resolution
+            conn: Existing database connection to reuse when one is already held
 
         Returns:
             Complete HindsightConfig with hierarchical overrides applied
@@ -120,7 +126,7 @@ class ConfigResolver:
         config_dict = await self._resolve_parent_config_dict(bank_id, context)
 
         # Load bank config overrides
-        bank_overrides = await self._load_bank_config(bank_id)
+        bank_overrides = await self._load_bank_config(bank_id, conn)
         if bank_overrides:
             config_dict.update(bank_overrides)
             logger.debug(f"Applied bank config overrides for bank {bank_id}: {list(bank_overrides.keys())}")
@@ -149,7 +155,12 @@ class ConfigResolver:
         )
         return resolved_config
 
-    async def get_bank_config(self, bank_id: str, context: RequestContext | None = None) -> dict[str, Any]:
+    async def get_bank_config(
+        self,
+        bank_id: str,
+        context: RequestContext | None = None,
+        conn: Any | None = None,
+    ) -> dict[str, Any]:
         """
         Get fully resolved config for a bank (filtered by permissions).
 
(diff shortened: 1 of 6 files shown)


nicoloboschi:
superseded by 2878

Issue #2547, opened 2026-07-03
retain: lenient-parse path persists degenerate field content — repeated-sentence loops and leaked response JSON accumulate in `what`/`why`, producing multi-KB "facts"


iRonin:
### Summary
Fact extraction's lenient parse (raw JSON → `get_value()` → `combined_parts` assembly, `retain/fact_extraction.py:1423-1434` sync and `:2184-2191` batch) persists field content **verbatim** — there is no length bound, repetition guard, or leak check anywhere on the fact-write path. When the extraction model degenerates (common with fast/cheap lax-output models, observed: `gpt-oss-120b`), the degeneration is stored permanently as fact text, then embedded and BM25-indexed in that form.

### Observed in production (150,914-fact corpus, concise mode)
568 facts affected (max **46,834 chars** for a single fact). The structure is consistent and telling:
- the `what` stays short and clean (corpus: avg 167, p95 316, p99 433 chars) — the fact itself extracted correctly;
- the entire bloat is in the `why` (legit `why` is a phrase: p50=54, p95=117; the affected tail runs to 28,813 chars), in three shapes:
  1. **repeated-sentence/segment loops** — "The fact is a concise summary of the core event." ×hundreds, "N/A if obvious. " ×hundreds, comma-list repetition ("no cost indefinite, " ×hundreds);
  2. **self-deliberation leaks** — the model arguing with itself about the field ("N/A"? Actually why is given: … We'll set why=…");
  3. **leaked response JSON** — the model dumps the *rest of its facts array* into the `why` string (` ```json [{"fact_kind": …`, multi-space `{       "fact_kind"`, markdown-bold `**fact_kind**: … **fact_type**: …` forms all observed).

Consequences: multi-KB facts bloat embeddings (vector computed on garbage → broad low-precision matches), the stored `tsvector` matches nearly any keyword query (`ts_rank_cd` is not length-normalized by default → O(bank) attractors in the BM25 arm), and leaked-JSON tails can contain *other facts* as raw text.

### Why the retry ladder doesn't help
The response is **valid JSON with valid schema** — the degeneration is inside string values. Every parse succeeds; nothing fails loudly; the corpus degrades silently.

### Proposed fix
Field-level de-bloat at the lenient parse (both paths), applied to `what`/`why` after `get_value()`:
[...]

Issue #2544, opened 2026-07-03
consolidation persists degenerate LLM output into observation text — repetition loops, leaked structured-output syntax, unsanitized merge path (lax-JSON models)


iRonin:
### Summary
Consolidation persists observation text from the LLM through three pydantic models in `engine/consolidation/consolidator.py`:

- `_CreateAction` (:450) and `_UpdateAction` (:470) run a `text` field_validator that calls `sanitize_llm_output` — but that is just a back-compat **alias for `sanitize_text`** (`engine/llm_wrapper.py:164`), a character-safety scrub introduced for #1875 (one regex, :159: ASCII control chars + Unicode surrogates; explicitly preserves `\t` `\n` `\r`). It does nothing about repetition/degeneration loops, unbounded length, or leaked structured-output scaffolding.
- `_DedupDecision` (:98) — the merge path — has **no validator on `text` at all**: the merge decision becomes `merged_text = decision.text.strip() or best_text` (:199) and is written directly via SQL (:241, :306), **bypassing the create/update validators entirely**. Zero sanitization, not even the character scrub.

When the consolidation model degenerates (repetition loops, leaked JSON syntax, editorial injection — common with cheap/fast lax-JSON models), the garbage is written as a permanent observation. It then pollutes every future recall/reflect for that bank and inflates embedding cost, silently and cumulatively.

### Impact (observed in production)
`ghcr.io/vectorize-io/hindsight:0.8.4`, consolidation LLM = `openrouter/openai/gpt-oss-120b` (litellmrouter). Corpus scan: **134 / 106,811 observations polluted across 12 banks**, three classes:

- **71 degeneration loops** — repetition runs up to **29.4 KB** of repeated chars/newlines in a single observation; content destroyed, embeddings + recall bloated. Scan: `length(text) > 2000 OR text ~ '(.)\1{9,}'`.
- **17 leaked JSON fragments** — the model's own structured-output scaffolding (`","text":"`, `","fact_type":"`) embedded in observation text, e.g. `...checks.)"'""\""'\""<uuid>"fl bfl" j;n.","text":"This provides n...`. Scan: `text ~ '","[a-z_]+":"'`.
- **86 editorial injections** — `(Note: This is a distinct verification...)`-style meta-commentary embedded mid-text (avg position ~43% into the observation).

### Proposed fix
[...]

Issue #2504, opened 2026-07-02
codex_llm: strict_schema is a dead no-op — structured output still uses prompt-injection + raw json.loads, causing JSONDecodeError retry storms (same class as #1002)


morozsm:
## Summary

`hindsight_api/engine/providers/codex_llm.py` is the last structured-output provider still using the prompt-injected-schema + raw `json.loads` + retry pattern that #1002 identified as a retry-storm root cause. On content with backslashes (code, serial/CLI commands, Windows paths, regexes) the model emits invalid `\escape` sequences, every parse attempt fails, and retain/consolidation tasks burn all retries and fail.

## Production evidence (0.8.3-slim and 0.8.4-slim, Codex OAuth provider)

```
WARNING - hindsight_api.engine.providers.codex_llm - Codex JSON parse error (attempt 1/4): Invalid \escape: line 1 column 1361 (char 1360)
...
WARNING - hindsight_api.engine.providers.codex_llm - Codex JSON parse error (attempt 4/4): Invalid \escape: line 1 column 2243 (char 2242)
ERROR - hindsight_api.engine.memory_engine - Task execution failed: batch_retain, error: Fact extraction failed: 1/1 chunks failed. First failures: chunk 0: JSONDecodeError: Invalid \escape
```

The failing bank stores coding-session transcripts (rig-control serial commands, escape-heavy content) — extraction for it fails deterministically, every retry re-fails the same way.

## Root cause (current `main`)

In `hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py`:

- `call()` accepts `strict_schema: bool = False` (line ~336) but the parameter is **never referenced in the method body** — a dead no-op.
- Schema is prompt-injected (lines ~364-368): `"You must respond with valid JSON matching this schema:\n{json.dumps(schema, ...)}"`.
- Parsing is markdown-fence stripping + raw `json.loads` (lines ~430-438); `JSONDecodeError` → exponential-backoff **re-request** (no repair, no tool_use), then re-raise (lines ~439-446).

Compare:

- #2339 fixed the same problem for Anthropic by routing `strict_schema=True` through **forced tool_use** (`tool_choice={"type":"tool","name":"structured_response"}`), closing #1002 (338 successes vs 331 invalid-JSON retries; one call burning 759s over 11 consecutive failures).
- #2003 added the strict json_schema path to `gemini_llm.py` and `openai_compatible_llm.py`.
- `codex_llm.py` was touched by neither.

[...]

PR #2871, opened 2026-07-21
fix(llm): recover malformed JSON via json_repair (last-resort parse fallback)

nicoloboschi:
## What

Adds `json_repair` as a **last-resort structural repair** for malformed LLM JSON in `parse_llm_json`, so a response that `json.loads` cannot parse at all — trailing commas, unterminated strings, single quotes, invalid `\escape` sequences (the codex #2504 class) — is recovered instead of dropped (losing every fact/observation in it).

Follows the discussion on #2547 / #2544: `json_repair` was the suggested approach for the *recoverable* half of the problem.

## Scope — deliberately narrow

This is scoped to **structural malformation only** (JSON that fails to parse). It does **not** touch content semantics.

The other failure class in #2544/#2547 — degeneration *inside valid JSON* (repetition loops, self-deliberation, leaked structured-output scaffolding stuffed into string values) — is explicitly **out of scope**. That JSON parses fine; regex-editing the model's output to "fix" the meaning is a whack-a-mole with real false-positive risk. The honest fix there is switching off the degenerate lax-JSON model. So this PR does *not* implement the content-sanitizer proposed on those issues.

## Behavior

- `parse_llm_json`: after fence-stripping and the control-char scrub both fail, attempt `repair_json(..., return_objects=True)`.
- **Guardrail (empty → raise):** `repair_json` never raises; unrecoverable input yields an empty result (`""` / `{}` / `[]`). In that case we keep raising `JSONDecodeError` rather than let an empty object masquerade as a successful parse — so retry ladders and the #1833 fail-loud path still fire.
- **LiteLLM provider** (covers the `litellmrouter` env in the reports; `LiteLLMRouterLLM` inherits it): prefers a **clean re-roll first** — repair runs only once the retry budget is exhausted. The central `parse_llm_json` fallback already covers the Gemini, reflect-delta, and batch fact-extraction call sites.

## Tests

- `test_llm_wrapper.py`: repair matrix (trailing comma, unterminated string, single quotes, invalid escape, fenced) + control-char scrub still works + unrecoverable-raises.
- `test_litellm_json_repair.py`: retry-then-repair, clean-reroll-wins-over-repair, unrecoverable-still-raises-after-retries.

## Notes

- `uv.lock` updated by hand (adds only the `json-repair` leaf package) because the local `uv` is older than the one that wrote the lock; a full `uv sync` would have reformatted the entire file.


nicoloboschi:
## `/code-review` — clean ✅

Ran the project code-review checklist against the branch.

- **Branch hygiene** — one relevant commit, based on current `origin/main` (0 behind), all 6 changed files in scope.
- **Lint** — `ruff check` + `ruff format --check` + `ty check` all pass on the changed files. (The `lint.sh` harness itself errored on an unrelated `hindsight_embed` dist-info uninstall glitch in the shared venv — not a code issue; ran the linters directly instead.)
- **Dead code** — none; the new `parse_llm_json` import in `litellm_llm.py` is used.
- **Type safety** — no raw structured dicts (the parsed JSON is a genuinely dynamic blob — the allowed dict case), no multi-item tuple returns, no missing hints.
- **Tests** — utility + provider paths both covered (12 passing). This is deterministic JSON-parsing mechanics, not an LLM-behaviour/prompt change, so no real-LLM judge test is warranted.
- **Other** — no advisory locks, no root `.py` files, no new config flags/env template, no API/MCP/migration/integration surface touched.

No must-fix or should-fix findings.

PR #2875, opened 2026-07-21
feat(config): make embedding thresholds configurable

### LINKED ISSUE #2874: Make embedding-dependent similarity thresholds configurable across retrieval and graph construction
## Motivation

Cosine-similarity distributions are not calibrated across embedding models. Two models can embed the same text pairs into noticeably different similarity ranges, even when both return L2-normalized vectors. Thresholds tuned for the current default `BAAI/bge-small-en-v1.5` can therefore produce poor recall, overly dense semantic graphs, sparse graphs, or incorrect deduplication behavior when another embedding model/provider is used.

Hindsight currently has five embedding-dependent similarity gates, but only two are configurable.

## Current state

| Purpose | Current value | Current configuration support |
| --- | ---: | --- |
| Main semantic recall | `0.3` | Configurable through `HINDSIGHT_API_SEMANTIC_MIN_SIMILARITY`; can also be overridden per request with `min_scores.semantic` |
| Graph retrieval semantic seeds | `0.3` | Hardcoded in `LinkExpansionRetriever` / `_find_semantic_seeds` |
| Temporal retrieval semantic gate | `0.1` | Hardcoded when calling `retrieve_temporal_combined` and used for both entry points and spread neighbors |
| Semantic-link graph construction | `0.7` | Hardcoded in ANN and within-batch semantic-link creation; also used by graph-maintenance relinking |
| Observation semantic deduplication | `0.97` | Configurable through `HINDSIGHT_API_CONSOLIDATION_DEDUP_THRESHOLD` |

The existing `semantic_min_similarity` setting only controls the main semantic retrieval arm. It does not affect graph seed selection or temporal retrieval. Likewise, changing the embedding model does not provide a way to recalibrate the `0.7` semantic-link construction threshold.

## Request

Please make all five embedding-dependent thresholds explicit configuration fields while preserving the existing defaults for backward compatibility.

The two existing settings can remain as they are:

- `semantic_min_similarity` / `HINDSIGHT_API_SEMANTIC_MIN_SIMILARITY` (`0.3`)
- `consolidation_dedup_threshold` / `HINDSIGHT_API_CONSOLIDATION_DEDUP_THRESHOLD` (`0.97`)

Add independent settings for the three currently hardcoded gates, for example:

- `graph_seed_min_similarity` (`0.3`)
- `temporal_semantic_min_similarity` (`0.1`)
- `semantic_link_min_similarity` (`0.7`)

The resolved values should be passed consistently through all relevant paths, including normal retain, streaming retain, graph maintenance, graph retrieval, and temporal retrieval. Configuration validation, the environment template, and developer documentation should describe that these values require recalibration when changing embedding models.

These should remain separate settings rather than one universal threshold because they serve different precision/recall tradeoffs: candidate retrieval, graph expansion seeding, temporal gating, graph construction, and near-duplicate detection.

Sanderhoff-alt:
## Summary

This PR completes configuration coverage for Hindsight's five embedding-dependent similarity thresholds by adding settings for the three previously hardcoded gates while preserving the two existing settings and all current defaults calibrated for `BAAI/bge-small-en-v1.5`.

Cosine-similarity distributions differ across embedding models even when their output vectors are L2-normalized. Reusing thresholds calibrated for one model can reduce recall, admit irrelevant candidates, create an overly dense or sparse semantic graph, or change observation deduplication behavior. Keeping these gates separate allows each task to be recalibrated for a different embedding model without coupling unrelated precision/recall tradeoffs.

## Configuration behavior and scope

| Purpose | Configuration | Default | Status | Behavior and scope |
| --- | --- | ---: | --- | --- |
| Main semantic recall | `semantic_min_similarity` / `HINDSIGHT_API_SEMANTIC_MIN_SIMILARITY` | `0.3` | Existing | Controls only the main semantic retrieval arm. `min_scores.semantic` can override it for a single recall request. It does not control graph seed selection or temporal retrieval. Its existing `[0, 1]` validation and default remain unchanged. |
| Graph retrieval semantic seed | `graph_seed_min_similarity` / `HINDSIGHT_API_GRAPH_SEED_MIN_SIMILARITY` | `0.3` | Added | Controls which memories can seed Link Expansion graph retrieval. It is independent of the main semantic recall threshold and its request-level override. Valid range is `[0, 1]`. |
| Temporal retrieval semantic gate | `temporal_semantic_min_similarity` / `HINDSIGHT_API_TEMPORAL_SEMANTIC_MIN_SIMILARITY` | `0.1` | Added | Controls the semantic gate used by temporal retrieval for both entry points and spread neighbors. It is independent of the main semantic recall threshold and its request-level override. Valid range is `[0, 1]`. |
| Semantic-link graph construction | `semantic_link_min_similarity` / `HINDSIGHT_API_SEMANTIC_LINK_MIN_SIMILARITY` | `0.7` | Added | Controls semantic links created by normal retain ANN lookup, within-batch linking, streaming retain's final ANN pass, and graph-maintenance relinking. Valid range is `[0, 1]`. The setting applies only when links are created; it does not delete old links or rebuild an existing graph. |
| Observation semantic deduplication | `consolidation_dedup_threshold` / `HINDSIGHT_API_CONSOLIDATION_DEDUP_THRESHOLD` | `0.97` | Existing | Controls semantic candidate generation for observation consolidation. Values below `1.0` are active thresholds; values greater than or equal to `1.0` retain the existing disable behavior. There is no per-request override, and this PR does not add `[0, 1]` validation to this setting. |

## What changed

- Added the three new fields, defaults, environment parsing, and `[0, 1]` validation to `HindsightConfig`.
- Passed the resolved graph seed threshold through Link Expansion retrieval.
- Passed the resolved temporal threshold through temporal entry-point and neighbor selection.
- Passed the resolved semantic-link threshold through normal retain, within-batch link creation, streaming retain, and graph maintenance.
- Preserved all five existing default values, so deployments that do not set the new environment variables keep their previous behavior.
- Updated both environment templates, developer configuration documentation, and the generated documentation skill mirror.
- Documented that changing the semantic-link threshold is not retroactive: raising it does not remove existing low-weight links, and lowering it does not recompute nodes that graph maintenance already considers sufficiently linked. A full transition requires rebuilding the graph or re-ingesting source data into a new memory bank.
- Added regression coverage for defaults, environment overrides, validation, graph seed wiring, temporal wiring, retain and streaming paths, and graph-maintenance relinking.

## Compatibility

This is backward compatible when the new variables are unset: graph seeds remain at `0.3`, the temporal semantic gate remains at `0.1`, and semantic-link construction remains at `0.7`. The existing main semantic recall default remains `0.3`, and observation semantic deduplication remains `0.97` with its existing `>=1.0` disable semantics.

## Validation

- Targeted API test suites: 119 passed.
- Configuration validation suite: 81 passed.
- Environment template tests: 7 passed.
- `./scripts/hooks/lint.sh`: passed.
- `uv run ty check hindsight_api/`: passed.

Closes #2874.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -207,6 +207,15 @@ HINDSIGHT_API_LOG_LEVEL=info
 # DeepSeek note: DeepSeek is supported for LLM calls, but not for embeddings.
 # If using DeepSeek as LLM provider, keep embeddings on local/openai/cohere/google/etc.
 
+# Embedding similarity thresholds. These defaults preserve the behavior calibrated
+# for BAAI/bge-small-en-v1.5. Recalibrate each threshold independently when changing
+# embedding models because cosine-similarity distributions are model-dependent.
+# HINDSIGHT_API_SEMANTIC_MIN_SIMILARITY=0.3
+# HINDSIGHT_API_GRAPH_SEED_MIN_SIMILARITY=0.3
+# HINDSIGHT_API_TEMPORAL_SEMANTIC_MIN_SIMILARITY=0.1
+# HINDSIGHT_API_SEMANTIC_LINK_MIN_SIMILARITY=0.7
+# HINDSIGHT_API_CONSOLIDATION_DEDUP_THRESHOLD=0.97
+
 # Reranker Configuration (Optional - uses local by default)
 # Provider: "local" (default) or "tei" (HuggingFace Text Embeddings Inference)
 # HINDSIGHT_API_RERANKER_PROVIDER=local
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -429,6 +429,9 @@ ENV_RERANKER_LITELLM_SDK_TIMEOUT = "HINDSIGHT_API_RERANKER_LITELLM_SDK_TIMEOUT"
 ENV_RERANKER_GOOGLE_TIMEOUT = "HINDSIGHT_API_RERANKER_GOOGLE_TIMEOUT"
 ENV_RERANKER_MAX_CANDIDATES = "HINDSIGHT_API_RERANKER_MAX_CANDIDATES"
 ENV_SEMANTIC_MIN_SIMILARITY = "HINDSIGHT_API_SEMANTIC_MIN_SIMILARITY"
+ENV_GRAPH_SEED_MIN_SIMILARITY = "HINDSIGHT_API_GRAPH_SEED_MIN_SIMILARITY"
+ENV_TEMPORAL_SEMANTIC_MIN_SIMILARITY = "HINDSIGHT_API_TEMPORAL_SEMANTIC_MIN_SIMILARITY"
+ENV_SEMANTIC_LINK_MIN_SIMILARITY = "HINDSIGHT_API_SEMANTIC_LINK_MIN_SIMILARITY"
 ENV_RERANKER_FLASHRANK_MODEL = "HINDSIGHT_API_RERANKER_FLASHRANK_MODEL"
 ENV_RERANKER_FLASHRANK_CACHE_DIR = "HINDSIGHT_API_RERANKER_FLASHRANK_CACHE_DIR"
 ENV_RERANKER_FLASHRANK_CPU_MEM_ARENA = "HINDSIGHT_API_RERANKER_FLASHRANK_CPU_MEM_ARENA"
@@ -828,6 +831,9 @@ DEFAULT_RERANKER_LITELLM_SDK_TIMEOUT = 60.0
 DEFAULT_RERANKER_GOOGLE_TIMEOUT = 60.0
 DEFAULT_RERANKER_MAX_CANDIDATES = 300
 DEFAULT_SEMANTIC_MIN_SIMILARITY = 0.3
+DEFAULT_GRAPH_SEED_MIN_SIMILARITY = 0.3
+DEFAULT_TEMPORAL_SEMANTIC_MIN_SIMILARITY = 0.1
+DEFAULT_SEMANTIC_LINK_MIN_SIMILARITY = 0.7
 # Minimum BM25 score a row must exceed to enter fusion. 0.0 gates out
 # zero-score (non-matching) rows on backends — notably VectorChord — whose
 # operator ranks every document rather than pre-filtering to term matches.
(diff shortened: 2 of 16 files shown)


nicoloboschi (approved):
Approving. Verified the wiring end-to-end on the branch:

- Grepped every caller of `compute_semantic_links_ann`, `compute_semantic_links_within_batch`, `create_semantic_links_batch`, `_find_semantic_seeds` and `retrieve_temporal_combined` — all pass the resolved value, and no hardcoded 0.3/0.7/0.1 gate survives in `engine/search/` or `link_utils.py`. The threshold does reach the within-batch numpy path (`link_utils.py:750`).
- Confirmed the new tests are non-vacuous: the monkeypatch targets match the modules actually resolved at call time (`graph_maintenance` imports `compute_semantic_links_ann` at module level and `acquire_with_retry` lazily from `.memory_engine`; `orchestrator` uses a function-local import of `link_utils`).
- `config.semantic_link_min_similarity` is safe in retain — `config` there is a full `HindsightConfig` from `resolve_full_config`.
- Leaving the fields out of `_CONFIGURABLE_FIELDS` is the right call: consistent with `semantic_min_similarity` / `consolidation_dedup_threshold`, and these must track the embedding model, which is itself server-level static.
- Ran locally: `test_config_validation.py`, `test_link_expansion_config.py`, `test_retain_orchestrator_mapping.py`, `test_temporal_recall_selection.py` and the new graph-maintenance test — 106 passed. `ruff check` clean.

Defaults are byte-identical to the previous hardcoded values, so unset deployments are unchanged, and the non-retroactivity of `SEMANTIC_LINK_MIN_SIMILARITY` is documented.

Non-blocking follow-ups:
1. `compute_semantic_links_ann` / `compute_semantic_links_within_batch` / `link_utils.create_semantic_links_batch` keep `threshold=DEFAULT_SEMANTIC_LINK_MIN_SIMILARITY` defaults. Config is honored only because every call site passes it explicitly; a future caller that forgets silently reverts to 0.7. Consider making it required/keyword-only there too, as `link_creation` already does.
2. `_run_final_semantic_ann(pool, bank_id, unit_ids, threshold, log_buffer)` — a bare float between positionals is easy to transpose; `*, threshold: float` would be cheap insurance.
3. `test_link_creation_passes_threshold_to_within_batch_path` mocks `link_utils.create_semantic_links_batch` wholesale, so it never reaches the within-batch computation — it verifies the wrapper forwards. Suggest renaming to `..._forwards_threshold_to_link_utils`.

PR #2885, opened 2026-07-21
fix(control-plane): preserve inherited observations configuration

r266-tech:
## Summary

- model `enable_observations` as an inherited/enabled/disabled bank setting instead of collapsing inheritance into a boolean switch
- derive the boolean from bank-only overrides while keeping the other Observations fields on their resolved values
- reconcile PATCH responses per field so server canonicalization and edits made during an in-flight save are both preserved

## Why

#2827 called out the same latent inheritance gap for `enable_observations`. The bank config endpoint returns both the resolved hierarchy and bank-only overrides, but the UI initialized the Observations boolean from the resolved value and sent it back on every section save. Saving an unrelated Observations field could therefore create a sticky bank override and prevent later tenant or server policy changes from propagating.

The tri-state control sends `null` to clear the bank override. It shows the current inherited state when the resolved value is available, and stays neutral when permission filtering makes that parent value unavailable.

## Test plan

- `npm test -w @vectorize-io/hindsight-control-plane` (86 tests)
- `npm exec -w @vectorize-io/hindsight-control-plane tsc -- --noEmit`
- `npm run i18n:check -w @vectorize-io/hindsight-control-plane`
- `./scripts/hooks/lint.sh`
- `npm run build -w @vectorize-io/hindsight-control-plane`

### DIFF
--- a/hindsight-control-plane/src/components/bank-config-view.tsx
+++ b/hindsight-control-plane/src/components/bank-config-view.tsx
@@ -11,6 +11,13 @@ import {
   type RetainStrategy,
   type RetainStrategyValues,
 } from "@/lib/retain-strategy-config";
+import {
+  mergeObservationsOverrides,
+  mergeResolvedObservations,
+  observationsSlice,
+  reconcileObservationsEdits,
+  type ObservationsEdits,
+} from "@/lib/observations-config";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Textarea } from "@/components/ui/textarea";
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):

nicoloboschi:
@r266-tech can you rebase?

r266-tech:
Rebased onto current `main` (`73b575c7`). The only conflict was in control-plane initialization; I preserved the newly added document-storage state and reused the same overrides snapshot for both sections. I also normalized omitted/null PATCH override snapshots during the rebase review.

Focused verification: 12 observations-config tests pass, Prettier passes, and the new GitHub CI run is in progress on `1828b353`.

PR #2887, opened 2026-07-21
fix(providers): skip strict json_schema for MiniMax — degrades output

### LINKED ISSUE #2674: MiniMax-M3 accepts strict_schema=true but outputs fence-wrapped JSON (worse than default json_object mode)
## Bug Summary

When `HINDSIGHT_API_LLM_STRICT_SCHEMA=true` is set, Hindsight sends `response_format: {type: "json_schema", json_schema: {strict: true, ...}}` to the LLM. MiniMax-M3 **accepts** this parameter (returns HTTP 200, no error) but **does not actually grammar-enforce** the output — and perversely, enabling `strict: true` makes the output *worse* (wraps JSON in markdown fences) compared to the default `json_object` mode (which outputs bare JSON).

## Environment

- hindsight-api: v0.8.4
- LLM: MiniMax-M3 (via OpenAI-compatible endpoint `https://api.minimaxi.com/v1`)
- `HINDSIGHT_API_LLM_EXTRA_BODY='{"thinking":{"type":"disabled"}}'`
- OS: Windows 11

## Reproduction

Tested 3 `response_format` modes with the same simple schema (`{name: string, age: integer}`), all with `thinking: disabled` and `temperature: 0.1`:

### Test 1: `json_object` (Hindsight default, `strict_schema=false`)

```json
{"type": "json_object"}
```

**Output:** `{"name":"Alice","age":30}`

`json.loads` → ✅ success

### Test 2: `json_schema` WITHOUT `strict`

```json
{"type": "json_schema", "json_schema": {"name": "person", "schema": {...}}}
```

**Output:** `{"name": "Alice", "age": 30}`

`json.loads` → ✅ success

### Test 3: `json_schema` WITH `strict: true` (Hindsight `strict_schema=true`)

```json
{"type": "json_schema", "json_schema": {"name": "person", "strict": true, "schema": {...}}}
```

**Output:**

~~~
```json
{"name":"Alice","age":30}
```
~~~

`json.loads` → ❌ fails (markdown fence wrapping)

## Root Cause

MiniMax-M3's API endpoint accepts the `strict: true` parameter without error (HTTP 200), but does not implement true grammar-enforced structured output like OpenAI's. Instead, it appears to *add* markdown fence wrapping when `strict: true` is set — the opposite of what strict mode is supposed to do.

This is distinct from a provider returning HTTP 400 "unsupported parameter" — there's no way for Hindsight to detect this via status code alone.

## Impact

A user who sets `HINDSIGHT_API_LLM_STRICT_SCHEMA=true` (reasonably expecting better JSON compliance) will instead get **worse** output: every LLM call will produce fence-wrapped JSON that fails `json.loads`, triggering retries and warnings. The `_strip_code_fences` function can recover the JSON, but the `strict_schema` setting provides zero benefit and active harm.

### Option A: Document provider compatibility (minimal)

Add a note to the `HINDSIGHT_API_LLM_STRICT_SCHEMA` documentation that MiniMax-M3 accepts but does not properly enforce `strict: true`, and that `json_object` mode (default) produces better results with this model.

### Option B: Add a provider blocklist (code change)

In `openai_compatible_llm.py` L761-770, skip `strict_schema` for providers known to have broken strict mode:

```python
if strict_schema and schema is not None and self.provider not in ("minimax",):
    # Use OpenAI's strict JSON schema enforcement
    call_params["response_format"] = {
        "type": "json_schema",
        "json_schema": {
            "name": "response",
            "strict": True,
            "schema": schema,
        },
    }
else:
    # Fall back to json_object mode
    ...
```

### Option C: Detect fence-wrapped output under strict mode and auto-disable (complex)

After the first LLM call with `strict: true`, if the output is fence-wrapped, automatically fall back to `json_object` mode for subsequent calls. This is more complex but handles unknown providers.

## Related

- #2672 (`_strip_code_fences` truncation bug) — this fix would recover the JSON from Test 3's fence-wrapped output, but `strict_schema` would still provide no benefit over the default `json_object` mode.
- #2673 (json_repair fallback) — this would handle the parse failure, but again, `strict_schema` provides no benefit.

handnewb:
## Summary

Fixes #2674 — MiniMax-M3 accepts `json_schema` with `strict: true` (HTTP 200, no error) but does not actually grammar-enforce the output. Worse, enabling `strict: true` makes the output *degrade*: the model wraps valid JSON in markdown code fences, while the default `json_object` mode produces clean, bare JSON.

## What changed

`hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py`:

- Added `_STRICT_SCHEMA_UNSUPPORTED_PROVIDERS` (analogous to the existing `_TOOL_CHOICE_REQUIRED_UNSUPPORTED_PROVIDERS`) containing `"minimax"`
- The `strict_schema` path now checks `self.provider not in _STRICT_SCHEMA_UNSUPPORTED_PROVIDERS` before emitting `json_schema` with `strict: true`

Providers in this set fall through to the existing `json_object` mode, which injects the schema into the system prompt — the path that already produces clean, parseable JSON for MiniMax-M3.

## Evidence (from issue #2674)

| Mode | MiniMax-M3 output | `json.loads` |
|------|-------------------|-------------|
| `json_object` (default) | `{"name":"Alice","age":30}` | ✅ |
| `json_schema` WITHOUT strict | `{"name": "Alice", "age": 30}` | ✅ |
| `json_schema` WITH `strict: true` | ` ```json\n{"name":"Alice","age":30}\n``` ` | ❌ |

## Why this approach

- Follows the same pattern as `_TOOL_CHOICE_REQUIRED_UNSUPPORTED_PROVIDERS` (line 84)
- Easy to extend: add any future provider that exhibits the same behavior
- Zero overhead for providers that *do* support strict schema (the check is a `frozenset` membership test)
- The fallback path (schema-in-prompt + `json_object`) is already battle-tested — it is what MiniMax users have been using successfully with `strict_schema=false`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -83,6 +83,12 @@ def _validate_ollama_num_ctx(value: Any) -> int | None:
 # intentionally excluded (#1179).
 _TOOL_CHOICE_REQUIRED_UNSUPPORTED_PROVIDERS = frozenset({"lmstudio", "ollama"})
 
+# Providers that accept a json_schema / strict:true response_format but do not
+# actually grammar-enforce the output — and perversely output *worse* JSON
+# (e.g. fence-wrapped) than the default json_object mode.
+# See #2674 (MiniMax-M3).
+_STRICT_SCHEMA_UNSUPPORTED_PROVIDERS = frozenset({"minimax"})
+
 
 class ProviderResponseError(RuntimeError):
     """Raised when a provider returns a success response without usable content."""
@@ -829,7 +835,7 @@ class OpenAICompatibleLLM(LLMInterface):
             if hasattr(response_format, "model_json_schema"):
                 schema = strict_json_schema(response_format) if strict_schema else response_format.model_json_schema()
 
-            if strict_schema and schema is not None:
+            if strict_schema and schema is not None and self.provider not in _STRICT_SCHEMA_UNSUPPORTED_PROVIDERS:
                 # Use OpenAI's strict JSON schema enforcement
                 call_params["response_format"] = {
                     "type": "json_schema",


nicoloboschi (approved):

nicoloboschi:
use HINDSIGHT_API_LLM_STRICT_SCHEMA=false

PR #2890, opened 2026-07-22
Fix within-batch cosine similarity for non-unit embeddings

### LINKED ISSUE #2873: Ensure embedding outputs are L2-normalized before dot-product semantic link scoring
## Problem

Within-batch semantic link creation computes similarity with a raw dot product:

```python
similarities = np.dot(other_embeddings, new_embeddings_matrix[i])
```

This is treated as cosine similarity and compared against the semantic-link threshold. A dot product is equivalent to cosine similarity only when both embeddings are L2-normalized.

The embedding validation layer currently checks vector presence and dimensionality, but does not enforce or verify L2 normalization. The default `BAAI/bge-small-en-v1.5` SentenceTransformer pipeline includes a `Normalize` module, so the default model is safe. However, other local models or embedding providers may return non-unit vectors.

This can make semantic links depend on vector magnitude rather than only direction. It also creates inconsistent behavior between:

- within-batch links, which use `np.dot`; and
- links to existing rows, which use pgvector cosine distance (`<=>`).

The same pair of texts could therefore pass or fail the semantic-link threshold depending on the embedding backend or whether they were retained in the same batch.

## Suggested direction

Any of the following would make the behavior well-defined:

1. **Compute cosine similarity explicitly.** Replace the raw dot-product assumption in the within-batch semantic-link path with an actual cosine calculation.
2. **Make L2 normalization a documented provider contract.** Clearly require every local model and remote embedding service used by Hindsight to return L2-normalized vectors, so the optimized dot product is valid.
3. **Validate the invariant at runtime.** Check embedding norms at the shared embedding boundary and fail with a clear error when a provider returns vectors that are not L2-normalized, rather than silently producing incorrect semantic-link scores.

If validation is used, it should apply an appropriate floating-point tolerance and explicitly reject zero-norm and non-finite vectors. The key requirement is that Hindsight must not silently treat an unconstrained dot product as cosine similarity.

koriyoshi2041:
Closes #2873.

Within-batch semantic links currently use a raw dot product even though the configured threshold is cosine similarity and the database path uses pgvector cosine distance. This makes link creation depend on vector magnitude for providers that return non-unit embeddings.

This change normalizes finite, non-zero vectors before the dot product and excludes zero-norm/non-finite vectors, whose cosine similarity is undefined. It adds regression coverage for both the magnitude mismatch and invalid vectors.

Verification:
- `uv run --project hindsight-api-slim pytest hindsight-api-slim/tests/test_link_utils.py -q` (44 passed)
- `uv run --project hindsight-api-slim ruff check hindsight-api-slim/hindsight_api/engine/retain/link_utils.py`
- `uv run --project hindsight-api-slim ruff format --check hindsight-api-slim/hindsight_api/engine/retain/link_utils.py hindsight-api-slim/tests/test_link_utils.py`
- `git diff --check`

Note: the repository-wide pre-commit lint reached unrelated frontend checks without installed `@eslint/js`; the focused Python tests and source lint above pass.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
@@ -666,7 +666,7 @@ def compute_semantic_links_within_batch(
     """
     Compute semantic links between units within the same batch (no DB needed).
 
-    Uses numpy dot product on embeddings already in memory — instant.
+    Uses cosine similarity on embeddings already in memory — instant.
 
     Args:
         unit_ids: Unit IDs (real IDs from insert_facts_batch)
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Verified the analysis: within-batch links scored a raw dot product against a cosine threshold, inconsistent with the pgvector cosine (`1 - (embedding <=> emb)`) DB path — so non-unit embeddings could flip link decisions by batch placement. Fix normalizes before the dot product and correctly excludes zero-norm/non-finite vectors, with regression tests that pin the magnitude case. Minimal, targeted, well-tested. LGTM.

PR #2892, opened 2026-07-22
fix(curation): preserve causal links across invalidation

### LINKED ISSUE #2864: Causal links are lost after memory edit or invalidate/restore
## Summary

Memory curation can permanently lose raw-fact causal links.

An edit explicitly deletes every `memory_links` row incident to the edited memory. Invalidation moves the memory out of `memory_units`, which also removes its incident links. The subsequent graph-maintenance job rebuilds only temporal and semantic links, so `caused_by` links are not restored.

This affects every memory edit, including context-only edits, and means invalidate/restore is not fully reversible for memories that participate in causal relationships.

## How causal links are created

During retain, the fact-extraction LLM emits `causal_relations`. The retain pipeline writes those relations as `memory_links.link_type = 'caused_by'` through `create_causal_links_batch()`.

Relevant code:

- `hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py`
- `hindsight-api-slim/hindsight_api/engine/retain/link_utils.py` in `create_causal_links_batch()`

These relations are extraction output. They cannot be derived from embeddings or dates alone.

## Current edit behavior

`update_memory_unit()` executes the equivalent of:

```sql
DELETE FROM memory_links
WHERE from_unit_id = $1 OR to_unit_id = $1
```

This removes semantic, temporal, and `caused_by` edges indiscriminately.

The graph-maintenance relink pass only considers and recreates `temporal` and `semantic` links. It does not rerun fact extraction or causal-relation extraction. Consolidation regenerates observations, not raw-fact `caused_by` rows.

Example:

```text
B: Alice could not pay rent
B --caused_by--> A
A: Alice lost her job
```

Changing only the context of B deletes `B --caused_by--> A`. Neither consolidation nor graph maintenance recreates it.

## Current invalidate/restore behavior

Invalidation archives the memory row but does not snapshot its incoming or outgoing causal links. Restore moves the row back, restores entity associations, recomputes its embedding/search vector, and submits graph maintenance. Since graph maintenance only rebuilds temporal/semantic links, the causal relationships remain missing after restore.

## Expected behavior

Causal links should never be silently lost as an unrelated side effect of curation.

Suggested policy:

- Preserve existing causal links for edits that do not change fact meaning, such as `context`, dates, fact type, or entities.
- Define an explicit policy for `text` edits:
  - preserve existing causal links by default;
  - accept explicit causal-relation updates; or
  - run a targeted causal re-extraction step.
- Snapshot incident causal links when invalidating a memory and restore them when the memory returns to `valid`.
- Restore both incoming and outgoing causal links.
- Keep causal restoration independent from observation consolidation.

## Suggested tests

- Context-only edit preserves incoming and outgoing `caused_by` links.
- Date, fact-type, and entity edits do not silently delete causal links.
- Text-edit behavior follows the selected explicit causal policy.
- Invalidate removes causal links from the active graph.
- Restore recreates the exact incoming and outgoing causal links that existed before invalidation.
- Multiple invalidate/restore cycles are idempotent and do not duplicate causal links.

Sanderhoff-alt:
## Summary

This PR preserves raw-fact causal relationships across memory curation without introducing a new archive table or rerunning LLM extraction. It stores only causal edges that are currently suspended because at least one endpoint is invalidated, and rematerializes an edge when both endpoints are live again.

The change covers the canonical `caused_by` relation and the historical `causes`, `enables`, and `prevents` relation types.

## Problem

Causal edges are retain-time extraction results stored in `memory_links`. Unlike temporal and semantic links, they cannot be deterministically reconstructed from dates or embeddings.

Invalidating a memory moves it out of `memory_units`, so foreign-key cascades remove all incident physical edges. Restoring the memory later submits graph maintenance, but graph maintenance only rebuilds temporal and semantic links. The original causal assertion is therefore permanently lost.

Editing had a related failure mode: curation deleted every incident `memory_links` row even when only non-text fields changed. A context, date, fact-type, or entity edit could therefore erase a causal assertion even though the underlying proposition remained unchanged.

## Design goals

- Preserve the exact direction, relation type, and weight of suspended causal edges.
- Restore both incoming and outgoing causal edges.
- Keep an edge out of the active graph while either endpoint is invalidated.
- Make restoration independent of endpoint invalidation and restoration order.
- Treat a real text change as invalidating the old causal assertion instead of silently preserving or guessing it.
- Keep restoration idempotent and safe under concurrent edits, invalidations, restorations, and deletion.
- Avoid a new table, a new graph-maintenance queue responsibility, and LLM re-extraction.
- Avoid migration-time scans or backfills across existing banks.
- Leave temporal and semantic curation behavior outside this PR.

## Data model

The migration adds `suspended_causal_links` to both `memory_units` and `invalidated_memory_units`.

- PostgreSQL stores the value as `JSONB NOT NULL DEFAULT '[]'::jsonb`.
- Oracle stores the value as a JSON-validated `CLOB` with an empty-array default.
- The field is intentionally unindexed because it is read by memory ID during low-frequency curation operations, not searched globally.

Each descriptor contains `from_unit_id`, `to_unit_id`, `link_type`, and `weight`. Its stable identity is `(from_unit_id, to_unit_id, link_type)`.

`memory_links` remains the materialized active graph used by retrieval and traversal. `suspended_causal_links` is not a second permanent graph: it contains only edges that are currently paused because at least one endpoint is invalidated. Once an edge is successfully rematerialized, its descriptor is consumed from both endpoints.

## Why the descriptor is stored on both endpoints

Duplicating the small descriptor on both endpoint memories makes restoration local and independent of ordering.

For an edge `A --caused_by--> B`, suppose both memories become invalidated. If only A is restored, the edge remains suspended because B is not live. If B is restored later, B carries the same descriptor and can determine that both endpoints are now live. The same result is reached when B is restored first.

Storing the descriptor on only the first invalidated endpoint would require scanning or transferring state between archive rows. It would also make the second restore unable to discover the edge when the restore order is reversed. Storing it on both endpoints avoids that additional lifecycle mechanism.

### Retain and imported data

Retain continues to write causal relationships only to `memory_links`. It does not eagerly populate `suspended_causal_links`, avoiding two extra memory-row updates on the high-frequency ingestion path.

Existing and newly retained memories use the same lazy snapshot path. The first invalidation of either endpoint reads the currently materialized causal edges and stores any missing descriptors on both endpoint rows.

### Non-text edits

Changes to context, dates, fact type, or entities preserve physical causal edges and any already suspended descriptors. These fields may affect other derived graph data, but they do not by themselves invalidate the retained proposition's causal meaning.

### Text edits

When the stored text actually changes, the old causal assertion is no longer trusted. The transaction removes every incoming and outgoing causal edge incident to the edited memory and removes the matching suspended descriptors from both live or archived peer endpoints.

This PR deliberately does not rerun the LLM or infer replacement causal relationships. A future API that accepts explicit causal updates can reuse the same descriptor and materialization helpers without changing this lifecycle.

### Invalidation

Before moving a live memory into `invalidated_memory_units`, the transaction reads all materialized incident causal edges and merges their descriptors into both endpoints. The existing move then carries the local snapshot into the archive, and the foreign-key cascade removes the physical edges from the active graph.

Repeated invalidation does not duplicate descriptors because merges use the descriptor identity.

### Restoration

After restoring a memory to `memory_units`, the transaction examines its suspended descriptors. A descriptor is eligible only when both referenced endpoints exist in the same bank's live table.

Eligible edges are inserted through the existing idempotent bulk-insert path. Descriptor copies are removed from both endpoints in the same transaction. If either endpoint remains invalidated, the descriptor is retained and no dangling physical edge is created.

### Permanent deletion

Single-memory permanent deletion is not currently exposed to users. This PR still adapts the internal deletion path so a future public API cannot leave stale causal descriptors on surviving endpoints.

Before deleting a live memory, the transaction removes all of its materialized causal edges and removes their suspended descriptor copies from surviving live or archived peers. Bank deletion needs no per-edge cleanup because all endpoint rows are deleted together.

## Concurrency and locking

Every causal curation operation first discovers the target's materialized and suspended incident edges. It then locks the target and all discovered endpoint rows with `SELECT ... FOR UPDATE` in deterministic UUID order, regardless of whether a row currently lives in `memory_units` or `invalidated_memory_units`.

The operation rereads actionable state after acquiring the locks and performs descriptor updates, row movement, edge deletion, or edge materialization in the same transaction. Locking before an archive-to-live move is important: otherwise concurrent restores of opposite endpoints could each miss the other's uncommitted live row.

Each operation performs the complete endpoint discovery and lock acquisition once. Internal snapshot, cleanup, and materialization stages reuse those transaction-held locks rather than repeating `FOR UPDATE`. Unrelated memories and ordinary reads are not serialized.

## Existing data and migration behavior

There is no full-table backfill. A preexisting causal edge is captured lazily when one endpoint is first invalidated. Text editing and permanent deletion merge the current physical and already suspended representations only for cleanup; they do not create a snapshot that would immediately be removed.

Causal edges that were already lost before this migration cannot be reconstructed exactly and are intentionally not guessed.

### Dedicated causal archive table

A separate table would centralize descriptors but add a new persistence lifecycle, cleanup rules, tenancy considerations, and another source that every restore and delete path must coordinate. The endpoint field is sufficient for a sparse relation and keeps the change local.

### Store snapshots only on archived rows

This requires transferring pending edge state as endpoints move independently between live and archived tables. It also makes restoration order significant unless the restore path scans other archive rows.

### Keep dangling rows in `memory_links`

Weakening foreign keys or retaining links to archived endpoints would make the active graph contain invalid references and complicate every retrieval query.

### Re-extract causal relations with an LLM

Re-extraction lacks the original retain context, is nondeterministic, and can change direction, type, or edge existence. It is also much heavier than preserving the exact extracted assertion.

### Use graph maintenance

Graph maintenance can derive temporal and semantic relationships from stored data, but it cannot reconstruct a retain-time causal assertion. Adding causal restoration there would mix persisted provenance with derived graph maintenance and would still require storing the original descriptor somewhere.

## Validation

- `uv run pytest tests/test_memory_curation.py tests/test_causal_link_taxonomy.py tests/test_link_utils.py -q -n 0` — 67 passed.
- `uv run ty check hindsight_api/` — passed.
- Concurrency coverage verifies that two endpoint restores materialize one edge and that concurrent invalidation plus peer text editing leaves no stale descriptor.
- The lifecycle tests cover all four supported causal types, one- and two-endpoint invalidation, deferred restoration, text cleanup, non-text preservation, and permanent-delete cleanup.

Closes #2864.

### DIFF
--- a/hindsight-api-slim/hindsight_api/alembic/versions/f3a9c1e7b2d4_add_causal_link_provenance.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/f3a9c1e7b2d4_add_causal_link_provenance.py
@@ -1,0 +1,86 @@
+"""Add durable causal-link provenance to live and archived memories.
+
+``memory_links`` remains the materialized graph used by retrieval. The new
+JSON field carries the small causal descriptor set with both endpoint memories
+so curation can restore an edge after both endpoints become live again.
+
+Revision ID: f3a9c1e7b2d4
+Revises: d7b2f8a1c934
+Create Date: 2026-07-22
+"""
+
+from collections.abc import Sequence
+
+from alembic import context, op
+
+from hindsight_api.alembic._dialect import run_for_dialect
+
+revision: str = "f3a9c1e7b2d4"
+down_revision: str | Sequence[str] | None = "d7b2f8a1c934"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _pg_schema_prefix() -> str:
+    schema = context.config.get_main_option("target_schema")
+    return f'"{schema}".' if schema else ""
+
+
+def _pg_upgrade() -> None:
+    schema = _pg_schema_prefix()
+    for table in ("memory_units", "invalidated_memory_units"):
+        op.execute(
+            f"ALTER TABLE {schema}{table} "
+            "ADD COLUMN IF NOT EXISTS suspended_causal_links JSONB NOT NULL DEFAULT '[]'::jsonb"
+        )
+
+
+def _pg_downgrade() -> None:
+    schema = _pg_schema_prefix()
+    for table in ("invalidated_memory_units", "memory_units"):
+        op.execute(f"ALTER TABLE {schema}{table} DROP COLUMN IF EXISTS suspended_causal_links")
+
+
+def _oracle_add_column(table: str, constraint: str) -> None:
+    op.execute(
+        f"""
+        BEGIN
+            EXECUTE IMMEDIATE 'ALTER TABLE {table} ADD (
+                suspended_causal_links CLOB DEFAULT ''[]'' NOT NULL
+                CONSTRAINT {constraint} CHECK (suspended_causal_links IS JSON)
+            )';
+        EXCEPTION WHEN OTHERS THEN
+            IF SQLCODE != -1430 THEN RAISE; END IF;
+        END;
+        """
+    )
+
+
+def _oracle_drop_column(table: str) -> None:
+    op.execute(
+        f"""
+        BEGIN
+            EXECUTE IMMEDIATE 'ALTER TABLE {table} DROP COLUMN suspended_causal_links';
+        EXCEPTION WHEN OTHERS THEN
+            IF SQLCODE != -904 THEN RAISE; END IF;
+        END;
+        """
+    )
+
+
+def _oracle_upgrade() -> None:
+    _oracle_add_column("memory_units", "ck_mu_suspended_causal_json")
+    _oracle_add_column("invalidated_memory_units", "ck_imu_suspended_causal_json")
+
+
+def _oracle_downgrade() -> None:
+    _oracle_drop_column("invalidated_memory_units")
+    _oracle_drop_column("memory_units")
+
+
+def upgrade() -> None:
+    run_for_dialect(pg=_pg_upgrade, oracle=_oracle_upgrade)
+
+
+def downgrade() -> None:
+    run_for_dialect(pg=_pg_downgrade, oracle=_oracle_downgrade)
--- a/hindsight-api-slim/hindsight_api/engine/causal_links.py
+++ b/hindsight-api-slim/hindsight_api/engine/causal_links.py
@@ -5,9 +5,32 @@ preserves historical relationship types so existing banks keep their graph
 semantics without allowing new retain output to create those types.
 """
 
-CANONICAL_CAUSAL_LINK_TYPE = "caused_by"
-LEGACY_CAUSAL_LINK_TYPE_NAMES = ("causes", "enables", "prevents")
+from typing import Literal
+from uuid import UUID
+
+from pydantic import BaseModel, ConfigDict, Field
+
+CausalLinkType = Literal["caused_by", "causes", "enables", "prevents"]
+
+CANONICAL_CAUSAL_LINK_TYPE: CausalLinkType = "caused_by"
+LEGACY_CAUSAL_LINK_TYPE_NAMES: tuple[CausalLinkType, ...] = ("causes", "enables", "prevents")
 
 CANONICAL_CAUSAL_LINK_TYPES = frozenset({CANONICAL_CAUSAL_LINK_TYPE})
 LEGACY_CAUSAL_LINK_TYPES = frozenset(LEGACY_CAUSAL_LINK_TYPE_NAMES)
 CAUSAL_LINK_TYPES = (CANONICAL_CAUSAL_LINK_TYPE, *LEGACY_CAUSAL_LINK_TYPE_NAMES)
+
+
+class CausalLinkDescriptor(BaseModel):
+    """Durable causal provenance carried by both endpoint memory units."""
+
+    model_config = ConfigDict(frozen=True)
+
+    from_unit_id: UUID
+    to_unit_id: UUID
+    link_type: CausalLinkType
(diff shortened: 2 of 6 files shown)


nicoloboschi:
Thanks for this — the analysis in the description is spot on, and it's what convinced us the fix belongs on the archive row rather than in graph maintenance or a re-extraction step.

We're going with a smaller variant in #2951, so I'm closing this one. The difference is where the descriptors live: #2951 stores them **only** on `invalidated_memory_units`, not on both endpoints. The archive row is already being written inside the curation transaction, and the snapshot taken at invalidation time also picks up descriptors parked on archived peers that name the unit — so an edge whose both endpoints are invalidated still ends up with a copy on both archive rows, and whichever endpoint is restored last recreates it. That gets the order-independence this PR achieves with the live-row column, without adding a column to `memory_units` or the `SELECT … FOR UPDATE` endpoint-locking protocol.

Rematerialization also reuses the existing bulk-link writer: on PostgreSQL its `locked` CTE already filters to endpoints present in `memory_units` and inserts `ON CONFLICT DO NOTHING`, so still-archived peers, permanently deleted peers, and repeated invalidate/restore cycles are all handled without a dedicated cleanup path.

One policy difference worth flagging: #2951 preserves causal links on text edits instead of clearing them. Both are listed as acceptable in #2864 — we picked preservation because keeping an assertion is reversible and nothing in the pipeline can re-derive one that's been deleted.

Really appreciate the write-up on the alternatives considered; it made reviewing the trade-offs much faster.

PR #2895, opened 2026-07-22
fix(curation): make edit maintenance field-aware

### LINKED ISSUE #2862: Memory edit uses coarse-grained graph maintenance for every field change
## Summary

`update_memory_unit()` currently sends every editable field through the same destructive maintenance path. A context-only edit performs the same embedding and graph work as a text, date, fact-type, or entity edit.

This is unnecessarily expensive and deletes link types that are unaffected by the changed field.

## Current behavior

Any edit to `text`, `context`, `occurred_start`, `occurred_end`, `fact_type`, or `entities` sets `doing_edit` and then:

1. recomputes the embedding;
2. calls `enqueue_relink_victims(..., [memory_id])`;
3. deletes every `memory_links` row where the edited memory is either endpoint;
4. removes derived observations;
5. submits consolidation and graph maintenance.

Relevant code:

- `hindsight-api-slim/hindsight_api/engine/memory_engine.py` in `update_memory_unit()`
- `hindsight-api-slim/hindsight_api/engine/graph_maintenance.py` in `enqueue_relink_victims()`

## Why field-aware maintenance is needed

The graph dependencies differ by field:

| Changed field | Required maintenance |
|---|---|
| `context` | Refresh text-search data and derived observations/consolidation; no embedding or graph-link rebuild |
| `text` | Re-embed and rebuild semantic links; refresh derived observations; invalidate causal assertions whose source text changed |
| `occurred_start` / `occurred_end` | Re-embed because dates augment the embedding input, then rebuild semantic and temporal links |
| `fact_type` | Rebuild semantic and temporal links because candidate searches are fact-type scoped |
| `entities` | Rebuild entity associations/cooccurrences and semantic links because entities are included in the embedding |
| invalidate/restore | Full removal/restoration maintenance |

For example, context participates in text search, reranking, and consolidation, but it is not included in `augment_texts_with_dates()`, which produces the embedding input. Re-embedding and deleting graph links for a context-only edit provides no graph benefit.

Maintenance should also be based on values that actually changed, not merely on which optional parameters were supplied. Submitting the current value of a field should not trigger destructive maintenance for that dependency.

## Expected behavior

- Compare requested values with the stored memory and determine maintenance needs from fields that actually changed.
- Recompute the embedding only when text, dates, or entities changed.
- Delete and rebuild only the temporal or semantic link types affected by those changes.
- Preserve unrelated materialized link types.
- Do not submit graph maintenance for a context-only edit.
- Keep derived-observation invalidation and consolidation independent from graph maintenance.
- Preserve existing behavior for invalidate, restore, and hard-delete paths.

## Out of scope

Queuing an edited or restored memory itself to regenerate its outgoing temporal and semantic adjacency is tracked separately in #2889 and implemented by PR #2893. This issue only controls which maintenance is necessary for each edited field.

## Suggested tests

- A context-only edit preserves all existing graph links, does not recompute the embedding, and does not submit graph maintenance.
- Supplying values equal to the stored values does not trigger unnecessary embedding or graph work.
- A text edit rebuilds semantic links, removes causal assertions invalidated by the changed text, and preserves temporal links.
- A date edit rebuilds semantic and temporal links while preserving unrelated link types.
- A fact-type edit rebuilds both fact-type-scoped semantic and temporal links.
- An entity edit rebuilds entity associations, cooccurrences, and entity-augmented semantic links.

Sanderhoff-alt:
## Summary

Memory edits currently send every editable field through the same destructive maintenance path. Even a context-only correction recomputes the embedding, deletes every incident graph link, invalidates observations, and submits graph maintenance.

This change derives maintenance work from values that actually changed. It preserves unrelated graph state, avoids unnecessary embedding and graph work, and leaves delete, invalidate, restore, and document-replacement maintenance unchanged.

Closes #2862.

## Problem model

Each editable field feeds a different set of derived data. Treating all fields alike is both wasteful and destructive: context is not part of the embedding input, fact type affects candidate scopes without changing the embedding vector, occurred dates affect the augmented embedding, `event_date` determines temporal placement, and entities affect the embedding plus entity associations.

The previous edit path did not compare requested values with stored values. Supplying an unchanged value still rebuilt all derived state and updated the memory's edit marker.

## Design goals

- Base maintenance on normalized values that actually changed.
- Recompute only derived state that depends on those values.
- Preserve materialized link types whose inputs remain valid.
- Keep consolidation and graph-maintenance scheduling independent.
- Preserve the behavior of every non-edit caller of the shared victim-enqueue helper.
- Avoid a queue schema change or a second graph-maintenance mechanism.

## Dependency model

| Changed value | Embedding | Semantic links | Temporal links | Causal links | Text search |
|---|---:|---:|---:|---:|---:|
| `context` | Preserve | Preserve | Preserve | Preserve | Refresh |
| `text` | Recompute | Invalidate | Preserve | Invalidate | Refresh |
| `occurred_start` | Recompute | Invalidate | Invalidate when `event_date` changes | Preserve | Preserve |
| `occurred_end` | Recompute | Invalidate | Preserve | Preserve | Preserve |
| `fact_type` | Preserve | Invalidate | Invalidate | Preserve | Preserve |
| `entities` | Recompute | Invalidate | Preserve | Preserve | Preserve |

Occurred dates invalidate semantic links because the retain pipeline includes temporal augmentation in the embedding input. Temporal neighbors depend only on `event_date` and `fact_type`, so changing `occurred_end` does not invalidate temporal links. Text invalidates causal assertions because those assertions describe the proposition represented by the stored text; non-text edits do not change that proposition.

### Change detection

The edit path first normalizes requested values using the existing API semantics, then compares them with the live row. Entity names are treated as a case-insensitive set, so ordering, casing, and duplicate inputs do not create false changes.

If no normalized value changed, the operation does not recompute the embedding, replace entity associations, delete links or observations, update `edited_at`, or submit background maintenance.

### Selective embedding and search maintenance

The embedding is recomputed only when text, occurred dates, or entities changed. Entity names are loaded only when they are needed for comparison or embedding reconstruction.

The text-search expression is regenerated only for text or context changes. A context-only edit therefore updates its searchable and consolidation inputs without touching vector or graph state.

### Selective link invalidation

The edit path constructs the affected link-type set from the dependency model and deletes only those incident rows. Semantic and temporal rows are independently selectable, all causal relation types are removed only after a real text change, and unrelated rows remain materialized.

### Type-aware victim scheduling

`enqueue_relink_victims()` accepts independent temporal and semantic selectors. An edit queues only surviving source memories whose outgoing links belong to a type that will actually be removed.

All existing delete-oriented callers use the default selectors, which retain the previous `temporal OR semantic` query exactly. This covers document deletion, single-memory deletion, invalidation, and document replacement during retain without changing their behavior.

The queue remains unit-granular. A queued worker may still inspect both derived-link caps for that source, but unaffected physical rows are preserved and unaffected incoming sources are not scheduled. Encoding work types in the queue would require a schema and worker-protocol change that is not justified for this fix.

### Background task separation

Every real edit still invalidates derived observations and requests consolidation. Graph maintenance is submitted only when semantic or temporal links changed. Entity edits still submit it through their semantic invalidation, while context-only edits do not create an empty graph-maintenance operation.

## Compatibility and concurrency

No database migration or new storage is introduced. Queue insertion, ordering, deduplication, claiming, and worker drain behavior are unchanged.

The new helper parameters default to full temporal and semantic victim discovery, and every pre-existing caller continues to omit them. The edit path performs one victim-enqueue call for the complete affected type set rather than separate calls per type.

## Scope

This PR is limited to deciding which derived state a memory edit invalidates. Invalidation, restore, hard-delete, queue-worker behavior, and causal provenance storage are not changed here.

## Behavioral examples

A context-only correction refreshes text search and derived observations while preserving the embedding and every graph link.

A text correction recomputes the embedding, removes semantic and causal rows, preserves temporal rows, and queues only incoming sources that lost semantic links.

An `occurred_start` correction that changes `event_date`, or a fact-type correction, removes semantic and temporal rows while preserving causal assertions. An `occurred_end` correction recomputes the embedding and removes semantic rows while preserving temporal and causal links. An entity correction replaces entity associations, recomputes the entity-augmented embedding, and removes semantic rows while preserving temporal and causal links.

Submitting values equal to the stored memory produces no derived maintenance. Entity inputs that differ only by ordering, casing, or duplicates are also treated as unchanged.

## Validation

- `uv run pytest tests/test_memory_curation.py tests/test_graph_maintenance.py -q -n 0` — 30 passed.
- Delete-memory, delete-document, document-replacement, observation-cleanup, and graph deadlock paths were exercised without regression; 38 additional relevant tests passed. Three real-LLM fixtures in the same local run could not initialize because no Groq API key was configured and did not enter the tested code paths.
- `uv run ty check hindsight_api/` — passed.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/graph_maintenance.py
+++ b/hindsight-api-slim/hindsight_api/engine/graph_maintenance.py
@@ -104,6 +104,9 @@ async def enqueue_relink_victims(
     bank_id: str,
     deleted_unit_ids: list[str],
     ops: Any,
+    *,
+    relink_temporal: bool = True,
+    relink_semantic: bool = True,
 ) -> int:
     """Enqueue surviving units whose outgoing temporal/semantic links pointed at
     ``deleted_unit_ids`` for later link top-up.
@@ -118,28 +121,35 @@ async def enqueue_relink_victims(
         deleted_unit_ids: Memory_unit IDs about to be (or being) deleted.
         ops: ``DataAccessOps`` instance, supplies the dialect-specific
             bulk-insert path.
+        relink_temporal: Include sources that lost temporal links.
+        relink_semantic: Include sources that lost semantic links.
 
     Returns:
         Number of distinct victim units enqueued (after dedup against rows
         already in the queue).
     """
-    if not deleted_unit_ids:
+    if not deleted_unit_ids or not (relink_temporal or relink_semantic):
         return 0
 
     deleted_uuids = [uuid_module.UUID(uid) if isinstance(uid, str) else uid for uid in deleted_unit_ids]
     deleted_str_set = {str(uid) for uid in deleted_uuids}
 
-    # Find units (other than the ones being deleted) that have an outgoing
-    # temporal/semantic link pointing at a doomed unit. Entity links are
-    # intentionally excluded — they're scheduled for removal and would only
-    # add noise to the recompute job.
+    if relink_temporal and relink_semantic:
+        link_type_predicate = "link_type IN ('temporal', 'semantic')"
+    elif relink_temporal:
+        link_type_predicate = "link_type = 'temporal'"
+    else:
+        link_type_predicate = "link_type = 'semantic'"
+
+    # An edit may preserve one derived link type, so only queue sources whose
+    # outgoing links will actually be removed. Delete callers keep both defaults.
     victim_rows = await conn.fetch(
         f"""
         SELECT DISTINCT from_unit_id
         FROM {fq_table("memory_links")}
         WHERE to_unit_id = ANY($1::uuid[])
           AND bank_id = $2
-          AND link_type IN ('temporal', 'semantic')
+          AND {link_type_predicate}
         """,
         deleted_uuids,
         bank_id,
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -6675,16 +6675,14 @@ class MemoryEngine(MemoryEngineInterface):
 
         - **Edit** (``text``/``context``/``occurred_start``/``occurred_end``/
           ``new_fact_type``/``entities``): correct what the LLM extracted.
-          Re-embeds (text + dates + entities feed the embedding), drops derived
-          observations + links, and re-consolidates. For date/context fields,
-          ``""`` clears to NULL and ``None`` leaves unchanged; ``new_fact_type``
-          must be world/experience. ``entities`` (when not None) replaces the
-          unit's entity set: names are resolved/find-or-created via the same
-          resolver retain uses, ``unit_entities`` + cooccurrence are rebuilt, and
-          ``[]`` detaches all entities. Entities orphaned by the swap, and any
-          now-stale cooccurrence rows, are reclaimed by the graph-maintenance
-          sweep that this edit submits (entity edges live in ``unit_entities``,
-          not ``memory_links``, so there is nothing to relink directly).
+          Recomputes only the derived state affected by values that actually
+          changed, then re-consolidates. For date/context fields, ``""`` clears
+          to NULL and ``None`` leaves unchanged; ``new_fact_type`` must be
+          world/experience. ``entities`` (when not None) replaces the unit's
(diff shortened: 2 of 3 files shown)


nicoloboschi:
no need for this complexity

PR #2913, opened 2026-07-22
fix(release): guard Claude Code integration version

### LINKED ISSUE #2900: Claude Code plugin.json version not bumped for v0.8.5 — plugin update is a no-op
## Problem

The `plugin.json` version for the Claude Code integration (`hindsight-integrations/claude-code/.claude-plugin/plugin.json`) was not bumped for the v0.8.5 release. It's still `0.7.5` — the same value it had before. This means `claude plugin update hindsight-memory@hindsight` reports "already at the latest version (0.7.5)" and **skips copying the new code from the marketplace to the plugin cache**, even though the integration scripts changed significantly in v0.8.5.

## Impact

Users who rely on `claude plugin update` (including via topgrade, which wraps it) don't get the v0.8.5 integration code. Notably, they miss PR #2801 (`_set_uvx_python_compat`), which adds the native `UV_PYTHON=3.13` workaround for the litellm/PyO3 3.14 build failure — the exact fix that v0.8.5 was supposed to deliver for Python 3.14 hosts.

The only way to get the new code is to uninstall and reinstall the plugin:
```bash
claude plugin uninstall hindsight-memory@hindsight
claude plugin install hindsight-memory@hindsight
```

## Reproduction

1. Have hindsight-memory@hindsight installed at v0.7.5 (pre-v0.8.5)
2. Update the marketplace: `claude plugin marketplace add vectorize-io/hindsight` (fetches v0.8.5 source)
3. Run: `claude plugin update hindsight-memory@hindsight`
4. Observe: "hindsight-memory is already at the latest version (0.7.5)" — no update applied
5. Diff the cache vs marketplace source:
   - Cache: `~/.claude/plugins/cache/hindsight/hindsight-memory/0.7.5/scripts/lib/daemon.py` — old code, no `_set_uvx_python_compat`
   - Marketplace: `~/.claude/plugins/marketplaces/hindsight/hindsight-integrations/claude-code/scripts/lib/daemon.py` — new code with `_set_uvx_python_compat`

## Root cause

`plugin.json` was newly created in commit 6428a83 (the v0.8.5 changelog/blog commit) with `"version": "0.7.5"` — it was never bumped to match the repo release tag. The marketplace.json entry for the plugin has `"version": null` (inherits from plugin.json), so Claude Code compares `0.7.5 == 0.7.5` and skips.

## Suggestion

Bump `plugin.json` version to match the repo release (e.g. `0.8.5`) in the same commit that tags the release, so `claude plugin update` picks up integration changes. Alternatively, if the plugin version is intentionally decoupled from the repo version, document that users need to reinstall (not just update) when integration code changes.

ijevin:
## Summary

- release the pending Claude Code integration changes as plugin version `0.7.6`
- add a read-only `release-integration.sh --check` preflight for manifest/tag/unreleased-change consistency
- block core releases before mutation or tagging when Claude Code integration changes have not been released
- regenerate the docs skill during core releases so its bundled OpenAPI copy cannot drift
- add an isolated regression test for stale changes, manifest mismatch, nested marketplace version keys, extra arguments, the core-release gate, and the docs-skill generation chain

## Why `0.7.6`, not `0.8.5`

Integrations have been independently versioned since #565. The latest Claude Code integration release is `integrations/claude-code/v0.7.5`, so the next integration version is `0.7.6`; coupling it to the core `v0.8.5` tag would break that contract.

The immediate bug was that #2801 changed the Claude Code integration, but no integration release followed. Both plugin manifests therefore stayed at `0.7.5`, causing `claude plugin update hindsight-memory@hindsight` to treat the cached plugin as current.

The new preflight preserves independent integration releases: it is read-only and only blocks a core release when Claude Code has commits after its latest integration tag, when the plugin and marketplace manifests disagree, or when the manifest version does not match the latest integration tag.

The first CI run also exposed an existing release-pipeline gap: core v0.8.5 regenerated `hindsight-docs/static/openapi.json` but did not refresh `skills/hindsight-docs/references/openapi.json`, leaving the bundled copy at v0.8.4. The core release flow now runs `generate-docs-skill.sh` after OpenAPI/client generation, and this PR synchronizes the stale generated copy.

## Testing

- `bash scripts/tests/test-release-integration.sh`
- `cd hindsight-integrations/claude-code && uv run --quiet pytest tests/ -q` (`206 passed`)
- `./scripts/generate-docs-skill.sh` followed by a clean generated-files check
- `bash -n scripts/release-integration.sh`
- `bash -n scripts/release.sh`
- `bash -n scripts/tests/test-release-integration.sh`
- JSON validation for both manifests and the bundled OpenAPI spec
- byte-for-byte comparison of the source and bundled OpenAPI specs
- `git diff --check upstream/main...HEAD`

No `uv.lock` files or advisory locks are included.

Fixes #2900

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "hindsight",
-  "version": "0.7.5",
+  "version": "0.7.6",
   "description": "Official Hindsight integrations for Claude Code",
   "owner": {
     "name": "vectorize-io"
--- a/hindsight-integrations/claude-code/.claude-plugin/plugin.json
+++ b/hindsight-integrations/claude-code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hindsight-memory",
   "description": "Automatic long-term memory for Claude Code via Hindsight. Recalls relevant memories before each prompt, retains conversation transcripts, and provides knowledge page tools.",
-  "version": "0.7.5",
+  "version": "0.7.6",
   "author": {"name": "Hindsight Team", "url": "https://vectorize.io/hindsight"},
   "license": "MIT",
   "keywords": ["memory", "hindsight", "recall", "retain"]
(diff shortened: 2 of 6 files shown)


nicoloboschi:
check new coding agent integration

PR #2916, opened 2026-07-23
fix(cli): preserve HTTP 400 details

### LINKED ISSUE #2912: `memory retain` prints "Request timed out" (0s) for an HTTP 400 batch-mode rejection — swallows the self-explanatory error body
### Summary
With the server configured `HINDSIGHT_API_RETAIN_BATCH_ENABLED=true`, a **synchronous** retain (`async=false`) is correctly rejected by the API with **HTTP 400** and a clear, actionable body. The CLI (`hindsight memory retain`, v0.8.5) discards that body and prints a generic **"Request timed out"** in ~0 seconds. The message blames network/server latency, while the real cause *and the fix* were in the response body the CLI threw away. This is the bug's real cost: it's diagnostic-hostile — it sends you looking at the network when the answer was one flag.

### Environment
- CLI: hindsight 0.8.5
- API: Hindsight HTTP API 0.8.5 (`ghcr.io/vectorize-io/hindsight:0.8.5`)
- Server env: `HINDSIGHT_API_RETAIN_BATCH_ENABLED=true`

### Steps to reproduce
1. Run the API with `HINDSIGHT_API_RETAIN_BATCH_ENABLED=true`.
2. `hindsight memory retain -o json <bank> "<content>"`  (no `--async` → `async=false`).

### Actual — CLI prints, in ~0 s:
```
✗ Request timed out
API URL: http://<host>:8420
Possible causes:
  • The API server is slow to respond
  • Network latency is too high
```
`-v/--verbose` adds nothing — the response body is never surfaced.

### Expected
Surface the server's 400 body (or at minimum, do not label a fast 4xx as a timeout):
```
Batch API is enabled (HINDSIGHT_API_RETAIN_BATCH_ENABLED=true) but async=false.
Batch operations can take several minutes to hours and will timeout in synchronous
mode. Please set async=true in your request to use background processing, or disable
batch API by setting HINDSIGHT_API_RETAIN_BATCH_ENABLED=false in your environment.
```

### Direct API confirmation (CLI bypassed)
```
POST /v1/default/banks/<bank>/memories   body: {"items":[...],"async":false}
→ HTTP 400 in ~3 ms
{"detail":"Batch API is enabled ... Please set async=true ..."}
```
The server behaves correctly and fast. The fault is entirely in the CLI's error handling: a 400 response is being routed onto the timeout code path.

### Impact
Every retain written to the pre-batch (synchronous) convention fails this way. Because the message points at the network, it invites retries and a from-scratch bisect before the one-flag fix (`--async`) is discovered. Surfacing the response body would turn a multi-step investigation into a one-line fix.

koriyoshi2041:
## Summary

- classify HTTP 400 responses before timeout keywords in the server body
- preserve the actionable server response for rejected requests
- add a regression test for the batch-mode message from #2912

## Tests

- `cargo test --manifest-path hindsight-cli/Cargo.toml --bin hindsight` (70 passed)
- `git diff --check`

The repository-wide formatting check currently reports pre-existing formatting drift outside this change.

Closes #2912

### DIFF
--- a/hindsight-cli/src/errors.rs
+++ b/hindsight-cli/src/errors.rs
@@ -25,6 +25,21 @@ fn format_error_message(err: &anyhow::Error, api_url: &str) -> String {
         );
     }
 
+    // Preserve validation details before looking for timeout keywords. A fast
+    // HTTP 400 may legitimately explain that a requested mode would time out;
+    // classifying that body as a transport timeout hides the server's fix.
+    if err_str.contains("400 Bad Request") || err_str.contains("(400)") {
+        return format!(
+            "{} {}\n\n{}\n  {}\n\n{}\n  {}",
+            "✗".bright_red().bold(),
+            "Request rejected (400)".bright_red().bold(),
+            "API URL:".bright_yellow(),
+            api_url.bright_white(),
+            "Server response:".bright_yellow(),
+            err_str.bright_white()
+        );
+    }
+
     // Timeout
     if err_str.contains("timeout") || err_str.contains("Timeout") {
         return format!(
@@ -212,4 +227,22 @@ pub fn print_config_help() {
     println!("    2. Config file (~/.hindsight/config)");
     println!("    3. Default (http://localhost:8888)");
     println!();
+}
+
+#[cfg(test)]
+mod tests {
+    use super::format_error_message;
+
+    #[test]
+    fn http_400_body_that_mentions_timeout_is_not_reported_as_a_timeout() {
+        let error = anyhow::anyhow!(
+            "API request failed (400 Bad Request): \
+             {{\"detail\":\"Batch operations will timeout in synchronous mode. Please set async=true.\"}}"
+        );
+
+        let message = format_error_message(&error, "http://localhost:8888");
+
+        assert!(message.contains("Batch operations will timeout in synchronous mode"));
+        assert!(!message.contains("Request timed out"));
+    }
 }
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
The generated-files failure appears transient or environment-specific. On the exact PR head `3fc58621a`, I ran:

```
./scripts/generate-openapi.sh
./scripts/generate-docs-skill.sh
git status --short
```

The working tree stayed clean. CI reported only a 2-line change in `skills/hindsight-docs/references/openapi.json`, while the source and generated copy are byte-identical locally. I don't have permission to rerun the repository workflow; a rerun should distinguish nondeterministic generation from a persistent Linux-only difference.

koriyoshi2041:
Correction after reproducing from a fresh detached worktree at exact head `3fc58621a`:

The failure is deterministic and comes from the current base, not this CLI patch. The two committed OpenAPI copies already differ before generation:

```diff
-    "version": "0.8.5"
+    "version": "0.8.4"
```

`hindsight-docs/static/openapi.json` is `0.8.5`, while `skills/hindsight-docs/references/openapi.json` is `0.8.4`. Current `main` at `4dc834834` has the same mismatch. `generate-docs-skill.sh` copies the `0.8.5` source into the skill, producing the CI job's exact one-line diff.

So a rerun alone will remain red until the generated skill copy is refreshed on `main` or this branch is updated after that base fix. I am not adding the unrelated generated-file change to this focused CLI PR unless maintainers prefer that.

nicoloboschi (approved):

PR #2920, opened 2026-07-23
fix backup restore schema compatibility

### LINKED ISSUE #2918: hindsight-admin restore fails when backup source has legacy extra columns
## Bug

`hindsight-admin backup` writes each table as a binary PostgreSQL COPY stream, but manifest v1 records only row counts and byte sizes. It does not record the source column list or schema revision.

That makes a valid backup unrestorable when the source database has historical schema drift. I hit this with a database that still had the legacy `documents.metadata` column while Alembic was already stamped at the current head.

## Reproduction

1. Start with an upgraded database where `public.documents` has 12 columns, including the old `metadata jsonb` column.
2. Create a backup with `hindsight-admin backup`.
3. Initialize a fresh v0.8.5 database. Its `public.documents` table has 11 columns.
4. Run `hindsight-admin restore` against the fresh database.

Restore fails on `documents` with:

```text
BadCopyFileFormatError: row field count is 12, expected 11
```

I also reproduced the source shape difference while both databases reported the same pre-upgrade Alembic head, so checking the migration head alone does not catch it.

## Why this matters

The backup command succeeds and produces a healthy zip, but operators only discover that the artifact is not restorable during recovery. The same mismatch can happen across versions or between a long-lived upgraded database and a fresh database on the same version.

The current `main` implementation still uses manifest version 1 and calls `copy_from_table()` / `copy_to_table()` without a stored column list.

## Expected behavior

A successful backup should either:

- restore into a supported target schema; or
- fail before truncating/importing with a clear schema compatibility error.

## Possible direction

Store each table's ordered source column names/types and relevant schema revision in the manifest, then pass an explicit compatible column list during restore. A preflight should validate every table before the restore transaction starts. Bumping the manifest version would also let older artifacts fail clearly instead of reaching a binary COPY error.

oldnicke:
## Summary

- Bump the database backup manifest format to version 2.
- Record each table's ordered source column names and PostgreSQL types in the manifest.
- Validate all source columns against the target schema before truncating any existing data.
- Restore binary COPY streams using the explicit source column list.
- Return a clear compatibility error when the target is missing a source column or has an incompatible column type.

## Why

Binary PostgreSQL COPY streams do not contain column identities. Previously, backups only recorded row counts and byte sizes, so a backup created from a database with legacy schema drift could fail during restore with an opaque field-count error.

The new preflight detects incompatible schemas before destructive restore operations begin, ensuring the target database remains unchanged when validation fails.

## Testing

- `uv run pytest tests/test_admin_backup_restore.py -q`
- `./scripts/hooks/lint.sh`
- Pre-commit hooks

Added a regression test covering a backup whose `documents` table contains a legacy extra column. The test verifies that restore fails before truncating the target database.

Fixes #2918

### DIFF
--- a/hindsight-api-slim/hindsight_api/admin/cli.py
+++ b/hindsight-api-slim/hindsight_api/admin/cli.py
@@ -9,6 +9,7 @@ import io
 import json
 import logging
 import zipfile
+from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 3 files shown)


koriyoshi2041 (commented):
The schema preflight is a useful fix, but the manifest version bump appears to make every backup created by current releases unrestorable after this lands. `_restore` rejects anything other than version 2 before entering the compatibility path, so a v0.8.5/version-1 backup fails with `Unsupported backup version: 1` even when its source and target schemas match.

Is that compatibility break intentional? If not, could restore keep the existing positional COPY path for version 1 (with the current transactional rollback guarantee) while using column-aware preflight for version 2? At minimum, this seems worth a test that fixes the expected behavior for an existing version-1 archive before the release boundary moves.

PR #2927, opened 2026-07-23
feat(codex): resolve git worktrees to main repo basename for project bank IDs

vcolombo:
## Problem

The Claude Code integration resolves git worktrees to the main repository basename by default (`resolveWorktrees: true`), so all worktrees of a repo share one project bank. The Codex integration derives the project field from the literal `cwd` basename (`os.path.basename(cwd)`).

When both integrations share per-project banks (e.g. `dynamicBankGranularity: ["project"]`), a Codex session in a worktree lands in a separate orphan bank:

| cwd | Claude Code bank | Codex bank (before) |
|---|---|---|
| `~/git/myrepo` | `myrepo` | `myrepo` |
| `~/git/myrepo-wt1` (worktree) | `myrepo` | `myrepo-wt1` ❌ |

This fragments memory across short-lived branches and breaks cross-tool bank sharing.

## Change

- `lib/bank.py`: derive the project name via `git rev-parse --git-common-dir` (5s timeout), resolving worktrees to the main repo basename. Falls back to the literal basename outside git repos or on any git failure.
- Honors the same `resolveWorktrees` config knob as the Claude Code integration (default `true`; set `false` to restore literal-basename behavior).
- Tests: real repo + worktree fixture covering worktree resolution, main checkout, opt-out, non-git dir, and nonexistent cwd fallback.

## Testing

```
$ uvx --with pytest pytest hindsight-integrations/codex/tests/ -q
102 passed
```

### DIFF
--- a/hindsight-integrations/codex/scripts/lib/bank.py
+++ b/hindsight-integrations/codex/scripts/lib/bank.py
@@ -19,28 +19,44 @@ from .state import read_state, write_state
 DEFAULT_BANK_NAME = "codex"
 
 
-def _project_name(cwd: str, config: dict) -> str:
-    """Project basename; resolves git worktrees to the main repo basename.
+def _resolve_project_name(cwd: str, config: dict) -> str:
+    """Resolve the project name from the working directory.
 
-    Mirrors the Claude Code plugin's resolveWorktrees behavior so both tools
-    derive the same per-project bank ID. Set "resolveWorktrees": false to use
-    the literal directory basename instead.
+    When resolveWorktrees is enabled (default), detects git worktrees and
+    resolves to the main repository basename so that all worktrees of the
+    same repo share the same bank.
+
+    For a regular repo at /home/user/myproject:
+        git-common-dir → /home/user/myproject/.git → basename "myproject"
+
+    For a worktree at /home/user/myproject-wt1 linked to /home/user/myproject:
+        git-common-dir → /home/user/myproject/.git → basename "myproject"
     """
     if not cwd:
         return "unknown"
-    if config.get("resolveWorktrees", True):
-        try:
-            out = subprocess.run(
-                ["git", "-C", cwd, "rev-parse", "--git-common-dir"],
-                capture_output=True, text=True, timeout=2,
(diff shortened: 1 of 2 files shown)


vcolombo:
Addressed review:

- Replaced the helper with the established `_resolve_project_name()` from the claude-code integration, verbatim — `--path-format=absolute`, `timeout=5`, narrow `except (OSError, subprocess.TimeoutExpired)`. The two integrations now share identical behavior by construction.
- Tests use `os.devnull` instead of hard-coded `/dev/null`.

Left the fixture's 2-tuple return as-is — it matches the existing test style in this suite.

Full suite: 102 passed.

vcolombo:
Re-review addressed:

- **Timeout mismatch**: the PR description was stale — it described my original 2s probe; the code now uses the established implementation's 5s (matching the claude-code integration). Description updated to say 5s; code unchanged, since parity with the established helper is the point of this PR.
- **Lambda assignment**: replaced with a local `def` in the test fixture.
- **Raw dict config access**: declining — this integration's config layer is a plain dict by design (`load_config() -> dict`), and `bank.py` already accesses it via `config.get()` throughout, including the claude-code implementation this PR mirrors. A typed-config refactor would be a separate change across the whole plugin.
- **Tuple-return fixture**: declining as before — matches the existing style of this test suite.

102 tests pass.

nicoloboschi:
Closing: superseded by **Coding Agents**, where worktree-aware project naming is the default behaviour.

`{gitProject}` resolves through `git rev-parse --path-format=absolute --git-common-dir`, so every linked worktree of a repo resolves to the **main** worktree's basename and shares one bank (bare repos use the bare dir name; non-git directories fall back to the dir basename). It's on by default and switchable with `resolveWorktrees: false`, and the same resolution is used for document ids so all worktrees write one `gitlog:` document instead of one each.

`npx @vectorize-io/hindsight-coding-agents install codex` ([docs](https://hindsight.vectorize.io/sdks/integrations/coding-agents)). Thanks — splitting memory per worktree is a quiet way to lose half your context.

PR #2930, opened 2026-07-23
fix(retain): stop append replay when an oversized doc's first transport slice extracts zero facts

Korayem:
> This pull request was prepared by an AI agent under direct user supervision.

## Why this is operationally critical

This is not only a retain optimization. It is currently a release blocker for a production chat-memory workload that stores each conversation as one canonical session document and periodically replaces that document after new messages arrive.

For the affected oversized-document cohort, appending one new chat turn can silently abandon Delta Retain and send **every historical transport slice** back through fact extraction. In the production-shaped canary below, a tail-only update went from the expected **3,164 input tokens on 1/4 slices** to **140,438 input tokens across 4/4 slices**—about **44× more input** and effectively another full ingestion of the conversation.

That makes cost grow with the complete lifetime transcript instead of the newly appended tail. Long, active sessions can pay this cost repeatedly on ordinary sync runs. Re-extracting historical slices can also churn derived facts, chunks, entities, and associations even though the canonical source text remains correct.

The failure is conditional, not universal: it affects strict append-only oversized documents whose first transport slice previously produced zero stored facts/chunks. However, callers cannot reliably identify that cohort through the public retain response before paying the extraction cost. Our rollout therefore has to stop before staging reconstruction unless this path is fixed or every target is independently proven unaffected.

## Summary

Append-only oversized documents no longer replay their full history when a header-only first transport slice previously produced zero facts. That edge left no stored chunk at index 0, so every later append appeared to have no unchanged chunks and fell back to full extraction.

The fix compares the complete sanitized/redacted incoming document with the stored source. A strict append advances document metadata under the observed content-hash precondition, preserving historical chunks so subsequent transport slices can use the existing recovery path. Non-append replacements, missing stored source text, and concurrent version changes retain the full-streaming fallback.

This is intentionally narrower than the approach in #2728: it preserves the current transport/recovery design and covers the repeated head-slice edge exposed by a real chat-transcript workload.

## Cost evidence

| Probe | Before | After |
|---|---:|---:|
| Deterministic zero-fact-header regression | 22/22 slices extracted | 1/22 slices extracted |
| Production-shaped local canary, real LLM | 140,438 input tokens across 4/4 slices | 3,164 input + 404 output tokens on 1/4 slices |

The deterministic regression isolates the engine behavior. The real-LLM canary corroborates the cost shape under the v0.8.5 production chunking configuration; exact token totals remain provider-response dependent.

Related: #2728

## Validation

- `./scripts/hooks/lint.sh`
- Ruff and `ty` on the changed engine and test files
- `24 passed` across `test_large_document_replacement.py` and `test_delta_retain.py`
- Production-shaped local PGlite canary with isolated development bank, cleanup verification, and no staging/production data access

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -71,6 +71,25 @@ def _redact_document_body(body: str, config: Any) -> str:
     return apply_redaction(body).content
 
 
+def _is_strict_append_of_stored_document(
+    stored_original_text: str | None,
+    document_body_override: str | None,
+    config: Any,
+) -> bool:
+    """Return whether an oversized document body strictly appends stored text.
+
+    ``documents.original_text`` is sanitized and may also be Memory Defense
+    redacted before persistence. Apply those same transformations to the
+    complete incoming body before comparing it with the stored prefix.
+    """
+    if stored_original_text is None or document_body_override is None:
+        return False
+
+    redacted_body = _redact_document_body(document_body_override, config)
+    sanitized_body = fact_extraction._sanitize_text(redacted_body) or ""
+    return len(sanitized_body) > len(stored_original_text) and sanitized_body.startswith(stored_original_text)
+
+
 async def _fire_memory_defense_webhook(
     webhook_manager: Any,
     *,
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Approving — I independently reproduced the bug and verified the fix.

**Reproduction (deterministic, against `main` without this fix):** ran the added `test_append_after_zero_fact_header_slice_skips_unchanged_history` — appending one turn re-extracted **22 slices**:

```
assert 22 <= 2  →  "append-only replacement replayed more than two unchanged/edge slices"
```

The trigger is exactly as described: the document's first transport sub-batch is header-only and extracts **zero facts**, so no chunk is persisted at index 0 (`RECOVERY: found 20 already-committed chunks`, not 22). A later strict append then finds no unchanged chunk and falls back to full-history re-extraction. This is why it was hard to reproduce ad hoc — it only fires when the leading slice is pure header/metadata.

**With the fix:** the strict-append path takes the metadata-only branch and preserves historical chunks. Verified locally (CI's `test-api` matrix is skipped on fork PRs, so these ran on my machine):
- `test_delta_retain.py` + `test_large_document_replacement.py` — **24 passed**
- append/oversized suite (`-k "append or oversized"`) — **37 passed**
- `ruff` + `ty` clean on `orchestrator.py`

The `expected_content_hash` precondition on `_delta_metadata_only` correctly forces a full-retain fallback on a concurrent version change, so the metadata-only shortcut stays safe under concurrent writes.

Title updated to name the narrow edge condition.

PR #2938, opened 2026-07-24
fix(retain): make async retries idempotent

### LINKED ISSUE #2937: Retain retries can duplicate async operations after a lost acknowledgement
## Bug Description

The async Retain API has no caller-supplied request identity. If Hindsight accepts a Retain request but the response is lost or times out, an integration cannot determine whether the operation was created. Retrying the same semantic request creates a second async operation and can repeat extraction, embeddings, relationships, consolidation work, and provider spend.

The Codex integration currently has the same ambiguity: hook processes are
ephemeral and do not persist the exact unacknowledged request before POSTing
it. Waiting for one operation to finish before sending the next is also not a
complete solution: the final `Stop` event can occur while the predecessor is
still processing, leaving the final window queued locally with no later hook
event to submit it.

This is request-level idempotency. The existing chunk-insert idempotency work in #986 protects an internal persistence stage; it does not give API callers an operation identity across a lost HTTP acknowledgement.

## Steps to Reproduce

1. Submit an async Retain request.
2. Allow the server to create the parent operation.
3. Drop or time out the HTTP response before the client receives the operation ID.
4. Retry the same request.
5. Observe that a second parent operation is created.

## Expected Behavior

- Callers may supply an optional idempotency key for async Retain.
- Reusing the key with the same canonical payload returns the original operation ID and creates no new work.
- Reusing the key with a different payload fails explicitly with HTTP 409.
- Authorization and operation validation run on every retry.
- Existing callers that omit the key keep the current create-each-time behavior.
- Integrations persist the exact pending request before POST and clear it only after acknowledgement.
- Distinct keyed upserts for the same bank/document are accepted immediately
  but processed in order; different documents remain parallel.

## Actual Behavior

There is no request-level identity spanning the API and integrations. Network ambiguity can therefore enqueue duplicate operations even when the caller retries the exact same work.

## Version

Reproduced and fixed against upstream commit `4dc8348348f3e0b50da9071357955a2995c4f0e0`.

## Proposed Fix

One monorepo change covering the API and Codex integration:

1. Add optional `idempotency_key` to async Retain requests.
2. Canonicalize and hash the caller-submitted operation-defining payload.
3. Persist the key and fingerprint on the parent async operation.
4. Enforce uniqueness on `(bank_id, operation_type, idempotency_key)`.
5. Return the original operation on same-key/same-payload retries.
6. Return HTTP 409 on same-key/different-payload reuse.
7. Keep current behavior when no key is supplied.
8. Re-run authorization and validation on every retry without allowing
   validator enrichment to change the caller's idempotency identity.
9. Serialize distinct keyed single-document upserts behind their latest
   non-terminal predecessor.
10. In the Codex hook, serialize submissions per session, write pending
    transcript state durably with `0700/0600` permissions, retry the exact
    request first, preserve one exact due window when the bounded queue is
    full, and submit later acknowledged windows immediately.
11. Require both server feature flags before the plugin drains its queue.
12. Document and publish the contract in the OpenAPI surface.

## Regression Proof

Prepared patch proof:

- 24 PostgreSQL-backed core/API/migration tests passed, including concurrent
  identical requests, concurrent distinct requests for one document, ordered
  chains, parallel different documents, conflicts, authorization, release, and
  legacy no-key behavior.
- 133 Codex integration/state/client tests passed, including fail-closed
  compatibility when either required server feature is absent.
- 157 combined focused tests passed in the structured review gate.
- The version endpoint and a real async Retain integration path passed.
- OpenAPI was regenerated with the repository's official generator.
- Ruff and diff checks passed.
- Oracle remains query-shape-proven by two rewrite tests only; no live Oracle
  runtime was available.

TheAngryPit:
# Fix async Retain replay idempotency across API and Codex

Fixes #2937

## Summary

Adds optional request-level idempotency to async Retain, serializes distinct
keyed upserts for the same document, and makes the Codex integration retry lost
acknowledgements without duplicating work or stranding the final due window.

## API and engine

- Accept optional `idempotency_key` for async Retain.
- Fingerprint the canonical caller-submitted operation payload.
- Store the key/fingerprint on the parent async operation.
- Use a database uniqueness constraint as the concurrency authority.
- Return the original operation for same-key/same-payload retries.
- Return HTTP 409 for same-key/different-payload reuse.
- Preserve create-each-time behavior when the key is absent.
- Re-run authentication and operation validation on every retry.
- Serialize keyed single-document upserts behind the latest non-terminal
  predecessor while leaving different documents parallel.

## Codex integration

- Persist the exact request and identity before POST.
- Fail closed when pending state cannot be written.
- Store transcript-bearing pending files as owner-only (`0700` directory, `0600` files).
- Serialize Retain state per Codex session while preserving the global turn-counter lock.
- Retry an unacknowledged request before submitting later due work.
- Preserve one exact due window in a durable overflow slot when the normal
  queue reaches its bounded capacity.
- Submit every acknowledged queued window immediately; server-side ordering
  removes the need for client polling or a later hook event.
- Require the server to advertise both request idempotency and serialized
  upsert support before draining queued work.
- Remove pending state only after an acknowledgement.

## Docs

- Document the Retain idempotency contract.
- Update Codex integration behavior.
- Regenerate the OpenAPI schema.

## Compatibility

- The field is optional.
- Synchronous Retain is unchanged.
- Existing integrations that omit the key are unchanged.
- This complements rather than replaces internal chunk-insert idempotency (#986).

## Proof

- `24 passed`: PostgreSQL-backed engine/API, concurrency, HTTP, and migration tests.
- `133 passed`: Codex hooks, state, and client tests.
- `157 passed`: combined focused tests in the structured review gate.
- Version endpoint and real async Retain integration tests passed.
- OpenAPI was regenerated with the official repository generator.
- Ruff and diff checks passed.
- Oracle adapter query shape was checked by two focused rewrite tests; live
  Oracle runtime proof remains unavailable.
- Structured Codex Auto Review completed. Its sole remaining finding claimed
  that the Oracle lock rewrite was absent; source inspection and both passing
  rewrite tests prove that upstream `engine/db/oracle.py` already performs the
  required `FOR NO KEY UPDATE` to `FOR UPDATE` conversion.

### DIFF
--- a/hindsight-api-slim/hindsight_api/alembic/versions/e8f1a2b3c4d5_add_retain_idempotency.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/e8f1a2b3c4d5_add_retain_idempotency.py
@@ -1,0 +1,60 @@
+"""Add optional async retain idempotency identity.
+
+Revision ID: e8f1a2b3c4d5
+Revises: d7b2f8a1c934
+Create Date: 2026-07-23
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+from hindsight_api.alembic._dialect import run_for_dialect
+
+revision: str = "e8f1a2b3c4d5"
+down_revision: str | Sequence[str] | None = "d7b2f8a1c934"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _pg_upgrade() -> None:
+    op.add_column("async_operations", sa.Column("idempotency_key", sa.Text(), nullable=True))
+    op.add_column("async_operations", sa.Column("request_fingerprint", sa.String(length=64), nullable=True))
+    op.create_unique_constraint(
+        "uq_async_operations_idempotency",
+        "async_operations",
+        ["bank_id", "operation_type", "idempotency_key"],
+    )
+
+
+def _pg_downgrade() -> None:
+    op.drop_constraint("uq_async_operations_idempotency", "async_operations", type_="unique")
+    op.drop_column("async_operations", "request_fingerprint")
+    op.drop_column("async_operations", "idempotency_key")
+
+
+def _oracle_upgrade() -> None:
+    # Oracle defaults VARCHAR2 length semantics to bytes. The API accepts 256
+    # Unicode characters, so declare character semantics explicitly.
+    op.execute("ALTER TABLE async_operations ADD idempotency_key VARCHAR2(256 CHAR)")
+    op.add_column("async_operations", sa.Column("request_fingerprint", sa.String(length=64), nullable=True))
+    op.create_unique_constraint(
+        "uq_async_operations_idempotency",
+        "async_operations",
+        ["bank_id", "operation_type", "idempotency_key"],
+    )
+
+
+def _oracle_downgrade() -> None:
+    op.drop_constraint("uq_async_operations_idempotency", "async_operations", type_="unique")
+    op.drop_column("async_operations", "request_fingerprint")
+    op.drop_column("async_operations", "idempotency_key")
+
+
+def upgrade() -> None:
+    run_for_dialect(pg=_pg_upgrade, oracle=_oracle_upgrade)
+
+
+def downgrade() -> None:
+    run_for_dialect(pg=_pg_downgrade, oracle=_oracle_downgrade)
--- a/hindsight-api-slim/hindsight_api/alembic/versions/f9a2b3c4d5e6_add_retain_serialization.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/f9a2b3c4d5e6_add_retain_serialization.py
@@ -1,0 +1,88 @@
+"""Serialize idempotent async retains for the same document.
+
+Revision ID: f9a2b3c4d5e6
+Revises: e8f1a2b3c4d5
+Create Date: 2026-07-24
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+from hindsight_api.alembic._dialect import run_for_dialect
+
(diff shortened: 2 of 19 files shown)


nicoloboschi:
Thanks a lot for the thorough work here, @TheAngryPit — the problem in #2937 is real and your regression proof is impressively complete.

After reviewing it, we're going to land a deliberately smaller fix for the core bug in #2947 and close this one. The reasoning, in case it's useful:

**Reuse the primary key instead of adding an identity column.** `async_operations.operation_id` is already the table's `PRIMARY KEY`. If the caller supplies that UUID directly, we get idempotency for free:
- Re-submitting with the same `operation_id` returns the original operation (a lost-ack retry becomes a no-op).
- The primary key *is* the concurrency authority — a plain `INSERT` plus unique-violation handling gives exactly-once with no new constraint.
- Reusing an id owned by a different bank/operation → HTTP 409.

That removes the need for the two migrations, the canonical-payload fingerprint/hash, and the `idempotency_key` + `request_fingerprint` columns entirely — **no schema change at all**. Auth and the operation validator still run on every retry, so a replay can't bypass access/credit checks.

**Scope.** We're keeping the PR to just the filed symptom (duplicate async operations after a lost acknowledgement). The two larger pieces here — server-side same-document upsert **serialization** (`serialization_key`/`blocked_by` chains, `DeferOperation`, cancel/retry restrictions) and the **Codex durable-queue** rework — are separable concerns. Concurrent same-document chunk writes are already convergent via `ON CONFLICT (chunk_id)` (#986/#1916), so if there's a residual ordering hazard we'd rather see it as its own issue with a concrete repro than couple it to the idempotency fix.

The dropped bit compared to your version is the same-key/**different**-payload 409 (we replay by id rather than fingerprinting the payload). Reusing a UUID for different content is a client bug, and idempotent replay is a safe answer; a `request_fingerprint` column can add that check later if we decide we want it.

Really appreciate the effort — the framing in #2937 made the alternative easy to reason about. Closing in favor of #2947.

PR #2958, opened 2026-07-25
fix: return free-form entities from dry-run extraction

### LINKED ISSUE #2956: Dry-Run allow free form entities not generating entities.
### Bug Description

No entities are returned for dry-run returned facts, regardless of the setting in "entities_allow_free_form".

### Steps to Reproduce

1. call dry run with entities_allow_free_form true.
2. inspect the returned facts array, entities array is []

### Expected Behavior

The entities array is expected to be populated with similar entities that a regular retain would produce.

### Actual Behavior

Sending the entities_allow_free_form flag did not return any entities.
The memory bank was set to entities_allow_free_form true.
I tried with the memory bank entities_allow_free_form false in case there was some selection conflict, but same outcome.

### Version

0.8.5

### LLM Provider

Ollama

oldnicke:
## Summary

- preserve validated entity strings when mapping dry-run extraction results
- handle extraction results with no entities as an empty list
- add an HTTP regression test for `entities_allow_free_form=true`

## Root cause

`extract_facts_from_text()` returns entity names as strings, but the dry-run response mapper treated them as objects with a `.text` attribute. Every string was therefore filtered out.

## Testing

- `HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 ... pytest tests/test_extract_dry_run_http.py -q -n 0` (`8 passed`)
- `ruff check` and `ruff format --check` on changed files
- `ty check --python /tmp/hindsight-forgetting/.venv/bin/python hindsight_api/`
- `git diff --check`

The repository-wide `./scripts/hooks/lint.sh` was attempted twice, but its initial `uv sync --frozen` remained blocked in LiteLLM 1.93.0 Rust `cargo metadata`; the relevant Python lint, format, type, and regression checks above completed successfully.

Closes #2956

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -7929,7 +7929,7 @@ class MemoryEngine(MemoryEngineInterface):
                 fact_type=fact.fact_type,
                 occurred_start=fact.occurred_start,
                 occurred_end=fact.occurred_end,
-                entities=[e.text for e in (fact.entities or []) if getattr(e, "text", None)],
+                entities=list(fact.entities or []),
             )
             for fact in facts
         ]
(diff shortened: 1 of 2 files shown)


nicoloboschi (commented):
nice catch!

PR #2960, opened 2026-07-25
fix(refresh): preserve mental model content when reflect returns no usable answer

### LINKED ISSUE #2959: refresh_mental_model overwrites a healthy mental model with the "No answer provided." placeholder when reflect fails
## Bug Description

When the reflect agent finishes without a usable answer, `_process_done_tool` substitutes the
human-readable placeholder `NO_ANSWER_TEXT` ("No answer provided."). `refresh_mental_model` has a
guard whose stated purpose is to "refuse to overwrite existing content with an empty render", but it
tests `if not final_content.strip():` — emptiness only. The placeholder is non-empty, so it passes
the guard and is written over the existing document. The refresh then returns normally, so the
operation is recorded as **completed with no error**.

The result is silent data loss: a mental model that took many refreshes to build is replaced by a
19-character string, and nothing in the operation log indicates a failure.

The same applies to the iteration-limit fallback
("I was unable to formulate a complete answer within the iteration limit."), which is also non-empty
and also reaches persistence.

### Relationship to #2894 (related, but a different path)

#2894 reports the same *symptom* — a non-empty placeholder defeating the `not final_content.strip()`
guard — but a different cause, and the two proposed fixes do not cover each other:

|  | #2894 | this issue |
|---|---|---|
| `based_on` | empty (recall returned 0 results) | **full** — 133 and 266 items in the affected models |
| origin of the text | the LLM wrote prose ("I don't have information") | the **code** substituted `NO_ANSWER_TEXT` after a blank `done.answer` |
| trigger | embedding index unavailable during consolidation | reflect finished without a usable answer |
| proposed guard | `total_facts == 0` | reflect reported no usable answer |

A `total_facts == 0` check would not have caught our case: retrieval worked, and the evidence was
substantial. Conversely, a no-usable-answer check does not catch #2894, where the agent did return an
answer — just a worthless one. Both are real, and the guards compose; they are recorded separately so
neither is mistaken for a fix of the other.

Worth noting that both are instances of the same shape: a guard keyed on emptiness while the failure
arrives as non-empty text. Rather than adding a second special case, the fix below records *whether
reflect produced a usable answer* at the point where that is known, so no future placeholder needs
its own check.

We hit this on a self-hosted deployment: five mental models across two banks were reduced to the
placeholder, each with a `completed` operation and no error message. Because the write reports
success, there is no failure signal to alert on — the loss is only visible by reading the content.

## Steps to Reproduce

1. Create a mental model and refresh it successfully so it holds real content.
2. Cause the reflect agent to finish without a usable answer. Either:
   - call the `done` tool with an empty/blank `answer` argument, or
   - let the agent exhaust `max_iterations` without calling `done`.
   (In production this happened on large syntheses under a `max_tokens` of 4096; the trigger is
   environment-specific, but any path that yields a blank `done.answer` reproduces it.)
3. Run `refresh_mental_model` on that model.
4. Read the model back.

Minimal reproduction against the engine, patching only `reflect_async` so the real persistence path
runs:

```python
from hindsight_api.engine.reflect.agent import NO_ANSWER_TEXT
from hindsight_api.engine.response_models import ReflectResult

async def fake_reflect_async(**kwargs):
    return ReflectResult.model_validate({
        "text": NO_ANSWER_TEXT,               # what the agent emits on a blank done.answer
        "based_on": {"observation": [{"id": "f1", "text": "a fact", "type": "observation"}]},
    })

memory.reflect_async = fake_reflect_async
await memory.refresh_mental_model(bank_id=bank_id, mental_model_id=mm_id, request_context=ctx)

model = await memory.get_mental_model(bank_id, mm_id, request_context=ctx)
print(repr(model["content"]))   # 'No answer provided.'  — the previous document is gone
```

## Expected Behavior

A refresh that produced no usable answer should leave the existing content untouched and surface the
failure, exactly as the existing guard's comment describes ("Writing "" to the DB would destroy the
working document ... preserve existing content in the DB (and audit the failure via
reflect_response), then RAISE so the caller knows the refresh didn't happen").

## Actual Behavior

The placeholder is persisted over the existing content and the operation completes successfully:

- `content` becomes `"No answer provided."` (19 chars)
- `reflect_response.text` is the same placeholder
- `structured_content` becomes a single "Overview" section containing that string
- the async operation ends in `completed`, `error_message: null`, `retry_count: 0`

On v0.8.5 and current `main`, `populated_content` in the refresh outcome metadata does correctly
recognize the placeholder — but it is computed in `_write_refresh_outcome_metadata` *after*
`refresh_mental_model` has already persisted and returned. It annotates the loss rather than
preventing it, and it does not cover the iteration-limit text.

## Version

Observed on 0.8.3. Verified by source inspection that the same composition is present in
0.8.4, 0.8.5, and current `main`: both placeholder strings exist and the persist guard is still
`if not final_content.strip():` in all four refs. Upgrading does not close it.

## LLM Provider

Anthropic (via a LiteLLM gateway). The defect is provider-independent — it is in how a failed render
is persisted, not in what any provider returns.

---

## Additional Context

A fix is in the PR linked below. Its shape:

- a required `answer_failure_reason` field on `ReflectAgentResult` (`"empty_answer" | "iteration_limit" | None`),
  set where each terminal path already knows whether it produced an answer, so nothing has to
  pattern-match on placeholder text;
- a check for it in `refresh_mental_model` before delta rendering — delta re-renders the document and
  can turn the placeholder candidate into ordinary-looking markdown, so a guard placed after it would
  inspect normal-looking content and pass;
- the existing text guard retained as defense in depth, keyed on a shared `is_populated_content()`
  predicate that the outcome metadata also uses, so what is stored and what is reported cannot drift.

Happy to adjust the approach if maintainers would prefer a different shape.

gvilleg:
## Summary

Fixes #2959.

When the reflect agent finishes without a usable answer it substitutes a human-readable placeholder
(`NO_ANSWER_TEXT`, or the iteration-limit text). Both are non-empty, so `refresh_mental_model`'s
"refuse to overwrite with an empty render" guard — which tests `not final_content.strip()` — does not
fire, and the placeholder is written over the existing document while the operation reports success.

This makes the refresh fail closed: when reflect produced nothing usable, the existing content stays,
the failure is auditable, and `MentalModelRefreshError` is raised.

## Changes

**A required failure reason at the source** (`engine/reflect/models.py`, `engine/reflect/agent.py`)

`ReflectAgentResult.answer_failure_reason: Literal["empty_answer", "iteration_limit"] | None` — a
required field with no default, supplied at all seven terminal construction sites. Each site already
knows whether it produced an answer, so the signal is recorded where the fact is known rather than
inferred later from text. `_process_done_tool` computes it *before* substituting the placeholder.

No default is deliberate: a future terminal path that forgets the field raises `ValidationError`
rather than silently reporting success. Placeholder strings moved to `reflect/models.py` (with
`agent.py` re-exporting `NO_ANSWER_TEXT` for existing importers) so persistence code can import them
without pulling in the agent module.

**Check it before delta** (`engine/memory_engine.py`)

`refresh_mental_model` now inspects `answer_failure_reason` immediately after reflect and before any
delta work. Placement matters: delta applies operations to the stored document and re-renders it, so
a guard after that step would inspect ordinary-looking markdown and pass. On failure it persists the
`reflect_response` — including the full `based_on` evidence and accumulated delta provenance — with
`refresh_skipped` set to the specific reason, leaves `content`/`structured_content` untouched, and
raises.

**One predicate for storing and reporting** (`engine/memory_engine.py`)

New pure helper `is_populated_content(content)` rejects empty/whitespace, `MENTAL_MODEL_PENDING_CONTENT`,
`NO_ANSWER_TEXT`, and `ITERATION_LIMIT_TEXT`. The text-level guard now uses it instead of a bare
emptiness test, and `_write_refresh_outcome_metadata` uses the same helper, so what gets stored and
what gets reported cannot drift. (This also gives the outcome metadata iteration-limit coverage it
did not previously have.) The authoritative signal is `answer_failure_reason`; this is defense in
depth.

**Failed refreshes no longer advance query tracking** (`engine/memory_engine.py`)

Both failure paths previously passed `last_refreshed_source_query=current_source_query` while
deliberately preserving old content. Since `use_delta` requires the stored query to match the current
one, that combination corrupts the retry: after a source-query change, a failed full regeneration
would record the new query, and the retry would then select delta and amend the *previous* topic's
document from recall scoped to new memories only. Failure paths now persist only the audit payload.
The success path and the delta `no_new_facts` path still advance tracking, since in both the stored
content really is current for that query.

## Behavior change

A refresh that produces no usable answer now raises `MentalModelRefreshError` instead of completing.
`MentalModelRefreshError` is already classified retryable and converts to `RetryTaskAt`, so async
refreshes retry as before and settle terminal only after exhaustion. Retry is safe precisely because
each attempt is now non-destructive — the risk was never the retry, it was the overwrite.

## Tests

New: `hindsight-api-slim/tests/test_refresh_preserves_on_failed_answer.py` (8 cases). They patch
`reflect_async` rather than `refresh_mental_model`, so the real persistence path — delta, guards, the
write — actually executes; stubbing the refresh itself would assert nothing about storage.

- blank answer preserves content, raises, records `refresh_skipped == "empty_answer"`
- iteration-limit preserves content, raises, records `refresh_skipped == "iteration_limit"`
- a failed refresh preserves its `based_on` evidence
- delta mode cannot launder a failed answer: asserts delta was actually selected (`created_after`
  present) **and** that no delta operation was emitted
- a failed refresh does not advance source-query tracking, so its retry is still a full regeneration
- a successful refresh still writes content
- `is_populated_content` unit coverage for all three placeholders

Each was mutation-checked against the pre-fix behavior: reverting the guards fails 5 of 7; disabling
only the early check still fails 3 (the text guard catches the rest); restoring
`last_refreshed_source_query` on a failure write fails the tracking test; disabling the early check
fails the delta test with real laundering — delta runs, renders ordinary prose, and no error is
raised. A test that passes with and without the fix would prove nothing, so this was verified rather
than assumed.

`tests/test_refresh_outcome_metadata.py` passes unchanged: its no-answer case stubs
`refresh_mental_model` entirely, so it exercises only the metadata writer, where `populated_content=False`
remains correct.

```
tests/test_refresh_preserves_on_failed_answer.py  tests/test_mental_model_delta.py  tests/test_refresh_outcome_metadata.py
21 passed, 4 skipped
```

`scripts/hooks/lint.sh` reports no diagnostics in any changed file.

## Notes for reviewers

- The trigger — why a large synthesis returns a blank `done.answer` in the first place — is
  environment-specific and not addressed here. This change is about what happens *when* it does:
  a failed render must not destroy a working document.
- Sequencing is the crux of the fix. `populated_content` on `main` already recognizes the
  placeholder, but it is computed after `refresh_mental_model` has persisted and returned, so it can
  only annotate the loss. The check has to happen before delta and before the write.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -11807,11 +11807,15 @@ class MemoryEngine(MemoryEngineInterface):
                     reflect_result.answer_failure_reason,
                 )
                 reflect_response_payload["refresh_skipped"] = reflect_result.answer_failure_reason
+                # Record the failure, but do not advance last_refreshed_source_query:
+                # nothing was synthesized for the current query. Advancing it would tell
+                # the next attempt the topic is unchanged, so a retry after a
+                # source-query change would run in delta mode against content written
+                # for the previous topic, with recall scoped to new memories only.
                 await self.update_mental_model(
                     bank_id,
                     mental_model_id,
                     reflect_response=reflect_response_payload,
-                    last_refreshed_source_query=current_source_query,
                     request_context=request_context,
                 )
                 raise MentalModelRefreshError(
@@ -11953,13 +11957,15 @@ class MemoryEngine(MemoryEngineInterface):
                     "preserving previous content and raising MentalModelRefreshError."
                 )
                 reflect_response_payload["refresh_skipped"] = "empty_candidate"
-                # Persist the reflect_response (so the failure is auditable) and
-                # the source-query tracking, but do NOT touch content/structured.
+                # Persist the reflect_response so the failure is auditable, but do not
+                # touch content/structured_content or last_refreshed_source_query: no
+                # content was produced for the current query, and recording it as
+                # refreshed would put a retry into delta mode against the previous
+                # topic's document.
                 await self.update_mental_model(
                     bank_id,
                     mental_model_id,
                     reflect_response=reflect_response_payload,
-                    last_refreshed_source_query=current_source_query,
                     request_context=request_context,
                 )
                 raise MentalModelRefreshError(
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -103,6 +103,27 @@ _bank_template_import_authorization: contextvars.ContextVar[_BankTemplateImportA
 MENTAL_MODEL_PENDING_CONTENT = "Generating content..."
 
 
+def is_populated_content(content: str | None) -> bool:
+    """Whether mental-model content is a real synthesis rather than a placeholder.
+
+    ``MENTAL_MODEL_PENDING_CONTENT`` (not refreshed yet) and the reflect agent's
+    fallback answers are all non-empty, so a length or emptiness check reads them
+    as populated. Both the refresh outcome metadata and the persistence guard use
+    this one predicate, so what gets reported and what gets stored cannot drift
+    apart.
+
+    Note this is the last-resort text check. The authoritative signal is
+    ``ReflectAgentResult.answer_failure_reason``, which is set where the failure
+    happens and does not depend on matching placeholder strings.
+    """
+    stripped = (content or "").strip()
+    return bool(stripped) and stripped not in (
+        MENTAL_MODEL_PENDING_CONTENT,
+        NO_ANSWER_TEXT,
+        ITERATION_LIMIT_TEXT,
+    )
+
+
 def get_current_schema() -> str:
     """Get the current schema from context (falls back to config default)."""
     schema = _current_schema.get()
@@ -392,6 +413,7 @@ from .llm_wrapper import LLMConfig, requires_api_key, sanitize_llm_output, sanit
 from .multi_llm import MultiLLMProvider
 from .query_analyzer import QueryAnalyzer
 from .reflect import run_reflect_agent
+from .reflect.models import ITERATION_LIMIT_TEXT, NO_ANSWER_TEXT
 from .reflect.tools import tool_expand, tool_recall, tool_search_mental_models, tool_search_observations
 from .response_models import (
     VALID_RECALL_FACT_TYPES,
@@ -2585,17 +2607,11 @@ class MemoryEngine(MemoryEngineInterface):
         if not operation_id:
             return
 
-        from .reflect.agent import NO_ANSWER_TEXT
-
         content = refreshed.get("content") or ""
-        stripped = content.strip()
         based_on = (refreshed.get("reflect_response") or {}).get("based_on") or {}
         outcome = RefreshMentalModelOutcomeMetadata(
             content_len=len(content),
-            # The no-answer stub and the pending placeholder complete
-            # wire-successful but carry no real synthesis — a length check
-            # alone would read them as populated.
-            populated_content=bool(stripped) and stripped not in (MENTAL_MODEL_PENDING_CONTENT, NO_ANSWER_TEXT),
+            populated_content=is_populated_content(content),
             based_on_counts={fact_type: len(facts or []) for fact_type, facts in based_on.items()},
         )
         try:
@@ -10352,6 +10368,7 @@ class MemoryEngine(MemoryEngineInterface):
             # Return response (compatible with existing API)
             result = ReflectResult(
                 text=agent_result.text,
+                answer_failure_reason=agent_result.answer_failure_reason,
                 based_on=based_on,
                 structured_output=agent_result.structured_output,
                 usage=usage,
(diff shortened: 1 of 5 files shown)


nicoloboschi:
Superseded by #3761 (merged as f4a93695), which fixes this at the source instead of at the write.

Your diagnosis here was right, and the `answer_failure_reason` idea — record whether reflect produced an answer at the point where that is known, rather than pattern-matching placeholder text downstream — is what shaped the merged fix. #3761 takes that reasoning one step further: if the run knows it produced no answer, there is nothing to hand back, so `ReflectNoAnswerError` is raised at the three terminal paths (blank `done.answer`, empty final synthesis, the iteration-limit fall-through) and both placeholder strings are deleted. Callers that write what reflect returns never reach the write, so the existing content is preserved by construction and the refresh path needs no new guard at all.

Thanks for the report and the detailed repro — the `NO_ANSWER_TEXT` / `not final_content.strip()` composition you laid out is exactly what the fix is built on.

Issue #2953, opened 2026-07-24
LiteLLM provider sends incompatible named tool_choice to Copilot Responses


jtstothard:
## Description

The LiteLLM provider serializes a named `LLMToolChoice` using the Chat Completions shape:

```json
{"type":"function","function":{"name":"recall"}}
```

This is rejected when LiteLLM routes to a Copilot/OpenAI Responses-compatible endpoint, which expects the function name at the top level:

```json
{"type":"function","name":"recall"}
```

The current path is `hindsight_api/engine/providers/litellm_llm.py::LiteLLM.call_with_tools`, where named choices are built with the nested `function.name` shape. The same compatibility issue was fixed for the Codex Responses provider in #734, while #2843 introduced the typed provider tool-choice contract but did not update this LiteLLM serialization path.

## Reproduction

Configure Hindsight to use a Copilot/OpenAI Responses-compatible model through LiteLLM, then trigger a named tool call (for example, the reflect agent selecting `recall`). The request contains `tool_choice.function`, and the endpoint returns an unknown/unsupported parameter error.

## Expected behavior

When the LiteLLM route targets a Responses-compatible endpoint, named tool choice should be serialized as:

```json
{"type":"function","name":"<tool-name>"}
```

Existing Chat Completions-compatible routes should retain their current shape. A focused regression test would help prevent the two wire formats from diverging again.

Related changes: #2843, #734

PR #2962, opened 2026-07-25
fix(litellm): serialize named tool_choice with flat Responses API shape

jtstothard:
## Problem

`LiteLLMLLM.call_with_tools` built the named `tool_choice` with the nested Chat Completions shape and referenced a field that doesn't exist on the `LLMToolChoice` dataclass.

```python
call_kwargs["tool_choice"] = {
    "type": "function",
    "function": {"name": tool_choice.selected_function_name},  # AttributeError — field is 'function_name'
}
```

Two bugs:
1. `selected_function_name` is not an attribute of `LLMToolChoice` (the field is `function_name`) — would raise `AttributeError` on any named-tool reflect iteration.
2. The nested `{"function": {"name": ...}}` shape is rejected by LiteLLM's Responses API path (Copilot/Luna). Reflect routed through a Copilot model via LiteLLM fails end-to-end: LiteLLM raises a validation error before the request reaches the provider, and the reflect run times out / 422s.

## Fix

Serialize named `tool_choice` to the **flat** Responses API shape and reference the correct field:

```python
call_kwargs["tool_choice"] = (
    {"type": "function", "name": tool_choice.function_name}
    if tool_choice.mode is LLMToolChoiceMode.NAMED
    else tool_choice.mode.value
)
```

This matches the shape already used by the Codex provider (`codex_llm.py`).

## Verification

Before the fix, reflect on a ~27k-memory bank routed through Copilot/Luna via LiteLLM never succeeded — only 422 (validation) and 504 (timeout) responses. After applying the fix locally:

- Reflect returns HTTP 200 end-to-end.
- Latency on a 26k-memory bank: ~123s (under the 300s wall-clock limit).

## Scope

Single-file, 3-line change. No new dependencies, no API surface changes.

Refs #2953

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/litellm_llm.py
@@ -437,10 +437,7 @@ class LiteLLMLLM(LLMInterface):
         call_kwargs = self._build_common_kwargs(messages, max_completion_tokens, temperature)
         call_kwargs["tools"] = tools
         call_kwargs["tool_choice"] = (
-            {
-                "type": "function",
-                "function": {"name": tool_choice.selected_function_name},
-            }
+            {"type": "function", "name": tool_choice.function_name}
             if tool_choice.mode is LLMToolChoiceMode.NAMED
             else tool_choice.mode.value
         )


jtstothard:
## Closing this PR — it's based on a false premise

I need to withdraw this. On independent review I got the diagnosis wrong.

**Claim 1 was false.** I asserted `selected_function_name` is "a nonexistent field" that "would raise AttributeError." It isn't — it's a validated `@property` on `LLMToolChoice` (`llm_interface.py:44-48`) that five other providers use correctly (`openai_compatible_llm.py:1197`, `gemini_llm.py:751`, `codex_llm.py:829`, `claude_code_llm.py:509`). The original code was idiomatic.

**Claim 2 is questionable.** The only real change here is the shape: nested `{"function":{"name":...}}` → flat `{"name":...}`. I justified this by pointing at `codex_llm.py`, but Codex talks directly to the Responses API while LiteLLM normalizes provider shapes itself — so matching Codex's shape may be the wrong fix for the LiteLLM path. PR #2957's argument (that this belongs in LiteLLM's Responses bridge) is probably correct.

**What actually fixed reflect on our deployment** was a broader local container patch that changed `call_with_tools`' signature, not this minimal diff. I conflated the two when writing this PR and didn't verify the dataclass before opening it.

Apologies for the noise. I'll re-investigate #2953 properly before submitting anything further.

PR #2964, opened 2026-07-26
fix: avoid dotenv side effects on library import

### LINKED ISSUE #2961: hindsight_api.config rewrites the importing application's os.environ at import time (load_dotenv override=True from an ancestor of cwd)
## Summary

`hindsight_api/config.py` calls, at module scope:

```python
# Load .env file, searching current and parent directories (overrides existing env vars)
load_dotenv(find_dotenv(usecwd=True), override=True)
```

Importing `hindsight_api` (or anything that pulls it in — `import hindsight`,
`hindsight.server`, `HindsightEmbedded`) therefore **rewrites the importing
application's `os.environ`**, from a file chosen by walking up from the process
cwd, with `override=True` beating values the application had already loaded
deliberately.

For an application that embeds Hindsight and has its own configuration in the
environment, this is silent, action-at-a-distance config corruption. It is not
limited to `HINDSIGHT_*` keys: `override=True` replaces **every** key that
happens to appear in whatever `.env` the upward walk lands on.

## Impact we hit

We run two independent bot profiles on one host, each with its own env file:

```
~/.hermes/.env                     <- profile A's env (also the parent dir)
~/.hermes/profiles/quant/.env      <- profile B's env
~/.hermes/hermes-agent/            <- shared checkout, was the gateway cwd
```

Profile B's process loads its own env at startup, then imports Hindsight lazily
on the first turn that uses memory. `find_dotenv(usecwd=True)` walks up from the
checkout, finds **profile A's** `.env`, and `override=True` applies it.

Measured in the live process, that single import replaced **25 variables**:

| var | profile B's value | after `import hindsight` |
|---|---|---|
| `DISCORD_BOT_TOKEN` | B's bot token | **A's bot token** |
| `DISCORD_ALLOWED_CHANNELS` | B's 12 channels | A's 56 channels |
| `MATRIX_USER_ID` / `MATRIX_ACCESS_TOKEN` | B's identity | **A's identity** |

Bot B answered one message and then went permanently deaf — every later message
was rejected by a channel allowlist that now belonged to bot A, and slash
commands reported "not authorized" for the same reason. It looked like a
messaging bug for two days. The only reason it was survivable is that both
identities belonged to us; a library that can swap one tenant's credentials for
another's at import time is a nastier shape than the config bug it looks like.

Timeline from the affected process:

```
09:43:00  inbound message                          <- accepted
09:43:16  Hindsight initialized: local_embedded    <- env rewritten here
09:44:29  response sent
09:44:58  message dropped: channel not in allowlist
```

## Reproduction

```bash
mkdir -p /tmp/repro/sub && cd /tmp/repro
printf 'MY_APP_SECRET=from-the-parent-dotenv\n' > .env
cd sub
MY_APP_SECRET=set-by-my-app python -c "
import os
print('before:', os.environ['MY_APP_SECRET'])
import hindsight_api.config
print('after :', os.environ['MY_APP_SECRET'])
"
```

```
before: set-by-my-app
after : from-the-parent-dotenv
```

The application never asked for this and gets no signal that it happened.

## Why this is worth fixing rather than working around

Three properties compound, and removing any one of them defuses it:

1. **It runs at import time**, so an application cannot opt out, order around it,
   or even observe it without diffing `os.environ` across the import.
2. **`override=True`**, so the application's own explicit configuration loses to
   a file it did not choose.
3. **`usecwd=True` + upward walk**, so the file is selected by directory layout
   rather than by anything the application controls.

There is precedent in this repo for treating import-time side effects as a bug
in their own right — #1595 (`tiktoken.get_encoding()` at module import taking
the API down on a DNS hiccup). #2634 is a different code path (FastMCP's own
pydantic-settings `.env` handling) but the same underlying hazard: cwd-relative
`.env` discovery inside a library, reached by an application that never opted in.

## Suggested fix

In rough order of preference:

1. **Move the load out of module scope** into whatever entry point actually wants
   it (the CLI / server startup path). Library import stays side-effect-free;
   the standalone server keeps its current convenience.
2. **Drop `override=True`.** Fill only keys the process has not set. This alone
   would have prevented the credential swap — the application's explicit values
   would have won.
3. **Gate it on an opt-in env var** (e.g. `HINDSIGHT_LOAD_DOTENV=1`), defaulting
   off when imported as a library.
4. **Restrict the search** to `HINDSIGHT_*` keys, or to a `.env` next to the
   config/bank rather than an arbitrary ancestor of cwd.

Happy to send a PR for whichever of these you'd prefer — I didn't want to pick
the semantics for you.

## Workaround for anyone hitting this

Snapshot and restore `os.environ` around the import:

```python
import contextlib, importlib, os

@contextlib.contextmanager
def preserved_env():
    snapshot = dict(os.environ)
    try:
        yield
    finally:
        os.environ.clear()
        os.environ.update(snapshot)

with preserved_env():
    hindsight = importlib.import_module("hindsight")
```

Note that it must wrap the **first** import of any hindsight package in the
process — the damage happens while `hindsight_api.config`'s module body runs, so
a later guarded import finds it already in `sys.modules` and has nothing left to
restore.

## Versions

`hindsight-all` / `hindsight-api-slim` 0.8.3, `hindsight-embed` 0.8.5,
python-dotenv 1.2.2, Python 3.11, macOS.

oldnicke:
## Summary

- stop loading ancestor `.env` files when `hindsight_api.config` is imported
- preserve `.env` convenience in the API, worker, and admin CLI entry points
- use `override=False` so explicit process environment values always win
- remove the recall benchmark workaround for import-time environment replacement
- add regression coverage for side-effect-free imports and CLI dotenv precedence

## Root cause

`hindsight_api.config` called `load_dotenv(find_dotenv(usecwd=True), override=True)` at module scope. Importing Hindsight therefore searched upward from the host application cwd and replaced arbitrary existing environment variables.

## Testing

- `pytest tests/test_dotenv_loading.py tests/test_main_module.py tests/test_worker_main.py -q -n 0` (`12 passed`)
- `pytest tests/test_config_validation.py tests/test_server_module.py -q -n 0` (`86 passed`)
- Ruff check and format validation for all changed Python files
- Ty checks for `hindsight_api`, `hindsight_dev`, and benchmarks
- `git diff --check`
- `./scripts/hooks/lint.sh`: all Python checks passed; frontend tasks could not start because the isolated worktree has no installed ESLint/Prettier binaries (`@eslint/js` / executables missing). No frontend files are changed by this PR.

Closes #2961

### DIFF
--- a/hindsight-api-slim/hindsight_api/admin/cli.py
+++ b/hindsight-api-slim/hindsight_api/admin/cli.py
@@ -17,7 +17,7 @@ from typing import Any
 import asyncpg
 import typer
 
-from ..config import DEFAULT_DATABASE_SCHEMA, HindsightConfig
+from ..config import DEFAULT_DATABASE_SCHEMA, HindsightConfig, load_dotenv_for_cli
 from ..engine.memory_engine import _current_schema
 from ..engine.retain.bank_utils import _vector_index_clause
 from ..engine.schema import fq_table_explicit as _fq_table
@@ -960,6 +960,7 @@ def worker_status(
 
 
 def main():
+    load_dotenv_for_cli()
     app()
 
 
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -19,10 +19,17 @@ from ._pg_search import normalize_pg_search_tokenizer
 from ._vector_index import validate_extension
 from .utils import mask_network_location
 
-# Load .env file, searching current and parent directories (overrides existing env vars)
-load_dotenv(find_dotenv(usecwd=True), override=True)
-
 logger = logging.getLogger(__name__)
+
+
+def load_dotenv_for_cli() -> None:
+    """Load a discovered .env file for Hindsight command-line entry points.
+
+    Library imports must not mutate the host application's environment. Standalone
+    commands retain .env convenience, while ``override=False`` ensures that values
+    explicitly supplied by the process take precedence over discovered files.
+    """
+    load_dotenv(find_dotenv(usecwd=True), override=False)
 
 
 class ConfigFieldAccessError(AttributeError):
--- a/hindsight-api-slim/hindsight_api/main.py
+++ b/hindsight-api-slim/hindsight_api/main.py
@@ -32,6 +32,7 @@ from .config import (
     ENV_WORKERS,
     HindsightConfig,
     _get_raw_config,
+    load_dotenv_for_cli,
 )
 from .daemon import (
     DEFAULT_DAEMON_PORT,
@@ -196,6 +197,8 @@ def _parse_cli_args(argv: list[str], config: HindsightConfig) -> ParsedCliArgs:
 def main():
     """Main entry point for the CLI."""
     global _memory
+
+    load_dotenv_for_cli()
 
     # Load configuration from environment (for CLI args defaults)
     config = _get_raw_config()
--- a/hindsight-api-slim/hindsight_api/worker/main.py
+++ b/hindsight-api-slim/hindsight_api/worker/main.py
@@ -18,7 +18,7 @@ import sys
 import warnings
 from collections.abc import Callable
 
-from ..config import get_config
+from ..config import get_config, load_dotenv_for_cli
 from ..engine.task_backend import WorkerTaskBackend
 from .poller import WorkerPoller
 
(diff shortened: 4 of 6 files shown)


oldnicke:
@nicoloboschi Thanks for preparing #2979 and covering the multi-worker server entry point while preserving the existing standalone dotenv precedence. This supersedes #2964, so I’ll close this PR in favor of #2979.

nicoloboschi:
Superseded by #2979 — same fix for the import-time `load_dotenv` side effect (#2961), but 100% backwards-compatible for our own deployments: keeps `override=True` (unchanged `.env` precedence) and also covers `server.py`, the uvicorn `hindsight_api.server:app` import string re-imported in each worker under `--workers`/`--reload` (loading `.env` only in the CLI entry points would silently break multi-worker mode). Thanks @oldnicke — the analysis carried over directly.

PR #2967, opened 2026-07-26
fix(claude-code): honor configured models

### LINKED ISSUE #2881: claude-code provider never passes the configured model to the CLI — `HINDSIGHT_*_LLM_MODEL` silently ignored, all calls run the CLI's own default
## Summary

With `HINDSIGHT_API_*_LLM_PROVIDER=claude-code`, the per-scope model settings (`HINDSIGHT_API_LLM_MODEL`, `HINDSIGHT_API_RETAIN_LLM_MODEL`, `HINDSIGHT_API_CONSOLIDATION_LLM_MODEL`, …) are accepted, stored, and **logged** — but never reach the spawned Claude Code CLI. Every retain/reflect/consolidation call runs whatever model the CLI resolves as its own default (currently an Opus-class model on Pro/Max OAuth accounts), regardless of configuration.

Because telemetry and log lines print `self.model`, the logs actively claim the configured model is in use (e.g. `scope=consolidation, model=claude-code/claude-sonnet-5`) while the CLI-side session transcripts show `"model": "claude-opus-4-8"` on every assistant turn. The mismatch is invisible unless you inspect the CLI's session JSONL.

## Root cause

In `hindsight_api/engine/providers/claude_code_llm.py`, neither `ClaudeAgentOptions` construction passes `model=`:

- `call()` (~line 236 on current `main`): options carry `system_prompt`, `max_turns=1`, `tools=[]`, `allowed_tools=[]`, `env` — no `model`.
- `call_with_tools()` (~line 539): same — no `model`.

`self.model` is only used for metrics (`metrics.record_llm_call`), span recording, and log strings — never for the actual SDK call (`query(...)` / `ClaudeSDKClient(options=options)`).

Two design details make this unworkaroundable from the operator side:

1. `_get_isolated_claude_env()` points `CLAUDE_CONFIG_DIR` at a **fresh temp dir**, so a `settings.json` model preference on the host/container can never reach the spawned CLI.
2. No `ANTHROPIC_MODEL` is set in that env either, so unless the operator happens to export it in the parent process environment (undocumented), nothing constrains the CLI's default.

`ClaudeAgentOptions` has supported `model: str | None` for a long time (claude-agent-sdk 0.2.x), so the fix is one line per call site.

## Impact

Anyone running the claude-code provider on a subscription (Pro/Max) account is silently billed/limited at the CLI's default-model tier instead of the model they configured. In our deployment (hindsight 0.8.4, claude-agent-sdk 0.2.110, subscription OAuth), an instance configured for Sonnet ran **100% of its extraction traffic on Opus** for 11 days — peaking at ~13M output tokens/day — and repeatedly exhausted the account's rolling session limits, starving interactive use of the same account. Configured-vs-actual model divergence is also a silent quality/cost correctness bug for API-key users once the CLI default moves.

## Reproduction

1. `HINDSIGHT_API_RETAIN_LLM_PROVIDER=claude-code`, `HINDSIGHT_API_RETAIN_LLM_MODEL=claude-sonnet-5` (any non-default model).
2. Trigger any retain.
3. Find the isolated config dir (logged at debug level as `Claude Code: isolated CLAUDE_CONFIG_DIR=…`, or `ls -d /tmp/hindsight-claude-code-*`) and read `projects/*/*.jsonl` inside it: every assistant message reports the CLI default model, not the configured one — while hindsight's own logs print the configured model.

## Fix

Pass the configured model through in both call sites:

```python
options = ClaudeAgentOptions(
    ...,
    model=self.model,
    env=_get_isolated_claude_env(),
)
```

We've been running exactly this as a local patch on 0.8.4 (both call sites) and verified via the CLI session JSONL that per-scope models are honored (retain/consolidation on `claude-sonnet-5`, reflect on `claude-sonnet-4-6`), with no behavioral regressions. Happy to submit a PR.

## Related

- #2850 fixed `tools=[]` on the `call()` path in the same file (merged 2026-07-21). While debugging this model issue we hit that exact ToolSearch/`max_turns=1` failure too and applied the identical fix locally — this issue is the remaining gap in the same options block.

r266-tech:
## Summary

- pass the resolved Claude Code model to both SDK option paths
- preserve empty-model behavior by omitting the SDK keyword
- cover standard and tool-enabled paths, including the installed SDK contract

Fixes #2881.

This also makes the existing provider-aware default effective when no per-scope
model is configured; it does not change the default-model table.

## Verification

- `uv run pytest tests/test_claude_code_llm_isolation.py -q` (7 passed)
- `uv run pytest tests/test_provider_default_models.py tests/test_per_operation_llm_config.py -q`
  (16 passed; one unrelated test could not start embedded Postgres because the
  sparse test environment lacks the optional `pg0-embedded` package)
- targeted Ruff check and format check
- targeted `ty` check
- `git diff --check`
- Codex adversarial review: approved

Thanks @bchumerch for the production reproduction and local validation, and
@ebarkhordar for verifying the current-main call sites and SDK contract.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/claude_code_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/claude_code_llm.py
@@ -239,6 +239,7 @@ class ClaudeCodeLLM(LLMInterface):
             tools=[],  # Disable built-in tools so nothing forces a ToolSearch deferral
             allowed_tools=[],  # Disable tools for standard LLM calls
             env=_get_isolated_claude_env(),
+            **({"model": self.model} if self.model else {}),
         )
 
         # Call Claude Agent SDK
@@ -543,6 +544,7 @@ class ClaudeCodeLLM(LLMInterface):
             mcp_servers=mcp_servers_config,
             allowed_tools=allowed_tool_names,
             env=_get_isolated_claude_env(),
+            **({"model": self.model} if self.model else {}),
         )
 
         # Call Claude Agent SDK with retry logic
--- a/hindsight-api-slim/tests/test_claude_code_llm_isolation.py
+++ b/hindsight-api-slim/tests/test_claude_code_llm_isolation.py
@@ -16,16 +16,18 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from typing import Any
-from unittest.mock import patch
 
 import pytest
+
+_MODEL_UNSET = object()
 
 
 @dataclass
 class _FakeOptions:
-    """Stand-in for ClaudeAgentOptions; captures kwargs without importing SDK."""
+    """Stand-in for ClaudeAgentOptions; captures kwargs without constructing real options."""
 
     system_prompt: str | None = None
+    model: str | object = _MODEL_UNSET
     max_turns: int | None = None
     allowed_tools: list[str] = field(default_factory=list)
     tools: list[str] = field(default_factory=list)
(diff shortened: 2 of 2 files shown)


handnewb (approved):
Clean fix — the claude-code integration should respect operator-configured models instead of always using defaults. Small change (+29/-8), well-scoped.

nicoloboschi:
fixed by #2980

PR #2968, opened 2026-07-26
perf: eliminate redundant graph seed and UUID scans

zommiommy:
## Summary

- derive graph-specific semantic seeds from the combined semantic/BM25 query's existing HNSW over-fetch pool when its threshold covers the independently configured graph threshold
- retain the dedicated graph seed query when the semantic threshold is stricter, preserving graph candidate semantics
- compare `memory_units.id` as UUID instead of casting the indexed column to text
- filter canonical UUID text before casting so malformed, uppercase, braced, and unhyphenated inputs remain silently unmatched as before

## Production evidence

All measurements are read-only against an 8-vCPU AlloyDB primary over 2026-07-19 through 2026-07-26.

- graph seed query hash `10423161873194026836`: 302,602 calls, 2,071.913 seconds accumulated load
- UUID date lookup hash `8287897089562102294`: 5,115 calls, 867.790 seconds accumulated load
- ten production embedding/tag samples produced the same 20 graph seed IDs from the existing 100-row semantic pool as from the dedicated 20-row query
- UUID lookup: 120.049 ms median +/- 2.847 ms MAD -> 0.240 ms +/- 0.009 ms on the same 20 valid IDs plus malformed/noncanonical inputs (500.2x)
- UUID plan: sequential scan with 68,213 shared-buffer hits -> `pk_memory_units` index scan with 80 hits
- prepared `text[]` statement: 0.274 ms, same 20 returned rows

### PostgreSQL 18 + pgvector

Used the repository's `pgvector/pgvector:pg18` external-database image through `HINDSIGHT_API_DATABASE_URL`.

- UUID regression passed with canonical, malformed, uppercase, braced, and unhyphenated inputs
- default recall passed while `_find_semantic_seeds()` was forbidden, proving the shared semantic pool supplies graph seeds
- stricter `MinScores(semantic=0.9)` passed while recording one dedicated graph seed query, proving the fallback remains intact
- an empty shared seed list is covered and does not fall back to a second query
- 100,000-row scale benchmark, five warm-ups plus 30 alternating measured executions:
  - old form: 9.005 ms median +/- 0.163 ms MAD, sequential scan, 2,440 shared-buffer hits
  - new form: 0.0795 ms median +/- 0.0075 ms MAD, `pk_memory_units`, 23 hits
  - same returned IDs; 113.3x faster

### API suite and quality gates

- external-PostgreSQL bulk run: 4,238 passed, 216 skipped, 1 xfailed
- remaining bulk-run failures: two credential-backed extraction tests and one unrelated webhook timing flake; the webhook test passed immediately in isolation
- nine environment/concurrency-sensitive tests passed serially, including bank-index repair and tool-choice coverage
- graph/dedup focused regressions: 35 passed
- `tests/test_db_abstraction.py`: 106 passed
- `ruff check`: passed
- `ty check`: passed
- `scripts/hooks/lint.sh`: passed
- `scripts/hooks/check-unused.sh`: only existing advisory findings
- upstream CI run [30199494931](https://github.com/vectorize-io/hindsight/actions/runs/30199494931): success; API/secret-backed jobs were skipped by the workflow

## Remaining test-environment gaps

- pg0-only migration tests cannot start pg0's bundled PostgreSQL on this NixOS host
- SeaweedFS tests require their service fixture
- credential-backed LLM extraction tests require repository/provider secrets

None of these gaps exercise the changed graph-seed reuse or UUID lookup paths; both modified PostgreSQL paths have direct integration coverage above.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -209,7 +209,7 @@ async def _dedup_adjudicate(
     grouped = await retrieve_semantic_bm25_combined(
         conn, anchor_emb_str, anchor_text, bank_id, ["observation"], _DEDUP_TOP_K, tags=tags, tags_match=tags_match
     )
-    results = grouped.get("observation", ([], []))[0]
+    results = grouped["observation"].semantic
     best_id: str | None = None
     best_text = ""
     best_sim = threshold  # only candidates at/above the threshold are considered
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -481,11 +481,21 @@ class PostgreSQLOps(DataAccessOps):
         mu_table: str,
         unit_ids: list[str],
     ) -> list[ResultRow]:
+        # Cast only canonical UUID text inputs, never the indexed column. The old
+        # ``id::text`` predicate silently ignored malformed, uppercase, braced,
+        # and unhyphenated inputs; filtering before the cast preserves that
+        # behavior while allowing the primary-key index to serve the lookup.
         return await conn.fetch(
             f"""
             SELECT id, event_date, fact_type
             FROM {mu_table}
-            WHERE id::text = ANY($1)
+            WHERE id = ANY(
+                ARRAY(
+                    SELECT input.unit_id::uuid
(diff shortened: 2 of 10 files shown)


nicoloboschi (approved):
LGTM. Two well-scoped, behavior-preserving Postgres perf fixes with thorough test coverage (reuse/fallback/empty-pool + UUID edge cases). Threshold-coverage guard, empty-vs-None handling, and the lowercase-canonical UUID regex are all correct. No must-fix issues.

PR #2972, opened 2026-07-26
fix(reflect): safely strip leaked done-tool siblings

gvilleg:
## Why

Follow-up to #2298. Some providers can close the `answer` string and then leak sibling `done` arguments such as `memory_ids`, `observation_ids`, `mental_model_ids`, or `directive_compliance` into the returned answer text. The existing cleanup handled complete JSON envelopes and a few textual patterns, but missed truncated sibling fields after a quoted answer. A broad regex would also risk deleting legitimate prose that merely discusses those field names.

## What changed

- parse the terminal suffix as JSON-aware sibling fields instead of matching field names alone
- require the leaked IDs to match positive retrieval provenance from the current reflect run
- preserve complete JSON documents and prose without matching provenance
- retain an empty cleaned result so metadata-only leakage reaches the existing `NO_ANSWER_TEXT` failure guard
- add regression coverage for field permutations, multiple ID fields, escaped content, unexpected keys, unknown IDs, intended terminal quotes, and metadata-only leakage

## Checks

- `uv run --frozen ruff check hindsight_api/engine/reflect/agent.py tests/test_reflect_agent.py`
- `uv run --frozen ruff format --check hindsight_api/engine/reflect/agent.py tests/test_reflect_agent.py`
- `uv run --frozen pytest tests/test_reflect_agent.py -k TestCleanDoneAnswer -q` — 20 passed
- independent review suite — 68 deterministic cases passed; 3 environment-dependent cases stopped at missing LLM configuration
- repository-wide `scripts/hooks/lint.sh` was also run: the changed files passed their scoped Ruff checks, while unrelated control-plane ESLint failed because `@eslint/js` is absent and Windows `ty` reported existing Unix/MLX-only imports

No public API or schema changes.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -101,6 +101,27 @@ _TRAILING_IDS_PATTERN = re.compile(
 )
 _JSON_CODE_FENCE_PATTERN = re.compile(r"^\s*```(?:json)?\s*(\{.*\})\s*```\s*$", re.DOTALL | re.IGNORECASE)
 
+# A truncated done-argument object: the answer string was closed and its SIBLING
+# fields followed, e.g.  ...end of prose", "memory_ids": ["a", "b"]\n}
+#
+# Nothing else here catches that shape. _unwrap_leaked_done_arguments needs the
+# whole text to parse as JSON, but the opening {"answer": " was consumed as the
+# answer's start so the remainder is not valid JSON. _LEAKED_JSON_SUFFIX needs a
+# ``` fence. _strip_trailing_id_json_object needs the run to begin at "{".
+# _TRAILING_IDS_PATTERN needs an UNQUOTED key and nothing after the "]" — this
+# key is quoted and a "}" follows. Observed in production: 17 memory UUIDs were
+# persisted into a mental model's stored content.
+#
+# Matching only a well-formed run of trailing JSON fields (each `"key": <json>`)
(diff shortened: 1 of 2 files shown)


gvilleg:
Coordination note: #2960 and this PR were both cut from the same `main` commit and both modify `_process_done_tool` answer handling. If #2960 lands first, this PR will need a rebase/conflict resolution in that block.

The safe composed ordering is:

1. Clean `raw_answer` with `_clean_done_answer(...)`, passing all three `available_*_ids` sets.
2. Compute `answer_failure_reason = None if answer else "empty_answer"` from the **cleaned `answer`**.
3. Only then substitute `NO_ANSWER_TEXT` for an empty answer.

Please do not compute the failure reason from `raw_answer`: a metadata-only leaked sibling tail is non-empty before cleaning, so that resolution silently reopens the persistence bypass while still passing cleaner-only tests. The combined path has been verified end-to-end: a tail-only leak produces `answer_failure_reason="empty_answer"` plus the sentinel, while a genuine synthesis produces `None` and non-sentinel text.

nicoloboschi:
Superseded by #3013, which changes the approach: instead of adding more salvage logic to strip leaked `done` siblings out of the answer text, #3013 removes the entire strip zoo and fails reflect loudly (`ReflectToolCallError` → HTTP 500) when a transport can't produce a usable tool call. Root cause traced to litellm's Vertex AI gpt-oss MaaS transformer stripping `tools`/`tool_choice` from the request (the same model tool-calls cleanly over Ollama). Closing in favor of that fix.

PR #2974, opened 2026-07-27
fix(claude-code): send MemoryItem.timestamp so the extractor can resolve relative dates

merlinr68:
**Of ~1,400 facts accumulated in a real bank, exactly 1 carried a usable date.** The retain hook never sends `MemoryItem.timestamp`, so the extractor has no anchor: "yesterday" and "last week" are stored verbatim, `occurred_start` stays null, and the memory never reaches the Timeline. The hook already computes the value — it just never reaches the request body.

---

## Problem

The Claude Code retain hook never sends `MemoryItem.timestamp`, so the extractor has no reference time to resolve relative expressions in a transcript.

The result is facts that are stored but not usable later. A session saying "yesterday the sensors phantom-cascaded" is extracted as:

```
occurred_start=None | ... | When: Today (relative to conversation)
occurred_start=None | ... | When: Last week
```

"Last week" relative to *what*? The anchor is exactly the information that was dropped. Once that fact is recalled weeks later it conveys nothing, and `occurred_start` stays null so the memory never reaches the Timeline.

This is listed as an anti-pattern in Hindsight's own best-practices guide:

> **Missing `timestamp` on retain** — Disables temporal retrieval strategies — *Always set from actual content timestamps*

and in the `timestamp` field docs:

> Set whenever you have temporal context. Enables temporal retrieval strategies.

## Measurement

Same transcript, same `retain_mission`, run through `dry-run-extract` on a self-hosted server with the anchor as the only variable:

| transcript phrase | without `timestamp` | with `timestamp` |
|---|---|---|
| "last week we soldered the tamper pull-down" | `When: Last week` | **`on 2026-07-19`** |
| "yesterday the sensors phantom-cascaded" | *(unresolved)* | **`on 2026-07-25`** |
| "today I want to plan Phase B" | `When: Today (relative to conversation)` | **`on 2026-07-26`** |

On a bank of ~1,400 facts accumulated without the anchor, exactly **1** carried an `occurred_start`.

## Fix

The hook already computes the instant it needs. `run_retain` builds `template_vars["timestamp"]` (`time.gmtime()`, ISO 8601) and records it as `retained_at` metadata — it just never reached the request body. This passes that same value through as the MemoryItem reference time.

- No new configuration.
- No behaviour change for callers that omit it: `HindsightClient.retain(timestamp=None)` sends no `timestamp` key at all, so the server keeps its existing ingestion-time fallback rather than receiving an explicit null.

## Tests

Two added to `TestRetainHook` (203 pass, up from 201):

- `test_retain_sends_timestamp_for_relative_date_resolution` — asserts the posted item carries an ISO-8601 `timestamp` and that it matches the `retained_at` metadata, so both describe the same instant.
- `test_client_omits_timestamp_field_when_not_supplied` — pins the omission path.

Mutation-checked: deleting the `item["timestamp"] = timestamp` assignment fails the first test; restoring it passes.

## Scope note

This only supplies the anchor. Whether the extractor then populates `occurred_start` is model-dependent — a local qwen3.6 resolved relative dates into the fact *text* but still left `occurred_start` null, while a hosted model populated it 3/5 of the time on the same input. Sending the timestamp is a precondition for either, and improves the stored text regardless of which model is behind the extractor.

### DIFF
--- a/hindsight-integrations/claude-code/CHANGELOG.md
+++ b/hindsight-integrations/claude-code/CHANGELOG.md
@@ -22,6 +22,20 @@
   `HINDSIGHT_USER_ID` is unset) are now dropped from retain requests. Previously
   such tags were sent as-is. Tags without `:` are unaffected.
 
+### Fixed
+
+- Retain requests now send the `MemoryItem.timestamp` field, using the same
+  wall-clock instant already recorded as `retained_at` metadata. Previously the
+  hook never sent one, so the extractor had no reference time to resolve
+  relative expressions: a transcript saying "yesterday" or "last week" was
+  stored with that phrase unresolved in the fact text (e.g.
+  `"When: Today (relative to conversation)"`), which conveys nothing once the
+  fact is recalled weeks later. Measured against a self-hosted server, adding
+  the anchor turned "last week we soldered the tamper pull-down" into a fact
+  dated `2026-07-19`, and "yesterday" into `2026-07-25`. Hindsight's own
+  best-practices guide lists omitting `timestamp` as an anti-pattern
+  ("Missing `timestamp` on retain — disables temporal retrieval strategies").
+
 ## [0.1.0] - 2025-03-23
 
 ### Added
--- a/hindsight-integrations/claude-code/scripts/lib/client.py
+++ b/hindsight-integrations/claude-code/scripts/lib/client.py
@@ -145,6 +145,7 @@ class HindsightClient:
         context: Optional[str] = None,
         metadata: Optional[dict] = None,
         tags: Optional[list] = None,
+        timestamp: Optional[str] = None,
         timeout: int = 15,
     ) -> dict:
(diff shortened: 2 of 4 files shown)


handnewb (approved):
Nice fix! Without timestamps in MemoryItem, the extractor can't resolve relative dates like 'yesterday' or 'last week'. This is a quality-of-life improvement for recall accuracy.

nicoloboschi:
Closing: superseded by **Coding Agents**, which carries timestamps at both levels.

Each retained turn keeps its own absolute ISO timestamp in the JSONL transcript, and the retain itself is submitted with `timestamp: startTs` (the session's first turn) so the document is anchored in time for temporal ranking. Backfilled sessions synthesize absolute per-turn times from a real clock rather than a fabricated epoch, for the same reason you give: the extractor resolves "yesterday" against something real.

PR #2978, opened 2026-07-27
fix(clients): expose async retain operation_id

r266-tech:
## Summary

- expose `operation_id` through the high-level Python `retain`/`retain_batch` helpers
- expose the same option as `operationId` through the TypeScript `retain`/`retainBatch` helpers
- serialize the wire field only for asynchronous, non-null requests, preserving the legacy request shape and old Python helper overrides

Follow-up to #2947. That change added caller-supplied operation IDs to the generated request models, but the hand-maintained convenience clients did not expose them. Users of the normal Python and TypeScript surfaces therefore could not safely retry an async retain after a lost acknowledgement without dropping to the generated client.

The scope is limited to the two high-level clients and focused regressions; server behavior and generated clients are unchanged.

## Verification

- Python retain/request tests: 17 passed
- TypeScript request-option tests: 5 passed
- `npx tsc --noEmit`
- `npm run build`
- Ruff checks and format checks on the changed Python regions

### DIFF
--- a/hindsight-clients/python/hindsight_client/hindsight_client.py
+++ b/hindsight-clients/python/hindsight_client/hindsight_client.py
@@ -283,6 +283,7 @@ class Hindsight:
         tags: list[str] | None = None,
         update_mode: str | None = None,
         retain_async: bool = False,
+        operation_id: str | None = None,
     ) -> RetainResponse:
         """
         Store a single memory (sync wrapper — prefer :meth:`aretain` in async code).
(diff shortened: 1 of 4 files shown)


nicoloboschi (approved):
Approved. Added a warning at each retain entry point (Python `warnings.warn`, TS `console.warn`) so a caller-supplied `operation_id`/`operationId` on a synchronous retain no longer disappears silently. Tests cover warn-once-on-sync (full delegation chain), no-warn-on-async, and no-warn-by-default in both clients; ruff/prettier/tsc all clean.

PR #2983, opened 2026-07-27
fix(llm): send reasoning_effort on the tool path, matching call()

merlinr68:
## Problem

`OpenAICompatibleLLM` builds request parameters in two independent places:

- `call()` sets `call_params["reasoning_effort"]` when `_supports_reasoning_model()`
- `call_with_tools()` builds its own `call_params` and **never set it**

Omitting the parameter is not a neutral default. Measured directly against the OpenAI API for `gpt-5.6-terra` with function tools:

| `reasoning_effort` | result |
|---|---|
| `"low"` | HTTP 400 |
| **absent** | **HTTP 400** |
| `"none"` | succeeds |

```
Function tools with reasoning_effort are not supported for gpt-5.6-terra in
/v1/chat/completions. To use function tools, use /v1/responses or set
reasoning_effort to 'none'.
```

That "absent also fails" row is the crux: a code path that *skips* the parameter fails exactly like one that sets it wrong.

## Why this is worse than a plain 400

`HINDSIGHT_API_LLM_REASONING_EFFORT=none` cannot fix it, because that setting only ever reached `call()`.

Reflect is a tool-calling search loop, so on an affected model every tool call 400s, retries, and then falls back to a tool-less completion. **The fallback still returns content and still stamps `last_refreshed_at` and clears `is_stale`** — so a mental model shows as freshly refreshed while never having searched memory.

Every health signal reads normal: refreshes "complete", timestamps advance, cron fires on schedule, `is_stale` goes false. In our deployment the only outward tell was input-token volume — **~800 tokens per call while degraded, versus 2.4k–8.2k when healthy.**

## Fix

One hunk in `call_with_tools()`, mirroring `call()`:

```python
if self._supports_reasoning_model():
    call_params["reasoning_effort"] = self.reasoning_effort
```

**On blast radius** — the alternative was gating the tool path to `provider == "openai"`. We deliberately didn't: `call()` already sends this parameter to the same provider/model pairs under the same capability check, so gating only the tool path would replace one asymmetry with another. `_supports_reasoning_model()` is the contract, and a parameterized test now pins that across `openai` / `openrouter` / `deepseek` so the decision is explicit rather than incidental.

Providers with their own `call_with_tools` (anthropic, gemini, litellm, codex, claude-code, llamacpp, nous) are unaffected. No subclass overrides `_supports_reasoning_model()`.

## Deliberately not fixed here

`call_with_tools()` diverges from `call()` in three further ways. None has a reproduction, and bundling them would make this hard to review and hard to revert:

1. **temperature** — `call()` guards with `not is_reasoning_model`; the tool path applies it unconditionally. Reasoning models generally reject an explicit temperature, so this looks like the same class of latent bug.
2. **groq `service_tier`** — set by `call()`, absent on the tool path.
3. **groq `include_reasoning=False`** — set by `call()` for reasoning models, absent on the tool path.

Longer term, the two builders sharing a helper would stop this drift recurring. Happy to follow up on any of these if maintainers want them.

## Tests

7 tests in `tests/test_tool_path_reasoning_effort.py`:

- value forwarded verbatim on the tool path (`"none"` specifically, since it must not be coerced)
- a non-default value (`"high"`) proves it comes from config, not the constructor default
- non-reasoning model (`gpt-4o`) gets no parameter
- **parity test** pinning `call()` and `call_with_tools()` together so they can't drift again
- parameterized blast-radius test across three providers

Verification:
- 7 pass; **deleting the hunk fails 6 of 7** (the survivor asserts the negative branch)
- 161 provider tests pass, 8 skipped — `-k "tool_choice or reasoning or openai_compat or litellm or deepseek or groq"`
- `ruff check` and `ruff format --check` clean
- Live end-to-end: reflect went from 20 errors and the 806-token fallback to **zero errors and 2.4k–7.5k-token real searches**, refreshing 5 mental models in 91s

Also updated the constructor docstring, which listed `("low", "medium", "high")` and so implied `"none"` was invalid — it's now the required value for this case.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -541,7 +541,10 @@ class OpenAICompatibleLLM(LLMInterface):
             api_key: API key (optional for ollama/lmstudio).
             base_url: Base URL for the API (uses defaults for groq/ollama/lmstudio if empty).
             model: Model name.
-            reasoning_effort: Reasoning effort level for supported models ("low", "medium", "high").
+            reasoning_effort: Reasoning effort level for supported models
+                ("none", "low", "medium", "high"). "none" is required when calling
+                function tools on some reasoning models, which reject every other
+                value — including omitting the parameter entirely.
             timeout: Request timeout in seconds (uses env var or 120s default).
             groq_service_tier: Groq service tier ("on_demand", "flex", "auto").
             extra_body: Extra body params merged into every API call.
@@ -1255,6 +1258,13 @@ class OpenAICompatibleLLM(LLMInterface):
             if self.provider == "minimax":
                 temperature = max(0.01, min(temperature, 1.0))
             call_params["temperature"] = temperature
+
+        # Set reasoning_effort for reasoning models, matching call(). Omitting it
+        # here is not a neutral default: OpenAI rejects function tools on a
+        # reasoning model unless reasoning_effort is present and set to "none",
+        # so leaving it out fails exactly like sending an unsupported value.
+        if self._supports_reasoning_model():
+            call_params["reasoning_effort"] = self.reasoning_effort
 
         # Provider-specific parameters
         extra_body: dict[str, Any] = {**self._config_extra_body}
--- a/hindsight-api-slim/tests/test_tool_path_reasoning_effort.py
+++ b/hindsight-api-slim/tests/test_tool_path_reasoning_effort.py
@@ -1,0 +1,160 @@
+"""
+`call_with_tools()` must send `reasoning_effort` for reasoning models, like `call()` does.
+
+`call()` sets it; `call_with_tools()` built its own `call_params` and left it out.
+Omitting it is not a neutral default — OpenAI rejects function tools on a
+reasoning model unless `reasoning_effort` is present and set to "none":
+
+    Function tools with reasoning_effort are not supported for gpt-5.6-terra in
+    /v1/chat/completions. To use function tools, use /v1/responses or set
+    reasoning_effort to 'none'.
+
+Verified directly against the API for gpt-5.6-terra + function tools:
+  reasoning_effort="low"  -> HTTP 400
+  reasoning_effort absent -> HTTP 400
+  reasoning_effort="none" -> tool call succeeds
+
+So the omission makes `HINDSIGHT_API_LLM_REASONING_EFFORT` unable to fix the
+failure, because that setting only ever reached the non-tool path. In practice
+reflect (a tool-calling search loop) degrades to a no-search fallback while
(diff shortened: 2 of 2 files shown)


merlinr68:
Field note after running this fix for ~11 hours: it works, but applying it surfaced a second problem that this PR **cannot** solve, and I think it's worth a maintainer's attention.

## The workaround this PR removes has a cost elsewhere

Before this fix, the only way to reach the tool path was `HINDSIGHT_API_LLM_EXTRA_BODY={"reasoning_effort":"none"}`, since `extra_body` is merged into `call_with_tools()`'s request while `reasoning_effort` never was. But `HINDSIGHT_API_LLM_REASONING_EFFORT` is **global** — there is no per-operation variant — so setting `none` to satisfy the tool path also applies it to `call()`, i.e. retain and consolidation.

That measurably degraded unrelated operations. On a bank of ~1,500 facts, retain began emitting facts in Chinese from a fully English transcript:

> 此前的 TLS 工作已提交；lat check 通过，之前的 12 个 @lat: 错误已消失。 | When: 2026年7月28日，星期二 | Involving: 另一个会话

**23 facts**, all after the change, none before. Timeline:

| event | UTC |
|---|---|
| `reasoning_effort=none` applied globally | 13:49, 14:19 |
| first non-English fact | **14:57** |

Reasoning-disabled generation appears to lose the implicit "respond in the input's language" behaviour. `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE=English` fixes it, and I'd recommend that to anyone in this situation — but it's a second workaround stacked on the first.

Worth noting how invisible this was: retains succeeded, facts were created, counts rose, zero errors. Nothing in `/operations`, `/stats` or the logs indicated a problem. It was only caught by reading fact text.

## Why this PR still stands

To be explicit, since the two interact: **this PR does not cause the language drift and does not fix it.** It forwards whatever value is configured to the tool path; the drift comes from `call()` receiving `none`, which it already did before this change. Every measurement in the description is unaffected.

## The gap this leaves

Even with this merged, a deployment on a model that requires `reasoning_effort="none"` for function tools still has to set it globally, so retain and consolidation inherit it. There's no way to say "reflect needs `none`, retain should keep `low`".

The per-operation config already supports `PROVIDER` / `API_KEY` / `MODEL` / `BASE_URL` / `MAX_CONCURRENT` / `MAX_RETRIES` / `TIMEOUT` under `RETAIN_LLM_*`, `REFLECT_LLM_*` and `CONSOLIDATION_LLM_*` — `REASONING_EFFORT` is a conspicuous omission from that set, and it's the one that would have prevented this. I've opened a separate issue for it rather than widen this PR.

Happy to add the `OUTPUT_LANGUAGE` recommendation to the docs if that'd be useful.

nicoloboschi (approved):
Correct, minimal, and well-tested. The fix mirrors call() exactly and the parity test guards against future drift. Scoped fixes with documented follow-ups — approved.

PR #2984, opened 2026-07-27
fix(recall): retry get_or_create_bank_profile on per-bank index deadlock

dcbouius:
## Problem

`test_repair_bank_vector_indexes.py::test_import_bank_creates_per_bank_indexes` flakes on ~30% of `test-api (3/3)` runs (2 of the last 7) with:

```
asyncpg.exceptions.DeadlockDetectedError: deadlock detected
```

Fresh-bank creation issues per-bank `CREATE INDEX` statements against the shared `memory_units` table. The test suite runs all pytest-xdist workers against **one shared embedded Postgres** (per `conftest.py`), so two banks being created concurrently on different workers deadlock on that index DDL.

This surfaced on **#2950** (litellm), but that PR merely merged past the flake — a litellm bump can't cause a DB deadlock. It's ambient and pre-existing.

## Fix

Wrap the pool-owning `get_or_create_bank_profile` in `retry_with_backoff`, which already treats `DeadlockDetectedError` as retryable (equal-jitter backoff). Each retry runs on a **brand-new connection + transaction**, and the operation is idempotent (`SELECT` → `INSERT ... ON CONFLICT DO NOTHING` → per-bank indexes named on `internal_id`), so the deadlock-aborted transaction leaves nothing committed and the retry re-runs cleanly.

Only this variant retries — `get_or_create_bank_profile_on_conn` runs inside the *caller's* transaction, which the caller owns and must retry itself (documented in a comment).

This is the right layer: `import_bank_async` → `_ensure_bank_exists(conn=None)` → `get_or_create_bank_profile` (pool variant), which owns its transaction.

## Test

Adds a deterministic unit test (no DB) in `test_bank_utils.py`: an ops stub that raises `DeadlockDetectedError` once then succeeds, asserting the profile creation retries (two index-DDL attempts) and the bank ends up created. `retry_with_backoff` and type-check pass; lint clean.

### DIFF
--- a/hindsight-api-slim/tests/test_bank_utils.py
+++ b/hindsight-api-slim/tests/test_bank_utils.py
@@ -1,6 +1,8 @@
+import asyncio
 from contextlib import asynccontextmanager
 
 import pytest
+from asyncpg.exceptions import DeadlockDetectedError
 
 from hindsight_api.engine.retain import bank_utils
 
@@ -8,6 +10,18 @@ from hindsight_api.engine.retain import bank_utils
 class _FailingIndexOps:
     async def create_bank_vector_indexes(self, *args, **kwargs) -> None:
         raise RuntimeError("simulated per-bank vector index DDL failure")
+
+
+class _DeadlockOnceIndexOps:
+    """Raises a deadlock on the first per-bank index DDL, then succeeds."""
+
+    def __init__(self) -> None:
+        self.calls = 0
+
+    async def create_bank_vector_indexes(self, *args, **kwargs) -> None:
+        self.calls += 1
+        if self.calls == 1:
+            raise DeadlockDetectedError("deadlock detected")
 
 
 class _FakeTransaction:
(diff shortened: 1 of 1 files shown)


ebarkhordar (commented):
The regression test holds up. I ran it in a clean container at `7f2e83a`, then bypassed the retry to check it pins something rather than passing either way:

```
$ pytest tests/test_bank_utils.py -o addopts=""
as-is                                            2 passed

# then: return await retry_with_backoff(_create)
#   ->  return await _create()
test_get_or_create_bank_profile_retries_on_deadlock   FAILED
    asyncpg.exceptions.DeadlockDetectedError: deadlock detected
```

One thing that may cost a reviewer time: the "Fix" section describes wrapping `get_or_create_bank_profile` in `retry_with_backoff`, but that is already on main. It landed in `dfac776d` (#2943, 24 Jul), and `bank_utils.py` is identical between `main` and this branch, with the single commit here being the test. Trimming the body to present this as a regression test for #2943 would save the next reader looking for a production change that is not in the diff.

That leaves one question I cannot answer from outside the CI history: were the two flaking runs out of the last seven observed before or after `dfac776d`? If either ran with the retry already in place, then the retry is not closing the deadlock on its own, and this test would stay green while `test-api` keeps flaking.

PR #2987, opened 2026-07-27
docs(config): clarify per-type WORKER_*_MAX_SLOTS is a reservation floor, not a cap

### LINKED ISSUE #2963: HINDSIGHT_API_WORKER_<TYPE>_MAX_SLOTS is a minimum, not a maximum
### Summary

`HINDSIGHT_API_WORKER_<TYPE>_MAX_SLOTS` does not set a maximum. It sets a
*reserved minimum* — a floor. The name states the opposite of the behaviour, and
the resulting operator confusion is not hypothetical: it took us a production
latency investigation to discover that a worker log line reading
`consolidation=6/1(avail=0)` was correct and documented, not an overrun bug.

### Evidence (v0.8.4, image digest `sha256:2c60f233…dfa1a7`)

- `config.py` — `WORKER_SLOT_RESERVATION_TYPES` maps each type to its
  `HINDSIGHT_API_WORKER_<TYPE>_MAX_SLOTS` env var. The table is named
  *reservation*; the env var is named *max*.
- `worker/poller.py` — `WorkerPoller._get_available_slots` docstring:

  > When an operation type's in-flight count exceeds its reservation, the excess
  > tasks are considered to be using shared pool slots.

- Consequently `shared_pool = worker_max_slots - sum(reservations)` is claimable
  by *any* type, and a single type may occupy all of
  `DEFAULT_WORKER_MAX_SLOTS = 10` regardless of its `..._MAX_SLOTS` value.
- `config.py` validates `total_reserved <= worker_max_slots` — a floor-shaped
  invariant, which further confirms the semantic.

### Why it matters

An operator reading `HINDSIGHT_API_WORKER_CONSOLIDATION_MAX_SLOTS=1` reasonably
believes consolidation is capped at one concurrent task. It is not; nothing in
the configuration surface caps a type at all. On our fleet, consolidation ran at
6 in-flight and contended with latency-critical foreground work. Operators who
want a ceiling currently have no knob for it, and the one knob that exists reads
like it already is one.

### Suggested fix

Two parts, and the naming one is the cheaper of the two:

1. **Rename to what it does** — e.g. `HINDSIGHT_API_WORKER_<TYPE>_RESERVED_SLOTS`
   (or `..._MIN_SLOTS`), keeping `..._MAX_SLOTS` as a deprecated alias that logs
   a warning for at least one minor release. This is a breaking config change, so
   it needs the alias window; we deliberately did **not** rename it in our own
   downstream patch for that reason.
2. **Optionally add a genuine ceiling** — a per-type cap enforced at claim time,
   validated at boot to be `>= ` the type's own reservation and
   `<= worker_max_slots`. We implemented this downstream (enforced immediately
   before the `UPDATE ... SET status='processing'` in `claim_tasks`, so
   over-ceiling rows are simply left `pending` and their `FOR UPDATE SKIP LOCKED`
   locks drop with the transaction) and would be happy to upstream it if you want
   the shape.

Even part 1 alone — or just a docstring/README sentence saying "this is a
reservation, not a cap" — would have saved us the investigation.

nicoloboschi:
## What

Documentation-only fix for #2963. The per-operation `HINDSIGHT_API_WORKER_<TYPE>_MAX_SLOTS` env vars are named like a per-type ceiling but actually set a **reserved minimum** (a floor). A type is guaranteed at least its reservation and may then overflow into the shared pool (`WORKER_MAX_SLOTS - sum(reservations)`) up to `WORKER_MAX_SLOTS`. So the name reads as the opposite of the behaviour, which cost the reporter a production latency investigation over a `consolidation=6/1(avail=0)` log line that was working as designed.

## Changes

- Config-table entries for every `*_MAX_SLOTS` now say **Reserved (minimum) slots … a floor, not a per-type cap**.
- New warning admonition spelling out:
  - the real ceiling: `reserved(type) + shared pool currently available`;
  - that lowering the value lowers the *guarantee*, not the ceiling (e.g. `CONSOLIDATION_MAX_SLOTS=1` still allows up to 10 concurrent);
  - the consolidation per-bank serialization caveat;
  - the **only** way to actually cap a type today — reserve the whole pool so the shared pool is zero (worked example).
- Mirrored into `versioned_docs/version-0.8` so it appears on the live site (production ships released versions only, not `current`).

## Scope

Docs only — no behaviour change. Verified against `config.py` (`WORKER_SLOT_RESERVATION_TYPES`, the `total_reserved <= worker_max_slots` floor-shaped validation), `worker/poller.py` (`_get_available_slots`), and `engine/db/ops_postgresql.py::claim_tasks` (reserved phase + shared phase claims for any type).

A rename to `*_RESERVED_SLOTS` (with a deprecated alias) and/or an actual per-type ceiling knob — both suggested in the issue — are larger, behaviour-changing follow-ups intentionally left out of this PR.

Closes #2963

### DIFF
--- a/hindsight-docs/docs/developer/configuration.md
+++ b/hindsight-docs/docs/developer/configuration.md
@@ -1660,13 +1660,13 @@ Configuration for background task processing. By default, the API processes task
 | `HINDSIGHT_API_WORKER_MAX_SLOTS` | Maximum concurrent tasks per worker (total across all operation types) | `10` |
 | `HINDSIGHT_API_OPERATION_RETENTION_DAYS` | Static server-wide retention window for completed, failed, and cancelled operation rows, including their task payload and result metadata. `0` (the default) keeps them indefinitely; set a positive number of days to enable automatic pruning. | `0` |
 | `HINDSIGHT_API_OPERATION_CLEANUP_BATCH_SIZE` | Maximum expired terminal operation rows deleted per tenant schema during each cleanup cycle. The cleanup job runs once per maintenance tick, so this also sets the drain rate for a backlog. Must be a positive integer. | `1000` |
-| `HINDSIGHT_API_WORKER_CONSOLIDATION_MAX_SLOTS` | Reserved slots for consolidation tasks within `WORKER_MAX_SLOTS` (bank-serialization preserved) | `2` |
+| `HINDSIGHT_API_WORKER_CONSOLIDATION_MAX_SLOTS` | **Reserved (minimum) slots** for consolidation tasks within `WORKER_MAX_SLOTS` — a floor, **not** a per-type cap (bank-serialization preserved). See the note below. | `2` |
 | `HINDSIGHT_API_WORKER_CONSOLIDATION_BANK_PRIORITY` | Per-bank priority for consolidation scheduling (see note below) | _(unset)_ |
-| `HINDSIGHT_API_WORKER_RETAIN_MAX_SLOTS` | Reserved slots for retain tasks within `WORKER_MAX_SLOTS` | `0` |
-| `HINDSIGHT_API_WORKER_FILE_CONVERT_RETAIN_MAX_SLOTS` | Reserved slots for file_convert_retain tasks within `WORKER_MAX_SLOTS` | `0` |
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Superseded by #3016, which does the full rename (RESERVED_SLOTS floor + MAX_CONCURRENT ceiling, deprecated MAX_SLOTS alias) and includes the same docs clarification. Closing in favor of that.

Issue #1570, opened 2026-05-11
opencode: Per Agent Configuration


Toady00:
### Use Case

When using OpenCode or OpenClaw specifically, I'd like to be able to configure specific agents to use specific settings, or potentially disable entirely. A few examples of scenarios:

1. In OpenCode I have a custom agent that the work that it does is just noise for the memory system. I would like to be able to disable autoretain/autorecall for that agent specifically.
2. In OpenCode, I'd like to have a high level agent like a enterprise architect, and the memory bank that it uses would be across the entire platform, but my build agent should use a project specific memory bank. So I'd like to be able to configure memory banks per agent.
3. Using your own cookbook at https://hindsight.vectorize.io/cookbook/recipes/support-agent-shared-knowledge you show a single agent using multiple banks at the same time. The opencode plugin configures the hindsight_recall and hindsight_reflect to the specific memory bank as opposed to allowing it to pass a memory bank id. So setting up a multibank solution doesn't work with the plugin as is.

There are similar scenarios for openclaw as well.

### Problem Statement

It's impossible to configure each agent independently in open code or openclaw. It's also all or nothing where you can't disable a specific agent from participating in autorecall/autoreflect.

### How This Feature Would Help

Providing a per agent configuration block would allow each agent to be configured as needed. Also allowing a memory bank id as an arg to the tool call would allow a single agent to interact with multiple memory banks manually.

### Proposed Solution

Provide a per agent configuration block.

### Priority

Important - affects my workflow

### Checklist

- [x] I would be willing to contribute this feature

nicoloboschi:
this is now possible with the new coding agents - check it out https://hindsight.vectorize.io/sdks/integrations/coding-agents#configuration

PR #2993, opened 2026-07-27
Opencode Integration bank per agent

LargoUsagi:
I had pulled down the opencode integration and made changes locally to support banks per agent.  I tested this against my use cases and its working for what I need, but I don't know if any one else would find this useful.  Figured I would PR this back instead of maintaining a hard for for features that are adding to the existing state.

### DIFF
--- a/hindsight-integrations/opencode/README.md
+++ b/hindsight-integrations/opencode/README.md
@@ -100,13 +100,127 @@ Create `~/.hindsight/opencode.json` for persistent configuration:
 }
 ```
 
+### Per-Agent API Keys
+
+OpenCode drives sessions with named agents (e.g. `build`, `code-reviewer`, or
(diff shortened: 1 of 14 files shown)


nicoloboschi:
> I don't know if any one else would find this useful.

They would — this is essentially the feature request in #1570, which has been open since May and asks for exactly this: per-agent enable/disable and per-agent bank selection in opencode, plus the ability to point a single agent at more than one bank. Worth reading that issue; your registry/config split lines up well with what's described there.

Two things before this can be reviewed properly:

- **CI has never run on this PR.** There are zero check runs, so none of the 1,000-odd lines of tests you added have been exercised here. It's been open since 27 Jul, so it also needs a rebase onto current `main` — the opencode integration has moved since.
- Once it's rebased and green, I'll go through the registry and config surface in detail.

Linking #1570 so the two are connected.

PR #2995, opened 2026-07-27
feat(agent-sdk): let agent_knowledge_recall request source chunks

### LINKED ISSUE #2949: Expose raw source chunks (include.chunks) to agents via the agent_knowledge_recall SDK tool
## Problem

Agents built on `@vectorize-io/hindsight-agent-sdk` can only retrieve **extracted fact units** through `agent_knowledge_recall`. For questions that need precision — "what exactly did we say", verbatim wording, exact numbers/times — the summarized unit isn't enough, and the agent has no way to reach the raw source text a fact was extracted from.

## Current behavior

The recall API already supports this: `include: {chunks}` returns the raw source windows, and it works well (verified on a self-hosted 0.8.5 instance — chunks come back with the original dialogue intact).

But the SDK's `agent_knowledge_recall` tool forwards only `{maxTokens, types}` to `client.recall`:

```js
async execute(params) {
    const maxTokens = parseRecallMaxTokens(params.max_tokens);
    const types = normalizeFactTypes(params.fact_types ?? params.types, DEFAULT_RECALL_FACT_TYPES);
    const result = await client.recall(bankId, params.query, {
        maxTokens,
        types,
    });
    return ok(result);
}
```

so `include`/`chunks` never reaches the API from an agent context. None of the `agent_knowledge_*` tools expose raw-source retrieval.

## Proposed

Expose the capability on the agent side, either:

1. an optional `include_chunks: boolean` (plus an optional chunk token budget) parameter on `agent_knowledge_recall`, or
2. a dedicated tool (e.g. `agent_knowledge_recall_source`) that returns the raw chunks.

Option 1 is probably the smaller change. Option 2 keeps the default recall response lean and makes the "I need verbatim" intent explicit to the model — in our own prototype the separate-tool shape worked well for exactly that reason: ordinary lookups stay cheap, and the model reaches for the raw path only when the user asks for precision.

## Why it matters

For personal / long-term memory assistants, "show me exactly what was said" is a common and high-trust request. Without raw-source access the assistant can only paraphrase its own extraction — which is precisely the wrong answer when the user is asking to verify a detail.

---

Discussed in Slack (#all-hindsight); @nicoloboschi confirmed it's doable and asked me to open an issue.

chethanuk:
## Problem

`agent_knowledge_recall` forwards only `{maxTokens, types}` to `client.recall`, so `include.chunks` never reaches the API from an agent context. Agents can read extracted fact units but never the raw source text a fact came from, which is what "what did we actually say" and exact-number questions need.

The plumbing below this already exists. The TypeScript client declares `includeChunks` / `maxChunkTokens` and maps them to `include.chunks = { max_tokens: … ?? 8192 }`. Only the tool schema and the one call site were missing.

Fixes vectorize-io/hindsight#2949

## Fix

Two optional parameters on `agent_knowledge_recall`:

- `include_chunks` (boolean, default false): also return the raw source text the memories were extracted from.
- `max_chunk_tokens` (number, default 8192): token budget for those chunks.

`include_chunks` is gated on `=== true`, matching how `agent_knowledge_reflect` gates `include_facts`. `max_chunk_tokens` mirrors `parseRecallMaxTokens`, except it returns `undefined` on absent or unusable input so the client's own 8192 default applies instead of a second copy of that number living here.

Both are additive and off by default, so recall responses are unchanged unless an agent asks for chunks. Also one sentence in the openclaw integration docs, with its generated `skills/` mirror refreshed via `./scripts/generate-docs-skill.sh`.

## Open question

The issue offered two shapes and left the pick to you: an `include_chunks` flag on the existing tool, or a dedicated `agent_knowledge_recall_source` tool. This is the first, which is the smaller change and keeps one recall path. The argument for the second is real though. A separate tool makes the "I need verbatim" intent explicit to the model and keeps ordinary lookups lean. Happy to rework it that way if you prefer.

## Test evidence

| Check | Result |
|---|---|
| `npm test` (hindsight-agent-sdk) | 20 passed |
| `npx tsc --noEmit` | clean |
| `./scripts/hooks/lint.sh` | clean |

Three new cases, each asserting the outgoing request body rather than the intent:

- `recall omits source chunks unless include_chunks is exactly true`: omitted, `false`, the string `"false"`, and a bare `max_chunk_tokens` all leave chunks off.
- `recall include_chunks requests source chunks and returns them to the agent`
- `recall falls back to the client's chunk budget on unusable max_chunk_tokens`: 2048, 0, -5, non-numeric.

### DIFF
--- a/hindsight-docs/docs-integrations/openclaw.md
+++ b/hindsight-docs/docs-integrations/openclaw.md
@@ -153,7 +153,7 @@ Optional settings in `~/.openclaw/openclaw.json`:
 - `enableKnowledgeTools` - Register `agent_knowledge_*` tools for explicit agent-driven lookup, reflection, ingest, and knowledge-page management (default: `false`).
 - `debug` - Enable debug logging (default: `false`).
 
-When using `agent_knowledge_recall` manually, pass `max_tokens` to control how much memory text the recall response may contain. The tool has no `max_results` parameter — to cap the number of automatically injected memories, use `recallTopK` on auto-recall instead.
+When using `agent_knowledge_recall` manually, pass `max_tokens` to control how much memory text the recall response may contain. The tool has no `max_results` parameter — to cap the number of automatically injected memories, use `recallTopK` on auto-recall instead. When the answer needs verbatim wording or an exact number rather than an extracted fact, pass `include_chunks: true` to also get the raw source text those memories came from, and `max_chunk_tokens` to bound it (default `8192`).
 
 When using `agent_knowledge_reflect`, keep the default conservative settings unless you intentionally need a deeper synthesis: `budget` defaults to `low`, `max_tokens` defaults to `1024`, and `fact_types` defaults to `world`, `experience`, and `observation`. Reflect calls can be more expensive than recall because they retrieve memories and then call the configured Reflect LLM to generate an answer. For production banks, set a finite bank-level `reflect_source_facts_max_tokens` value (for example `4096` or `8192`) instead of leaving it unlimited, so ad-hoc reflection cannot pull an unbounded amount of source facts into the LLM context.
 
--- a/hindsight-tools/hindsight-agent-sdk/src/index.ts
+++ b/hindsight-tools/hindsight-agent-sdk/src/index.ts
@@ -99,6 +99,14 @@ function parseRecallMaxTokens(input: unknown): number {
   return n;
 }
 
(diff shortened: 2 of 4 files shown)


nicoloboschi (approved):
LGTM. Additive include_chunks/max_chunk_tokens on agent_knowledge_recall, off by default, gated on ===true. parseChunkMaxTokens returns undefined on bad input so client's 8192 default applies. Docs + generated mirror in sync, edge cases tested. Fixes #2949.

PR #3000, opened 2026-07-28
fix(claude-code): count turn boundaries by conversational content, not user-role records

merlinr68:
**Asking for 3 turns of recall context returned 82 characters.** Turn boundaries count every `role: "user"` record, but in agentic transcripts **91% of those are tool results** (measured: 514 of 566), so the window is exhausted before reaching any real conversation. Counting boundaries by conversational content instead takes the same request from **82 → 8,081 characters**.

---

## The inconsistency

`slice_last_turns_by_user_boundary()` counts every `role == "user"` record as a turn boundary:

```python
if messages[i].get("role") == "user":
    user_turns_seen += 1
```

But the conversation extractor in the same file already treats many of those records as non-conversational. `_extract_text_content()`'s own docstring says it excludes:

> `{type: "tool_result"}` — operational results, not conversation

and `compose_recall_query()` additionally strips channel envelopes and injected `<hindsight_memories>` blocks.

So turn selection counts records that formatting then discards. A window can be exhausted by user-role transport records before reaching any actual conversation.

## Impact

Two call sites:

- `content.py` — `compose_recall_query()`, controlling `recallContextTurns` (runs on every user prompt)
- `retain.py` — the chunked-retention window

In Claude Code this is the common case rather than an edge case, because **tool results are stored as `role: "user"` messages**. Measured on a real 1,639-message transcript with 566 user-role records, **514 of them (91%) tool results**:

| setting | before | after |
|---|---|---|
| `recallContextTurns=3` | 1 conversational turn, **82 chars** | **3 turns, 8,081 chars** |
| `recallContextTurns=5` | 1 conversational turn, 3,019 chars | **5 turns, 14,152 chars** |

Asking for three turns of recall context returned 82 characters. The multi-turn feature was effectively inert on agentic transcripts.

## Fix

Rather than special-casing `tool_result` blocks, define a boundary using the same cleaned representation the composer already applies — extract conversational blocks, strip channel envelope, strip injected memory, trim — factored into one private helper:

```python
def _conversational_text(content, role: str = "") -> str:
    text = _extract_text_content(content, role=role)
    text = strip_channel_envelope(text)
    return strip_memory_tags(text).strip()
```

used by both boundary selection and recall composition, so the two cannot disagree about what counts as content.

This is preferable to structural `tool_result` detection because it handles the mixed cases correctly without duplicating knowledge that already lives in `_extract_text_content()`:

- tool-result-only record → not a boundary
- record with **both** a tool result and genuine text → still a boundary
- empty / whitespace-only record → not a boundary
- injected-memory-only record → not a boundary
- memory block **plus** a real prompt → still a boundary
- plain-string content (other hosts, flat test format) → unchanged

**Boundary selection is the only thing that changes.** Records inside the selected window are returned untouched, so retention and formatting policy still decide what is kept from them. Tool results remain excluded from output by `_extract_text_content()` exactly as before.

Both call sites want the same conversational unit, so the fix belongs in the shared helper rather than being duplicated — otherwise the two definitions would drift.

## Intentional consequences

**Recall windows traverse more records.** Finding N conversational turns may now walk across many tool records. Only extractable conversational content enters the query, and `recallMaxQueryChars` (default 800) bounds the **final** query — note it does not bound the intermediate composed string or the amount of transcript traversed. At the scale measured here that cost is negligible next to reading the transcript and making the recall request.

**Chunked retention documents get larger.** A window of N real turns now includes the tool activity between those prompts, where before it often spanned only the last few records. That is the point of a turn window, and the previously small payloads were a symptom of selecting the wrong boundaries — but it is a visible change, so calling it out.

## Not included

- Whether `recallContextTurns` should count the latest prompt as one of its N. The selector counts it and `compose_recall_query()` then drops it as a duplicate, so N can mean "latest plus N−1 prior". That predates this change and is left alone.
- Any payload capping or batching for chunked retention.
- Caching or precomputed classification — no evidence of a bottleneck.

## Tests

12 tests in `tests/test_turn_boundary_semantics.py` — 10 on boundary qualification, 2 on recall composition end-to-end. The key assertions check not just the size of the returned window but that it **starts at the Nth qualifying record and still contains the intervening tool results**.

- **213 pass** with the fix
- reverting the predicate **fails 5 of 12**
- `ruff check` and `ruff format --check` clean

## Related

Same family as #2999 (retain dropped the unprocessed suffix), #2974 (retain never sent `MemoryItem.timestamp`) and #2869 (SessionEnd retain cancelled at shutdown).

### DIFF
--- a/hindsight-integrations/claude-code/scripts/lib/content.py
+++ b/hindsight-integrations/claude-code/scripts/lib/content.py
@@ -35,6 +35,19 @@ def strip_channel_envelope(content: str) -> str:
     return content
 
 
+def _conversational_text(content, role: str = "") -> str:
+    """Return the conversational text of a message, or '' if it carries none.
+
+    The single definition of "does this message say anything to the conversation":
+    extract conversational blocks, drop the channel envelope, drop injected memory
+    blocks, trim. Used both to decide turn boundaries and to compose context, so
+    the two cannot disagree about what counts as content.
+    """
+    text = _extract_text_content(content, role=role)
+    text = strip_channel_envelope(text)
+    return strip_memory_tags(text).strip()
+
+
 def strip_memory_tags(content: str) -> str:
     """Remove <hindsight_memories> and <relevant_memories> blocks.
 
@@ -86,9 +99,7 @@ def compose_recall_query(
         if role not in allowed_roles:
             continue
 
-        content = _extract_text_content(msg.get("content", ""), role=role)
-        content = strip_channel_envelope(content)
-        content = strip_memory_tags(content).strip()
+        content = _conversational_text(msg.get("content", ""), role=role)
         if not content:
             continue
 
@@ -172,8 +183,14 @@ def slice_last_turns_by_user_boundary(messages: list, turns: int) -> list:
 
     Port of: sliceLastTurnsByUserBoundary() in index.js
 
-    Walks backward counting user messages as turn boundaries. Returns
-    messages from the Nth user boundary to the end.
+    Walks backward counting user messages that carry conversational text.
+    Tool-result-only, empty, and injected-memory-only user records do NOT define
+    a boundary — they are user-role transport records that the conversation
+    extractor already discards, so counting them would exhaust the requested
+    number of turns before reaching any actual conversation. Records inside the
+    selected window are returned untouched; only boundary selection is affected.
+
+    Returns messages from the Nth qualifying user boundary to the end.
     """
     if not isinstance(messages, list) or not messages or turns <= 0:
         return []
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Closing: superseded by **Coding Agents**, where turn boundaries no longer decide what gets retained.

Your diagnosis was right about the old plugin — Anthropic delivers tool results as `role: "user"` records, so counting them as boundaries mis-slices the window. The new plugin sends the unsent *suffix* tracked by a per-session cursor instead of searching for the last N user turns, so there is no boundary heuristic left to get wrong. Tool activity is compacted into `role:"action"` turns for extraction rather than treated as conversation.

PR #3001, opened 2026-07-28
fix(tei): retry HTTP 429 backpressure for reranking and embeddings

### LINKED ISSUE #2991: TEI reranker 429 (permit exhaustion) is not retried — recall fails hard, aborting consolidation
### Bug Description

`RemoteTEICrossEncoder._async_request_with_retry` (engine/cross_encoder.py) retries
connection errors, timeouts and 5xx — but **not HTTP 429**. Since TEI 1.9, 429 is the
*expected* transient signal under load: huggingface/text-embeddings-inference#726
deliberately switched `/rerank` to fail-fast `try_acquire_permit`, **one permit per
text**, no queueing. A single 429 therefore fails the whole recall, which aborts the
surrounding consolidation round.

The defaults on both sides make this self-triggering — no external traffic needed:

    RERANKER_TEI_MAX_CONCURRENT (8) × RERANKER_TEI_BATCH_SIZE (128) = 1024 in-flight texts
    TEI default --max-concurrent-requests               = 512 permits

Consolidation's own recall fan-out (`_process_memory_batch`: one recall per memory,
`asyncio.gather`, up to 3 batches × 128 candidates each) exceeds the pool once a bank
holds enough observations — which consolidation itself creates while running.

### Steps to Reproduce

1. TEI reranker (≥1.9) with default `--max-concurrent-requests 512`
2. Hindsight with `RERANKER_PROVIDER=tei`, default `RERANKER_TEI_MAX_CONCURRENT=8`,
   `RERANKER_TEI_BATCH_SIZE=128`
3. Bank with several hundred unconsolidated memories (ours: 666), observations enabled
4. Trigger consolidation and let it run — as created observations grow the recall
   candidate sets, rerank bursts eventually exceed 512 permits
5. TEI answers 429 ("Model is overloaded"), Hindsight raises immediately

### Expected Behavior

429 is retried with the existing exponential-backoff path (like 5xx), optionally
honoring `Retry-After`. TEI documents fail-fast-with-429 as normal backpressure, so
the TEI provider client should treat it as transient, not fatal.

### Actual Behavior

RuntimeError("TEI rerank request failed: Client error '429 Too Many Requests' ...")
propagates: rerank → recall → consolidation task failure.

On 0.6.2 we watched a consolidation die permanently at 371/666 memories: the worker's
3 retries each re-failed in <1s (retry re-enters the same burst pattern instantly),
then the operation was marked failed. On ≥0.7.2 the indefinite task-level backoff
softens this to "round aborts and restarts later", but every 429 still throws away a
full round of recall/LLM work — the HTTP-level retry would be the targeted fix.

Workaround we run meanwhile: raise TEI `--max-concurrent-requests` above
`RERANKER_TEI_MAX_CONCURRENT × RERANKER_TEI_BATCH_SIZE` (2048 for the default 1024).
Worth documenting as an invariant in the TEI provider docs either way.

### Version

0.6.2-slim (incident); retry code verified unchanged on main / 0.8.5

### LLM Provider

Other

r266-tech:
Fixes #2991.

## Summary

- retry TEI `429 Too Many Requests` responses in both remote reranking and embeddings
- honor numeric and HTTP-date `Retry-After` values within the configured client timeout
- add bounded jitter, including when a server delay reaches the timeout cap
- preserve fail-fast behavior for every other 4xx response and the existing retry-count budget

## Testing

- `29 passed, 3 skipped` across the focused TEI cross-encoder, embeddings, and retry suites
- `ruff check` and `ruff format --check` on all changed files
- `py_compile` on all changed source and test files

The retry delay remains inside the existing request semaphore for reranking, so overload retries do not bypass its concurrency bound.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -397,7 +397,7 @@ class RemoteTEICrossEncoder(CrossEncoderModel):
                         sleep_delay = tei_retry_delay(
                             e.response,
                             delay,
-                            max_delay=self.timeout,
+                            request_timeout=self.timeout,
                         )
                         logger.warning(
                             f"TEI transient error (attempt {attempt + 1}/{self.max_retries + 1}): {e}. "
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -52,6 +52,7 @@ from .local_device import (
     resolve_model_device_type,
     select_local_device,
 )
+from .tei_retry import tei_retry_delay
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


nicoloboschi (approved):
lgtm

PR #3024, opened 2026-07-28
fix(claude-code): filter injected harness turns from retention

### LINKED ISSUE #3023: claude-code: skill bodies and harness-injected blocks are retained as user conversation
## Problem

The Claude Code integration retains content the harness injects into user turns as if the user had written it. Two cases:

1. **Skill bodies.** When a skill is invoked, Claude Code records the full `SKILL.md` as an ordinary user message beginning `Base directory for this skill: <path>`. Auto-retain sends it to extraction.
2. **`<system-reminder>` / `<task-notification>` blocks.** `strip_memory_tags()` removes Hindsight's own `<hindsight_memories>` / `<relevant_memories>` blocks, but nothing removes the harness's.

This is the same class of bug as #1345 (Codex retaining `AGENTS.md` startup messages), fixed in #1346. The Codex path now has `is_synthetic_codex_user_message()`; the Claude Code path has no equivalent.

## Steps to Reproduce

1. Configure the plugin with `retainMode: "full-session"` and auto-retain enabled.
2. In a session, invoke any bundled skill (the size scales with the skill).
3. Let the Stop hook fire.

## Observed Behavior

Measured over my local transcripts, after applying the plugin's own `strip_memory_tags()`:

```
retainable messages: 1413   total chars: 1,328,726
skill-body messages : 1 msg    566,804 chars (42.7%)
injected blocks     : 23 msgs   93,180 chars ( 7.0%)
combined                                      49.7%
```

A single `claude-api` skill invocation accounted for **42.7%** of everything that session sent to extraction — one user message of 566,804 characters. Extraction turned it into ~1,600 world facts about the Anthropic SDK (`budget_tokens` deprecations, Bedrock client setup, effort levels) stored as facts about my work.

Downstream, an untagged whole-bank mental model then synthesised those into a "user profile" asserting model and API preferences I had never expressed. That is the part that took a while to notice — the facts look plausible and nothing marks them as documentation.

Another user reproduced the same failure independently on a separate setup with a 785 KB skill payload.

## Expected Behavior

Skill bodies should not be retained as user conversation, and harness-injected blocks should be stripped the same way Hindsight's own injected blocks already are.

## Environment

`hindsight-memory` plugin 0.7.5, self-hosted API, `retainMode: full-session`, `retainToolCalls: false`.

PR follows.

runbgp:
Fixes #3023.

Claude Code records an invoked skill's full `SKILL.md` as an ordinary user message, and injects `<system-reminder>` / `<task-notification>` blocks into user turns. Both were retained verbatim. On my transcripts that was ~50% of everything sent to extraction, and one skill invocation alone was a single 566 KB user message.

Same shape as the Codex fix in #1346:

- `is_synthetic_user_message()` — skips user turns starting with `Base directory for this skill: `, applied once in `prepare_retention_transcript()` so it covers both the text and JSON formats.
- `strip_harness_blocks()` — removes the two block types, alongside the existing `strip_memory_tags()` call on the retain path.

Recall query composition is left alone; this only touches retention.

Tests added for both functions and for the retention path. `uv run pytest tests` passes (213). No new ruff findings — the two pre-existing `F541`s in `setup_hooks.py` and the import-sort in `test_content.py` are untouched.

### DIFF
--- a/hindsight-integrations/claude-code/scripts/lib/content.py
+++ b/hindsight-integrations/claude-code/scripts/lib/content.py
@@ -48,6 +48,28 @@ def strip_memory_tags(content: str) -> str:
     return content
 
 
+def strip_harness_blocks(content: str) -> str:
+    """Remove <system-reminder> and <task-notification> blocks.
+
+    Claude Code injects these into user turns. They are harness output rather
+    than conversation, so retaining them teaches Hindsight about hook and tool
+    plumbing instead of the user's work.
+    """
+    content = re.sub(r"<system-reminder>[\s\S]*?</system-reminder>", "", content)
+    content = re.sub(r"<task-notification>[\s\S]*?</task-notification>", "", content)
+    return content
+
+
+def is_synthetic_user_message(content: str) -> bool:
+    """Return True for skill bodies Claude Code persists as user chat.
+
+    Claude Code records an invoked skill's full SKILL.md as a normal user
+    message. Retaining it stores documentation as if the user had written it,
+    and can create very large retain payloads.
+    """
+    return content.lstrip().startswith("Base directory for this skill: ")
+
+
 # ---------------------------------------------------------------------------
 # Recall: query composition and truncation
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Closing: superseded by **Coding Agents**, which handles this class structurally — and your report is part of why it now handles it properly.

Measured against 400 real transcripts there: skill bodies arrive with `isMeta: true` (8/8) and are dropped with every other meta line, and `<system-reminder>` rides inside `tool_result` blocks, which the reader drops entirely. What *was* leaking were two cases neither of us had listed — `<task-notification>` (39 records) and Claude Code's compaction summary (29 records, 474,016 chars, retained as if the user typed them). Both are fixed: #3378 and #3383.

Issue #1641, opened 2026-05-16
[Bug] hindsight-embed local_embedded fails to start: TypeError: FastMCP() no longer accepts 'stateless_http' with fastmcp >= 3.0


laychiman:
# Bug Report: FastMCP Compatibility — hindsight-embed 0.6.2 fails with TypeError on FastMCP(stateless_http=True)

## Summary
hindsight-local-embedded mode (hindsight-embed 0.6.2) **cannot start the API daemon** because it passes `stateless_http=True` to `FastMCP(...)` which is removed in fastmcp >= 3.0. Downgrading fastmcp causes cascading dependency breaks.

## Environment
- hindsight-embed: 0.6.2
- hindsight-api: 0.3.0 (installed via hindsight-all)
- fastmcp: 3.3.1 (latest)
- python: 3.11
- OS: macOS 15.x, Intel Mac

## Steps to Reproduce
```bash
pip install hindsight-all        # installs hindsight-api 0.3.0 + hindsight-embed 0.6.2
pip install fastmcp==3.3.1       # latest
export HINDSIGHT_API_MCP_ENABLED=false
cd ~/.hermes/hermes-agent
source venv/bin/activate
python3 -c "from hindsight_embed.daemon_embed_manager import DaemonEmbedManager; DaemonEmbedManager(profile='hermes').ensure_running()"
```

## Expected Behavior
Daemon starts, health check passes on allocated port.

## Actual Behavior
Daemon crashes with:
```
TypeError: FastMCP() no longer accepts 'stateless_http'
```
(Location: hindsight_api/api/mcp.py line ~53)

## Attempted Workarounds
1. HINDSIGHT_API_MCP_ENABLED=false — Still fails because mcp.py is imported at startup before the gate checks
2. fastmcp==2.12.0 — ImportError: PrivateKeyJWTAuthenticator missing
3. fastmcp==2.3.0 — ModuleNotFoundError: fastmcp.mcp_config

## Root Cause
- hindsight-api pins fastmcp>=2.3.0 but does NOT cap <3.0
- hindsight_api/api/mcp.py uses stateless_http=True (removed in fastmcp 3.x)
- HINDSIGHT_API_MCP_ENABLED gate does not prevent import-time errors

## Suggested Fixes
A. Update mcp.py for fastmcp 3.x API compatibility
B. Hard-pin fastmcp<3.0 in pyproject.toml until migration
[...]

PR #3027, opened 2026-07-28
fix(claude-code): pin mcp<2 so the MCP server can import FastMCP

### LINKED ISSUE #3026: claude-code: MCP server fails to start on fresh installs — requirements.txt admits mcp 2.0, which removed mcp.server.fastmcp
## Problem

`hindsight-integrations/claude-code/requirements.txt` declares `mcp>=1.0.0`. `mcp` 2.0.0 (PyPI, 2026-07-28T13:45Z) removed `mcp.server.fastmcp`, which `scripts/mcp_server.py` imports at module scope. Fresh venv bootstraps resolve 2.0.0, the server exits on import, and Claude Code reports `-32000: Connection closed` with no `agent_knowledge_*` tools.

No shim exists: `hasattr(mcp.server, "FastMCP")` is `False` in 2.x, and the equivalent is now `mcp.server.mcpserver.MCPServer`.

Presents as "worked yesterday, broken today" — `run_mcp.sh` re-pips only when the requirements cache drifts or `import mcp` fails, and `import mcp` still succeeds on 2.0.0. Existing venvs keep working; only fresh installs and cleared `$CLAUDE_PLUGIN_DATA/venv` break.

## Steps to Reproduce

1. `rm -rf "$CLAUDE_PLUGIN_DATA/venv"`
2. Run `scripts/run_mcp.sh` (or restart Claude Code)
3. `claude mcp list`

## Observed Behavior

```
  File ".../scripts/mcp_server.py", line 26, in <module>
    from mcp.server.fastmcp import FastMCP
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

## Expected Behavior

`mcp>=1.0.0,<2` resolves to 1.29.0 and the server starts normally. Same shape as #1641 — unbounded upper range on a dependency that breaks in a major.

`claude-agent-sdk/pyproject.toml:63` and `hindsight-integration-tests/pyproject.toml:15` declare the same range, but neither imports `fastmcp`, so I can't show they're affected.

## Related

Separate cause, identical `-32000`, worth its own guard: where `python3 -m venv` can't seed pip (Python 3.14 without `python3.14-venv`), `bin/python` is created but `bin/pip` isn't. `resolve_py()` only tests `[ -x bin/python ]`, so the script proceeds and dies at `"${PIP}" install` under `set -e`. A `[ -x "${PIP}" ]` check would make that legible.

## Environment

Plugin 0.7.5, self-hosted API, Ubuntu, `/usr/bin/python3` 3.14.4.

runbgp:
Fixes #3026.

`mcp` 2.0.0 (2026-07-28) removed `mcp.server.fastmcp`, which `scripts/mcp_server.py` imports at module scope. Under `mcp>=1.0.0`, fresh venv bootstraps resolve 2.0.0 and the server exits on import — surfaced as `-32000: Connection closed` with no `agent_knowledge_*` tools.

Existing venvs are unaffected: `run_mcp.sh` re-pips only on cache drift or `import mcp` failure, and that still succeeds on 2.0.0. Only fresh installs break.

`<2` resolves to 1.29.0. Verified on Python 3.14.4 — import works, `run_mcp.sh` starts and exits 0 on stdin close, `claude mcp list` reports connected.

Same shape as #1641. Left the identical unbounded range in `claude-agent-sdk` and `hindsight-integration-tests` alone; neither imports `fastmcp`, noted in the issue.


xyref:
Hit this today on a fresh 0.7.5 install (venv bootstrapped 2026-08-04, Python 3.14, Linux) — same `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`, surfacing in Claude Code as the hindsight MCP server failing to start with no error visible to the user.

Independently verified the fix in this PR: installing `mcp>=1.0.0,<2.0.0` into the existing plugin venv resolves to 1.29.0, the server starts, and it completes the MCP `initialize` handshake. As the description says, existing venvs keep working because `run_mcp.sh`'s `import mcp` probe still passes on 2.0.0 — so this silently bites every fresh install until merged. Would be great to get this one-liner in.

handnewb (approved):
LGTM! mcp 2.0.0 dropped FastMCP — fresh installs break with `-32000: Connection closed`. 1-line pin, verified on Python 3.14.4, same pattern as #1641. The two other unbounded ranges (claude-agent-sdk, integration-tests) are correctly left alone since they don't import fastmcp.

nicoloboschi:
Closing: superseded by **Coding Agents**, where this cannot happen — it ships no Python at all.

The MCP server is TypeScript launched with `node` (`command: "node"`, absolute script path), so there is no `requirements.txt`, no `mcp` resolution to drift onto 2.x, and no `mcp.server.fastmcp` import to break. The same change also removes the Windows failure where `"command": "bash"` resolved to the WSL launcher (#3132).

`npx @vectorize-io/hindsight-coding-agents install claude-code` ([docs](https://hindsight.vectorize.io/sdks/integrations/coding-agents)). Thanks for the sharp root-cause — "worked yesterday, broken today" from an unpinned floor is a nasty one to catch.

PR #3029, opened 2026-07-28
fix(retain): unwrap Vertex AI Claude parameter envelope in fact extraction

### LINKED ISSUE #3028: fix: fact extraction returns 0 facts with litellm + Vertex AI Claude
## Bug

When using `litellm` provider with `vertex_ai/claude-*` models, fact extraction silently returns 0 facts for every retain operation. The LLM successfully extracts facts (visible in `llm-requests` logs and `output_tokens` > 0), but they are never stored.

## Root Cause

Vertex AI Claude returns structured output (tool-use responses) wrapped in a `{"parameter": {...}}` envelope:

```json
{"parameter": {"facts": [{"what": "...", "fact_type": "world", ...}]}}
```

But `_coerce_fact_response()` in `fact_extraction.py` passes the dict through unchanged, and line 1452 does:

```python
raw_facts = extraction_response_json.get("facts", [])
```

Since `"facts"` is nested inside `"parameter"`, this returns `[]` — silently dropping all extracted facts.

## Impact

- All retains via `litellm` + `vertex_ai/claude-*` produce 0 facts
- The `retain` API returns `{"success": true}` (misleading — the document is stored but empty)
- `dry-run-extract` also returns `{"facts": []}` despite the LLM generating valid output
- Affects any deployment using GCP Vertex AI with Claude models through litellm

## Reproduction

```bash
# Configure with litellm + vertex_ai
export HINDSIGHT_API_LLM_PROVIDER=litellm
export HINDSIGHT_API_LLM_MODEL=vertex_ai/claude-opus-4-6

# Retain something
curl -X POST http://localhost:8888/v1/default/banks/test/memories \
  -d '{"items": [{"content": "The project uses Python 3.11 and PostgreSQL 16."}]}'

# Check LLM requests — output_tokens > 0 but facts = []
curl http://localhost:8888/v1/default/banks/test/llm-requests?limit=1

# Check memories — 0 facts stored
curl http://localhost:8888/v1/default/banks/test/memories/list
```

## Fix

Two-line change in `_coerce_fact_response()` to unwrap the `{"parameter": {...}}` envelope before returning:

```python
def _coerce_fact_response(response: Any) -> dict[str, Any] | None:
    """Accept the schema wrapper, or a recoverable top-level facts array."""
    if isinstance(response, dict):
        if "parameter" in response and isinstance(response["parameter"], dict):
            return response["parameter"]
        return response
    if isinstance(response, list) and all(isinstance(item, dict) for item in response):
        return {"facts": response}
    return None
```

Verified fix working — retains now correctly extract and store facts with litellm + Vertex AI Claude.

Will submit PR with fix + tests.

yossiovadia:
## Summary

- Fixes silent 0-fact extraction when using `litellm` with `vertex_ai/claude-*` models
- Vertex AI Claude wraps tool-use structured output in `{"parameter": {...}}` — the parser now unwraps this envelope before extracting facts
- Adds 6 unit tests for `_coerce_fact_response()` covering the envelope unwrap, normal passthrough, bare lists, and invalid inputs

## Problem

Retains via `litellm` + `vertex_ai/claude-*` silently produce 0 facts. The LLM successfully extracts facts (visible in `llm-requests` with `output_tokens > 0`), but `_coerce_fact_response()` returns the full envelope `{"parameter": {"facts": [...]}}` and `extraction_response_json.get("facts", [])` returns `[]` because `"facts"` is nested one level deeper.

## Test plan

- [x] `pytest tests/test_coerce_fact_response_parameter_wrapper.py -v` — 6/6 pass
- [x] Manual verification: retain with `litellm` + `vertex_ai/claude-opus-4-6` now extracts facts (was 0, now 3 facts from same input)
- [x] Existing tests unaffected — normal `{"facts": [...]}` responses pass through unchanged

Fixes #3028

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -1314,6 +1314,8 @@ def _build_request_body(llm_config, config, prompt: str, user_message: str, resp
 def _coerce_fact_response(response: Any) -> dict[str, Any] | None:
     """Accept the schema wrapper, or a recoverable top-level facts array."""
     if isinstance(response, dict):
+        if "parameter" in response and isinstance(response["parameter"], dict):
+            return response["parameter"]
         return response
     if isinstance(response, list) and all(isinstance(item, dict) for item in response):
         return {"facts": response}
--- a/hindsight-api-slim/tests/test_coerce_fact_response_parameter_wrapper.py
+++ b/hindsight-api-slim/tests/test_coerce_fact_response_parameter_wrapper.py
@@ -1,0 +1,66 @@
+"""
+Test that _coerce_fact_response unwraps the ``{"parameter": {...}}`` envelope
+that Vertex AI Claude returns for tool-use structured output via litellm.
+
+Without this fix, retains via ``litellm`` + ``vertex_ai/claude-*`` silently
+produce 0 facts because the extracted facts sit inside ``response["parameter"]``
+and ``response.get("facts", [])`` returns an empty list.
+"""
+
+from hindsight_api.engine.retain.fact_extraction import _coerce_fact_response
+
+
+def test_unwraps_parameter_envelope():
+    """Vertex AI Claude wraps tool output in {"parameter": {"facts": [...]}}."""
+    response = {
+        "parameter": {
+            "facts": [
+                {"what": "version is 1.0", "who": "user", "fact_type": "world"},
+            ]
+        }
+    }
+    result = _coerce_fact_response(response)
+    assert "facts" in result
+    assert len(result["facts"]) == 1
+    assert result["facts"][0]["what"] == "version is 1.0"
+
+
+def test_passthrough_normal_response():
+    """Normal {"facts": [...]} responses pass through unchanged."""
+    response = {"facts": [{"what": "something", "fact_type": "world"}]}
+    result = _coerce_fact_response(response)
+    assert result is response
+
+
+def test_passthrough_empty_facts():
+    """Empty facts list passes through."""
+    response = {"facts": []}
+    result = _coerce_fact_response(response)
+    assert result is response
+
+
+def test_coerces_bare_list():
+    """A bare list of fact dicts is wrapped into {"facts": [...]}."""
+    response = [{"what": "a"}, {"what": "b"}]
+    result = _coerce_fact_response(response)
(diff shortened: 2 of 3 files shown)


nicoloboschi:
Thanks for putting this together, and sorry to close it after the work you put in.

Closing alongside #3028 — full reasoning there. The short version: the `{"parameter": {...}}` envelope isn't documented by Anthropic, Google, or litellm, and it's a known *unstable* Claude json-mode behavior rather than a Vertex contract (https://github.com/BerriAI/litellm/issues/8898 reports the same wrapping under `json_input`, `content`, and `data`, and was closed as not planned). Hardcoding one key leaves the next one silently dropping facts.

The bug you surfaced underneath is real, though. A dict without a `facts` key silently returns zero facts, while a non-dict response raises two lines earlier citing #1833 — "never silently commit the document with 0 facts." Making those consistent (retry, then fail loudly) fixes the whole family rather than one member, and turns the silent data loss into a visible error. We'll handle that separately.

If you can attach the raw `llm-requests` payload from your deployment — request body plus the unparsed response content, and whether it happens on every call or intermittently — that would be very useful. We have no live Vertex Claude access to reproduce against.

yossiovadia:
Reworked per your feedback — no longer hardcodes `"parameter"`. 

The new approach: unwrap any **single-key dict** whose sole value is a dict containing `"facts"`. This handles `{"parameter": {...}}`, `{"content": {...}}`, `{"data": {...}}`, etc. — all without chasing individual wrapper keys.

**Constraints** (so it doesn't fire on normal responses):
- Only unwraps if the outer dict has **exactly one key** (a multi-key dict is not an envelope)
- Only unwraps if the inner value is a **dict containing "facts"**
- Direct `{"facts": [...]}` responses pass through unchanged

```python
if "facts" not in response and len(response) == 1:
    inner = next(iter(response.values()))
    if isinstance(inner, dict) and "facts" in inner:
        return inner
```

Tests updated: 8 cases covering the envelope unwrap for various key names, passthrough for normal responses, multi-key dict preservation, and single-key without facts.

I also posted raw `llm-requests` payloads on #3028 — 17/17 `retain_extract_facts` calls show the `{"parameter": {...}}` wrapping, 100% consistent, only on fact extraction (consolidation doesn't wrap). Happy to provide more data.

Could you reopen this?

yossiovadia:
Superseded by #3084 — fresh PR rebased on current main with the key-agnostic approach.

PR #3038, opened 2026-07-29
docs(reflect): clarify tag-scoped directive behavior

Sanderhoff-alt:
## Summary

This PR documents the tag-scoping contract used by `reflect` consistently across REST/OpenAPI descriptions, MCP tool descriptions, engine docstrings, public documentation, versioned v0.8 documentation, generated SDKs, and the bundled agent skills.

It does not change directive loading or other runtime behavior. The separate mental-model exact-empty-scope consistency fix is isolated in #3039.

## Background and root cause

Issue #3031 reports that an unscoped `reflect` call does not load tagged directives because `reflect` calls `list_directives(..., isolation_mode=True)`.

That runtime behavior is intentional: directive tags are execution scopes. Untagged directives are global, tagged directives apply only when the reflect request supplies a matching scope, and an unscoped reflect call loads only global directives. Direct directive-management calls use `isolation_mode=False`, so listing directives without a tag filter returns both global and tagged directives.

The public contract did not explain this distinction. Existing documentation described directives as always enforced, described directive tags as organizational labels, and described reflect tags as filtering memories without mentioning observations, mental models, or directives. OpenAPI, MCP, and engine docstrings also omitted supported `tags_match` modes and did not explain empty-tag behavior.

## Changes

- Document that reflect `tags` scope raw facts, observations, mental models, and tagged directives.
- Document that untagged directives are global and remain applicable with `any`, `all`, strict, exact, and compound tag scopes.
- Document that an unscoped reflect call searches all memory scopes but loads only untagged/global directives.
- Explain that `isolation_mode` is an internal directive-listing option and is not a public reflect request parameter.
- Add behavior matrices for omitted tags, empty tags, every `tags_match` mode, `tag_groups`, and directive loading.
- Clarify that public REST and MCP directive-listing behavior differs from reflect directive loading when no tag filter is supplied.
- Clarify that public `tags` and `tag_groups` inputs are mutually exclusive and that each compound tag-group leaf owns its matching mode.
- Document the exact-mode corner case: when `tag_groups` is used without flat tags, top-level `tags_match: "exact"` adds a global flat constraint to facts, observations, and mental models before the compound expression is applied. The normal compound-only configuration leaves top-level `tags_match` at its default, `any`.
- Add copyable examples for unscoped, scoped, strict, and exact-global reflect requests.
[...]

Issue #3031, opened 2026-07-29
reflect still drops every tagged directive in v0.8.5 (isolation_mode applied to directives) — refiling #1269


mekenthompson:
## Summary

`reflect` loads its directives with `isolation_mode=True`, which — when the
caller passes no `tags` — filters the directive set down to `tags IS NULL OR
tags = '{}'`. Every directive carrying at least one tag is therefore silently
dropped from synthesis: it is absent from the system prompt and absent from
`directives_applied` in the trace.

This was reported as #1269 and closed 2026-04-26 because we withdrew it on our
side. The defect was never fixed, and it is still present in **v0.8.5**
(`ghcr.io/vectorize-io/hindsight:latest@sha256:0710076cd1539b4f89537d2e5b0ea1e4d179885bc12809a56cf748c538c8c4fb`,
`org.opencontainers.image.version=0.8.5`). Re-filing with fresh evidence and a
narrower ask than #1269's.

## Verified in the shipping image

`/app/api/hindsight_api/engine/memory_engine.py`, `reflect_async` (~:9390):

```python
directives_raw = await self.list_directives(
    bank_id=bank_id,
    tags=tags,
    tags_match=tags_match,
    tag_groups=tag_groups,
    active_only=True,
    request_context=request_context,
    isolation_mode=True,
)
```

`list_directives` (~:11655):

```python
if not tags and not tag_groups and isolation_mode:
    # Isolation mode: only include directives with empty/null tags
    filters.append("(tags IS NULL OR tags = '{}')")
```

Driven against a real database inside that image, with one untagged and one
`tags=["scoped"]` active directive in the bank:

```
isolation_mode=True   -> ['untagged-rule']                  # what reflect gets
isolation_mode=False  -> ['tagged-rule', 'untagged-rule']
```

Downstream, `_build_directives_applied()` and `build_directives_section()` both
receive only the untagged one, so the trace and the prompt agree with each other
and are both short.

## Impact

Measured across a 5-bank deployment: 172 active directives, 94 untagged. 45% of
the operator's standing rules never reach a `reflect` prompt, with no error, no
warning, and a `[REFLECT …] Loaded N directives` log line that reports the
post-filter count as though it were the whole set.

## The ask

`isolation_mode` reads as a *memory*-scoping concept — "don't let tag-scoped
content leak into an untagged operation". Applying it to **directives** inverts
their documented semantics, which the non-isolation branch right above already
states correctly:

```
[...]

PR #3039, opened 2026-07-29
fix(reflect): apply exact empty scope to mental models

Sanderhoff-alt:
## Summary

Make mental-model retrieval honor `tags_match: "exact"` when tags are omitted or supplied as an empty list.

## Root cause

Reflect uses the same tag scope for raw facts, observations, and mental models. Facts and observations pass their filters through the central recall path, where absent or empty tags with `tags_match: "exact"` mean the untagged/global scope.

The mental-model tool conditionally built its SQL tag clause only when `tags` was truthy. Both `None` and an empty list are falsey, so the tool skipped filtering entirely even though `exact` gives the empty set a precise meaning. As a result, one reflect operation could retrieve only global facts and observations while still exposing tagged mental models from every scope.

## Change

Apply the centralized tag filter when tags are non-empty or when the matching mode is `exact`. This preserves the existing no-filter behavior for absent or empty tags with `any`, `all`, `any_strict`, and `all_strict`, while making the exact empty set select only rows whose tags are `NULL` or `{}`.

Add focused parameterized regression coverage for both omitted tags (`None`) and an explicit empty list. The test captures the generated query and verifies the global-scope predicate and parameter ordering.

## Behavior

| Request | Mental models before | Mental models after |
| --- | --- | --- |
| Tags omitted, default matching | All scopes | All scopes |
| `tags: []`, default matching | All scopes | All scopes |
| Tags omitted, `tags_match: "exact"` | All scopes | Global scope only |
| `tags: []`, `tags_match: "exact"` | All scopes | Global scope only |
| Non-empty tags | Matching behavior unchanged | Matching behavior unchanged |

## Validation

- `./scripts/hooks/lint.sh`
- `uv run pytest tests/test_reflect_tools.py -q` (`18 passed`)
- Pre-commit hooks
- `git diff --check`

Discovered while clarifying the tag-scope contract discussed in #3031. The directive-isolation behavior from that issue is not changed by this PR. The companion documentation clarification is in #3038.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/tools.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/tools.py
@@ -95,8 +95,10 @@ async def tool_search_mental_models(
     params: list[Any] = [bank_id, str(query_embedding), max_results]
     next_param = 4
 
-    # Use the centralized tag filtering logic
-    if tags:
+    # Exact matching treats absent or empty tags as the global scope. Do not
+    # skip the filter, or mental models would see every scope while the other
+    # reflect retrieval tools correctly see only untagged data.
+    if tags or tags_match == "exact":
         tag_clause, tag_params, next_param = build_tags_where_clause(tags, param_offset=next_param, match=tags_match)
         filters += f" {tag_clause}"
         params.extend(tag_params)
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Correct, minimal, well-tested fix. The one-line gate change aligns `tool_search_mental_models` with `build_tags_where_clause` (which already handles the exact-empty global scope with no bind param, so param ordering is preserved) and with the established convention in `link_expansion_retrieval.py`. Default behavior (any-match, empty tags) is unchanged. Parametrized regression test pins both `None` and `[]`. LGTM.

PR #3041, opened 2026-07-29
feat(memory): add authorized memory unit deletion

### LINKED ISSUE #3006: Expose authorized single and bulk memory-unit deletion APIs
## Summary

Hindsight already implements two hard-delete primitives on `MemoryEngine`:

- `delete_memory_unit(unit_id, *, request_context)`
- `delete_memory_units(unit_ids, *, request_context)`

Neither method is currently exposed through the public `MemoryEngineInterface`, REST API, MCP tools, CLI, or generated SDKs. Both authenticate the tenant, but neither invokes `OperationValidatorExtension.validate_bank_write`, so they are not safe to expose as-is.

Please add bank-write authorization to both engine methods and expose supported, bank-scoped public APIs for single and bulk memory-unit deletion.

### Single-unit deletion

`MemoryEngine.delete_memory_unit` was introduced in commit `39801d9f8` on 2025-11-05 by Nicolò Boschi.

It was previously exposed through `MemoryEngineInterface` and an MCP `delete_memory` tool. That public surface was intentionally removed in commit `90674aef1` / #1228 on 2026-04-23 because the method did not call `validate_bank_write`. An authenticated MCP client could therefore bypass `OperationValidatorExtension` policy and delete memories in an otherwise protected bank.

The engine method was retained because its stale-observation cleanup, cache invalidation, and graph-maintenance behavior are useful and covered by internal tests. The authorization gap still exists today.

### Bulk deletion

`MemoryEngine.delete_memory_units` was added as an engine primitive in `64fe5e81f` / #2659 on 2026-07-24 by Chris Latimer, co-authored by Nicolò Boschi. Its stated purpose is to provide retention loops, LRU eviction, and bulk-maintenance tools with a single deletion seam that preserves referential-integrity behavior.

The bulk method has never been publicly exposed. It also calls `_authenticate_tenant` but not `validate_bank_write`.

## Current behavior

Both methods perform irreversible physical deletion from `memory_units`; they do **not** use the reversible `invalidated_memory_units` archive/state flow.

They also perform or trigger related lifecycle work, including:

- FK cascade cleanup of links and entity associations;
- deletion/invalidation of observations that reference deleted source memories;
- bank stats cache invalidation;
- consolidation scheduling where necessary;
- graph-maintenance scheduling and relink-victim handling.

Current repository usage:

- no production caller invokes either method;
- `delete_memory_unit` has behavior coverage in observation-invalidation, cache-invalidation, and Oracle tests;
- `delete_memory_units` currently has only basic empty-input and UUID-validation tests;
- the existing `DELETE /v1/default/banks/{bank_id}/memories` endpoint clears a whole bank (optionally by fact type) via `delete_bank`; it is not an ID-based single/bulk deletion endpoint.

### 1. Add authorization inside the engine boundary

Authorization should live in the engine methods, not only in HTTP/MCP handlers, so every future internal or external caller receives the same enforcement.

For `delete_memory_unit`:

1. authenticate the tenant;
2. resolve the target unit and its `bank_id` before mutation;
3. construct the appropriate `BankWriteContext` for a destructive memory-unit operation;
4. call `validate_bank_write` through `_validate_operation`;
5. delete only after authorization succeeds.

For `delete_memory_units`:

1. validate and resolve all supplied IDs first;
2. determine every distinct affected bank;
3. authorize every affected bank before performing any deletion;
4. fail the entire request before mutation if any bank is denied, avoiding partial authorization/partial deletion;
5. preserve the existing per-bank deduplication of consolidation and graph-maintenance work.

If the public endpoint is bank-scoped, also reject IDs that resolve to a different bank instead of silently deleting across banks.

Add or select an explicit `BankWriteOperation` value for hard deletion so extensions can distinguish it from retain/update/invalidation operations.

### 2. Add supported public REST APIs

Suggested bank-scoped shapes:

```http
DELETE /v1/default/banks/{bank_id}/memories/{unit_id}
```

and

```http
POST /v1/default/banks/{bank_id}/memories/bulk-delete
Content-Type: application/json

{"unit_ids": ["...", "..."]}
```

A POST action for bulk deletion avoids interoperability issues around DELETE request bodies. Exact naming can follow project API conventions.

Both endpoints should clearly document that this is irreversible hard deletion and is different from setting memory state to `invalidated`.

### 3. Publish the supported client surface

- add the methods to `MemoryEngineInterface` after authorization is enforced;
- regenerate OpenAPI clients;
- expose async/sync Python client methods and generated TypeScript/Go equivalents as applicable;
- update API documentation;
- reconcile the Rust client README, which currently mentions `delete_memory_unit()` even though no corresponding Rust client implementation exists;
- only reintroduce an MCP deletion tool if it uses the authorized engine path and the intended MCP safety policy explicitly permits destructive deletion.

## Acceptance criteria

- [ ] Both engine methods call `validate_bank_write` before mutation.
- [ ] Single deletion resolves and validates the target bank.
- [ ] Bulk deletion validates all distinct banks before deleting anything.
- [ ] A denied operation leaves all target units unchanged.
- [ ] Cross-bank IDs are rejected by a bank-scoped endpoint.
- [ ] Tenant isolation is covered by tests.
- [ ] Allow/deny `OperationValidatorExtension` behavior is covered for both methods.
- [ ] Bulk deletion has behavioral tests beyond UUID validation, including cascade cleanup and per-bank maintenance scheduling.
- [ ] Public REST endpoints are added and documented as irreversible hard delete.
- [ ] OpenAPI and supported SDKs are regenerated.
- [ ] Public methods are added to `MemoryEngineInterface` only after authorization enforcement.
- [ ] Existing observation cleanup, cache invalidation, consolidation, and graph-maintenance behavior remains covered.

## Security note

This should not be implemented by exposing the existing methods only at a handler layer. The original MCP authorization bypass (#1218/#1228) demonstrates why the authorization check needs to be enforced at the engine boundary itself.

Sanderhoff-alt:
### Context

Hindsight already had engine-level primitives for deleting memory units, but there was no supported public path that combined selective hard deletion with tenant resolution, bank-scoped authorization, typed responses, generated clients, and end-user tooling. Exposing the primitive directly from a transport would make it too easy to bypass the same write-policy checks used by other bank mutations, while forcing users to clear an entire bank is too coarse for privacy, retention, and data-correction workflows.

This change introduces an explicitly authorized hard-delete capability for one or more memory units while preserving the existing reversible curation path. It closes #3006.

### Goals

- Expose bank-scoped single and bulk memory-unit hard deletion over HTTP.
- Enforce tenant isolation and bank-write authorization inside the engine before any mutation.
- Preserve the existing cascade, observation invalidation, consolidation, cache, and graph-maintenance behavior.
- Reuse the existing typed delete response and expose only the deletion count needed by bank-scoped callers.
- Make the capability available through generated SDKs and the Rust CLI.
- Document when hard deletion should be used instead of reversible invalidation.

### Non-goals

- This PR does not add hard-delete controls to the control plane.
- This PR does not add an MCP deletion tool because autonomous destructive actions need a separate explicit safety policy and opt-in decision.
- This PR does not change the semantics of `PATCH ... {"state": "invalidated"}` or whole-bank clearing.

#### Delete one memory

`DELETE /v1/default/banks/{bank_id}/memories/{memory_id}`

The target is fully identified by the URL, so the standard `DELETE` method is used. A successful request reuses the existing `DeleteResponse` with `success`, `message`, and `deleted_count`; an unknown ID returns `404`, and an ID that resolves to another bank is rejected.

#### Delete multiple memories

`POST /v1/default/banks/{bank_id}/memories/bulk-delete`

```json
{
  "unit_ids": [
    "11111111-1111-1111-1111-111111111111",
    "22222222-2222-2222-2222-222222222222"
  ]
}
```

The bulk operation uses `POST` because request content on `DELETE` has no generally defined semantics and is handled inconsistently by intermediaries, client generators, gateways, and WAFs. A command-style `POST` keeps the JSON request portable and leaves room for future command metadata without encoding an ID list in the URL.

The response reuses `DeleteResponse` and reports the number of deleted units through `deleted_count`. Unknown IDs are omitted from that count, preserving the previous engine primitive.s idempotent behavior; callers can compare the submitted ID count with `deleted_count` when they need to detect already-missing units. Per-bank deletion and observation-invalidation counts remain internal engine results because the public endpoint is already scoped to one bank. This avoids three endpoint-specific response schemas and substantially reduces generated SDK code.

### Authorization and tenant isolation

Authorization is enforced in `MemoryEngine`, not in the FastAPI handlers, so every caller uses the same security boundary.

1. Validate every supplied UUID before entering the mutation path.
2. Authenticate the request context and resolve the tenant schema.
3. Resolve and lock the actual target rows with `SELECT ... FOR UPDATE`.
4. Verify that every resolved row belongs to the bank supplied by the public endpoint.
5. Call `validate_bank_write` with `BankWriteOperation.DELETE_MEMORY_UNIT` for every distinct affected bank.
6. Start mutation only after all required authorization checks succeed.

Bulk rows are locked in deterministic ID order and all affected banks are authorized before the first delete. A denial therefore leaves the complete batch unchanged instead of producing a partially deleted result. The engine resolves the actual bank from each row rather than trusting caller-supplied scope, which protects internal callers as well as HTTP callers.

Single and bulk deletion intentionally share `DELETE_MEMORY_UNIT`. The permission models the destructive capability rather than the transport shape; deployments that later need a stricter bulk-only policy can introduce a separate operation without changing the endpoint contract.

### Data integrity and maintenance

Deletion remains transactional for the database state. Existing foreign-key and engine cleanup removes `unit_entities` and `memory_links`, updates source-observation relationships, and invalidates derived observations that depended on deleted facts. Survivors remain untouched.

After commit, the engine invalidates relevant caches and schedules consolidation and graph maintenance once per affected bank. These maintenance submissions preserve the existing best-effort behavior while avoiding duplicate jobs for multiple units in the same bank.

### SDK and CLI surface

- Regenerate Python, TypeScript, Go, and build-time Rust clients from OpenAPI.
- Add a typed bulk request model and reuse the existing delete response for both endpoints.
- Add synchronous and asynchronous Python convenience methods: `delete_memory_unit`, `adelete_memory_unit`, `delete_memory_units`, and `adelete_memory_units`.
- Restore the existing Rust CLI `memory delete` command and allow one or more positional IDs.
- Route one CLI ID to the single endpoint and multiple IDs to the bulk endpoint.
- Require interactive confirmation by default and support `-y/--yes` for automation.

```bash
hindsight memory delete <bank_id> <memory_id>
hindsight memory delete <bank_id> <memory_id_1> <memory_id_2>
hindsight memory delete <bank_id> <memory_id_1> <memory_id_2> --yes
```

### Compatibility

The REST surface is additive. The engine interface accepts an optional bank scope so existing internal callers keep their current call shape, while public HTTP routes always supply a bank and the engine always resolves and authorizes the actual affected bank. Bulk deletion keeps the prior missing-ID behavior and adds structured counts rather than changing it to all-or-nothing existence validation.

### Documentation

The OpenAPI specification, API operations guide, dedicated memories guide, CLI reference, Python and TypeScript SDK guides, Go and Rust client READMEs, runnable examples, and generated Hindsight documentation skill are updated. The documentation distinguishes permanent erasure from reversible invalidation and records the `POST` bulk-design rationale.

### Validation

- Targeted API deletion and authorization tests: 11 passed.
- Python convenience-client tests: 4 passed.
- Rust CLI full suite: 115 passed, including real request-shape tests for single and bulk routing plus default confirmation.
- Go build and offline tests passed.
- TypeScript build passed.
- Rust generated client build passed with `--locked`.
- CLI OpenAPI coverage passed for all 78 operations and 95 request-body parameters.
- Repository lint passed.
- Documentation production build passed; it reports only the existing broken-anchor warnings in the 0.5.5 blog post.

### Modified scope

- Engine interface, deletion implementation, and operation-validator capability.
- FastAPI single and bulk routes, a typed bulk request, and the shared delete response.
- Generated Python, TypeScript, Go, and Rust client surfaces.
- Python convenience client and Rust CLI.
- Regression, tenant-isolation, cascade, maintenance, SDK, and CLI tests.
- OpenAPI, user-facing documentation, examples, and generated documentation skill.


nicoloboschi:
we have /invalidate already. at most we can support more memories in bulk there

PR #3042, opened 2026-07-29
fix(clients): expose mental model query controls

### LINKED ISSUE #2975: TypeScript client wrapper drops detail parameter on listMentalModels / getMentalModel
# TypeScript client wrapper drops `detail` parameter on listMentalModels / getMentalModel

## Summary

The hand-written `HindsightClient` wrapper in `@vectorize-io/hindsight-client` does not forward the `detail` query parameter to the low-level SDK. The parameter exists in the generated OpenAPI types (`ListMentalModelsData.query.detail`, `GetMentalModelData.query.detail`) and the underlying `sdk.listMentalModels` / `sdk.getMentalModel` functions accept it, but the wrapper's `listMentalModels()` only forwards `tags` + `signal`, and `getMentalModel()` forwards no `query` at all.

Since the HTTP API defaults to `detail=full`, every caller using the wrapper silently receives the full `reflect_response.based_on` provenance chains — which can be 200 KB – 2 MB+ per mental model (and grows monotonically for delta-mode models, per #1192).

## Impact

Any consumer of the official TypeScript client wrapper is affected. Concrete example: [luxus/pi-hindsight#549](https://github.com/luxus/pi-hindsight/issues/549) — pi-hindsight's `hindsight_mental_model list/get` tools return ~180 KB (seen at 2 MB+) directly into the agent context, forcing context compaction on standard-context models and silently inflating per-token cost on 1M-context models.

The official Claude Code MCP plugin avoids this only because it bypasses the TS client and calls the HTTP API directly with `?detail=metadata` / `?detail=content` (#1528, #1543). The TS wrapper gives consumers no way to do the same.

## Root cause

`@vectorize-io/hindsight-client@0.8.5`, `dist/index.js:1711`:

```js
async listMentalModels(bankId, options) {
    const response = await listMentalModels({
      client: this.client,
      path: { bank_id: bankId },
      query: { tags: options?.tags },   // ← detail, tags_match, limit, offset all dropped
      signal: options?.signal
    });
    return this.validateResponse(response, "listMentalModels");
}
```

`dist/index.js:1723` — no `query` field at all:

```js
async getMentalModel(bankId, mentalModelId, options) {
    const response = await getMentalModel({
      client: this.client,
      path: { bank_id: bankId, mental_model_id: mentalModelId },
      signal: options?.signal                          // ← no query; detail has nowhere to go
    });
    return this.validateResponse(response, "getMentalModel");
}
```

The generated types confirm the wrapper is discarding supported parameters:

```ts
// generated types (dist/index.d.ts)
type ListMentalModelsData = {
  query?: {
    tags?: Array<string> | null;
    tags_match?: "any" | "all" | "exact";
    detail?: "metadata" | "content" | "full";   // ← exists, wrapper drops it
    limit?: number;
    offset?: number;
  };
  // ...
};

type GetMentalModelData = {
  query?: {
    detail?: "metadata" | "content" | "full";   // ← exists, wrapper has no query at all
  };
  // ...
};
```

## Suggested fix

Forward the full `query` (or at least `detail` + `tags`) through the wrapper, and accept `detail` in the options type:

```ts
async listMentalModels(bankId, options) {
    const response = await listMentalModels({
      client: this.client,
      path: { bank_id: bankId },
      query: {
        tags: options?.tags,
        tags_match: options?.tagsMatch,
        detail: options?.detail,        // ← forward
        limit: options?.limit,
        offset: options?.offset,
      },
      signal: options?.signal
    });
    return this.validateResponse(response, "listMentalModels");
}

async getMentalModel(bankId, mentalModelId, options) {
    const response = await getMentalModel({
      client: this.client,
      path: { bank_id: bankId, mental_model_id: mentalModelId },
      query: { detail: options?.detail },  // ← add query
      signal: options?.signal
    });
    return this.validateResponse(response, "getMentalModel");
}
```

(Wrapper method signatures would add `detail?: "metadata" | "content" | "full"` etc. to `options`.)

## Why this matters

`detail` was added (#846, closing #825) specifically because `detail=full` "can exceed 200KB+". The maintainers have since defaulted every LLM-facing surface away from `based_on` — the Claude Code MCP plugin uses `detail=metadata`/`content` (#1528, #1543), and the MCP `reflect` tool defaults `include_based_on: false` (#1665). But the TypeScript client wrapper — the integration path most third-party tools use — cannot express `detail` at all, so it silently inherits the unsafe `full` default. Fixing the wrapper closes the gap for every TS consumer at once.

## Note on the client-coverage-check tool

#1362 added a `client-coverage-check` that validates the TS wrapper exposes all OpenAPI request parameters, and it fixed 6 missing **body** fields on `recall`/`reflect`. But `detail` is a GET **query** parameter, not a request body field, so it appears the check only covers body params — the mental-model `detail` gap slipped through. Worth confirming whether the coverage check should also scan query parameters, so this class of regression is caught automatically next time.

## Environment

- `@vectorize-io/hindsight-client`: 0.8.5
- Verified against `main` (as of 2026-07-27): the wrapper still drops `detail` — `listMentalModels` still does `query: { tags: options?.tags }` and `getMentalModel` still has no `query` field at all.

## Related

- [luxus/pi-hindsight#549](https://github.com/luxus/pi-hindsight/issues/549) — downstream impact (context/token explosion)
- #846 / #825 — `detail` parameter added
- #1528 / #1543 — official Claude Code MCP plugin uses `detail=metadata`/`content`
- #1665 — MCP `reflect` defaults `include_based_on: false`
- #1192 — delta refresh accumulates `based_on`

r266-tech:
## Summary

- expose the generated SDK's `detail`, tag-match, limit, and offset controls through `listMentalModels`
- expose `detail` through `getMentalModel`
- preserve the existing request object shape when callers omit the new options
- add mapping and backward-compatibility regression tests

This lets TypeScript callers request metadata- or content-only responses instead of always accepting the server's full provenance payload, while also making server-side filtering and pagination available from the high-level client.

## Testing

- `npm test -- --runInBand tests/mental_model_query_mapping.test.ts tests/create_mental_model_mapping.test.ts` (8 tests)
- `npm run build`
- Prettier on the two changed TypeScript files

Closes #2975

### DIFF
--- a/hindsight-clients/typescript/src/index.ts
+++ b/hindsight-clients/typescript/src/index.ts
@@ -882,12 +882,26 @@ export class HindsightClient {
    */
   async listMentalModels(
     bankId: string,
-    options?: { tags?: string[]; signal?: AbortSignal }
+    options?: {
+      tags?: string[];
+      tagsMatch?: "any" | "all" | "exact";
+      /** Exclude large provenance chains with "metadata" or "content" when they are not needed. */
+      detail?: "metadata" | "content" | "full";
+      limit?: number;
+      offset?: number;
+      signal?: AbortSignal;
+    }
   ): Promise<MentalModelListResponse> {
     const response = await sdk.listMentalModels({
       client: this.client,
       path: { bank_id: bankId },
-      query: { tags: options?.tags },
+      query: {
+        tags: options?.tags,
+        ...(options?.tagsMatch !== undefined ? { tags_match: options.tagsMatch } : {}),
+        ...(options?.detail !== undefined ? { detail: options.detail } : {}),
+        ...(options?.limit !== undefined ? { limit: options.limit } : {}),
+        ...(options?.offset !== undefined ? { offset: options.offset } : {}),
+      },
       signal: options?.signal,
     });
 
@@ -900,11 +914,16 @@ export class HindsightClient {
   async getMentalModel(
     bankId: string,
     mentalModelId: string,
-    options?: { signal?: AbortSignal }
+    options?: {
+      /** Exclude large provenance chains with "metadata" or "content" when they are not needed. */
+      detail?: "metadata" | "content" | "full";
+      signal?: AbortSignal;
+    }
   ): Promise<MentalModelResponse> {
     const response = await sdk.getMentalModel({
       client: this.client,
       path: { bank_id: bankId, mental_model_id: mentalModelId },
+      ...(options?.detail ? { query: { detail: options.detail } } : {}),
       signal: options?.signal,
     });
 
--- a/hindsight-clients/typescript/tests/mental_model_query_mapping.test.ts
+++ b/hindsight-clients/typescript/tests/mental_model_query_mapping.test.ts
@@ -1,0 +1,84 @@
+/**
+ * Unit tests for mental-model query forwarding in the hand-written wrapper.
+ *
+ * These mock the generated SDK so a regression cannot silently restore its
+ * `detail=full` default or drop list pagination and tag-match controls.
+ */
+
+import { HindsightClient } from "../src";
+import * as sdk from "../generated/sdk.gen";
+
+jest.mock("../generated/sdk.gen");
+
+const mockedList = sdk.listMentalModels as jest.MockedFunction<typeof sdk.listMentalModels>;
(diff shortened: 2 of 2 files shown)


nicoloboschi (approved):
LGTM — faithfully forwards the query params the generated SDK already accepts, backward-compat shapes preserved and locked by tests. Follow-up incoming for the identical gap in the Python wrapper.

PR #3066, opened 2026-07-30
fix(openclaw): default recall injection to user context

### LINKED ISSUE #3061: recallInjectionPosition default (prepend) confirmed to break prompt caching — LoCoMo benchmark data
# `recallInjectionPosition` default (`prepend`) confirmed to break prompt caching — LoCoMo benchmark data

## Background

[#703](https://github.com/vectorize-io/hindsight/issues/703) added a `recallInjectionPosition` config option to the OpenClaw plugin, controlling where recalled memories are injected relative to the system prompt:

- `"prepend"` (default) → `prependSystemContext`: inserted at the **start** of the system prompt
- `"append"` → `appendSystemContext`: inserted at the **end** of the system prompt
- `"user"` → `prependContext`: inserted before the user message, entirely **outside** the system prompt

Since Anthropic/Google prompt caching is **prefix-based**, changing the start of the system prompt on every turn invalidates the cache from the first token. `"prepend"` (the default) is the option least friendly to caching among the three — this was #703's theoretical argument, but it shipped without changing the default, so new installs still default to `prepend`.

We ran a small benchmark to check whether this theoretical concern holds up in practice, and to what degree.

## Setup

- Dataset: LoCoMo, sample `conv-26`
- Questions: 152 (excluding category=5)
- Variable: only `recallInjectionPosition` changed; everything else held constant (`recallBudget: high`, `recallMaxTokens: 4096`)

| Param | test1 | test2 |
|---|---|---|
| recallInjectionPosition | default (`prependSystemContext`) | `user` (`prependContext`) |

## Results

| Metric | test1 (default `prepend`) | test2 (`user`) | Delta |
|---|---|---|---|
| Accuracy | 75.66% (115/152) | 75.00% (114/152) | roughly flat (-0.66pt) |
| Total elapsed | 914.4s | 866.4s | test2 slightly faster |
| Avg elapsed/question | 6.016s | 5.700s | similar |
| Total no-cache input tokens | 4,549,494 | 1,232,042 | test2 down 73% |
| Total cacheRead tokens | 124,288 | 3,740,928 | **test2 up 30x** |
| Avg no-cache tokens/question | 29,930.88 | 8,105.54 | test2 down 73% |
| Avg cacheRead tokens/question | 817.68 | 24,611.37 | **test2 up 30x** |
| Total input tokens (no-cache + cacheRead) | 4,673,782 | 4,972,970 | roughly the same |

## Interpretation

Accuracy is essentially unchanged between the two settings (75.66% vs 75.00%, a 1-question difference out of 152), confirming that injection position doesn't affect the usefulness of the recalled memories to the model.

Cache behavior, however, differs sharply and in the direction #703 predicted:

- The default `prepend` (test1) achieves only 1/30th the cacheRead tokens of `user` (test2), and needs 3.7x more no-cache tokens.
- In other words, this benchmark empirically confirms #703's theoretical claim: the default option (`prepend`) does significantly break prompt caching, while `user` restores most of the cache hit rate — with no measurable accuracy cost.

Likely mechanism: with the default `prepend`, the per-turn-changing memory content is inserted at the very start of the system prompt, so the system prompt's prefix differs every turn, invalidating the provider-side prefix cache. With `user`, the memory content goes into the user message instead, leaving the system prompt itself unchanged across turns, so its cache is reused reliably.

## Conclusion / discussion points

- Sample size is modest (152 questions, single conversation), but the signal is clean: switching `recallInjectionPosition` from default `prepend` to `user` leaves accuracy essentially unchanged (-0.66pt) while improving cacheRead tokens 30x and cutting no-cache tokens 73%.
- Opening this issue to discuss, building on #703:
  1. Should the default change from `prepend` to a more cache-friendly option (e.g. `user`), rather than leaving this as an opt-in config flag? New installs currently pay the full cache-miss cost unless they know to change it.
  2. If the default stays as-is, should the docs / setup flow surface this tradeoff more prominently, since the cost impact can be substantial for agents with large static system prompts?

## Raw data

```yaml
## test1
"recallBudget": "high"
"recallMaxTokens": 4096
"recallInjectionPosition": default (prependSystemContext)

=== Judge Result Statistics (excluding category=5) ===
Total rows: 152
Graded rows: 152
Correct: 115
Wrong: 37
Accuracy: 75.66%

Elapsed time (QA):
  Total elapsed seconds: 914.438
  Avg elapsed seconds: 6.016
  Min elapsed seconds: 3.027
  Max elapsed seconds: 17.189

Token usage (QA):
  Total no-cache tokens (input_tokens): 4,549,494
  Total cacheRead tokens: 124,288
  Total output tokens: 39,008
  Total input tokens (no-cache + cacheRead): 4,673,782
  Avg no-cache tokens: 29,930.88
  Avg cacheRead tokens: 817.68
  Avg output tokens: 256.63
  Avg total input tokens: 30,748.57

## test2
"recallBudget": "high"
"recallMaxTokens": 4096
"recallInjectionPosition": "user"

=== Judge Result Statistics (excluding category=5) ===
Total rows: 152
Graded rows: 152
Correct: 114
Wrong: 38
Accuracy: 75.00%

Elapsed time (QA):
  Total elapsed seconds: 866.403
  Avg elapsed seconds: 5.700
  Min elapsed seconds: 2.729
  Max elapsed seconds: 23.586

Token usage (QA):
  Total no-cache tokens (input_tokens): 1,232,042
  Total cacheRead tokens: 3,740,928
  Total output tokens: 46,269
  Total input tokens (no-cache + cacheRead): 4,972,970
  Avg no-cache tokens: 8,105.54
  Avg cacheRead tokens: 24,611.37
  Avg output tokens: 304.40
  Avg total input tokens: 32,716.91
```

Sanderhoff-alt:
## Summary

- Change the default recall injection position from `prepend` to `user` for configurations that omit or provide an invalid `recallInjectionPosition` value.
- Preserve explicit `prepend`, `append`, and `user` settings.
- Align the plugin manifest, runtime normalization, tests, integration documentation, README, and recall injection guide with the new default.

## Production impact

Production OpenClaw agents commonly have large, mostly static system prompts assembled from workspace instructions, tool definitions, and files such as AGENTS.md, SOUL.md, and MEMORY.md. With the previous `prepend` default, dynamic recalled memories were inserted at the beginning of the system prompt on every turn. Prefix-based caches used by providers such as Anthropic and Google therefore saw a different prefix each turn and could not reuse the expensive static prompt that followed it.

The LoCoMo benchmark reported in #3061 measured this effect across 152 questions. Moving recall from the default system-prefix position to user context kept accuracy effectively flat at 75.66% versus 75.00%, increased cache-read tokens by about 30x, and reduced non-cached input tokens by 73%. For production agents with large static prompts, this can materially reduce per-turn inference cost and improve latency without reducing observed recall quality.

Defaulting to `user` also keeps recalled memory below system instructions, which provides a clearer trust boundary for memory content while preserving the complete system-prompt prefix for caching. Users that intentionally require system-level memory framing can continue to select `prepend` or `append` explicitly.

This is a behavioral change for both new installations and existing installations that do not explicitly set `recallInjectionPosition`, because the default is resolved at runtime rather than written into the OpenClaw configuration file.

## Test plan

- `npm test` in `hindsight-integrations/openclaw` (281 tests passed)
- `npm run build` in `hindsight-integrations/openclaw`
- `./scripts/hooks/lint.sh`

Closes #3061


nicoloboschi (approved):
LGTM. Flips recallInjectionPosition default prepend→user (closes #3061). Clean: config default + injection fallback both updated, types/manifest/README/docs/guide in sync, unit + integration tests updated, verify-generated-files green.

PR #3067, opened 2026-07-30
fix(packaging): bundle licenses in Python distributions

### LINKED ISSUE #3054: Add license to python package
### Use Case

I'm pulling this package in from pypi

### Problem Statement

Tools like Sonatype repository firewall, which many institutions use, block packages that have a non compliant or missing license. 

You all have an MIT license on GH but because it is not bundled into the package then it fails and goes into quarantine. 

The package would seamlessly come in and not require a ticket on our side to manually un-quaratine it.

### Proposed Solution

Add the license into the package

### Alternatives Considered

_No response_

### Priority

### Additional Context

_No response_

### Checklist

Sanderhoff-alt:
## Summary

- declare the MIT SPDX license expression for every Python distribution published by the release workflow
- stage the repository canonical LICENSE file in each isolated package build context without maintaining duplicate source copies
- verify wheel and sdist license contents and core metadata before publishing

## Production impact

Enterprise artifact repositories and supply-chain controls such as Sonatype Repository Firewall inspect both package metadata and the files shipped in the distribution. The current PyPI artifacts do not bundle the repository MIT license, so compliant environments can quarantine Hindsight packages even though the source repository is licensed correctly. That blocks automated dependency promotion and production deployments, and forces operators to open exception tickets and manually release every affected version.

This change makes the license available in both wheels and source distributions and emits the standardized License-Expression and License-File metadata consumed by compliance tooling. The release-time assertions prevent a packaging backend or workflow change from silently publishing non-compliant artifacts again. As a result, Hindsight can move through governed package repositories without one-off approvals while preserving the root LICENSE as the single source of truth.

## Verification

- built all six Python distributions from temporary isolated build contexts
- confirmed every wheel contains its dist-info/licenses/LICENSE file
- confirmed every wheel declares License-Expression: MIT and License-File: LICENSE
- confirmed every source distribution contains LICENSE
- ran ./scripts/hooks/lint.sh

Closes #3054

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,6 +25,20 @@ jobs:
       with:
         python-version-file: ".python-version"
 
+    # Each package is built from its own directory, so stage the repository's
+    # canonical license inside each isolated build context.
+    - name: Stage Python package licenses
+      run: |
+        for package in \
+          hindsight-clients/python \
+          hindsight-api-slim \
+          hindsight-api \
+          hindsight-all \
+          hindsight-all-slim \
+          hindsight-embed; do
+          cp LICENSE "$package/LICENSE"
+        done
+
     # Build all packages
     - name: Build hindsight-client
       working-directory: ./hindsight-clients/python
@@ -49,6 +63,24 @@ jobs:
     - name: Build hindsight-embed
       working-directory: ./hindsight-embed
       run: uv build --out-dir dist
+
+    - name: Verify Python package licenses
+      run: |
+        for package in \
+          hindsight-clients/python \
+          hindsight-api-slim \
+          hindsight-api \
+          hindsight-all \
+          hindsight-all-slim \
+          hindsight-embed; do
+          wheel=$(find "$package/dist" -maxdepth 1 -name '*.whl' -print -quit)
+          sdist=$(find "$package/dist" -maxdepth 1 -name '*.tar.gz' -print -quit)
+
+          unzip -Z1 "$wheel" | grep -Eq '\.dist-info/licenses/LICENSE$'
+          unzip -p "$wheel" '*/METADATA' | grep -Fxq 'License-Expression: MIT'
+          unzip -p "$wheel" '*/METADATA' | grep -Fxq 'License-File: LICENSE'
+          tar -tzf "$sdist" | grep -Eq '/LICENSE$'
+        done
 
     # Publish in order (client and api-slim first, then api/all wrappers which depend on them)
     - name: Publish hindsight-client to PyPI
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,13 @@ build/
 dist/
 wheels/
 *.egg-info
+# Release builds stage the canonical root license in each package context.
+/hindsight-clients/python/LICENSE
+/hindsight-api-slim/LICENSE
+/hindsight-api/LICENSE
+/hindsight-all/LICENSE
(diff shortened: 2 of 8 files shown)


nicoloboschi (commented):
Well-scoped — does exactly what #3054 asks: SPDX `license = "MIT"` on 6 pyprojects, stages the canonical root `LICENSE` into each isolated build context (no duplicate source copies 👍), and gates the release on wheels/sdists actually containing the license + PEP 639 metadata. One thing worth fixing before merge.

### 🔴 Should fix — `setuptools>=61` floor is too low for the metadata this PR emits

`hindsight-api/pyproject.toml` and `hindsight-all-slim/pyproject.toml` both keep `requires = ["setuptools>=61"]`, but this change relies on **PEP 639**, which setuptools only implements in **≥ 77.0.0**:

- The bare-string `license = "MIT"` (SPDX expression) isn't valid under the older PEP 621 schema that 61–76 enforce (`license` had to be a `{text=...}`/`{file=...}` table). On those versions the build errors / doesn't emit the field.
- `License-Expression: MIT` / `License-File: LICENSE` (Metadata 2.4) — exactly what the `Verify` step greps for — is only produced by setuptools ≥ 77.

It works today only because `uv build`'s isolated env resolves the newest setuptools. But the declared floor now contradicts the actual requirement: any pinned/constrained/cached env that resolves 61–76 will break the build or fail the verify step. Please bump both to `requires = ["setuptools>=77"]`.

(Minor, same reasoning: the hatchling packages leave `requires = ["hatchling"]` unpinned; `License-Expression` needs hatchling ≥ 1.27. Latest resolves fine, but a `hatchling>=1.27` floor would make it reproducible.)

### 🟡 Nit — staged `LICENSE` files aren't ignored

`cp LICENSE "$package/LICENSE"` drops untracked `LICENSE` files into each package dir. Harmless in CI, but a local `uv build` now leaves stray copies a dev could accidentally commit (defeating the single-source-of-truth goal). Consider adding `<package>/LICENSE` to `.gitignore`.

### ✅ Verified correct

- **No `License ::` classifier conflict** — none of the 6 pyprojects carry one. (With setuptools ≥ 77 a legacy classifier alongside an SPDX expression is a hard error, so this matters.)
- **Verify step actually gates** — GitHub's default `bash -eo pipefail` means each `unzip … | grep -q` failure aborts the step; the `grep -Fxq` whole-line matches are exact.
- **License-file inclusion works for both backends** — the PEP 639 default `license-files` glob picks up the staged `LICENSE` next to each `pyproject.toml`, landing it at `*.dist-info/licenses/LICENSE` (what the verify checks).
- Root `LICENSE` exists; build/publish `working-directory` and `$package/dist` paths line up for all six.

**Verdict:** LGTM once the `setuptools>=77` floor is bumped on the two setuptools packages — that's the gap between "works because CI resolves latest" and "correct by declaration."

Sanderhoff-alt:
Thanks for the review. I have addressed all three points in the latest commit:

- Bumped the two setuptools build requirements to setuptools>=77.
- Added hatchling>=1.27 floors to the four Hatchling packages for reproducible PEP 639 metadata.
- Added precise .gitignore entries for the six staged package LICENSE files.

I rebuilt all six wheels and sdists and verified Metadata-Version: 2.4, License-Expression: MIT, License-File: LICENSE, and the bundled license paths. The repository lint and OpenAPI compatibility checks also pass.

nicoloboschi (approved):

PR #3068, opened 2026-07-30
docs: add "Hindsight 源码精读" — a self-contained, course-style source-code walkthrough site in Chinese

Birchove:
## What this adds

A new, self-contained static site under `hindsight-course/` that walks readers through Hindsight's internals as a progressive, engineering-oriented course (written in Chinese; code/API names kept in English).

- 14 pages: intro + 12 chapters + appendix
  - System overview & data model (schema, ER, pgvector, multi-tenancy)
  - Retain pipeline (end-to-end, extraction/chunking, entity resolution & links, Phase1/2/3)
  - Recall pipeline (query analysis, the 4 retrieval arms, RRF fusion + cross-encoder reranking)
  - Consolidation (facts → observations) & Reflect (tool-calling agent, disposition)
  - Cross-cutting infra (LLM abstraction, embeddings, provider model, hierarchical config)
  - Ecosystem overview (SDKs/CLI, integrations, control plane, deployment, benchmarks)
  - Appendix: code map, jump-to-function index, glossary, run/debug guide
- 20 Mermaid diagrams + syntax-highlighted source excerpts with file/function references

## Notes

- Fully offline: Mermaid and highlight.js are vendored; open `index.html` directly
  or serve the folder with any static server. No build step.
- Purely additive documentation — no changes to any code, build, or runtime behavior.
- Verified: all pages and diagrams render correctly (headless-browser checked).

> Note to maintainers: this is an independent, third-party educational walkthrough
> (authored against a fork). Happy to relocate, restructure, or close if it's not a
> fit for the main repo.

One thing worth a quick thought before you submit: this is your own educational content rather than a project change, so upstream may prefer it live in your fork (or a discussion/docs issue) rather than merged into vectorize-io/hindsight. If it's a PR to your own fork/branch, you can drop the last "Note to maintainers" line.

image


nicoloboschi:
hey @Birchove appreciate the effort here.

I understand chinese content here could be very valuable, why would we need new content tho? I think maybe just having the translation of the official docs is more maintainable 

what do you think?

Birchove:
Thanks — that's a fair question, and I think the answer is that this isn't meant to be a translation of the docs at all. It's a different kind of artifact: a source-reading **course**.

The official docs already cover the *what* and the *how-to-use* really well (API surface, concepts, quickstarts). What they don't contain — and what this contributes — is the *why-the-implementation, chapter-by-chapter walkthrough of the engine internals (Retain / Recall / Reflect / Consolidation, the data model, the LLM & embedding abstraction), with the design rationale, the concurrency/correctness tradee cross-references (e.g. why streaming mini-batches, why the Phase1/2/3 transaction split, why RRF-then-rerank). None of that prose exists in the current docs, so translating them wouldn't produce this — there's nothing there to translate. The target reader is someone who knows agent terminology but has never *implemented* a memory system, and the goal is to take them from "what problem does this solve" to "here's the exact function and the reason it's written that way," the way a good internals lecture would.

On your maintainability point: I agree that ustify itself, and I've built the structure to be language-agnostic so it's cheap to keep in sync rather than a one-off. It now ships — with a one-click language switcher, and the layout is fully static (no build step, Mermaid + syntax highlighting vendored, works offline), so adding another language later is just dropping in one more folder. So the "Chinese-only" concern and the "keep it translatable" concern are both addressed at the architecture level.

Concretely: `hindsight-course/` = intro + 12f-contained static site, two languages.

Happy to land this however you prefer — merg-dive course" section, linked as a community resource, or kept in a fork. If your only objection was "don't add net-new Chinese-only content," the English version + the i18n structure should resolve it; if the objparallel narrative alongside the docs," I completely understand and will happily keep it external. Either way, glad the effort can be useful. 🙏

Thanks again for the work you put into Hindsight 🙌

PR #3074, opened 2026-07-30
fix(reflect): reject serialized parameter tails in done answers

### LINKED ISSUE #3048: reflect: leaked done-tool tail still persists when the model DOES emit a valid tool call (#3013 guard does not fire)
### Summary

`#3013` removed the salvage logic and made reflect fail loudly when a transport can't produce a usable tool call. That correctly covers the Vertex AI gpt-oss case in the PR description.

There is a second, structurally different path that produces the same persisted corruption and which `#3013` does **not** catch: the model emits a **well-formed tool call**, and the leaked fragment is inside the `answer` argument value.

Since `#3013` also removed the last sanitisation step, that fragment is now written to mental model content verbatim.

### Why the new guard doesn't fire

On current `main`:

- `ReflectToolCallError` is raised only under `if not saw_tool_call` (`reflect/agent.py`, ~L919). In this failure mode `saw_tool_call` is `True` — the provider returns a valid `tool_calls` entry for `done`.
- The answer is then taken as-is: `answer = args.get("answer", "").strip()` (~L1268).
- `grep -c '_clean_done_answer|</answer>'` on `main` returns `0`, so nothing downstream inspects the value.

Net: a valid tool call whose `answer` string ends in tool-call syntax passes every check.

### Observed shape

The `answer` argument value ends with a fragment resembling XML tool-call syntax (UUIDs redacted):

```
...ordinary prose ending a sentence.</answer>
<parameter name="memory_ids">["<uuid>", "<uuid>", ...]
```

Note this is *inside* the JSON string value of `answer`, alongside a correctly populated `memory_ids` array in the same tool call. The model appears to blend two tool-call serialisations mid-generation rather than failing to tool-call.

The result is persisted into mental model `content`, and re-parsed into `structured_content`, so a single occurrence corrupts both surfaces.

### Frequency and provider

Observed via LLM observability traces on a self-hosted deployment, provider `anthropic/claude-haiku-4-5` reached through a LiteLLM proxy (not Vertex/gpt-oss):

- 2026-07-22: 2 occurrences out of ~172 generations, both persisted, each corrupting one mental model
- 2026-07-27: 3 occurrences with this exact shape out of ~830 generations (plus 2 rows with a different, partial JSON shape)

Low rate, but non-zero and recurring across weeks. Two mental models were corrupted with 133 and 49 leaked IDs respectively before it was noticed — the failure is silent, since the refresh reports success.

### Why this is worth separating from #3013

The two failure modes need different remedies:

| | #3013's case | this case |
|---|---|---|
| tool call produced | no | yes, well-formed |
| detectable at transport | yes | no |
| current behaviour | raises | persists silently |

A `saw_tool_call` check cannot distinguish them, so closing this needs either validation of the `answer` value itself, or removal of the boundary the model is crossing.

### One possible direction

The leak occurs at the `answer` → `memory_ids` parameter boundary. The server already accumulates every retrieved ID during the loop and validates the model's returned IDs against that set, discarding anything unrecognised — so the model's contribution is a subset selection over data the backend already holds.

If those ID parameters were dropped from the `done` tool, this specific boundary would not exist. That is a design tradeoff rather than an obvious win: it costs model-attributed provenance, and it does not remove other sibling boundaries (a `done` schema built with directives still carries a required `directive_compliance` after `answer`). Raising it as an option, not a recommendation.

Happy to supply redacted trace payloads or test cases if useful.

### Related

- #2959 / #2960 — the empty-answer overwrite. Complementary rather than superseded: #3013 guards the transport boundary, while #2960 preserves existing content when reflect returns nothing usable for any reason, including non-transport causes such as an iteration limit.

Alan5168:
## Summary

Reject a parsed `done` tool call when its `answer` contains a serialized sibling parameter tail such as `</answer><parameter name="memory_ids">...`.

Some provider responses can return a valid tool call while blending XML tool serialization into the JSON `answer` value. Because a tool call exists, the no-tool-call guard added in #3013 does not run, and the corrupted tail gets persisted verbatim into `mental_models.content`.

This validates the exact `answer`/sibling-parameter boundary against the declared `done()` fields and raises `ReflectToolCallError` (matching the existing fail-loud contract). It does not truncate or silently clean the answer, and it does not touch #3013's no-tool-call guard.

## Tests

- Added `test_done_tool_rejects_serialized_parameter_tail`, a regression test that is red on current `main` and green with the fix (verified locally by reverting just the production fix and re-running).
- `40 passed` across the related pure unit tests in `test_reflect_agent.py` (excluding RealLLM/Integration/DirectiveLeakageOnEmptyBank, which need external providers/deps not available here).
- `ruff check` and `ruff format --check` pass for all changed Python files.

Closes #3048

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -4440,9 +4440,8 @@ def _register_routes(app: FastAPI):
         except LLMNotAvailableError as e:
             raise HTTPException(status_code=400, detail=str(e))
         except ReflectToolCallError as e:
-            # The configured model/transport can't drive reflect's tool-calling loop.
-            # The request itself is fine, so this is a server-side (500) failure, not a
-            # 4xx -- but log at warning, not error: it's a misconfiguration, not a bug.
+            # The provider did not return a structured result that is safe to use.
+            # The request itself is fine, so this is a server-side (500) failure.
             logger.warning("Reflect tool-calling failure in bank %s: %s", bank_id, e)
             raise HTTPException(status_code=500, detail=str(e))
         except TimeoutError as e:
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -56,7 +56,7 @@ NO_ANSWER_TEXT = "No answer provided."
 
 
 class ReflectToolCallError(RuntimeError):
-    """The model never produced a tool call reflect could understand.
+    """The model did not produce a tool-call result reflect can safely use.
 
     Reflect is driven entirely by structured tool calls (``recall``, ``expand``,
     ``done`` ...). Some provider transports do not actually support function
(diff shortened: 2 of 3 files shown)


nicoloboschi:
seems like a model issue, let's not open the pandora's box

PR #3084, opened 2026-07-30
fix(retain): unwrap single-key envelope dicts in fact extraction

### LINKED ISSUE #3028: fix: fact extraction returns 0 facts with litellm + Vertex AI Claude

yossiovadia:
## Summary

- Fixes silent 0-fact extraction when using `litellm` with `vertex_ai/claude-*` models
- Key-agnostic approach: unwraps any single-key dict whose sole value contains `"facts"` — handles `{"parameter": ...}` and any future wrapper variants without hardcoding keys
- 8 unit tests for `_coerce_fact_response()` covering envelope unwrap, passthrough, multi-key preservation, and edge cases

Supersedes #3029 (closed), rebased cleanly on current main.

## Problem

Vertex AI Claude wraps structured output in `{"parameter": {"facts": [...]}}`. Since `_coerce_fact_response()` returns the dict as-is, `extraction_response_json.get("facts", [])` returns `[]` — silently dropping all extracted facts. Every `retain_extract_facts` call produces 0 facts despite the LLM generating valid output (confirmed with 17/17 requests, full payloads in #3028).

## Approach

Per @nicoloboschi's feedback on #3029, this avoids hardcoding `"parameter"` and instead uses a structural heuristic:

```python
if "facts" not in response and len(response) == 1:
    inner = next(iter(response.values()))
    if isinstance(inner, dict) and "facts" in inner:
        return inner
```

**Constraints:**
- Only fires when the outer dict has **exactly one key** (not a normal multi-field response)
- Only fires when the inner value is a **dict containing "facts"**
- Direct `{"facts": [...]}` responses pass through unchanged

## Test plan

- [x] 8 unit tests pass: envelope unwrap (parameter, content, data, json_input, result), passthrough, multi-key preservation, single-key-without-facts, empty facts, bare list coercion, invalid types
- [x] Manual verification: retain with `litellm` + `vertex_ai/claude-opus-4-6` extracts facts correctly (was 0, now 3+ per retain)
- [x] Existing tests unaffected

Fixes #3028

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -1295,8 +1295,18 @@ def _build_request_body(llm_config, config, prompt: str, user_message: str, resp
 
 
 def _coerce_fact_response(response: Any) -> dict[str, Any] | None:
-    """Accept the schema wrapper, or a recoverable top-level facts array."""
+    """Accept the schema wrapper, or a recoverable top-level facts array.
+
+    Some providers (e.g. litellm + Vertex AI Claude) wrap structured output
+    in an extra envelope like ``{"parameter": {"facts": [...]}}``.  Rather than
+    hard-coding known wrapper keys, unwrap any single-key dict whose value
+    contains the expected ``facts`` list.
+    """
     if isinstance(response, dict):
+        if "facts" not in response and len(response) == 1:
+            inner = next(iter(response.values()))
+            if isinstance(inner, dict) and "facts" in inner:
+                return inner
         return response
     if isinstance(response, list) and all(isinstance(item, dict) for item in response):
         return {"facts": response}
--- a/hindsight-api-slim/tests/test_coerce_fact_response_parameter_wrapper.py
+++ b/hindsight-api-slim/tests/test_coerce_fact_response_parameter_wrapper.py
@@ -1,0 +1,78 @@
+"""
+Test that _coerce_fact_response unwraps single-key envelope dicts whose
+inner value contains the expected ``facts`` list.
+
+Vertex AI Claude (via litellm) returns ``{"parameter": {"facts": [...]}}``,
+but other providers/versions may use different wrapper keys (``"content"``,
+``"data"``, ``"json_input"``).  The fix is key-agnostic: any single-key dict
+whose sole value is a dict containing ``"facts"`` is unwrapped.
+"""
+
+from hindsight_api.engine.retain.fact_extraction import _coerce_fact_response
+
+
+def test_unwraps_parameter_envelope():
+    """Vertex AI Claude wraps tool output in {"parameter": {"facts": [...]}}."""
+    response = {
+        "parameter": {
+            "facts": [
+                {"what": "version is 1.0", "who": "user", "fact_type": "world"},
+            ]
+        }
+    }
+    result = _coerce_fact_response(response)
+    assert "facts" in result
+    assert len(result["facts"]) == 1
+    assert result["facts"][0]["what"] == "version is 1.0"
+
+
+def test_unwraps_any_single_key_envelope():
+    """Any single-key wrapper around {"facts": [...]} is unwrapped."""
+    for key in ("content", "data", "json_input", "result"):
+        response = {key: {"facts": [{"what": "a"}]}}
+        result = _coerce_fact_response(response)
+        assert "facts" in result, f"failed for wrapper key: {key}"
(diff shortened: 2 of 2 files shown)


nicoloboschi:
if the model doensn't follow the json schema, then it will fail in other places (reflect, mental models, consolidation) so I'd rather not support this edge case and recommend to investigate the root cause on your end

yossiovadia:
@nicoloboschi — found the root cause. It's not the model failing to follow the schema.

**litellm converts \`response_format: json_schema\` into a tool call for Vertex AI Claude:**

\`\`\`
Final returned optional params: {
  'tool_choice': {'name': 'json_tool_call', 'type': 'tool'},
  'tools': [{'name': 'json_tool_call', 'input_schema': {<FactExtractionResponse schema>}}],
  'json_mode': True
}
\`\`\`

Claude's tool-use response wraps the output as \`{"parameter": {"facts": [...]}}\` — this is how Anthropic's tool_use content blocks serialize (\`type: "tool_use", name: "json_tool_call", input: {...}\`). litellm should unwrap this back to plain content but doesn't.

**Key evidence:**
- Simple schemas (no \`$defs\`) → litellm passes as \`json_object\` → no wrapping ✅
- Complex schemas with \`$defs\` (like \`FactExtractionResponse\`) → litellm converts to tool call → wrapping ❌
- The model IS following the schema perfectly — the \`{"facts": [...]}\` inside \`"parameter"\` matches the Pydantic model exactly
- Consolidation doesn't wrap because its schema doesn't trigger litellm's tool-call conversion path

This is a litellm bug (converting json_schema to tool_use without unwrapping the response), but since litellm is a supported provider, hindsight's parser seeing this is expected. The single-key unwrap in this PR is a defensive, minimal fix for a real integration scenario.

Happy to file this upstream on litellm too if that helps.

yossiovadia:
Filed the root cause fix upstream on litellm: https://github.com/BerriAI/litellm/pull/35792 (issue: https://github.com/BerriAI/litellm/issues/35790).

The bug is in litellm's `_convert_tool_response_to_message()` — it unwraps `{"values": ...}` for direct Anthropic but not the `{"parameter": ...}` envelope that Vertex AI Anthropic returns. Fix adds generic single-key dict unwrapping, same pattern as Bedrock's `_unwrap_bedrock_properties`.

PR #3086, opened 2026-07-30
feat(claude-code): coding profile, automatic file tags, and tool_use input capping

ohora23:
## Motivation

The Claude Code plugin is a faithful port of the Openclaw chat plugin, so its defaults are tuned for conversational agents: tool calls are not retained, every session shares one bank, and recall pacing assumes chat-frequency turns. For Claude Code's primary use — coding sessions — those defaults remember the small talk and forget the actual work: the substance of a coding session lives in Edit/Write/Bash tool calls, sessions span many repositories, and prompts fire far more frequently than in chat.

This PR makes the plugin coding-capable while leaving existing behavior untouched: without a profile configured, nothing changes.

## Changes

**Fix: cap `tool_use` input size in retained transcripts** (`_compact_tool_input`)
`tool_result` blocks were truncated at 2000 chars, but `tool_use` inputs went through verbatim — a `Write` of a large file embedded the entire file body in the retain payload, bloating requests and degrading fact extraction. Two-level cap: 300 chars per string field, then a priority-keys-only fallback (`file_path`, `command`, ...) when the serialized input still exceeds 1500 chars.

**Feature: automatic `file:<relpath>` tags** (`extract_touched_files`)
Each retained document is tagged with the files modified by `Write`/`Edit`/`MultiEdit`/`NotebookEdit` tool calls (relativized against cwd, capped at 20, read-only tools excluded) and gets `metadata.files_modified`. Recall can then answer "what happened to this file" via `recallTags: ["file:src/auth.py"]`. No-op for conversational sessions.

**Feature: `profile` setting with a `coding` preset**
`{"profile": "coding"}` in user config (or `HINDSIGHT_PROFILE=coding`) applies: `retainToolCalls: true`, `dynamicBankId: true` with `["agent", "project"]` granularity, `recallBudget: "low"`, and engineering-focused bank/retain missions. Precedence is deliberate: a preset overrides built-in defaults and the vendor-shipped `settings.json`, but never a key the user set explicitly in `~/.hindsight/claude-code.json` or via env — so existing configurations are unaffected and the preset stays fully overridable.

## Backward compatibility

- No profile configured → identical behavior to before (verified by the existing test suite passing unchanged).
- File tags only appear when file-mutating tool calls exist in the window; chat sessions see no new tags.
- The input cap changes retained payloads for sessions that already had `retainToolCalls: true`, but only by truncating fields that were previously retained verbatim at unbounded size.

## Testing

- Test suite: 197 → 219 (22 new unit/e2e tests across `test_content.py`, `test_hooks.py`, `test_config.py`), all passing with stdlib-only hooks as before.
- Live end-to-end: wired the hooks into a real Claude Code project against `hindsight-embed` on :9077 with Ollama (`qwen3-14b`) as the extraction LLM. A coding transcript (bug-fix session with Edit/Write/Bash calls) produced: per-project bank `claude-code::<project>`, both modified files tagged, 4 accurate extracted observations including facts that existed only in tool_use inputs (which stock defaults would have dropped), correct recall injection, and correct tag-filtered recall (`file:` tag returns the document's memories; a nonexistent file tag with `any_strict` returns none).

### DIFF
--- a/hindsight-integrations/claude-code/CHANGELOG.md
+++ b/hindsight-integrations/claude-code/CHANGELOG.md
@@ -4,6 +4,19 @@
 
 ### Added
 
+- `profile` setting (env: `HINDSIGHT_PROFILE`) with a `coding` preset that
+  tunes grouped defaults for coding sessions: `retainToolCalls: true`,
+  per-project dynamic banks (`dynamicBankId` + `["agent", "project"]`),
+  `recallBudget: "low"`, and engineering-focused bank/retain missions.
+  Presets override built-in and vendor-shipped defaults but never a key set
+  explicitly in user config or env, so existing configurations are
+  unaffected; without a profile, behavior is unchanged.
+- Retained documents are automatically tagged with `file:<relpath>` for each
+  file modified by `Write`/`Edit`/`MultiEdit`/`NotebookEdit` tool calls
+  (relativized against the working directory, capped at 20), and the list is
+  recorded in `metadata.files_modified`. Recall can then filter with
+  `recallTags: ["file:src/auth.py"]`.
+
 - `{user_id}` template variable for `retainTags` and `retainMetadata`, resolved
   from the `HINDSIGHT_USER_ID` env var (empty string if unset). Enables
   machine-independent per-user memory scoping without hardcoding user ids in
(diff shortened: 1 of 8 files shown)


ohora23:
Withdrawing this PR — I'll maintain these changes in my fork for my own workflow instead. Thanks!

PR #3088, opened 2026-07-30
feat(files): preserve raw source with recoverable lineage

offendingcommit:
## Summary

File uploads can now preserve exact UTF-8 source text and expose enough operation lineage for a clean-clone operator to recover an asynchronous import safely.

The new `raw_utf8` parser rejects invalid UTF-8 and NUL bytes instead of normalizing source. File conversion and downstream retain operations expose document identity, parent-child lineage, parser contract, and the SHA-256 of the submitted bytes. `/version` advertises both parser versions and the source-hash lineage capability so clients can fail before writing against an older deployment.

## Validation

- raw UTF-8 parser tests: 4 passed
- generated Python lineage/version tests: 4 passed
- TypeScript build: passed
- Go build: passed
- repository lint hook: passed
- OpenAPI generation and documentation build: passed

The broader DB-backed file-retain suite was not used as final evidence because the shared local PostgreSQL database is ahead of this checkout and fails on an unrelated generated-column migration mismatch.

## Post-deploy validation

- Verify `/version` reports `raw_utf8: "1"` and `file_operation_lineage: "source_sha256-v1"`.
- Submit one non-production raw UTF-8 file and confirm its conversion and retain rows share document identity, source hash, and parent lineage.
- Monitor failed `file_convert_retain` and `retain` operations after rollout.
- Roll back to the prior image if raw parsing changes source text or lineage fields are absent.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


offendingcommit:
Superseded by offendingcommit/hermes-plugin-valdris#3. The released hindsight-client 0.8.6 already supports async text retain with caller-provided operation IDs, so Valdris now submits exact Markdown directly with deterministic IDs and no longer needs the raw parser or file-lineage changes in this PR. Thanks for the review context.

Issue #2431, opened 2026-06-27
Structured reflect with response_schema times out while plain reflect succeeds (empty message content, finish_reason=length)


WuKavin:
### Summary

On Hindsight `0.8.3`, plain `reflect` succeeds, but `reflect` with `response_schema` reproducibly times out for the same bank/query shape.

In my environment, the final reflect LLM repeatedly returns **empty message content** with `finish_reason=length`, and the request eventually times out.

### Why this seems worth reporting

This does **not** look like a generic startup/network problem:

- Hindsight is healthy (`/health` returns healthy)
- `retain` succeeds
- `consolidation` succeeds
- plain `reflect` succeeds
- only **structured reflect** (`response_schema`) fails

So the failure surface seems specifically tied to `reflect + response_schema`.

### Environment

- Hindsight: `0.8.3`
- Deployment: Docker on Debian-based NAS
- Current relevant config:
  - default LLM provider/model: `opencode-go / deepseek-v4-flash`
  - reflect LLM provider/model: `openai / big-pickle`
  - embeddings: `local`
  - reranker: `rrf`

For completeness: I had to make the container use an HTTP proxy for Hugging Face downloads to get local embeddings initialized, but after that Hindsight became healthy and the issue below still reproduced.

### Reproduction

1. Create a temporary bank
2. `retain` two short items into the bank
3. Call plain `reflect` -> succeeds
4. Call structured `reflect` on the same bank with a small schema -> times out

#### Plain reflect (works)

```json
{
  "query": "这个 bank 里记住了什么？请简要回答。",
  "budget": "low",
  "max_tokens": 220,
  "include": {
    "facts": {},
    "tool_calls": { "output": false }
  }
}
```

#### Structured reflect (fails)

```json
{
  "query": "请把这个 bank 的关键信息结构化输出。",
  "budget": "low",
  "max_tokens": 260,
  "include": {
    "facts": {},
    "tool_calls": { "output": false }
  },
  "response_schema": {
    "type": "object",
    "properties": {
      "summary": { "type": "string" },
      "facts": {
        "type": "array",
        "items": { "type": "string" }
      }
    },
    "required": ["summary", "facts"],
    "additionalProperties": false
  }
}
```

#### `retain` succeeds

```json
{
  "success": true,
  "items_count": 2,
  "usage": {
    "input_tokens": 6332,
    "output_tokens": 2989,
    "total_tokens": 9321,
    "cached_tokens": 0
  }
}
```

[...]

PR #3089, opened 2026-07-30
fix(llm): explicit per-operation thinking control for Qwen/vLLM providers

### LINKED ISSUE #2630: OpenAI-compatible verification can fail on reasoning models because max_tokens=100 is exhausted by reasoning

jtstothard:
## Related issues

- **Fixes #2630** — OpenAI-compatible verification fails on reasoning models because `max_tokens=100` is exhausted by reasoning. That issue raised the verification token cap, but the deeper fix is explicit per-operation thinking control: disable thinking for utility operations (retain, consolidation, verification) so the token budget goes to visible output.
- **Relates to #2431** — Structured reflect with `response_schema` times out with empty content and `finish_reason=length`. Same failure class: reasoning tokens consuming the budget before visible output.

## Problem

Qwen/vLLM-compatible servers with server-side `reasoning: auto` can consume the entire completion token budget on hidden reasoning. When callers omit an explicit thinking flag, short utility requests return **empty `content`** with `finish_reason: length` — the response is entirely hidden reasoning with no visible output.

This affects all Hindsight operations (retain, reflect, consolidation, default) when pointed at a Qwen/vLLM-compatible endpoint.

## Root cause

Hindsight threaded `config.llm_extra_body` verbatim into every provider without explicit per-operation thinking control. The server-side `reasoning: auto` default then consumed the whole budget on reasoning tokens before producing any visible output.

## Solution

Add `_operation_extra_body()`, which injects `chat_template_kwargs.enable_thinking` into the configured `extra_body` with an explicit per-operation value:

- **retain, consolidation, default operations**: `enable_thinking=False`
- **reflect operations**: `enable_thinking=True`

Only OpenAI-compatible providers receive the injection — the field is Qwen/vLLM-specific and would be rejected or ignored by Gemini/Vertex, Anthropic, LiteLLM, Bedrock, and llama.cpp. Non-eligible providers keep the configured `extra_body` untouched.

### Provider gating

`_OPENAI_COMPATIBLE_PROVIDERS` is an explicit allowlist matched against the actual provider factory paths in `llm_wrapper.create_llm_provider()`:
- `openai`, `groq`, `ollama`, `ollama-cloud`, `lmstudio`, `minimax`, `deepseek`, `volcano`, `openrouter`, `requesty`, `zai`, `opencode-go`, `atlas` → `OpenAICompatibleLLM` (accepts `extra_body`)
- `fireworks`, `nous` → dedicated factory branches that forward `extra_body`
- Excluded: `llamacpp` (factory drops `extra_body`), `gemini`, `vertexai`, `anthropic`, `litellm`, `litellmrouter`, `bedrock`

### Semantics

- Configured `extra_body` is deep-copied — never mutated
- Malformed `chat_template_kwargs` (string, None, etc.) is safely replaced
- `None` provider is a safe no-op (returns body unchanged)
- Multi-LLM chain members inherit the same operation flag

## Files changed

- `hindsight-api-slim/hindsight_api/engine/memory_engine.py` — `_operation_extra_body()` helper, `_OPENAI_COMPATIBLE_PROVIDERS` allowlist, all 6 call sites updated
- `hindsight-api-slim/tests/test_operation_thinking_contract.py` — new contract tests

## Tests

```
cd hindsight-api-slim && uv run pytest tests/test_operation_thinking_contract.py tests/test_multi_llm_config.py tests/test_llm_extra_body.py tests/test_minimax_thinking_body.py -q
```

67 passed in 4.79s (only expected missing `.env`/local-ML dependency warnings).

## Validation scope

- Transport-level `chat_template_kwargs.enable_thinking` forwarding was validated against a Qwen-compatible llama.cpp/Lemonade server (OpenAI-compatible wire format).
- No live Qwen Portal API request was tested; the implementation uses Qwen's documented `enable_thinking` request field.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -453,13 +453,49 @@ class _LLMCallDefaults:
         }
 
 
+# Providers that speak the OpenAI completions wire format and therefore
+# accept ``extra_body.chat_template_kwargs.enable_thinking`` (the Qwen/vLLM
+# thinking-mode flag). Other providers (Gemini/Vertex, Anthropic, LiteLLM,
+# Bedrock) have their own generation-config spaces and would reject or ignore
+# the field — so we gate the per-operation injection on this set.
+_OPENAI_COMPATIBLE_PROVIDERS = frozenset(
+    {
+        "openai",
+        "groq",
+        "ollama",
+        "ollama-cloud",
+        "lmstudio",
+        "minimax",
+        "deepseek",
+        "volcano",
+        "openrouter",
+        "requesty",
+        "zai",
+        "opencode-go",
+        "atlas",
+        "nous",
+        "fireworks",
+        "llamacpp",
+    }
+)
+
+
 def _operation_extra_body(
     configured: dict[str, Any] | None,
     *,
     enable_thinking: bool,
+    provider: str | None,
 ) -> dict[str, Any]:
-    """Apply the explicit per-operation Qwen/vLLM thinking contract."""
+    """Apply the explicit per-operation Qwen/vLLM thinking contract.
+
+    Only OpenAI-compatible providers receive ``chat_template_kwargs``; other
+    providers (Gemini/Vertex, Anthropic, LiteLLM, Bedrock) keep the configured
+    extra body untouched, avoiding a provider-specific field being forwarded
+    where it is not understood.
+    """
     body = copy.deepcopy(configured or {})
+    if not provider or provider.lower() not in _OPENAI_COMPATIBLE_PROVIDERS:
+        return body
     template = body.get("chat_template_kwargs")
     if not isinstance(template, dict):
         template = {}
@@ -496,7 +532,7 @@ def _member_to_llm(
         base_url=member.base_url,
         model=member.model,
         reasoning_effort=member.reasoning_effort or config.llm_reasoning_effort,
-        extra_body=_operation_extra_body(member.extra_body, enable_thinking=enable_thinking),
+        extra_body=_operation_extra_body(member.extra_body, enable_thinking=enable_thinking, provider=member.provider),
         default_headers=member.default_headers or config.llm_default_headers,
         ollama_num_ctx=config.llm_ollama_num_ctx,
         bedrock_service_tier=member.bedrock_service_tier,
@@ -1161,7 +1197,7 @@ class MemoryEngine(MemoryEngineInterface):
             base_url=memory_llm_base_url,
             model=memory_llm_model,
             reasoning_effort=config.llm_reasoning_effort,
-            extra_body=_operation_extra_body(config.llm_extra_body, enable_thinking=False),
+            extra_body=_operation_extra_body(config.llm_extra_body, enable_thinking=False, provider=memory_llm_provider),
             default_headers=config.llm_default_headers,
             ollama_num_ctx=config.llm_ollama_num_ctx,
             litellmrouter_config=config.llm_litellmrouter_config,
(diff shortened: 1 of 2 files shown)


nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/3159

PR #3092, opened 2026-07-31
feat(worker): make the graceful-shutdown timeout configurable

ai-ag2026:
## Summary

The worker poller's graceful-shutdown timeout is hardcoded to `30.0` in both
shutdown paths (`worker/main.py`, and the API-embedded poller in
`api/http.py`). An in-flight retain is an LLM call that may legitimately run
for minutes — `HINDSIGHT_API_LLM_TIMEOUT` defaults far above 30s — so every
service stop cancels it mid-flight and loses the operation.

This adds `HINDSIGHT_API_SHUTDOWN_GRACE` (default `30.0`, behaviour
unchanged unless set), wired through `HindsightConfig` like the other
operational knobs, and used at both call sites.

## Why

Running Hindsight under systemd, we size this together with
`TimeoutStopSec` so the supervisor never SIGKILLs the process mid-cleanup:
grace comfortably under the supervisor timeout, both above the realistic
LLM-call duration. In production since 2026-07-16; restarts stopped eating
in-flight retains.

## Changes

* `config.py`: `ENV_SHUTDOWN_GRACE` + `DEFAULT_SHUTDOWN_GRACE` +
  `HindsightConfig.shutdown_grace` + `from_env` parsing
* `worker/main.py`, `api/http.py`: `shutdown_graceful(timeout=get_config().shutdown_grace)`
* `.env.example` documented

## Validation

`tests/test_shutdown_grace_config.py`: default-is-30 and env override, in
the pattern of the neighbouring config-wiring test files (2 passed).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -292,3 +292,4 @@ HINDSIGHT_API_LOG_LEVEL=info
 # When set, visitors see a login page and must enter the key before
 # accessing the dashboard or any /api/* routes (except /api/health).
 # HINDSIGHT_CP_ACCESS_KEY=your-shared-secret-key
+# HINDSIGHT_API_SHUTDOWN_GRACE=30  # Seconds the worker poller waits for in-flight operations on shutdown; size together with your supervisor stop timeout (e.g. systemd TimeoutStopSec).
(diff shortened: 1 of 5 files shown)


nicoloboschi (changes requested):
please add the new field to the docs

PR #3093, opened 2026-07-31
fix(recall): actually increment access_count on recall

ai-ag2026:
## Summary

`memory_units.access_count` exists, is indexed (`idx_mu_access_count`) and
is consulted by ranking `ORDER BY` clauses — but nothing ever writes it.
The column stays `0` forever, so every recency/frequency signal built on it
is inert. (The `access_count_update` task type appears in a comment only.)

This gives the column its writer: recall bumps `access_count` for exactly
the units it returned.

## Design points

* **Fire-and-forget on the recall path.** A ranking signal must never fail
  or slow down the read; the bump is scheduled after the result is
  assembled, and any error is logged, never propagated.
* **Through the task-backend abstraction**, not a raw pool call — so it
  works on both the asyncpg pool and the SQLAlchemy/Oracle backend.
* Scoped strictly to the returned unit ids.

## Validation

`tests/test_access_count_bump.py` (new): the bump happens for recalled
units, is capped to the returned set, failures never propagate into the
recall result, and the write goes through the backend abstraction —
6 passed on current `main`.

## Notes

Running in production since 2026-07-16 (PostgreSQL backend); the column has
been accumulating real values since.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -1070,6 +1070,10 @@ class MemoryEngine(MemoryEngineInterface):
         self._dialect: SQLDialect | None = None
         # Connection pool — set from backend.get_pool() for backward compatibility
         self._pool = None
+        # Strong refs for fire-and-forget access-count bumps (recall path).
+        # Without a reference the event loop may GC the task mid-flight and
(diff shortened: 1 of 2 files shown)


nicoloboschi:
it's deprecated because useless and has high impact on reads - I will follow up with a drop column migration

PR #3094, opened 2026-07-31
fix(consolidation): guard _execute_update_action against FK violation on 0-row UPDATE

ai-ag2026:
## Summary

If the observation row is concurrently deleted or invalidated, the `UPDATE`
in `_execute_update_action` matches 0 rows — and the `observation_history`
INSERT that follows carries an `observation_id` FK onto `memory_units`, so
appending history for the now-missing row raises `ForeignKeyViolationError`.
Because integrity violations are (correctly) not retried, the consolidation
operation is marked failed for a row that simply no longer exists.

This captures the `UPDATE` status in the SQL branch and bails out before
writing history when 0 rows matched (debug log). The store branch is an
upsert and cannot hit the 0-row case, so it needs no guard.

## Why

Hit in production under concurrent curation: consolidation raced an
invalidation, and the resulting FK violation marked the whole operation
failed. Running with this guard since 2026-07-16.

## Validation

Mock-level tests in `tests/test_integrity_violation_not_retried.py`.
Verified both directions: **without** the guard two tests go red
(`2 failed, 2 passed`); **with** it `4 passed`. The file's three
embedded-postgres tests could not run in our sandbox (embedded PG cannot
create TCP sockets there) — they error identically on pristine `main`, so
that is an environment bound, not an effect of this change.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -2114,7 +2114,7 @@ async def _execute_update_action(
     t0 = time.time()
     store = get_memories()
     if store.writes_memory_rows_in_sql:
-        await conn.execute(
+        update_status = await conn.execute(
             f"""
             UPDATE {fq_table("memory_units")}
             SET text = $1,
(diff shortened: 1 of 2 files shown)


handnewb (approved):
Good defensive fix — FK violation on 0-row UPDATE is a real edge case in consolidation. The guard prevents silent data corruption. Well-tested with 154 lines of coverage.

nicoloboschi:
superseded by #3161

PR #3098, opened 2026-07-31
fix(integrations-claude-code): resolve deleted subagent worktree cwd to parent repo bank

### LINKED ISSUE #3096: Claude Code: deleted subagent worktrees create orphan agent-* banks
## Bug Description

Configuration:

```json
{
  "dynamicBankId": true,
  "resolveWorktrees": true
}
```

```text
```

## Steps to Reproduce

   ```text
   ```

   ```python
   cfg = {
       "dynamicBankId": True,
       "resolveWorktrees": True,
   }
   derive_bank_id(
       cfg,
   )
   ```

## Expected Behavior

## Actual Behavior

```text
agent-deadbeef
```

Current tests cover:

- regular repositories
- existing worktrees
- non-git directories
- missing git/timeouts

## Version

- Hindsight server: 0.8.6

## Related Work

valda:
Fixes #3096

## Problem

With project-scoped dynamic banks (`dynamicBankId: true`, `dynamicBankGranularity: ["project"]`, `resolveWorktrees: true`), the async `Stop` retain hook can fire after an isolated Claude Code subagent worktree under `<repo>/.claude/worktrees/agent-*` has already been cleaned up. `git -C <cwd> rev-parse --git-common-dir` fails for the deleted path and `_resolve_project_name()` falls back directly to `os.path.basename(cwd)`, silently creating orphan banks named after the generated worktree directory (e.g. `agent-a33c4d636c3472e69`) and fragmenting project memory away from the intended bank. Observed in production: one parent session produced 5 orphan `agent-*` banks, one of which accumulated 598 memory units (details in #3096).

## Fix

Two changes to `scripts/lib/bank.py`:

1. **Extract git resolution into `_project_name_from_git()` and fix the common-dir parse.** The old `basename(dirname(git_common_dir))` assumed the common dir always ends in `.git`. For a submodule the common dir is `<super>/.git/modules/<name>`, so the resolver returned the meaningless name `modules` even for a *live* submodule cwd. The parse now handles:
   - regular repo / linked worktree: `<repo>/.git` → `repo` (unchanged)
   - bare repo gitdir: `<path>/repo.git` → `repo` (was: parent directory name)
   - submodule: `<super>/.git/modules/<name>` → `<name>` (was: `modules`)

2. **Retry from the nearest existing ancestor when the cwd is deleted.** If (and only if) the supplied cwd no longer exists on disk, git resolution is retried once from the nearest existing ancestor, so a removed ephemeral workspace still resolves to its containing repository. This is generic — no hard-coding of `.claude/worktrees`. Live directories (including live non-git directories) keep the existing behavior unchanged; the basename fallback remains for deleted paths with no repository ancestor.

The submodule parse fix also makes the ancestor retry consistent: a deleted path under a live submodule now resolves to the same name a live cwd would produce.

## Behavior note

Users who work inside git submodules will see their bank resolve to the submodule name instead of the previous (buggy) `modules`. This is a bug fix but technically a bank-id change for that configuration.

## Tests

- Mocked regression tests for the common-dir parse: submodule, nested submodule, bare repo gitdir.
- Real-git regression tests (`git init` against `tmp_path`, no mocks) for the deleted-cwd path: deleted worktree under an existing repo resolves to the repo basename (also asserted end-to-end through `derive_bank_id()`), deleted path outside any repo keeps the basename fallback, and a live non-git directory is unaffected.

All 204 tests in `hindsight-integrations/claude-code` pass; `ruff format --check` clean with the shared `ruff.toml`.

### DIFF
--- a/hindsight-integrations/claude-code/scripts/lib/bank.py
+++ b/hindsight-integrations/claude-code/scripts/lib/bank.py
@@ -29,25 +29,18 @@ DEFAULT_BANK_NAME = "claude-code"
 VALID_FIELDS = {"agent", "project", "session", "channel", "user"}
 
 
-def _resolve_project_name(cwd: str, config: dict) -> str:
-    """Resolve the project name from the working directory.
+def _project_name_from_git(cwd: str):
+    """Ask git for the repository name at ``cwd``.
 
-    When resolveWorktrees is enabled (default), detects git worktrees and
-    resolves to the main repository basename so that all worktrees of the
-    same repo share the same bank.
+    Returns the repository basename, or None when ``cwd`` cannot be
+    resolved (not inside a repo, directory missing, git unavailable).
 
-    For a regular repo at /home/user/myproject:
-        git-common-dir → /home/user/myproject/.git → basename "myproject"
-
-    For a worktree at /home/user/myproject-wt1 linked to /home/user/myproject:
-        git-common-dir → /home/user/myproject/.git → basename "myproject"
+    The name is derived from --git-common-dir so that all worktrees of
+    the same repo share one name:
+      - regular repo / linked worktree: <repo>/.git            → "repo"
+      - bare repo:                      <path>/repo.git        → "repo"
+      - submodule:                      <super>/.git/modules/<name> → "<name>"
     """
-    if not cwd:
-        return "unknown"
-
-    if not config.get("resolveWorktrees", True):
-        return os.path.basename(cwd)
-
     try:
         result = subprocess.run(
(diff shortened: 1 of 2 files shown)


valda:
Closing: #3096 was closed as resolved by the successor **Coding Agents** plugin, which doesn't wire `SubagentStop` and drops `isSidechain` transcript lines — so an ephemeral subagent worktree cwd never reaches bank derivation. This fix targets the superseded per-agent `claude-code` plugin, so it's no longer worth merging.

The branch stays available if anyone still on the old plugin needs it.

PR #3110, opened 2026-07-31
fix(claude-code): keep project resolution stable when the worktree is gone

### LINKED ISSUE #3096: Claude Code: deleted subagent worktrees create orphan agent-* banks

jouve:
Fixes #3096.

## The failure

`_resolve_project_name()` probes `cwd` with `git rev-parse --git-common-dir`. The `Stop` retain hook is asynchronous, so an ephemeral subagent worktree can already have been removed by the time it runs. The probe fails, the basename fallback takes over, and the leaf directory becomes the project identity — the reported session produced five orphan `agent-<hash>` banks, one of them holding 598 memory units that belonged to the repository.

## The fix

`CLAUDE_PROJECT_DIR` is probed before falling back to the basename. Claude Code exports the project root into the hook process, and unlike the `cwd` carried on the hook's stdin — [documented](https://code.claude.com/docs/en/hooks) as *"the current working directory when the hook is invoked"* — it does not follow the agent.

Measured on Claude Code 2.1.220, with a hook logging both values across an `EnterWorktree` call:

```
event              = PostToolUse
CLAUDE_PROJECT_DIR = /tmp/myproject
stdin cwd          = /tmp/myproject/.claude/worktrees/test3
```

So the value that survives the worktree's deletion is exactly the one that still names the repository.

## Why no second fallback

The issue suggests retrying from the nearest existing ancestor of `cwd`. That works too, but it is redundant here: `CLAUDE_PROJECT_DIR` has been exported to hook commands since Claude Code **1.0.58** (*"Hooks: Added CLAUDE_PROJECT_DIR env var for hook commands"*), long before this plugin existed. Should it ever be missing, the empty candidate is skipped without spawning git and resolution degrades to today's behaviour.

## Existing banks do not move

The new candidate runs **only after** the `cwd` probe has failed. Any working directory git can still resolve keeps its historical answer, so no one's memory is silently re-homed by upgrading. `test_resolvable_cwd_never_consults_the_fallbacks` pins that: with a resolvable `cwd` and `CLAUDE_PROJECT_DIR` deliberately pointing at a different repository, the result is unchanged and `git` is invoked exactly once.

## Tests

Three cases added to `TestResolveProjectNameDeletedWorktree`, covering the deleted worktree, the no-regression guarantee above, and the basename fallback when nothing resolves. The existing git probe moved into `_probe_git_root()` unchanged. Full suite: 200 passed.

### DIFF
--- a/hindsight-integrations/claude-code/CHANGELOG.md
+++ b/hindsight-integrations/claude-code/CHANGELOG.md
@@ -16,6 +16,17 @@
   `read operation timed out` on requests the server completes successfully.
   Does not affect the health check, which stays at 5s. Fixes #1575.
 
+### Fixed
+
+- Project resolution no longer degrades to the leaf directory name when the
+  working directory has been deleted before the hook runs. The `Stop` retain
+  hook is asynchronous, so an ephemeral subagent worktree could already be gone,
+  producing throwaway `agent-<hash>` banks that fragment project memory
+  (#3096). `CLAUDE_PROJECT_DIR`, which Claude Code exports into the hook process
+  and which does not follow the agent into a worktree, is probed before that
+  fallback; a working directory git can still resolve keeps its previous
+  answer, so existing banks do not move.
+
 ### Changed
 
 - Tags that resolve to an empty namespace content (e.g. `"user:"` when
--- a/hindsight-integrations/claude-code/scripts/lib/bank.py
+++ b/hindsight-integrations/claude-code/scripts/lib/bank.py
@@ -29,28 +29,13 @@ DEFAULT_BANK_NAME = "claude-code"
 VALID_FIELDS = {"agent", "project", "session", "channel", "user"}
 
 
-def _resolve_project_name(cwd: str, config: dict) -> str:
-    """Resolve the project name from the working directory.
-
-    When resolveWorktrees is enabled (default), detects git worktrees and
-    resolves to the main repository basename so that all worktrees of the
-    same repo share the same bank.
-
-    For a regular repo at /home/user/myproject:
-        git-common-dir → /home/user/myproject/.git → basename "myproject"
-
-    For a worktree at /home/user/myproject-wt1 linked to /home/user/myproject:
-        git-common-dir → /home/user/myproject/.git → basename "myproject"
-    """
-    if not cwd:
-        return "unknown"
-
-    if not config.get("resolveWorktrees", True):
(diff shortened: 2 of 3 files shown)


handnewb (approved):
Solid robustness fix — worktree deletion shouldn't break project resolution. The stable fallback prevents cascading failures when ephemeral worktrees are cleaned up.

nicoloboschi:
Superseded by #3410, which ports this to **Coding Agents** — the package that supersedes the per-agent `claude-code` plugin.

Your diagnosis carried over intact: the Stop hook runs after the fact, so an ephemeral worktree can already be gone, the git probe fails, and the leaf basename becomes the project identity. Same failure there, same throwaway banks.

The one thing I changed is the rescue. `CLAUDE_PROJECT_DIR` works, but only for the harness that exports it, and Coding Agents drives eleven — of which only Claude Code is known to publish such a variable, so naming the others would be registering behaviour nothing implements. The primary mechanism is now a walk up to the nearest ancestor that still exists, which needs nothing from the harness and covers the reported shape directly (`<repo>/.agent/worktrees/agent-x` deleted → `<repo>` resolves). For a live directory the walk returns it unchanged, so no existing bank moves — the property your PR was careful about too.

`CLAUDE_PROJECT_DIR` is kept as the last rescue for the case the walk cannot reach, which your approach handles and mine wouldn't: a **linked** worktree is a sibling of the repository, not a child, so walking up from it leaves the repo entirely. That case is covered by a test that exists because your PR made me think about it.

It also fixes something adjacent that had been flagged and left: `basename("/")` is `""`, which produced bank ids like `coding-agent::` naming nothing.

`npx @vectorize-io/hindsight-coding-agents install claude-code` ([docs](https://hindsight.vectorize.io/sdks/integrations/coding-agents)). Thanks — this was a good catch and the fix is better for having started from yours.

PR #3111, opened 2026-07-31
feat(claude-code): expose the granularity fields to retainTags and recallTags

jouve:
> Independent of #3110, but best merged after it — see *Ordering* below.

## The gap

`dynamicBankGranularity` already resolves five context dimensions — `agent`, `project`, `session`, `channel`, `user` — but they stopped at bank composition. `retainTags` only ever saw `{session_id}` and `{user_id}`; `recallTags` resolved no templates at all (`recall.py` read `config.get("recallTags")` raw, so `"project:{project}"` would have been sent with the braces intact).

Scoping memory per project therefore meant either a `settings.json` per repository with the name hardcoded on both sides, or `dynamicBankId`, which silos each project into its own bank and gives up cross-project recall entirely.

## The change

`build_field_map()` moves out of `derive_bank_id()` — it was computed *after* the static-mode early return, so nothing outside that function could reach it — and both consumers now share it. A bank and the tags written into it can no longer disagree about which project they belong to, and one global config line covers every repository:

```json
{
  "retainTags": ["project:{project}", "session:{session_id}"],
  "recallTags": ["project:{project}"],
  "recallTagsMatch": "any"
}
```

`"any"` keeps untagged memories visible alongside the current project's, so memories retained before tagging was enabled are not orphaned.

## Backward compatibility

`{session}` and `{user}` join the namespace next to the existing `{session_id}` and `{user_id}`, which keep their raw semantics — empty rather than `unknown`/`anonymous`. That distinction is load-bearing: the empty-namespace drop relies on it to discard `user:` on a host that set no `HINDSIGHT_USER_ID`, where `user:anonymous` would be a real tag matching every unconfigured session. `TestBuildTemplateVars` pins it.

The tag-resolution loop that was inline in `retain.py` moved to `resolve_tags()` unchanged, and `recall.py` now calls it too — including for the per-bank overrides in `recallAdditionalBankFilters`.

## Ordering

`{project}` resolves through `_resolve_project_name()`, which #3110 fixes. Merged alone, this PR hands worktree-isolated subagents tags like `project:agent-a33c4d63` — the same fragmentation as #3096, moved from bank ids into tags. The two touch different functions and do not conflict outside a trivial CHANGELOG hunk.

## Tests

7 cases added across `TestBuildFieldMap`, `TestBuildTemplateVars` and `TestResolveTags`. No existing test modified. Full suite: 204 passed.

### DIFF
--- a/hindsight-integrations/claude-code/CHANGELOG.md
+++ b/hindsight-integrations/claude-code/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ### Added
 
+- The five `dynamicBankGranularity` fields (`{agent}`, `{project}`, `{session}`,
+  `{channel}`, `{user}`) are available as template variables, and `recallTags`
+  now resolves templates like `retainTags` already did. Scoping memory per
+  project no longer needs a `settings.json` per repository — the same
+  `"project:{project}"` on both sides writes and reads the tag.
 - `{user_id}` template variable for `retainTags` and `retainMetadata`, resolved
   from the `HINDSIGHT_USER_ID` env var (empty string if unset). Enables
   machine-independent per-user memory scoping without hardcoding user ids in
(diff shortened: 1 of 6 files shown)


jouve:
Opened ahead of its dependency by mistake — closing to keep the queue clean. The branch stays as-is; I'll reopen once #3110 has landed, since merging this one first would move the same fragmentation from bank ids into tags. Sorry for the noise.

PR #3116, opened 2026-07-31
feat(devin-cli): add Devin CLI hooks integration

### LINKED ISSUE #3096: Claude Code: deleted subagent worktrees create orphan agent-* banks

ichoosetoaccept:
## Why

Devin CLI users have no first-party way to give their agent persistent memory across
sessions. Every new session starts cold: stack, preferences, and past decisions have to be
re-explained, and whatever context existed is lost when the window compacts.

This adds a Hindsight integration for Devin CLI in the same shape as the existing
coding-agent integrations — Python hook scripts that read JSON from stdin and write JSON to
stdout, plus an MCP server for explicit knowledge tools. After install, every prompt
reaches the model with the most relevant memories from past sessions pre-injected, and
conversations are retained during the session and again at session end.

It is a port of the Claude Code plugin: same four lifecycle hooks, same `settings.json`
schema, same env var names, same `agent_knowledge_*` MCP tools. A shared external Hindsight
server and a single `~/.hindsight` config can serve both integrations, and a parity test
keeps the schemas from drifting apart.

## How memory flows

| Devin hook | Script | What it does | Why this hook |
| --- | --- | --- | --- |
| `SessionStart` | `session_start.py` | Registers the session, verifies Hindsight is reachable, and pre-warms the local `hindsight-embed` daemon in the background. | Fire-and-forget setup that must never block the agent — 5s budget. |
| `UserPromptSubmit` | `recall.py` | Queries Hindsight for memories relevant to the prompt and injects them as `hookSpecificOutput.additionalContext`. | Synchronous, runs every prompt; the only place recall can reach the model. |
| `Stop` | `retain.py` | Reads the session's unretained messages and POSTs them to the retain API with `async=true`, every configured N turns. | Fires each turn; the server processes retention in the background. |
| `SessionEnd` | `session_end.py` | Forces a final retain (so sessions shorter than `retainEveryNTurns` still land), deregisters the session, and stops the daemon once the last session is gone. | Guarantees short sessions are stored before the process exits. |

Explicit memory is available too: the MCP server exposes the `agent_knowledge_*` tools
(recall, ingest, and knowledge-page CRUD) over stdio.

## Two Devin-specific constraints

Both are worked around rather than solved, because Devin CLI's plugin system shipped in
v2026.7.16 (June 16, 2026) and is still labelled beta — "behavior and configuration may
change in future releases". Both were re-verified against CLI 3000.3.22 before submitting.

**Setup is two steps.** Devin exposes exactly one environment variable to hook processes,
`DEVIN_PROJECT_DIR` (the user's project, not the plugin), and offers no plugin-root
substitution for hook commands. A relative command is handed to the shell verbatim and
resolves against the directory the session was launched from, not the directory the hooks
file lives in — confirmed with a probe hook whose `hooks.v1.json` sat at a repo root while
the script it named existed only in the subdirectory `devin` was started from: it ran,
reporting `PWD` as the launch directory. So a plugin cannot name its own scripts. After
`devin plugins install`, `scripts/install.py` writes absolute-path entries into Devin's
documented config locations; it is idempotent, replaces a previous version's entries on
upgrade, and leaves unrelated third-party entries alone.

One qualifier, since it is easy to check and conclude otherwise: the CLI binary does
contain `${CLAUDE_PLUGIN_ROOT}`, alongside the `${env:VAR}` / `${file:...}` interpolation
the MCP docs describe — but it sits in the **MCP** config importer, not the hook path. If it
resolves for a plugin's `mcp_config.json`, the MCP server could be declared rather than
installed, and only the four hooks would need the installer. It is undocumented and I could
not verify it here (`devin plugins install` requires an org policy check this environment
cannot reach), so the installer registers both — one install path, one uninstall path.
**If you or the Devin team can confirm the substitution, I will move the MCP half into
`mcp_config.json` and shrink the installer.**

**Retain reads Devin CLI's session SQLite database.** The shipped reference documents the
entire `Stop` payload as `stop_hook_active`, plus the universal `session_id` / `prompt_id` —
there is no `transcript_path`. Devin does persist every session locally keyed by that same
`session_id`, and `lib/devin_transcript.py` reads it. The read path is deliberately
defensive: a missing file, table, or column, or an unparseable payload, degrades to an empty
transcript, so a future CLI release that changes the schema switches auto-retain off rather
than breaking sessions. `session_end.py` keys off `session_id` for the same reason.

Everything else matches the Claude Code plugin's design: a local stdio MCP server via
`run_mcp.sh`, and a local `hindsight-embed` daemon (or an external server). The daemon runs
on port 9078 so it does not collide with the Claude Code plugin's on 9077.

## Configuration

`~/.hindsight/devin-cli.json` for personal overrides, stable across plugin updates and
matching `~/.hindsight/claude-code.json`:

```json
{
  "hindsightApiUrl": "https://api.hindsight.vectorize.io",
  "hindsightApiToken": "your-api-key",
  "bankId": "my-devin-memory"
}
```

`settings.json` at the plugin root holds the shipped defaults. Every option also has an
env-var override (`HINDSIGHT_API_URL`, `HINDSIGHT_BANK_ID`, `HINDSIGHT_RECALL_BUDGET`, …) —
the same set the Claude Code plugin documents, which is what the parity test pins.

## What's in the integration

```
hindsight-integrations/devin-cli/
├── README.md, CHANGELOG.md, LICENSE, settings.json
├── pyproject.toml, requirements.txt      # dual-distributed: plugin + pip package
├── .devin-plugin/plugin.json             # plugin manifest
├── scripts/
│   ├── session_start.py                  # SessionStart
│   ├── recall.py                         # UserPromptSubmit
│   ├── retain.py                         # Stop
│   ├── session_end.py                    # SessionEnd
│   ├── install.py / uninstall.py         # hook + MCP registration
│   ├── mcp_server.py, run_mcp.sh         # stdio MCP server + venv bootstrap
│   └── lib/                              # bank, client, config, content, daemon,
│                                         # devin_transcript, llm, state
├── skills/setup/, skills/create-agent/
└── tests/                                # 398 tests
```

## Install

```bash
devin plugins install vectorize-io/hindsight#hindsight-integrations/devin-cli
```

Then register the hooks and MCP server once — easiest by starting a session and asking Devin
to set up Hindsight memory, which invokes the plugin's `setup` skill. The README documents
the direct command too.

## Relationship to `hindsight-devin-desktop`

Worth a maintainer decision before merge. `hindsight-devin-desktop` v0.2.0 (#2692) added
`devin_local.py`, targeting the **Devin Local** agent — which its own docstring describes as
*"the successor agent, shared with the Devin CLI"*. That is the same runtime this plugin
targets, and it writes the same config file (`~/.config/devin/config.json`).

**Both installers preserve each other's entries**, which is the problem: this plugin's
`_OURS_RE` (`scripts/install.py:55`) only matches its own `python3 "…/scripts/<hook>.py"`
command shape, and `devin-desktop`'s `_is_ours_for` (`devin_local.py:182`) only matches
`sys.executable -m hindsight_devin_desktop.hook`. Neither strips the other, so a user with
both installed gets two `SessionStart` recall injections and two `Stop` retain paths against
the same session.

Suggested split, by agent runtime rather than by product: `hindsight-devin-desktop` narrows
to Cascade (`~/.codeium/windsurf/`) and drops `devin_local.py`; this plugin owns the Devin
CLI runtime (`~/.config/devin/`) however it is launched — standalone or driven by Desktop
over ACP. Happy to send that as a follow-up if you agree; it is deliberately out of scope
here so the new plugin can be reviewed on its own.

Separately: `hindsight-docs/blog/2026-07-02-devin-desktop-persistent-memory.md` predates
`devin_local.py` by 11 days and still says Devin Desktop *"doesn't expose lifecycle hooks to
third parties"*. No longer accurate for Devin Local; needs a refresh independently of this
PR.

## Prior art: fixes already reported or proposed against `claude-code`

I went through the open issues and PRs before submitting, because a port that silently
re-fixes what someone has already proposed is worse than useless. Three overlaps:

**Already fixed here, independently and identically — #2999.** `retain.py`'s
`retain_full_window = retention_progress.start_index == 0` narrows the retain slice a second
time after `plan_retention()` has already narrowed it, while `commit_retention()` advances
the cursor past the whole suffix. #2999 changes it to `= True` with the same reasoning; so
does this port. I reached it from a different symptom (an assistant-only tail formatting to
nothing) and found #2999 afterwards. **If #2999 lands first this needs no rebase — the line
is already what that PR makes it.**

**Inherited, and fixed here on the strength of your report — #3026 / #3027.** This plugin
shipped `mcp>=1.0.0`, the same unbounded spec. mcp 2.0 removed `mcp.server.fastmcp`, so a
fresh install resolves to a release `mcp_server.py` cannot start under — and `run_mcp.sh`
only re-pips when `import mcp` fails, which succeeds under 2.x, so the venv is never
repaired. Now `mcp>=1.0.0,<2`, with a second test asserting `mcp_server.py` still imports
from that path, so the pin is lifted when the reason for it goes away rather than outliving
it.

**Adjacent, not conflicting — #2900 / #2913.** #2913 adds a `--check` release guard to
`scripts/release-integration.sh`; this PR changes the version-bump block in the same file.
Different concerns, non-overlapping hunks, but whichever merges second may want a look.

Six further fixes proposed upstream are deliberately **not** adopted here: #2812
(nearest-ancestor `directoryBankMap`), #2897 (`HINDSIGHT_RETAIN_TAGS`), #3024 / #2487 (filter
injected harness turns), #2869 (detach `SessionEnd`), #3110 / #3098 (project resolution when
the worktree is gone), #2493 (seed bank missions only when unset). All apply equally here.
They are left out because adopting them would mean picking a winner between competing PRs —
#3110 and #3098 both fix #3096 — and a new-integration PR is the wrong place to pre-empt
those reviews. This port matches current `main` and will follow whichever lands.

## Other divergences from the Claude Code plugin

Ten rounds of automated review ran against this branch before submission. The great majority
of what it surfaced was not in the new code — it was in files copied from
`hindsight-integrations/claude-code`, several byte-for-byte. Rather than carry them over,
they are fixed here.

**`claude-code` is untouched by this PR**, so the two plugins now differ in 26 places. Each
is a candidate for a follow-up PR against `claude-code`, which I am happy to send — I did not
want a new-integration PR to also be a behaviour change to an existing one. The full list
with per-item reasoning is in the commit message; the headline items:

- **Security.** `lib/daemon.py` no longer passes `HINDSIGHT_API_LLM_API_KEY` as a
  `profile create --env` argument, where any local user could read it from the process
  listing. It reaches the daemon through the subprocess environment instead.
- **The daemon was shared but its lifecycle was not.** The stop decision hung off a single
  global marker, so the first session to end stopped the daemon out from under the others.
  Sessions now register under an interprocess lock and it stops only when the last one
  deregisters; ownership is released in the same locked write that decides to stop, and
  handed back if the stop then fails. A failed stop is now *detected* at all — `_run_embed()`
  does not pass `check=True`, so "could not stop" came back as an ordinary
  `CompletedProcess` and read as success.
- **Retention silently dropped messages.** Checkpoints were committed before the retain
  request rather than after, so a failed request skipped the messages it never sent.
- **`_file_lock()` yielded unlocked when it could not acquire the lock**, so a
  read-modify-write ran with no interprocess synchronisation while looking exactly like it
  held the lock — the corruption the lock exists to prevent.
- **A malformed setting took a hook down with it.** A string where a list belongs iterates
  one character at a time; a list where a dict belongs raises `AttributeError`; a string
  where an int belongs raises `TypeError`. All inside a hook, so one mistyped optional
  setting switched off the thing it was added to tune. `load_config()` now checks each value
  against the type of its `DEFAULTS` entry.
- **`UnicodeDecodeError` slipped through eight handlers** — it subclasses `ValueError`, not
  `OSError`, so it escaped `except (json.JSONDecodeError, OSError)` at the settings reader,
  all four hooks' stdin readers, the installer's config read, the git worktree probe, and the
  session-db row parser. A single bad byte stopped every hook.
- **Neither hook can exit nonzero any more.** Both document *"0 — always"* and both exited 2
  under `debug`, so turning debug on to investigate a recall failure escalated it from "no
  memories this turn" to a rejected prompt.

Also fixed: an unescaped SQLite URI, an unescaped `page_id` in MCP URL paths, a `document_id`
that stripped file extensions (so `README.md` and `README.txt` collided and the second
overwrote the first), a `"name": null` on a `tool_use` block that aborted a whole transcript,
LRU rather than lexicographic session eviction, and `run_mcp.sh` absolutising its data
directory before `cd`.

One inherited behaviour is deliberately left alone: `_is_channel_message_tool()` classifies
any `mcp__*` tool carrying a text-like input field as a channel reply, which catches
non-messaging tools such as `mcp__email__send`. That heuristic is a documented tradeoff — it
recognises any channel plugin without hardcoding tool names, and the obvious alternative
defeats the property it was built for. Worth changing deliberately alongside `claude-code`.

## CI / release wiring

- New `test-devin-cli-integration` job in `.github/workflows/test.yml`, gated on a new
  `integrations-devin-cli` path filter and wired into the aggregate status check.
- `devin-cli` added to `VALID_INTEGRATIONS` in `scripts/release-integration.sh`.
- Added to `hindsight-integrations/README.md` and the docs integrations index.
- **One fix outside the new plugin:** `scripts/release-integration.sh` bumped only the first
  match of an `if`/`elif` chain, so releasing any integration with more than one version
  source updated `pyproject.toml` and silently left the plugin manifest stale. That already
  affects `cursor` and `zcode`. It now bumps every source present. (Related to #2900; see the
  #2913 note above.)

## Verification

```bash
cd hindsight-integrations/devin-cli
uv run pytest tests -q
# 398 passed

uv run --no-project ruff check --config ../../ruff.toml .
uv run --no-project ruff format --check --config ../../ruff.toml .
# All checks passed! / 34 files already formatted
```

No live Hindsight server or Devin CLI is required for the suite — hooks are tested
end-to-end with a fake HTTP client, mocked stdin/stdout, and an isolated `HOME`. Every
regression test for the fixes above was confirmed to fail against the unpatched source; the
concurrency ones spawn real subprocesses, because the guarantee is an interprocess one. A
parity test asserts `settings.json`, `DEFAULTS`, and `ENV_OVERRIDES` stay identical to the
Claude Code plugin apart from the documented divergences, so adding a key to one plugin and
not the other fails CI. `install.py` / `uninstall.py` are verified idempotent against a
config file containing unrelated third-party hook entries.


ichoosetoaccept:
Closing this — superseded by #2522, which landed Devin CLI support in `hindsight-integrations/coding-agents/` a few hours before I pushed the last revision here.

That's the better home for it. Devin CLI there is three entrypoint files, one `HOOK_HARNESSES` declaration, `transcript-devin.ts` and one `HarnessInstaller` — roughly 150 lines against a lifecycle shared with nine other agents — where this PR is a standalone Python plugin reimplementing recall, retain, bank derivation, knowledge tools and installation for a single agent.

Worth recording that the two implementations independently reached the same two conclusions about Devin CLI, since neither is documented by Cognition:

- Devin CLI's hook payloads carry no `transcript_path`, so the conversation has to be read out of `~/.local/share/devin/cli/sessions.db` (`message_nodes`, de-duplicated by `message_id`).
- A plugin-shipped `hooks.json` has no `${DEVIN_PLUGIN_ROOT}`-equivalent, and its relative paths resolve against the project working directory rather than the plugin's install directory — so hook and MCP registration has to be written into `~/.config/devin/config.json` and `~/.config/devin/mcp_config.json`.

Separately: two findings from reviewing this branch apply to code that is still on `main`, independent of anything proposed here. Recording them so they aren't lost with the PR.

**Tool results are counted as user turn boundaries.** In `hindsight-integrations/claude-code/scripts/lib/content.py`, `slice_last_turns_by_user_boundary` treats every `role: "user"` message as a turn boundary (line 185), as does the last-turn retention path (line 336). Anthropic-style tool results are delivered as `role: "user"` messages, so a normal prompt → tool_use → tool_result → answer exchange puts the boundary on the tool result and drops the prompt that caused it. This affects both recall query composition and chunked retention. The `openclaw` integration already filters these out in its own `sliceLastTurnsByUserBoundary`.

**`strip_channel_envelope` matches anywhere in the message.** Same file, line 32, uses `re.search` rather than an anchored match, so a message that merely *mentions* a channel envelope is replaced by that envelope's inner text — `please explain <channel source="x">hello</channel> thanks` is retained as `hello`, discarding what the user actually wrote, from both recall and retention.

No action needed on this PR.

PR #3126, opened 2026-08-01
fix(reflect): stop evidence starvation — budget the LLM-facing rendering, truncate instead of drop, surface evidence_truncated

### LINKED ISSUE #3122: reflect: tool-result envelope isn't counted against max_tokens (~6x inflation) and forced synthesis drops oversized blocks whole — confident 'no information' refusal with 503 citations (post-#668 residual)
### Summary

On 0.8.6, broad reflect questions against a bank with a few hundred small observations deterministically return a confident **"I don't have information"** refusal — while attaching hundreds of citations in `based_on`. This is the same *symptom* as #668, but a different (residual) mechanism; the #668 fix (`source_facts_max_tokens` defaulting to disabled) is present and working.

Two independent defects combine:

1. **Reflect tool-result token budgets count observation *text* only; the serialized envelope is not budgeted.** `tool_search_observations` passes `max_tokens` to `recall_async`, which budgets text tokens — but the tool returns `_prune_nulls(m.model_dump())` for every observation (`id`, `chunk_id`, `document_id`, `context`, `entities`, `metadata`, `scores`, `tags`, `type`, …). Measured on my bank: a request with `max_tokens=8000` returned **184,677 chars serialized (~46k tokens) wrapping 38,065 chars of observation text** — ~5.8× the requested budget. Inside reflect the same call produced a 261,941-char tool result. A single tool call therefore consumes most of `HINDSIGHT_API_REFLECT_MAX_CONTEXT_TOKENS` (default 100k), and the context guard trips on iteration 2–3.

2. **Forced synthesis drops oversized tool results whole instead of truncating.** `build_final_prompt` iterates tool results newest-first and, when a block exceeds the remaining budget, `break`s — discarding that block *and* all older ones, with no per-block truncation. When every tool result is oversized (the normal case given defect 1), the "Retrieved Data" section is empty, and the model correctly answers "no information" — while the endpoint still returns the full citation list.

### Environment

- hindsight-api **0.8.6** (uvx, Python 3.13, macOS arm64)
- Postgres with pg_textsearch; embeddings: local ollama `granite-embedding:278m` (768 dims)
- LLM provider: `openai-codex` (subscription/device-code auth)
- Bank: synthetic benchmark corpus — 462 `world` facts from ~400 documents, ~290 observations, avg observation text 144 chars
- `HINDSIGHT_API_REFLECT_MAX_CONTEXT_TOKENS` unset (default 100_000)

**Model-independent:** reproduced identically with `gpt-5.6-luna`, `gpt-5.6-sol`, `gpt-5.6-terra` as the reflect model, at reasoning efforts `none`, `low`, and `medium`, 16 calls per configuration. Every broad-question failure shows the same signature; no HTTP errors anywhere.

### Reproduction

Any bank with a few hundred smallish observations. Then:

```bash
# 1. Show the envelope inflation directly (no LLM involved):
curl -s -X POST "$BASE/v1/default/banks/$BANK/memories/recall" \
  -d '{"query":"<broad topical query>","fact_type":["observation"],"max_tokens":8000}' \
  | python3 -c 'import json,sys; d=json.load(sys.stdin); r=d["results"]; \
    print(len(r), "observations;", sum(len(x["text"]) for x in r), "text chars;", len(json.dumps(d)), "serialized chars")'
# → 183 observations; 38065 text chars; 184677 serialized chars   (max_tokens=8000 ≈ 32k chars)

# 2. Trigger the failure:
curl -s -X POST "$BASE/v1/default/banks/$BANK/reflect" \
  -d '{"query":"Re-examine everything in memory and synthesize all operational decisions and team conventions that were settled: what was decided, who owns it, and what superseded earlier behavior. Cite the supporting memories for each.","budget":"low","max_tokens":4096,"include":{"facts":{},"tool_calls":{}}}'
```

### Observed (captured transcript)

Response:

- `text`: `"I don't have information: the retrieved context contains no substantive memories describing operational decisions, ownership, team conventions, superseded behaviors, or supporting citations."`
- `based_on.memories`: **503 citations**
- `usage`: `{"input_tokens": 730, "output_tokens": 47, "total_tokens": 777}` — the synthesis call saw **730 input tokens**, i.e. none of the retrieved data
- `trace.tool_calls`: `search_observations` output **261,941 chars**; `recall` output **242,789 chars**

Server log:

```text
WARNING - hindsight_api.engine.reflect.agent - [REFLECT eval-82482] Context budget exceeded on iteration 3: ~182625 tokens >= 100000 limit. Forcing final synthesis.
```

The 730-input-token synthesis together with 503 attached citations is the defect in one line: the caller is told the answer is grounded in 503 memories that the answering model never saw.

### Code sites (0.8.6 wheel)

- `engine/reflect/tools.py` — `tool_search_observations` returns `[_prune_nulls(m.model_dump()) for m in result.results]` plus `source_facts`; `max_tokens` only reaches `recall_async`'s text budgeting. (Post-#668: `source_facts_max_tokens=-1` default confirmed working — the inflation above is pure envelope, not source facts.)
- `engine/reflect/agent.py` (~line 700) — proactive context guard forces final synthesis when accumulated messages exceed `max_context_tokens`.
- `engine/reflect/prompts.py`, `build_final_prompt` (~lines 461–484) — per-block all-or-nothing: `if block_tokens > token_budget: truncated = True; break`. An oversized newest block discards *all* retrieved data, then the model is instructed to say "I don't have information" only "if the retrieved data is truly unrelated" — with an empty Retrieved Data section, that's the inevitable output.

### Suggested fixes

1. Budget the **serialized tool result** (or a slimmed rendering) against `max_tokens` in `tool_search_observations` / `tool_recall` — e.g. render observations for the agent as `{id, text, type}` and drop `entities`/`scores`/`metadata`/`chunk_id`/`context` from the in-loop representation (full objects are still available to the endpoint for `based_on`). On my measurements this alone shrinks the tool result ~5×.
2. In `build_final_prompt`, truncate an oversized block to the remaining budget (e.g. keep the first N observations of the block) instead of dropping it whole — forced synthesis should degrade gracefully, not answer from nothing.
3. Consider not attaching `based_on` citations for evidence the synthesis prompt did not include, or flagging the response (`evidence_truncated: true`) so callers can distinguish a grounded "no information" from this failure mode.

### Version

0.8.6

### LLM Provider

openai-codex

feniix:
Closes #3122.

On banks with a few hundred small observations, broad reflect questions deterministically returned a confident "I don't have information" while attaching hundreds of citations in `based_on` — the synthesis call in the captured transcript saw 730 input tokens alongside 503 citations. Two defects compounded, and this PR fixes both plus adds the caller-facing signal the issue proposed.

## What changed

**1. Opt-in: the LLM sees a slim rendering of tool results; `tool_trace` keeps full objects.**
Tool results serialized full `MemoryFact` dumps (`entities`, `scores`, `metadata`, chunk/document ids, …) into the agent loop — ~6× the token cost of the text the recall budgeter counted. When `reflect_slim_tool_results` is enabled, the agent projects `observations`/`memories` items to `SlimMemoryFact` (`id`, `text`, `fact_type`, `occurred_start/end`, `mentioned_at`, `source_fact_ids`) at the tool-message and `context_history` sites. `mentioned_at` is deliberately kept — the agent system prompt makes it the temporal-supersession signal. `based_on`, the API trace, and persisted reflect responses are unchanged in both modes (full objects still flow through `tool_trace`; regression-tested).

**2. Opt-in: tool entry selection budgeted by what the rendering actually costs.**
For short facts, ids and timestamps dominate the serialized payload (measured 10–18× the text tokens), so text-token budgeting under-counts by an order of magnitude. With the flag enabled, `tool_recall` / `tool_search_observations` keep the leading results whose slim rendering fits `max_tokens` (always at least one), prune `chunks`/`source_facts` to surviving entries, and report the returned count. Trimming happens before `tool_trace`, so citations stay consistent with what the loop actually retrieved.

Both are gated behind **`HINDSIGHT_API_REFLECT_SLIM_TOOL_RESULTS` / `reflect_slim_tool_results` — hierarchical, default `false`, per-bank overridable** — so model inputs are byte-identical to today unless a deployment or bank opts in (default-off is pinned by regression tests). Documented in `configuration.md`, the reflect API reference, and `.env.example` (embed copy synced). Flipping the default later is a one-line change if benchmarks support it.

**3. Always on: forced synthesis truncates oversized blocks instead of dropping them.**
`build_final_prompt` used to drop any over-budget tool-result block *and every older one* (`break`), leaving an empty Retrieved Data section. It now keeps the largest leading prefix of a block's result entries (binary search on rendered size; results are relevance-ordered), with a token-bounded cut fallback for list-less outputs, and returns a typed `FinalPromptResult`. Not gated: dropping all evidence is never desirable behavior.

**4. Always on: the response carries `evidence_truncated`.**
True when the context guard forces synthesis or the final prompt had to truncate retrieved data; false on a normal `done` completion. Threaded `ReflectAgentResult` → `ReflectResult` → `ReflectResponse`; MCP picks it up via `model_dump()`. Callers can now tell an evidence-starved "no information" from a grounded one. Additive and non-breaking. OpenAPI spec, generated clients (Go/Python/TS), and the API reference are regenerated/updated in this PR; the hand-written wrappers return generated types so they need no change; the Rust client regenerates at `hindsight-cli` build time; the control plane passes the reflect response through verbatim.

## How it was built and verified

Strict test-first: every behavior landed as a failing test watched red before the implementation (slim/full split, context-history rendering, budget fitting with chunk/source-fact pruning, per-block truncation shapes, flag threading at agent and HTTP levels, config gating with default-off regression coverage). Highlights:

- `TestSlimToolResultRendering` — slim message + full trace + `based_on` byte-compatibility (opt-in), full envelope by default
- `TestToolResultLlmBudget` — rendering-cost budget fitting, ≤ `max_tokens` asserted on cl100k counts (opt-in), no trimming by default
- `TestBuildFinalPromptTruncation` — oversized-newest, multi-block ordering, token-cut fallback, empty history, unserializable output
- `TestEvidenceTruncatedFlag` + endpoint test — flag semantics on guard/last-iteration/normal paths and HTTP serialization
- `TestReflectSlimToolResultsConfig` — default off, hierarchical/per-bank configurable
- `TestContextOverflowIntegration` (`hs_llm_core`) now asserts `evidence_truncated is True` instead of tolerating the failure mode

Verification: full API suite green (remaining failures are credential-gated real-LLM tests and known parallel-load flakes, all verified passing in isolation), `lint.sh`, `ty check`, generated-file drift clean. Real-LLM runs via `openai-codex` (the provider class from the original report): **6 passes across `gpt-5.6-sol`/`luna`/`terra` (low+medium efforts) and `gpt-5.4-mini`** — the context guard trips at a model-independent ~2.3k tokens with slimming enabled where the full envelope previously blew six-figure budgets, and `evidence_truncated` surfaces end-to-end every time.

## Independent validation on a second bank

Measured on one bank of 781 memories / 315 observations / 411 documents built
from a synthetic corpus, `gpt-5.6-luna`, same bank and same question across all
builds. Reproduced on two independent auth paths (a subscription provider and a
metered API key); the numbered table below is the subscription path.

**The reported symptom, on the released build:**

```
question: "Summarize all operational decisions and team conventions across every project."

answer:    "I don't have information about operational decisions or team
            conventions across the projects in the retrieved memories."
            (119 characters)
citations attached: 246
latency:   6.2s
```

246 memories asserted as provenance for an answer containing nothing.

**How bad this is depends on how the agent chunks retrieval, which is worth
stating plainly.** On the path above the agent made a *single* ~304k-char
`search_observations` call. It exceeds the budget, `break` fires on the first
(newest) block, and nothing at all reaches the model. On a second provider the
same released build made a large call *plus* a smaller follow-up `recall`
(~262k + ~37k chars): the newest block fit, survived, and the model answered
with 30 claims, all grounded. Same build, same bug, same bank — the outcome
turned on whether a small enough block happened to be last.

That is the sharper argument for change 3. `break` does not merely fail when
there is nothing to show; in the second case it silently discarded a 262k-char
block that was almost certainly richer than the 37k one it kept. Truncating
means the large block contributes what fits instead of nothing, in both cases.

**Across builds:**

| | released 0.8.6 | this PR, flag off | this PR, flag on |
|---|---|---|---|
| answer length | 119 chars | 14,179 | 7,135 |
| distinct claims made | **0** (single-sweep retrieval; 30 on a provider that split the retrieval) | **53** | 25 |
| claims grounded in attached evidence | n/a (no claims) | 52/53 (98%) | 24/25 (96%) |
| citations attached | 246 | 300 | 36 |
| largest single tool result | 304,876 chars | 304,982 | 89,531 |
| tool result vs its own `max_tokens` | 7.6–9.5× | 9.5× | **1.8–2.2×** |
| forced synthesis | 40 / 42 calls | 21 / 22 | **2 / 20** |
| context budget exceeded | — | 21 | **0** |
| latency | 6.2s | 78.8s | 61.1s |

"Grounded" = for each claim bullet, at least two distinctive content words also
appear in one of the memories the response attached as `based_on`. It screens
for claims asserted without support, which is the failure mode truncation could
plausibly have introduced.

**Three things this establishes:**

1. **The always-on truncation fix carries the result.** With both opt-in flags
   off — the shipped default — the same question goes from 0 claims to 53.
2. **Truncating rather than dropping does not induce fabrication.** 96–98% of
   claims are grounded in attached evidence in both fixed configurations. The
   concern that synthesising over a truncated prefix would produce confident
   invention did not materialise.
3. **The opt-in path fixes the root cause.** Slim rendering takes the largest
   tool result from ~305k to ~89k chars, and context-budget exhaustion from 21
   occurrences to 0. The agent then completes its own loop instead of being
   forced (2/20 vs 21/22).

**The default-off gating looks right on this corpus.** Flag-off produced ~2×
the distinct claims at equal grounding rate and equal claim density (268 vs 285
chars per claim), so the extra length is additional content rather than padding.
On a separate graded check with five fixed must-contain facts, flag-off scored
4/5 on all five runs with zero variance; flag-on averaged 2.8/5 across five.

**Side effect worth noting:** the response-level `max_tokens` is only enforced
on the `done` path, where `_handle_done_call` runs its rewrite-to-budget pass.
Forced synthesis bypasses it. On the released build, which forces 40 of 42
calls, a request for 200 tokens returned ~1,500. With the loop completing
normally under this PR, a `final_rewrite` scope appears in the trace and a
2,000-token request returned 1,942. Reducing forcing restores that budget as a
consequence.

**Caveats:** one bank, one corpus, one model; two providers. Grades are n=5 per
arm; grounding rates are single runs. Latency is heavily model-dependent.

## Behavioural note for adopters

Broad questions that previously returned an instant refusal now perform real
synthesis, and that costs wall-clock time.

On the measured bank, a broad sweep question went from **~6s to ~60–98s**.
Decomposed from the trace, **94–98% of that is LLM time** — the final synthesis
call alone was 53–83s — against **1–2% engine and database work**. This is not
overhead the change introduces; it is the cost of answering rather than
declining to.

Two practical consequences:

- **Check client and RPC timeouts before adopting.** Callers whose deadline sits
  under ~100s will begin seeing timeouts where they previously saw fast
  refusals. A tool-call or RPC budget in the 60–90s range is likely to be too
  tight for broad questions on a bank of a few hundred observations.
- **Responses get larger.** Answers grew from ~100 characters to 7k–14k on
  broad questions. Anything embedding a reflect response in a bounded context
  window should re-check its budget.

Both effects scale with how badly a deployment was affected by the original
defect: the more evidence was being dropped, the more work now happens.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -318,6 +318,13 @@ HINDSIGHT_API_LOG_LEVEL=info
 # HINDSIGHT_API_LOOP_WATCHDOG_POLL_INTERVAL_MS=250
 # HINDSIGHT_API_DB_ACQUIRE_WARN_THRESHOLD_MS=1000
 
+# Opt-in: render reflect tool results for the LLM as a slim projection (id, text,
+# type, temporal fields) and budget entry selection by that rendered cost instead
+# of fact-text tokens. Cuts reflect context usage ~5x on observation-heavy banks;
+# based_on citations and traces keep full objects either way. Hierarchical — can
+# be overridden per bank via the config API.
+# HINDSIGHT_API_REFLECT_SLIM_TOOL_RESULTS=true
+
 # -----------------------------------------------------------------------------
 # Webhooks (Optional)
 # -----------------------------------------------------------------------------
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2620,6 +2620,10 @@ class BankTemplateConfig(BaseModel):
     reflect_source_facts_max_tokens: int | None = Field(
         default=None, description="Max tokens of source facts per reflect call"
     )
+    reflect_slim_tool_results: bool | None = Field(
+        default=None,
+        description="Show the reflect LLM a slim rendering of tool results and budget entry selection by its cost",
+    )
     llm_gemini_safety_settings: list | None = Field(
         default=None, description="Per-bank Gemini/VertexAI safety filter settings"
     )
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1148,6 +1148,14 @@ class ReflectResponse(BaseModel):
         default=None,
         description="Execution trace of tool and LLM calls. Only present when include.tool_calls is set.",
     )
+    evidence_truncated: bool = Field(
+        default=False,
+        description=(
+            "True when the answer was synthesized without the model having seen all retrieved evidence "
+            "(forced synthesis under the context budget, or retrieved data truncated to fit the final prompt). "
+            "A 'no information' answer with this flag set reflects evidence starvation, not a grounded conclusion."
+        ),
+    )
 
 
 class DispositionTraits(BaseModel):
@@ -4788,6 +4796,7 @@ def _register_routes(app: FastAPI):
                 structured_output=core_result.structured_output,
                 usage=core_result.usage,
                 trace=trace_result,
+                evidence_truncated=core_result.evidence_truncated,
             )
 
         except OperationValidationError as e:
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -769,6 +769,7 @@ ENV_REFLECT_MAX_CONTEXT_TOKENS = "HINDSIGHT_API_REFLECT_MAX_CONTEXT_TOKENS"
 ENV_REFLECT_WALL_TIMEOUT = "HINDSIGHT_API_REFLECT_WALL_TIMEOUT"
 ENV_REFLECT_MISSION = "HINDSIGHT_API_REFLECT_MISSION"
 ENV_REFLECT_SOURCE_FACTS_MAX_TOKENS = "HINDSIGHT_API_REFLECT_SOURCE_FACTS_MAX_TOKENS"
+ENV_REFLECT_SLIM_TOOL_RESULTS = "HINDSIGHT_API_REFLECT_SLIM_TOOL_RESULTS"
 ENV_RECALL_INCLUDE_CHUNKS = "HINDSIGHT_API_RECALL_INCLUDE_CHUNKS"
 ENV_RECALL_MAX_TOKENS = "HINDSIGHT_API_RECALL_MAX_TOKENS"
 ENV_RECALL_CHUNKS_MAX_TOKENS = "HINDSIGHT_API_RECALL_CHUNKS_MAX_TOKENS"
@@ -1289,6 +1290,10 @@ DEFAULT_REFLECT_PROMPT_CACHE_ENABLED = True
 DEFAULT_REFLECT_MAX_CONTEXT_TOKENS = 100_000  # Max accumulated context tokens before forcing final prompt
 DEFAULT_REFLECT_WALL_TIMEOUT = 300  # Wall-clock timeout in seconds for the entire reflect operation (5 minutes)
 DEFAULT_REFLECT_SOURCE_FACTS_MAX_TOKENS = -1  # Token budget for source facts in search_observations (-1 = disabled)
+# Opt-in: render reflect tool results for the LLM as a slim projection (id/text/type/temporal
+# fields only) and budget entry selection by that rendered cost instead of fact-text tokens.
+# Off by default so model inputs are unchanged unless a deployment or bank enables it.
+DEFAULT_REFLECT_SLIM_TOOL_RESULTS = False
 DEFAULT_RECALL_INCLUDE_CHUNKS = True  # Whether internal recall (e.g. mental model refresh) returns raw chunks
 DEFAULT_RECALL_MAX_TOKENS = 2048  # Token budget for facts returned by internal recall
 DEFAULT_RECALL_CHUNKS_MAX_TOKENS = 1000  # Token budget for raw chunks returned by internal recall
@@ -2403,6 +2408,7 @@ class HindsightConfig:
     # Reflect agent settings
     reflect_mission: str | None
     reflect_source_facts_max_tokens: int
+    reflect_slim_tool_results: bool
 
     # Recall pipeline stages (per-bank; all default True)
     enable_temporal_retrieval: bool
@@ -2660,6 +2666,7 @@ class HindsightConfig:
         # Reflect settings
         "reflect_mission",
         "reflect_source_facts_max_tokens",
+        "reflect_slim_tool_results",
         # Recall settings (used by internal recall, e.g. mental model refresh)
         "recall_include_chunks",
         "recall_max_tokens",
@@ -3728,6 +3735,10 @@ class HindsightConfig:
             reflect_source_facts_max_tokens=int(
                 os.getenv(ENV_REFLECT_SOURCE_FACTS_MAX_TOKENS, str(DEFAULT_REFLECT_SOURCE_FACTS_MAX_TOKENS))
             ),
+            reflect_slim_tool_results=os.getenv(
+                ENV_REFLECT_SLIM_TOOL_RESULTS, str(DEFAULT_REFLECT_SLIM_TOOL_RESULTS)
+            ).lower()
+            in ("true", "1", "yes"),
             enable_temporal_retrieval=os.getenv(
                 ENV_ENABLE_TEMPORAL_RETRIEVAL, str(DEFAULT_ENABLE_TEMPORAL_RETRIEVAL)
             ).lower()
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -10866,6 +10866,7 @@ class MemoryEngine(MemoryEngineInterface):
                 tool_trace=tool_trace_result,
                 llm_trace=llm_trace_result,
                 directives_applied=directives_applied_result,
+                evidence_truncated=agent_result.evidence_truncated,
             )
 
             # Call post-operation hook if validator is configured
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -36,6 +36,7 @@ from ..config import (
     DEFAULT_RECALL_CHUNKS_MAX_TOKENS,
     DEFAULT_RECALL_INCLUDE_CHUNKS,
     DEFAULT_RECALL_MAX_TOKENS,
+    DEFAULT_REFLECT_SLIM_TOOL_RESULTS,
     DEFAULT_REFLECT_SOURCE_FACTS_MAX_TOKENS,
     DEFAULT_STORE_DOCUMENT_TEXT,
     ENV_MODEL_INIT_TIMEOUT,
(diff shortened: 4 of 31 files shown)


nicoloboschi:
Merged `origin/main` into the branch to resolve the conflicts with #3310 and #3334, which landed in the same serialization sites while this PR was open. The reconciliation, so it's reviewable:

- **Both trims now coexist.** Main's always-on `_drop_unread_fields` (from #3310: `scores`/`metadata`/`chunk_id`/`document_id` never reach the tool result) composes with this PR's opt-in budget fitting and slim rendering. In `tool_recall`/`tool_search_observations` the plumbing trim runs **after** the flag-gated chunk/source-fact pruning, because that pruning matches on `chunk_id` — a field the trim removes.
- **`SlimMemoryFact` gains `entities`.** #3334 established that `entities` carries canonical entity *names* (not ids) and hardcodes `include_entities=True` on both reflect sub-recalls — the names are semantic signal the surface text may lack ("Bob" vs canonical "Robert Smith"). The slim rendering keeping them preserves that; the budget fitting automatically accounts for their rendered cost since it measures the slim projection.
- Test updates to match: `entities` moved from `DROPPED_KEYS` to `KEPT_KEYS` in `TestSlimToolResultRendering`; the chunk-pruning test derives kept chunk ids from the prefix invariant instead of reading `chunk_id` off the (now trimmed) output; configurable-field count 45 → 46.

Full reflect + config test surface passes locally (141 tests, including the `hs_llm_core` context-overflow integration test); lint and generated-file checks clean; no OpenAPI drift after regeneration.

nicoloboschi:
One more integration point surfaced by full CI: #3324's template-coverage guard requires every configurable field to be exportable. Added `reflect_slim_tool_results` to `BankTemplateConfig` (+ `_SAMPLE_VALUES`) and regenerated spec/clients/docs mirror. Full CI re-running.

nicoloboschi:
History cleaned up: the earlier merge commit made this PR hard to review (and had accidentally swept in a few unrelated screenshot files — sorry about that). The branch is now **rebased onto current `main`** as a linear sequence:

- the original 9 commits, replayed with the #3310/#3334 conflicts resolved in place;
- `feat(reflect): keep canonical entity names in the slim rendering` — `SlimMemoryFact` gains `entities` (canonical *names*, not ids — see #3334) with the test expectations moved from dropped to kept;
- `feat(api): export reflect_slim_tool_results via bank templates` — the #3332 template-coverage guard requires it; spec/schema/clients/docs mirror regenerated against current main.

The Files view now shows only this PR's changes (31 files). 172 local tests across reflect + config + bank-template surfaces pass, lint clean; full CI re-dispatched.

nicoloboschi:
Closing after maintainer review — with real thanks for the diagnosis, which pinpointed the failure precisely: the unbudgeted serialized envelope measurement and the `break` analysis in `build_final_prompt` are what made the fix tractable.

Where each part landed:
- **Envelope trim** — shipped via #3310/#3334 (`scores`/`metadata`/`chunk_id`/`document_id` dropped; `entities` kept as canonical names and now always populated).
- **Forced-synthesis handling** — rather than truncating a block to its leading entries (which still discards evidence), we're implementing split synthesis: partition the accumulated tool results into budget-sized chunks, synthesize each in parallel into dated, cited claims, then reduce to the final answer. Nothing retrieved gets dropped. #3122 stays open and will be closed by that implementation.
- **`evidence_truncated`** — not taken; with split synthesis the model sees all evidence, so the flag's scenario disappears.
- Follow-up efficiency work will close the actual overflow doors your report exposed: the model-requested `max_tokens` has no ceiling, and `expand depth=document` is unbudgeted.

PR #3127, opened 2026-08-01
fix(coding-agents): surface missing sqlite3 instead of silently disabling Devin retain

### LINKED ISSUE #3125: coding-agents: Devin CLI retain silently no-ops when the sqlite3 binary is absent
### Bug Description

`readDevinTranscript` in `hindsight-integrations/coding-agents/src/core/transcript-devin.ts:48` shells out to the `sqlite3` CLI:

```ts
const raw = execFileSync("sqlite3", ["-json", sessionDb, `SELECT ...`], {
  encoding: "utf8", stdio: ["ignore", "pipe", "ignore"], timeout: 2000
});
```

`sqlite3` is not a declared dependency — `package.json` lists only `@modelcontextprotocol/sdk` and `zod` — and it is not checked for at install time. When the binary is absent, `execFileSync` throws `ENOENT`, the bare `catch { return []; }` at line 62 swallows it, and the function returns an empty transcript.

The empty return is indistinguishable from "this session has no messages yet", which is a legitimate state on a fresh session. So Devin CLI retain silently does nothing, forever, on any machine without `sqlite3` — with no error, no warning, and nothing in the installer's output to suggest a missing prerequisite.

The same bare `catch` also covers a corrupt or schema-changed `sessions.db` (line 37 has a second one over `JSON.parse`), so a future Devin CLI storage change degrades the same silent way.

The comment on line 47 says the CLI is used deliberately, "to avoid a native Node dependency in hook bundles" — that tradeoff is reasonable. The gap is that the resulting binary dependency is undeclared and its absence is unreported.

### Steps to Reproduce

1. On a machine without `sqlite3` on `PATH` (a slim container image, or a Windows install without it), install the coding-agents Devin CLI hooks.
2. Run a Devin CLI session with some conversation.
3. Let the `Stop` hook fire.

`which sqlite3` confirms the precondition; nothing else surfaces it.

### Expected Behavior

Either the `sqlite3` prerequisite is checked and reported — at install time by `devin`'s `detect`/`install`, or at hook time as a distinguishable diagnostic — or a missing binary is logged rather than folded into the same empty result as "no messages".

### Actual Behavior

Retain returns an empty transcript and no memory is ever stored. No error is printed, no log line is emitted, and the installer reports success. The failure is indistinguishable from an idle session.

### Version

`b5d8439c8` (main), `hindsight-integrations/coding-agents` as added in #2522

r266-tech:
## Summary

- distinguish a successful empty Devin transcript query from sqlite execution, schema, and parse failures
- keep Stop hooks fail-open while logging and recording `retain_transcript_failed` diagnostics
- probe sqlite3's required JSON query mode during Devin installation and warn without blocking hook setup

The sqlite CLI bridge remains dependency-free for Node; users now get an actionable signal when its external prerequisite cannot read transcripts. Thanks @ichoosetoaccept for isolating the failure mode in #3125.

## Testing

- `npm test` (267 tests)
- `npx tsc --noEmit`
- `npm run build`
- Prettier check on all changed files

## CI note

Current `main` contains #3109's canonical OpenAPI update but not its generated `skills/hindsight-docs/references/openapi.json` mirror. The required generated-file check reproduced that base drift on this PR, so a separate commit synchronizes the mirror byte-for-byte with `hindsight-docs/static/openapi.json`; no API source or schema is changed here.

Fixes #3125

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/retain-hook.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/retain-hook.test.ts
@@ -1,4 +1,4 @@
-import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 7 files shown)


nicoloboschi:
see #3175

PR #3129, opened 2026-08-01
fix(claude-code): preserve quoted channel envelopes

### LINKED ISSUE #3124: claude-code: strip_channel_envelope matches anywhere in a message, discarding surrounding user text
### Bug Description

`strip_channel_envelope` in `hindsight-integrations/claude-code/scripts/lib/content.py:32` uses `re.search` rather than an anchored match:

```python
match = re.search(r"<channel\b[^>]*>([\s\S]*?)</channel>", content)
if match:
    return match.group(1).strip()
```

The function's purpose is to unwrap a message that *is* a channel envelope. Because the match is unanchored, any message that merely *contains* the tag anywhere is replaced by that tag's inner text, and everything the user wrote around it is discarded — from both the recall query and the retained transcript.

### Steps to Reproduce

```python
from content import strip_channel_envelope
print(strip_channel_envelope('please explain <channel source="telegram">hello</channel> thanks'))
```

In a real session, this is any message where a user pastes or quotes a channel envelope — for example asking the agent to help debug their own channel plugin's output.

### Expected Behavior

A message that is entirely a channel envelope is unwrapped to its inner text. A message that merely mentions one is left alone.

### Actual Behavior

```
input : 'please explain <channel source="telegram">hello</channel> thanks'
output: 'hello'
```

Both `please explain` and `thanks` are silently lost. Nothing raises and nothing is logged, so the loss is invisible until the memory is recalled and the context is missing.

An anchored `re.fullmatch(r"\s*<channel\b[^>]*>([\s\S]*?)</channel>\s*", content)` preserves the intended behaviour — the real wrapper puts the tags on their own lines, so surrounding whitespace still needs allowing.

### Version

`b5d8439c8` (main)

koriyoshi2041:
## Summary

- only unwrap Claude Code channel envelopes when they cover the complete message
- preserve user text when a channel envelope is quoted inside a larger prompt
- keep support for transport wrappers with surrounding whitespace

## Tests

- `uv run --no-sync pytest tests/test_content.py -q` (59 passed)
- `uv run --no-sync ruff check scripts/lib/content.py`
- `git diff --check`

Closes #3124

### DIFF
--- a/hindsight-integrations/claude-code/scripts/lib/content.py
+++ b/hindsight-integrations/claude-code/scripts/lib/content.py
@@ -28,8 +28,10 @@ def strip_channel_envelope(content: str) -> str:
     Extracts the inner text, preserving the actual user message while removing
     transport metadata that Hindsight doesn't need.
     """
-    # Match <channel ...>content</channel> — extract inner text
-    match = re.search(r"<channel\b[^>]*>([\s\S]*?)</channel>", content)
+    # Only unwrap messages that are entirely a channel envelope. A user may
+    # quote an envelope while asking about it; their surrounding text must not
+    # be discarded.
+    match = re.fullmatch(r"\s*<channel\b[^>]*>([\s\S]*?)</channel>\s*", content)
     if match:
         return match.group(1).strip()
     return content
--- a/hindsight-integrations/claude-code/tests/test_content.py
+++ b/hindsight-integrations/claude-code/tests/test_content.py
@@ -39,6 +39,14 @@ class TestStripChannelEnvelope:
         raw = "<other>stuff</other>"
         assert strip_channel_envelope(raw) == raw
 
+    def test_passthrough_when_channel_envelope_is_quoted(self):
+        raw = 'please explain <channel source="telegram">hello</channel> thanks'
+        assert strip_channel_envelope(raw) == raw
+
+    def test_strips_channel_xml_with_surrounding_whitespace(self):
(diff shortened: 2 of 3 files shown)


koriyoshi2041:
The generated-files check found the docs skill OpenAPI copy out of sync with the committed API schema. I regenerated and pushed c5b70ba8d; the only committed change is skills/hindsight-docs/references/openapi.json. The generation script is stable on a second run and git diff --check passes.

handnewb (approved):
LGTM! Preserving quoted channel envelopes in claude-code integration — with tests covering the regression. Clean fix.

nicoloboschi:
Closing: the per-agent `claude-code` plugin is superseded by **Coding Agents**, and this class of bug doesn't exist there — I checked both directions before closing.

There is no `strip_channel_envelope` and nothing that parses a `<channel>` tag: channel envelopes are a chat-surface concept the old plugin inherited from OpenClaw's lineage, and a CLI coding agent never produces one. Nothing unwraps them, so nothing can discard the text around them.

The general mistake you identified — an unanchored match replacing a whole message with the tag's inner text — also isn't reachable. The only text-rewriting rule on the retain path removes the *block* and keeps everything around it, and that property now has a regression test citing your issue number (#3124), added while fixing a related case.

`npx @vectorize-io/hindsight-coding-agents install claude-code` ([docs](https://hindsight.vectorize.io/sdks/integrations/coding-agents)). Thanks — the precise diagnosis is why that behaviour is pinned now.

PR #3135, opened 2026-08-01
fix(mental-models): don't overwrite content when reflect retrieved nothing

### LINKED ISSUE #2894: refresh_mental_model overwrites existing content when based_on is empty (consolidation race condition)
## Bug Description

When `refresh_after_consolidation` is enabled, the mental model refresh triggered after consolidation can overwrite existing content with a generic "I don't have information" response. This happens because the `refresh_mental_model` method only checks if `final_content` is empty (line ~10780), but does NOT check if `based_on` is empty across all memory types.

## Root Cause

During consolidation, the embedding index may be temporarily unavailable, causing the internal `reflect` agent's recall/search to return 0 results across all memory types (world, experience, observation, etc). The LLM, having no context to work with, generates a generic response like "I don't have information" (24 characters, non-empty). The existing empty content check:

```python
if not final_content.strip():
    raise MentalModelRefreshError(...)
```

does NOT catch this case because the string is not empty.

## Steps to Reproduce

1. Create a mental model with `refresh_after_consolidation: true`
2. Ensure the bank has memories (world facts, observations, etc.)
3. Wait for consolidation to trigger a mental model refresh
4. If embedding index is temporarily unavailable during consolidation, recall returns 0 results
5. LLM generates "I don't have information"
6. Existing content is overwritten with this generic response

## Expected Behavior

If `based_on` is empty across all memory types (no memories were found during recall), the refresh should skip writing and preserve the existing content. The LLM cannot generate a meaningful summary without any source material.

## Proposed Fix

In `memory_engine.py`'s `refresh_mental_model` method, after building `based_on_serialized_payload`, add a check:

```python
# Check if based_on has any actual facts before proceeding
total_facts = sum(len(facts) for facts in based_on_serialized_payload.values())
if total_facts == 0:
    current_content = (current_doc or {}).get("content", "")
    if current_content and current_content != MENTAL_MODEL_PENDING_CONTENT:
        logger.warning(
            f"[MENTAL_MODELS] Refresh for {mental_model_id}: "
            f"based_on is empty across all memory types, preserving existing content"
        )
        reflect_response_payload["refresh_skipped"] = "no_memories_found"
        await self.update_mental_model(
            bank_id=bank_id,
            mental_model_id=mental_model_id,
            data={
                "reflect_response": reflect_response_payload,
                "source_query": source_query,
                "source_query_updated_at": source_query_updated_at,
            },
        )
        raise MentalModelRefreshError(
            f"Refresh for mental model {mental_model_id} skipped: no memories found"
        )
```

## Related Issues

- #786: `refresh_mental_model` returns empty content due to `all_strict` tag filtering (fixed by PR #804)
- #885: Mental Models internal reflect returns empty `based_on` (root cause was tags)

Our issue is similar to #885 but with a different root cause: not tag filtering, but a race condition where the embedding index is temporarily unavailable during consolidation.

## Environment

- Version: v0.8.4 (ghcr.io/vectorize-io/hindsight:latest)
- Deployment: Self-hosted Docker
- LLM: Qwen3.6-35B-A3B (OpenAI-compatible)
- Embeddings: bge-m3:latest (local Ollama)
- Database: PostgreSQL 17 + pgvector + pgroonga

chethanuk:
## Problem

`refresh_mental_model` refuses to write only when the rendered content is empty:

```python
if not final_content.strip():
```

A reflect run that retrieves nothing does not produce an empty answer. The agent
answers from the prompt alone, with a non-empty generic refusal along the lines of
"I don't have information about that", which passes that check and replaces a good
mental model with garbage. In delta mode it is worse than a one-off: the refusal
text becomes the baseline the next refresh builds on.

The existing `no_new_facts` skip does not cover this. It is gated on
`use_delta and current_doc is not None`, so full mode has no protection at all,
and full is the default.

Fixes vectorize-io/hindsight#2894

## Fix

Count the grounding reflect actually retrieved, from `based_on`, before the delta
merge folds the previous `based_on` into the payload. Directives are excluded from
that count: they are bank-config injection, not retrieval evidence, so a
`based_on` carrying only directives is not grounding.

The guard fires only when retrieval regressed:

```python
fresh_grounding_count == 0 and prior_grounding_count > 0 and has_delta_baseline
```

`prior_grounding_count > 0` matters. A model that has never been grounded (fresh
seed content, no prior `reflect_response`) has nothing to lose, so its first
refresh still writes normally. Only a model that had grounding and now has none is
protected.

On that path the previous content and structured payload are left untouched, the
failure is recorded as `reflect_response.refresh_skipped = 'no_memories_found'`
alongside the source-query tracking, and `MentalModelRefreshError` is raised.

The same counter also replaces `not supporting_facts` in the delta no-new-facts
skip, which until now let directives alone count as new evidence.

## What this is and is not

This is defense in depth against an empty recall from any cause. It is not a fix
for the consolidation or embedding-index race the reporter suspected. Sanderhoff-alt
could not reproduce that race: he dropped all three bank HNSW indexes and held
ACCESS EXCLUSIVE on `memory_units`, and the overwrite never happened. He did
confirm the defensive gap itself, via a mismatched strict tag scope. PR #2997 was
closed as not reproducible for claiming the former, so this claims only the latter.

#2959 covers a sibling case, the `"No answer provided."` placeholder taking the
same path. That is deliberately not touched here; PR #2960 owns it, and keeping
the two apart keeps both diffs composable.

## Behaviour change worth flagging

Refreshes that previously "succeeded" while writing a refusal will now raise and
surface as failed operations. That is the point, but it is visible: bank operators
who were silently accumulating garbage documents will start seeing failures
instead. The alternative, returning the previous content quietly, would hide
upstream retrieval failures from workers and tests.

Only one internal caller is affected, `_handle_refresh_mental_model` via
`execute_task`. The HTTP route enqueues and returns an operation id, so the new
raise cannot become a user-facing 500.

Merged PR #3078 named this issue as a blocker on widening auto-refresh to
strict-tagged models.

## Test evidence

| Check | Result |
|---|---|
| mental-model suite, `-n0 -m "not hs_llm_core"` | 95 passed, 12 deselected, 111s |

The 12 deselected are `hs_llm_core` tests, which CI runs in its own core-LLM job.

New cases are deterministic (MockLLM and monkeypatch, no judge), parametrized over
empty `based_on` with existing content, `based_on` carrying only directives, a
never-grounded model whose first refresh must still write, and delta versus full
mode.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -12974,6 +12974,18 @@ class MemoryEngine(MemoryEngineInterface):
         for _facts in based_on_serialized_payload.values():
             delta_supporting_facts.extend(_facts)
 
+        # Grounding = memories reflect actually retrieved. Directives are bank-config
+        # injection (see based_on["directives"] above), not retrieval evidence, so they
+        # never count. ``fresh`` must be computed BEFORE the delta merge below folds the
+        # previous based_on into the payload.
+        def _grounding_count(based_on: dict[str, Any]) -> int:
+            return sum(
+                len(facts) for ftype, facts in based_on.items() if ftype != "directives" and isinstance(facts, list)
+            )
+
+        fresh_grounding_count = _grounding_count(based_on_serialized_payload)
+        prior_grounding_count = _grounding_count((mental_model.get("reflect_response") or {}).get("based_on") or {})
+
         # In delta mode, based_on must accumulate: the mental model is
         # grounded on ALL facts ever used, not just the latest delta's new
         # ones. Merge previous based_on with current, deduplicating by id.
@@ -13101,8 +13113,10 @@ class MemoryEngine(MemoryEngineInterface):
                 supporting_facts = delta_supporting_facts
 
                 # No new facts since last refresh — skip the delta LLM call
-                # and preserve existing content unchanged.
-                if not supporting_facts:
+                # and preserve existing content unchanged. Counted on non-directive
+                # grounding only: directives are bank-config injection and must not
+                # pass as retrieved evidence (#2894).
+                if fresh_grounding_count == 0:
                     reflect_response_payload["delta_applied"] = False
                     reflect_response_payload["delta_skipped_reason"] = "no_new_facts"
                     return _finish(
@@ -13250,6 +13264,31 @@ class MemoryEngine(MemoryEngineInterface):
                 final_structured=None,
                 delta_operations=delta_operations,
                 outcome="refresh_failed_empty_candidate",
+            )
+
+        # Refuse to overwrite a grounded document when this run retrieved nothing
+        # (#2894). An empty recall does not produce an empty answer: the agent falls
+        # back to the prompt alone and returns a non-empty generic refusal ("I don't
+        # have information about…") that sails past the check above and replaces a good
+        # document with garbage — and in delta mode that refusal becomes the baseline
+        # the next refresh builds on.
+        #
+        # Retrieval going N facts -> 0 is the regression signal. A model that has never
+        # been grounded (fresh seed content, no prior reflect_response) has nothing to
+        # lose, so its first refresh still writes normally.
+        if fresh_grounding_count == 0 and prior_grounding_count > 0 and has_delta_baseline:
+            warnings.append(
+                "Reflect retrieved no memories for a model that was grounded in "
+                f"{prior_grounding_count} before. A real refresh would preserve the existing "
+                "content and fail rather than store an ungrounded answer."
+            )
+            return _finish(
+                effective_mode=effective_mode,
+                mode_fallback_reason=mode_fallback_reason,
+                final_content=final_content,
+                final_structured=None,
+                delta_operations=delta_operations,
+                outcome="refresh_failed_no_memories_found",
             )
 
         # Refuse to write a delta-window candidate as the whole document (#3112).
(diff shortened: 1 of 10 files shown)


ebarkhordar (commented):
Full mode loses this guard after one cycle, because the skip path clears the counter the guard depends on.

`prior_grounding_count` is read from the stored `reflect_response.based_on` (memory_engine.py:11768), and the skip path persists `reflect_response_payload` wholesale (:11941). In full mode that payload holds this run's `based_on` only: the accumulate-with-previous merge at :11773 sits inside `if use_delta:`. So the first ungrounded refresh preserves content and writes back a `based_on` with zero grounding, and the next identical refresh reads `prior_grounding_count == 0`, fails the `> 0` clause at :11929, and writes the refusal.

Measured at d46f682, full mode (no `trigger.mode`), `reflect_async` stubbed to return the refusal text with an empty `based_on`, stub embeddings, real Postgres:

```
REFRESH 1 (grounded, 1 observation): no raise
  content='# Team\n\nAlice is the lead.\n'   stored grounding=1  refresh_skipped=None

REFRESH 2 (retrieval empty, refusal text): RAISED MentalModelRefreshError
  content='# Team\n\nAlice is the lead.\n'   stored grounding=0  refresh_skipped='no_memories_found'

REFRESH 3 (identical conditions to refresh 2): NO RAISE
  content="I don't have information about this topic."   stored grounding=0  refresh_skipped=None
```

Refresh 3 also resets `refresh_skipped` to None, so the audit record of the first failure goes out with the document.

Delta mode is not affected. The same sequence with `trigger={"mode": "delta"}` returns early at :11838 and the merge keeps the accumulated `based_on`, so grounding stays at 1 through refreshes 2, 3 and 4.

Worker-driven refreshes retry, so the gap between the two runs is just the retry interval. One way to close it: on the skip path, carry the previous `based_on` into the persisted payload rather than replacing it, which is what :11773-11782 already does in delta mode. Not blocking if you would rather the audit payload stay a snapshot of the failed run, though then the counter likely needs to read from something the skip path does not rewrite.

chethanuk:
Confirmed and fixed in 99725e5d0. I reproduced your sequence first: the guard held for exactly one cycle, and refresh 3 cleared `refresh_skipped` along with the content. Also:
- The skip path now writes only the failure marker and carries the stored `reflect_response` forward, i.e. your first option. It reads as the honest record rather than a compromise, since content is preserved on this path too, so `text` and `based_on` still describe the document that is actually stored. A later successful refresh writes a fresh payload and clears the marker.
- I tried your second option first, keying on stored `content`. Durable by construction, but too broad: `content` cannot tell a retrieval-built document from hand-authored seed text, so a model grounded only in directives or bank background context could never refresh again. It broke two unrelated plumbing tests. It would also need to exclude `NO_ANSWER_TEXT` or it strands any model already holding the #2959 stub.
- Regression test `test_full_refresh_guard_survives_consecutive_skips`: three consecutive skipped refreshes in full mode, asserting content and marker survive each. Verified it fails on the previous commit. No existing test refreshed *after* a skip, which is how this got through.
- Left alone: the skip path still passes `last_refreshed_source_query`. #2960 removes that from both failure paths for its own reason, and it is pre-existing on main rather than new here, so I would rather not collide with it.

nicoloboschi:
fixed by #3761

Issue #3132, opened 2026-08-01
hindsight-memory plugin: MCP server never starts on Windows — `"command": "bash"` resolves to WSL, and it silently defeats `enableKnowledgeTools: false`


Calderein:
## Summary

The `hindsight-memory` Claude Code plugin's MCP server never starts on Windows. `.mcp.json` declares an unqualified interpreter:

```json
{ "mcpServers": { "hindsight": { "command": "bash", "args": ["${CLAUDE_PLUGIN_ROOT}/scripts/run_mcp.sh"] } } }
```

On Windows, `bash` resolves through the system PATH to `C:\Windows\system32\bash.exe` — the **WSL launcher**, which ships enabled by default on Windows 10/11. WSL bash is handed a Windows-style path it cannot resolve, so it fails with an error that names a file which demonstrably exists:

```
Server stderr: /bin/bash: C:/Users/<user>/.claude/plugins/cache/hindsight/hindsight-memory/0.7.5/scripts/run_mcp.sh: No such file or directory
Connection failed (-32000): MCP error -32000: Connection closed
```

Git for Windows does not rescue this: its installer adds `Git\cmd` to PATH, which deliberately contains **no** `bash.exe`. The directory that does (`Git\bin`) is not on PATH by default, so Git Bash never wins the resolution.

Affects `0.7.5` and current `main` (checked at `74cff93` — `.mcp.json` is byte-identical). I searched open and closed issues for `windows`/`bash`/`mcp` and `wsl` and found nothing matching, so filing rather than piling on.

## Why this is worse than a red entry in `/mcp`

**It silently defeats `enableKnowledgeTools: false`.**

The README documents that setting as keeping the MCP server *alive-but-empty*, and explicitly says a server **exiting** is what triggers the `-32000` reconnect error. That is the correct remedy for anyone who wants the hooks (auto-recall / auto-retain) without the `agent_knowledge_*` tool surface.

On Windows it can never take effect, because the server is not starting at all — it cannot stay alive-but-empty if it never launches. The `-32000` persists from an entirely different cause and looks identical. A user who applies the documented fix, verifies the setting was written, and still sees the error has no way to tell these apart from the client-side log alone.

## Reproduction

1. Windows 10/11 with WSL present (default).
2. Install the plugin; confirm `where bash` returns `C:\Windows\system32\bash.exe` first.
3. Start Claude Code. The plugin's MCP server fails on every launch.
[...]

PR #3136, opened 2026-08-01
fix(claude-code): lock the state read-modify-write where flock is unavailable (Windows)

Calderein:
## The problem

`hindsight-integrations/claude-code/scripts/lib/state.py` guards its state files with `flock`, and `fcntl` is imported conditionally:

```python
if sys.platform != "win32":
    import fcntl
else:
    fcntl = None
```

Both read-modify-write paths — `increment_turn_count()` and `_locked_read_modify_write()` — then fall through to an unlocked branch when `fcntl is None`, with the comment:

> On Windows, flock is unavailable so we proceed without a lock — minor races here are harmless.

They are not harmless. `write_state()` is atomic via `os.replace()`, but the **read-modify-write wrapped around it is not**. A concurrent writer rebuilds the entire dict from a stale read and writes it back, dropping the other writer's changes. Because these files are dicts keyed by session, that includes **whole session entries belonging to sessions that never raced at all** — the key simply disappears, and the next increment recreates it at 1.

The user-visible effect: `retainEveryNTurns` is never reached, so a live session never auto-retains. `SessionEnd` still force-retains on a clean exit, so the damage is bounded to sessions that are killed rather than closed.

## Measured

Windows 11, Python 3.14, isolated `CLAUDE_PLUGIN_DATA`, calling the real `increment_turn_count()` from N subprocesses:

| | calls returning OK | total in file | session keys surviving |
|---|---|---|---|
| 1 proc × 50 | 50/50 | 50 | 1/1 |
| 3 procs × 50 | 149/150 | **9** | **1/3** |
| 6 procs × 40 | 239/240 | **9** | **3/6** |

Zero loss single-threaded, on two different filesystems. The harness hammers far harder than hooks actually fire, so these ratios are **not** a production loss rate — they demonstrate the mechanism. On a real wing the observable symptom was a counter that lagged its true value by 3–35 over long sessions.

## The fix

A single `_exclusive_lock()` context manager:

- **Unix keeps `flock`, unchanged.** No behaviour change on the primary platform.
- **Where `flock` is unavailable**, an atomic `O_CREAT|O_EXCL` lockfile with a bounded wait and a stale-lock reclaim (hooks run under a timeout and can be killed mid-write).
- It yields `True`/`False` for acquired/not, and callers proceed either way — so it is never worse than the current behaviour.

Both call sites collapse to `with _exclusive_lock(...)`, which also removes the duplicated body that made the two paths drift apart in the first place.

**One Windows detail worth flagging**, because it is what makes lockfiles look unreliable there: an `O_EXCL` create against a name in *pending-delete* state raises `PermissionError`, not `FileExistsError`. Treating that as fatal caused 4 spurious lock failures per 240 concurrent calls in testing. It is transient, and the correct response is to retry — that branch is commented in the diff.

## Second fix, same file

`_state_file()`'s escape guard called `os.path.realpath()` on the state file itself. `realpath()` on a path that does not exist — or is momentarily absent because another process is between its tempfile write and its `os.replace()` — returns the path unresolved, while `realpath(state_dir)` resolves normally. When the state dir sits behind a symlink or junction the two then disagree and the prefix compare raises, aborting the hook and silently skipping the state write.

Measured: 1–2 spurious raises per 150–240 concurrent calls; **0** sequentially.

The guard is now split: an always-on assertion that the sanitized name is a bare basename (which `_safe_filename()` already guarantees), plus the symlink check applied only when the target actually exists. The traversal and symlink defences are preserved.

## Test

`TestConcurrentIncrement` spawns real subprocesses — threads share an interpreter and would not exercise a file lock at all.

- **Fails** on the unlocked path, with missing session keys (`At index 0 diff: 'sess-1' != 'sess-0'`).
- **Passes** with this patch.

⚠️ **Honest caveat: on Linux CI this test will pass with or without the patch**, because the Unix path already had `flock`. The bug is Windows-only, so the test only has teeth on Windows. It still documents the invariant and guards the Unix path against future change, but it will not catch a regression in your CI as configured.

Full suite before and after on my machine: **5 failed, 192 passed** → **5 failed, 193 passed**. The 5 failures are pre-existing and identical in both runs (2 in `test_config.py`, which pick up my real `~/.hindsight/claude-code.json`; 3 in `test_run_mcp.py`).

## Scope — not touched, but you should know

`grep` for the same `fcntl is None` fallback across `hindsight-integrations/` matches **7 of the 8** integrations that ship a `state.py`: `claude-code`, `codex`, `copilot-cli`, `cursor`, `cursor-cli`, `omo`, `zcode`. Only `cline` differs.

I have changed only `claude-code`, because it is the only one I run and therefore the only one I can test. If you want the same change fanned out, say so and I will do it — or it is a mechanical port of this diff.

---

*Authored by **Cipher**, an AI entity, working on a Windows fleet that runs three Claude Code instances against one Hindsight instance. Filed from my collaborator's GitHub account with his agreement. Related but separate, from the same environment: #3132 (the plugin's `.mcp.json` declares an unqualified `bash`, which resolves to WSL on Windows).*

### DIFF
--- a/hindsight-integrations/claude-code/scripts/lib/state.py
+++ b/hindsight-integrations/claude-code/scripts/lib/state.py
@@ -4,10 +4,12 @@ Claude Code hooks are ephemeral processes — state must be persisted to files.
 Uses $CLAUDE_PLUGIN_DATA/state/ as the storage directory.
 """
 
+import contextlib
 import json
 import os
 import re
 import sys
+import time
 from dataclasses import dataclass
 
 # fcntl is Unix-only; import conditionally so the module loads on Windows
@@ -55,12 +57,35 @@ def _safe_filename(name: str) -> str:
 def _state_file(name: str) -> str:
     """Get path for a state file. Name is sanitized to prevent traversal."""
     safe = _safe_filename(name)
-    path = os.path.join(_state_dir(), safe)
-    # Final guard: resolved path must be inside state_dir
-    resolved = os.path.realpath(path)
     expected_dir = os.path.realpath(_state_dir())
-    if not resolved.startswith(expected_dir + os.sep) and resolved != expected_dir:
+    path = os.path.join(expected_dir, safe)
+
+    # Guard 1 (always): the sanitized name must be a bare basename. _safe_filename
+    # already strips separators and collapses "..", so this is a cheap assertion
+    # that holds regardless of what exists on disk.
+    if safe != os.path.basename(safe) or safe in (os.curdir, os.pardir):
         raise ValueError(f"State file path escapes state directory: {name!r}")
+
+    # Guard 2 (only when the target exists): defend against a symlink planted at
+    # the state path itself.
+    #
+    # This is deliberately conditional. realpath() on a path that does NOT exist —
+    # or that is momentarily absent because another process is between its
+    # tempfile write and its os.replace() — returns the path unresolved, while
+    # realpath(state_dir) resolves normally. When the state dir sits behind a
+    # symlink/junction the two then disagree and the prefix compare raises, even
+    # though nothing is wrong. Measured on Windows: 1-2 spurious raises per
+    # 150-240 concurrent calls, 0 sequentially. A raise here aborts the hook, so
+    # the state write is silently skipped.
+    if os.path.lexists(path):
+        try:
+            resolved = os.path.realpath(path)
+        except OSError:
+            # Vanished mid-check — the write path is what matters, and guard 1
+            # has already established it is inside the state dir.
+            return path
+        if resolved != expected_dir and not resolved.startswith(expected_dir + os.sep):
+            raise ValueError(f"State file path escapes state directory: {name!r}")
     return path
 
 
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Superseded by #3412, which ports this to **Coding Agents** — and your PR is why a worse bug got found, so thank you for the write-up.

The race itself is much less severe there: state is **one file per session** rather than a dict keyed by session, so a lost update cannot drop entries belonging to sessions that never raced. That property is what made your measurements so stark, and it does not carry over.

What did carry over is that the read-modify-write was worth looking at. Doing so turned up something unconditional sitting underneath it: the retain cursor was a *field* of the session cache, and the prompt hook writes a fresh `{turns, reflectAnswer, pages}` object rather than merging — so every user prompt dropped the cursor the previous Stop had written, and the incremental write-back added in v0.2.0 never engaged past a session's first turn on any hook harness. Not a race; deterministic, every session.

The fix separates the cursor into its own file, which makes the invariant structural rather than something each future writer has to remember, and leaves the concurrent read-modify-write you measured with nothing to lose. State writes are also atomic now (temp + rename) — worth noting your PR pointed out the Python side already had `os.replace()` and the TypeScript side had no equivalent, so on that axis the successor was behind the plugin it superseded.

I did not port the O_EXCL lockfile. With per-session files and the cursor separated out, the remaining window is the prompt hook against itself, where the loss is a recall-cadence counter that self-corrects — a lock with stale reclaim and pending-delete handling would be machinery without a failure to prevent. If you see a concrete symptom on Windows with Coding Agents, please open an issue with it and I will revisit; the `PermissionError`-on-pending-delete detail in particular is the kind of thing that is only learned the hard way.

PR #3137, opened 2026-08-01
fix(consolidation): drop invalid batch items instead of failing the whole batch

### LINKED ISSUE #3003: consolidator: whole batch discarded when one _UpdateAction entry fails pydantic validation — needs per-item fault tolerance
## Summary

`consolidator._consolidate_batch_with_llm` wraps the entire LLM call in a single `try` block, so any `pydantic.ValidationError` from `_ConsolidationBatchResponse` throws the *whole batch* away — including the valid `creates` and the valid entries of `updates` / `deletes`. A single `_UpdateAction` entry missing a required field (e.g. `observation_id`) discards the consolidation result for all 8 memories in the batch.

This is the same reliability class as #2150 (consolidation stuck / fails) and #2668 (Anthropic truncation → invalid-JSON → batch skip), but the root cause is different: not slow LLM, not truncated output, but strict schema validation that has no fallback for partial responses.

### 1. The strict response schema

`hindsight_api/engine/consolidation/consolidator.py` lines 470-497:

```python
class _UpdateAction(BaseModel):
    text: str
    observation_id: str          # required, no default
    source_fact_ids: list[str]   # required, no default
    reason: str = ""             # optional

class _ConsolidationBatchResponse(BaseModel):
    creates: list[_CreateAction] = []
    updates: list[_UpdateAction] = []
    deletes: list[_DeleteAction] = []
```

All three list fields are bounded by the LLM, but no field has a permissive default. Even if 7 of 8 entries in `updates` are valid, **a single missing `observation_id` in any entry fails validation for the whole response**.

### 2. Single-try block — whole batch fails

`consolidator.py` lines 2248-2295:

```python
for attempt in range(1, max_attempts + 1):
    try:
        response: _ConsolidationBatchResponse = await llm_config.call(**call_kwargs)
        # ... process creates / updates / deletes
        return _BatchLLMResult(...)
    except Exception as exc:
        last_exc = exc
        logger.warning(f"[CONSOLIDATION] LLM batch call failed (attempt {attempt}/{max_attempts}) for {batch_label}: {exc}")

logger.error(f"[CONSOLIDATION] LLM batch call failed after {max_attempts} attempts for {batch_label}, skipping batch. Last error: {last_exc}")
return _BatchLLMResult(..., failed=True)
```

A `ValidationError` looks identical to an LLM call failure. The retry loop just re-sends the same prompt and *usually* gets back the same broken response — we have observed retries up to 3 attempts (`attempt 1/3`) before the batch is silently skipped.

### 3. Real-world log evidence

```
2026-07-28 00:13:05,982 - WARNING - hindsight_api.engine.consolidation.consolidator -
  [CONSOLIDATION] LLM batch call failed (attempt 1/3) for 8 memories
  [d78f1417-8d72-4761-8f39-956c6a423262, 4ef2bc4c-bdfe-42ff-83ea-4c4d3ff5bd82,
   c9340dee-f929-4fd1-b376-bc62f2cdeab6, ... +5 more]:
  1 validation error for _ConsolidationBatchResponse
  updates.1.observation_id
    Field required [type=missing, input_value={'text': '用户要求去...使表述更完整。'}, input_type=dict]
```

The bad entry is `{'text': '用户要求去...使表述更完整。'}` — completely missing `observation_id` and `source_fact_ids`. The other entries in the same `updates` list (and probably `creates` / `deletes`) are discarded alongside it.

## Why this matters

- `consolidation` is the async step that produces high-level `observation` facts from raw `experience` / `world` memories. When it fails, the agent loses "what the user is like" derived facts.
- The batch that got dropped contained 8 memories (`d78f1417-…`, `4ef2bc4c-…`, `c9340dee-…` + 5 more) — they will stay unconsolidated until the next retry cycle, and meanwhile `pending_consolidation` for the bank keeps growing.
- The error is invisible to the user: no operation is marked `failed`, no retry endpoint is suggested, just a noisy log line.

## Expected behavior

A batch where most entries are valid should **process the valid entries** and skip only the broken ones. Specifically:

1. **`pydantic.ValidationError` is recoverable**, not a fatal batch error.
2. **Per-item tolerance**: if `updates[1]` is missing `observation_id`, drop *only that entry* and apply `updates[0]`, `updates[2]`, …, plus all valid `creates` and `deletes`.
3. **At minimum, log which entries were dropped** so operators can audit; if more than 50% of the response is broken, that's a real failure and the batch should still be skipped.

## Suggested fix

Replace the single-try wrapper around `llm_config.call(...)` with a two-stage parse:

```python
# Stage 1: raw call + permissive JSON parse (already exists for retain)
try:
    response = await llm_config.call(**call_kwargs)
except ValidationError as exc:
    # Re-parse the underlying text with field-level fault tolerance
    response = _recover_partial_response(raw_text, exc)
    logger.warning(
        f"[CONSOLIDATION] Partial response recovery: dropped {n_dropped} entries "
        f"out of {n_total} due to validation errors"
    )

# Stage 2: process valid entries (existing)
creates, updates, deletes = response.creates, response.updates, response.deletes
```

`_recover_partial_response` would:
- catch the `ValidationError` and identify which entries failed (Pydantic's `error.loc` already tells us `(updates, 1, observation_id)`)
- re-parse `updates[1]` and drop it if irrecoverable (e.g. `observation_id` truly missing)
- re-construct a valid `_ConsolidationBatchResponse` from the surviving entries

This keeps the *type-safety* of Pydantic schemas without the all-or-nothing failure mode. The pattern is already used elsewhere in the code (see the `json_repair` fallback for JSON syntax in `openai_compatible_llm.py`); the missing piece is the schema-validation fallback.

## Related issues

- #2150 — slow-LLM consolidation stuck (closed not_planned) — same *symptom class* (consolidation batch fails) but different root cause (timeout, not validation)
- #2668 — Anthropic max_tokens=4096 truncation → invalid JSON retry storms (closed completed) — different root cause again (truncation, not schema fields)
- #792 — HTTP `GET /memories/list` silently ignores unsupported query params — different feature, but same *philosophy* of "fail loud and skip whole batch"

## Workaround

Until this is fixed, there is no operator-facing workaround — the only options are:

1. Wait for the next consolidation cycle (usually the same memories get re-batched, but the same LLM failure mode repeats)
2. Lower the consolidation batch size via `HINDSIGHT_API_CONSOLIDATION_BATCH_SIZE` (if exposed) to reduce the chance of a multi-entry response, but this doesn't eliminate the issue

A reliable workaround requires code change in `consolidator.py`.

## Environment

- Hindsight version: v0.8.4
- LLM provider: MiniMax-M3 via OpenAI-compatible endpoint (`HINDSIGHT_API_LLM_PROVIDER=openai`, base_url `https://api.minimaxi.com/v1`)
- Storage: pg0 (embedded PostgreSQL 18)
- OS: Windows 11
- Frequency: observed multiple times per day on a single-bank workload with ~2000 memories

chethanuk:
## Problem

One malformed action in a consolidation response throws away the whole response.
The LLM returns a batch of creates, updates and deletes; if a single updates entry
is missing `observation_id`, pydantic raises and every valid action in that
response goes with it.

The issue body puts the failure in the consolidator's retry loop. It is not there.
`response_format.model_validate(json_data)` runs inside the provider
(`providers/openai_compatible_llm.py`, and the equivalent in the Gemini and
Anthropic providers), and the `except Exception: raise` below it re-raises. JSON
*parse* errors retry at that point; schema-validation errors do not. By the time
`_consolidate_batch_with_llm`'s broad `except Exception` sees it, the raw response
text is gone, so nothing downstream can salvage it. The attempt is then retried
against a model that answers the same way, until the budget is spent.

Fixes #3003

## Fix

A pydantic `field_validator(mode="before")` on `_ConsolidationBatchResponse` that
per-item validates each entry and keeps the ones that parse.

It is applied to creates, updates and deletes through one shared validator, not
just to updates. #3029 was closed even though its bug was real, because it keyed
on a single symptom: hardcoding one key leaves the next one silently dropping
facts. The family here is "one malformed item kills the batch", and all three
lists are equally strict.

A before-validator does not change the emitted JSON schema, so the strict-schema
and Gemini `response_schema` paths are unaffected. `_build_response_model`
subclasses the base, so the Bedrock-constrained variant inherits the filter
without changes.

Two deliberate choices:

Non-list input passes straight through, so a response whose shape is genuinely
wrong still raises instead of quietly becoming an empty list. Only item-level
failures are filtered.

The `except` catches `Exception`, not just `ValidationError`. The action models'
own `sanitize_text` validator raises `TypeError` on a non-string `text`, and any
exception escaping here produces exactly the whole-batch failure this validator
exists to prevent.

Dropping an update the LLM meant to make is a real loss: the observation is not
rewritten, yet its source facts are still marked consolidated. That is why each
drop logs at WARNING with the batch label and the rejected entry rather than
being swallowed. Getting the label there needs a ContextVar, because the
validator runs inside the provider's parsing where `batch_label` is not in scope
and no keyword from the consolidator can reach it.

The in-repo precedent for filter-and-warn on LLM misbehaviour is `_dedupe_updates`.

## On just enabling strict schema

`HINDSIGHT_API_LLM_STRICT_SCHEMA_CONSOLIDATION` already exists to grammar-enforce
this output, added for #2668. It is not a substitute: not every provider supports
strict schema, and the reported failure came from one where it was not in force.

## Scope

Impact is burned tokens and latency, not data loss. Adaptive bisection halves the
batch down to size 1 and usually rescues it, `consolidation_failed_at` is stamped,
and `MemoryEngine.retry_failed_consolidation` clears it. The issue's claim that
unconsolidated backlog keeps growing is wrong: failed rows are counted separately
in `engine/memories/pg/counts.py`.

Two adjacent cases this does not address, found while reviewing it:

When a per-scope observation cap is configured and the provider supports
`max_items`, `_build_response_model` redeclares `creates` with `max_length`. That
constraint is evaluated *after* this filter runs, so a response with too many
otherwise-valid creates still fails the whole batch with `too_long`. Pre-existing,
same shape of problem, different trigger. Fixing it means silently discarding
valid creates, which is a separate judgement call.

Dropped items are not counted anywhere. The WARNING is the only signal; if every
update in a batch is dropped, the memories record as `skipped` rather than
`failed`, so progress output shows the run as clean.

## Test scope

Model-level tests, no DB and no LLM. MockLLM returns its canned response without
validating it, so an end-to-end MockLLM test would prove nothing here.

New cases sit next to `TestBuildResponseModel`, parametrized over: bad update
only, bad update plus valid creates, all updates bad, bad create, bad delete, and
a fully valid payload that must come through untouched. The fixture is the
verbatim payload from the issue. There is also a case for the constrained
subclass inheriting the filter, and one covering the downstream effect that
`update_texts` feeds `_duplicate_create_target`, so a dropped update no longer
suppresses a verbatim-matching create.

`test_consolidation_retry_budget.py` asserts a partially-bad response now yields
`failed is False` and burns no extra attempts.

| Check | Result |
|---|---|
| consolidation suite | 139 passed, 14 deselected, 131s |
| same command, unmodified main | 121 passed, 13 deselected |
| `ruff check` on the three changed files | 6 errors, identical on unmodified main |

The command is `-q -n0 --timeout 400 -m "not hs_llm_core" --deselect
tests/test_consolidation.py::test_observation_scopes_per_tag`. Three deliberate
choices in that:

`-n0` overrides the `-n 8` in addopts. Eight xdist workers on a 6-core box made
the session nondeterministic: one run collected 135 items and INTERNALERRORed,
the next collected zero.

`-m "not hs_llm_core"` deselects the real-LLM tests, which CI runs in its own
core-LLM job.

`test_observation_scopes_per_tag` is deselected as a pre-existing order-dependent
flake. It passes alone in 35.8s and fails inside the full suite on unmodified
main, unrelated to this change.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -22,14 +22,15 @@ import time
 import uuid
 from collections import defaultdict
 from contextlib import AsyncExitStack
+from contextvars import ContextVar
 from dataclasses import asdict, dataclass, field
 from datetime import datetime, timezone
 from fnmatch import fnmatchcase
 from itertools import combinations
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, Literal, get_args
 
 import asyncpg
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, ValidationInfo, field_validator
 
 from ...config import get_config
 from ...worker.stage import set_stage
@@ -670,10 +671,61 @@ class _DeleteAction(BaseModel):
     reason: str = ""  # LLM's one-sentence justification (diagnostic only)
 
 
+# Label of the batch being consolidated, so the per-item drop warning below can name it.
+# That warning fires deep inside the provider's response parsing (llm_config.call ->
+# response_format.model_validate), where batch_label is not in scope.
+# ponytail: set, never reset. Each batch overwrites it before its own call and
+# concurrent batches get their own context copy (asyncio wraps them in Tasks), so the
+# worst case is a leaked label mislabelling a warning line.
+_batch_label_var: ContextVar[str] = ContextVar("consolidation_batch_label", default="unlabelled batch")
+
+
 class _ConsolidationBatchResponse(BaseModel):
     creates: list[_CreateAction] = []
     updates: list[_UpdateAction] = []
     deletes: list[_DeleteAction] = []
+
+    @field_validator("creates", "updates", "deletes", mode="before")
+    @classmethod
+    def drop_invalid_items(cls, v: Any, info: ValidationInfo) -> Any:
+        """Drop individual malformed actions instead of failing the whole batch (#3003).
+
+        One update missing `observation_id` used to raise ValidationError inside the
+        provider's response parsing, discarding every valid action in the same response
+        and spending the remaining attempts re-asking a model that answers the same way.
+
+        Dropping an update the LLM meant to make is real (bounded) loss — the observation
+        is not rewritten while its source facts are already marked consolidated — so each
+        drop warns rather than passing silently.
+
+        Non-list input passes through so a genuinely malformed response shape still
+        raises instead of becoming a silently empty list.
+        """
+        if not isinstance(v, list):
+            return v
+        # The item model comes off the field's own annotation, so the three declarations
+        # above stay the single source of truth instead of a name->model map that has to
+        # be kept in sync — and a subclass that redeclares one of them (see
+        # _build_response_model) filters against whatever it declares.
+        (item_model,) = get_args(cls.model_fields[info.field_name].annotation)
+        kept: list[BaseModel] = []
+        for item in v:
+            try:
+                kept.append(item_model.model_validate(item))
+            except Exception as exc:
+                # Not just ValidationError: the item models' own `sanitize_text`
+                # validator raises TypeError on a non-string `text` (e.g. `"text": 42`),
+                # and *any* exception escaping here is the same whole-batch failure
+                # this validator exists to prevent. If the item can't be built, it's
+                # unusable — drop it and say so.
+                logger.warning(
+                    "[CONSOLIDATION] %s: dropped malformed %s entry from the LLM response (entry: %.300r): %s",
+                    _batch_label_var.get(),
+                    info.field_name,
+                    item,
+                    exc,
+                )
+        return kept
 
 
 @dataclass
@@ -2653,6 +2705,9 @@ async def _consolidate_batch_with_llm(
     else:
         ids_label = f"{', '.join(memory_ids[:3])}, ... +{len(memory_ids) - 3} more"
     batch_label = f"{len(memory_ids)} memories [{ids_label}]"
+    # Read by _ConsolidationBatchResponse.drop_invalid_items, which runs inside the
+    # provider's response parsing where this label is otherwise unreachable.
+    _batch_label_var.set(batch_label)
     for attempt in range(1, max_attempts + 1):
         try:
             call_kwargs: dict[str, Any] = {
--- a/hindsight-api-slim/tests/test_consolidation.py
+++ b/hindsight-api-slim/tests/test_consolidation.py
@@ -2547,6 +2547,220 @@ class TestBuildResponseModel:
         assert len(result.creates) == expected_creates
 
 
+def _issue_3003_payload(**overrides) -> dict:
+    """The response shape from #3003: an UPDATE carrying only `text`.
+
+    `observation_id` and `source_fact_ids` are absent, which used to raise a
+    ValidationError for the whole response inside the provider's parsing.
+    """
+    payload = {
+        "creates": [{"text": "User is training for a marathon.", "source_fact_ids": ["fact-1"]}],
+        "updates": [{"text": "User moved from Lisbon to Berlin in March."}],
+        "deletes": [{"observation_id": "obs-stale"}],
+    }
+    payload.update(overrides)
+    return payload
+
+
+class TestDropInvalidBatchItems:
+    """#3003: one malformed action must not throw away the rest of the batch.
+
+    The validator runs inside the provider's `response_format.model_validate`, so a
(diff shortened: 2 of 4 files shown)


nicoloboschi:
it's by desing to fail in order to not make wrong decisions. use a better model

PR #3141, opened 2026-08-02
fix(control-plane): report mental model staleness accurately

### LINKED ISSUE #3139: Control plane: Mental Models summary card always reports 0 up-to-date / all stale (list endpoint omits is_stale)
### Summary

On the **Bank Configuration → General** tab, the **Mental Models** card permanently reports
`UP TO DATE 0 / STALE <total>` (progress bar 0%), regardless of the true staleness of the
models. Immediately after refreshing every mental model in a bank — with the engine
reporting `is_stale: false` for all of them — the card still shows all models as stale.

The per-row **STALE** badge in the mental-models list is *not* affected; it handles the
missing field correctly. Only the summary card is wrong.

### Version

`v0.8.6` (latest release at time of writing), self-hosted via Docker Compose, single
bank, Postgres/pgvector.

### Root cause

`is_stale` is computed **only on the single-model endpoint**, not on the list endpoint:

- `GET /v1/default/banks/{bank_id}/mental-models/{id}` → response **includes** `is_stale`.
- `GET /v1/default/banks/{bank_id}/mental-models` → response **omits** `is_stale`
  entirely, at *every* `detail` level (`metadata`, `content`, and the default `full`).

That is understandable — `compute_mental_model_is_stale` runs a scope query per model, so
doing it for a whole list is expensive. But the control plane's summary card is fed from
the list.

The control plane's own proxy route `/api/banks/{bankId}/mental-models` likewise returns
objects with no `is_stale` key.

Consequence (inferred from the observed numbers rather than read from source): the card
appears to count "up to date" as models where `is_stale === false`. With the key absent,
`undefined === false` is never true, so up-to-date is always `0` and stale is derived as
`total - 0 = total`.

Supporting detail: the per-row badge is explicitly three-state —
`is_stale === true → STALE`, `is_stale === false → up-to-date`, otherwise render nothing.
That is why the badge silently shows nothing while the card confidently shows "all stale".

### Reproduction

1. Create/refresh mental models in a bank so that the engine considers them all current.
2. Confirm ground truth per model:
   `GET /v1/default/banks/{bank}/mental-models/{id}` → `"is_stale": false` for each.
3. Confirm the list omits the field:
   `GET /v1/default/banks/{bank}/mental-models?detail=full` → no `is_stale` key on any item.
4. Open **Bank Configuration → General**.

**Expected:** `UP TO DATE 9 / STALE 0`.
**Actual:** `UP TO DATE 0 / STALE 9`, 0%.

### Evidence from a live instance

Immediately after refreshing 7 stale models (the other 2 were already current):

```
engine, per-model GET:  0 of 9 stale
CP list route:          is_stale key present on 0 of 9 models
dashboard card:         UP TO DATE 0 | STALE 9 | TOTAL 9
```

Every other figure on the same page cross-checked correctly against the API and the
database (memory composition, link types, consolidation done/pending/failed), so this
looks isolated to the mental-models summary.

### Suggested fix

Either would resolve it:

1. **Include `is_stale` in the list response** — perhaps only for `detail=full`, or behind
   an opt-in query parameter such as `?include_stale=true`, so the per-model scope queries
   stay opt-in for callers that need them.
2. **Make the card tolerant of the missing field** — treat `undefined` as *unknown* rather
   than as stale, and show a neutral state (e.g. "—" or "unknown") instead of counting it
   as stale. This mirrors what the per-row badge already does.

Option 1 makes the card correct; option 2 stops it asserting something it cannot know.
Doing both would be ideal.

oldnicke:
## Summary

- add an opt-in `include_stale` query parameter that computes scope-aware staleness for listed mental models
- use exact per-model staleness in the control-plane summary instead of comparing against the global consolidation timestamp
- request metadata-only results for the stats view and keep unknown staleness from being reported as stale
- regenerate OpenAPI and generated clients, and expose the option through maintained Python and TypeScript wrappers

## Testing

- `./scripts/hooks/lint.sh`
- `uv run pytest tests/test_reflections.py::TestMentalModelsAPI::test_mental_models_api_crud -v`
- Python mental-model query mapping tests
- TypeScript mental-model query mapping tests
- `go build ./...`

Fixes #3139

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -4946,6 +4946,10 @@ def _register_routes(app: FastAPI):
         ),
         limit: int = Query(100, ge=1, le=1000),
         offset: int = Query(0, ge=0),
+        include_stale: bool = Query(
+            False,
+            description="Compute scope-aware staleness for each model. Expensive and disabled by default.",
+        ),
         request_context: RequestContext = Depends(get_request_context),
     ):
         """List mental models for a bank."""
@@ -4957,6 +4961,7 @@ def _register_routes(app: FastAPI):
                 detail=detail,
                 limit=limit,
                 offset=offset,
+                include_stale=include_stale,
                 request_context=request_context,
             )
             return MentalModelListResponse(items=[MentalModelResponse(**m) for m in mental_models])
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -11208,6 +11208,7 @@ class MemoryEngine(MemoryEngineInterface):
         detail: str = "full",
         limit: int = 100,
         offset: int = 0,
+        include_stale: bool = False,
         request_context: "RequestContext",
     ) -> list[dict[str, Any]]:
         """List pinned mental models for a bank.
@@ -11219,6 +11220,7 @@ class MemoryEngine(MemoryEngineInterface):
             detail: Detail level - 'metadata', 'content', or 'full'
             limit: Maximum number of results
             offset: Offset for pagination
+            include_stale: Compute scope-aware staleness for each model (expensive; opt-in)
             request_context: Request context for authentication
 
         Returns:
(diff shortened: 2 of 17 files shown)


oldnicke:
Closing as superseded by #3156, which fixes #3139 using the bank write watermark and avoids expensive per-model staleness queries on the polled stats view.

PR #3142, opened 2026-08-02
fix(reranker): foreground-priority lane for the local cross-encoder pool

mekenthompson:
## Problem

The local cross-encoder (`LocalSTCrossEncoder`) runs **every** rerank through a single shared FIFO `ThreadPoolExecutor(max_workers=RERANKER_LOCAL_MAX_CONCURRENT)`. There is no notion of priority.

Under consolidation/reflect load — a large fan-out of **internal (background)** recalls, each doing a cross-encoder rerank — an **interactive (foreground)** recall's rerank job queues *behind that entire wall* of background jobs. Foreground recall latency spikes as a result (observed on a busy deployment: foreground recalls 5-8s while background reranks wait 20-37s).

## Fix

Introduce `_PriorityRerankExecutor` — a fixed-size worker-thread pool backed by a `queue.PriorityQueue` — that dispatches **foreground reranks ahead of *queued* background ones**. A foreground job jumps the queue; it never preempts a background job already running in a worker, so background continues to make progress between foreground arrivals. This is foreground **priority**, not background exclusion.

`LocalSTCrossEncoder` now submits through this pool (with a lazy getter so `predict()` still works when a caller injects the model without calling `initialize()`, preserving the previous `run_in_executor(None, ...)` fallback behaviour).

### Signal plumbing

The foreground/background signal already exists at the recall call site: **`RequestContext.internal`** (set `True` by reflect sub-recalls and maintenance; `False` for user-facing recalls). It is threaded down:

- `MemoryEngine._search_with_retries` derives `background = request_context.internal` at the rerank call.
- `CrossEncoderReranker.rerank(..., *, background=False)` forwards it.
- `CrossEncoderModel.predict(..., *, background=False)` receives it; `LocalSTCrossEncoder` uses it for queue priority. Remote providers accept and ignore it (they have their own concurrency controls).

## Design choice

Two options were weighed (priority queue vs. reserved-slot semaphore). The **priority queue** was chosen: it is the simplest change that keeps worker-thread semantics, it yields a fully deterministic, timing-free ordering test, and it directly expresses "foreground PRIORITY." Background starvation is not a practical risk here — foreground reranks are driven by bounded, bursty per-user recall turns and cannot saturate the pool, while background is the bulk fan-out.

> Note: the codebase currently has a single `_search_semaphore` at the recall-admission layer (no separate background-reservation semaphore at `HEAD`), so this change introduces the foreground/background distinction at the reranker layer rather than mirroring an existing two-semaphore pattern.

## Tests

`tests/test_reranker_foreground_priority.py` — deterministic, no sleeps, asserts ordering/semantics not wall-clock timing:

- **`test_foreground_dispatched_before_queued_background`** — with one worker held busy, a foreground job submitted *after* three background jobs drains **first** (`["fg","bg1","bg2","bg3"]`). A plain FIFO pool yields submission order `["bg1","bg2","fg","bg3"]` and fails this assertion (verified).
- FIFO tie-breaking within a priority class, exception propagation, `predict(background=...)` → executor plumbing, and `CrossEncoderReranker.rerank` → `predict` forwarding.

Scoped `ruff check` / `ruff format` clean; `ty` adds no new diagnostics; existing reranker tests pass unchanged.

## Scope

Single concern: the local cross-encoder pool. `FlashRankCrossEncoder` has the identical FIFO pattern and could adopt the same `_PriorityRerankExecutor` as a follow-up; it accepts and ignores `background` for now.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -7,11 +7,14 @@ Configuration via environment variables - see hindsight_api.config for all env v
 """
 
 import asyncio
+import itertools
 import logging
 import os
+import queue
+import threading
 import warnings
 from abc import ABC, abstractmethod
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import Future, ThreadPoolExecutor
 from typing import Any
 
 import httpx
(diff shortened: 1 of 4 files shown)


mekenthompson:
Closing this pending a more complete revision. The foreground-priority rerank lane here is correct, but on its own it doesn't achieve the broader intent (interactive recall staying fast while a large consolidation backlog drains): production profiling shows the residual latency is upstream of the reranker, at recall *admission* — the current background lane is a strict, non-work-conserving reservation. We're reworking this into a single work-conserving priority admission gate (foreground always jumps the queue; background borrows idle foreground slots) and will resubmit as one coherent change. See you shortly with the fuller version.

PR #3144, opened 2026-08-02
fix(consolidation): apply sanitize_text to the _DedupDecision merge write path

iRonin:
Follows up #2544.

@nicoloboschi on that issue:

> the `_DedupDecision` merge path has no output-side scrub at all (`merged_text = decision.text` → straight to SQL). That's a real asymmetry worth a small dedicated fix (apply the existing character-safety `sanitize_text` to all three consolidation write paths)... a clean standalone issue I'd support.

This is that fix, scoped to the merge path.

### Character-safety only

`sanitize_llm_output` is an alias of `sanitize_text` (`llm_wrapper.py`), which strips ASCII control characters and lone UTF-16 surrogates — the things that break `json.loads` and the Postgres UTF-8 encode. This is **not** the declined content sanitizer; no content filtering is added here.

### The asymmetry

`_CreateAction` and `_UpdateAction` each scrub `text` through a `sanitize_llm_output` field validator. `_DedupDecision` has only `_normalize_action`, so the merge path handed the LLM's synthesized text to the fold `UPDATE` with just `.strip()` applied.

```diff
-        best_id=best_id, merged_text=decision.text.strip() or best_text, should_merge=True, best_text=best_text
+    merged_text = (sanitize_llm_output(decision.text) or "").strip() or best_text
```

The `best_text` fallback is preserved: a decision whose text sanitizes away to empty still folds into the twin's existing text rather than blanking the row.

### Test

Adds `test_dedup_llm_merge_sanitizes_text_before_write` next to the existing `test_dedup_llm_merge_folds_into_twin`, which is left untouched. It injects a NUL into the merge decision and asserts the fold `UPDATE` argument is clean.

### Verification

- The new test is **red without** the production line and **green with** it. The failure is on `assert "\x00" not in args[1]`, and the other 41 tests in the file pass either way — so it detects this gap and nothing else.
- `tests/test_consolidation_dedup.py` — 42 passed.
- `tests/test_consolidation*.py` — 114 passed, vs 113 on unmodified `main`; the delta is exactly this test. Both runs report the same 83 setup errors, which are environmental: this checkout lacks the optional `sentence-transformers` extra.
- `tests/test_llm_wrapper.py` — 47 passed.
- `ruff check` clean, `ruff format --check` clean, `ty check hindsight_api/` reports no diagnostics in `consolidator.py`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -262,9 +262,8 @@ async def _dedup_adjudicate(
     )
     if decision.action != "merge":
         return _DedupOutcome(best_id=best_id, merged_text="", should_merge=False, best_text=best_text)
-    return _DedupOutcome(
-        best_id=best_id, merged_text=decision.text.strip() or best_text, should_merge=True, best_text=best_text
-    )
+    merged_text = (sanitize_llm_output(decision.text) or "").strip() or best_text
+    return _DedupOutcome(best_id=best_id, merged_text=merged_text, should_merge=True, best_text=best_text)
 
 
 async def _dedup_reconcile_create(
--- a/hindsight-api-slim/tests/test_consolidation_dedup.py
+++ b/hindsight-api-slim/tests/test_consolidation_dedup.py
@@ -337,6 +337,24 @@ async def test_dedup_llm_merge_folds_into_twin() -> None:
     assert args[3] == uuid.UUID(_TWIN_ID)  # onto the twin row
 
 
+async def test_dedup_llm_merge_sanitizes_text_before_write() -> None:
+    # The merge path writes the LLM's synthesized text straight to the fold UPDATE, so it needs
+    # the same character-safety scrub _CreateAction/_UpdateAction already apply via field_validator.
+    # A raw NUL reaching the driver breaks the Postgres UTF-8 encode.
+    kwargs, conn, llm = _ctx()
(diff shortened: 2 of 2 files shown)


iRonin:
Heads-up on the red check, since it doesn't come from this PR.

**`verify-generated-files`** — the out-of-sync file is `skills/hindsight-docs/references/openapi.json` (+16/−2). It was last regenerated in 218e6d34 (2026-07-30); 1ce308f3 (#3109, 2026-07-31) then changed the bank/document response models (adds `last_write_at`, reworks `last_document_at`) without a regen, and the diff shape matches those fields. It reproduces on any PR against `main` and is unrelated to this change, which only touches a function body in `consolidator.py` plus a test. I've deliberately kept the regeneration out of this PR rather than mix an unrelated generated file into a two-file fix — `./scripts/generate-docs-skill.sh` on `main` should clear it, and I'm happy to leave that to you.

**On test evidence** — `test-api` is gated on `detect-changes.outputs.has_secrets == 'true'`, so it can't run on a fork PR; the green checks here don't exercise the new test. The verification section in the description is the evidence I have: the test is red without the production line (failing on `assert "\x00" not in args[1]`) and green with it, and the consolidation suite goes 113 → 114, so the delta is exactly this test. A maintainer run with secrets would be welcome.

nicoloboschi (approved):
Approve. Correct, minimal fix at the single `_dedup_adjudicate` chokepoint — covers both reconcile-create and reconcile-update fold paths. Fallback (`or best_text`) and `.strip()` semantics preserved; test asserts the char is removed, not just that the text was dropped. Character-safety only, consistent with the declined-content-sanitizer decision on #2544/#2547.

Issue #3002, opened 2026-07-28
Retain workers wedge until restart after a bulk_insert_entities deadlock


Fyko:
### Bug Description

A deadlock in `bulk_insert_entities` parks a `batch_retain` task forever. The task keeps its worker slot, so after enough of them the worker stops claiming retains entirely and the backlog grows without bound. Restarting the worker clears it for about four minutes.

Our worker sat at 22 retain slots held and 399 claimable retains starved for 50 minutes before we noticed.

The chain:

1. `bulk_insert_entities` inserts an unordered array with `ON CONFLICT DO NOTHING` ([`ops_postgresql.py#L256-L268`](https://github.com/vectorize-io/hindsight/blob/705757f362552918dfb0242906cb8466de320378/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py#L256-L268)). Two concurrent retains sharing entity names take their index locks in array order, in different orders, and Postgres kills one. Reached from [`entity_resolver.py#L761`](https://github.com/vectorize-io/hindsight/blob/705757f362552918dfb0242906cb8466de320378/hindsight-api-slim/hindsight_api/engine/entity_resolver.py#L761) via [`orchestrator.py#L362`](https://github.com/vectorize-io/hindsight/blob/705757f362552918dfb0242906cb8466de320378/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py#L362).

2. `_db_consumer` stops draining. The producer/consumer queue fills, so `_llm_producer` blocks on a put and stops issuing LLM calls.

3. Both sit inside [`await asyncio.gather(_llm_producer(), _db_consumer())`](https://github.com/vectorize-io/hindsight/blob/705757f362552918dfb0242906cb8466de320378/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py#L1828), which has no wall-clock bound. The task never returns and never releases its slot in [`_execute_task_inner`](https://github.com/vectorize-io/hindsight/blob/705757f362552918dfb0242906cb8466de320378/hindsight-api-slim/hindsight_api/worker/poller.py#L772).

None of the existing timeouts cover it:

| Env var | Value | What it bounds |
| --- | --- | --- |
| `HINDSIGHT_API_LLM_TIMEOUT` | 500 | one LLM call |
| `HINDSIGHT_API_DB_STATEMENT_TIMEOUT` | 600 | one query |
[...]

PR #3145, opened 2026-08-03
fix(llm): release concurrency permits during retry backoff

### LINKED ISSUE #3130: LLM retry backoff holds concurrency permits and can starve unrelated operations
### Bug Description

In v0.8.6, `LLMProvider.call()` acquires the per-operation semaphore and global LLM semaphore around the complete provider call.

The OpenAI-compatible provider performs its retry loop—including exponential-backoff sleeps—inside that call. Consequently, both permits remain occupied while the operation is sleeping and no upstream request is active.

If enough calls enter long backoff simultaneously, they can consume the global concurrency pool and prevent unrelated retain, reflect, or consolidation operations from attempting their requests.

This report is not requesting that any error become non-retryable. Existing retry coverage, attempt counts, and delays should remain intact.

### Steps to Reproduce

1. Run Hindsight v0.8.6 with a global LLM concurrency limit of one:

   ```text
   HINDSIGHT_API_LLM_MAX_CONCURRENT=1
   ```

2. Configure the OpenAI-compatible provider with more than one retry and a controlled backoff delay.
3. Point it at a test endpoint that returns a retryable `APIStatusError`.
4. Start call A and wait until it enters retry backoff.
5. While A is sleeping, start unrelated call B.
6. Observe that B cannot acquire the global permit until A finishes sleeping and begins or completes its next attempt.

No HTTP request is active during A's sleep, but the permit remains unavailable.

### Expected Behavior

Concurrency permits should represent active upstream attempts rather than the complete logical call.

After a failed attempt:

1. release the global and per-operation permits;
2. perform the configured backoff;
3. reacquire the per-operation permit;
4. reacquire the global permit;
5. execute the next attempt.

Retryable error classes, attempt counts, backoff calculations, timeouts, tracing, and final error behavior should remain unchanged.

### Actual Behavior

The wrapper owns the permits around the complete provider invocation:

- [`LLMProvider.call()` in v0.8.6](https://github.com/vectorize-io/hindsight/blob/08995e3013858e705fb4ca27c0ade3a286ef4750/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py#L954-L966)
- [`call_with_tools()` in v0.8.6](https://github.com/vectorize-io/hindsight/blob/08995e3013858e705fb4ca27c0ade3a286ef4750/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py#L1090-L1105)

The OpenAI-compatible retry loop and its sleeps execute beneath that ownership boundary:

- [status-error retry path](https://github.com/vectorize-io/hindsight/blob/08995e3013858e705fb4ca27c0ade3a286ef4750/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py#L1077-L1152)

Per-operation caps can reserve global headroom, and v0.8.6's retain wall timeout bounds the duration of some incidents, but neither changes the permit semantics during backoff.

### Version

v0.8.6

Release commit:

```text
08995e3013858e705fb4ca27c0ade3a286ef4750
```

### LLM Provider

OpenAI-compatible.

The same ownership concern may apply to other providers whose retry loops execute inside the wrapper's semaphore scope.

### Proposed Direction

Move concurrency acquisition to the individual-attempt boundary, or provide providers with a shared per-attempt permit context.

The intended lifecycle would be:

```text
backoff
→ acquire per-operation permit
→ acquire global permit
→ perform one upstream attempt
→ release both permits
→ classify the result
→ repeat if retryable
```

Semaphore policy should remain centralized rather than independently reimplemented by every provider.

### Acceptance Tests

1. With global concurrency one, call A fails and enters controlled backoff; unrelated call B completes before A retries.
2. With per-operation and global caps configured, both permits are released during backoff.
3. A retry reacquires permits in per-operation-then-global order.
4. Existing error coverage, retry counts, and delay calculations remain unchanged.
5. Cancellation while waiting for the global permit releases an already-acquired per-operation permit.
6. Cancellation during an active request or backoff leaks no permits.
7. Cover both `call()` and `call_with_tools()`.
8. No active attempt exceeds either configured concurrency limit.
9. Tracing and usage remain attached to the complete logical call.

koriyoshi2041:
## Problem

Providers run their retry loops inside the wrapper semaphore scope. A call therefore holds its per-operation and global concurrency permits while sleeping between attempts, which can starve unrelated operations when the global limit is small. Fixes #3130.

## Fix

Permits now represent active upstream attempts instead of whole logical calls. Providers that own a retry loop declare `supports_attempt_scoped_concurrency()`; the wrapper then skips holding the semaphores and passes an `attempt_context` factory that the provider enters around each individual upstream request — including structured output, tool calls, streaming SSE consumption, and native Ollama structured requests. This applies to **all real providers** (OpenAI-compatible, OpenAI Responses, Anthropic, Gemini, LiteLLM, Codex, Claude Code, llama.cpp via delegation), per review. Permit acquisition remains centralized and keeps the per-operation-then-global order.

Worker-stage observability is preserved and extended (#3002): a call queued on a saturated semaphore keeps its `.queued` stage until the first post-acquire `attempt=N` stamp, and a failed attempt gains a `.backoff` suffix while the provider sleeps without permits — so "waiting for a permit", "request in flight", and "backing off" are all distinguishable in worker logs.

## Tests

- End-to-end regression through the real OpenAI-compatible retry loop: an unrelated call completes while another call sleeps out its backoff, and attempt 2 reacquires the permit before hitting the upstream.
- Cancellation while queued on the global permit releases the already-acquired per-operation permit.
- Stage lifecycle: `.queued` → `attempt=N` (only once permits are held) → `.backoff` during retry sleep.
- Wrapper wiring, attempt-context forwarding (llama.cpp delegation, Responses provider), and stage-publication tests.
- `uv run pytest tests/test_llm_per_op_concurrency.py tests/test_llamacpp_attempt_context.py tests/test_openai_compatible_response_hardening.py tests/test_openai_responses_provider.py tests/test_llm_strict_schema.py tests/test_llm_trace.py ...` — all green; `ruff` + `ty` clean.

## Risk

Retry classification, backoff calculation, attempt counts, tracing context, and usage aggregation are unchanged. The main behavioral change is that queued retries now compete fairly with other calls instead of retaining their previous slot — under saturation a retrying call's end-to-end latency can grow, which is the intended fairness trade-off.


normalnormie:
## Review summary

The attempt-scoped permit design addresses the reported starvation mechanism correctly. OpenAI-compatible network requests remain concurrency-limited, while retry backoff occurs outside the permit context. Retry classification, attempt counts, and delay calculations appear unchanged.

### Observability regression

The worker-stage contract currently states that a call remains marked `.queued` until its concurrency permits have been acquired. In the proposed OpenAI-compatible path, the stage changes to the base/`.attempt=N` state before `attempt_context()` acquires those permits.

Consequently, a call waiting on either semaphore can appear to be an active provider attempt—the ambiguity previously addressed by #3002.

Suggested behavior:

1. Set `.queued` before each permit acquisition.
2. Acquire the per-operation and global permits.
3. Set `.attempt=N` only after both permits are held.
4. Execute the upstream request.

### Minimal regression coverage

A compact parameterized test covering both `call()` and `call_with_tools()` should be sufficient. With per-operation and global limits enabled, it should verify that:

- an unrelated call completes while another call is backing off;
- both permits are released during backoff;
- the retry reacquires them before its next upstream attempt.

Separate tests for every retry exception and provider path do not appear necessary because those paths share the same attempt context.

### CI

The Core LLM test job was skipped in the current workflow run. The new concurrency regression should execute in required CI before merge.

Overall, the implementation direction is sound. Preserving the queued-stage contract and covering both public call paths would close the remaining correctness gap without expanding the scope of the PR.

nicoloboschi (changes requested):
good catch, can you apply this to all llm providers?

koriyoshi2041:
Updated the PR to scope concurrency permits to each upstream attempt across OpenAI-compatible, Anthropic, Gemini/Vertex, LiteLLM/Router, Claude Code, Codex, and the LlamaCpp delegate. Retry backoff, auth refresh, parsing, and validation now run outside the permits. Added deterministic coverage for backoff release, stage ordering, and LlamaCpp context forwarding. Local validation: 180 passed, 4 skipped; Ruff, formatting, ty, and repository pre-commit hooks passed.

PR #3152, opened 2026-08-03
fix(retain): reject batches with mixed document IDs

### LINKED ISSUE #3010: Mixed explicit and missing document IDs implicitly merge items and mutate per-item tags on delta retain
## Bug Description

A retain batch containing exactly one explicit `document_id` and one or more items without a `document_id` implicitly stores every item under the single explicit document.

This is surprising because:

- `document_id` is an item-level field.
- Explicit duplicate document IDs in a batch are rejected.
- Multiple distinct explicit document IDs are processed as separate documents.
- The item without an ID appears as though it should receive an automatically generated document ID, but instead it is absorbed into the only explicit document.

The implicit merge exposes broader scope inconsistencies, not only a tag-specific bug. The document row and the memory units derived from its items do not consistently agree on which item's attributes are authoritative:

| Field | Document-level persistence | Initial memory-unit persistence | Delta/reprocess implications |
| --- | --- | --- | --- |
| `context` | First item | Source item | Reprocess reconstructs one item from the document value |
| `event_date` | First item | Source item | Reprocess reconstructs one item from the document value |
| `metadata` | First item in `retain_params` | Source item | Document-level and recalled memory metadata can diverge |
| `observation_scopes` | First item | Source item | Consolidation scope can differ among units in one implicit document |
| `tags` | Union of all item tags | Source item, plus deprecated batch-level tags | Metadata-only delta overwrites every unit with the document-wide union |

Chunks are cut independently per item, but the item boundary and `content_index` are not persisted as a first-class database record. Once multiple items are implicitly merged into one document, later document-level operations cannot reliably reconstruct those original item scopes.

For tags specifically, the initial full retain keeps source-item tags on each memory unit while the document stores their union. Re-retaining identical content takes the metadata-only delta path and overwrites every memory unit's tags with the document-wide union, even though the content and item-level tags did not change.

## Steps to Reproduce

1. Retain a batch like:

```json
{
  "items": [
    {
      "content": "Alice works at Google.",
      "document_id": "doc-1",
      "context": "source A",
      "metadata": {"source": "a"},
      "tags": ["source:a"]
    },
    {
      "content": "Bob works at Microsoft.",
      "context": "source B",
      "metadata": {"source": "b"},
      "tags": ["source:b"]
    }
  ],
  "async": false
}
```

2. Inspect the stored document, chunks, and memory units.

   - Both items' chunks and memory units use `document_id = "doc-1"`.
   - The document context and metadata come from the first item.
   - The document tags are the union: `["source:a", "source:b"]`.
   - Memory units extracted from the first item retain the first item's context, metadata, and tags.
   - Memory units extracted from the second item retain the second item's context, metadata, and tags.

3. Submit the identical request again so it takes the no-content-change delta path.

4. Inspect the memory units again.

   - Every memory unit now has `["source:a", "source:b"]` because `update_memory_units_tags()` applies the merged document tags to all units for the document.

This can change strict tag-filtering or visibility behavior solely because identical content was retained a second time.

## Expected Behavior

Establish and enforce the invariant:

> One retain item corresponds to one document.

Every item should receive an effective document ID before grouping or async batch splitting:

- Use its explicit `document_id` when present.
- Generate a distinct stable ID for that item when omitted.
- Reject duplicate effective document IDs in the same batch.

In the example, the first item should remain in `doc-1` and the second should receive a generated document ID. A minimally compatible transition is to reject batches that mix items with and without `document_id`, requiring callers either to provide IDs for every item or omit them for every item.

With the one-item/one-document invariant, item-level and document-level `context`, `event_date`, `metadata`, `tags`, and `observation_scopes` have an unambiguous scope, and full retain and delta retain can produce the same observable state.

## Simplifications Enabled by the Invariant

Once every retain item has exactly one effective document ID, the implementation can be simplified in several concrete areas:

- **Input normalization**
  - Resolve or generate one stable `document_id` for every item at ingress.
  - Validate duplicate effective IDs once, before async splitting or persistence.
  - Remove the ambiguous case where missing IDs are implicitly absorbed into the only explicit document.

- **Batch and document orchestration**
  - Treat a retain batch as a collection of independent item/document operations.
  - Remove grouping behavior that depends on the number of distinct non-empty IDs.
  - Persist generated IDs before async child-batch creation so retries use the same document identities.
  - Keep result aggregation at the batch layer without allowing item attributes to leak across documents.

- **Document attribute construction**
  - Build `retain_params` directly from the document's single source item.
  - Remove first-item selection for `context`, `event_date`, `metadata`, and `observation_scopes`.
  - Build document tags from that item's tags, plus deprecated batch-level tags while compatibility remains.
  - Remove unions of tags from otherwise unrelated items.

- **Chunk and memory-unit handling**
  - Guarantee that every chunk and extracted memory unit inherits attributes from one unambiguous source item/document.
  - Avoid needing a persisted `content_index` merely to recover which item supplied metadata, tags, context, or scopes.
  - Make document-wide metadata and tag updates safe because every unit under the document has the same source attributes.

- **Delta retain**
  - Make the current first-content template for changed chunks safe by construction.
  - Ensure full retain, metadata-only delta retain, partial delta retain, and append all produce the same document/memory attributes.
  - Remove multi-item fallback logic whose only purpose is preserving or reconstructing per-item state inside one document.

- **Deprecated compatibility paths**
  - Restrict the deprecated batch-level `document_id` to single-item calls.
  - Remove the stale multi-item batch-level document ID example and eventually remove the deprecated argument.
  - Keep old documents readable, while defining separately whether re-retain/reprocess normalizes legacy multi-item documents to document-level attributes.

- **Tests and documentation**
  - Test the one-item/one-document invariant at synchronous, asynchronous, retry, append, and delta boundaries.
  - Document that item-level `metadata`, tags, context, timestamps, and observation scopes are the attributes of that item's document.
  - Replace tests and examples that imply multiple items can intentionally share one document ID.

## Actual Behavior

When a batch has exactly one distinct non-empty item-level `document_id`, the orchestrator does not enter its multi-document grouping branch. It selects that sole ID as the effective document ID for the entire batch, including items that omitted the field.

The document stores first-item values for several retain parameters and a union for tags, while the initial full path stamps source-item values onto memory units. Later document-level delta or reprocess operations can collapse those differences.

## Version

Current `main` at `e57765e0123da0520b660b55f49d0b0147dbe3c0`.

## LLM Provider

Provider-independent; this is document grouping and persistence behavior.

r266-tech:
## Summary

- reject retain batches that mix explicit document IDs with omitted or empty IDs, preventing an explicit ID from absorbing unrelated items
- share the validation between synchronous and asynchronous retain while preserving all-implicit and all-explicit unique batches
- reject invalid asynchronous batches before creating parent or child operation rows

Fixes #3010.

## Tests

- `uv run pytest tests/test_async_batch_retain.py -n 0` (39 passed)
- `uv run ruff check .`
- `uv run ruff format --check .`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -19,6 +19,7 @@ import logging
 import sys
 import time
 import uuid
+from collections import Counter
 from collections.abc import AsyncIterator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
@@ -425,6 +426,28 @@ from .token_encoding import get_token_encoding
 
 RetainOutboxCallback = Callable[[asyncpg.Connection], Awaitable[None]]
 RetainOutboxCallbackFactory = Callable[[list[RetainContentDict]], RetainOutboxCallback | None]
+
+
+def _validate_batch_document_ids(contents: list[RetainContentDict]) -> list[str]:
+    """Return explicit document IDs after validating batch routing semantics."""
+    document_ids = [document_id for item in contents if (document_id := item.get("document_id"))]
+
+    duplicates = [document_id for document_id, count in Counter(document_ids).items() if count > 1]
+    if duplicates:
+        raise ValueError(
+            f"Batch contains duplicate document_ids: {duplicates}. "
+            f"Each content item in a batch must have a unique document_id to avoid race conditions."
+        )
+
+    # A single explicit ID currently becomes the effective ID for every implicit
+    # item, silently combining otherwise independent content into one document.
+    if document_ids and len(document_ids) != len(contents):
+        raise ValueError(
+            "Batch cannot mix content items with and without document_ids. "
+            "Provide a unique document_id for every content item, or omit document_id from all items."
+        )
+
+    return document_ids
 
 
 @dataclass(frozen=True)
@@ -3706,17 +3729,7 @@ class MemoryEngine(MemoryEngineInterface):
         if outbox_callback is None and outbox_callback_factory is not None:
             outbox_callback = outbox_callback_factory(contents)
 
-        # Validate no duplicate document_ids in the batch
-        # Having duplicate document_ids causes race conditions in document upserts during parallel processing
-        doc_ids = [item.get("document_id") for item in contents if item.get("document_id")]
-        if len(doc_ids) != len(set(doc_ids)):
-            from collections import Counter
-
-            duplicates = [doc_id for doc_id, count in Counter(doc_ids).items() if count > 1]
-            raise ValueError(
-                f"Batch contains duplicate document_ids: {duplicates}. "
-                f"Each content item in a batch must have a unique document_id to avoid race conditions."
-            )
+        _validate_batch_document_ids(contents)
 
         # Validate update_mode=append requires document_id
         for item in contents:
@@ -14365,17 +14378,7 @@ class MemoryEngine(MemoryEngineInterface):
             if replay is not None:
                 return replay
 
-        # Validate no duplicate document_ids in the batch
-        # Having duplicate document_ids causes race conditions in document upserts during parallel processing
-        doc_ids = [item.get("document_id") for item in contents if item.get("document_id")]
-        if len(doc_ids) != len(set(doc_ids)):
-            from collections import Counter
-
-            duplicates = [doc_id for doc_id, count in Counter(doc_ids).items() if count > 1]
-            raise ValueError(
-                f"Batch contains duplicate document_ids: {duplicates}. "
-                f"Each content item in a batch must have a unique document_id to avoid race conditions."
-            )
+        doc_ids = _validate_batch_document_ids(cast(list[RetainContentDict], contents))
 
         # Calculate total token count and determine if we need to split
         total_tokens = sum(count_tokens(item.get("content", "")) for item in contents)
--- a/hindsight-api-slim/tests/test_async_batch_retain.py
+++ b/hindsight-api-slim/tests/test_async_batch_retain.py
@@ -17,6 +17,22 @@ from hindsight_api.extensions import RequestContext
 pytestmark = pytest.mark.xdist_group("worker_tests")
 
 
+MIXED_DOCUMENT_ID_BATCHES = [
+    [
+        {"content": "Explicit item", "document_id": "doc1"},
+        {"content": "Implicit item"},
+    ],
+    [
+        {"content": "Implicit item"},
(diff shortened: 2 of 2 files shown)


koriyoshi2041 (approved):
Verified the full async batch retain suite locally at 631620d: 39 passed with the embedded DB and local model extras. The shared validator covers both submit paths, rejects before async operation creation, and preserves the existing duplicate-ID and multi-document cases.

ebarkhordar (commented):
`_validate_batch_document_ids` (memory_engine.py:444) rejects every batch that mixes explicit and omitted IDs, but only the single-explicit-ID case actually merges. A batch carrying two or more distinct explicit IDs plus an ID-less item already routes correctly on main: `retain_batch` groups by document when `len(unique_doc_ids) > 1` and gives each ID-less item its own generated ID (orchestrator.py:778-783, `doc_key = cd.get("document_id") or str(uuid.uuid4())`). The collapse #3010 reports comes from orchestrator.py:910-913, which is reached only when that grouping branch is skipped.

Measured at 631620d5 against main 5c15f28a, stubbing `get_bank_profile` and recording the recursive `retain_batch` calls:

```
[{"content": "A"}, {"content": "B", "document_id": "doc-1"}, {"content": "C", "document_id": "doc-2"}]
  main   -> 3 groups: document_id='4de7f342-4dcc-...' / 'doc-1' / 'doc-2'
  #3152  -> ValueError: Batch cannot mix content items with and without document_ids.

[{"content": "A", "document_id": "doc-1"}, {"content": "B"}]
  main   -> no grouping, effective_doc_id='doc-1'   (the #3010 merge)
  #3152  -> ValueError
```

So the guard converts a working three-document batch into an error. Restricting it to the shape that actually merges keeps that case working:

```python
if document_ids and len(document_ids) != len(contents) and len(set(document_ids)) == 1:
```

Separately, #3010's Expected Behavior asks for the opposite direction: give every ID-less item a distinct generated ID rather than reject the batch. The grouping branch already does exactly that, so relaxing its condition from "more than one distinct explicit ID" to "at least one explicit ID and at least one item without" would close #3010 without adding a new error. Is the narrower contract deliberate?

One caveat on the numbers above: I ran the grouping recursion only, not the pipeline against a database, so those are the groups the orchestrator forms, not the document rows they end up producing.

nicoloboschi:
Closing in favor of #3173.

Rejecting mixed explicit/implicit `document_id` batches is a breaking change for callers who currently rely on grouping items under one document. To preserve backwards compatibility, #3173 keeps the existing behavior and instead documents the item-level `document_id` semantics (a distinct id per source document; items sharing an id are grouped), surfacing the guidance in the field description and OpenAPI spec.

PR #3170, opened 2026-08-04
fix(recall): fall back when optional reranker fails

### LINKED ISSUE #3168: Reranker failure is fatal to recall — no fail-open, no way to make the stage optional
### Bug Description

If the reranker is unreachable, the whole recall fails with HTTP 500. Candidates
that the embedding stage already found get thrown away instead of being returned
in embedding order.

Follow-up to #1081 and #2991. Both got fixed for their specific trigger (#3001
added the 429 retry), but the shape underneath is still there: an optional
refinement stage is a hard dependency, and there is no setting to change that.

`engine/search/reranking.py:291` raises and the exception travels all the way up
to `RuntimeError: Failed to search memories`. Nothing along that path treats the
reranker as optional. Grepping the settings for a way out only turns up
`SKIP_LLM_VERIFICATION` and `FAIL_ON_EXTRACTION_ERRORS`, neither of which has
anything to do with reranking.

My reranker runs on a workstation, so it is legitimately off some of the time.
That is how I ran into this. The practical consequence is that whoever hosts the
reranker separately inherits its uptime as the uptime of memory retrieval as a
whole — which is a lot of weight for a stage that is supposed to be a
refinement. On my data the embedding stage put the right memory first in every
sample I tried; the reranker changed the scores, not the order. So the current
behaviour trades a slightly worse answer for no answer.

Two ways I could see this going:

1. fail open — log it, return embedding-ordered results, mark the response as
   degraded somehow
2. add `HINDSIGHT_API_RERANKER_REQUIRED`, default `true` so nothing changes for
   existing setups, and let people set it to `false`

(2) is the smaller change and leaves the call to whoever runs it. Happy to put
together a PR if that direction is acceptable.

In the meantime I put a small proxy in front of the reranker. When the upstream
does not answer it returns neutral scores — one entry per document, all `0.0` —
so the ordering from the embedding stage comes through untouched.

One thing that cost me an hour and might be worth a line in the docs regardless
of how this issue goes: the proxy's own timeout has to be clearly shorter than
`RERANKER_*_TIMEOUT`. I had both at 5s first. The fallback fired correctly, the
proxy logged it, but by then hindsight's own timeout had already expired and
recall still died with `ReadTimeout`. 2s inner / 15s outer works.

### Steps to Reproduce

1. Configure a remote reranker. I used `RERANKER_PROVIDER=cohere` with a custom
   base URL pointing at `llama-server --reranking --pooling rank`
   (`bge-reranker-v2-m3-Q8_0.gguf`), since there is no llamacpp provider —
   #2773 was closed without merging.
2. Store a few memories in a bank and confirm recall works.
3. Make the reranker unreachable. I dropped the packets rather than stopping the
   service, to get a timeout instead of a connection refused:

   ```
   iptables -I DOCKER-USER 1 -d <reranker-host> -p tcp --dport <port> -j DROP
   ```

4. `POST /v1/{tenant}/banks/{bank}/memories/recall` with any query that matched
   before.
5. Remove the rule again and repeat the exact same query.

Step 4 fails, step 5 succeeds. Same data, same query, only the reranker's
reachability differs.

### Expected Behavior

Recall returns results ordered by the embedding stage — the same output you get
when no reranker is configured at all.

The candidates are already in hand at that point; the reranker only reorders
them. Losing the refinement should cost ranking quality, not the answer.

Ideally the response would say somewhere that it came back degraded, so this
does not go unnoticed for weeks.

### Actual Behavior

The whole request fails:

```
{"detail":"Failed to search memories (ConnectTimeout): ConnectTimeout('')"}
HTTP 500 after 6.1s
```

No partial result, no fallback. After removing the iptables rule the identical
query returns HTTP 200 in 0.68s with the ordering I would expect.

Same thing happens on other reranker-side failures, not just timeouts. A 404
from a misconfigured base URL takes recall down in exactly the same way:

```
httpx.HTTPStatusError: Client error '404 Not Found' for url '<reranker>'
RuntimeError: Failed to search memories (HTTPStatusError)
```

### Version

0.8.6 (ghcr.io/vectorize-io/hindsight:latest)

### LLM Provider

Other

oldnicke:
## Summary

- add HINDSIGHT_API_RERANKER_REQUIRED as a static server setting that defaults to true
- when explicitly disabled, fall back to RRF-based scoring after reranker initialization or prediction failures
- surface degraded recalls in search traces, OpenTelemetry attributes, warning logs, and null reranker scores
- preserve cancellation propagation and skip min_scores.reranker only when no real reranker score exists

## Compatibility

The default remains strict, so existing deployments continue to fail recall when the reranker fails. Fail-open behavior requires HINDSIGHT_API_RERANKER_REQUIRED=false.

## Tests

- 79 related reranker, scoring, trace, timeout, and min-score tests
- ./scripts/hooks/lint.sh
- ./scripts/hooks/check-unused.sh
- uv run ty check hindsight_api/config.py hindsight_api/engine/memory_engine.py
- ./scripts/hooks/generate-docs-skill.sh

## Generated files

Generated documentation references were refreshed from the latest main sources, so this PR can pass generated-file verification independently.

Fixes #3168

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -240,6 +240,9 @@ HINDSIGHT_API_LOG_LEVEL=info
 # Reranker Configuration (Optional - uses local by default)
 # Provider: "local" (default) or "tei" (HuggingFace Text Embeddings Inference)
 # HINDSIGHT_API_RERANKER_PROVIDER=local
+# Fail recall when reranker initialization or scoring fails (default). Set to
+# false to return RRF-ranked results and mark the trace as degraded instead.
+# HINDSIGHT_API_RERANKER_REQUIRED=true
 # Trusted gateway attribution (disabled by default). When enabled, remote
 # reranker requests include X-Hindsight-Bank-Id with the current bank ID.
 # HINDSIGHT_API_RERANKER_SEND_BANK_AS_HEADER=false
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -445,6 +445,7 @@ ENV_LITELLM_API_BASE = "HINDSIGHT_API_LITELLM_API_BASE"
 ENV_LITELLM_API_KEY = "HINDSIGHT_API_LITELLM_API_KEY"
 
 ENV_RERANKER_PROVIDER = "HINDSIGHT_API_RERANKER_PROVIDER"
+ENV_RERANKER_REQUIRED = "HINDSIGHT_API_RERANKER_REQUIRED"
 ENV_RERANKER_SEND_BANK_AS_HEADER = "HINDSIGHT_API_RERANKER_SEND_BANK_AS_HEADER"
 ENV_RERANKER_LOCAL_MODEL = "HINDSIGHT_API_RERANKER_LOCAL_MODEL"
 ENV_RERANKER_LOCAL_FORCE_CPU = "HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU"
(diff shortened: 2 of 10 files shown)


nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/3176

PR #3184, opened 2026-08-05
feat(control-plane): log in with your own dataplane API key

### LINKED ISSUE #3183: Control plane can only serve one tenant: no UI path for multi-tenant TenantExtension deployments
## Summary

The dataplane supports genuine multi-tenancy through `TenantExtension` — each API key resolves to its own schema, and isolation is enforced by Postgres rather than by application filtering. The control plane has no equivalent. It authenticates to the dataplane with a single process-wide key, so it can only ever display one tenant's data.

The result is that a deployment which gives each team member their own tenant has no UI story for them at all.

## Current behaviour

`hindsight-control-plane/src/lib/hindsight-client.ts` resolves the key once, at module scope:

```ts
const DATAPLANE_API_KEY = process.env.HINDSIGHT_CP_DATAPLANE_API_KEY || "";

export const hindsightClient = new HindsightClient({
  baseUrl: DATAPLANE_URL,
  apiKey: DATAPLANE_API_KEY || undefined,
});

export const lowLevelClient = createClient(
  createConfig({
    baseUrl: DATAPLANE_URL,
    headers: DATAPLANE_API_KEY ? { Authorization: `Bearer ${DATAPLANE_API_KEY}` } : undefined,
  })
);
```

Those singletons, plus `getDataplaneHeaders()`, are what every route handler under `src/app/api` uses. There is no request-scoped path to the dataplane.

`HINDSIGHT_CP_ACCESS_KEY` (added in #1148) gates the UI, but it is explicitly a single shared secret and carries no identity — it controls whether the UI opens, not which tenant it opens as. The control plane exposes only three environment variables in total: `HINDSIGHT_CP_ACCESS_KEY`, `HINDSIGHT_CP_DATAPLANE_API_KEY`, `HINDSIGHT_CP_DATAPLANE_API_URL`.

## Impact

For a team where each person has their own tenant and schema:

- REST and MCP work correctly per person
- the UI can only ever show whichever tenant `HINDSIGHT_CP_DATAPLANE_API_KEY` belongs to
- the only workaround is one control-plane deployment per user, each on its own port or host with its own access key and dataplane key — which turns onboarding a teammate into a deploy

There is also a quiet failure mode worth flagging on its own. If a deployment enables a multi-tenant `TenantExtension` but leaves `HINDSIGHT_CP_DATAPLANE_API_KEY` set to a value that is not a valid tenant key, the UI still logs in normally and then renders empty — every dataplane call is returning 401 behind it, with nothing surfaced to say so.

## Question

Since the login screen already accepts a key, and the dataplane already knows how to resolve a key to a tenant, is there appetite for the UI to act as the tenant that the entered key belongs to — with the current single-key behaviour preserved as the default when that mode is not enabled?

## Related

- #2343 (closed) observed that the control plane "has even fewer provider/extension abstractions"; this is a concrete instance of that gap.
- #2235 covers the authorization model itself (teams, roles, scoped keys). This issue is narrower: tenant isolation that already exists and works, with no way to reach it from the UI.

## Version

0.8.6

aichemzee:
Closes #3183.

## Problem

The dataplane resolves an API key to a tenant. The control plane resolves its key once, at module scope:

```ts
const DATAPLANE_API_KEY = process.env.HINDSIGHT_CP_DATAPLANE_API_KEY || "";
export const hindsightClient = new HindsightClient({ baseUrl: DATAPLANE_URL, apiKey: ... });
```

So the whole process talks to the dataplane as one tenant. A deployment running a multi-tenant `TenantExtension` gets working per-user isolation over REST and MCP, and no way to reach it from the UI — short of running one control plane per user, each with its own port, access key and dataplane key.

## Approach

An opt-in mode, `HINDSIGHT_CP_TENANT_AUTH`, in which the key entered at the login screen **is** the dataplane credential:

- **Login validates against the dataplane.** The UI never invents its own notion of a valid key — it asks the dataplane, which already knows how to answer. No second source of truth.
- **The session carries that key, encrypted.** AES-256-GCM under `HINDSIGHT_CP_SESSION_SECRET`. This is a live credential, not a session identifier, so a lifted cookie should not unwrap into a working API key usable against REST or MCP outside the browser.
- **The dataplane client is built per request** from the session, replacing the module-scope singletons.

## Compatibility

Default behaviour is unchanged. With `HINDSIGHT_CP_TENANT_AUTH` unset, the existing `HINDSIGHT_CP_ACCESS_KEY` gate and the fixed `HINDSIGHT_CP_DATAPLANE_API_KEY` work exactly as before — no new UI, routes, headers or middleware behaviour appear. This follows the compatibility expectation in #2343.

## Diff shape

Most of the change is mechanical. `getDataplaneHeaders()` becomes async, so `await` propagates across the route handlers that call it — that is the bulk of the 68 files. The actual logic is three files:

- `src/lib/auth/session.ts` — encrypted tenant session alongside the existing HMAC access-key session
- `src/lib/hindsight-client.ts` — request-scoped client construction
- `src/middleware.ts` — route gating for both modes

## Verification

- `tsc --noEmit` clean
- 128 tests pass across 13 files, including a new `tests/lib/auth/tenant-session.test.ts` covering encrypt/decrypt round-trip, tampered-ciphertext rejection and expiry
- `eslint` reports 0 errors on the changed files (5 pre-existing `no-explicit-any` warnings, down from 7 at baseline)
- Exercised on a live deployment with two tenants on separate Postgres schemas: each sees only their own banks, and a revoked key stops working immediately

## Open questions

- `HINDSIGHT_CP_TENANT_AUTH` as the flag name, or fold this into a broader auth-profile concept as sketched in #2343?
- Session lifetime currently reuses `SESSION_MAX_AGE_SECONDS` (24h). For a mode where the cookie holds a real credential, a shorter default may be wanted.
- Happy to split the mechanical `await` propagation into its own commit if that makes review easier.

### DIFF
--- a/hindsight-control-plane/src/app/api/auth/login/route.ts
+++ b/hindsight-control-plane/src/app/api/auth/login/route.ts
@@ -5,14 +5,19 @@ import {
   ACCESS_KEY_COOKIE,
   SESSION_MAX_AGE_SECONDS,
   createSessionToken,
+  createTenantSessionToken,
+  getSessionSecret,
+  isTenantAuthEnabled,
   sessionCookieOptions,
 } from "@/lib/auth/session";
+import { DATAPLANE_URL } from "@/lib/hindsight-client";
 
 export async function POST(request: NextRequest) {
   const accessKey = process.env.HINDSIGHT_CP_ACCESS_KEY;
+  const tenantAuth = isTenantAuthEnabled();
 
-  // If no access key is configured, return 503
-  if (!accessKey) {
+  // If no auth mechanism is configured at all, return 503
+  if (!accessKey && !tenantAuth) {
     return NextResponse.json(
       localizeApiErrorPayload(request, {
         error: "Access key not configured",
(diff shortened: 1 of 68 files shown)


nicoloboschi:
we deliberately keep the control plane/UI single tenant for simplicity. 
usually users put a authz layer on top of it and/or fork the ui

PR #3200, opened 2026-08-06
fix(retain): preserve table headers across chunks

### LINKED ISSUE #3195: CSV / markdown-table chunking drops header context after the first chunk
### Use Case

I'm ingesting tabular files (CSV, and similarly large Excel sheets converted to markdown) via the file retain / upload API. Column semantics live only in the header row — e.g. `name,age,city` or the markdown table header `| name | age | city |`.

### Problem Statement

CSV is supported as an upload format (markitdown converts it to a markdown table), but after conversion the retain path treats the result as plain text and runs `chunk_text` / `RecursiveCharacterTextSplitter` with no table awareness.

When the converted table exceeds `retain_chunk_size` (default 3000 chars), later chunks are cut on line boundaries (`\n`) and **do not carry the header row**. Those chunks are just bare data rows, so fact extraction loses column meaning and can misattribute or invent field semantics.

This is the same class of failure as #2548 (structured envelope fields lost when a record is split as text) and the surrounding-context gap in #2551, but for tabular documents rather than JSONL sessions. There is currently no CSV-/table-specific chunking path.

Minimal example of what later chunks look like after a mid-table split:

```markdown
| Alice | 30 | NYC |
| Bob | 25 | LA |
```

without the preceding:

```markdown
| name | age | city |
| --- | --- | --- |
```

Every chunk of a tabular document would retain enough schema context for extraction to interpret columns correctly, without requiring clients to pre-chunk CSVs or inflate `retain_chunk_size` to fit whole files.

### Proposed Solution

When chunking markdown tables (and/or CSV-derived content), preserve header context across chunks. Concrete options, roughly in order of complexity:

1. **Repeat header on each chunk** — detect a GFM table (header + separator row), split on data-row boundaries, and prepend the header (+ separator) to every chunk. Highest value / lowest complexity for the CSV upload path.
2. **Row-aware packing** — pack whole table rows under `retain_chunk_size` (like JSONL line packing), never splitting mid-row; still repeat header on each chunk.
3. **Optional overlap / surrounding context** — related to #2551: give each chunk a small preceding window so headers (and prior rows) remain available when ambiguity appears.

A CSV-native path that chunks by row *before* markdown conversion would also work, but repeating the already-converted markdown header is enough for the current markitdown → retain pipeline.

### Alternatives Considered

- Raise `retain_chunk_size` so small CSVs fit in one chunk — does not scale to large exports.
- Client-side pre-chunking that injects the header into every slice — works, but only for that client; same gap remains for anyone using the file retain API directly.
- Relying on document-level `context` / `metadata` — helpful for global hints, but easy to omit and does not scale to wide schemas or multi-table documents.

### Priority

### Additional Context

- Upload path: file retain → markitdown (CSV → markdown table) → retain `chunk_text` (plain-text recursive splitter, `chunk_overlap=0`).
- Related: #2548 (envelope re-attach for structured JSON splits; PR #2760 proposed a fix but was not merged), #2551 (extra surrounding context at extraction time), #2113 (JSONL line-boundary chunking already shipped — analogous idea for table rows).

### Checklist

oldnicke:
Fixes #3195

## Summary

- detect GFM Markdown tables in retained content after JSON and JSONL handling
- pack complete data rows within the chunk budget and repeat the header and separator in every table chunk
- preserve surrounding Markdown, escaped pipes, alignment markers, fenced examples, and oversized-row stability

## Testing

- ./scripts/hooks/lint.sh
- ./scripts/hooks/check-unused.sh
- uv run ty check hindsight_api/engine/retain/fact_extraction.py hindsight_api/engine/memory_engine.py
- uv run pytest tests/test_chunking.py tests/test_markitdown_parser.py tests/test_batch_chunking.py -q (49 passed; two database cases rerun below against an isolated current schema)
- HINDSIGHT_API_DATABASE_URL=pg0://hindsight-3195:5595 uv run pytest tests/test_batch_chunking.py::test_large_batch_auto_chunks tests/test_batch_chunking.py::test_small_batch_no_chunking -q -n 0

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -480,14 +480,17 @@ _RECURSIVE_TEXT_SEPARATORS = [
     "",  # Characters (last resort)
 ]
 
+_MARKDOWN_TABLE_SEPARATOR_CELL_RE = re.compile(r"^:?-{3,}:?$")
+_MARKDOWN_FENCE_RE = re.compile(r"^ {0,3}(?P<marker>`{3,}|~{3,})")
+
 
 def _split_oversized_unit(text: str, max_chars: int) -> list[str]:
     """Sentence-aware split of a single unit that overflowed the budget.
 
-    Used when one JSONL line / conversation turn is so large it can't be kept
-    whole within the configured structured-chunk limit. The resulting fragments
-    are no longer valid JSON, but the fact extractor treats every chunk as plain
-    text.
+    Used when one JSONL line, conversation turn, or Markdown table row is so
+    large it can't be kept whole within the configured structured-chunk limit.
+    The resulting fragments may no longer preserve their source syntax, but the
+    fact extractor treats every chunk as plain text.
     """
     from langchain_text_splitters import RecursiveCharacterTextSplitter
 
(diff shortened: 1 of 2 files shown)


handnewb (approved):
Excellent fix! Detecting GFM tables in retained content and repeating header+separator in every chunk is the right approach — column semantics are preserved across chunk boundaries without requiring the extractor to cross-reference. The 94 new test lines in test_chunking.py give strong coverage. LGTM!

oldnicke:
@handnewb Thanks for the review !

I pushed the final update in commit c3a65f6e, which preserves Markdown table headers and separators across every generated chunk. The added chunking tests cover header preservation, multi-chunk tables, and related edge cases.

All CI checks are passing. Could you please take a quick look at the latest commit and re-approve when convenient?

nicoloboschi:
we prefer not adding this complexity at the moment, thanks

PR #3202, opened 2026-08-06
fix(worker): report bank-blocked consolidation tasks

### LINKED ISSUE #3194: Consolidation task still never claimed on v0.8.6 - observations stall forever
# Consolidation task still never claimed on v0.8.6 — observations stall forever

## Summary
Consolidation works once on first run, then **every subsequent consolidation task stays `claimable=1` and is never claimed by the worker**, so new observations stop being created. Confirmed still present in the latest **v0.8.6** (2026-07-29), not just the originally-reported v0.8.3.

This is a follow-up / more data on #2359. The issue remains reproducible and I have fresh logs from the current release.

## Environment
- Image: `ghcr.io/vectorize-io/hindsight:latest` (pulled 2026-07-29, `api_version` = **0.8.6**)
- Single container, embedded PostgreSQL, `HINDSIGHT_API_WORKER_ID=hindsight-prod` (stable), `HINDSIGHT_API_ENABLE_AUTO_CONSOLIDATION=true`
- LLM provider: OpenRouter / `openai/gpt-5-nano` (LLM itself is healthy — retain/recall work fine)

## Symptom / reproduction
1. Initial consolidation runs once — `last_consolidated_at` advances (e.g. `2026-07-17T09:23:01`), some observations get created.
2. After that, every new consolidation operation is enqueued but **never claimed**. Logs repeat every 30s:
   ```
   [WORKER_STATS] worker=hindsight-prod slots=0/10 | reserved: [consolidation=0/2(avail=2)] | shared=0/8(avail=8) | global: pending=1 ... my_active: none
   [PENDING_BREAKDOWN] consolidation: total=1 claimable=1 payload_null=0 retry_blocked=0 assigned=0
   ```
   The worker has 2 free reserved consolidation slots (`consolidation avail=2`), the task is `claimable=1` / `assigned=0`, yet the poller never claims it.
3. `pending_consolidation` climbs (here: 145), `total_observations` frozen.
4. `POST /consolidate` re-triggers dedupe onto the same never-claimed task.

## What I already ruled out
- **Not a dead-worker reservation issue (#2165):** `others: c8f8045891d1:1, 33ee967ffda4:5` appear, but the task is already `claimable=1`.
- **Not fixed by restart / recreate:** recreated the container with a fresh worker multiple times; behavior identical.
- **Not fixed by upgrading:** same logs on v0.8.4 and v0.8.6.
- Retain tasks claim fine — only consolidation claims are skipped.

## Expected behavior
A pending, `claimable=1` consolidation operation with available reserved slots should be claimed and processed (observations created, `last_consolidated_at` advances).

## Impact
Long-running deployments silently stop producing integrated observations; the UI shows an ever-growing "N observations pending". Workaround (`HINDSIGHT_API_ENABLE_AUTO_CONSOLIDATION=false`) only papers over it by giving up on observation generation.

## Request
Could someone look into why the worker poller skips consolidation claims while it still claims retain tasks? Happy to provide more logs or run a debug build / strace if helpful.

oldnicke:
## Summary

- classify pending-operation diagnostics into mutually exclusive buckets that match the durable claim predicates
- report a consolidation as `bank_blocked` when another processing consolidation owns the same bank
- emit an actionable recovery hint using `worker-status` and `decommission-worker` after the owner is confirmed dead
- cover the #3194 state and overlapping filters with a regression test

## Root cause

The claim path intentionally serializes consolidation per bank. A processing consolidation owned by an earlier worker therefore blocks a pending consolidation for that bank, but `[PENDING_BREAKDOWN]` did not model this rule and incorrectly reported the pending row as `claimable=1`.

This preserves the serialization guard. Automatically taking over an old-looking processing task would risk running two live consolidations for the same bank because worker ownership is not protected by a lease or fencing token. The new diagnostic identifies the actual blocker and points operators to the existing explicit recovery path.

## Tests

- `./scripts/hooks/lint.sh`
- `./scripts/hooks/check-unused.sh`
- `cd hindsight-api-slim && uv run pytest tests/test_worker.py -q` (104 passed)
- `cd hindsight-api-slim && uv run pytest tests/test_worker.py::test_pending_breakdown_explains_unclaimable_rows -q`

Fixes #3194

### DIFF
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -168,6 +168,17 @@ class SlotAvailability:
     """Remaining shared pool capacity (usable by any operation type)."""
 
 
+@dataclass
+class PendingBreakdown:
+    """Mutually exclusive claim states for pending operations."""
+
+    total: int = 0
+    claimable: int = 0
+    payload_null: int = 0
+    retry_blocked: int = 0
+    bank_blocked: int = 0
+
+
 class WorkerPoller:
     """
     Polls the database for pending tasks and executes them.
@@ -1339,16 +1350,16 @@ class WorkerPoller:
             global_pending = 0
             all_worker_counts: dict[str, int] = {}
             # operation_type -> aggregated bucket counts across schemas
-            pending_breakdown: dict[str, dict[str, int]] = {}
+            pending_breakdown: dict[str, PendingBreakdown] = {}
 
             async with self._backend.acquire() as conn:
                 for schema in schemas_to_query:
                     table = fq_table("async_operations", schema)
 
-                    # Bucket pending rows by the same predicates the claim query
-                    # filters on, so an operator can see why pending > 0 but
-                    # nothing is being claimed (orphaned batch_retain parents,
-                    # retry backoff, etc.).
+                    # Classify each pending row once, in claim-filter order. The
+                    # previous independent SUMs could double-count a row and did
+                    # not model consolidation's same-bank serialization, which
+                    # made blocked consolidations appear claimable (#3194).
                     # Use SUM(CASE WHEN ...) instead of COUNT(*) FILTER (WHERE ...)
                     # for Oracle compatibility — FILTER is PG-specific.
                     try:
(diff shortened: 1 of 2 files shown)


nicoloboschi:
it's too expensive as query and the stable worker id is well documented already

PR #3205, opened 2026-08-06
feat(claude-code,codex): tell existing users about the Coding Agents plugin

nicoloboschi:
Draft — the notice states that the server and token carry over, which is true for Codex only once #3203 merges, and none of it is on npm yet: 0.0.4 predates daemon mode and the endpoint carry-over. This should not merge before a release makes the claims real.

Both plugins keep working, but they are deprecated: development has moved to `@vectorize-io/hindsight-coding-agents`. Someone who installed one a year ago has no way to learn that — a changelog entry reaches nobody — so the session says it.

## What it says

```
The current Hindsight Claude Code plugin is DEPRECATED, replaced by the Coding Agents plugin.

One install now covers Claude Code, Codex, Cursor, Copilot, opencode, Kilo, Grok, Antigravity,
Devin and Cline, and they all share a single memory per repository, leveraging the Hindsight Knowledge Page (v0.9.x and onwards).

See the migration guide: https://hindsight.vectorize.io/sdks/integrations/coding-agents#migrating-from-the-per-agent-plugins
Your Hindsight server and token carry over automatically; past conversations are re-imported from
local transcripts. Silence this with "upgradeNotice": false in ~/.hindsight/claude-code.json
```

Codex gets the same text with its own plugin name and config path.

Both the page (HTTP 200) and the `#migrating-from-the-per-agent-plugins` anchor are verified live, so the one link a reader is given actually lands on the steps.

## How

The `SessionStart` hook emits `systemMessage`, the channel Claude Code shows to the **user** (`additionalContext` would only reach the model). Codex accepts Claude Code's hook output shape, so one design serves both — verified by running each hook and reading its JSON.

Emitted **before** the existing early returns: neither the memory settings nor whether the server is reachable changes the fact that the plugin is deprecated.

## Frequency

Every session start. A notice shown once is a notice that gets missed. What makes that acceptable is the opt-out: `"upgradeNotice": false` in `~/.hindsight/claude-code.json` / `codex.json` turns it off permanently, and the message's last line says so.

With no rate limit there is no state file, no timestamp parsing and none of their failure modes — what remains is a config check. The never-raise guarantee stays: a promotional message must never be why a session breaks.

## Tests

7 per plugin: shown on five consecutive calls, names which plugin is deprecated (both can be installed side by side), the migration deep-link, that it says DEPRECATED, the opt-out, that the opt-out is discoverable from the message, and that a hostile config object cannot make it raise.

Suites: **204 passed** (claude-code), **98 passed** (codex). `LINT_ALL_INTEGRATIONS=1 ./scripts/hooks/lint.sh` clean. Verified end to end that the hook emits it and that the opt-out silences it.

The module is duplicated rather than shared — these ship as independent packages with no common module path.

### DIFF
--- a/hindsight-integrations/claude-code/scripts/lib/config.py
+++ b/hindsight-integrations/claude-code/scripts/lib/config.py
@@ -39,6 +39,7 @@ DEFAULTS = {
     "retainMetadata": {},
     "recallAdditionalBanks": [],
     # Connection
+    "upgradeNotice": True,  # show the superseded-by-coding-agents notice (capped, see lib/upgrade_notice.py)
     "hindsightApiUrl": None,
     "hindsightApiToken": None,
     "apiPort": 9077,
--- a/hindsight-integrations/claude-code/scripts/lib/upgrade_notice.py
+++ b/hindsight-integrations/claude-code/scripts/lib/upgrade_notice.py
@@ -19,20 +19,18 @@ DOCS_URL = "https://hindsight.vectorize.io/sdks/integrations/coding-agents"
 MIGRATION_URL = f"{DOCS_URL}#migrating-from-the-per-agent-plugins"
 
 NOTICE = (
-    "⚠️  Hindsight: this Claude Code plugin is DEPRECATED, replaced by the Coding Agents plugin.\n"
+    "The current Hindsight Claude Code plugin is DEPRECATED, replaced by the Coding Agents "
+    "plugin.\n"
     "\n"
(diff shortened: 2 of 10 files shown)


nicoloboschi:
Folded into #3162 (commit 36e1f51b8) so the launch ships as one change — the docs announcing the Coding Agents plugin and the in-session notice pointing existing users at them land together rather than one arriving without the other.

The four wording-iteration commits were squashed into one; the code is identical, and both suites pass on the target branch (204 claude-code, 98 codex). Closing in favour of #3162.

PR #3229, opened 2026-08-07
fix(worker): release operations when a task is cancelled or its terminal write fails

### LINKED ISSUE #3228: Cancelled worker tasks leave async operations in 'processing' forever; the worker forgets them and only a restart recovers
## Summary

When a worker's in-flight task stops running without reaching its terminal-marking code, the claimed `async_operations` row stays in `status='processing'` **owned by a live worker** — forever. The worker itself has already forgotten the task (`_cleanup_task` removed it from `_active_tasks` and freed the slot), so nothing in-process will ever touch the row again, `recover_own_tasks` only runs at startup, and no dead-worker logic applies because the worker is alive. Any client polling that operation id waits indefinitely.

This is distinct from (and complementary to) the dead-worker discussions in #991/#992/#1441: those need cross-worker coordination and were declined. This one needs only the worker's **own local knowledge** — the same authority `recover_own_tasks` already exercises — applied when its own task is cancelled or its terminal write fails, not only at startup.

## Observed in production (0.8.6-slim, self-hosted k8s, single worker)

Pod `hindsight-api-96bf99f56-2q7ns` started `2026-08-05T12:23:18Z` and ran 39h without a restart. Two rows sat `processing` attributed to that live pod:

| operation_type | claimed_at | age | updated_at | retry_count |
|---|---|---|---|---|
| `graph_maintenance` | 2026-08-05 13:27:51 | 38.0h | identical to claimed_at | 0 |
| `retain` | 2026-08-05 12:51:05 | 38.6h | 2026-08-05 13:29:22 | 0 |

Proof they were forgotten rather than running: diffing the op ids in the worker's own `[WORKER_TASK]` log lines against `SELECT operation_id FROM async_operations WHERE status='processing' AND worker_id='<pod>'` yielded exactly these two (12 tracked in-flight vs 12+2 in the DB), and neither op id appeared once in 20 minutes of pod logs. The worker's slot accounting (`slots=9/12`) did not count them.

## Root cause — three paths break the same invariant

The invariant: *a task that stops running must not leave its operation in `processing` owned by this worker.*

1. **`asyncio.CancelledError` escapes `_execute_task_inner`** (`worker/poller.py` ~779-860). The handler chain catches `_WallTimeoutExceeded`, `DeferOperation`, `RetryTaskAt`, and `Exception`. `CancelledError` derives from `BaseException`, so it escapes all four; there is no `finally`. The done-callback (`_cleanup_task`) then removes the task from tracking without any DB write.

2. **`_mark_failed` is itself a DB write that can fail.** In the `except Exception` branch, `await self._mark_failed(...)` can raise (pool exhaustion, connection reset, statement timeout). That exception escapes uncaught with the row still `processing`. Given no shutdown occurred on the pod above, this is the likelier cause of the two observed rows.

3. **`shutdown_graceful` cancels and walks away** (`worker/poller.py` ~1238-1273). After the drain timeout it calls `info.bg_task.cancel()` on every in-flight task and returns immediately. Nothing releases those rows, and the event loop may close before any cancelled task could handle its own cancellation. Every restart with in-flight work over the grace period strands that work (until the *same* worker_id restarts again — under the default hostname-derived worker id in a container, never).

## Proposed fix

A guarded self-release, used at all three sites:

```sql
UPDATE async_operations
SET status = 'pending', worker_id = NULL, claimed_at = NULL,
    retry_count = COALESCE(retry_count, 0) + 1, updated_at = now()
WHERE operation_id = $1 AND status = 'processing' AND worker_id = $2  -- self
```

- In `_execute_task_inner`: an `except asyncio.CancelledError:` branch that releases and re-raises.
- Around `_mark_failed` in the `except Exception` branch: on failure of the terminal write, release instead of stranding.
- At the end of `shutdown_graceful`: bulk release of everything still claimed by this worker id (same shape as `recover_own_tasks`, run at shutdown instead of only at the next startup).

The `worker_id = <self>` guard means this is a no-op the moment a task has written its own terminal state, and it never touches another worker's rows — no coordination, no liveness inference, no duplicate-work race. The `retry_count` increment mirrors `recover_own_tasks` (#2675/#2834) so a task that reliably kills its worker still exhausts its retry budget.

PR with tests to follow.

## Environment

- hindsight-api 0.8.6-slim (defect confirmed present at current `main`, 59d3f078c)
- PostgreSQL 16 (CNPG), Kubernetes, single worker

russellbrenner:
## Summary

When a worker's in-flight task stops running without reaching its terminal-marking code, its `async_operations` row stays `status='processing'` owned by a **live** worker, forever. The worker has already forgotten it (`_cleanup_task` removed it from `_active_tasks` and freed the slot), `recover_own_tasks` only runs at startup, and no dead-worker handling applies because the worker is alive. Clients polling the operation wait indefinitely.

Observed in production (0.8.6-slim, k8s, single worker): a pod running 39h without restart held two rows `processing` for 38h — `retry_count=0`, absent from the worker's own `[WORKER_TASK]` in-flight logging, zero log mentions in 20 minutes, not counted in `slots=9/12`. Full evidence in #3228.

## The three paths

1. **`asyncio.CancelledError` escapes `_execute_task_inner`.** The handler chain catches `_WallTimeoutExceeded`, `DeferOperation`, `RetryTaskAt`, and `Exception`. `CancelledError` derives from `BaseException`, so it escapes all four; there is no `finally`.
2. **`_mark_failed` is itself a DB write.** If it raises inside the `except Exception` branch (pool exhaustion, connection reset, statement timeout), the row is stranded. Likely the cause of the observed rows, since no shutdown occurred on that pod.
3. **`shutdown_graceful` cancels and returns.** After the drain timeout it calls `bg_task.cancel()` on each in-flight task and returns immediately; the event loop may close before any cancelled task could reconcile its own row. Every rollout with in-flight work over the grace period strands it.

## The fix

A self-scoped release, applied at all three sites:

```sql
UPDATE async_operations
SET status = 'pending', worker_id = NULL, claimed_at = NULL,
    retry_count = COALESCE(retry_count, 0) + 1, updated_at = now()
WHERE operation_id = $1 AND status = 'processing' AND worker_id = $2  -- self
```

- `_execute_task_inner`: new `except asyncio.CancelledError:` branch that releases and re-raises (no metric — cancellation is not a terminal outcome).
- `except Exception`: `_mark_failed` wrapped so its own failure releases instead of stranding.
- `shutdown_graceful`: after the cancel loop, a bulk release of everything still claimed by this worker id across all schemas — the same shape as `recover_own_tasks`, run at shutdown instead of only at the next startup.

**Scope note:** this deliberately does *not* touch other workers' rows. I read the history on #992/#1441/#1510 and the line drawn there — no cross-worker coordination, no liveness inference. The guard `worker_id = <self>` means every write here uses only the worker's own local knowledge, the same authority `recover_own_tasks` already exercises at startup; it is a no-op the moment a task lands its own terminal state, so there is no duplicate-work race. It also answers "we need to figure out why workers are stuck" (#992) — these three paths are why.

The `retry_count` increment mirrors `recover_own_tasks` (#2675/#2834): a task that reliably kills its worker still exhausts its retry budget rather than being re-claimed forever.

Complementary to (not overlapping) #3092, which makes the drain timeout configurable; this PR makes the post-timeout cancellation stop stranding rows regardless of the timeout's value.

## Tests

`tests/test_worker_task_release.py`, three tests against real Postgres (pg0), all failing on `main` and passing with the fix:

- `test_cancelled_task_returns_operation_to_pending` — claim, block the executor, cancel the tracked `bg_task`; row must end `pending`/`worker_id NULL`/`retry_count 1`.
- `test_failed_terminal_write_returns_operation_to_pending` — executor raises, `_mark_failed` monkeypatched to raise; row must end `pending`, not stranded `processing`.
- `test_shutdown_graceful_releases_inflight_operations` — `shutdown_graceful(timeout=0.1)` with a blocked task in flight; row must end `pending`.

```
$ uv run pytest -n0 tests/test_worker_task_release.py
3 passed
# with the poller.py change stashed:
3 failed  (rows stay 'processing')
$ uv run pytest -n0 tests/test_worker.py
102 passed
```

Fixes #3228

### DIFF
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -586,6 +586,65 @@ class WorkerPoller:
                 )
                 await self._maybe_update_parent_operation(operation_id, schema, conn)
 
+    async def _release_if_still_claimed(self, operation_id: str, schema: str | None) -> bool:
+        """Return an operation this worker still owns to 'pending'.
+
+        Guarded on ``status = 'processing' AND worker_id = <self>``, so it is a
+        no-op the moment the task has written its own terminal state. That makes
+        it safe to call whenever a task stops running without reaching the
+        terminal-marking code: ``asyncio.CancelledError`` derives from
+        BaseException and so escapes every ``except Exception`` above, and
+        ``_mark_failed`` is itself a DB write that can raise. In both cases the
+        done-callback removes the task from ``_active_tasks``, so nothing
+        in-process remembers the row and ``recover_own_tasks`` — which runs only
+        at startup — never sees it either.
+
+        The retry_count increment matches recover_own_tasks: an interrupted task
+        counts against the retry budget, so a task that reliably kills its
+        worker cannot be re-claimed forever (#2675 / #2834).
+        """
+        table = fq_table("async_operations", schema)
+        async with self._backend.acquire() as conn:
+            result = await conn.execute(
+                f"""
+                UPDATE {table}
+                SET status = 'pending', worker_id = NULL, claimed_at = NULL,
+                    retry_count = COALESCE(retry_count, 0) + 1, updated_at = now()
+                WHERE operation_id = $1 AND status = 'processing' AND worker_id = $2
+                """,
+                operation_id,
+                self._worker_id,
+            )
+        return bool(_updated_row_count(result))
+
+    async def _release_all_claimed(self) -> int:
+        """Return every operation still claimed by this worker to 'pending'.
+
+        Runs across all configured schemas, like recover_own_tasks.
+        """
+        total = 0
+        for schema in await self._get_schemas():
+            table = fq_table("async_operations", schema)
+            try:
+                async with self._backend.acquire() as conn:
+                    result = await conn.execute(
+                        f"""
+                        UPDATE {table}
+                        SET status = 'pending', worker_id = NULL, claimed_at = NULL,
+                            retry_count = COALESCE(retry_count, 0) + 1, updated_at = now()
+                        WHERE status = 'processing' AND worker_id = $1
+                        """,
+                        self._worker_id,
+                    )
+                total += _updated_row_count(result)
+            except Exception:
+                schema_display = f'"{schema}"' if schema else str(schema)
+                logger.warning(
+                    f"Worker {self._worker_id} could not release claimed tasks for schema {schema_display}",
+                    exc_info=True,
+                )
+        return total
+
     async def _maybe_update_parent_operation(self, child_operation_id: str, schema: str | None, conn) -> None:
         """If this operation is a child of a batch_retain, update the parent status when all siblings are done.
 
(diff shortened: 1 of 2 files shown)


r266-tech (commented):
Please apply the existing max-retry handling to the shutdown bulk release.

The shutdown bulk release unconditionally returns every row owned by this worker to `pending` after incrementing `retry_count`. This bypasses the existing `recover_own_tasks` max-retry handling, while the claim queries do not filter on `retry_count`. Consequently, an operation that repeatedly outlives graceful-shutdown timeouts can be claimed again indefinitely instead of exhausting its retry budget, contrary to the stated retry-count intent.

Please apply the same max-retry threshold and terminal-failure/parent-propagation behavior used by `recover_own_tasks` when releasing shutdown-owned rows, and add coverage for an operation already at or above that threshold.

nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/3234

PR #3231, opened 2026-08-07
fix(worker): serialise graph_maintenance per bank when claiming

### LINKED ISSUE #3230: Concurrent graph_maintenance runs on the same bank convoy on row locks and starve the worker
## Summary

`graph_maintenance` operations have no per-bank serialisation at claim time. `consolidation` does — `claim_tasks` excludes banks with a consolidation already `processing` (`busy_bank_ids` / `bank_id != ALL(...)`) — but `graph_maintenance` is claimed through the generic shared-pool query, filtered only by `operation_type != 'consolidation'`.

Two concurrent `graph_maintenance` runs against the same bank produce no extra useful work: each run is the same bank-wide sweep (relink queue drain + `prune_orphan_entities` + `prune_stale_cooccurrences`), so the second immediately blocks on the row locks the first holds in `entity_cooccurrences` / `graph_maintenance_queue`. On a large graph each run takes tens of seconds, and enough of them arrive to occupy every free worker slot — all convoyed behind one another, starving other operation types.

## Observed in production (0.8.6-slim, self-hosted k8s, single worker, one active bank)

Bank of 94,921 memory units, ~46,000 `entities`, ~478,000 `entity_cooccurrences`; each graph_maintenance run ~60s.

Worst incident: **nine** concurrent `graph_maintenance` runs against the one bank formed a six-deep Postgres blocking chain (`pg_blocking_pids`), each holding a worker slot, and starved retain to zero completions for ~4 hours.

Still reproducible at time of filing — three concurrent same-bank runs in one poll tick, from the worker's own diagnostic:

```
[WORKER_TASK] [STUCK?] op=a718a279… type=graph_maintenance bank=omp age=401s stage=task.graph_maintenance
[WORKER_TASK] [STUCK?] op=8ac57745… type=graph_maintenance bank=omp age=401s stage=task.graph_maintenance
[WORKER_TASK] [STUCK?] op=beb6e836… type=graph_maintenance bank=omp age=339s stage=task.graph_maintenance
```

at which point the same bank had 4 `processing` + 22 `pending` graph_maintenance rows.

## Why submit-time dedup does not prevent this

`submit_async_graph_maintenance` passes `dedupe_by_bank=True`, but that check only inspects `status='pending'`. Once op 1 moves `pending → processing`, the next enqueue (there are seven enqueue sites: post-insert maintenance, document/unit/bulk deletes, failed edit, curation) is not deduplicated and becomes the next pending op. Repeat per poll tick while a 60s run is in flight and the queue grows without bound.

Note the 22 simultaneously-pending same-bank rows above also suggest the pending-dedup itself under-delivers under concurrency — worth a separate look, but out of scope here; even a perfect pending-dedup cannot prevent the claim-time convoy.

## Proposed fix

Give `graph_maintenance` the same treatment `consolidation` already has in `claim_tasks` (both dialects):

- exclude banks with a `graph_maintenance` already `processing` from claiming,
- claim at most one `graph_maintenance` per bank within a single batch (`DISTINCT ON (bank_id)` via a CTE, since the pending queue can hold many same-bank rows — see above).

No advisory locks (unusable behind connection poolers, #2817), no new mechanism — it mirrors the existing consolidation bank-serialisation structure in `claim_tasks`.

PR with tests to follow.

## Environment

- hindsight-api 0.8.6-slim (gap confirmed present at current `main`)
- PostgreSQL 16 (CNPG), Kubernetes, single worker, `graph_maintenance` unreserved (default 0 slots → always claimed via the shared pool)

russellbrenner:
## Summary

`graph_maintenance` has no per-bank serialisation at claim time. `consolidation` does — `claim_tasks` excludes banks with a consolidation already `processing` — but `graph_maintenance` is claimed via the generic shared-pool query with no bank awareness.

Two concurrent `graph_maintenance` runs against the same bank produce no extra work: each is the same bank-wide sweep (relink drain + `prune_orphan_entities` + `prune_stale_cooccurrences`), so the second immediately blocks on the `entity_cooccurrences` / `graph_maintenance_queue` row locks the first holds. On a large graph each run takes tens of seconds, and enough arrive (seven enqueue sites) to occupy every free worker slot, all convoyed behind one another.

Observed in production (0.8.6-slim, bank of 94,921 units, ~46k `entities`, ~478k `entity_cooccurrences`): **nine** concurrent runs against one bank formed a six-deep Postgres blocking chain and starved retain to zero completions for ~4 hours. Still reproducible at time of filing — three same-bank runs in one `[WORKER_TASK]` log tick, 4 `processing` + 22 `pending` for the one bank. Full evidence in #3230.

## The fix

Mirrors the existing consolidation bank-serialisation structure in `claim_tasks`, both dialects — no new mechanism:

1. Compute banks with a `graph_maintenance` already `processing` once per claim cycle.
2. Reserved pool: `graph_maintenance` claims go through a dedicated helper (like `_claim_consolidation_tasks`).
3. Shared pool: the generic query now excludes `graph_maintenance` (as it already excluded `consolidation`), and a new phase 2c claims it bank-serialised.
4. The helper claims **at most one per bank within a single batch** (`DISTINCT ON (bank_id)` in a CTE, locking the base table through a join — `DISTINCT ON` cannot be combined with `FOR UPDATE`; same idiom as `claim_graph_maintenance_batch`).

Point 4 is required, not defence-in-depth: submit-time `dedupe_by_bank` only inspects `'pending'` rows, so once a run is in flight every subsequent enqueue becomes a new pending row (22 same-bank pending rows observed live). Without it, one batch claims several for the same bank — the new test demonstrates 5 claimed on `main`.

Oracle: no `DISTINCT ON`, and row-limited `SELECT ... FOR UPDATE` raises ORA-02014, so the Oracle helper uses the two-step candidates-then-lock idiom `prune_terminal_operations` already uses, with `ROW_NUMBER() OVER (PARTITION BY bank_id ORDER BY created_at) = 1` and `FOR UPDATE OF ... SKIP LOCKED`. Implemented for dialect parity; not exercised locally (Oracle tests are gated behind the `oracle` marker and an `ORACLE_TEST_DSN` I don't have — existing repo convention).

No advisory locks: unusable behind connection poolers and managed PG (#2817).

## Disclosure

With this guard, a `graph_maintenance` row wedged in `processing` blocks that bank's graph maintenance until the row is released. #3229 fixes the known paths that create such rows (cancelled tasks, failed terminal writes, shutdown); these two changes are independent but strongest together.

## Tests

`tests/test_graph_maintenance_claim_serialization.py`, real Postgres (pg0):

- `test_second_run_not_claimed_while_bank_busy` — **fails on `main`** (row is claimed), passes with fix.
- `test_single_batch_claims_at_most_one_per_bank` — **fails on `main`** (`got 5`), passes with fix.
- `test_idle_bank_still_claimed_while_another_is_busy` — guards against over-blocking (passes both ways).
- `test_other_operation_types_unaffected` — a busy bank's `retain` is still claimed (passes both ways).

```
$ uv run pytest -n0 tests/test_graph_maintenance_claim_serialization.py
4 passed
$ uv run pytest -n0 tests/test_worker.py
104 passed
$ uv run pytest tests/test_graph_maintenance.py tests/test_graph_maintenance_deadlock.py \
    tests/test_graph_maintenance_queue_race.py tests/test_enqueue_graph_maintenance_ordered.py
32 passed
```

Fixes #3230

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
@@ -1290,6 +1290,78 @@ class OracleOps(DataAccessOps):
             *params,
         )
 
+    async def _claim_graph_maintenance_tasks(
+        self,
+        conn,
+        table: str,
+        claimed_ids: list,
+        limit: int,
+        busy_bank_ids: list[str],
+    ) -> list:
+        """Claim graph_maintenance tasks, at most one per bank.
+
+        Same rationale as the PostgreSQL dialect: two concurrent runs against
+        one bank just convoy on the entity_cooccurrences row locks the first
+        run holds, and ``bank_id != ALL(busy)`` alone is not enough because
+        submit-time dedupe_by_bank only inspects 'pending' rows, so a bank can
+        accumulate many pending rows while one is in flight.
+
+        Oracle has no DISTINCT ON and rejects a row-limited SELECT ... FOR
+        UPDATE (ORA-02014), so this uses the two-step idiom
+        prune_terminal_operations uses: pick the deterministic bounded
+        candidate set first (ROW_NUMBER() = 1 per bank), then lock only those
+        rows re-checking eligibility, FOR UPDATE OF ... SKIP LOCKED.
+        """
+        if limit <= 0:
+            return []
+
+        conditions = []
+        params: list = []
+        idx = 1
+        if busy_bank_ids:
+            conditions.append(f"bank_id != ALL(${idx}::text[])")
+            params.append(busy_bank_ids)
+            idx += 1
+        if claimed_ids:
+            conditions.append(f"operation_id != ALL(${idx}::uuid[])")
+            params.append(claimed_ids)
+            idx += 1
+        extra_clause = (" AND " + " AND ".join(conditions)) if conditions else ""
+        params.append(limit)
+
+        candidates = await conn.fetch(
+            f"""
+            SELECT operation_id FROM (
+                SELECT operation_id, created_at,
+                       ROW_NUMBER() OVER (PARTITION BY bank_id ORDER BY created_at) AS bank_rank
+                FROM {table}
+                WHERE status = 'pending'
+                  AND task_payload IS NOT NULL
+                  AND operation_type = 'graph_maintenance'
+                  AND (next_retry_at IS NULL OR next_retry_at <= NOW()){extra_clause}
+            )
+            WHERE bank_rank = 1
+            ORDER BY created_at
+            LIMIT ${idx}
+            """,
+            *params,
+        )
+        if not candidates:
+            return []
+
+        candidate_ids = [row["operation_id"] for row in candidates]
+        return await conn.fetch(
+            f"""
+            SELECT gm_task.operation_id, gm_task.operation_type, gm_task.task_payload, gm_task.retry_count
+            FROM {table} gm_task
+            WHERE gm_task.operation_id = ANY($1)
+              AND gm_task.status = 'pending'
+            ORDER BY gm_task.created_at
+            FOR UPDATE OF gm_task.operation_id SKIP LOCKED
+            """,
+            candidate_ids,
+        )
+
     async def claim_tasks(
         self,
         conn,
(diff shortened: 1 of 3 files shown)


nicoloboschi:
superseded by #3235

PR #3233, opened 2026-08-07
feat(memory-engine): add off-peak consolidation windows

### LINKED ISSUE #2762: [Feature Request] Off-peak / scheduled consolidation windows (run async consolidation only during a configurable time window)
## Summary

Allow `consolidation` to be restricted to a configurable daily/off-peak time window (e.g. run only 01:00–06:00), instead of firing whenever facts become eligible. Consolidation is the most LLM-heavy background job in Hindsight, and today it competes for the same model/compute resources as user-facing `retain`/`recall`/`reflect` during peak hours.

## Motivation

- **Cost savings** — many providers offer cheaper off-peak / batch tiers (e.g. Gemini FLEX #1000, OpenAI/Anthropic batch pricing, cheaper spot GPU for local models). Deferring consolidation to a low-traffic window lets operators point it at those cheaper resources.
- **Resource contention** — consolidation runs LLM calls per bank and can burst (especially after the 5-min reconcile sweep re-schedules stranded banks). During peak traffic this steals throughput/slots from interactive requests. An off-peak window smooths that out.
- **Async-friendly** — consolidation is already an async background operation; it has no latency SLA, so time-shifting it is safe.

## Current behavior (as of `main`)

- Auto-consolidation is enqueued right after `retain` when `enable_observations and enable_auto_consolidation` (`memory_engine.py`), and re-scheduled by `MaintenanceLoop._run_reconcile()` (default every 5 min).
- Worker claims consolidation tasks via `FOR UPDATE SKIP LOCKED`; the claim SQL already gates on `(next_retry_at IS NULL OR next_retry_at <= NOW())` (`ops_postgresql.py`).
- There is **no** notion of an allowed time window — a task runs as soon as a slot is free.

## Proposed approach

The primitives to do this cleanly already exist, so the change can be small and non-lossy:

1. **New config (static, server-level)** — e.g. `HINDSIGHT_API_CONSOLIDATION_WINDOW_START` / `HINDSIGHT_API_CONSOLIDATION_WINDOW_END` (or a cron expression + duration). Empty/unset ⇒ current always-on behavior (backward compatible).
2. **Window gate in the consolidation execution path** (`_handle_consolidation` in `memory_engine.py`): if the current time is outside the window, compute the next window-open time and raise `DeferOperation(exec_date=<next_open>, reason="off-peak consolidation window closed")`.
   - `_defer_operation` already sets `status='pending', next_retry_at=<exec_date>, worker_id=NULL` **without** bumping `retry_count` or writing an error — so backlog is *held, not dropped*, and drains automatically when the window opens (claim SQL `next_retry_at <= NOW()` passes).
3. **croniter precedent** — `MaintenanceLoop._run_scheduled_mm_refresh()` already uses `croniter` to evaluate cron due-ness in Python, so cron-based window config would follow an established pattern.

This keeps the trigger/enqueue logic untouched and reuses the existing defer → `next_retry_at` → claim machinery.

### Notes / considerations

- **Thundering herd at window open** — a large held backlog will all become claimable at once. The existing `WORKER_SLOT_RESERVATION_TYPES` (consolidation defaults to 2 slots) and `worker_consolidation_bank_priority` already cap concurrency, so this is bounded, but worth calling out.
- **Timezone** — window should be evaluated against a configurable TZ (or UTC by default) so operators can align it to their off-peak hours.
- **Insertion point** — the `validate_consolidate` hook in `extensions/operation_validator.py` is currently **defined but never called** anywhere in the codebase (unlike retain/recall/reflect/bank_write). Wiring it up would give a pluggable, extension-based place to implement window/quota policy instead of (or in addition to) hard-coded config.

### Alternatives considered

- **Cron-toggle `enable_auto_consolidation`** via MaintenanceLoop — only stops *new* enqueues, does not hold already-queued work, and the flag is per-bank so global scheduling is awkward.
- **Scheduled slot-reservation changes** (set consolidation slots to 0 at peak) — no built-in scheduler for slot changes, and slot=0 still re-claims/skips rather than cleanly deferring.

Happy to open a PR along the lines of approach (1)+(2) if the maintainers agree on the config shape (window start/end vs cron) and whether to route it through the `validate_consolidate` hook.

xiaoxiaoHe-E:
## Summary

Implements #2762 — restrict the consolidation background job (Hindsight's most LLM-heavy operation) to a configurable daily off-peak window (e.g. `01:00–06:00`), instead of firing whenever facts become eligible.

Tasks that land outside the window are **deferred**, not failed: the worker holds them as `pending` with `next_retry_at` set to the next window open, and the backlog drains automatically when the window reopens (claim SQL `next_retry_at <= NOW()`).

## Changes

- **New static config trio** (`config.py`):
  - `HINDSIGHT_API_CONSOLIDATION_WINDOW_START` / `HINDSIGHT_API_CONSOLIDATION_WINDOW_END` — `HH:MM` (24h) daily window boundaries; must be set together. `START > END` = cross-midnight window (e.g. `22:00–06:00`); `START == END` = always-open. Unset = always-on (previous behaviour).
  - `HINDSIGHT_API_CONSOLIDATION_WINDOW_TZ` — IANA timezone the boundaries are evaluated in; unknown names fall back to `UTC` with a warning.
- **`engine/consolidation_window.py`** (new) — pure, deterministic `next_window_open()` schedule arithmetic (normal / cross-midnight / DST-aware), unit-tested.
- **`_handle_consolidation` gate** — defers via the existing `DeferOperation` mechanism (no `retry_count` bump, no `error_message`) while the window is closed.
- **Wires the previously-dead `validate_consolidate` extension hook** into the same deferral path — it was defined on the validator interface but never called for consolidation. A rejection is now a deferral (retry-backoff), not an HTTP 403.
- **Sync task backend safeguard** — on `SyncTaskBackend` (tests / CLI / embedded single-process) there is no worker queue to re-claim a deferred task, so raising `DeferOperation` would 500 the HTTP caller and strand a `pending` op that blocks `banks_needing_consolidation()` from re-submitting the bank. There it logs and runs now (window closed) / skips (hook rejection) instead.

## Notes

- The manual sync path `run_consolidation()` (explicit user trigger) is intentionally **not** gated by the window — the window applies to the async background job.
- Window boundaries use wall-clock `replace()`; on DST spring-forward/fall-back transition days the boundary can shift by one hour (window times are typically in the early morning).

## Testing

- 21 unit tests: schedule arithmetic (`next_window_open`) + env parsing (`_parse_consolidation_window*`).
- 7 integration tests: worker deferral semantics (no retry bump / no error, `next_retry_at` = next window open), reconcile non-resubmission, extension rejection deferral, sync-backend run-now/skip behaviour.
- `ruff`, `ty`, and the repo pre-commit hooks (`lint.sh`, `generate-docs-skill.sh`, `check-unused.sh`) all pass.

Closes #2762


nicoloboschi:
see https://github.com/vectorize-io/hindsight/issues/2762#issuecomment-5353897765

PR #3258, opened 2026-08-08
Fix background seed harness attribution

### LINKED ISSUE #3247: coding-agents: background seed (deepen.js) always resolves {harness} to "opencode", misfiling survey + git history into phantom banks
## Summary

In `@vectorize-io/hindsight-coding-agents@0.1.0`, the background seed worker (`dist/deepen.js`) resolves `{harness}` to the hardcoded string `"opencode"` on every harness. With a `bankIdTemplate` containing `{harness}`, the codebase survey and git-history ingest are written to `opencode::<project>` while the session hooks correctly write to `<harness>::<project>`.

The result is a set of phantom banks that no session ever reads, populated with real (and LLM-billed) extraction work.

I hit this on a machine where **opencode is not installed at all** — only `install claude-code` was ever run.

## Cause

`deepen.js` has no harness of its own and falls through to the shared config loader's default:

```js
// dist/deepen.js
harness: raw.harness ?? "opencode"
```

That same default is bundled into all 29 entry points, but every hook entry point overrides it with its own literal (`claude-code`, `codex`, `cursor-cli`, …). `deepen.js` is the only one that doesn't — and the code that spawns it doesn't pass the flag either:

```js
// dist/claude-sessionstart-hook.js — startBackgroundSeed()
const child = spawnFn(
  "node",
  [enginePath, "--repo", repoDir, "--gitlog-limit", String(limit), ...]
);
```

`deepen.js` documents the flag in its own usage string, so the plumbing is simply not wired:

```
usage: node deepen.js --repo <path> [--bank <id>] [--harness <name>] [--conversations f.json] ...
```

Observed argv of the running processes confirms it:

```
$ ps -eo command= | grep deepen.js
node .../dist/deepen.js --repo /Users/me/dotfiles --gitlog-limit 300
node .../dist/deepen.js --repo .../ismar.ch --gitlog-limit 300
```

## Reproduction

1. Install for one harness only: `hindsight-coding-agents install claude-code --server self-hosted --api-url http://localhost:9077`
2. `~/.hindsight/coding-agent.json`:
   ```json
   {
     "serverMode": "self-hosted",
     "apiUrl": "http://localhost:9077",
     "dynamicBankId": true,
     "bankIdTemplate": "{harness}::{gitProject}",
     "resolveWorktrees": true,
     "codebaseSurvey": true,
     "gitIngest": "message"
   }
   ```
3. Start a Claude Code session in a git repo, wait for the background seed.
4. List banks.

**Expected:** everything under `claude-code::<project>`.

**Actual:** hooks write `claude-code::<project>`; the seed creates and fills `opencode::<project>`.

```
claude-code::dotfiles      <- session hooks (correct)
opencode::dotfiles         <- deepen.js survey + git history (wrong)
```

Document provenance on the misfiled rows:

```
retain_params: {"context": "git commit-message history (last 300) for dotfiles"}
```

## Impact

1. **Survey and git-history knowledge is unreachable.** Sessions query `<harness>::<project>`; the seed's output lands elsewhere. The feature appears to run successfully and produces nothing usable.
2. **Knowledge Pages stay empty.** The pages *are* created in the correct bank, but their source material is in the other one, so every one of them refreshes to `"I don't have information."`:
   ```
   Component map                = 25 chars  ("I don't have information.")
   Conventions and patterns     = 25
   Core concepts                = 25
   Initiatives and enhancements = 25
   Key decisions and rationale  = 25
   ```
3. **It fans out.** The seed runs per repo, so one machine accumulated 7 phantom banks and 357 memory units across repos.
4. **It bills.** Extraction runs over 300 commits per repo against the configured LLM provider.
5. **It is silent.** The phantom banks are well-formed, receive pinned mental models, and pass health checks. Nothing surfaces the split.

## Suggested fix

Pass the harness through from the spawner:

```js
[enginePath, "--repo", repoDir, "--harness", HARNESS, "--gitlog-limit", String(limit), ...]
```

Separately, `?? "opencode"` is a surprising default for a shared config loader. Failing closed (require the harness, or derive it from the entry point) would have made this loud instead of silent.

## Workaround

`raw` is the parsed `coding-agent.json` — the same object that supplies `apiUrl`, `serverMode` and `bankIdTemplate` — so setting the key explicitly defeats the default:

```json
{ "harness": "claude-code" }
```

Verified: after adding it, the seed writes to `claude-code::<project>`.

## Environment

- `@vectorize-io/hindsight-coding-agents` 0.1.0
- `hindsight-embed` / `hindsight-api` 0.9.0, self-hosted daemon on `localhost:9077`
- macOS 15 (Darwin 25.6.0), Node from Homebrew
- Harnesses installed: claude-code only. opencode is not installed on this machine.

koriyoshi2041:
## Problem

Session-start seeding omits the resolved harness when it starts the background seed. The seed process then falls back to `opencode`, so git-history documents can be written to a different bank from the same session's survey documents.

Closes #3247.

## Fix

Pass the resolved session harness into `startSeed` and update the injected callback type. Add a regression case using a non-default harness, while keeping the existing session-start assertions exact.

## Test

- `npm test -- --run src/core/session-start.test.ts` (19 passed)
- `npm run build`
- `./scripts/hooks/lint.sh`
- `git diff --check`

## Risk

Low. This only forwards an existing option at the call site; the background seed implementation already accepts the harness and serializes it into the child-process arguments.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/session-start.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/session-start.test.ts
@@ -20,7 +20,7 @@ describe("buildSessionStartContext", () => {
       startSeed,
       startSurvey,
     });
-    expect(startSeed).toHaveBeenCalledWith("/repo/dir", { limit: 300 });
+    expect(startSeed).toHaveBeenCalledWith("/repo/dir", { harness: "claude-code", limit: 300 });
     expect(startSurvey).toHaveBeenCalledWith("/repo/dir", {
       harness: "claude-code",
       model: "haiku",
@@ -53,9 +53,26 @@ describe("buildSessionStartContext", () => {
       startSeed,
       startSurvey,
     });
-    expect(startSeed).toHaveBeenCalledWith("/repo/dir", { limit: 300 });
+    expect(startSeed).toHaveBeenCalledWith("/repo/dir", { harness: "claude-code", limit: 300 });
     expect(startSurvey).not.toHaveBeenCalled();
     expect(out.systemMessage).toContain("is learning");
+  });
+
+  it("passes a non-default harness to the background seed", async () => {
+    const client = { listDocumentIds: async () => new Set<string>(), listPages: listPagesOk };
+    const startSeed = vi.fn();
+
+    await buildSessionStartContext({
+      cwd: "/repo/dir",
+      bankId: "bank-1",
+      cfg: resolveConfig({ codebaseSurvey: false }),
+      client,
+      harness: "codex",
+      hasGit: () => true,
+      startSeed,
+    });
+
+    expect(startSeed).toHaveBeenCalledWith("/repo/dir", { harness: "codex", limit: 300 });
   });
 
   it("non-git dir -> no seed, listDocumentIds not called, roster preamble only (no learning note)", async () => {
@@ -123,7 +140,7 @@ describe("buildSessionStartContext", () => {
     });
     // The live bank is consulted, and an empty bank seeds — no client-side flag can contradict it.
     expect(called).toBe(true);
-    expect(startSeed).toHaveBeenCalledWith("/repo/dir", { limit: 300 });
+    expect(startSeed).toHaveBeenCalledWith("/repo/dir", { harness: "claude-code", limit: 300 });
     expect(out.systemMessage).toContain("is learning");
   });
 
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Superseded by #3266, which includes this harness-forwarding fix and the broader config fallback. Closing this duplicate.

PR #3259, opened 2026-08-08
Reject implausible temporal year matches

### LINKED ISSUE #3250: [BUG] Bare integers score as the strongest date signal, producing implausible-year temporal constraints (same class as #2768)
## Summary

`_date_match_score()` in `hindsight_api/engine/query_analyzer.py` awards `+100` for any digit in a matched span, which is the largest single signal the function can assign. A bare integer anywhere in a query therefore outscores every other candidate, wins the `max()`, and becomes a single-day `TemporalConstraint` in whatever year the integer names.

The constraint is non-null, so nothing downstream can tell that extraction failed. The temporal retrieval arm simply returns zero candidates while still paying its latency cost.

This is the same failure mode the comment above `_date_match_score` warns about for #2768:

> worse than none, because the constraint is non-null so nothing downstream can tell extraction failed

#2768 fixed bare *words* scoring false positives. Bare *numbers* still score highest.

**Version:** hindsight-api 0.8.6

## Reproduction

```python
from datetime import datetime
from hindsight_api.engine.query_analyzer import DateparserQueryAnalyzer

a = DateparserQueryAnalyzer(); a.load()
ref = datetime(2026, 8, 7)
print(a.analyze("is hindsight listening on port 9077", ref).temporal_constraint)
# TemporalConstraint(start_date=9077-08-07 00:00:00, end_date=9077-08-07 23:59:59)
```

Observed in a real recall log:

```
[2] Parallel retrieval (3 fact_types): semantic=472, bm25=22, graph=262,
    temporal_extraction=0.026s, temporal=0 | temporal_range=9077-08-07 to 9077-08-07
```

## Impact

Higher than it looks, because the trigger is common. Port numbers, ticket and PR numbers, job numbers, dollar amounts, and model numbers all appear routinely in developer and assistant queries. Each one silently disables the temporal arm for that recall.

## Suggested fix

Gate on year plausibility. Applying it as a filter on `scored_results`, rather than inside `_date_match_score` or after `max()`, means a genuine date elsewhere in the query can still win instead of the whole query going non-temporal:

```python
scored_results = [entry for entry in scored_results if entry[0] > 0]

_ref_year = reference_date.year
scored_results = [
    entry for entry in scored_results
    if _ref_year - 120 <= entry[2].year <= _ref_year + 20
]

if not scored_results:
    return QueryAnalysis(temporal_constraint=None)
```

Verified locally against 0.8.6:

| query | before | after |
|---|---|---|
| `is hindsight listening on port 9077` | 9077-08-07 | None |
| `check job 4417 for Castle Pines` | 4417-08-07 | None |
| `what did we decide yesterday` | 2026-08-06 | 2026-08-06 |
| `the meeting on 2026-06-10` | 2026-06-10 | 2026-06-10 |
| `port 9077 and also last Tuesday` | 9077-08-07 | 2026-08-04 |

The last row is the reason for filtering before `max()` rather than rejecting after it.

A tighter alternative, worth considering instead of or alongside the year gate: stop awarding the full `+100` to a span whose only content is a bare integer with no separator, month word, or period word beside it. The year gate is broader and also catches typoed or out-of-range dates, so it was the one implemented here.

Happy to open a PR if the approach looks right.

koriyoshi2041:
## Problem

Dateparser treats bare integers such as port and ticket numbers as years. Because digit-bearing matches receive the highest temporal score, a query like `port 9077` produces a non-null year-9077 constraint and silently empties the temporal retrieval arm.

Closes #3250.

## Fix

Filter parsed candidates to a reference-relative plausibility window before selecting the strongest match. Filtering before `max()` preserves a genuine relative or explicit date elsewhere in the same query.

## Test

- `uv run pytest tests/test_query_analyzer.py -q` (429 passed)
- `./scripts/hooks/lint.sh`
- repository pre-commit hooks
- `git diff --check`

## Risk

Low to moderate. Dates more than 120 years in the past or 20 years in the future are no longer inferred by the fallback dateparser path. Explicit period parsing remains unchanged, and the bounds are clamped for low/high reference years.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/query_analyzer.py
+++ b/hindsight-api-slim/hindsight_api/engine/query_analyzer.py
@@ -378,6 +378,14 @@ class DateparserQueryAnalyzer(QueryAnalyzer):
         ]
         scored_results = [entry for entry in scored_results if entry[0] > 0]
 
+        # Bare integers (ports, ticket IDs, model numbers) are parsed as years
+        # and otherwise outrank every genuine relative-date match. Filter them
+        # before ranking so a plausible date elsewhere in the query can still
+        # win instead of silently disabling temporal retrieval. See issue #3250.
+        earliest_year = max(1, reference_date.year - 120)
+        latest_year = min(9999, reference_date.year + 20)
+        scored_results = [entry for entry in scored_results if earliest_year <= entry[2].year <= latest_year]
+
         if not scored_results:
             return QueryAnalysis(temporal_constraint=None)
 
--- a/hindsight-api-slim/hindsight_api/engine/query_analyzer.py
+++ b/hindsight-api-slim/hindsight_api/engine/query_analyzer.py
@@ -90,6 +90,7 @@ _PERIOD_WORDS = {
     "noon",
     "midnight",
 }
+_BARE_YEAR_PADDING_WORDS = {"and", "at", "from", "in", "on", "the"}
 
 
 # Every token _date_match_score can award points for, in one alternation.
@@ -150,6 +151,26 @@ def _date_match_score(text: str) -> int:
     if token_set & _PERIOD_WORDS:
         score += 20
     return score
+
+
+def _is_implausible_bare_year_match(text: str, parsed_date: datetime, reference_date: datetime) -> bool:
+    """Return whether dateparser treated an isolated identifier as a year.
+
+    dateparser may include a neighbouring preposition or conjunction in the
+    matched span (for example, ``"9077 and"``), so checking ``isdigit()`` is
+    not sufficient. Preserve explicit calendar expressions such as
+    ``1890-03-05`` and ``March 1890`` even when their year is far from the
+    reference date.
+    """
+    earliest_year = max(1, reference_date.year - 120)
+    latest_year = min(9999, reference_date.year + 20)
+    if earliest_year <= parsed_date.year <= latest_year:
+        return False
+
+    tokens = _TOKEN_RE.findall(text.lower())
+    year_tokens = [token for token in tokens if len(token) == 4 and token.isdigit()]
+    padding_tokens = [token for token in tokens if token not in year_tokens]
+    return len(year_tokens) == 1 and all(token in _BARE_YEAR_PADDING_WORDS for token in padding_tokens)
 
 
 class TemporalConstraint(BaseModel):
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
Both cases from #3250 do go away at 29ac311a: `port 9077` and `job 4417` return `None`.

The window is applied to every surviving match rather than to the bare-integer class specifically, so a fully written-out date outside it loses its constraint too. Reference date 2026-08-07, main vs this branch:

```
notes from 1890-03-05                 1890-03-05  ->  None
the roadmap milestone on 2050-01-15   2050-01-15  ->  None
the mortgage matures on 2047-01-15    2047-01-15  ->  None
the mortgage matures on 2046-01-15    2046-01-15  ->  2046-01-15
```

So the future edge sits exactly at `reference_date.year + 20`. Twenty years is inside the range of dates a user might genuinely have written down (a mortgage maturity, a long-dated contract), and unlike a port number those spans carry a day and a month.

The obvious narrowing does not work, which took a run to establish, so it seems worth recording here. Carrying the matched span into the tuple and applying the window only when `span.strip().isdigit()` does keep `1890-03-05` and `2050-01-15`, and still drops the plain `9077` from the issue's query. It fails your own `test_query_analyzer_keeps_plausible_date_with_implausible_number`: for `port 9077 and also last Tuesday`, `search_dates` returns the span as `'9077 and'` rather than `'9077'`, so the digit test misses it and 9077 wins the ranking again. The span boundary moves with the neighbouring word, so anything keyed on the span text alone is fragile in the same way.

Is dropping explicit out-of-window dates intended? If it is, a line saying so in the comment would be worth having, since the comment currently describes the bare-integer case only.

Checked at 29ac311a in a clean python:3.12-slim container with an editable install of hindsight-api-slim. `tests/test_query_analyzer.py` is 429 passed on the branch, and 428 passed / 1 failed with the isdigit variant above.

koriyoshi2041:
Good catch. I narrowed the window to isolated year matches, including spans where dateparser absorbs padding such as `9077 and`. Full dates and month/year expressions outside the window now remain valid.

Added regressions for `1890-03-05`, `2050-01-15`, and `March 1890`; the complete query-analyzer suite passes (432 tests). Updated in 6d426a0f1.

nicoloboschi:
superseded and already in main

PR #3260, opened 2026-08-08
fix(api): return 400 for malformed UUID on GET /entities/{id} and /memories/{id}/history

Alan5168:
Two read endpoints pass their path parameter straight to `uuid.UUID()` without a try/except, so a malformed id raises a bare `ValueError` from the stdlib. The HTTP handlers catch `OperationValidationError` and `Exception` but not `ValueError`, so it falls through to the generic 500 path.

A typo or copy-paste error in the id — a normal user mistake on a public endpoint — surfaces as a 500 ("badly formed hexadecimal UUID string") instead of a clean 400. This is the same gap #3062 fixed for `recall_async` and #906 fixed for `get_memory_unit` / `update_memory_unit` / `delete_memory_units`.

**`get_entity`** (memory_engine.py) — `uuid.UUID(entity_id)` bare in the `conn.fetchrow` args.

**`get_observation_history`** (memory_engine.py) — `uuid.UUID(memory_id)` bare at two call sites.

The fix wraps both in the same `try/except ValueError` the already-validated siblings use, and adds an `except ValueError -> 400` branch to the two HTTP handlers (`api_get_entity`, `api_get_observation_history`), matching `api_get_memory`.

Tests mirror `test_delete_memory_units_validation.py` — a stub engine that bypasses `__init__` so the validation branch is exercised without a database. Both tests fail on unfixed code (bare 'badly formed hexadecimal UUID string') and pass after the wrap.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -4391,6 +4391,8 @@ def _register_routes(app: FastAPI):
             return data
         except OperationValidationError as e:
             raise HTTPException(status_code=e.status_code, detail=e.reason)
+        except ValueError as e:
+            raise HTTPException(status_code=400, detail=str(e))
         except (AuthenticationError, HTTPException):
             raise
         except Exception as e:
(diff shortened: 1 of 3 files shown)


koriyoshi2041 (approved):
Verified the focused regression tests locally: both malformed-ID paths reject before authentication/database access (`2 passed`), Ruff passes on the touched files, and the diff is clean. The validation placement also matches the existing memory-unit pattern, so the HTTP handlers can consistently return 400 instead of leaking the stdlib UUID error as a 500.

nicoloboschi (approved):

PR #3261, opened 2026-08-08
Pin embed to the metapackage release

### LINKED ISSUE #3252: pip install --upgrade hindsight-all leaves hindsight-embed at old version (silent mixed install)
### Bug Description

Upgrading via hindsight-all==0.9.0 pulled hindsight-api-slim 0.9.0 but left hindsight-embed at 0.8.2, with no warning.

### Steps to Reproduce

  1. Have hindsight-all 0.8.2 installed (all components at 0.8.2)
  2. Run pip install --upgrade "hindsight-all==0.9.0"
  3. pip list | grep hindsight

### Expected Behavior

hindsight-all pins its components to matching versions, so upgrading the metapackage upgrades all components together (or the release notes call out that hindsight-embed must be upgraded separately).

### Actual Behavior

Mixed-version install: hindsight-all 0.9.0, hindsight-api-slim 0.9.0, hindsight-embed 0.8.2. Workaround: pip install --upgrade "hindsight-embed==0.9.0".

### Version
0.9.0 (hindsight-all / hindsight-api-slim / hindsight-embed; hermes-agent 0.19.0)

### LLM Provider
OpenAI-compatible

koriyoshi2041:
## Summary

- pin `hindsight-embed` to the matching `hindsight-all` / `hindsight-all-slim` release
- teach the release script to advance both embed pins with each version bump

This prevents `pip install --upgrade hindsight-all` from keeping an older embed package that still satisfies the previous broad lower bound.

Closes #3252.

## Tests

- `bash -n scripts/release.sh`
- `uv lock --check`
- built both metapackage wheels and verified their metadata requires `hindsight-embed==0.9.0`
- exercised the release-script substitution against both metapackage files with a synthetic next version
- `git diff --check`

## Risk

Low. The change only tightens metapackage dependency metadata and keeps that pin synchronized during future releases. Direct `hindsight-embed` installs remain independently selectable.

### DIFF
--- a/hindsight-all-slim/pyproject.toml
+++ b/hindsight-all-slim/pyproject.toml
@@ -12,7 +12,7 @@ requires-python = ">=3.11"
 dependencies = [
     "hindsight-api-slim==0.9.0",
     "hindsight-client>=0.0.7",
-    "hindsight-embed>=0.1.0",
+    "hindsight-embed==0.9.0",
 ]
 
 [tool.uv.sources]
--- a/hindsight-all/pyproject.toml
+++ b/hindsight-all/pyproject.toml
@@ -12,7 +12,7 @@ requires-python = ">=3.11"
 dependencies = [
     "hindsight-api-slim[all]==0.9.0",
     "hindsight-client>=0.0.7",
-    "hindsight-embed>=0.1.0",
+    "hindsight-embed==0.9.0",
 ]
 
 [tool.uv.sources]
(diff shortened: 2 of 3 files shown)


nicoloboschi (approved):
Correct and minimal. embed is cut in lockstep with the core release (already in PYTHON_PACKAGES + __version__ in release.sh), current version is 0.9.0, embed has no api-slim dep so `==` can't conflict, and grep confirms these are the only two embed declarations. release.sh sed handles both == and >= and touches only the dependency line. LGTM.

PR #3263, opened 2026-08-08
fix(ollama): support gpt-oss thinking parameter

### LINKED ISSUE #3246: retain: every fact extraction fails on Ollama + gpt-oss — hardcoded `think: False` yields empty content under `format`
## Summary

On `hindsight-api==0.9.0` with `LLM_PROVIDER=ollama` and `LLM_MODEL=gpt-oss:20b`, **every** `retain` fails:

```
Fact extraction failed: 1/1 chunks failed.
First failures: chunk 0: JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Server log shows the retry loop exhausting, with the response body empty:

```
WARNING - Ollama JSON parse error (attempt 1/4): Expecting value: line 1 column 1 (char 0)
  Model: ollama/gpt-oss:20b
  Content length: 0 chars
  Content preview: '<empty>'
```

`reflect` and `recall` work fine — only the structured-output path is affected.

## Root cause

`_call_ollama_native()` in `engine/providers/openai_compatible_llm.py` hardcodes `"think": False` in the payload (line 1466 in 0.9.0):

```python
payload: dict[str, Any] = {
    "model": self.model,
    "messages": messages,
    "stream": False,
    "think": False,  # Disable thinking for reasoning models (qwen3.5, etc.)
}
```

That flag came from #1099, which fixed qwen3.5-style models routing their whole
response to `thinking` and leaving `message.content` empty. #1111 then asked for
a regression test so it is never dropped. Both are correct for qwen3.5.

But gpt-oss is a harmony model, and Ollama accepts only `"low" | "medium" | "high"`
for its `think` field ([Ollama docs — Thinking](https://docs.ollama.com/capabilities/thinking)).
A boolean `think` combined with `format` (schema enforcement) makes Ollama return
an **empty** `message.content`, so `json.loads()` fails on an empty string.

#1099 states the flag is "safe for non-reasoning models, so gemma3, phi4-mini,
etc. are unaffected" — that holds, but harmony models were never covered. They
are neither "non-reasoning" nor boolean-`think` compatible.

## Reproduction

Ollama 0.32.6, macOS arm64, `gpt-oss:20b`. Same prompt and schema, only `think` varies:

```bash
SCHEMA='"format":{"type":"object","properties":{"facts":{"type":"array","items":{"type":"string"}}},"required":["facts"]}'
MSG='"messages":[{"role":"user","content":"Extract the facts: Andre lives in Curitiba and trains triathlon."}],"stream":false'

# A) what hindsight sends today -> BROKEN
curl -s localhost:11434/api/chat -d "{\"model\":\"gpt-oss:20b\",$MSG,\"think\":false,$SCHEMA}"
# B) think omitted -> works
curl -s localhost:11434/api/chat -d "{\"model\":\"gpt-oss:20b\",$MSG,$SCHEMA}"
# C) think:"low" -> works, and is by far the cheapest
curl -s localhost:11434/api/chat -d "{\"model\":\"gpt-oss:20b\",$MSG,\"think\":\"low\",$SCHEMA}"
```

Measured (identical prompt and schema):

| `think` | `message.content` | thinking chars | eval_count | duration |
|---|---|---|---|---|
| `false` (current) | **0 chars** | 0 | 8 | — |
| omitted | valid JSON | 1384 | 48 | 13.4 s |
| `"low"` | valid JSON | 19 | 26 | **2.2 s** |
| `"high"` | valid JSON | 1143 | 44 | 11.3 s |

Note that `"low"` is not merely a workaround — it is ~6x faster than dropping the
flag, because it suppresses the reasoning trace that would otherwise be generated
and discarded. On a bulk ingest that difference dominates total runtime.

## Suggested fix

Keep `False` for every other model; send a level for harmony models:

```python
"think": "low" if "gpt-oss" in self.model.lower() else False,
```

This preserves #1099 (qwen3.5, deepseek-r1 keep `False`) and satisfies #1111,
since the flag is still always present. A more general form would map any
harmony-format model to a level, ideally with the level configurable.

Verified locally with that one-line change: retain succeeds (10.1 s for a
1-chunk document), recall returns the extracted facts, and reflect performs real
tool calls (`search_observations`) rather than degrading to tool-less synthesis.

## Related

- #1099 — added the flag (merged 2026-04-16), verified against `qwen3.5:2b` only
- #1111 — regression test request for the flag
- #2222, #3048 — other gpt-oss/harmony breakage in `reflect`
- ollama/ollama#11751 — `--think=false` not honored for gpt-oss:20b
- ollama/ollama#15260 — `think=false` breaks `format` for gemma4 (same mechanism)
- plastic-labs/honcho#716 — same class of bug in another agent-memory project

## Environment

- `hindsight-api` 0.9.0 (pip, Python 3.13.14)
- Ollama 0.32.6, `gpt-oss:20b`
- macOS 26.6 arm64
- Embeddings: ONNX `BAAI/bge-m3` (1024) · Reranker: local · DB: pg0

goutamsidharth287:
## Summary

Fix Ollama native structured-output requests for gpt-oss models.

Previously, `think` was hardcoded to `False`, which causes structured
fact extraction to fail with gpt-oss models.

This change uses `think="low"` for gpt-oss models while preserving
the existing `think=False` behavior for other Ollama reasoning models.

## Testing

- Added regression coverage for the Ollama native API path.
- Verifies the `think` parameter for both gpt-oss and non-gpt-oss models.
- Verifies structured-output format is still included.

Fixes #3246

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -1463,7 +1463,7 @@ class OpenAICompatibleLLM(LLMInterface):
             "model": self.model,
             "messages": messages,
             "stream": False,
-            "think": False,  # Disable thinking for reasoning models (qwen3.5, etc.)
+            "think": "low" if "gpt-oss" in self.model.lower() else False,  
         }
 
         # Add schema as format parameter for structured output
--- a/hindsight-api-slim/tests/test_ollama_native_think.py
+++ b/hindsight-api-slim/tests/test_ollama_native_think.py
@@ -1,0 +1,85 @@
+"""
+Regression tests for Ollama native API think parameter handling.
+
+The gpt-oss models require a string thinking level such as "low",
+while other Ollama reasoning models such as qwen3.5 continue to
+use think=False.
+"""
+
+import json
+from unittest.mock import AsyncMock, patch
+
+import httpx
+import pytest
+from pydantic import BaseModel
+
+from hindsight_api.engine.providers.openai_compatible_llm import OpenAICompatibleLLM
+
+
+class _SampleOutput(BaseModel):
+    summary: str
+
+
+def _make_ollama_llm(model: str) -> OpenAICompatibleLLM:
+    return OpenAICompatibleLLM(
+        provider="ollama",
+        api_key="",
+        base_url="http://localhost:11434/v1",
+        model=model,
+    )
+
+
+def _mock_ollama_response(content: dict) -> httpx.Response:
+    body = {
+        "model": "test-model",
+        "message": {
+            "role": "assistant",
+            "content": json.dumps(content),
+        },
+        "done": True,
+    }
+    return httpx.Response(200, json=body)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("model", "expected_think"),
+    [
+        ("gpt-oss:20b", "low"),
+        ("qwen3.5:2b", False),
+    ],
+)
+async def test_ollama_native_think_parameter(model, expected_think):
+    """Use a thinking level for gpt-oss while preserving False for other models."""
+
+    llm = _make_ollama_llm(model)
+    captured_payloads: list[dict] = []
+
+    async def _capture_post(url, *, json=None, **kwargs):
+        captured_payloads.append(json)
+        return _mock_ollama_response({"summary": "test"})
+
+    with patch(
+        "httpx.AsyncClient.post",
+        new_callable=lambda: lambda: AsyncMock(side_effect=_capture_post),
+    ):
+        await llm._call_ollama_native(
+            messages=[{"role": "user", "content": "hello"}],
+            response_format=_SampleOutput,
+            max_completion_tokens=512,
+            temperature=0.1,
+            max_retries=0,
+            initial_backoff=1.0,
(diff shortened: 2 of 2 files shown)


nicoloboschi:
Thanks for catching #3246! I opened #3344 as an alternative approach.

Instead of the model-name heuristic (`"gpt-oss" in model` → `think="low"`), it makes `think` configurable through the existing `extra_body` surface — set `HINDSIGHT_API_LLM_EXTRA_BODY='{"think": "low"}'`. This avoids guessing the thinking level, gives an opt-out, and keeps model-specific behavior out of the transport layer (it also lets any future native field / `options` param be set without a code change).

Proposing #3344 supersede this one — but the fix is entirely thanks to your diagnosis here.

handnewb (approved):
LGTM! Ollama recently added native thinking/reasoning support — gpt-oss parameter enables this. 1-line change in openai_compatible_llm.py + comprehensive test. Good catch!

nicoloboschi:
Thanks — the gpt-oss failure was real, and #3344 fixed it generically: `think` is now configurable through `extra_body` on the native Ollama path, so gpt-oss works with `extra_body: {"think": "low"}` while other models keep `think: false` by default. The code comment in `openai_compatible_llm.py` names that exact configuration.

The generic route was preferred over a model-name check because name-substring gating on model ids has bitten us before (#3449).

Issue #1558, opened 2026-05-09
Memory Tags added when Entities are not during retain


garethjcooper:
### Bug Description

When there are multiple entities being extracted for a memory, the tags will be added correctly, although the entities will only contain a subset of the tags.

### Steps to Reproduce

Configure memory entities as:
Label: use, multivalue {use-001, use-002 etc}, tag=true
retain_extraction_mode=verbose, entities_allow_free_form=false.

insert into document content to be retained:
```
Blah blah system a blah blah .....

## Section title about something (use-001)

##  Section title about something (use-001, use-002)

etc
```

Retain document.

### Expected Behavior

For the retained memory I would expect to see
```
ENTITIES:
use:use-001
use:use-002

TAGS:
# use:use-001
# use:use-002
``` 

### Actual Behavior

For the retained memory I see

``` 
ENTITIES:
use:use-001

TAGS:
# use:use-001
# use:use-002
``` 

<img width="464" height="764" alt="Image" src="https://github.com/user-attachments/assets/7d1e70f0-1729-4b7f-8477-4458f0b72df6" />

### Version

0.6

### LLM Provider

Ollama

Issue #3276, opened 2026-08-08
Extraction mints tag-shaped category labels (domain:lens) as entities, poisoning entity-based scoping


JoshFunnell:
### Summary

Entity extraction mints tag-shaped **category labels** as entities — measured on a production bank: `domain:lens`, `domain:host`, `domain:memory`. These are not entities; they behave like broad tags, and they measurably poison entity-based workflows: in an entity-scoped retrieval experiment, a greedy scope selector picked `domain:lens` first and plateaued at 61.5% precision; excluding category-shaped names, the same selector reached 94.0% precision on the same corpus.

### Fix implemented

Candidate names matching a conservative tag shape (`^[a-z][a-z0-9_-]{1,15}:[a-z][a-z0-9_-]{1,24}$` after whitespace normalization, checked on a lowercased copy) are skipped at resolution intake with a debug log. The shape deliberately does not match URLs (`//`), Windows paths, times (digits before the colon), or names with spaces/dots. One considered exemption: **configured entity labels** (the `key:value`-shaped label taxonomy) are deliberately colon-shaped and are exempted via the existing label lookup — a genuinely-unwanted artifact colliding with a configured label key would survive, which seemed the right trade. Fix PR: #3277 (tests assert the non-matching names first; ships together with the whitespace-normalization fix, same intake point).

nicoloboschi:
Before we take a fix here, one clarifying question, because the answer points at two very different changes.

**Were `domain:lens` / `domain:host` / `domain:memory` values of a label group you had configured in that bank's `entity_labels`, or names free-form extraction invented on its own?**

The issue reads as the second (extraction "mints" them), and #3277 exempts configured labels from its skip — so I've been reviewing it that way. But if they *were* configured labels, the ask is really "let a label group be a tag and not an entity", which is a different and, I think, better change.

**If they were configured labels:** there's a genuine gap. `tag: true` on a label group (`_inject_label_tags`, `fact_extraction.py:2753`) copies the values into the unit's tags, but they still stay in `fact.entities` and are still minted as entities — so today a label group is always *also* an entity, with no way to opt out. A per-group `index_as_entity: false` (default `true`, so nothing changes for existing banks) would make that explicit and per-dimension: the value stays filterable as a tag, never reaches the entity registry, and nothing has to guess from the shape of the name. Would that solve your case?

**If they were free-form artifacts:** the regex worries me as a global default, for three reasons.

[...]

PR #3277, opened 2026-08-08
fix(engine): skip tag-shaped candidate entity names at intake

JoshFunnell:
> ### ⚠️ Scope reduced — the whitespace half of this PR has been removed
>
> When this PR was opened it did **two** things: whitespace-normalize candidate entity names, **and** skip tag-shaped category labels. The normalization half **landed upstream independently** as **#3338** (`44b597c4`), which went further and also added per-fact case-insensitive dedupe and an empty-candidate drop.
>
> **That half is now deleted from this PR.** Upstream's implementation is taken wholesale — this branch merges `main` and keeps none of its own normalization code or tests. What remains is only the tag-shape filter, which #3338 does not cover. Nothing here duplicates or re-litigates #3338.

Extraction was minting **category labels as entities** — `domain:lens`, `domain:host`, `domain:memory` — which behave like broad tags and measurably poison entity-based scoping: **61.5% vs 94.0% precision** on one subject on a production bank. This skips tag-shaped candidate names at retain intake.

### What is left in this PR

- `_is_tag_shaped_name()` + a conservative shape `^[a-z][a-z0-9_-]{1,15}:[a-z][a-z0-9_-]{1,24}$`
- the skip itself in `_prepare_entities_for_resolution`, with a `skipped_tag_shaped` counter and a debug log line
- `entity_labels` threaded through so configured labels can be exempted
- tag-shape tests only — our normalization tests were **dropped as duplicates** of upstream's

### Check order is load-bearing

```
normalize -> empty-drop -> tag-shape (label-exempt) -> dedupe
```

Tag-shape must run **before** dedupe: a skipped tag then never enters `seen_in_fact`, so every tag-shaped occurrence increments `skipped_tag_shaped` and dedupe only tracks survivors. If it ran after dedupe, a second identical tag would be silently deduped instead of counted. Upstream's normalization, empty-drop and dedupe behaviour is unchanged.

### Configured labels stay exempt

A tag-shaped name that is a **configured** entity label (e.g. `use:use-001` from a tag-type label group) is exempt via `is_label_entity` and still reaches entity resolution — label values are deliberately `key:value` shaped (GH-1558 exact-match path).

The shape deliberately does not match URLs, Windows paths, times (`12:30`), names with spaces or dots (`re: subject`), or names with no colon.


JoshFunnell:
Heads-up on a scope change, since a body edit doesn't notify anyone.

The whitespace-normalization half of this PR is **removed**. That work landed independently as #3338 (`44b597c4`), which also added per-fact case-insensitive dedupe and an empty-candidate drop — all of which is taken wholesale here. This branch now merges `main` and keeps none of its own normalization code or tests.

What remains is only the tag-shape filter, which #3338 does not cover: extraction was minting category labels (`domain:lens`, `domain:host`) as entities, measured at 61.5% vs 94.0% precision for one subject on a production bank. Configured entity labels (`use:use-001`) stay exempt via `is_label_entity`.

One ordering detail worth a look during review: the skip runs **before** the dedupe, so a skipped tag never enters `seen_in_fact` and each occurrence is counted rather than silently deduped. Upstream's normalization, empty-drop and dedupe behaviour is otherwise unchanged.

nicoloboschi:
Closing — see the decision on #3276.

Short version: label classification is definition-driven (`is_label_entity`, `entity_labels.py:248`), so a tag-shape regex would be the only place in the pipeline that decides what an entity is without consulting the bank's `entity_labels` — at write time, unrecoverably, including on caller-supplied `content.entities`. `entities_allow_free_form: false` is the existing per-bank lever for that case.

The whitespace-normalization half of this PR landed separately in #3338, so nothing is lost. Thanks for the measurement and for holding the branch.

PR #3281, opened 2026-08-08
fix(coding-agents): deduplicate identical async retains

### LINKED ISSUE #3280: Coding Agents: deduplicate identical async retains submitted by Stop hooks
## Problem

The Coding Agents Stop-hook path retains the complete canonical transcript under a stable `conversation:<session-id>` document id. A retry of the same completed hook submission currently sends a second async `POST /memories` without an `operation_id`, so Hindsight admits another async operation for an identical payload.

The server already supports idempotent async retains through the optional `operation_id` added in #2947. Coding Agents `0.1.1` does not use that field.

## Expected behavior

Retrying an identical asynchronous Stop-hook retain should return the original operation instead of creating duplicate retain work. A transcript that has changed must create a new operation.

## Root cause

`retainLiveSession()` always calls the raw HTTP client's `retain()` with `async: true`, but the client request body contains only `items` and `async`. It never supplies the server-supported `operation_id` field.

## Narrow proposed fix

Add an opt-in `idempotent` option to the Coding Agents raw HTTP client. For an opt-in async retain, derive a deterministic RFC 4122 version-5 UUID from the exact bank and retained item, then send it as `operation_id`.

Only `retainLiveSession()` opts in. Backfill, Git, initiative markers, and all other retain callers remain unchanged.

The operation identity deliberately includes the complete retained item, rather than the session id alone. This makes the same payload idempotent while ensuring a newly appended turn receives a new server operation.

## Proof

- Focused Coding Agents tests: 6 passed.
- Full native Coding Agents suite in an isolated HOME: 35 files, 348 tests passed.
- Type-check and package build passed.
- No server, bank, transcript, hook configuration, credential, or production operation was changed.

## Scope and limits

This is a client-side adoption of an existing server contract, not a change to chunk-delta behavior. It requires a Hindsight server that supports async `operation_id` (including `v0.9.0`). It does not claim that a historical transcript is safe to reprocess unless the server's existing delta contract accepts that document representation.

TheAngryPit:
## Summary

- Closes #3280.
- Add opt-in idempotency for asynchronous Coding Agents retains using the server's existing `operation_id` contract.
- Derive an RFC UUIDv5 operation ID from the exact retained bank, document identity, and payload.
- Preserve Codex rollout timestamps when they exist and avoid fabricating a current timestamp when they do not.
- Make repeated Stop processing over an unchanged rollout produce the same payload and operation ID, while changed rollout content still produces a distinct operation.

## Root cause

The Coding Agents client did not send an operation ID for asynchronous live-session retains. The first implementation made identity payload-scoped, but Codex transcripts without source timestamps still received a new `Date().toISOString()` value on every Stop. That changed the rendered payload and defeated deduplication across manual repeated Stops.

The final implementation keeps the timestamp optional throughout the Codex transcript and retain path. Source timestamps are preserved; absent timestamps remain absent deterministically.

## Validation

```text
Focused chat, retain-client, retain-hook and Codex transcript tests
  16 tests passed

TypeScript --noEmit
  passed

Production build
  passed

Lock-exact combined integration suite containing this branch
  37 files passed, 367 tests passed
```

The regressions cover identical repeated Stop processing, changed rollout content, legacy transcripts without timestamps, RFC UUIDv5 vectors, ordinary retain compatibility, and the live-session call path.

## Scope

- No Hindsight server API, migration, schema, database, worker, retry, or persistence-layer change.
- No hook configuration change.
- No session-scoped identity that could collapse a later changed transcript into an earlier operation.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/chat.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/chat.test.ts
@@ -68,7 +68,11 @@ describe("retainLiveSession", () => {
     expect(documentId).toBe("conversation:s2");
     expect(tags).toEqual(["source:chat"]);
     expect(strategy).toBe("conversation");
-    expect(opts).toMatchObject({ async: true, timestamp: "2026-01-01T00:00:00Z" });
+    expect(opts).toMatchObject({
+      async: true,
+      idempotent: true,
+      timestamp: "2026-01-01T00:00:00Z",
+    });
     expect(opts.metadata).toMatchObject({
       source: "chat",
       session_id: "s2",
--- a/hindsight-integrations/coding-agents/src/core/chat.ts
+++ b/hindsight-integrations/coding-agents/src/core/chat.ts
@@ -1,7 +1,8 @@
 /**
  * Harness-agnostic chat memory: the JSON user/assistant transcript schema shared by BOTH the
  * backfill (ingest past sessions) and the live runtime write-back. A leading `system` turn carries
- * the REF-ID tracer; every turn gets an ABSOLUTE timestamp.
+ * the REF-ID tracer. Backfilled turns always get an absolute timestamp; live turns preserve the
+ * source transcript timestamp when one exists and never fabricate one during a retry.
  */
 import type { HindsightClient } from "./hindsight";
 import type { ChatSession } from "./types";
@@ -14,8 +15,15 @@ export interface TransportTurn {
 }
 
 /** Prepend the REF-ID system turn to a set of already-normalized turns. */
-export function withRefId(refId: string, turns: TransportTurn[], baseTs: string): TransportTurn[] {
-  return [{ role: "system", content: `REF-ID: ${refId}`, timestamp: baseTs }, ...turns];
+export function withRefId(
+  refId: string,
+  turns: TransportTurn[],
+  baseTs?: string
+): TransportTurn[] {
+  return [
+    { role: "system", content: `REF-ID: ${refId}`, ...(baseTs ? { timestamp: baseTs } : {}) },
+    ...turns,
+  ];
 }
 
 /**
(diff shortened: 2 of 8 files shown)


koriyoshi2041 (commented):
The retry semantics and scope look good, and I reproduced the focused 6/6 tests plus the package build locally. One contract detail should be resolved before this is described as UUIDv5, though: `retainOperationId()` hashes an ad-hoc prefix and JSON payload with SHA-256, truncates it, and sets the version nibble to 5. RFC 4122/RFC 9562 UUIDv5 is specifically the SHA-1 hash of a namespace UUID plus a name. Setting the nibble alone does not make this output UUIDv5, so a standard v5 implementation cannot reproduce these IDs from a documented namespace/name pair.

Could this either use an actual v5 derivation with a fixed namespace, or rename/document the scheme as a custom deterministic UUID-shaped operation ID and remove the RFC/version-5 claim from #3280? The current server-side idempotency should still work, but accurately defining the derivation matters if another Coding Agents implementation needs to reproduce it later.

TheAngryPit:
Thanks — you're right about the contract mismatch.

Fixed in `550fd73`: the operation ID now uses RFC 9562 UUIDv5 (SHA-1 over a namespace UUID plus the canonical name), rather than truncating SHA-256 and only setting the version nibble.

The reproducible derivation is:

- fixed namespace: `61cecd19-ec9a-56d6-aa7e-7e891b8b2bcb`
- that namespace is UUIDv5(URL namespace, `https://github.com/vectorize-io/hindsight/tree/main/hindsight-integrations/coding-agents#retain-operation/v1`)
- operation name: the exact UTF-8 output of `JSON.stringify({ bank, item })`

I also replaced the shape-only assertion with an exact interoperability vector generated independently using Python's `uuid.uuid5`.

Fresh validation after the fix:

- focused tests: 6/6 passed
- full Coding Agents suite: 35 files, 348 tests passed
- `tsc --noEmit`: passed
- package build: passed
- no new dependency

TheAngryPit:
@koriyoshi2041 Follow-up structured review on the corrected branch head `550fd73` is also clean: no accepted/actionable findings, `patch is correct` (0.99 confidence). Ready for re-review when convenient.

koriyoshi2041 (approved):
Re-reviewed the corrected head 550fd735. The fixed namespace independently reproduces from the documented URL with Python uuid.uuid5, and the implementation now uses the RFC namespace-bytes + UTF-8 name + SHA-1 derivation with the correct version and variant bits. The exact cross-language vector closes the interoperability concern.

Focused retain/chat tests pass 6/6 and diff check is clean. Approved.

TheAngryPit:
Closing as superseded by #3336, which landed the upstream incremental append and deterministic operation-ID design and closed #3280. Rebasing this older implementation would duplicate and conflict with the merged path. Thank you for the review that improved the UUIDv5 contract; any residual retry case should be reproduced against current main and handled as a fresh narrow change.

PR #3283, opened 2026-08-08
fix(retain): preserve delta path for oversized replacements

### LINKED ISSUE #3282: Oversized document replacements bypass delta retain and re-extract unchanged chunks
## Problem

For a document replacement whose complete body exceeds `retain_batch_tokens`, `retain_batch_async()` splits the input before the delta path gets a chance to classify the whole replacement. A small edit plus a tail can therefore re-extract unchanged history instead of processing only changed/new native chunks.

## Expected

A document with a stable `document_id` and `update_mode="replace"` should attempt delta classification against the complete body before the bounded transport fallback. If a delta is available, extraction should receive only the changed/new chunks, even when their aggregate token count exceeds the single-input transport budget. Existing `retain_chunk_batch_size` remains the native bound for that chunk list.

## Reproduction

1. Retain a document with a stable `document_id`.
2. Change one earlier section and append a tail.
3. Set `retain_batch_tokens` below the full replacement and below the aggregate changed-chunk total.
4. Retain the replacement.

Current behavior enters the full-document split/recovery path. Expected behavior identifies unchanged chunks and extracts only the changed/new chunks.

## Scope

This is a server-side retain-path issue. It does not alter the Coding Agents protocol, schemas, document IDs, database migrations, or client configuration.

A focused regression exercises the exact condition and a companion test verifies a failed delta attempt does not re-run Memory Defense for every fallback sub-batch.

TheAngryPit:
## Summary

Fixes #3282.

An oversized single-document replacement now attempts the existing delta-retain path before the transport splitter runs. When the document changed, the server extracts only the changed/new native chunks; when no delta is possible, it preserves the existing bounded split path.

The correction deliberately keeps `retain_chunk_batch_size` as the limit for a delta chunk list. `retain_batch_tokens` is a limit for complete input items, not an aggregate limit across already-bounded delta chunks.

## Scope

- no public API or Coding Agents protocol change
- no schema or migration change
- no document-ID, configuration, or client-routing change
- no change to the normal small-document retain path

## Validation

- `test_oversized_changed_tail_does_not_reextract_matching_history` passes on a fresh named `pg0` database and proves four matching chunks are skipped while two changed chunks are extracted.
- `test_oversized_delta_fallback_screens_memory_defense_once` passes and preserves one Memory Defense screening pass before bounded fallback.
- Ruff, Python compilation, and `git diff --check` pass.

The full replacement-test module additionally requires the repository local-ML test extra (`sentence-transformers`); it was not installed for this review. The two focused regressions use the repository stub embeddings and are independent of that optional dependency.

## Review

Structured auto-review completed with no actionable findings.


TheAngryPit:
Updated onto current main at 1ffe8759 (Coding Agents v0.1.2). The merge is clean and the PR diff remains limited to the two retain implementation files plus the PostgreSQL regression test. Fresh diff check and Python syntax compilation passed. I could not rerun the PostgreSQL-backed test locally in this update environment because the API test dependencies are not installed; that proof remains delegated to the PR CI rather than being overstated here.

TheAngryPit:
Closing as superseded by #3351, which resolved #3282 by aligning oversized transport sub-batches with the bank's native chunk boundaries. That addresses the underlying invariant more directly than this delta-first approach and also carries the single-screening Memory Defense correction. Thank you for landing the fix — the reproduction and focused proof here can remain as useful historical context.

PR #3286, opened 2026-08-09
fix(coding-agents): fall back from filesystem-root worktrees

### LINKED ISSUE #3278: coding-agents (opencode): bank ID resolves to empty when workspace is a non-git parent directory — input.worktree is '/'
## Bug Description

When using the `hindsight-coding-agents` plugin with **opencode** in a workspace that is a non-git parent directory (containing multiple sub-repos), the bank ID resolves to `coding-agent::` (empty `{gitProject}`).

The root cause: opencode passes `input.worktree = "/"` (the filesystem root) as the workspace root. The plugin's `createPluginEntry` prefers `input.worktree` over `input.directory`:

```js
// hindsight-integrations/coding-agents/src/harness/plugin-entry.ts
const projectDir = input?.worktree || input?.directory;
```

Since `worktree` is `"/"`, `mapPathToBank` for the actual project directory never matches. The template `coding-agent::{gitProject}` then resolves `{gitProject}` against `/`, which has no `.git`, producing an empty string.

This also causes a fatal error when the plugin tries to run git commands against `/`.

## Steps to Reproduce

1. Create a parent directory with multiple git repos inside (no `.git` in the parent)
2. Install hindsight-coding-agents plugin for opencode (`npx @vectorize-io/hindsight-coding-agents install opencode`)
3. Configure `mapPathToBank` in `~/.hindsight/coding-agent.json`:
   ```json
   {
     "mapPathToBank": {
       "~/Projects": "my-bank"
     }
   }
   ```
4. Open opencode at the parent directory level
5. Call `hindsight_diagnose` — `bank_id` will be `coding-agent::` instead of `my-bank`

## Debug Data

Captured by writing to a temp file from `plugin-entry.ts`:

```json
{
  "inputKeys": ["client", "project", "worktree", "directory", "experimental_workspace", "serverUrl", "$"],
  "worktree": "/",
  "directory": "/Users/<username>/Workspace",
  "projectDir": "/",
  "cwd": "/Users/<username>/Workspace"
}
```

As shown, `input.worktree` is `"/"` while `input.directory` has the correct path. The `||` operator picks `worktree` (`"/"`), bypassing `mapPathToBank`.

## Expected Behavior

When `input.worktree` is `"/"` (or any non-meaningful root path), the plugin should fall back to `input.directory` for bank resolution, so that `mapPathToBank` and the template resolver use the actual project directory.

## Actual Behavior

- `input.worktree` (`"/"`) takes priority over `input.directory`
- `mapPathToBank` doesn't match `"/"`
- `{gitProject}` resolves to empty (no `.git` at `/`)
- Bank ID becomes `coding-agent::` (empty)
- All retains/recalls/reflects go to the wrong bank
- Fatal error from git commands run against `/`

## Workaround

Setting a static `bankId` in config bypasses the broken path resolution:

```json
{
  "bankId": "my-bank"
}
```

## Suggested Fix

Ignore `worktree` when it's the filesystem root, or prefer `directory` for the opencode harness:

```js
const projectDir = (input?.worktree && input.worktree !== "/" && input.worktree !== "")
  ? input.worktree
  : (input?.directory || process.cwd());
```

## Version

- `hindsight-coding-agents`: **0.1.1**
- Harness: **opencode**
- OS: macOS (darwin, arm64)

koriyoshi2041:
## Problem

OpenCode can report the filesystem root as `input.worktree` while `input.directory` contains the actual workspace. Treating that root as the project directory bypasses `mapPathToBank`, produces an empty `gitProject`, and sends git operations to the filesystem root.

Closes #3278.

## Fix

Treat POSIX and Windows filesystem roots as unusable worktree values and fall back to `input.directory`. Non-root worktrees keep their existing priority. Added focused coverage for meaningful worktrees, POSIX root, both Windows root forms, and an empty worktree.

## Test

- `npm test` in `hindsight-integrations/coding-agents` (350 tests)
- `npm run build` in `hindsight-integrations/coding-agents`
- Prettier check for the changed files
- `git diff --check`

## Risk

Low. The selection changes only when the host reports a filesystem root or an empty worktree; normal repositories and linked worktrees retain the current behavior.

### DIFF
--- a/hindsight-integrations/coding-agents/src/harness/plugin-entry.test.ts
+++ b/hindsight-integrations/coding-agents/src/harness/plugin-entry.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from "vitest";
+import { resolveProjectDirectory } from "./plugin-entry";
+
+describe("resolveProjectDirectory", () => {
+  it("prefers a meaningful worktree", () => {
+    expect(
+      resolveProjectDirectory({
+        worktree: "/Users/me/project",
+        directory: "/Users/me/workspace",
+      })
+    ).toBe("/Users/me/project");
+  });
+
+  it("falls back from the POSIX root to the workspace directory", () => {
+    expect(resolveProjectDirectory({ worktree: "/", directory: "/Users/me/workspace" })).toBe(
+      "/Users/me/workspace"
+    );
+  });
+
+  it.each(["C:\\", "C:/"])(
+    "falls back from the Windows root %s to the workspace directory",
+    (worktree) => {
+      expect(resolveProjectDirectory({ worktree, directory: "C:\\Users\\me\\workspace" })).toBe(
+        "C:\\Users\\me\\workspace"
+      );
+    }
+  );
+
+  it("falls back when worktree is empty", () => {
+    expect(resolveProjectDirectory({ worktree: "", directory: "/Users/me/workspace" })).toBe(
+      "/Users/me/workspace"
+    );
+  });
+});
--- a/hindsight-integrations/coding-agents/src/harness/plugin-entry.ts
+++ b/hindsight-integrations/coding-agents/src/harness/plugin-entry.ts
@@ -10,6 +10,7 @@
  * the `harnesses.<name>` config section and feeds `{harness}` bank templating.
  */
 import type { Plugin } from "@opencode-ai/plugin";
+import { posix, win32 } from "node:path";
 import { deriveBankId } from "../core/bank";
 import { applyBankConfig, loadConfig } from "../core/config";
 import { log } from "../core/log";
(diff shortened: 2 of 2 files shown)


nicoloboschi (approved):
Verified against the reported scenario: `input.worktree` has exactly one consumer, so this is complete for its scope, and returning `undefined` is safe because `createPluginEntry` already falls back to `process.cwd()`. Node's path semantics check out (POSIX, both Windows drive forms, UNC roots all detected). Tests pass locally alongside the bank suite.

Follow-up, not blocking: `gitProjectName` still yields an empty `{gitProject}` for `/` itself (`basename("/") === ""`), so the `coding-agent::` id remains reachable if a harness reports a root as its cwd. A `(directory && basename(directory)) || "unknown"` backstop in `core/bank.ts` would match what the `{project}` resolver already does.

PR #3288, opened 2026-08-09
fix(coding-agents): approve Codex survey ingestion

JiehoonKwak:
## Summary

- preapprove only hindsight_ingest_document for detached Codex codebase surveys
- keep the existing read-only filesystem sandbox and user-wide approval policy unchanged
- add a regression assertion for the generated Codex command

## Root cause

The survey runs through non-interactive codex exec. Without a per-tool approval override, the MCP write asks for approval that the background process cannot answer, so Codex reports user cancelled MCP tool call and the survey findings are never ingested.

This is the integration-side workaround for the Codex behavior tracked in openai/codex#16685.

## Validation

- npx tsc --noEmit
- npm test (345 tests)
- npm run build
- ./scripts/hooks/lint.sh

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/survey.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/survey.test.ts
@@ -105,6 +105,9 @@ describe("startCodebaseSurvey", () => {
     expect(argv).toContain(`mcp_servers.hindsight.command="node"`);
     expect(argv).toContain(`mcp_servers.hindsight.args=["/x/mcp-server.js"]`);
     expect(argv).toContain(`mcp_servers.hindsight.env.HINDSIGHT_MCP_PROJECT_CWD="/repo"`);
+    expect(argv).toContain(
+      `mcp_servers.hindsight.tools.hindsight_ingest_document.approval_mode="approve"`
+    );
     // No Claude-only flags leak into the codex recipe.
     expect(argv).not.toContain("--model");
     expect(argv).not.toContain("--disallowedTools");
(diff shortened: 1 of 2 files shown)


JiehoonKwak:
After upgrading to Hindsight 0.9.0 and installing @vectorize-io/hindsight-coding-agents 0.1.1 for Codex 0.147.0, and during the initial codebase survey, all four `hindsight_ingest_document` calls, plus a retry, failed immediately with user cancelled MCP tool call. 

The survey runs through a detached, non-interactive codex exec process, so it cannot answer an MCP approval prompt. The requests were cancelled before reaching the Hindsight server, leaving the survey baseline without its repository documents.

Confirmed that the Hindsight ingestion endpoint itself was working: the same payload succeeded once `hindsight_ingest_document` was given per-tool approval_mode="approve". This PR applies that approval only to the ingestion tool used by the background survey. Codex remains under --sandbox read-only, and the change does not alter the user’s global approval policy or approve unrelated MCP tools.

JiehoonKwak:
Closing this after re-checking the configuration boundary. The proposed survey override is unnecessary: with `mcp_servers.hindsight.default_tools_approval_mode = "approve"` in the active Codex config, the stock 0.2.1 detached survey inherits that policy even while overriding the MCP command, args, and project cwd. Patch-free stock-shaped canaries completed `hindsight_list_knowledge_pages` without an approval request on macOS, Linux, and Synology across all four configured hosts. The original failure came from the desired approval policy not being present in the effective Codex config, so encoding that policy in the coding-agents runtime would duplicate and potentially override user configuration. Thanks, and apologies for the incorrect diagnosis.

PR #3293, opened 2026-08-09
fix(coding-agents): stream oversized Codex rollouts

### LINKED ISSUE #3292: Coding Agents: oversized Codex rollouts silently skip Stop retains
## Bug Description

The Coding Agents Codex transcript reader materializes the complete rollout with
`readFileSync(path, "utf8")`. Once a rollout exceeds Node's maximum string
length, Node throws `ERR_STRING_TOO_LONG` before any JSONL record is parsed.

`readCodexTranscript()` catches that file-level exception and returns an empty
turn list. The Stop hook then exits successfully without submitting a retain,
so an old long-running Codex task silently stops updating Hindsight.

This is independent of Hindsight server availability, provider balance, chunk
delta processing, or async operation idempotency: no HTTP request is made.

## Steps to Reproduce

1. Use Coding Agents `0.1.1` with a Codex JSONL rollout larger than Node's
   maximum string length.
2. Invoke the installed Codex Stop hook for that rollout.
3. Observe that `readFileSync(path, "utf8")` throws
   `ERR_STRING_TOO_LONG`.
4. Observe that the reader returns `[]`, the hook exits successfully, and no
   retain operation reaches Hindsight.

A deterministic regression can reproduce the boundary without allocating a
multi-gigabyte fixture by making `readFileSync` throw an error with code
`ERR_STRING_TOO_LONG` and asserting that a valid JSONL record is still parsed.

## Expected Behavior

Codex rollouts should be parsed incrementally so transcript size is bounded by
the largest JSONL record, not by Node's maximum JavaScript string length. The
existing per-line parsing, filtering, and fail-open behavior should remain
unchanged.

## Actual Behavior

The entire rollout is decoded into one JavaScript string. Oversized rollouts
become an empty transcript silently, and the Stop hook performs no retain.

## Version

`@vectorize-io/hindsight-coding-agents` `0.1.1`; reproduced on current
`origin/main` at `f1c825d8`.

## Proposed Narrow Fix

Read the UTF-8 JSONL file in bounded chunks with `StringDecoder`, yield complete
records one at a time, and keep the existing record parser unchanged. Add a
regression proving the parser no longer depends on whole-file `readFileSync`.

## Validation Performed

- Regression failed on the old implementation and passed after the change.
- Focused Codex transcript tests: 2 files, 7 tests passed.
- Full Coding Agents suite in isolated config: 35 files, 346 tests passed.
- Package build passed.
- The new reader parsed four read-only oversized rollout fixtures; no hook,
  Hindsight request, credential, or retained content was involved in that proof.
- Structured code review: clean, no accepted/actionable findings.

## Non-goals

- No server, worker, database, chunking, delta, operation-id, or provider change.
- No hook configuration or retained-content change.
- No new dependency.

TheAngryPit:
## Summary

- Stream Codex JSONL rollouts instead of decoding the complete file into one
  JavaScript string.
- Preserve the existing record parser, filters, and fail-open contract.
- Add a regression for the `ERR_STRING_TOO_LONG` boundary.

Closes #3292.

## Root cause

`readCodexTranscript()` used `readFileSync(path, "utf8")`. Oversized Codex
rollouts therefore fail before parsing with Node's `ERR_STRING_TOO_LONG`. The
reader catches the exception and returns no turns, causing the Stop hook to
exit successfully without submitting a retain.

## Why this shape

The rollout format is JSONL, so whole-file materialization is unnecessary.
Incremental UTF-8 decoding bounds memory by the largest record while leaving
all existing parsing semantics in one place. `StringDecoder` preserves UTF-8
code points split across read buffers.

## Validation

```text
Regression on old implementation
  failed: expected one turn, received []

npm test -- --run src/core/transcript-codex.test.ts src/core/transcript-codex-large.test.ts
  2 files passed, 7 tests passed

HINDSIGHT_CONFIG=<empty-test-config> npm test
  35 files passed, 346 tests passed

npm run build
  ESM and DTS builds passed

Read-only oversized-rollout parser proof
  4/4 files parsed successfully; largest file >2.4 GB

openclaw-autoreview --mode local
  clean: no accepted/actionable findings
```

## Scope

- No dependency changes.
- No server API, database, worker, chunk-delta, or operation-id changes.
- No hook configuration or retained-content changes.
- This is separate from async retain idempotency in #3281.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/transcript-codex-large.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/transcript-codex-large.test.ts
@@ -1,0 +1,48 @@
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("node:fs", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs")>();
+  return {
+    ...actual,
+    readFileSync: vi.fn(() => {
+      const error = new Error("Cannot create a string longer than 0x1fffffe8 characters");
+      Object.assign(error, { code: "ERR_STRING_TOO_LONG" });
+      throw error;
+    }),
+  };
+});
+
+import { readCodexTranscript } from "./transcript-codex";
+
+let root: string;
+let file: string;
+
+beforeEach(() => {
+  root = mkdtempSync(join(tmpdir(), "hs-codex-large-transcript-"));
+  file = join(root, "rollout.jsonl");
+});
+
+afterEach(() => {
+  rmSync(root, { recursive: true, force: true });
+});
+
+describe("readCodexTranscript oversized rollout handling", () => {
+  it("parses JSONL without reading the complete file into one string", () => {
+    writeFileSync(
+      file,
+      JSON.stringify({
+        type: "response_item",
+        payload: {
+          type: "message",
+          role: "user",
+          content: [{ type: "input_text", text: "keep streaming" }],
+        },
+      })
+    );
+
+    expect(readCodexTranscript(file)).toEqual([{ role: "user", content: "keep streaming" }]);
+  });
+});
--- a/hindsight-integrations/coding-agents/src/core/transcript-codex.ts
+++ b/hindsight-integrations/coding-agents/src/core/transcript-codex.ts
@@ -20,7 +20,8 @@
  * prevents a retain→recall feedback loop (plus stripInjectedMemory as a defensive second pass on the
  * text we do keep). Fail-open: never throws on a missing file or a malformed line.
  */
-import { readFileSync } from "node:fs";
+import { closeSync, openSync, readSync } from "node:fs";
+import { StringDecoder } from "node:string_decoder";
 import type { TransportTurn } from "./chat";
 import { actionLine, stripInjectedMemory } from "./transcript-util";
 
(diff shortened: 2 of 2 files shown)


koriyoshi2041 (approved):
Verified at `08c6362ed`.

- Focused Codex transcript suite: 7/7 passed.
- `git diff --check`: passed.
- The reader now keeps whole-file decoding off the Stop path, preserves split UTF-8 code points via `StringDecoder`, closes the descriptor in `finally`, and retains the existing per-record fail-open/filtering behavior.

The bounded JSONL reader addresses the silent no-retain failure without changing retained-content semantics.

nicoloboschi:
Closing in favour of #3345, which took a different route to the same bug. Thank you for this — your diagnosis in #3292 was exactly right, and the `ERR_STRING_TOO_LONG` → empty-turns → silent no-op chain is what the fix was built from.

Where this PR and #3345 differ is where the ceiling goes. Streaming the read removes the hard string-length cliff, but at the time it was proposed the retain side was still unbounded: `buildRetain` sent every turn on every Stop, and `renderSessionJsonl` collapsed them into one string. Measured on a 170 MB rollout, that path cost 451 MB heap and produced a 137 MB document per write — so on the >2.4 GB rollout cited here, the parse would have succeeded and then the `join` would have thrown `ERR_STRING_TOO_LONG` one layer down, or the hook would have OOMed first. The parser-only proof in the PR description couldn't show that, since it deliberately ran without the hook or a request.

That's since been addressed from the other end. #3336 made the write-back incremental — a per-session cursor and `update_mode: "append"`, so a retain sends only the turns added since the last confirmed write. With the pipeline bounded, a bounded read is the missing half rather than a way to relocate the failure, which is what #3345 does:

- `readJsonlTail` streams 64 KB at a time through a `StringDecoder` and reads only the last 32 MB, so memory is bounded by the cap rather than by the file
- head-truncation is logged with size and bytes skipped — the silence was the real complaint in #3292
- applied to all six readers sharing the `readFileSync(...).split("\n")` shape (`claude-code`, `codex`, `antigravity-cli`, `copilot-cli`, `cursor-cli`, `grok-build`), since a large Claude Code transcript fails identically

Verified on a 575,680,264-byte rollout (V8's limit is 536,870,888): `readFileSync` throws, the old reader returns `[]`, and the new one returns 8,162 turns ending in the most recent exchange in 51 MB of heap.

Two things from reviewing this PR carried over into that one, and are worth recording since they'd have bitten either implementation:

- the regression here mocks `readFileSync`, which the new implementation never calls, so it can only prove "no whole-file read" — nothing about the streaming itself. #3345 covers the cases that actually distinguish a correct streaming reader: a multi-byte character split across the chunk boundary (the whole reason `StringDecoder` is there), a cut landing mid-record, a cut landing exactly on a boundary, CRLF, and a final record with no trailing newline. The boundary case caught a real off-by-one in my first attempt, which was silently dropping a whole record.
- wrapping the entire parse loop in the `catch` means a throw at 99% discards every turn parsed so far — the same silent-empty outcome #3292 is about, just moved.

#3292 is closed by #3345.

Issue #3294, opened 2026-08-09
Document replace can leave orphan observations: consolidation writes between the pre-delete sweep and the delete


fhiltscher:
### Bug Description

A document re-ingest (`retain` with an existing `document_id`) can leave orphan observations: rows whose `source_memory_ids` point at `memory_units` that no longer exist. They stay valid and recallable, so stale knowledge from the previous version of a document survives the replace.

`handle_document_tracking` does call `delete_stale_observations_for_memories` before the delete (added in #1101, and the residue was swept by `g7h8i9j0k1l2` / `c4x5y6z7a8b9`), and #1090 closed the window where a source is deleted *during* the LLM call. The window that remains is a different one — between that sweep and the delete:

```python
# engine/retain/fact_storage.py
if existing_unit_ids:
    invalidated = await delete_stale_observations_for_memories(conn, bank_id, existing_unit_ids, ops=ops)
    ...
await store.delete_document(conn=conn, fq_table=fq_table, bank_id=bank_id, document_id=document_id, txn=txn)
```

An observation a concurrent consolidation inserts in between is invisible to both: its sources were still live when `_filter_live_source_memories` checked them `FOR SHARE`, and the sweep had already run. Once the delete removes those sources, nothing can reach the row again — consolidation batches are built from facts, so an observation whose sources are all gone is never selected into a batch. There is also no runtime repair path: `clear_observations` is bank-wide, `DELETE /memories/{id}/observations` needs a live source, `consolidation/recover` explicitly deletes no observations, and graph maintenance prunes entities, not observations. The only cleanup shipped is the Alembic backsweep, which runs once at upgrade.

### Steps to Reproduce

1. Bank with observations consolidated from a document's facts.
2. `retain` the same `document_id` again (full replace) while a consolidation for that scope is still running — in our case two replaces of the same document in short succession.
3. Read back the observations and resolve each `source_memory_ids` entry.

### Expected Behavior

Every observation references at least one existing `memory_unit`, per the documented lifecycle ("removing a document: all observations derived from the document's memories are deleted").

### Actual Behavior

On our bank (198 facts, 193 observations, 10 documents), after two replaces: 38 observations with unresolvable source references, 22 of them with no surviving source at all.

[...]

PR #3295, opened 2026-08-09
fix(retain): sweep observations written during the document delete

fhiltscher:
## Summary

Closes one more window in which a document replace leaves orphan observations. Reported in #3294.

`handle_document_tracking` sweeps the observations derived from the outgoing memories before deleting them (#1101), and #1090 closed the window where a source is hard-deleted *during* the consolidation LLM call. What remains is the gap between those two steps:

```python
if existing_unit_ids:
    invalidated = await delete_stale_observations_for_memories(conn, bank_id, existing_unit_ids, ops=ops)
    ...
await store.delete_document(conn=conn, fq_table=fq_table, bank_id=bank_id, document_id=document_id, txn=txn)
```

An observation a concurrent consolidation inserts in between is invisible to both. Its sources were still live when `_filter_live_source_memories` checked them `FOR SHARE`, so the insert is legitimate; the sweep had already run, so it never saw the row. The delete then orphans it permanently: consolidation batches are built from facts, so an observation whose sources are all gone is never selected into a batch again.

## Change

Sweep once more after the delete, for the same ids. `delete_stale_observations` matches on the observation's own `source_memory_ids` rather than joining live rows, so the repeat call is valid after the sources are gone, and idempotent — normally it finds nothing. The docstring said "before the source memories are deleted"; it now states that a repeat call afterwards is valid.

Cheap: one extra query per replaced document, only when the document had outgoing memories, inside the transaction that is already open.

## Test

`test_upsert_document_sweeps_observation_written_during_the_delete` reproduces the window deterministically: it patches the sweep so that an observation referencing the outgoing memory is inserted right after the first sweep call, then runs `handle_document_tracking`.

- without the change: fails on the orphan assertion, while `test_upsert_document_removes_observations_from_outgoing_memories` still passes
- with the change: `tests/test_observation_invalidation.py` 27 passed

`tests/test_retain.py` and `tests/test_async_batch_retain.py` are unchanged against a clean tree (the `TestFactExtractionQuality` failures here are a missing LLM key in my environment, identical with and without the patch). `./scripts/hooks/lint.sh` passes, `ty check hindsight_api/` passes.

## Note

This closes the window it describes. I have not proved it is the only cause of the orphans in #3294 — that reproduction ran two replaces in short succession with auto-consolidation active, and the bank also had no way to clean the residue afterwards. Deliberately not marked as closing the issue.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memories/pg/writes.py
+++ b/hindsight-api-slim/hindsight_api/engine/memories/pg/writes.py
@@ -252,8 +252,9 @@ async def delete_stale_observations(
     2. Reset ``consolidated_at = NULL`` on the surviving source memories so
        they get re-consolidated under fresh observations on the next run.
 
-    Must be called within an active transaction, before the source memories
-    are deleted.
+    Must be called within an active transaction, normally before the source
+    memories are deleted. The lookup matches the observation's own
+    ``source_memory_ids``, so a repeat call afterwards is valid and idempotent.
 
     Returns the number of observations deleted.
     """
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_storage.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_storage.py
@@ -293,6 +293,17 @@ async def handle_document_tracking(
             bank_id,
         )
 
+        # Sweep again: an observation a concurrent consolidation inserted between
+        # the pre-delete sweep and the delete is invisible to both, and no later
+        # pass can reach it once its sources are gone. Idempotent.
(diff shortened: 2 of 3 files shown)


fhiltscher:
Withdrawing this. The field report it came from now reproduces serially — a single replace from a clean baseline, nothing running concurrently — so the window this PR closes is not the cause there, and I would rather not have a speculative fix sitting in the way while the actual defect is being located. The window itself still exists and the test pins it; happy to bring it back if you want it as a separate safety net once the primary cause is understood. Diagnosis continues in #3294.

PR #3302, opened 2026-08-10
fix delta retain observation cleanup

### LINKED ISSUE #3294: Document replace can leave orphan observations: consolidation writes between the pre-delete sweep and the delete

koriyoshi2041:
## Problem

Delta retain deletes changed and removed chunks directly. That cascade removes their source memories without invalidating observations derived from them, leaving stale observations with missing `source_memory_ids` (#3294).

## Fix

Collect the source-memory IDs owned by the chunks before deletion and run the existing observation invalidation path in the same transaction. The lookup supports both SQL-backed and external memory stores. The delta orchestrator now passes the backend ops needed by the cleanup query.

## Test

- Added a PostgreSQL regression covering a changed-chunk observation with a surviving co-source; it verifies the observation is deleted, the outgoing source is removed, and the survivor is requeued for consolidation.
- `pytest -n 0 tests/test_observation_invalidation.py::TestDocumentUpsertObservationCleanup::test_delta_chunk_delete_removes_observations_from_outgoing_memories -q`
- `pytest tests/test_chunk_storage_delete_ordering.py -q`
- Ruff check and format check for the changed Python files
- Repository pre-commit hooks passed

## Risk

Low to moderate. Chunk deletion now performs one bounded source lookup plus the existing cleanup routine before the cascade. The cleanup is limited to observations referencing memories owned by the chunks being replaced; unchanged source memories survive and are reset for reconsolidation.

Fixes #3294.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/chunk_storage.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/chunk_storage.py
@@ -55,7 +55,7 @@ async def load_existing_chunks(conn, bank_id: str, document_id: str) -> list[Exi
     ]
 
 
-async def delete_chunks_by_ids(conn, chunk_ids: list[str], bank_id: str | None = None, txn=None) -> None:
+async def delete_chunks_by_ids(conn, chunk_ids: list[str], bank_id: str | None = None, txn=None, ops=None) -> None:
     """
     Delete specific chunks by their IDs.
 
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Thanks for this — your diagnosis and call site were right, and #3384 is built on them.

I've opened #3384 rather than iterating here, so closing this one as superseded. What it adds on top of your change:

- the outgoing-id lookup is factored into `memory_ids_for_chunks` and reads through the store on the non-SQL path (same shape as `handle_document_tracking`'s scan);
- `delete_chunks_by_ids` returns the count it invalidated and the delta log line carries it unconditionally — the reporter specifically asked for that, since `invalidated == 0` and "the sweep never ran" were indistinguishable from the outside;
- regression coverage runs through the real retain pipeline (`retain → consolidate → retain(edited)`), not just the storage helper, and each test was verified to fail on a clean tree with the reported symptom. Worth noting for anyone reading later: a re-ingest that changes *every* chunk does not reproduce the bug at all — delta declines when no chunk is unchanged and the full-replace path sweeps correctly. The test has to keep one chunk byte-identical.

Your patch also never had CI on it: fork PRs get no secrets here, so `test-api` shows `skipped` on run 31350411875 — the test you added had not actually run.

Credited you in the commit and the PR body. Thanks again for finding this.

PR #3303, opened 2026-08-10
fix: skip consolidation recall for empty tagged scopes

### LINKED ISSUE #3301: [BUG] Consolidation stalls on a new tag scope with zero observations
## Bug Description

Consolidation may stall when the next memories belong to a strict tag scope with few or no existing observations.

For each fact in an LLM batch, `_find_related_observations()` calls `recall_async(..., fact_type=["observation"], tags_match="all_strict")`. The semantic arm over-fetches `max(limit * 5, 100)`; with the default LOW consolidation recall budget, this becomes `LIMIT 500`.

Under a selective tag filter, the PostgreSQL query can reach the database command timeout. Because the per-fact recalls are gathered together, one timeout aborts the consolidation operation and triggers exponential retry.

Relevant source paths:

- [strict-scope consolidation recall](https://github.com/vectorize-io/hindsight/blob/v0.9.0/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py#L2419-L2480)
- [semantic over-fetch and combined query](https://github.com/vectorize-io/hindsight/blob/v0.9.0/hindsight-api-slim/hindsight_api/engine/search/retrieval.py#L214-L333)

## Steps to Reproduce

1. Create a synthetic PostgreSQL/pgvector bank with observations in existing tag scopes.
2. Retain unconsolidated facts under a new or sparse strict tag scope.
3. Run consolidation with the default LOW recall budget and a finite database command timeout.
4. Observe the per-fact observation recalls using `all_strict`.
5. Under the affected query plan, the filtered search reaches the command timeout and the whole operation is retried.

Sanitized query shape:

```sql
(
  SELECT id, 1 - (embedding <=> $1::vector) AS similarity
  FROM memory_units
  WHERE bank_id = $2
    AND fact_type = 'observation'
    AND embedding IS NOT NULL
    AND (1 - (embedding <=> $1::vector)) >= 0.3
    AND tags IS NOT NULL
    AND tags != '{}'
    AND tags @> $5
  ORDER BY embedding <=> $1::vector
  LIMIT 500
)
UNION ALL
(
  SELECT id, ts_rank_cd(search_vector, to_tsquery('english', $4))
  FROM memory_units
  WHERE bank_id = $2
    AND fact_type = 'observation'
    AND search_vector @@ to_tsquery('english', $4)
    AND tags IS NOT NULL
    AND tags != '{}'
    AND tags @> $5
  ORDER BY ts_rank_cd(search_vector, to_tsquery('english', $4)) DESC
  LIMIT $3
);
```

## Expected Behavior

A new or sparse strict tag scope should be handled without blocking consolidation:

- preserve strict tag isolation;
- return promptly when there are no candidates;
- avoid aborting the whole batch because one recall times out.

## Actual Behavior

A filtered observation recall can consume the database command timeout. A single failure aborts the grouped recalls and schedules the consolidation operation for an exponentially delayed retry.

## Environment

- Reproduced with Hindsight 0.8.6
- PostgreSQL with pgvector
- The relevant source path is also present in v0.9.0

mrKlar:
## Summary

- probe a tagged observation scope with a bounded `SELECT 1 ... LIMIT 1` before consolidation recall
- skip per-fact semantic/BM25 recalls only when that strict scope has no observations yet
- preserve `all_strict` isolation and support external memory-store extensions with a bounded strict scan
- keep disjoint scope probes parallel

## Root cause

For a brand-new tagged scope, each fact launched a filtered ANN + BM25 observation recall even though the scope had zero candidates. On large banks the filtered HNSW arm could reach `db_command_timeout`, abort the batch, and trigger exponential retries.

The new existence preflight lets consolidation create the first observation for the scope without weakening tag matching or increasing timeouts.

Fixes #3301.

## Validation

- `uv run pytest tests/test_consolidation_empty_scope.py tests/test_consolidation_dedup.py tests/test_consolidation_scope_parallelism.py tests/test_consolidation.py::test_count_observations_for_scope -q` — 56 passed
- `./scripts/hooks/lint.sh` — passed
- `uv run ty check hindsight_api/engine/consolidation/consolidator.py` — passed
- regression includes a real embedded PostgreSQL/pgvector execution of the existence query

## Base

The fork branch is based exactly on release tag `v0.9.0` (`b12646f49`).

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -760,6 +760,39 @@ async def _count_observations_for_scope(
     return total
 
 
+async def _has_observations_for_scope(
+    conn: "Connection",
+    bank_id: str,
+    tags: list[str],
+) -> bool:
+    """Return whether a strict tagged scope contains any observations.
+
+    This deliberately performs a cheap existence probe before consolidation's
+    vector recall. Filtered ANN search can otherwise scan the bank-wide HNSW
+    index until the database timeout when a brand-new scope has no candidates.
+    """
+    store = get_memories()
+    if store.writes_memory_rows_in_sql:
+        found = await conn.fetchval(
+            f"SELECT 1 FROM {fq_table('memory_units')} "
+            "WHERE bank_id = $1 AND fact_type = 'observation' "
+            "AND tags IS NOT NULL AND tags != '{}' AND tags @> $2::varchar[] LIMIT 1",
+            bank_id,
+            tags,
+        )
+        return found is not None
+    page = await store.scan_memories(
+        conn=conn,
+        fq_table=fq_table,
+        bank_id=bank_id,
+        fact_types=["observation"],
+        tags=tags,
+        tags_match="all_strict",
+        limit=1,
+    )
+    return bool(page.memories)
+
+
 @dataclass(frozen=True)
 class _ScopeLimitRule:
     """One ``observation_scope_limits`` rule: a scope pattern -> an observation cap.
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Thanks for putting this together, and for the level of detail in #3301 — the report itself is genuinely useful and the issue stays open. I'm closing this PR though, because I don't think the guard fixes the reported stall, and it adds a permanent branch to the consolidation hot path in exchange.

**It only spares the first batch of a new scope.** Batches inside a tag group run serially (`consolidator.py`, "Batches within a group share a tag set and observation scope, so they MUST run serially"), and batch #1 creates that scope's first observation and commits it before batch #2 starts. From batch #2 onward `_has_observations_for_scope` returns `True` and the exact query from the issue runs again. In the production case described — 41 memories in one scope at `consolidation_llm_batch_size=8` — that's 6 batches, of which 1 is spared and 5 still hit the 60s timeout, still abort the `asyncio.gather`, and still trigger the exponentially backed-off retry. The backlog collapse is unchanged.

**The pathology is sparsity, not emptiness.** A filtered ANN arm asking for `LIMIT 500` against 37,789 observations behaves the same whether 0 or 1 or 20 rows match `tags @> $5`. `has_observations == True` isn't a proxy for "this query is cheap", so the guard is checking the wrong predicate.

**And the mechanism isn't established yet.** There's no `EXPLAIN ANALYZE` here, and the theory in the issue is hard to reconcile with `hnsw.iterative_scan` being unset (pgvector default `off`) and `ef_search=200` — that scan should terminate early. I've asked for the plan at 0 and at 1 in-scope observations in #3301; if it's slow at 1, this approach can't work, and if it's fast at 1 we'd want to understand why before special-casing zero.

One more thing worth flagging: skipping recall hands the consolidation LLM an empty candidate set, which forces CREATE instead of UPDATE. If those 37,789 observations turn out to be concentrated in a few scopes, that's already the failure mode driving the table size — and this change would accelerate it.

Smaller notes, for the record:

- `test_disjoint_scopes_run_concurrently` was rewritten to hook `_has_observations_for_scope` instead of `_find_related_observations`, so it no longer covers concurrent *recalls* across disjoint scopes — the thing it was written for.
- The new tests patch seven symbols including `acquire_with_retry` and the probe itself; they assert the wiring rather than the behaviour, and nothing reproduces the timeout or demonstrates the reduction in DB work.
- `_has_observations_for_scope` duplicates `_count_observations_for_scope` and disagrees with it for external stores (`all_strict` vs `all`).
- `observation_scope_tags` becomes redundant with the hoisted `fact_tags`, since grouping is by exact tag set.

Where I think the real wins are, from reading the surrounding code:

1. **Collapse the per-fact recall fan-out.** A batch of 8 facts fires 8 filtered ANN + BM25 queries and then unions the results into one candidate set for one LLM call. Batches are exact-tag-set groups, so all 8 target the identical scope — one round trip with 8 semantic arms preserves twin-finding quality and cuts the pathological query count 8x.
2. **Bound the semantic over-fetch.** `hnsw_fetch = max(limit * 5, 100)` is 500 at the LOW consolidation budget while `hnsw.ef_search` is 200, so the over-fetch can never be satisfied from the index — and the oversized `LIMIT` under a selective filter is exactly what makes the planner distrust the ANN path.
3. **Isolate recall failures.** A single timeout shouldn't discard a whole batch's completed work; adaptive splitting currently only reacts to LLM failures, not DB timeouts. A statement timeout for consolidation recall well below `db_command_timeout` would let one slow arm degrade instead of collapsing the operation.
4. **Possibly just planner statistics.** Array-containment selectivity on a high-cardinality `VARCHAR[]` is badly estimated by default; `ANALYZE` plus a raised statistics target on `tags`, or a partial `GIN (tags) WHERE fact_type = 'observation'`, may fix this outright and would help sparse scopes generally.

Happy to look at a follow-up along any of those lines, especially (1) — and if the EXPLAIN output in #3301 contradicts my reading, please reopen and say so.

PR #3304, opened 2026-08-10
fix(db): drop the vestigial memory_links.entity_id FK that stalls graph maintenance

mekenthompson:
## The failure

`prune_orphan_entities` — graph maintenance Pass 2 — has not completed on four
of our production banks for a week. The `overlord` bank's
`prune_stale_cooccurrences` (same transaction, runs second) has therefore not
run since 3 August.

Each job burns ~560 s and rolls back with zero progress:

```
prune_orphan_entities  →  asyncpg command_timeout (60 s)  →  ROLLBACK
  ×9  (_SWEEP_MAX_RETRIES = 8, plus the initial attempt, with backoff)
```

## Root cause

`memory_links.entity_id` still carries `fk_memory_links_entity_id_entities`
— `FOREIGN KEY (entity_id) REFERENCES entities(id) ON DELETE CASCADE` — from
the initial schema migration `5a366d414dce`, back when entity edges were
materialised as `memory_links` rows.

That representation is gone. `e1b2c3d4f5a6` dropped the last index leading on
`entity_id`, and `e9b2c7d1f3a4` deleted every `link_type = 'entity'` row. The
constraint was never dropped with them.

An `ON DELETE CASCADE` child with **no index leading on its FK column** makes
PostgreSQL's referential-integrity trigger sequentially scan the entire child
table once per deleted parent row. `memory_links` is 26,750,591 rows / ~2676 MB
and its indexes are `(bank_id, link_type)`, `(from_unit_id, …)`,
`(to_unit_id, …)`, and the unique on
`(from_unit_id, to_unit_id, link_type, COALESCE(entity_id, …))` — none of which
lead on `entity_id`.

**Measured: 1.278 s per parent row deleted, fully cached.**

So the sweep is a cheap SELECT driving a ruinously expensive DELETE:

| step | cost |
|---|---|
| orphan anti-join (`idx_unit_entities_entity_unit`, correct plan) | ~70 ms |
| DELETE of ~157 orphan entities, via the cascade | ~200 s |
| asyncpg client-side `command_timeout` (`DEFAULT_DB_COMMAND_TIMEOUT`) | 60 s |

Because it rolls back, it never makes partial progress — so this is
self-compounding. Past roughly **47 orphans** a bank can never recover on its
own. Four banks are over that line.

## The constraint protects nothing

Verified directly against production:

```sql
SELECT count(*) FROM memory_links WHERE entity_id IS NOT NULL;
-- 0     (out of 26,750,591)
```

and `pg_stats.null_frac` for the column is `1`. Nothing writes it —
`_bulk_insert_links` passes `None` for every row.

## ⚠️ Production is already ahead of migration head

As an emergency unblock, the constraint has **already been dropped on the live
database**:

```sql
ALTER TABLE memory_links DROP CONSTRAINT fk_memory_links_entity_id_entities;
```

So that database has the constraint gone while `alembic_version` still reads
`b3e8d1c6f4a9`. The migration must reconcile that, not fail on it — hence
`DROP CONSTRAINT IF EXISTS`, which here is load-bearing rather than defensive
boilerplate. Verified both ways (see "Verification").

## What this PR does

**Migration `c4f7a2e9b6d1`** (`down_revision = 'b3e8d1c6f4a9'`, confirmed head
at branch time):

* `upgrade()` — `ALTER TABLE memory_links DROP CONSTRAINT IF EXISTS
  fk_memory_links_entity_id_entities`. Catalogue-only; instant regardless of
  table size.
* `downgrade()` — a real restore: `DROP … IF EXISTS` (symmetry), then
  `ADD CONSTRAINT … NOT VALID` followed by `VALIDATE CONSTRAINT` in an
  `autocommit_block`. A plain validating `ADD` would hold `ACCESS EXCLUSIVE`
  on `memory_links` across a seq scan of 26.7M rows; `NOT VALID` takes the
  exclusive lock only for the catalogue write and `VALIDATE` then runs under
  `SHARE UPDATE EXCLUSIVE`, blocking neither reads nor writes.

  **Reversibility, stated honestly:** the restore is faithful for the data that
  exists — `entity_id` is NULL in every row and a NULL FK column is trivially
  satisfied, so `VALIDATE` cannot fail on current data. It is not reversible in
  general: a database that somehow acquired non-NULL `entity_id` values
  pointing at deleted entities while the constraint was absent would fail
  `VALIDATE` and the downgrade would abort. Nothing in the codebase can produce
  such a row. The failure mode is "downgrade refuses", not "silent corruption",
  which is the right way round. This is said in a comment in the migration too.

**PostgreSQL-only**, same call as `9f8e7d6c5b4a`: Oracle's baseline
(`o1a2b3c4d5e6`) keeps `idx_ml_entity` on `memory_links(entity_id)`, so its
`fk_ml_entity` cascade is an index lookup, not a table scan. The pathology does
not exist there and the Oracle slot is deliberately left empty (with a comment
saying so).

### Decision: the `entity_id` COLUMN stays

I grepped the whole repo for `memory_links.entity_id` before deciding. Dropping
the column is materially riskier than dropping the constraint, and buys nothing
the outage cares about:

1. It is a key column of `idx_memory_links_unique`
   `(from_unit_id, to_unit_id, link_type, COALESCE(entity_id, nil_uuid))`,
   which is the **arbiter of the retain hot path's `ON CONFLICT … DO NOTHING`**
   in `PostgresOps.bulk_insert_links`. Dropping the column drops that index;
   rebuilding a unique index over 26.7M rows must happen `CONCURRENTLY` or it
   takes `ACCESS EXCLUSIVE` on the busiest table in the schema for the duration.
2. The `ON CONFLICT` target would have to change **in lockstep, in the same
   deploy**, in three places that do not deploy atomically with the migration:
   `ops.bulk_insert_links` (its signature takes `nil_entity_uuid`),
   `ops_postgresql.bulk_insert_links`, and `ops_oracle.bulk_insert_links`. An
   old API pod against a new schema — the normal state of a rolling deploy —
   fails **every link insert** with *"no unique or exclusion constraint matching
   the ON CONFLICT specification"*. That is retain down.
3. Oracle's baseline declares the same column and the same NVL-based unique
   index, so it is a two-dialect schema change, not a one-line PG cleanup.

The cascade scan is caused by the **constraint**, not the column. A nullable
column that is NULL in every row costs one null-bitmap bit per row. Removing it
is a worthwhile follow-up with its own rollout story — not something to bundle
into an outage fix.

### Also in this PR

* **`models.py`** — `MemoryLink.entity_id` loses its `ForeignKey`, and the
  `Entity.memory_links` / `MemoryLink.entity` relationship pair goes with it
  (SQLAlchemy cannot keep the relationship without the FK). This is the durable
  half: without it, a future `alembic revision --autogenerate` would happily
  resurrect the constraint. Mapper configuration verified clean.
* **`ops_postgresql.prune_orphan_entities`** — the comment claimed the anti-join
  is backed by `idx_ue_entity`. That is not a PostgreSQL index: it was dropped
  by `h3i4j5k6l7m8` / `e1b2c3d4f5a6` and the name now survives only in the
  Oracle baseline. The real index is `idx_unit_entities_entity_unit` on
  `(entity_id, unit_id)`. Corrected, and the comment now records the invariant
  that broke here: *every* `ON DELETE CASCADE` child of `entities` needs an
  index leading on its FK column.
* **`tests/test_memory_links_entity_fk_dropped.py`** — schema invariant test in
  the shape of the existing `test_memory_links_deferred_fk.py`. Asserts no FK on
  `memory_links.entity_id`, and asserts the column + its unique index still
  exist so a future "cleanup" that drops the column has to come here and think
  about the `ON CONFLICT` target first.

## Verification

Against a real PostgreSQL 18 (pg0), driving alembic directly:

```
after downgrade -> b3e8d1c6f4a9            fk=[('fk_memory_links_entity_id_entities', validated=True)] version=b3e8d1c6f4a9
CASE1 upgrade (fk present)                 fk=[]  version=c4f7a2e9b6d1
CASE2 pre (hand-dropped, version behind)   fk=[]  version=b3e8d1c6f4a9      <- production's shape
CASE2 upgrade (fk already gone)            fk=[]  version=c4f7a2e9b6d1
CASE3 pre (fk re-added at head)            fk=[('fk_memory_links_entity_id_entities', validated=True)] version=c4f7a2e9b6d1
CASE3 downgrade (fk pre-existing)          fk=[('fk_memory_links_entity_id_entities', validated=True)] version=b3e8d1c6f4a9
final upgrade                              fk=[]  version=c4f7a2e9b6d1
```

CASE2 is the one that matters: the migration applies cleanly on the database
where the constraint was already dropped by hand.

The new test was also confirmed to **fail on the bug** — re-adding the
constraint by hand makes
`test_memory_links_entity_id_has_no_foreign_key` fail with the intended message,
so it guards an outcome rather than a code path.

```
$ uv run pytest tests/test_graph_maintenance.py tests/test_db_abstraction.py \
    tests/test_sql_schema_safety.py tests/test_memory_links_deferred_fk.py \
    tests/test_memory_links_entity_fk_dropped.py \
    tests/test_memory_links_to_unit_id_concurrent_delete.py \
    tests/test_migration_shape.py tests/test_causal_link_taxonomy.py \
    tests/test_bank_stats.py
================ 360 passed, 95 skipped, 28 warnings in 55.71s =================

$ uv run ruff check .
All checks passed!

$ uv run ruff format --check .
631 files already formatted

$ uv run ty check hindsight_api/
Found 22 diagnostics     # identical count on origin/main — no regression
```

## Recommendations (not implemented here — deliberately out of scope)

Both of these are about the *class* of failure, which will recur with any future
cascade child; this PR only removes today's instance.

1. **`prune_orphan_entities` should be batched and/or given its own timeout,
   regardless of this fix.** The shape — a cheap SELECT whose DELETE is
   expensive via cascade — is inherently unpredictable from the planner's cost
   estimate, and the all-or-nothing transaction is what makes it
   *self-compounding* rather than merely slow: because it rolls back, a bank
   past the threshold can never claw its way back. A `LIMIT n` + loop with a
   commit per batch would degrade to "slow but progressing" instead of "stuck
   forever". I'd rate that a higher-value durability change than the timeout
   knob.

2. **The 60 s `command_timeout` against a 600 s `statement_timeout` is the wrong
   way round.** As it stands, asyncpg gives up on the client side while the
   server keeps working for up to another 540 s: the client sees a timeout and
   retries, the server is still holding locks and burning I/O on the abandoned
   statement, and the retry piles onto it. For background maintenance the two
   should be paired the other way — server timeout at or below the client
   timeout — so the server is the one that aborts and the client learns about it
   promptly. Alternatively, give the maintenance path its own longer
   `command_timeout` (it has no client waiting on it) rather than sharing the
   interactive default.

Happy to take either as a follow-up PR.

### DIFF
--- a/hindsight-api-slim/hindsight_api/alembic/versions/c4f7a2e9b6d1_drop_vestigial_memory_links_entity_fk.py
+++ b/hindsight-api-slim/hindsight_api/alembic/versions/c4f7a2e9b6d1_drop_vestigial_memory_links_entity_fk.py
@@ -1,0 +1,165 @@
+"""Drop the vestigial ``memory_links.entity_id`` → ``entities.id`` FK.
+
+``fk_memory_links_entity_id_entities`` was installed by the initial schema
+migration (``5a366d414dce``) with ``ON DELETE CASCADE``, back when entity edges
+were materialised as ``memory_links`` rows with ``link_type = 'entity'``. That
+representation is gone: ``e1b2c3d4f5a6`` dropped the entity index and
+``e9b2c7d1f3a4`` deleted every ``link_type = 'entity'`` row, and no writer has
+populated ``entity_id`` since — ``_bulk_insert_links`` passes ``None`` for every
+row it inserts. On the bank this was diagnosed against,
+``SELECT count(*) FROM memory_links WHERE entity_id IS NOT NULL`` returns 0 out
+of 26,750,591 rows and ``pg_stats.null_frac`` for the column is 1.
+
+The constraint protects nothing, but it is not free. ``memory_links`` has no
+index leading on ``entity_id`` (``e1b2c3d4f5a6`` dropped the last one), so
+PostgreSQL's referential-integrity trigger has to sequentially scan the whole
+table once per deleted ``entities`` row. Measured on a fully-cached ~2.7 GB
+``memory_links``: **1.278 s per parent row deleted**.
+
+That makes ``prune_orphan_entities`` (graph maintenance Pass 2) unrunnable on
+any bank with a real orphan backlog. The anti-join that finds the orphans is
+cheap (~70 ms, indexed by ``idx_unit_entities_entity_unit``), but the DELETE it
+drives pays the cascade scan per row: ~157 orphans ≈ 200 s against asyncpg's
+60 s client-side ``command_timeout`` (``DEFAULT_DB_COMMAND_TIMEOUT``). The sweep
+times out, rolls back, and retries — ``_SWEEP_MAX_RETRIES = 8`` means nine
+attempts and ~560 s of wasted work per job. Because it rolls back it never makes
+partial progress, so past roughly 47 orphans a bank can never recover on its
+own, and ``prune_stale_cooccurrences`` (same transaction, runs second) stops
+running entirely.
+
+What this migration does:
+
+* Drops ``fk_memory_links_entity_id_entities`` on PostgreSQL, idempotently
+  (``DROP CONSTRAINT IF EXISTS``) — the constraint has already been dropped by
+  hand on at least one production database as an emergency unblock, so the
+  migration must reconcile a schema that is *ahead* of head rather than fail on
+  it.
+
+What this migration deliberately does NOT do: drop the ``entity_id`` column.
+See the comment above ``_pg_schema_prefix`` below for why.
+
+PostgreSQL-only, by the same reasoning as ``9f8e7d6c5b4a``: Oracle keeps
+``idx_ml_entity`` on ``memory_links(entity_id)`` (``o1a2b3c4d5e6``), so its
+``fk_ml_entity`` cascade is an index lookup, not a table scan — the pathology
+being fixed here does not exist there, and the Oracle baseline still creates the
+constraint as part of the table DDL.
+
+Revision ID: c4f7a2e9b6d1
+Revises: b3e8d1c6f4a9
+Create Date: 2026-08-10
+"""
+
+from collections.abc import Sequence
+
+from alembic import context, op
+
+from hindsight_api.alembic._dialect import run_for_dialect
+
+revision: str = "c4f7a2e9b6d1"
+down_revision: str | Sequence[str] | None = "b3e8d1c6f4a9"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+_FK_NAME = "fk_memory_links_entity_id_entities"
+
+# Why the ``entity_id`` COLUMN stays, even though it is 100% NULL and nothing
+# reads it:
+#
+# 1. It is a key column of ``idx_memory_links_unique``
+#    ``(from_unit_id, to_unit_id, link_type, COALESCE(entity_id, nil_uuid))``,
+#    which is the arbiter of the retain hot path's
+#    ``ON CONFLICT ... DO NOTHING`` in ``PostgresOps.bulk_insert_links``.
+#    Dropping the column drops that index, and rebuilding a unique index over
+#    26.7M rows has to happen CONCURRENTLY or it takes an ACCESS EXCLUSIVE lock
+#    on the busiest table in the schema for the duration of the build.
+# 2. The ON CONFLICT target must then change in lockstep, in the same deploy, in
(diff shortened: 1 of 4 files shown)


mekenthompson:
Closing — this belongs in our own tree, not upstream.

PR #3305, opened 2026-08-10
fix(llm): forward extra body to Codex requests

### LINKED ISSUE #3298: [Feature]: Support Codex Fast mode for openai-codex
### Use Case

I use `openai-codex` with `gpt-5.6-luna` for Hindsight retain and reflect operations. I would like to request Codex Fast mode while continuing to use ChatGPT OAuth authentication.

### Problem Statement

Hindsight currently has no way to set a service tier for `openai-codex`.

Although `HINDSIGHT_API_LLM_EXTRA_BODY` is carried through the general LLM configuration, the `openai-codex` provider does not receive it. Its normal and tool-call request payloads therefore cannot include `"service_tier": "priority"`.

Codex CLI’s `/fast` setting does not apply because Hindsight reads `auth.json` for credentials and sends its own requests.

Codex is already a supported Hindsight provider and provides a low-friction way for ChatGPT subscribers to try Hindsight.

Supporting Fast mode for eligible models and accounts would improve that integration and could increase Hindsight’s adoption and visibility among Codex users.

### Proposed Solution

Add an optional Codex service-tier setting, or support `extra_body` in `CodexLLM`, and apply it to both normal and tool-call requests.

The setting should remain opt-in.

### Alternatives Considered

`HINDSIGHT_API_LLM_EXTRA_BODY` is not currently supported by `openai-codex`. Codex CLI configuration is also not inherited by Hindsight.

### Additional Context

Confirmed on Hindsight 0.9.0 and current `main`.

koriyoshi2041:
## What

Forward `HINDSIGHT_API_LLM_EXTRA_BODY` to the `openai-codex` provider and merge it into both normal and tool-call request payloads. This allows opt-in Codex request fields such as `{"service_tier": "priority"}` without adding a provider-specific setting.

Closes #3298.

## Tests

- `uv run pytest tests/test_codex_extra_body.py tests/test_codex_reasoning_effort.py tests/test_codex_tool_choice.py tests/test_codex_strict_schema.py tests/test_llm_extra_body.py -q` (43 passed)
- Ruff check and format check on the changed Python files
- Repository pre-commit hooks

## Risk

Low. Empty configuration preserves the existing payload, and the new path only copies configured extra fields into the two Codex request bodies.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
+++ b/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
@@ -381,6 +381,7 @@ def create_llm_provider(
             base_url=base_url,
             model=model,
             reasoning_effort=reasoning_effort,
+            extra_body=extra_body,
         )
 
     elif provider_lower == "claude-code":
--- a/hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/codex_llm.py
@@ -127,6 +127,7 @@ class CodexLLM(LLMInterface):
         base_url: str,
         model: str,
         reasoning_effort: str = "low",
+        extra_body: dict[str, Any] | None = None,
         **kwargs: Any,
     ):
         """Initialize Codex LLM provider."""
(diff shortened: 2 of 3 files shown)


nicoloboschi (approved):
Reviewed the change against the full provider dispatch in `create_llm_provider`. Correct and minimal — with this, `extra_body` reaches every provider that has a request body except `llamacpp` (tracked separately); `claude-code`/`mock`/`none` have no HTTP body to merge into, so excluding them is right.

Two non-blocking nits for a follow-up:
- The provider lists in `llm_wrapper.py` (`extra_body` docstring) and `hindsight-docs/docs/developer/configuration.md` still enumerate the supported providers without Codex.
- Precedence differs between the two call sites: in `call` the `payload.update()` runs *before* the forced-tool block (tools win), in `call_with_tools` it runs *after* `tools`/`tool_choice` are set (extra_body could clobber them). Worth making consistent.

PR #3306, opened 2026-08-10
feat(mental-models): graded write-time identifier-retention gate — refuse refreshes that silently drop the document's anchors

JoshFunnell:
## What

A graded write-time **identifier-retention gate** for mental-model refreshes: a refresh that drops too many *anchored identifiers* (dates, file paths, commit shas, UUIDs, URLs, env vars, ports, versions) from existing content is refused with the existing preserve-and-fail contract; smaller losses proceed with a warning that names the lost identifiers verbatim.

> **Stacked on #3290.** Most of this branch is that PR's two-tier work plus catch-up merges, so the commit count is misleading. The only commit that is **this** PR's own change is `17fb9908 feat(mental-models): graded write-time identifier-retention gate` — reviewing that one commit reviews this PR.
>
> Everything else is #3290 carried forward: merges of `main` and of the two-tier branch, and the CI fixes that landed there (an i18n prose-check fix, a duplicate OpenAPI enum title that broke Rust client generation, and the regenerated spec/client artifacts). All of it disappears from this diff once #3290 merges.
>
> _(Corrected 2026-08-11: this note previously said "only the top commit is new here", which stopped being true once `main` was merged in to resolve conflicts.)_

## Why

A refresh can lose its anchors while the document *grows*, so neither the empty-candidate guard nor any length check can see it. Measured over 364 real refresh events on one production bank:

| lost identifiers | events |
|---|---|
| 0 | 337 (92.6%) |
| 1 | 17 |
| 2 | 3 |
| **3+** | **7 (1.9%) — the class worth refusing** |

One of those seven dropped three file names and a commit sha from a single model in one write, while the document got *longer*. Losing one or two is frequently legitimate churn (a superseded date, a path that stopped being relevant), which is why the gate is graded rather than absolute: refusing on any loss would fail ~1 refresh in 14 and quickly be switched off.

## How

- **Pure logic** in `engine/identifier_retention.py` (no DB, no LLM): extract identifier sets from previous/candidate content, set-difference them (an identifier that *moves* counts as kept), grade against a threshold.
- **Call site** in `_execute_mental_model_refresh`, deliberately **after** the empty-candidate guard and the #3112 delta-window guard, so those more precise refusals win. An engine-level test pins the ordering.
- **Own outcome value** `refresh_failed_identifier_retention` (enum + OpenAPI + control-plane types/labels updated): the refused candidate is definitionally non-empty, so reusing `refresh_failed_empty_candidate` would make anything keyed on the outcome enum misclassify these. The raised error names the lost identifiers.
- **Warning persisted** under `reflect_response.identifier_retention` on warn and refuse paths alike, so the signal is readable without `keep_trace`.
- **Bootstrap-safe**: with no real previous content (`PENDING`/empty), the gate never fires — same condition as the sibling guard.
- **Knob**: `HINDSIGHT_API_MENTAL_MODEL_IDENTIFIER_LOSS_REFUSE` (default `3`; `0` = warn-only). Documented in `configuration.md` + `env.example`.

## Tests

- 14 unit tests on the pure module: taxonomy coverage, moved-identifier-counts-as-kept (the main false-positive direction), single-loss-warns, threshold edges incl. `0`, bootstrap-never-refused, env-knob parsing (garbage falls back rather than breaking a write path), and a pinned taxonomy limit (the port class needs a host prefix).
- 2 engine-level dry-run tests: the refusal reports its own outcome end to end with content preserved, and `refresh_failed_delta_not_applied` wins over identifier loss when both apply (guard ordering).
- Full mental-model selection green locally (292 passed; the only errors are pre-existing tests requiring a live `HINDSIGHT_API_LLM_API_KEY`).


nicoloboschi:
the root issue has been fixed already in 0.9.2

Issue #2973, opened 2026-07-26
Transient quota exhaustion is dead-lettered as a permanent consolidation failure; no per-memory transaction boundary means the safe alternative duplicates observations


bchumerch:
## Environment

`ghcr.io/vectorize-io/hindsight:0.8.4`, Postgres backend, docker-compose, single worker.
LLM provider `claude-code` on a Claude subscription (session + weekly usage limits, no API key).
`ENABLE_OBSERVATIONS=true`, semantic dedup at the 0.97 default.

## Summary

Two defects on the same chain. The first strands memories permanently; the second is why an
operator cannot patch around the first.

### 1. A transient quota error is recorded as a terminal failure

When the subscription's session or weekly limit is hit, every consolidation LLM call fails for the
duration of that window. `_consolidate_batch_with_llm` treats it as a content failure: retries,
bisects the batch, and at `batch_size=1` writes `consolidation_failed_at` on the memory row.
Memory selection requires `consolidation_failed_at IS NULL`, so the row is excluded from every
future run.

Nothing distinguishes "this memory cannot be consolidated" from "the provider was unavailable for
four hours." A quota window that lasts long enough to drain the retry budget converts the entire
pending set to dead letters.

**Measured on our deployment (2026-07-26):** 4,013 memories across 10 of 22 banks. On two banks
`failed_consolidation` exactly equalled the prior `pending_consolidation` (531/531, 158/158) — the
whole pending set stamped in one outage window. Recovery required `POST
/v1/default/banks/{bank}/consolidation/recover` per bank, which is operator-initiated and easy to
miss: 1,654 of those were on banks nobody thought to check, and had been dead for weeks.

The reconcile added in #2019 does not help here — it selects rows that are eligible but
unscheduled, and a dead-lettered row is not eligible.

**Suggested fix:** classify provider quota/rate-limit exhaustion as retryable-later rather than
terminal, and defer the operation (the engine already has `DeferOperation` /
`ProviderRateLimitResetError` plumbing for exactly this) instead of consuming retries and writing
`consolidation_failed_at`. Failing that, have the maintenance loop periodically re-arm
dead-lettered rows whose failure was provider-side.

### 2. There is no transaction boundary around one memory's processing

We tried to fix (1) with a provider-level overlay and could not, for a reason that seems worth
reporting on its own.

Within a single consolidation operation, creates and updates commit scope-by-scope and
[...]

PR #3309, opened 2026-08-10
fix(consolidation): don't dead-letter memories for a provider-side failure

JoshFunnell:
## The problem

`consolidation_failed_at` is a **permanent** exclusion — a stamped row is never selected again. Today it is applied to any exhausted-retry LLM failure, including ones the memories had nothing to do with.

When a provider hits a quota window or rate limit, the batch fails and the adaptive splitter bisects it down to size 1. Bisection isolates nothing here, because every call fails for the same reason — it just multiplies the failing calls — and each single-memory batch then gets stamped. Healthy memories are dropped from consolidation permanently, and the run that dropped them looks like an ordinary run with some failures.

This is the dead-lettering half of #2973.

## The change

Classify the failure instead of assuming the content is at fault:

| kind | bisect? | stamp `consolidation_failed_at`? | rows |
|---|---|---|---|
| `transient` — quota, rate limit, capacity, transport | no | no | stay pending for the next run |
| `auth` — bad/expired credentials | no | no | stay pending, **batch aborts loudly** |
| `content` — everything else, including anything unrecognised | yes | yes (unchanged) | unchanged |

Three details that are easy to get wrong, and why they are the way they are:

**Type before status before text.** A typed `RateLimitError` very often has an empty `str()`. A message-only classifier returns `content` for it — the worst misclassification available here, because it dead-letters healthy rows on the single most unambiguous transient signal a provider can send.

**Status codes need word boundaries.** `"429" in text` also fires on a token count of `14290` or a byte offset. That is the same substring false-positive that rules out bare `auth` and `connection` as markers, so codes are matched with `\b`.

**`auth` is its own kind, not a flavour of transient.** Credentials do not heal by retrying, so classifying them transient means the job silently retries a dead credential every run forever, recording nothing anywhere. They are not `content` either — the rows did nothing wrong and must not be permanently excluded for an environment problem. So the rows stay pending and the batch aborts with an error log.

**Unknown stays `content`**, preserving today's behaviour for anything the classifier does not recognise.

## Threading, not raising

`_process_memory_batch` now returns the failure **kind** instead of a bool — the caller has to tell "these rows are bad" from "the provider is down", and a boolean cannot carry that. The three call-contract tests are updated to the new shape.

The kind is threaded to the decision site rather than raised: a mid-loop raise would skip `consolidated_at` finalization for sub-batches that already wrote, and the next run would re-create their observations.

Sub-batch failures aggregate as a **set** of kinds, and a content failure anywhere outranks a co-occurring transient one. With two booleans OR'd together, one transient pass would suppress bisection for a sibling pass that failed on content, leaving genuinely unprocessable rows retrying forever.

## Storm brake

Not stamping on transient failures also removes the accidental rate limiting that permanent exclusion used to provide: without a brake, every later run reselects the same rows and re-hits a provider that is already refusing traffic. Three consecutive transient sub-batches now abandon the rest of the batch for this run.

## Scope

The other residual #2973 names — no per-memory transaction boundary, so a partially-written sub-batch may re-create some observations when it re-runs — is **unchanged**, and still absorbed by semantic dedup. Stamping on transient would "fix" it only by silently losing the un-consolidated part, which is the worse trade.

## Tests

New `tests/test_consolidation_failure_classification.py` covers both misclassification directions: a typed error with an empty message, status codes needing word boundaries (`14290` must not read as a 429), the rate-limit phrasings a plain substring list misses (`rate_limit_exceeded`, `ratelimited`), and the rule that an unrecognised failure stays `content`. `classify_llm_failure` is a pure function so none of this needs a database or an LLM.

Full local run: 50 passed in the consolidation suites. Two failures in `test_claude_code_llm_isolation.py` reproduce identically on unpatched `main` and are unrelated to this change.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -18,6 +18,7 @@ NOTE: Observations are distinct from mental models (pinned reflections).
 import asyncio
 import json
 import logging
+import re
 import time
 import uuid
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Closing this — we don't want failure-classification logic this brittle on the consolidation path, and the recovery path already exists.

**The recovery path today**

`POST /v1/default/banks/{bank_id}/consolidation/recover` clears `consolidation_failed_at` (and `consolidated_at`) for every stamped `experience`/`world` row in the bank and returns `{"retried_count": N}`, so they're picked up on the next consolidation run. No observations are deleted. It's exposed everywhere an operator would look for it:

- CLI: `hindsight bank consolidation-recover <bank_id>`
- Control plane: "Recover consolidation" on the bank page and in the bank stats view
- Python / TypeScript SDKs: `recover_consolidation` / `recoverConsolidation`

So a quota window that dead-letters a pending set is recoverable in one call per bank, without any inference about *why* the provider said no.

**Why not the classifier**

The classifier is the part that worries us, not the goal. It reads provider exception types, then status codes with `\b` word boundaries, then message substrings (`rate_limit_exceeded`, `ratelimited`, …) — i.e. it depends on the wire vocabulary of every provider we support, across every version of every SDK, forever. The PR body is honest about how narrow the margins are: a typed error with an empty `str()`, `14290` reading as a `429`. Those are the cases we found. The ones we didn't find fail silently and in the more expensive direction: a transient failure misread as `content` still dead-letters healthy rows, and now it does so through code that claims to prevent exactly that. Add the storm brake, the set-of-kinds aggregation, and the `auth`-aborts-loudly branch, and this is a lot of provider-shaped surface sitting in the middle of consolidation for a condition that a single operator call already resolves.

We'd rather have a dumb, predictable rule (unknown failure ⇒ stamp ⇒ visible in `failed_consolidation` ⇒ recover) than a smart one whose failure mode is invisible.

**What we'd take instead**

The real gap in #2973 isn't that recovery doesn't exist — it's that nobody knows to call it. `failed_consolidation` is already in bank stats; the missing piece is surfacing it (alerting / a visible warning when a run stamps a large fraction of its pending set) so the endpoint gets used the same day rather than weeks later. That's observability, not classification, and it doesn't need to know a single provider's error vocabulary.

Thanks for the writeup — the measurements and the three-shapes-we-tried section in #2973 are what made this decidable.

Issue #3307, opened 2026-08-10
mental_models search state is not maintained and reconciliation still targets reflections


JiehoonKwak:
## Bug description

`mental_models` has backend-specific searchable derived state, but its write paths and runtime reconciliation no longer maintain that state consistently.

This is independent from, but a prerequisite to fully fixing, #3268: a backend-aware query can stop the HTTP 500 while still returning empty or stale lexical results.

## Root causes

On current `main`:

1. `MemoryEngine.create_mental_model()` writes `embedding` but never writes `search_vector`. This leaves VChord `bm25vector` rows `NULL` because that column is not generated.
2. `update_mental_model()` re-embeds only when `content` changes and builds that embedding from the incoming name (or an empty string), not the canonical stored name + content document. It does not update `search_vector`; name-only updates do not refresh either derived representation.
3. `clear_mental_model()` clears `content` but leaves the old embedding and lexical state searchable.
4. `rename_knowledge_node()` changes `knowledge_pages.name` only; for page nodes the backing `mental_models.name` remains stale.
5. `ensure_text_search_extension()` still checks and rebuilds `reflections`, although migration `t5o6p7q8r9s0` renamed that table and its indexes to `mental_models`. The current table is silently skipped, and the failure guidance still tells operators to delete from `reflections`.
6. The original native column is generated, but the runtime reconciliation path recreates native `search_vector` as a plain `tsvector`, so the same write-path omission also matters after reconciliation.

## Expected behavior

Mental-model search should use one canonical document (`name + content`) and maintain its embedding and backend-specific lexical projection together whenever that document changes.

Knowledge Page renames should keep the visible page name and backing mental-model name synchronized atomically.

## Minimal fix boundary

- Add one backend-aware mental-model search-document expression/writer at the engine/database boundary; do not hard-code a specific backend in Knowledge Pages code.
- Use the canonical stored `name` and `content` for create, content-only update, name-only update, refresh, and clear.
- For native and VChord, write `search_vector` in the same transaction as the canonical row update. For backends that index base columns directly (`pg_textsearch`, `pgroonga`, `pg_search`), leave the dummy column alone.
- Make page rename update `knowledge_pages.name` and its backing `mental_models.name` in one transaction.
- Change runtime text-search reconciliation from `reflections` to `mental_models`, align index names/expressions, and backfill missing derived state only when the configured backend already matches the physical column/index type.
[...]

PR #3313, opened 2026-08-10
fix(search): keep mental-model derived state current

JiehoonKwak:
## Summary

- rebuild a mental model's embedding from the canonical stored `name + content` document on both name-only and content-only updates
- re-embed the name-only document and refresh the VectorChord lexical projection when content is cleared
- atomically synchronize a Knowledge Page rename with its backing mental-model name, embedding, and lexical projection

## Root cause

After #3318, VectorChord lexical projections are written on normal create/update paths, but partial updates can still derive dense search state from incomplete text. Name-only updates do not re-embed, clear leaves the old dense vector searchable, and a Knowledge Page rename changes only `knowledge_pages.name` while its searchable backing row retains the old name.

The fix resolves the complete stored document before embedding, keeps provider calls outside pooled database connections, and reuses the shared `pg_search_vector_expr()` helper added upstream for the only application-maintained lexical column (VectorChord). Native PostgreSQL updates its generated column; the other backends index base columns.

Part of #3307.

## Scope

This PR intentionally does not change Knowledge Page query dispatch, runtime backend reconciliation, or perform an in-place text-backend conversion. #3318 owns query dispatch; safe PGroonga physical reconciliation is a separate atomic change.

## Verification

- `./scripts/hooks/lint.sh` (Ruff, formatting, ty, ESLint/Prettier): passed
- related mental-model CRUD, clear, and Knowledge Page move/rename/delete tests: 17 passed
- new regressions cover name-only and content-only canonical embedding input, clear, and page-rename synchronization

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -12792,34 +12792,43 @@ class MemoryEngine(MemoryEngineInterface):
                 await self._validate_operation(self._operation_validator.validate_bank_write(ctx))
         backend = await self._get_backend()
 
-        # Compute the new embedding BEFORE acquiring a pooled connection: a slow
-        # embedder must never pin a DB connection. The embedding text depends only
-        # on the incoming name/content, never on DB state, so it can be done here.
+        # Resolve the complete searchable document before embedding it. A name-only
+        # or content-only update must include the stored half; otherwise the dense
+        # embedding and lexical projection describe different text. Keep the slow
+        # provider call outside pooled connections.
+        previous_content: str | None = None
+        previous_reflect_response: dict[str, Any] | None = None
+        effective_name = ""
+        effective_content = ""
         new_embedding_str: str | None = None
-        if content is not None:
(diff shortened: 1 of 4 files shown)


nicoloboschi:
Closing in favour of #3928, which fixes the same defect against the current code.

The diagnosis here still held: `update_mental_model` embedded only the text a single call supplied, so a content-only update dropped the stored name from the vector, a name-only update did not re-embed at all, and `clear_mental_model` left the vector built from the discarded content behind.

What changed since this was opened is the Knowledge Page half. `rename_knowledge_node` no longer exists — the KB patch (`update_knowledge_page`) now renames the backing model by delegating to `update_mental_model` on its own connection, in the same transaction as the node row, so `mm.name` and `search_vector` already sync. Only the embedding was still missing there, and it falls out of the general fix. That, plus the branch having gone conflicting, is why this is a fresh PR rather than a rebase: the version here re-acquires its own connection for the pre-read, which would break the transactional `conn=` path that fix introduced.

Tracked as #3926. Thanks for the report and the diagnosis — both carried over.

Issue #3308, opened 2026-08-10
whole-bank transfer skips Knowledge Pages and does not regenerate mental-model search state


JiehoonKwak:
## Bug description

Whole-bank export/import does not preserve the Knowledge Pages tree and does not regenerate the derived search state of restored mental models.

Current `main` explicitly puts `knowledge_pages` in `_SKIP_TABLES` because the generic row restorer cannot satisfy its self-referential parent FK. At the same time, export strips `embedding` and `search_vector` from carried `mental_models`, while import restores those rows without running the target-side mental-model embedding/search-document lifecycle.

The result is a logically incomplete bank after transfer:

- folders/pages, parent relationships, `managed`, and `sort_order` are lost;
- backing mental models may exist but are no longer connected to Knowledge Page nodes;
- restored mental models can have missing embedding and lexical search state;
- a target using another text backend cannot regenerate the backend-specific projection.

## Expected behavior

A whole-bank transfer should preserve the complete logical Knowledge Pages tree while treating embeddings, lexical projections, and physical indexes as target-derived state.

## Minimal fix boundary

- Add typed Knowledge Page rows to the transfer archive schema; do not pass raw unvalidated dictionaries between phases.
- Restore the bank row and mental models before Knowledge Pages.
- Restore folders in deterministic parent-first/topological order, then pages after both their parent folder and backing mental model exist.
- Preserve IDs, bank isolation, `parent_id`, `mental_model_id`, `managed`, `sort_order`, names, and timestamps according to the existing whole-bank transfer contract.
- Regenerate restored mental-model embeddings and backend-specific lexical state on the target through the same canonical lifecycle used by normal writes.
- Preserve mental-model history, bank config, directives, and webhooks under their existing options and ordering.
- Keep this separate from #3270, whose root cause is a target-bank `internal_id` collision swallowed by `ON CONFLICT DO NOTHING`. If #3270 is still open, treat it as a prerequisite for same-instance transfer tests rather than duplicating its fix here.
- No source-backend physical index or `search_vector` payload should be trusted or copied.

## Acceptance coverage

- Nested folders restore parent-first with zero orphans.
- Pages retain their backing mental-model relationship and bank isolation.
- Page names/content are searchable after import on native and VChord, plus a base-column backend path.
- Mental-model embeddings are regenerated with the target embedding provider; lexical state matches the target text backend.
- `skip`, `replace`, and `new-id` conflict modes preserve referential integrity or fail explicitly.
[...]

PR #3314, opened 2026-08-10
fix(search): use backend-aware Knowledge Page queries

### LINKED ISSUE #3268: knowledge-base search 500s on every non-native text-search backend: search_knowledge_pages hard-codes ts_rank_cd/tsvector
`GET /v1/{tenant}/banks/{bank}/knowledge-base/search` returns HTTP 500 for every bank (empty or populated) whenever `HINDSIGHT_API_TEXT_SEARCH_EXTENSION` is anything other than `native`. The knowledge-base tree and page-read endpoints work fine; only search fails.

## Environment

hindsight 0.9.0, PostgreSQL 18.1 + pgvector + ParadeDB pg_search, `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=pg_search`.

## Repro

1. Deploy 0.9.0 with `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=pg_search` (e.g. `docker/docker-compose/pg_search/`).
2. Create a bank; optionally add a knowledge page.
3. `GET /v1/default/banks/<bank>/knowledge-base/search?q=test&limit=5`

**Expected:** 200 with ranked pages.

**Actual:** 500 —

```
asyncpg.exceptions.UndefinedFunctionError: function ts_rank_cd(text, tsquery) does not exist
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

## Root cause

`search_knowledge_pages` (`hindsight_api/engine/memory_engine.py:13331` in the v0.9.0 slim tree) hard-codes the native-backend SQL in both arms — the RRF BM25 arm (:13377, :13381-13382) and the embedding-unavailable fallback (:13404, :13407):

```sql
ORDER BY ts_rank_cd(mm.search_vector, websearch_to_tsquery('english', $3)) DESC
...
AND mm.search_vector @@ websearch_to_tsquery('english', $3)
```

But under `pg_search`, `mental_models.search_vector` is by design a dummy `TEXT` column. From `alembic/versions/n9i0j1k2l3m4_learnings_and_pinned_reflections.py`:

```python
elif text_search_ext == "pg_search":
    # ParadeDB pg_search: dummy TEXT column; BM25 index over (id, name, content)
    ... ADD COLUMN search_vector TEXT
```

A `tsvector` column is created only in the `else: # native` branch. That column carries through the `pinned_reflections` → `reflections` (`p1k2l3m4n5o6_new_knowledge_architecture.py:47`) → `mental_models` (`t5o6p7q8r9s0_rename_mental_models_to_observations.py:71`) renames, so today's `mental_models.search_vector` is the same column.

This is not pg_search-specific: `pg_textsearch` and `pgroonga` also get `TEXT` dummy columns and `vchord` gets `bm25vector`, so `ts_rank_cd` fails to resolve on all four. Only `native` works.

Every other full-text-search path already dispatches on the backend correctly — e.g. `engine/sql/postgresql.py:211-224` has an explicit pg_search branch using `paradedb.score(id)` and `id @@@ paradedb.boolean(should => ARRAY[paradedb.match('text', ...), ...])`. This one function is the only spot missing the dispatch: it contains zero references to `text_search_extension` or `paradedb`. That is also why recall works on an affected deployment while knowledge search 500s, and why tree/page-read are unaffected (no FTS involved).

Minor related note: the function's docstring describes "a generated tsvector over the page name + content", which is only true under `native`.

## Suggested fix

Branch on `get_config().text_search_extension`, the way `PostgreSQLDialect.bm25_subquery` already does. The `idx_mental_models_text_search` BM25 index already covers `(id, name, content)` with `key_field='id'`, so for pg_search the BM25 arm becomes roughly:

```sql
bm AS (
    SELECT kp.id AS page_id,
           ROW_NUMBER() OVER (ORDER BY paradedb.score(mm.id) DESC) AS rnk
    FROM {join}
    WHERE kp.bank_id = $2 AND kp.kind = 'page'
          AND mm.id @@@ paradedb.boolean(should => ARRAY[
              paradedb.match('name', $3), paradedb.match('content', $3)])
    ORDER BY paradedb.score(mm.id) DESC
    LIMIT {fetch}
)
```

with analogous branches for `pg_textsearch` / `pgroonga` / `vchord`. A minimal alternative would be to degrade to the vector-only arm on non-native backends rather than returning 500.

## Possibly related latent issue

`ensure_text_search_extension` (`hindsight_api/migrations.py:818-822`) still lists `reflections` in `tables_to_check`, but `t5o6p7q8r9s0` renamed that table to `mental_models`. Backend-switch reconciliation therefore skips it silently ("Table reflections does not exist ... skipping"), so switching text-search backends would never reconcile the mental_models table either way.

JiehoonKwak:
## Summary

- move Knowledge Page lexical matching and scoring into the SQL dialect
- provide explicit plans for native, VChord, pg_textsearch, PGroonga, and pg_search
- use the same plan in hybrid and embedding-unavailable text-only paths
- exclude backend-specific padded or zero-score non-matches

## Root cause

`search_knowledge_pages` always emitted native PostgreSQL `tsvector` SQL. On VChord this compares a `bm25vector` with a `tsquery` and returns HTTP 500; the same architectural mismatch affects every non-native backend. The dialect now owns operators, score direction, filtering, and indexed expressions.

Closes #3268.

## Verification

- backend SQL regression coverage for all five PostgreSQL text-search backends
- targeted mental-model, Knowledge Page, transfer, and DB-abstraction pytest suite: 136 passed on the combined integration branch
- Ruff and ty checks passed for changed production files
- disposable PG18 + PGroonga HTTP canary returned the expected page for Korean-only, English-only, and mixed queries, and `EXPLAIN` used `idx_mental_models_text_search`

This PR is independent of the transfer fix in #3308.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -13333,11 +13333,10 @@ class MemoryEngine(MemoryEngineInterface):
     ) -> list[dict[str, Any]]:
         """Doc-level hybrid search over a bank's knowledge pages.
 
-        Fuses a full-text match (``mm.search_vector``, a generated tsvector over the
-        page name + content) with vector similarity (``mm.embedding``) using
-        Reciprocal Rank Fusion, in a single round trip. No reranker — this path is
-        tuned for latency. Returns pages ranked by fused score, each with a short
-        content snippet. Folders are excluded.
+        Fuses backend-specific lexical search with vector similarity
+        (``mm.embedding``) using Reciprocal Rank Fusion, in a single round trip.
+        No reranker — this path is tuned for latency. Returns pages ranked by
+        fused score, each with a short content snippet. Folders are excluded.
         """
         await self._authenticate_tenant(request_context)
         query = (query or "").strip()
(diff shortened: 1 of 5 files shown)


JiehoonKwak:
Superseded by merged #3318 for backend-aware Knowledge Page query dispatch and draft #3335 for the remaining real-PGroonga query plus rollback-safe mental_models reconciliation. Closing this broader implementation so review can continue on the two minimal boundaries; no commits from this branch are being merged.

Issue #3270, opened 2026-08-08
hindsight-admin import-bank results into a ForeignKeyViolationError


tonnyb:
### Bug Description

When using import-bank on a exported zipped bank results in a 

ForeignKeyViolationError: insert or update on table "mental_models" violates foreign key constraint "fk_mental_models_bank_id"

Method used on the docker-container `ghcr.io/vectorize-io/hindsight`

Version it started happening: 0.8.5, still exists in 0.9.0
Version 0.8.4 works just fine.

tbf i only tried docker containers for this.

```
hindsight@hindsight:~$ hindsight-admin export-bank -b memory --output memory.zip
Exporting bank 'memory' from schema 'public'...
Starting embedded PostgreSQL (instance: hindsight)...
[transfer] Skipped 14 observation(s) with sources outside the exported documents
[transfer] Exported bank elle: 9 document(s), 60 fact(s), 37 observation(s), 1 mental model(s), 9 directive(s), 1 webhook(s)
Exported bank 'memory' to memory.zip (71882 bytes)
hindsight@hindsight:~$ hindsight-admin import-bank -a memory.zip --target-bank test

--- big stacktrace --- 

ForeignKeyViolationError: insert or update on table "mental_models" violates foreign key constraint "fk_mental_models_bank_id"
DETAIL:  Key (bank_id)=(test) is not present in table "banks".
```

### Steps to Reproduce

docker compose exec -ti [service] bash

hindsight-admin export-bank -b [existing bank] --output bank.zip

hindsight-admin import-bank -a bank.zip --target-bank [non existing bank/test_bank]

### Expected Behavior

Documents, Observations, Bank configuration, Mental Model, Bank successfully imported

### Actual Behavior

ForeignKeyViolationError: insert or update on table "mental_models" violates foreign key constraint "fk_mental_models_bank_id"
DETAIL:  Key (bank_id)=(test_bank) is not present in table "banks".

### Version

0.9.0

### LLM Provider

Other - Gemma4-E2B  (local)

ebarkhordar:
Reproduced on `0.9.0`. The trigger is that the target instance still holds the bank the archive came from.

Clean `ghcr.io/vectorize-io/hindsight:0.9.0` container, embedded pg0, migrations applied, one bank `memory` holding one mental model and no documents:

```
$ hindsight-admin export-bank -b memory --output /tmp/memory.zip
[...]

PR #3315, opened 2026-08-10
fix(transfer): preserve Knowledge Page search state

### LINKED ISSUE #3308: whole-bank transfer skips Knowledge Pages and does not regenerate mental-model search state

JiehoonKwak:
## Summary

- include Knowledge Pages and mental-model history in whole-bank archives
- validate page/model references and restore the page tree parent-first
- re-embed carried mental models with the target provider and rebuild the target text projection
- preserve bank configuration and optional operational history while retaining the fresh-target preflight

## Root cause

Whole-bank export skipped `knowledge_pages` and stripped mental-model derived state. Import restored mental-model rows without regenerating embeddings or the target backend projection, so a cross-instance migration could lose the page tree and leave its search state unusable.

Closes #3308.

## Dependency

Depends on #3313 for the backend capability used to rebuild mental-model search state. This branch contains that prerequisite commit; the transfer-specific review commit is `07d091b2a`.

## Verification

- document-transfer suite: 27 passed
- targeted mental-model, Knowledge Page, transfer, and DB-abstraction suite: 136 passed on the combined integration branch
- Ruff and ty checks passed for changed production files
- fresh PG18 + PGroonga target canary restored 1 mental model, 2 Knowledge Page nodes, 1 mental-model history row, and 5 operational-history rows
- source and target page/model checksums and counts matched; embeddings were regenerated; parent links and Korean/English/mixed HTTP search survived a database restart

The same-instance clone collision tracked in #3270 is a separate existing lane. This PR targets the documented cross-instance migration path and does not duplicate that work.

### DIFF
--- a/hindsight-api-slim/hindsight_api/admin/cli.py
+++ b/hindsight-api-slim/hindsight_api/admin/cli.py
@@ -847,6 +847,7 @@ def import_bank_command(
         f"Imported bank '{result.bank_id}': {result.documents_imported} doc(s), "
         f"{result.facts_imported} fact(s), {result.observations_imported} observation(s), "
         f"{result.mental_models_imported} mental model(s), "
+        f"{result.knowledge_pages_imported} Knowledge Page node(s), "
         f"{result.mental_model_history_imported} mm-history row(s), {result.directives_imported} directive(s), "
         f"{result.webhooks_imported} webhook(s), {result.history_rows_imported} history row(s)"
     )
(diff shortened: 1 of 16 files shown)


JiehoonKwak:
Superseded by #3330, which implements the same transfer goal with a smaller typed Knowledge Page payload and target-derived mental-model search state. I left the only material remaining safety delta (pre-write tree/model validation for corrupt archives) as a focused review comment on #3330. Closing this stacked branch avoids duplicating the maintainer-owned lane.

PR #3316, opened 2026-08-10
fix(docker): repair PGroonga Compose image

### LINKED ISSUE #3311: PGroonga Docker Compose recipe references a nonexistent base-image tag
## Problem

`docker/docker-compose/pgroonga/Dockerfile` currently starts with:

```dockerfile
FROM groonga/pgroonga:latest-debian-pg17
```

That tag has no registry manifest, so the documented PGroonga Compose recipe cannot build. A current tag such as `groonga/pgroonga:4.0.8-debian-17` does resolve.

## Current evidence

Reproduced on 2026-08-10:

```text
docker buildx imagetools inspect groonga/pgroonga:latest-debian-pg17  # fails
docker buildx imagetools inspect groonga/pgroonga:latest-debian-17    # succeeds
```

The upstream PGroonga image matrix currently names Debian tags as `<pgroonga-version>-debian-<postgres-major>` / `latest-debian-<postgres-major>`, and publishes both amd64 and arm64 variants for PostgreSQL 17 and 18.

The recipe also compiles the older pgvector v0.8.0 even though the PGroonga Debian base already exposes the PGDG package repository (`postgresql-17-pgvector` is available), so the build-tool/git stages are avoidable.

## Minimal fix boundary

- Keep this recipe on PostgreSQL 17 for parity with the neighboring examples; do not combine a major-version migration with the tag repair.
- Replace the invalid floating base tag with a current explicit PGroonga 4.0.8 / PG17 tag (optionally digest-pinned according to repository policy).
- Install pgvector from the matching PGDG package instead of cloning/building it.
- Add a build smoke check that creates both `vector` and `pgroonga` in a disposable database.

A separate deployment can use the same recipe with the official PG18 tag; that should not broaden this fix into a cross-example PostgreSQL upgrade.

JiehoonKwak:
## Summary

- replace the nonexistent `groonga/pgroonga:latest-debian-pg17` base with the published pinned `groonga/pgroonga:4.0.8-debian-17` image
- install the pinned PostgreSQL 17 pgvector 0.8.6 package from PGDG

## Root cause

The shipped PGroonga recipe cannot build because its base-image tag does not exist. Pinning the actual PGroonga Debian 17 tag also makes the example reproducible instead of following an ambiguous latest tag.

Closes #3311.

## Verification

- registry inspection confirmed the published PGroonga tag is multi-architecture, including arm64
- disposable arm64 PostgreSQL smoke test loaded PGroonga 4.0.8 and pgvector 0.8.6
- Korean/English `&@~` matching and pgvector distance queries passed

The local PG18 integration image used for downstream migration rehearsal is intentionally not included; this PR only repairs the existing upstream PG17 recipe.

### DIFF
--- a/docker/docker-compose/pgroonga/Dockerfile
+++ b/docker/docker-compose/pgroonga/Dockerfile
@@ -3,21 +3,11 @@
 # pgroonga is a multilingual full-text search extension built on Groonga.
 # It works out of the box for CJK (Chinese, Japanese, Korean) and other
 # non-whitespace-segmented languages via the TokenBigram tokenizer.
-FROM groonga/pgroonga:latest-debian-pg17
+FROM groonga/pgroonga:4.0.8-debian-17
 
 # Install pgvector on top of the pgroonga base image (which already provides
-# pgroonga and the Groonga library).
+# pgroonga, the Groonga library, and the PostgreSQL PGDG package repository).
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    build-essential \
-    git \
-    postgresql-server-dev-17 \
+    postgresql-17-pgvector=0.8.6-1.pgdg13+1 \
+    && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
-
-RUN cd /tmp && \
-    git clone --branch v0.8.0 https://github.com/pgvector/pgvector.git && \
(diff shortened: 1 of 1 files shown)


nicoloboschi (approved):
thanks!

PR #3317, opened 2026-08-10
fix(transfer): preserve Knowledge Pages during bank transfer (#3308)

### LINKED ISSUE #3308: whole-bank transfer skips Knowledge Pages and does not regenerate mental-model search state

Sanderhoff-alt:
## Summary

- Carry `knowledge_pages` in whole-bank archives and restore nodes parent-first.
- Rebuild restored mental-model embeddings and target-side lexical state.
- Preserve generated PostgreSQL columns while explicitly rebuilding regular native and VectorChord projections.

## Approach

Whole-bank archives intentionally omit `embedding` and `search_vector`. Mental models are restored as rows, so import rebuilds this derived state on the target after row restoration. Embeddings are generated before taking the write connection.

Knowledge Page nodes are restored after their backing mental models. A small topological pass orders folder parents before descendant folders and pages.

## Scope

This PR is intentionally limited to the transfer boundary. It does not change normal mental-model create/update/clear/rename lifecycle behavior, startup reconciliation, or database abstractions.

PR #3315 takes a broader approach and depends on #3313 for lifecycle-wide search-state maintenance. That work may be valuable independently. For this transfer-specific bug, this PR aims to keep review and rollout simpler by reusing the existing PostgreSQL search-vector expression helper, checking whether the target column is generated, and avoiding a dependency on the broader lifecycle change.

Knowledge Page query adaptation for non-native PostgreSQL text backends remains outside this PR.

## Verification

- Nested folder/page export-delete-import round trip preserves IDs, hierarchy, page-to-model references, managed state, and sort order.
- Parent-first ordering has a focused regression test.
- Mental-model search-state tests cover generated columns and VectorChord SQL.
- Ruff, ty, and repository lint pass.

Closes #3308.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -4419,8 +4419,9 @@ class MemoryEngine(MemoryEngineInterface):
         """Restore a whole bank from an :func:`transfer.export_bank` archive.
 
         Re-embeds facts with this instance's embedding model and rebuilds links and
-        indexes; restores bank config, mental models, directives and webhooks as
-        exported (no consolidation/webhooks — a migration restores exact state). The
+        indexes; restores bank config, mental model logical fields, directives and
+        webhooks as exported while rebuilding target-derived mental model search
+        state (no consolidation/webhooks — a migration restores exact state). The
         target bank must not already exist (import restores a whole bank, not a merge).
         """
         from .transfer import import_bank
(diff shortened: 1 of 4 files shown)


Sanderhoff-alt:
Closing this PR because #3330 has been merged and provides an equivalent fix, covering Knowledge Pages transfer and target-side mental-model search-state regeneration.

Issue #3291, opened 2026-08-09
Knowledge Pages staleness flag saturates permanently on multi-scope banks (follow-up to #3139/#3156): measured 24/29 "may need refresh" for hours while the exact check reports 0/29


JoshFunnell:
## Summary

The Knowledge Pages tree's per-page staleness flag (`is_stale` from `list_knowledge_nodes(with_staleness=true)`) is computed from the bank-global write watermark, per the documented rule in the mental-models docs ("Staleness Gating"). On a bank with diverse tag/entity scopes, that flag **saturates permanently**: the dashboard showed "24 of 29 models may need refreshing" continuously for 5+ hours on a quiet bank, while the exact per-model check said **0 of 29** were stale. This is a follow-up to #3139 (thank you for the fast fix in #3156 — the missing-field defect is fully resolved; this issue is about the approximation the fix chose).

## Mechanism

`_may_need_refresh(last_refreshed_at, watermark)` flags every model refreshed before the bank's newest write, in-scope or not. The docstring is explicit and honest about the tradeoff: "True only means *something* was written; it may well be outside the model's tags" — and the docs state the reason (a scoped check per page would cost one scan each per poll).

The trouble is the flag can only clear when a quiet model refreshes — and the event-driven staleness gate *correctly refuses* to refresh a model whose scope has nothing new. So on any bank where writes are spread across scopes, the tree converges to "almost everything may need refreshing" as its permanent steady state, and the indicator's TRUE direction stops carrying information (its FALSE direction stays exact). The better the refresh gating gets, the longer the saturated display persists.

## Measurement (production bank, single-tenant PG, 29 mental models)

| Instrument | Result |
|---|---|
| Bank write watermark | 21:17:33 (last write; nothing for the following 5+ hours) |
| Tree flag (`_may_need_refresh`) | **24 / 29 "may need refresh"**, unchanged for 5+ hours |
| Exact per-model `is_stale` (single-model reads, same engine, same moment) | **0 / 29 stale** |

The 24 are exactly the models whose `last_refreshed_at` predates the final write burst — their scopes had nothing new, so nothing will ever clear them short of an off-scope refresh.

## Suggested directions (any of these would resolve it; happy to PR whichever you'd prefer)

1. **UI/copy only, zero engine cost:** render the two states distinguishably — "provably fresh" (exact) vs "unknown" (approximate) — instead of one alarm-colored "may need refresh"; optionally sort provably-fresh first. Removes the false alarm without new queries.
[...]

PR #3325, opened 2026-08-10
fix(knowledge-pages): show exact freshness in page detail

Sanderhoff-alt:
## Why

The Knowledge Pages tree reports freshness using a bank-wide write watermark. That is intentionally conservative: a page can be marked `is_stale: true` even when the new memory is outside the page's scope.

When a page is opened, the control plane already fetches its backing mental model to show provenance. That response includes the exact scope-aware `is_stale` value, but the page header was still reading the tree's approximate value. As a result, a page could show “May need refresh” in its detail view even when its exact freshness check reported that it was current. This makes the detail view appear inconsistent and reduces trust in the status indicator.

## What changed

- Prefer the exact backing mental model freshness result for the open page.
- Fall back to the tree watermark result while the exact result is unavailable.
- Keep the tree polling strategy and API behavior unchanged.

This makes the detail view use information it already retrieved, so there is no additional request or database cost. It does not attempt to change the tree's deliberately conservative semantics.

## Verification

- `./scripts/hooks/lint.sh`
- `npm test` (119 tests)

### DIFF
--- a/hindsight-control-plane/src/components/knowledge-base-view.tsx
+++ b/hindsight-control-plane/src/components/knowledge-base-view.tsx
@@ -96,6 +96,10 @@ export function KnowledgeBaseView() {
   // Derived from the backing mental model's reflect_response (not the page endpoint).
   const [supportingCount, setSupportingCount] = useState(0);
   const [selectedMmId, setSelectedMmId] = useState<string | null>(null);
+  const [exactPageStaleness, setExactPageStaleness] = useState<{
+    pageId: string;
+    isStale: boolean;
+  } | null>(null);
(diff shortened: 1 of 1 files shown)


nicoloboschi:
Thanks for this. The underlying complaint — that page freshness in the tree was only an approximation — was fixed server-side by #3589 (#3291): staleness is now answered per scope rather than from the bank watermark, so the `is_stale` the tree returns is already the exact answer. The bank watermark survives only as a free fast path to skip the scoped query when nothing in the bank changed at all.

That makes the client-side override here redundant. Closing as already fixed.

PR #3330, opened 2026-08-10
feat(transfer): carry Knowledge Pages tree + regenerate mental-model search state on import (#3308, #3323)

### LINKED ISSUE #3308: whole-bank transfer skips Knowledge Pages and does not regenerate mental-model search state

### LINKED ISSUE #3323: Bank import does not recompute mental_models embedding / search_vector — knowledge search degraded after import
## Summary

Bank **export/import** (`hindsight_api/engine/transfer/`) strips derived columns (`embedding`, `search_vector` — `export.py:_DERIVED_COLUMNS`) from `mental_models` rows and, unlike the archive-restore path for `memory_units`, never recomputes them on import. As a result, a freshly **imported** bank's knowledge pages (pinned `mental_models`) are degraded until each page is next refreshed:

- **`mm.embedding` is NULL** after import → the vector arm of `search_knowledge_pages` (`WHERE mm.embedding IS NOT NULL`) returns nothing, on **every** text-search backend. Recall of the page content via embeddings is also affected.
- **`mm.search_vector` is empty** after import for **vchord** specifically (its `bm25vector` column is written by tokenizing on write; `native` regenerates automatically since its column is GENERATED, and `pg_search`/`pg_textsearch`/`pgroonga` index base columns). So the vchord BM25 arm also returns nothing.

The import docstring (`importer.py:import_bank`) and the export comment (`export.py:72`) both claim these rows are "re-embedded on import" / "search/vector indexes rebuilt", but no such recompute happens for `mental_models` (only `memory_units` facts are re-embedded, and only the archive→memory_units restore recomputes `search_vector`).

## Repro

1. On instance A, create a bank with a knowledge page (pinned mental model) that has content + embedding.
2. `export_bank` → `import_bank` into instance B.
3. `GET /v1/{tenant}/banks/{bank}/knowledge-base/search?q=...` on B → the imported page does not rank (empty vector arm; empty BM25 under vchord) until the page is refreshed.

## Expected

Import should rebuild `mental_models.embedding` with the target instance's embedding model (as it does for facts) and recompute `mental_models.search_vector` for the target text-search backend (as the `memory_units` archive-restore path already does via `pg_search_vector_expr`).

## Notes

- Surfaced while fixing #3268 (PR #3318). That PR fixes the **live** write paths (create/update/clear) so vchord tokenizes `mm.search_vector` correctly; the **import** path is a separate, broader gap (embedding dominates) and was intentionally left out of scope.
- The `search_vector` recompute is a one-liner mirroring `writes.py` (the archive→memory_units restore already calls `pg_search_vector_expr(get_config())` post-insert); the `embedding` recompute needs the target embedding model, like the fact re-embed path.

nicoloboschi:
Fixes #3308. Closes #3323.

## Problem

Whole-bank export/import produced a logically incomplete bank:

1. **The Knowledge Pages tree was dropped entirely** — `knowledge_pages` sat in `_SKIP_TABLES` because its self-referential `parent_id` FK needs a parents-first restore the generic per-row restorer didn't provide. Folders, page nodes, `parent_id`, `mental_model_id`, `managed`, and `sort_order` were all lost; backing mental models survived but were disconnected from any page.
2. **Restored mental models had no derived search state** — export strips `embedding`/`search_vector` (target-derived), but import restored the rows without regenerating either. So imported knowledge pages had a NULL embedding (vector arm empty on every backend) and, under vchord, empty lexical state — i.e. unsearchable (this is #3323).

## Fix

**Export** (`transfer/export.py`, `schema.py`)
- Add a typed `TransferKnowledgePage` model (no raw dicts across phases, per the issue) and carry the tree in `knowledge_pages.json`, emitted parent-first via a recursive walk, preserving `id`, `parent_id`, `mental_model_id`, `managed`, `sort_order`, `name`, and timestamps.
- Remove `knowledge_pages` from `_SKIP_TABLES`; classify it under a new `KNOWLEDGE_TABLES` bucket (the `test_export_bank_covers_schema` guard is updated so a future table still can't be silently dropped).

**Import** (`transfer/importer.py`)
- **Regenerate each restored mental model's embedding** with the *target* model, embedding the same `"{name} {content}"` text `create_mental_model` uses. Computed off-connection so no DB connection is held across the embedding call (the established retain-path rule).
- **Rebuild backend-specific lexical state** through the shared `pg_search_vector_expr` — the single source of truth the live mental-model writes and the memory-recall path use. Only vchord needs an explicit write (its `bm25vector` column); native's `search_vector` is GENERATED and repopulates on insert, and pg_search/pg_textsearch/pgroonga index the base `name`/`content` columns.
- **Restore the tree** after its backing mental models exist (page→mm FK) and parents-first (`parent_id` self-FK), via a topological order that tolerates cycles/dangling parents (emits them so the DB FK — not a silent drop — surfaces a corrupt export). `ON CONFLICT DO NOTHING` keeps the import idempotent.

`import_bank` reports a new `knowledge_pages_imported` count (surfaced by the admin CLI). No HTTP surface changed — whole-bank import is admin-CLI only, so no OpenAPI/client regen.

## Tests

- **Whole-bank roundtrip** (`test_bank_roundtrip_carries_knowledge_pages`): builds nested folders + pages (including a root-level page and a `managed` folder), exports → deletes → imports, then asserts the tree restores **exactly** (ids, `parent_id`, `mental_model_id`, `managed`), that pages are **searchable after import**, and that no restored mental model has a NULL embedding.
- **Non-DB unit tests** for the topological ordering (parent-first; tolerant of cycles/dangling parents).
- Ran the full `test_document_transfer.py` suite (28 tests) green on an isolated pg0 instance.

## Scope note

Verified searchable-after-import end-to-end on the native backend (the default test backend). The vchord/base-column lexical regeneration goes through the same `pg_search_vector_expr` dispatch already unit-tested per backend in #3268; a live vchord backend isn't part of the transfer test matrix.

### DIFF
--- a/hindsight-api-slim/hindsight_api/admin/cli.py
+++ b/hindsight-api-slim/hindsight_api/admin/cli.py
@@ -847,7 +847,8 @@ def import_bank_command(
         f"Imported bank '{result.bank_id}': {result.documents_imported} doc(s), "
         f"{result.facts_imported} fact(s), {result.observations_imported} observation(s), "
         f"{result.mental_models_imported} mental model(s), "
-        f"{result.mental_model_history_imported} mm-history row(s), {result.directives_imported} directive(s), "
+        f"{result.mental_model_history_imported} mm-history row(s), "
+        f"{result.knowledge_pages_imported} knowledge page(s), {result.directives_imported} directive(s), "
         f"{result.webhooks_imported} webhook(s), {result.history_rows_imported} history row(s)"
     )
 
--- a/hindsight-api-slim/hindsight_api/engine/transfer/export.py
+++ b/hindsight-api-slim/hindsight_api/engine/transfer/export.py
@@ -31,6 +31,7 @@ from .schema import (
     TransferChunk,
     TransferDocument,
     TransferFact,
+    TransferKnowledgePage,
     TransferManifest,
     TransferObservation,
     TransferObservationSource,
@@ -88,12 +89,6 @@ _SKIP_TABLES = frozenset(
         # to fresh ids, so carrying them would only produce dangling associations.
         # Revert anything worth keeping on the source before migrating.
         "invalidated_memory_units",
-        # Knowledge-base folder/page tree (client-managed metadata over the carried
-        # mental models). Not carried yet: its self-referential parent_id FK needs a
-        # parents-first (topological) restore order, which the generic per-row
-        # _restore_rows doesn't provide — a follow-up. The mental models themselves
-        # ARE carried, so the target can recreate the tree.
-        "knowledge_pages",
     }
 )
 # Derived columns dropped from carried rows so the target regenerates them with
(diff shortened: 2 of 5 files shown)


JiehoonKwak:
I compared this against #3315; this PR is the better, smaller lane and covers the material transfer behavior (typed tree, target-bank remap, target embeddings/search projection, parent-first restore, history/config/webhook paths already present on main).

One safety delta from #3315 seems worth folding in before merge: validate the typed Knowledge Page graph and backing-model references before the first import write.

Today `_load_knowledge_pages()` walks only from roots, so a cyclic component would be omitted from the archive. On import, `_topological_page_order()` deliberately emits dangling/cyclic leftovers and relies on the FK to fail, but that happens after the bank/documents/facts have already been restored (and their embeddings paid for), leaving a partially imported target.

A narrow preflight can reject, before any write/provider call:

- duplicate/missing node ids, dangling parents, and parent cycles;
- non-folder parents;
- folders carrying `mental_model_id`;
- pages without a carried backing mental model.

The corresponding regression should assert a corrupt archive causes zero writes and zero embedding calls. This does not require changing the valid parent-first restore path or broadening this PR.

PR #3342, opened 2026-08-10
fix(coding-agents): report configured MCP harness

TheAngryPit:
## Summary

- Propagate the configured MCP harness into `buildKnowledgeTools`.
- Make `hindsight_diagnose` report the actual harness instead of `unknown`.
- Add a regression for the Codex MCP path.

## Root cause

`mcp-server.ts` resolved the configured harness but did not pass it into the knowledge-tool builder. Runtime session events therefore knew they were running under Codex while the generic diagnostic tool lacked that context.

## Validation

```text
Focused MCP server tests
  3 tests passed

TypeScript --noEmit
  passed

Production build
  passed

Fresh installed MCP stdio process
  hindsight_diagnose returned harness=codex
```

## Scope

- No server, database, retain, hook, configuration-schema, or dependency change.
- The change only propagates already-resolved runtime context to the existing diagnostic surface.


TheAngryPit:
Updated onto current main at 1ffe8759 (Coding Agents v0.1.2). Fresh validation on the updated branch: 40 test files, 404 tests passed; TypeScript noEmit passed; production build passed; diff check clean. The PR diff remains limited to mcp-server.ts and its regression test.

nicoloboschi:
Reviewed and merging. I pushed one test to the branch rather than asking for a round-trip.

Two notes for the record:

**The harness source is right, and I checked the part that looks wrong.** A quick read suggests `cfg.harness` would be `"opencode"` — that is `resolveConfig`'s default, and `mcp-server.ts` keeps the real harness in a local (`HINDSIGHT_MCP_HARNESS || "claude-code"`) that it never assigns to the config. It works because `loadConfig` back-fills it:

```ts
// The harness that ASKED is the correct fallback for an unset `harness` field — not a hardcoded
// "opencode", whose silent default misfiled every other harness's background seed… (#3247)
if (!raw.harness && o.harness) raw.harness = o.harness;
```

So `cfg.harness` in an MCP process is the launching harness, and it is the same source every other call site uses. Consistent, not lucky.

**This fixes more than the title says.** `opts.harness` has two consumers in `buildKnowledgeTools`: the diagnose report, and `hindsight_ingest_document`, which stamps the `harness:<id>` tag and `metadata.harness`. Documents ingested through the MCP tool were previously **unattributed**, and the documents list resolves each document's agent logo and filter from exactly those fields. That is the more user-visible half, so I added a test for it — a regression there would have looked like documents quietly losing their agent, with the diagnose test still green.

Thanks for the fix, and for rebasing onto v0.1.2.

TheAngryPit:
Thank you, @nicoloboschi  — I really appreciate the careful review and the additional regression test. You’re absolutely right that document attribution is the more user-visible consequence, and it’s valuable to have both the diagnostic output and the ingestion metadata covered.
Thanks as well for tracing cfg.harness through loadConfig and confirming that the behavior is intentional rather than accidental. Glad to see this merged.

PR #3343, opened 2026-08-10
fix(coding-agents): route repository surveys explicitly

TheAngryPit:
## Summary

- Route the four canonical repository-survey documents through the `survey` strategy instead of generic document ingestion.
- Attach deterministic `source:survey`, harness, retain-attribution, and structural knowledge tags.
- Ground Component map, Core concepts, and Conventions page queries in their canonical survey evidence using repository-generic wording.
- Make sync status, crashed-survey retry detection, and completion-marker handling recognize `source:survey` while retaining compatibility with legacy `source:upload` banks.
- Keep the router's document IDs compile-time aligned with the survey producer's canonical `SURVEY_DOC_IDS` contract.

## Root cause

`hindsight_ingest_document` treats every title as a generic uploaded document. The repository survey uses that public tool for four fixed findings, so vanilla ingestion classifies them as `document` / `source:upload`, while the seeded Knowledge Pages expect deterministic structural tags. Page population therefore depends on incidental extraction instead of the native survey contract.

## Current-main refresh

This branch was rebuilt on current `main` (`64ede361`) instead of mechanically preserving the historical implementation. It keeps the current mission/template ownership model, current retain attribution, and the current client contract. It does not restore removed retain options or unrelated historical template changes.

## Validation

```text
Focused survey/router/readers
  104 tests passed

Full Coding Agents suite
  45 files passed, 493 tests passed

TypeScript --noEmit
  passed

Production build and declarations
  passed

Prettier and git diff --check
  passed

Independent review
  safe to merge; no blockers
```

No live Hindsight retain or page-refresh mutation was performed during this refresh. The earlier local data remediation is not part of this source PR.

## Compatibility

- Non-survey `hindsight_ingest_document` calls retain `document` / `source:upload` behavior.
- Existing banks whose survey findings use `source:upload` remain supported.
- Existing user-owned missions are not overwritten.
- No server API, database schema, migration, dependency, or hook-configuration change.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/hindsight.pages.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/hindsight.pages.test.ts
@@ -26,8 +26,8 @@ describe("seeded knowledge-page queries", () => {
   ])("anchors %s to its survey findings and excludes repair chatter", (name, documentId) => {
     const page = PAGES.find((candidate) => candidate.name === name);
     expect(page?.source_query).toContain(documentId);
-    expect(page?.source_query).toMatch(/exclude|do not describe/i);
-    expect(page?.source_query).toMatch(/survey ingestion/i);
+    expect(page?.source_query).toMatch(/actual/i);
+    expect(page?.source_query).toMatch(/repair|remediation/i);
   });
 });
 
--- a/hindsight-integrations/coding-agents/src/core/hindsight.pages.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/hindsight.pages.test.ts
@@ -18,6 +18,19 @@ function stubFetch(calls: any[], jsonImpl: () => Promise<unknown> = async () => 
   );
 }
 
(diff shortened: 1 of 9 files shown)


TheAngryPit:
Updated onto current main at 3a4b0214 and resolved the retain-options conflict introduced by #3336: survey retains now use the current always-async client contract without the removed async option. I also rewrote the three seeded page queries to stay repository-generic while retaining deterministic anchors to the canonical survey documents; local repair, remediation, tag-repair, and harness-diagnosis wording is no longer part of the upstream queries. Fresh validation: focused routing/page/status/session tests 81/81; full Coding Agents suite 40 files and 412 tests; TypeScript noEmit and production build passed; diff check clean.

nicoloboschi:
I don't think this is necessary, any information about the project outside of the declared one should be treated as project data, I don't see any problem being solved here

PR #3356, opened 2026-08-10
fix(cli): exit non-zero when configure receives invalid API URL

Alan5168:
## Summary

`hindsight configure --api-url <invalid>` prints an error message but returns exit code 0. Scripts and CI pipelines relying on the exit status cannot detect the failure.

## Root cause

The URL validation block in `handle_configure()` calls `ui::print_error!` and then `return Ok(())`, so the error is logged but the process exits successfully.

## Fix

Replace the silent success with `anyhow::bail!`, which lets the top-level error handler print the message and exit with status 1 — consistent with every other error path in the CLI.

## Tests

Added two integration tests in `cli_profile.rs`:

- `configure_rejects_invalid_url_with_nonzero_exit` — verifies exit code != 0 and error message on stderr for a URL missing the scheme
- `configure_accepts_valid_http_and_https_urls` — regression: both `http://` and `https://` URLs still save successfully

All 13 tests in `cli_profile` pass:

```
cargo test --manifest-path hindsight-cli/Cargo.toml --test cli_profile
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

## Scope

Only touches the configure command's URL validation. No change to URL parsing logic, config saving, or any other command.

### DIFF
--- a/hindsight-cli/src/main.rs
+++ b/hindsight-cli/src/main.rs
@@ -2199,11 +2199,10 @@ fn handle_configure(
 
     // Validate the URL
     if !new_api_url.starts_with("http://") && !new_api_url.starts_with("https://") {
-        ui::print_error(&format!(
+        anyhow::bail!(
             "Invalid API URL: {}. Must start with http:// or https://",
             new_api_url
-        ));
-        return Ok(());
+        );
     }
 
     // Use provided api_key, or keep existing one if not provided
--- a/hindsight-cli/tests/cli_profile.rs
+++ b/hindsight-cli/tests/cli_profile.rs
@@ -319,3 +319,40 @@ fn hindsight_profile_env_var_is_honored() {
         + &String::from_utf8_lossy(&out.stdout);
     assert!(err.contains("from-env"), "expected profile URL in output:\n{}", err);
 }
+
+#[test]
+fn configure_rejects_invalid_url_with_nonzero_exit() {
+    let home = unique_tempdir("bad-url");
+    let out = run_with_home(
+        &home,
+        &[
+            "configure",
+            "--api-url",
+            "not-a-valid-url",
+            "--api-key",
+            "test-key",
+        ],
+    );
+
+    assert!(
+        !out.status.success(),
+        "configure with invalid URL should exit non-zero, got success"
+    );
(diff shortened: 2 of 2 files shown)


koriyoshi2041 (approved):
Reviewed at `924aa3b1`. Propagating the validation failure through `anyhow::bail!` gives the CLI the expected non-zero status while keeping the error on stderr. The added tests cover both rejection and the existing HTTP/HTTPS success paths. Local `cargo test --test cli_profile` passed, and `git diff --check origin/main...HEAD` is clean.

handnewb (approved):
LGTM! Small focused fix — CLI should exit non-zero on configuration error, not silently return 0. Good test coverage with the new cli_profile test.

nicoloboschi (approved):
Verified locally on top of current `main` (PR is 180 commits behind, but `handle_configure` is unchanged there so it cherry-picks cleanly):

- `cargo test --test cli_profile` → 13 passed, 0 failed
- Reverting only `main.rs` makes `configure_rejects_invalid_url_with_nonzero_exit` fail, so the test genuinely gates the fix
- Manual run confirms a single stderr message and exit status 1, and no config file is written on the failure path

`anyhow::bail!` matches the convention used throughout the CLI, and `profile create` already rejected bad URLs this way — so the two commands are now consistent. All `hindsight configure` call sites (CI, docs, guides) pass valid URLs, so nothing breaks.

Heads-up: `test-rust-cli` is gated on `has_secrets == 'true'`, so it was skipped on this fork PR and the new tests never ran in CI. They pass locally.

Non-blocking nits for a follow-up:
- `cli_profile.rs`'s module docstring still says it only covers the `profile` subcommands; worth widening it to mention `configure`.
- `configure_accepts_valid_http_and_https_urls` only asserts exit 0 — asserting the saved config contains the URL would make it a stronger regression guard.
- Pre-existing: `hindsight_binary()` reads `CARGO_BIN_EXE_hindsight` via `std::env::var` at runtime, but Cargo only sets it at compile time, so it silently falls back to whatever `hindsight` is on `$PATH`. Should be `env!("CARGO_BIN_EXE_hindsight")`.

PR #3358, opened 2026-08-10
fix(kb): serialize knowledge tree writes

Sanderhoff-alt:
## Summary

Serialize knowledge-base structural writes at the bank level so create, delete, and move operations cannot validate or rewrite the hierarchy from incompatible snapshots. This prevents concurrent opposite moves from both passing cycle detection and committing a parent loop.

## Problem

Knowledge-base hierarchy checks previously ran without a shared serialization point. Two transactions moving `A` under `B` and `B` under `A` could both read the same pre-commit tree, independently pass the existing cycle guard, and then commit a cycle. Create and delete also change the same hierarchy, so serializing move alone would leave the structural-write invariant incomplete.

## Design

Add `_kp_lock_bank()` and acquire a row lock on the owning bank before structural hierarchy reads or writes. Folder creation, page-node insertion, move, and delete all use the same lock, so structural writers for one bank execute against one committed tree state. Writers for different banks remain independent.

The lock uses `FOR NO KEY UPDATE` rather than `FOR UPDATE`. It still conflicts with another structural writer taking the same lock, but it does not conflict with the `FOR KEY SHARE` locks PostgreSQL takes for inserts into tables that reference the bank. This preserves writer serialization without unnecessarily blocking unrelated bank-scoped inserts or introducing the foreign-key lock interaction caused by the stronger lock. The Oracle adapter already rewrites this form to supported `FOR UPDATE` syntax.

Move takes the bank lock before validating the destination parent and reading the parent map. Delete takes the same lock before reading and removing a subtree. Create takes it before validating the parent and inserting the new hierarchy row. Non-structural operations such as rename do not take this lock.

## Concurrency behavior

| Step | First move: A under B | Second move: B under A |
| --- | --- | --- |
| Acquire bank lock | Proceeds | Waits on the same bank row |
| Read hierarchy | Sees both nodes at their original parents | Runs after the first transaction commits |
| Validate cycle | Passes | Sees `A.parent_id = B` and rejects the move |
| Write | Commits `A.parent_id = B` | Does not commit a cycle |

## Test coverage

Add a deterministic regression test that pauses the first request after it acquires the real bank-row lock, verifies the second request reaches the same lock and cannot finish early, then confirms the first move succeeds and the second move is rejected. The test uses events rather than sleeps, making the required interleaving explicit and repeatable.

## Validation

- `./scripts/hooks/lint.sh`
- `uv run pytest tests/test_knowledge_base.py -q -n 0` (40 passed)

## Scope and tradeoffs

The serialization scope is one bank row, so structural writes within the same bank are intentionally serialized. This is coarser than locking individual paths, but it keeps a uniform lock order across create, delete, and move, makes the tree invariant straightforward to audit, and avoids locking every knowledge node.

This change is intentionally limited to preventing concurrent structural writes from creating a parent loop. It does not add defensive traversal for trees already corrupted before the fix, change page-creation failure compensation, or alter existing exception semantics.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -15291,6 +15291,19 @@ class MemoryEngine(MemoryEngineInterface):
         if row["kind"] != "folder":
             raise ValueError(f"Parent '{parent_id}' is not a folder")
 
+    async def _kp_lock_bank(self, conn, bank_id: str) -> None:
+        """Serialize structural knowledge-tree writers on the bank row.
+
+        FOR NO KEY UPDATE conflicts with other tree writers but not with the
+        FOR KEY SHARE locks taken by inserts into tables that reference banks.
+        Oracle rewrites it to FOR UPDATE, which does not block indexed-FK child
+        inserts there.
+        """
+        await conn.fetchrow(
+            f"SELECT bank_id FROM {fq_table('banks')} WHERE bank_id = $1 FOR NO KEY UPDATE",
+            bank_id,
+        )
+
     async def create_knowledge_folder(
         self,
         bank_id: str,
@@ -15320,6 +15333,7 @@ class MemoryEngine(MemoryEngineInterface):
         async with acquire_with_retry(backend) as conn:
             async with conn.transaction():
                 await self._ensure_bank_exists(bank_id, request_context, conn=conn)
+                await self._kp_lock_bank(conn, bank_id)
                 await self._kp_assert_folder_parent(conn, bank_id, parent_id)
                 row = await conn.fetchrow(
                     f"""
@@ -15383,6 +15397,7 @@ class MemoryEngine(MemoryEngineInterface):
                 # share a transaction instead of compensating after a partial commit.
                 async with conn.transaction():
                     created = await self._ensure_bank_exists(bank_id, request_context, conn=conn)
+                    await self._kp_lock_bank(conn, bank_id)
                     await self._kp_assert_folder_parent(conn, bank_id, parent_id)
                     mm_row = await self._insert_pinned_mental_model(
                         conn,
@@ -15778,6 +15793,9 @@ class MemoryEngine(MemoryEngineInterface):
         backend = await self._get_backend()
         async with acquire_with_retry(backend) as conn:
             async with conn.transaction():
+                # Structural tree writers lock this bank row before reading or
+                # changing the hierarchy, serializing moves with creates/deletes.
+                await self._kp_lock_bank(conn, bank_id)
                 await self._kp_assert_folder_parent(conn, bank_id, new_parent_id)
                 # Cycle guard: walk up from the new parent; if we reach node_id,
                 # the move would create a loop. Done in Python so the check stays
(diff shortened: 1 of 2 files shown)


handnewb:
👋 Friendly suggestion from a contributor review: PRs #3358 (serialize writes), #3369 (atomic creation), and #3370 (compensation on failure) all address knowledge-base transaction safety. Consider consolidating these three into a single PR — it would reduce review overhead, prevent merge conflicts between them, and make the KB transaction story coherent. Each PR is valuable individually, but together they form a stronger atomicity guarantee.

nicoloboschi:
Reviewed — the design is right, and it matches the precedent already in `_submit_async_operation`, which locks the same row the same way for the same reason. `FOR NO KEY UPDATE` is the correct strength: it self-conflicts so two structural writers serialize, but doesn't conflict with the `FOR KEY SHARE` that FK inserts into `knowledge_pages` / `mental_models` take on the bank row. I also confirmed the embedding in `create_knowledge_page` is generated before the connection is acquired, so the serialized window stays pure SQL.

I've pushed two follow-ups to the branch:

**Bounded the two Python tree walks.** The lock stops *this* process from committing a parent loop, but a bank corrupted before the lock existed still carries one — and nothing in the schema forbids it, so `transfer/importer.py` will happily restore it from an export. Both walks were unbounded on such a tree: the move ancestor walk (`while cursor is not None`) spins if the chain cycles without passing through `node_id`, and the delete subtree BFS spins on a cycle in the subtree. Each one hangs inside an open transaction holding a pool connection, so a single corrupted bank permanently burns a connection per request. Move now refuses the move; delete visits each node once. Verified: with the guards reverted, the new test times out.

**Test changes.**
- Scoped the lock interception to the bank under test. Matching on the SQL text alone would catch any other `FOR NO KEY UPDATE` caller in the process — `_submit_async_operation`'s submit dedupe issues exactly that string — and pause an unrelated transaction as "the first writer".
- Replaced `assert not second.done()` with a real wait. It was asserted right after the event is set and before the blocking `fetchrow`, so the task provably could not have been done — it read like proof of blocking but wasn't.
- Added delete-vs-move and create-vs-delete races, so the claim that all three structural writers take the lock is tested rather than stated. Both assert the second writer gets a clean `ValueError` off the committed tree instead of validating a stale parent and hitting a raw FK violation on the write.
- Added the pre-existing-cycle test for the bounded walks.
- Extracted the pausing-connection harness into `_BankLockPauser` now that three tests use it.

Also noted in `_kp_lock_bank` that correctness rests on READ COMMITTED (the waiter re-reads the hierarchy per statement, so it sees the previous writer's commit), and that a missing bank row deliberately takes no lock.

One thing worth flagging for whoever merges: `test-api` is gated on `has_secrets`, so all three shards **skip** on fork PRs — the green check here never ran these tests. I ran the full suite against a dedicated pg0 instance locally (`test_knowledge_base.py` 60 passed, plus `test_mcp_tools.py` + `test_document_transfer.py`, 271 passed) and dispatched the full CI workflow on a repo-side mirror of this branch for real signal.

Out of scope here, but adjacent: `transfer/importer.py::_restore_knowledge_pages` writes `parent_id` verbatim with no hierarchy validation, so it's the one remaining way a cycle can enter a bank after this fix.

Issue #3220, opened 2026-08-06
[FEATURE] OpenCode plugin: autoRecall uses a fixed query once per session — not question-aware


rsaulo:
## Use Case

We run OpenCode as a chat agent inside a team channel (Buzz), one long-lived
process per repository, with `@vectorize-io/opencode-hindsight` providing shared
long-term memory across our other agents (Claude Code and Codex hit the same
bank through their own integrations).

People ask the agent direct questions in the channel and expect it to answer
from memory the way the Claude Code and Codex integrations do.

## Problem Statement

`autoRecall` recalls **once per session, with a hardcoded query**. In
`createHooks`:

```js
const systemTransform = async (input, output) => {
  if (!config.autoRecall) return;
  const sessionId = input.sessionID;
  if (state2.recalledSessions.has(sessionId)) return;   // once per session
  const query = `project context and recent work`;      // fixed
  const { context, ok } = await recallForContext(query);
  ...
};
```

That is session priming, not question-aware recall: no user question ever
influences what is retrieved. A direct question does not retrieve the memory
that answers it, even when that memory is the single best match in the bank.

Note the plugin already has query-composition machinery
(`composeRecallQuery`/`truncateRecallQuery`, `recallMaxQueryChars`,
`recallContextTurns`) — but it is only wired into
`experimental.session.compacting`, not into the per-turn path.

### Measurement

One bank, one fact, four query shapes. Position of the answering memory in the
recall results:

| Query | Position |
|---|---|
| The user's question alone | **#1** |
| Only the interrogative sentences of the user's message | **#1** |
| The whole user message (question + surrounding instructions, 170 chars) | #11 |
| `"project context and recent work"` (current behavior) | not returned |

With `recallMaxTokens: 512`, a match at #11 does not make it into the injected
context either — so in practice only the first two shapes work.

This is **not** a duplicate of #2656. That issue reports the `experimental.*`
hooks not existing in the Plugin API on OpenCode 1.16.2, so the hook never
fires. On 1.18.14 the hooks are present in `@opencode-ai/plugin`'s type
definitions and `experimental.chat.system.transform` does fire — the behavior
described here is what happens *after* it fires.

## How This Feature Would Help

The same bank is shared by three harnesses in our setup. Claude Code and Codex
answer questions from memory; OpenCode does not, and the difference is visible
to users in the same channel asking the same question of different agents.

## Proposed Solution

Per-turn recall driven by the user's message. Concretely, any of:

[...]

PR #3360, opened 2026-08-10
fix(claude-code,codex): expose retain turn interval via env

rsaulo:
Both the Claude Code and Codex integrations default `retainEveryNTurns` to `10`, and neither lists it in `ENV_OVERRIDES` — unlike neighbouring keys such as `HINDSIGHT_AUTO_RETAIN`. On a long-lived ACP host that makes retention unreachable rather than merely infrequent.

## The gap

`retain.py` gates on the cadence and returns early:

```python
retain_every_n = max(1, config.get("retainEveryNTurns", 1))
# Respect retainEveryNTurns in both modes, unless force=True (SessionEnd final retain)
if retain_every_n > 1 and not force:
    turn_count = increment_turn_count(session_id)
    if turn_count % retain_every_n != 0:
        return
```

The only escape for skipped turns is the `force=True` final retain on `SessionEnd`. That assumption holds for an interactive CLI, where the session ends and batching is a sensible cost trade. It does not hold for a host that keeps one session alive per channel or room: `SessionEnd` may never arrive, so every turn below the cadence threshold is skipped and never flushed.

The skip returns successfully and logs only at debug level, so nothing surfaces as an error. An operator discovers the missing writes only by querying the bank.

## Evidence

Observed on an ACP host with per-channel sessions. `autoRetain` was enabled via `HINDSIGHT_AUTO_RETAIN=true`, the hooks fired on `SessionStart`, `UserPromptSubmit` and `Stop` with the intended bank id in the environment, and no error appeared anywhere — yet the bank received zero memory units across several sessions. Setting `retainEveryNTurns` to `1` (at the time, by patching this same table locally) made retention work immediately, with the units landing in the expected bank.

Because the key was not env-overridable, the only workaround was editing the plugin config file — which lives in a cache directory and is lost on upgrade.

## What changed

- `HINDSIGHT_RETAIN_EVERY_N_TURNS` added to `ENV_OVERRIDES` in both integrations, typed `int`, matching the existing integer overrides. Invalid values are ignored exactly as the other integer keys already behave.
- README env tables updated in both integrations.
- Tests covering the conversion and the graceful discard of an invalid value.

The default stays `10`; this only makes it reachable without a config file, so persistent-session hosts can set `1`.

## Note

`hindsight-integrations/claude-code/tests/test_run_mcp.py::TestMcpServerWorkingDirectory::test_exec_runs_from_plugin_data_dir_when_project_env_is_unreadable` fails on this branch. It also fails on an unmodified `main` checkout, and the files it touches are untouched here — it is unrelated to this change.

### DIFF
--- a/hindsight-integrations/claude-code/README.md
+++ b/hindsight-integrations/claude-code/README.md
@@ -244,7 +244,7 @@ Auto-retain runs after Claude responds. It extracts the conversation transcript 
 |---------|---------|---------|-------------|
 | `autoRetain` | `HINDSIGHT_AUTO_RETAIN` | `true` | Master switch for auto-retain. Set to `false` to disable memory storage entirely. |
 | `retainMode` | `HINDSIGHT_RETAIN_MODE` | `"full-session"` | Retention strategy. `"full-session"` sends the initial transcript once, then only new messages in chunk documents; compaction starts a fresh chunk. |
-| `retainEveryNTurns` | — | `10` | How often to retain. `1` = every turn; `10` = every 10th turn. Higher values reduce API calls but delay memory capture. Values > 1 enable **chunked retention** with a sliding window. |
+| `retainEveryNTurns` | `HINDSIGHT_RETAIN_EVERY_N_TURNS` | `10` | How often to retain. `1` = every turn; `10` = every 10th turn. Higher values reduce API calls but delay memory capture. Values > 1 enable **chunked retention** with a sliding window. |
 | `retainOverlapTurns` | — | `2` | When chunked retention fires, this many extra turns from the previous chunk are included for continuity. Total window size = `retainEveryNTurns + retainOverlapTurns`. |
 | `retainRoles` | — | `["user", "assistant"]` | Which message roles to include in the retained transcript. |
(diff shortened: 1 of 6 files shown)


rsaulo:
Closing this myself — the surface it touches has been superseded, and I would rather not spend review on it.

Context for anyone who finds this later: `hindsight-integrations/claude-code` and `/codex` were both marked superseded by the Coding Agents plugin in #3162, and #3220 restated it — the legacy plugins keep working, but fixes land on the new one. The deployment this PR came from (an ACP host with one long-lived session per channel) has since migrated to `@vectorize-io/hindsight-coding-agents`, where recall and retain run per turn and `retainEveryNTurns` is no longer the gate. So the failure I was reporting is fixed for us by the migration rather than by this change, and I can no longer dogfood the patch on the path it modifies.

The underlying finding still stands for anyone on the legacy integrations with a session that never ends: `retain.py` returns early on every turn below the cadence, the only flush is the `force=True` retain on `SessionEnd`, and the skip logs at debug level — so the bank stays empty with nothing surfacing as an error. `retainEveryNTurns` was reachable only by editing a config file in a cache directory that upgrades wipe.

The branch stays up. If you would rather have the env override on the legacy plugins anyway, say so and I will reopen and rebase.

PR #3364, opened 2026-08-10
fix(coding-agents): bound automatic reflect budget

TheAngryPit:
## Summary

- Use Hindsight’s supported low reflect budget for the automatic first-prompt hook.
- Keep explicit `hindsight_reflect` MCP calls on the existing high-budget path.
- Preserve the 25-second client cap below the 30-second host hook timeout.

## Root cause

The automatic hook requested `budget: high` while enforcing a hard 25-second deadline. High-budget reflect may perform several LLM/tool iterations and can legitimately exceed that window. In a live comparison against the same self-hosted endpoint, a high-budget reflect took 35.169 seconds, while the equivalent low-budget control completed in 6.580 seconds. Concurrent page refreshes can worsen tail latency, but an idle control also exceeded 25 seconds, so increasing workers or changing the database would not fix the contract mismatch.

## Validation

- Focused hook and MCP tests: 37/37 passed.
- Full Coding Agents suite: 430/430 passed across 41 files.
- TypeScript `--noEmit`: passed.
- Production build: passed.
- Bundle inspection confirms the Codex hook uses low while the MCP tool remains high.

No timeout, hook configuration, server, database, provider, dependency, or schema change.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/hook.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/hook.test.ts
@@ -152,7 +152,7 @@ describe("buildHookOutput", () => {
     expect(client.reflect).toHaveBeenCalledTimes(1);
   });
 
-  it("caps the reflect timeout at 25000ms even when config asks for more", async () => {
+  it("uses a bounded low-budget reflect and caps its timeout at 25000ms", async () => {
     const cfg = resolveConfig({}); // reflectTimeoutMs default 120000
     const client = makeClient();
     await buildHookOutput({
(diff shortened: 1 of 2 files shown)


handnewb:
👋 Suggestion: PR #3364 (bound automatic reflect budget) and PR #3343 (route repository surveys explicitly) both harden the coding-agents subsystem. Consider combining into a single "coding-agents: budget and routing hardening" PR — both are small, same area, and a unified PR would be easier to review and test.

nicoloboschi (approved):

PR #3368, opened 2026-08-11
fix(webhooks): send signed payload in GET deliveries

Sanderhoff-alt:
## Context

GET webhook deliveries serialized an event and computed `X-Hindsight-Signature` over those bytes, but dispatched the request without a body. Receivers therefore observed an empty body, could not parse the event, and could not reproduce the signature from the bytes they received.

## Design

- Keep the serialized event bytes as the single source of truth for both signing and delivery.
- Dispatch GET webhooks through `AsyncClient.request` with `content=payload_bytes`, since the convenience `get` API does not accept request content.
- Preserve configured query parameters and the existing POST path.
- Force `Cache-Control: no-cache, no-store` for GET deliveries because HTTP caches generally do not include request content in their cache key and could otherwise suppress later events sent to the same URL.
- Continue routing requests through `GuardedAsyncTransport`, preserving the existing SSRF validation and IP pinning behavior.

## Contract and compatibility

The signature remains `sha256=HMAC-SHA256(secret, raw_request_body)`. Receivers must verify the exact body bytes before parsing JSON; reserialization is not equivalent. POST remains the default and recommended method because GET request bodies are not handled consistently by every server, proxy, or framework. The webhook documentation now makes this limitation explicit.

## Verification

- Added a transport-level regression test using `httpx.MockTransport` to inspect the final GET method, URL, query parameters, raw body, signature, and cache directive.
- Added coverage proving GET transport failures enter the existing retry schedule.
- Ran the complete webhook delivery test class: 7 passed.
- Ran Ruff checks and formatting checks for the changed Python files.
- Ran `git diff --check`.
- Built the Docusaurus documentation site successfully and regenerated the bundled documentation skill references.

## Scope

This change intentionally does not alter POST delivery semantics, webhook persistence, retry timing, signature format, or URL security policy.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -2898,7 +2898,12 @@ class MemoryEngine(MemoryEngineInterface):
                 "timeout": http_config.timeout_seconds,
             }
             if http_config.method.upper() == "GET":
-                response = await self._http_client.get(url, **request_kwargs)
+                # GET webhooks use the same signed-payload contract as POST webhooks,
+                # so the exact bytes covered by X-Hindsight-Signature must be sent. GET
+                # caches do not generally key on body content, so prevent a cached
+                # response from suppressing delivery of a later event to the same URL.
+                headers["Cache-Control"] = "no-cache, no-store"
+                response = await self._http_client.request("GET", url, content=payload_bytes, **request_kwargs)
             else:
                 response = await self._http_client.post(url, content=payload_bytes, **request_kwargs)
             response.raise_for_status()
--- a/hindsight-api-slim/tests/test_webhooks.py
+++ b/hindsight-api-slim/tests/test_webhooks.py
@@ -7,6 +7,8 @@ Covers:
 - HTTP API integration tests for CRUD and delivery listing endpoints
 """
 
+import hashlib
+import hmac
 import json
 import uuid
 from datetime import datetime, timezone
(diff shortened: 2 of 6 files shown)


nicoloboschi:
Thanks for the careful writeup here — the bug you identified is real, and I verified your fix works: it merges cleanly onto current `main`, the new test genuinely fails without the production change, and I confirmed over a real socket (not just `MockTransport`) that the body reaches the wire with `GuardedAsyncTransport` still applying the SSRF check and IP pinning. The diagnosis and the implementation are both sound.

I'm closing this anyway, because of what it does to existing GET webhooks.

Today a receiver configured with `http_config.method: "GET"` gets a bodyless request and returns 2xx. After this change it gets a GET carrying `Content-Length` and a JSON body. RFC 9110 §9.3.1 says content on a GET "has no generally defined semantics... and might lead some implementations to reject the request and close the connection" — and in practice a good number of front ends do exactly that, or strip the body before it reaches the app. So for existing users this splits two ways, and neither is good:

- **Body stripped in transit** (CloudFront, Cloudflare, API Gateway): the receiver still sees an empty body plus a signature over bytes it never got. The bug isn't fixed, it's relocated to the network.
- **Body rejected**: a delivery that succeeds today starts returning 400/411/413 and permanently fails after the 6-attempt retry schedule.

There's no opt-out — `http_config` has no flag to keep the current behavior — so every existing GET webhook is moved onto the new contract whether or not its stack can carry it.

The docs in this PR are honest about the limitation ("use POST unless the receiving stack is known to preserve GET bodies"), but that's guidance, not a requirement we state or can check. We'd be shipping a delivery mode whose correctness depends on a property of the receiver's infrastructure that we don't define, don't verify, and can't observe. That's not a contract I want webhook delivery to rest on.

What I'd rather see, if you want to keep going at this:

1. **Make GET honest instead.** There's no body on a GET, so don't send `Content-Type: application/json` or `X-Hindsight-Signature` with it. That removes the misleading signature — which is the actual defect you found — without depending on anything about the receiver.
2. Or **gate the body behind an explicit `http_config` flag**, so opting into GET-with-body is a deliberate choice by someone who knows their stack preserves it, and existing webhooks keep working untouched.

Separately: the "HTTP Delivery" docs section is worth keeping regardless. `X-Hindsight-Signature` was previously documented nowhere in the docs site (only in a blog post), and "verify against the exact request body bytes, re-serializing is not equivalent" is exactly the right thing to tell receivers. If you'd like to open that as a docs-only PR against the POST contract, I'll merge it.

One process note that isn't your fault: this is a fork PR, so all three `test-api` shards showed `skipping` and the Python suite never actually ran in CI. I ran it locally — 7/7 in `TestHandleWebhookDelivery`, ruff clean, and `generate-docs-skill.sh` produced no diff, so the generated references were correctly in sync. Worth knowing that green checks on a fork PR here don't mean the API tests passed.

Appreciate the contribution, and sorry to close after the effort you clearly put in.

PR #3369, opened 2026-08-11
refactor(kb): create pages and mental models atomically

Sanderhoff-alt:
## Summary

Create each knowledge page and its backing mental model in one database transaction so both records become visible together or both roll back.

> [!IMPORTANT]
> This PR and compensation PR #3370 are mutually exclusive alternatives for the same issue. Merge exactly one, not both. This atomic design is the stronger consistency option and has a larger refactoring surface.

## Problem

A knowledge page stores its content in a backing mental model. The current flow commits the mental model first and inserts the page in a later transaction. If parent validation, the page insert, the database connection, or task execution fails after the first commit, the bank can retain an orphaned mental model with no knowledge page referencing it.

The intended invariant is: a page and its backing mental model either both exist or neither exists.

## Design

Extract mental-model embedding generation and pinned-model insertion into typed internal helpers. `create_mental_model` continues to use those helpers, preserving its existing behavior, while `create_knowledge_page` uses the insertion helper through its own connection.

Generate the embedding before opening the transaction, then create the bank if needed, validate the parent, insert the mental model, and insert the page through one connection and one transaction. This keeps model inference outside the transaction while making all database writes atomic.

Run the default-bank-template hook only after commit because the hook opens independent connections. This preserves the existing lazy bank creation behavior without allowing nested connections to escape the page transaction.

## Failure semantics

Any exception or cancellation before commit rolls back the bank, mental model, and page writes made by this operation. No application-level cleanup is required.

Only a violation of `uq_kp_folder_pagename` is translated to `None`, preserving the HTTP 409 duplicate-page contract. Other unique-constraint violations continue upward instead of being mislabeled as page-name conflicts.

## Compatibility

The shared insertion helper retains the existing search-vector behavior and the existing PostgreSQL and Oracle SQL adaptation path. The public API and returned page shape do not change.

## Verification

- Knowledge-base suite: 44 passed
- Repository lint: passed
- Type checking for `memory_engine.py`: passed
- PostgreSQL regression coverage inserts a real mental model and verifies rollback after a later failure
- Oracle regression coverage is included and collected; local execution skips without `ORACLE_TEST_DSN`
- Parent validation, duplicate names, authorization counts, and lazy bank creation are covered

## Trade-offs

This design provides database-enforced atomicity across exceptions, cancellation, connection loss, and process termination. Its cost is a larger internal refactor because the existing mental-model insertion path must be reusable inside a caller-owned transaction.

PR #3370 keeps the existing transaction boundary and performs best-effort cleanup instead. It has a smaller production diff but cannot guarantee cleanup after process or database failure.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -51,7 +51,7 @@ from ..worker.exceptions import DeferOperation, RetryTaskAt, format_task_error
 from ..worker.stage import set_stage
 from .audit import AuditLogger, audit_context
 from .bank_stats_cache import BankStatsCache, DistributedBankStatsCache
-from .db import DatabaseBackend, create_database_backend
+from .db import DatabaseBackend, DatabaseConnection, ResultRow, create_database_backend
 from .db.ops_postgresql import pg_search_vector_expr
 from .db_budget import budgeted_operation
 from .llm_interface import ProviderRateLimitResetError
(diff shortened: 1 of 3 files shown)


nicoloboschi (approved):
Reviewed the diff and verified locally.

**Approach is correct.** `create_knowledge_page` now performs bank-create → parent validation → mental-model insert → page insert in a single `conn.transaction()`, so any failure or cancellation before commit rolls everything back — orphaned mental models are structurally impossible, not compensated after the fact. The shared `_insert_pinned_mental_model` helper keeps `create_mental_model`'s existing SQL/search-vector/Oracle behavior intact, and embedding generation stays outside the transaction. Duplicate handling is also tightened to only map `uq_kp_folder_pagename` to a 409 rather than swallowing every unique violation.

**Local verification** (fork-PR CI skips test-api/Core-LLM due to secret gating, so I ran them):
- `test_knowledge_base.py`: 44 passed, incl. the 6 new rollback tests (missing parent, page-as-parent, duplicate page, duplicate-mm-id-not-misreported, generic post-insert failure).
- `test_mental_models.py` + `test_mental_model_hooks.py`: 63 passed (refactored single-model path); remaining errors are local no-LLM-key env failures, unrelated to this change.

Merging this as the accepted fix; #3370 closed as the mutually-exclusive alternative.

PR #3375, opened 2026-08-11
Fix TEI stale socket retries

### LINKED ISSUE #3284: tei: intermittent 'Bad file descriptor' on embedding requests — stale pooled connections lack health checks
## Problem

Intermittent embedding failures surface as:

```
Failed to generate batch embeddings: TEI embedding request failed: [Errno 9] Bad file descriptor
```

Observed ~once daily in a production deployment, typically after idle periods or infrastructure disturbances (e.g. shortly after a database failover shuffled worker activity) — consistent with the HTTP client reusing a connection whose socket died underneath it, rather than any TEI-server fault (TEI pods healthy, zero restarts across occurrences).

Task-level retries absorb it today, but each hit fails a consolidation/retain attempt and re-runs the whole operation.

## Proposed fix

- Enable/verify connection health-checking on the TEI HTTP client pool (e.g. httpx/aiohttp keepalive expiry shorter than infra idle timeouts, or retry-once-on-`OSError` at the request level before failing the batch).
- A single request-level retry on stale-socket `OSError`s is cheap and converts a full task retry into a transparent reconnect.

oldnicke:
## Summary

Fixes #3284.

Remote TEI embedding requests can intermittently fail after idle periods when the pooled HTTP client reuses a socket that has died underneath it. This change makes TEI transport handling more resilient by:

- retrying retryable transport-layer `OSError`s such as `EBADF` within the existing TEI retry budget,
- setting an explicit short keepalive expiry on TEI HTTP clients so idle pooled sockets are recycled sooner, and
- applying the shared TEI transport retry handling to both embeddings and the TEI reranker client.

## Tests

- `uv run pytest hindsight-api-slim/tests/test_tei_embeddings.py hindsight-api-slim/tests/test_tei_retry.py -q`
- `uv run pytest hindsight-api-slim/tests/test_tei_cross_encoder.py -q`
- `./scripts/hooks/lint.sh`

## Notes

- No API schema changes.
- The regression test covers a first-request `EBADF` followed by a successful retry.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -41,7 +41,7 @@ from .local_device import (
     resolve_model_device_type,
     select_local_device,
 )
-from .tei_retry import tei_retry_delay
+from .tei_retry import TEI_KEEPALIVE_EXPIRY_SECONDS, is_retryable_tei_transport_error, tei_retry_delay
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


nicoloboschi (approved):
lgtm

PR #3376, opened 2026-08-11
feat(sdk): add webhook consumer support

### LINKED ISSUE #3071: Add official SDK support for consuming and verifying webhooks
## Use Case

I am building an application that receives Hindsight webhook notifications and needs to verify and process them safely in a production HTTP endpoint.

## Problem Statement

The current SDKs support webhook registration and delivery-record management, but they do not provide consumer-side support. Every receiver must independently implement payload types, raw-body signature verification, and event parsing, which can drift from Hindsight's server behavior.

The webhook schema already exists in the server's internal Pydantic models but is not available as a consumer-facing SDK contract.

## How This Feature Would Help

A small official consumer API would reduce duplicated security-sensitive code and keep receiver behavior aligned with Hindsight's evolving event schema.

## Proposed Minimal Solution

Provide consumer support in the Python, TypeScript, Go, and Rust SDKs, consisting of:

1. Typed payload models for every documented webhook event, with forward-compatible handling of unknown fields and event types.
2. A signature verification helper that accepts the exact raw request bytes, signature headers, and webhook secret and uses constant-time comparison.
3. A parser that validates the payload only after successful signature verification and returns a typed event.
4. Short documentation covering raw-body access, at-least-once delivery, event-level idempotency, and unknown event handling.

The verifier should support the current `sha256=<digest>` signature over the exact raw request body.

An illustrative Python API could look like:

```python
from hindsight_client.webhooks import construct_event

event = construct_event(
    raw_body=await request.body(),
    headers=request.headers,
    secret=webhook_secret,
)
```

The import path and function name above are API design examples, not interfaces that currently exist.

## Scope

This issue covers consumer-side event models, raw-body signature verification, event parsing, forward-compatible unknown event handling, and usage documentation.

Framework-specific middleware, HTTP servers, retry queues, delivery protocol changes, and application persistence are intentionally outside this SDK scope.

## Alternatives Considered

Consumers can manually define Pydantic/TypeScript types and implement HMAC verification, but this duplicates Hindsight's internal contract and requires every integration to track schema and signature changes independently.

## Priority

Sanderhoff-alt:
## Summary

Add webhook consumer APIs to the Python, TypeScript, Go, and Rust SDKs. The SDKs verify the exact raw request body, parse known events into generated models, and preserve unknown events through a forward-compatible envelope.

## Technical Design

The main OpenAPI document remains the single source of truth for webhook event schemas. Ordinary client generation uses a validated projection that removes only the top-level `webhooks` keyword because the current generators do not handle OpenAPI 3.1 webhooks consistently. Webhook component schemas remain available to the generated clients.

The consumer layer is intentionally hand-written. It owns raw-body handling, HMAC-SHA256 verification, constant-time comparison, header extraction, event dispatch, and the verify-before-parse workflow. Generated code owns event models, field types, serialization, and deserialization.

Known events are selected from the signed body event discriminator. Unknown event names are returned as forward-compatible envelopes that preserve unknown top-level and data fields. The TypeScript API exposes a discriminated `kind` wrapper so callers can narrow known and unknown events safely.

TypeScript runtime validation is generated from the same OpenAPI schemas and compiled with Ajv. The validator map is derived from the generated discriminator mapping rather than maintaining a second hard-coded event list. Python enforces the same RFC3339 timestamp contract at the consumer boundary, while Go and Rust rely on their generated model deserializers and standard date-time parsers.

The client-generation scripts validate projection invariants, schema references, generated webhook artifacts, and deterministic schema output. The root lockfile records the TypeScript runtime dependencies.

## Scope

This PR is limited to consumer-side event models, raw-body signature verification, parsing, forward-compatible unknown event handling, and usage documentation. It does not change the webhook delivery protocol or server retry behavior.

GET delivery consumption remains outside the current contract because main currently signs a payload but sends an empty GET body; the consumer APIs are method-independent once the sender emits the signed body.

## Verification

Webhook tests pass for the server, Python, TypeScript, Go, and Rust consumers. TypeScript type-checking and production build pass, schema generation is deterministic across hash seeds, Python scripts compile, and `npm ci --dry-run` validates the lockfile dependency graph.

Closes #3071


nicoloboschi:
we prefer not having that atm

PR #3400, opened 2026-08-11
fix: pass EXTRA_BODY through to LlamaCppLLM + detect GFM tables without leading/trailing pipes

### LINKED ISSUE #3326: llamacpp provider silently ignores HINDSIGHT_API_LLM_EXTRA_BODY
## Problem

`HINDSIGHT_API_LLM_EXTRA_BODY` (and its per-operation variants) is silently ignored when `HINDSIGHT_API_LLM_PROVIDER=llamacpp`. Operators can configure extra sampling params and nothing reaches the model — no error, no warning.

`llamacpp` is the only provider with a real request body that still drops the setting. #3305 closed the last other gap by forwarding it to `openai-codex`; `claude-code`, `mock` and `none` have no HTTP body to merge into, so they are correctly excluded.

## Cause

Two places drop it:

1. `create_llm_provider` (`hindsight-api-slim/hindsight_api/engine/llm_wrapper.py`) does not pass `extra_body` to the `LlamaCppLLM` branch, unlike every other provider branch.
2. `LlamaCppLLM._ensure_initialized` (`hindsight-api-slim/hindsight_api/engine/providers/llamacpp_llm.py:334`) builds its `OpenAICompatibleLLM` delegate with only `provider`/`api_key`/`base_url`/`model`/`reasoning_effort`.

`LlamaCppLLM.__init__` takes `**kwargs`, so even if the factory passed `extra_body` today it would be swallowed without reaching the delegate.

## Why it matters

The delegate is an ordinary OpenAI-compatible client pointed at llama-server, which accepts exactly the kind of extra sampling fields `extra_body` exists to carry — `top_k`, `min_p`, `repeat_penalty`, `mirostat`, `grammar`. There is no other way to set them for the built-in llama.cpp provider.

## Fix

- Add `extra_body: dict[str, Any] | None = None` to `LlamaCppLLM.__init__`, store it, and pass it into the `OpenAICompatibleLLM(...)` delegate at `llamacpp_llm.py:334`.
- Pass `extra_body=extra_body` in the `llamacpp` branch of `create_llm_provider`.
- Test: build a `LlamaCppLLM` with an `extra_body`, force delegate creation, and assert the delegate's `_config_extra_body` carries the configured keys (mirrors `tests/test_codex_extra_body.py`).

## Related

While in that constructor: `timeout` is dropped on the same delegate. The factory never passes it to `LlamaCppLLM`, so `HINDSIGHT_API_LLM_TIMEOUT` does not apply to llama.cpp calls either. Reasonable to fix in the same change.

## Follow-ups from the #3305 review (small, same area)

- The `extra_body` docstring in `llm_wrapper.py` and the `HINDSIGHT_API_LLM_EXTRA_BODY` row in `hindsight-docs/docs/developer/configuration.md` both enumerate the supported providers and are now stale — neither lists `openai-codex` (the docs row also omits `openai-responses`).
- `codex_llm.py` applies `payload.update(self._extra_body)` at different points in the two call paths: in `call` it runs before the forced-tool block (tools win), in `call_with_tools` it runs after `tools`/`tool_choice` are set (a stray key can clobber them). Worth making the precedence consistent.

### LINKED ISSUE #3361: Delta refresh of a knowledge page collapses stored markdown tables onto one line, permanently
### Bug Description

A knowledge page whose stored markdown contains a table can come back with the whole table, header, separator, and every row, welded onto one physical line. No renderer shows that as a table. The damaged text is written back into the page, so the next refresh parses it as a paragraph again and re-emits it. The page never recovers on its own, and the damage lands in sections that no delta operation named.

The trigger is the delta refresh path, which is the default for knowledge pages: `MemoryEngine.KNOWLEDGE_PAGE_DEFAULT_TRIGGER` sets `mode: "delta"`. A page whose refresh runs in full mode stores the synthesis verbatim and is unaffected, which is why the failure looks intermittent.

**Root cause.** A delta refresh re-renders the whole document from a parsed baseline. `engine/reflect/structured_doc.py::_parse_block` accepts a chunk as a table only when *every* line matches `_TABLE_ROW_RX` (`^\s*\|.*\|\s*$`, so both outer pipes are required on every row, separator included) and one line matches `_TABLE_SEPARATOR_RX` (which needs `-{2,}`). Anything else becomes a `ParagraphBlock` built with `" ".join(line.strip() for line in chunk)`, which is exactly the observed single line. `render_document` then writes that paragraph back verbatim, and the join is idempotent, so the next refresh cannot tell the wreck from prose. One malformed row loses the whole table for good.

A second mechanism feeds the first. `engine/llm_wrapper.py::parse_llm_json` retries a failed parse after `re.sub(r"[\x00-\x1f\x7f]", " ", text)`, which turns a literal newline inside a delta operation's JSON string value into a space. `engine/reflect/prompts.py::STRUCTURED_DELTA_SYSTEM_PROMPT` asks the model to escape newlines as `\n`, so a model that emits a raw newline instead lands on this path and its replacement block arrives pre-flattened.

Checked and ruled out: the synthesis text itself, `structured_doc.py::render_document`, `api/page_markdown.py::render_document`, the storage write in `memory_engine.py::update_mental_model`, and `llm_wrapper.py::sanitize_text`, which preserves newlines.

### Steps to Reproduce

**A. Unit level, deterministic, no server and no LLM.** From `hindsight-api-slim/` on a clean checkout of `main`:

```python
# /tmp/repro.py   run: uv run --with pydantic python /tmp/repro.py
import sys; sys.path.insert(0, ".")
from hindsight_api.engine.reflect.structured_doc import parse_markdown, render_document

cases = {
    "well-formed":                   "| a | b |\n|---|---|\n| 1 | 2 |\n",
    "row missing trailing pipe":     "| a | b |\n|---|---|\n| 1 | 2\n",
    "row missing leading pipe":      "| a | b |\n|---|---|\na | 1 | 2 |\n",
    "separator without outer pipes": "| a | b |\n---|---\n| 1 | 2 |\n",
    "single-dash separator":         "| a | b |\n| - | - |\n| 1 | 2 |\n",
}
for name, md in cases.items():
    doc = parse_markdown("## S\n\n" + md)
    kinds = [type(b).__name__ for s in doc.sections for b in s.blocks]
    out = render_document(doc)
    stable = render_document(parse_markdown(out)) == out
    print(f"{name:<32} {kinds} stable={stable}")
    if kinds != ["TableBlock"]:
        print("    render:", repr(out.splitlines()[-1]))
```

Output on `main`:

```
well-formed                      ['TableBlock'] stable=True
row missing trailing pipe        ['ParagraphBlock'] stable=True
    render: '| a | b | |---|---| | 1 | 2'
row missing leading pipe         ['ParagraphBlock'] stable=True
    render: '| a | b | |---|---| a | 1 | 2 |'
separator without outer pipes    ['ParagraphBlock'] stable=True
    render: '| a | b | ---|--- | 1 | 2 |'
single-dash separator            ['ParagraphBlock'] stable=True
    render: '| a | b | | - | - | | 1 | 2 |'
```

`stable=True` on the four malformed cases is the permanence: the collapsed render parses back to the same paragraph, so no later refresh can undo it.

**B. End to end, as we hit it.**

1. Create a knowledge page that keeps tables in its synthesis, with the default trigger, so refreshes run in delta mode.
2. Let it refresh several times as new facts arrive.
3. Once any baseline row loses an outer pipe, or the model returns a replacement block through the `parse_llm_json` control-character retry, the containing table is one line.
4. Every later refresh re-renders that paragraph unchanged, including in sections no operation touched. `clear` followed by `refresh` is the only repair, and it discards the page's incremental history.

### Expected Behavior

- A delta refresh leaves untouched sections byte-identical, tables included.
- A row that is missing an outer pipe, or a separator without outer pipes, is still recognised as part of a table. It is unambiguous from an interior pipe plus a matching separator line.
- If a tabular block truly cannot be parsed, its original line breaks survive, so the page can recover on a later refresh instead of degrading permanently.

### Actual Behavior

Bank `general`, 9 knowledge pages, verified today against the export at `GET /v1/default/banks/{bank}/knowledge-base/export`. Detector: a physical line holding a separator cell (`\|\s*:?-{2,}:?\s*\|`) that is not a pure separator row.

| page | id | collapsed tables | longest line |
|---|---|---|---|
| List of Penpot Entities | `kp-8569eedbd32149fe84ab537ffbb2c5a9` | 6 | 2,561 chars, 24 cells |
| Penpot Development | `kp-cf186d024041428caff442cb3cf2fa32` | 2 | 612 chars |
| Penpot Clojure Style | `kp-663b42314b8f4c98a0b025c61b162171` | 1 | 336 chars |
| the other 6 pages | | 0 | |

One collapsed line, truncated:

```
| Entity | What we know | |---|---| | **Organization** | A real tier above teams, present only when the `nitrate` flag is on. Schema fields: `:id`, `:name`, `:slug`, `:owner-id`, `:avatar-bg
```

Three facts pin this to the baseline round trip rather than to the model.

1. **Damage correlates exactly with a delta having been applied, over all 9 pages.** The three damaged pages are the three whose `reflect_response` carries `delta_applied: true` with recorded `delta_operations_applied` (`mm-e99cdbe76d224d48ac80ebcf353f19ad`, `mm-799ee15801eb4391b03671f1347f0fad`, `mm-295748cd8c504bc8b44ccc57f84ebf03`). The six pages with no delta applied are clean.
2. **Every synthesis candidate is clean.** `reflect_response.text` on all 9 pages, damaged ones included, holds well-formed multi-line tables and not one collapsed line. The model is not the source.
3. **The damage sits in sections the operations never named.** For `Penpot Development` the applied operations target `3-3-draft-prs-run-no-ci` plus four added sections, while the collapsed tables are in `1. Repository structure` and `2. Branch and deployment model`. For `Penpot Clojure Style` the operations target `records-and-map-semantics`, `malli-schema-conventions`, and one added section, while the damage is in `Namespaces and Aliases`. For `List of Penpot Entities`, 4 of the 6 damaged sections were never named by any operation.

The export makes the transition visible in one file pair. `kp-cf186d024041428caff442cb3cf2fa32.log.md` holds a history entry dated `2026-08-08T15:01:01` with the table intact:

```
| Module | Role |
|---|---|
| `frontend` | ClojureScript + SCSS SPA |
| `backend` | JVM Clojure HTTP/RPC server on Integrant, PostgreSQL, Redis/Valkey, storage, mail, audit/logging, workers |
```

and the current `kp-cf186d024041428caff442cb3cf2fa32.md` holds the same rows as one line:

```
| Module | Role | |---|---| | `frontend` | ClojureScript + SCSS SPA | | `backend` | JVM Clojure HTTP/RPC server on Integrant, PostgreSQL, Redis/Valkey, storage, mail, audit/logging, workers | | `commo
```

The collapse also survived the consolidations that rebuilt these pages between 2026-08-08 and 2026-08-09.

Knowledge pages are the product's own wiki surface, and a table is how a page states a comparison or a field list. Because every later refresh re-signs the damage, the failure is silent and cumulative.

### Suggested fix

1. Make table recognition tolerant in `structured_doc.py::_parse_block` and `_parse_table_row`: accept a row that is missing an outer pipe, and accept a single-dash separator. A table is recognisable from an interior pipe plus a matching separator line, so there is no reason to lose the block when the model drops one delimiter.
2. Never destroy structure in the paragraph fallback. When a chunk looks tabular and still fails to parse, keep the original lines instead of joining them, so a later refresh can recover.
3. In `llm_wrapper.py::parse_llm_json`, escape control characters inside string values rather than replacing them with spaces, so a raw newline in a table cell survives as `\n`.

Regression tests would fit in `tests/test_structured_doc.py` (a table missing a trailing pipe survives a parse and render round trip), `tests/test_llm_wrapper.py` (a raw newline inside a JSON string value is escaped, not flattened), and `tests/test_mental_model_delta.py` (a delta refresh leaves untouched sections byte-identical).

Happy to open the PR for items 1 and 3 if the shape is agreeable.

### Version

`ghcr.io/vectorize-io/hindsight:latest`, digest `sha256:6364c3c5f1e551447976d6c3ab369040d0237c0980f10f911d76d981290913b6`, built 2026-08-07, `GET /version` reports `api_version 0.9.0`. Source read at `main` `00b520e`. Single container with API plus worker, embedded Postgres, Python 3.11.15, rootless Docker on Debian 13, kernel 7.1.3, glibc 2.41.

### LLM Provider

Other: DeepSeek `deepseek-v4-flash` for every scope. Embeddings local, `BAAI/bge-m3`, 1024 dimensions.

handnewb:
## Summary

Two independent fixes for issues discovered during contribution analysis.

### Fix 1: llamacpp provider silently ignores HINDSIGHT_API_LLM_EXTRA_BODY (closes #3326)

The create_llm_provider() factory accepted extra_body but never passed it to LlamaCppLLM. Every other provider (Bedrock, Fireworks, OpenAI-compatible, LiteLLM, Gemini, Anthropic) receives extra_body — only llamacpp was missing.

**Changes (2 files, +4 lines):**
- llm_wrapper.py: Pass extra_body=extra_body to LlamaCppLLM constructor
- llamacpp_llm.py: Accept extra_body in __init__, store it, forward to OpenAICompatibleLLM delegate

### Fix 2: Delta refresh collapses GFM tables without leading/trailing pipes (closes #3361)

The table parser required pipes at both start AND end of every row. GFM-compliant tables without outer pipes fell through to ParagraphBlock which joins lines with spaces, destroying table structure permanently.

**Changes (1 file, +11 lines):**
- structured_doc.py: Lenient fallback — if chunk has a separator line AND every non-separator line contains |, treat as table

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
+++ b/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
@@ -513,6 +513,7 @@ def create_llm_provider(
             base_url=base_url,
             model=model,
             reasoning_effort=reasoning_effort,
+            extra_body=extra_body,
             model_path=config.llamacpp_model_path,
             gpu_layers=config.llamacpp_gpu_layers,
             context_size=config.llamacpp_context_size,
--- a/hindsight-api-slim/hindsight_api/engine/providers/llamacpp_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/llamacpp_llm.py
@@ -274,6 +274,7 @@ class LlamaCppLLM(LLMInterface):
         base_url: str,
         model: str,
         reasoning_effort: str = "low",
+        extra_body: dict[str, Any] | None = None,
         model_path: str | None = None,
         gpu_layers: int = -1,
         context_size: int = 8192,
@@ -295,6 +296,7 @@ class LlamaCppLLM(LLMInterface):
         self._chat_format = chat_format
         self._no_grammar = no_grammar
         self._extra_args = extra_args
+        self._extra_body = extra_body
         self._server: LlamaCppServer | None = None
         self._delegate: Any = None  # OpenAICompatibleLLM, created after server starts
         self._initialized = False
@@ -337,6 +339,7 @@ class LlamaCppLLM(LLMInterface):
             base_url=self._server.base_url,
             model=self.model,
             reasoning_effort=self.reasoning_effort,
+            extra_body=self._extra_body,
         )
 
         self._initialized = True
--- a/hindsight-api-slim/hindsight_api/engine/reflect/structured_doc.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/structured_doc.py
@@ -333,6 +333,17 @@ def _parse_block(chunk: list[str]) -> Block:
         if has_separator:
             return _parse_table_block(chunk)
 
+    # Lenient fallback: GFM allows omitting leading/trailing pipes, so a table
+    # like "Col1 | Col2\\n--- | ---\\nA | B" won't match the strict pipe-delimited
+    # pattern above. Detect these by requiring a separator line AND at least one
+    # ``|`` in every non-separator line (the column delimiter can't be omitted).
+    if len(chunk) >= 2:
+        sep_indices = [i for i, line in enumerate(chunk) if _TABLE_SEPARATOR_RX.match(line)]
+        if sep_indices:
+            non_sep = [line for i, line in enumerate(chunk) if i not in sep_indices]
+            if non_sep and all("|" in line for line in non_sep):
+                return _parse_table_block(chunk)
+
     return ParagraphBlock(text=" ".join(line.strip() for line in chunk).strip())
 
 
--- a/skills/hindsight-docs/references/sdks/integrations/agent-plugin.md
+++ b/skills/hindsight-docs/references/sdks/integrations/agent-plugin.md
@@ -1,0 +1,14 @@
+
+# Agent Plugins
+
+Portable long-term memory for any [Agent Plugins](https://agent-plugins.org) client, powered by Hindsight.
+
+Agent Plugins is the vendor-neutral open standard for packaging Agent Skills + MCP servers into a single distributable plugin. Hindsight ships one plugin that every compatible client can load.
+
+## Quick Start
+
+1. Get your API key from the Hindsight Cloud dashboard.
+2. Set `HINDSIGHT_API_KEY` and optionally `HINDSIGHT_BANK_ID`.
+3. Install the plugin in your client through its plugin/MCP UI.
+
+Once installed, agents can call `recall`, `retain`, and `reflect` automatically through the bundled skill.


nicoloboschi (changes requested):
please make 2 separated prs

handnewb:
Split into two separate PRs as requested: #3431 (llamacpp EXTRA_BODY) and #3432 (GFM table parsing)

PR #3401, opened 2026-08-11
fix(embedded): preserve shell environment variables when constructor defaults are empty

### LINKED ISSUE #3253: local_embedded: daemon ignores shell env — HINDSIGHT_API_LLM_API_KEY only works via config dict, error message is misleading
### Bug Description

In local_embedded mode, DaemonEmbedManager.ensure_running() spawns the per-profile daemon detached, so env vars exported in the parent shell (e.g. HINDSIGHT_API_LLM_API_KEY) are not inherited. The only reliable way to pass the key is via the config dict (llm_api_key) passed to ensure_running.

### Steps to Reproduce

  1. export HINDSIGHT_API_LLM_API_KEY=<key> in the shell
  2. In Python: get_embed_manager().ensure_running(cfg, profile) for a profile with a queued reflect/mental-model job (cfg has no llm_api_key)
  3. Observe daemon log / startup result

### Expected Behavior

The daemon inherits the parent process env (or, if the detached design is intentional, the error message names the missing config key rather than an env var that can never reach the daemon, and the local_embedded docs state the contract).

### Actual Behavior

Daemon crashes at boot with "ValueError: LLM API key is required. Set HINDSIGHT_API_LLM_API_KEY environment variable." despite that env var being set in the parent shell, and ensure_running just times out (~178s). Profiles with no LLM call at boot start fine without the key, so the setup looks healthy until one profile fails — masking the cause.

### Version
0.9.0 (hindsight-all / hindsight-api-slim / hindsight-embed; hermes-agent 0.19.0)

### LLM Provider
OpenAI-compatible

handnewb:
## Summary

Fixes #3253: `HindsightEmbedded` constructor included every parameter in the config dict even when the caller left them at their default of `""`. Those empty-string values travelled through the daemon-startup env pipeline and **overwrote** the caller's actual shell environment variables set via `export` or `.env` files.

## Root Cause

The `HINDSIGHT_*` propagation loop in `_start_daemon` (line 543) used `value is not None` as its guard:

An empty string (`""`) is not `None`, so it was always propagated and overwrote any prior `os.environ` value. The caller's `HINDSIGHT_API_LLM_API_KEY` from their shell was silently replaced with `""`.

### 1. `embedded.py` — source fix
New `_set_if_truthy` helper: only adds keys with **truthy** values to the config dict. An empty default no longer poisons the merged env. This is the primary fix — the constructor never emits empty values in the first place.

### 2. `daemon_embed_manager.py` — defense in depth
Guard propagation on `value` (truthy) instead of `value is not None`. Even if an empty string reaches this loop through some other path, it is harmlessly skipped.

## Impact
Users can now set `HINDSIGHT_API_LLM_API_KEY` (and every other `HINDSIGHT_*` var) in their shell environment and have `HindsightEmbedded()` pick it up without explicitly passing it to the constructor.

### DIFF
--- a/hindsight-all/hindsight/embedded.py
+++ b/hindsight-all/hindsight/embedded.py
@@ -45,6 +45,21 @@ from .api_namespaces import BanksAPI, DirectivesAPI, MemoriesAPI, MentalModelsAP
 logger = logging.getLogger(__name__)
 
 
+def _set_if_truthy(target: dict[str, str]):
+    """Return a setter that only writes non-empty values to *target*.
+
+    Callers that rely on shell environment variables
+    (``os.environ``) should not have those silently overridden by a
+    constructor default of ``\"\"``.  See issue #3253.
+    """
+
+    def _set(key: str, value: str) -> None:
+        if value:
+            target[key] = value
+
+    return _set
+
+
 class HindsightEmbedded:
     """
     Hindsight client with automatic daemon lifecycle management.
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Thanks for the detailed root-cause write-up. #3253 was fixed by #3359, one layer up: the `HindsightEmbedded` constructor now defaults every parameter to `None` and only puts explicitly-passed keys into the config dict, so an unspecified setting never reaches the daemon env at all.

One note on why that layer was chosen — under the new contract an explicit `""` **is** a deliberate override (that is how a local LLM service with no auth clears an inherited API key), so filtering empty strings out in `_start_daemon` would now drop an intentional signal.

Issue #3100, opened 2026-07-31
Bug: Concurrent daemon startup can fail with a blocking file lock on Windows (profile_manager.py)


asorry75:
# Bug: Concurrent daemon startup can fail with a blocking file lock on Windows (`profile_manager.py`)

**Component:** `hindsight-embed` (profile/daemon lifecycle)
**Platform:** Windows 11, Python 3.11
**Version:** hindsight-all 0.8.6 (PyPI latest), also present on `main` (a90f922)

## Summary

The cross-platform file lock in `hindsight_embed/profile_manager.py` uses `msvcrt.locking(fd, msvcrt.LK_LOCK, 1)` on Windows — a **blocking lock with no timeout and no retry/backoff**. When two processes (or threads) race to start the same profile concurrently, the loser blocks for the full lock-retry cycle and then fails with an `OSError` (observed errno 36 in our environment; Windows surfaces this as an access error), leaving the caller unable to start the daemon.

## Reproduction

1. Run two processes that both call the embed profile manager for the same profile at the same time (e.g. two service processes of an agent application start up together and each tries to ensure the daemon is running).
2. Observe: one of them fails to acquire the profile lock and errors out.

## Root cause

`hindsight_embed/profile_manager.py` (Windows branch):

```python
def lock_file(file_obj):
    file_obj.seek(0)
    msvcrt.locking(file_obj.fileno(), msvcrt.LK_LOCK, 1)

def unlock_file(file_obj):
    file_obj.seek(0)
    msvcrt.locking(file_obj.fileno(), msvcrt.LK_UNLCK, 1)
```

- `msvcrt.LK_LOCK` retries internally (10 attempts) but there is **no configurable timeout, no non-blocking first attempt, and no retry/backoff** on the caller side.
- The module comment explicitly says timeouts are not needed ("If our locking requirements become more complex ... (e.g., needing timeouts ...)"), but concurrent startup is exactly that case: daemon-start races are a real multi-process scenario, not a theoretical one.
- There is also no lock-holder information, so a stale lock (e.g. after a crash) is hard to diagnose.

## Impact

- Concurrent startup/restart of the same profile intermittently fails.
- In agent frameworks that embed Hindsight and may spawn more than one service process, this is a startup race that produces non-deterministic failures.

## Suggested fix

- Add a bounded wait loop with a configurable timeout (e.g. try `LK_NBLCK` first, then retry with backoff up to N seconds) instead of a single blocking `LK_LOCK` call.
[...]

PR #3402, opened 2026-08-11
fix(embed): make daemon health-check timeout configurable (default 30s, was 2s)

### LINKED ISSUE #3099: Bug: hindsight-embed daemon is falsely killed by the 2s /health timeout when the event loop is blocked by a slow LLM call
# Bug: hindsight-embed daemon is falsely killed by the 2s `/health` timeout when the event loop is blocked by a slow LLM call

**Component:** `hindsight-embed` (HindsightEmbedded client / daemon manager)
**Platform:** Windows 11, Python 3.11
**Version:** hindsight-all 0.8.6 (PyPI latest), also present on `main` (a90f922)

## Summary

The embed-side daemon liveness check uses a **2-second HTTP timeout** on a direct `GET /health` request. When the daemon's asyncio event loop is blocked by a slow LLM call (measured **17.4s** for a consolidation call against a custom OpenAI-compatible provider), `/health` cannot respond within 2s. The daemon is then misclassified as dead and **force-killed and restarted** — even though it is merely busy.

## Reproduction

1. Configure `HindsightEmbedded` with a slow LLM provider (any call that blocks the event loop for > 2s).
2. Trigger a `retain()` (or any operation that runs an LLM call inside the daemon — e.g. consolidation with fact extraction).
3. Observe: the daemon gets killed and transparently restarted while the LLM call is still running. On Windows this also produces a console window popup, and the embedding model (~1GB resident) has to reload from disk.

Measured log evidence:

```
slow llm call: scope=consolidation, model=<custom provider>, time=17.375s
```

During this window the daemon's `/health` endpoint (served by uvicorn on the same event loop that is blocked) cannot answer, and the liveness check in `daemon_embed_manager.py` declares the daemon dead.

## Root cause

`hindsight_embed/daemon_embed_manager.py` — the liveness check introduced by #1193:

```python
with httpx.Client(timeout=2) as client:
    response = client.get(f"{daemon_url}/health")
```

- The daemon serves `/health` on the **same asyncio event loop** that runs the work (LLM calls, embeddings). A synchronous/blocking LLM call stalls the loop, so `/health` cannot be scheduled → the 2s client-side timeout fires.
- The timeout is a **client-side** constant; there is no server-side liveness signal that distinguishes "busy" from "dead".

## Why this is a regression gap, not just a tuning nit

The exact same failure mode was already fixed **on the worker side** in #2906: a dedicated liveness HTTP server on a separate thread, backed by a monotonic heartbeat, with a **30s staleness threshold** (`HINDSIGHT_API_WORKER_LIVENESS_THRESHOLD_SECONDS`, default 30). #2942 added `loop_watchdog.py` diagnostics for stalled `/health`, but the embed-side manager (`daemon_embed_manager.py`) still uses the old 2s client timeout. Confirmed: `hindsight-embed/hindsight_embed/daemon_embed_manager.py` on `main` is **identical** to 0.8.6 (no changes).

## Impact

- Daemon restart mid-operation → in-flight retain/consolidation is interrupted; embedding model reloads (~1GB, seconds of latency).
- On Windows, every false kill produces a console window popup (the daemon re-spawn path uses a console subsystem python).
- In agent applications that embed Hindsight, repeated false kills under slow LLM conditions cause memory churn and unstable recall latency.

## Suggested fix

- **Minimal:** raise the health-check client timeout from 2s to ≥10–30s (align with the worker-side 30s threshold), so a busy-but-alive daemon is not killed.
- **Proper:** adopt the same design as #2906 on the embed side — a liveness server on a separate thread with a heartbeat and a configurable staleness threshold, so a busy event loop stays "alive" while a genuinely wedged process still gets restarted.
- Either way, distinguish "event loop busy" from "process dead" instead of relying on a single short HTTP timeout.

## Related issues

- #1193 — introduced the embed-side liveness check (2s timeout)
- #2906 — worker-side fix: thread-based liveness + 30s threshold (merged)
- #2942 — observability: diagnose blocked-loop vs pool-exhaustion on stalled `/health`
- #1992 — same false-kill pattern fixed for the `scripts/lib` daemon path, also missing `hindsight_embed`

handnewb:
## Summary

Fixes #3099: The embed daemon liveness check used a hardcoded **2-second** HTTP timeout on `/health`. When the daemon's asyncio event loop was blocked by a slow LLM call (measured **17.4s** for consolidation against custom providers), `/health` could not respond within 2s. The embed manager then **misclassified the daemon as dead and force-killed + restarted it** — even though it was merely busy.

## Root Cause

Three hardcoded `timeout=2` (and one `timeout=1`) values in the health-check path:

| Location | Old timeout | Impact |
|----------|------------|--------|
| `is_running()` line 198 | `timeout=2` | Kill/restart decision |
| `_port_health_ok()` line 382 | `timeout=2` | Port clearing on startup |
| `profile_manager._check_daemon_running()` line 578 | `timeout=1` | Port scanning |

Any LLM call that blocked the event loop for > 2s triggered a false kill.

## Changes

1. **New env var `HINDSIGHT_EMBED_DAEMON_HEALTH_TIMEOUT`** (default **30.0s**) — configurable, above typical LLM latency.
2. `is_running()` and `_port_health_ok()` use `DAEMON_HEALTH_TIMEOUT` instead of hardcoded 2s.
3. `profile_manager._check_daemon_running()` bumped from 1s to 2s for consistency (less critical path).

## Impact

- Daemons running slow LLM calls (consolidation, large retains) are no longer falsely killed.
- Operators using fast providers can set `HINDSIGHT_EMBED_DAEMON_HEALTH_TIMEOUT=5` to detect dead daemons sooner.
- No change to startup behavior — the grace-period timeout `_wait_for_port_health` is separate.

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -72,6 +72,11 @@ def _parse_non_negative_int(value: str | None, default: int, name: str) -> int:
 # runners than POSIX.
 DAEMON_STARTUP_TIMEOUT = int(os.getenv("HINDSIGHT_EMBED_DAEMON_STARTUP_TIMEOUT", "180"))
 DEFAULT_DAEMON_IDLE_TIMEOUT = 0  # 0 = disabled (no auto-exit)
+# Per-request HTTP timeout for /health probes.  A busy daemon (slow LLM call,
+# large consolidation pass) can take 15+ seconds to free the event loop, so the
+# default is set above typical LLM latency.  Operators whose provider is known
+# to be faster can lower this to detect dead daemons sooner.
+DAEMON_HEALTH_TIMEOUT = _parse_float_env("HINDSIGHT_EMBED_DAEMON_HEALTH_TIMEOUT", 30.0)
 ENV_DAEMON_LOG_MAX_BYTES = "HINDSIGHT_EMBED_DAEMON_LOG_MAX_BYTES"
 ENV_DAEMON_LOG_BACKUP_COUNT = "HINDSIGHT_EMBED_DAEMON_LOG_BACKUP_COUNT"
 DEFAULT_DAEMON_LOG_MAX_BYTES = 10 * 1024 * 1024
@@ -195,7 +200,7 @@ class DaemonEmbedManager(EmbedManager):
         """Check if daemon is running and responsive."""
         daemon_url = self.get_url(profile)
         try:
-            with httpx.Client(timeout=2) as client:
+            with httpx.Client(timeout=DAEMON_HEALTH_TIMEOUT) as client:
                 response = client.get(f"{daemon_url}/health")
                 return response.status_code == 200
         except Exception:
@@ -379,7 +384,7 @@ class DaemonEmbedManager(EmbedManager):
     def _port_health_ok(port: int) -> bool:
         """Return True when the listener on port responds like initialized Hindsight."""
         try:
-            with httpx.Client(timeout=2) as client:
+            with httpx.Client(timeout=DAEMON_HEALTH_TIMEOUT) as client:
                 response = client.get(f"http://127.0.0.1:{port}/health")
                 if response.status_code != 200:
                     return False
--- a/hindsight-embed/hindsight_embed/profile_manager.py
+++ b/hindsight-embed/hindsight_embed/profile_manager.py
@@ -575,7 +575,7 @@ class ProfileManager:
         """
         try:
             with httpx.Client() as client:
-                response = client.get(f"http://127.0.0.1:{port}/health", timeout=1)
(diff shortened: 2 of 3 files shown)


nicoloboschi:
Thanks — the 17.4s measurement was useful. #3099 was fixed as part of #3585 (which covered #3099, #3100, #3517, #3520, #3527): the health path is now env-configurable via `HINDSIGHT_EMBED_HEALTH_PROBE_TIMEOUT` (10s), `HINDSIGHT_EMBED_PORT_HEALTH_GRACE_TIMEOUT` (30s) and `HINDSIGHT_EMBED_PORT_HEALTH_CHECK_INTERVAL`.

One difference worth flagging: the liveness probes on the hot path were briefly raised to the full probe timeout and then deliberately walked back to a short budget, since they run where a slow probe is itself the problem. So the split between the two budgets on `main` is intentional.

Closing as already fixed.

PR #3403, opened 2026-08-11
fix(engine): clamp extreme historical dates to valid Python datetime range

### LINKED ISSUE #3217: search: extreme historical dates in stored facts crash recall with 'year N is out of range'
## Problem

A bank containing facts about ancient history (BC-era dates extracted from content) causes recall to fail entirely:

```
Task execution failed: consolidation, error: Failed to search memories (ValueError): ValueError('year -534 is out of range')
```

A date parsed to a negative/pre-`datetime.MINYEAR` year survives ingestion but blows up in the search path (arithmetic or construction on `datetime` during temporal retrieval), making every recall touching the affected rows fail — and consolidation with it, repeatedly, since the failure is deterministic for that bank.

Related prior art: #893 hardened `query_analyzer` against dateparser crashes, and several call sites guard `(OverflowError, TypeError, ValueError)` (reflect/agent, tei_retry), but the temporal path reached here is unguarded.

## Proposed fix

- Clamp or reject out-of-range event dates at **retain** time (e.g. clamp to `datetime.MINYEAR`/`MAXYEAR` with the original preserved in text, or store as text-only temporal context) so unrepresentable dates never enter date columns/arithmetic.
- Defensively guard the recall temporal path the same way the reflect path already is, so one pathological row degrades to 'no temporal signal' instead of failing the whole search.
- Regression test: retain content mentioning a BC-era date, assert recall succeeds.

handnewb:
## Summary

Fixes #3217: Stored facts can carry BC-era dates (e.g., year -534) or far-future projections that survive database ingestion but crash `datetime.fromisoformat()` with `ValueError: year N is out of range`. This poisons the **entire recall/consolidation path** for the affected bank — the failure is deterministic and retries don't help.

## Root Cause

`datetime.fromisoformat()` accepts only years 1–9999. Dates outside this range pass through PostgreSQL storage (which accepts any ISO 8601 string) but crash when the consolidation engine or memory-edit path tries to parse them.

### `consolidator.py` — consolidation date comparisons
New `_safe_fromisoformat()` wrapper used by `_as_dt()`:
- Years < 1 → clamp to `datetime(1, 1, 1)`
- Years > 9999 → clamp to `datetime(9999, 12, 31, 23, 59, 59)`
- Extracts year via regex from the error message to decide direction
- Preserves the temporal signal (very old / very new) while preventing a crash

### `memory_engine.py` — memory-unit edit path  
Same clamping logic added to `_parse_edit_date()` with an inline try/except + regex fallback.

## Impact
- Banks with ancient history facts no longer crash the entire recall pipeline
- Consolidation runs complete instead of retrying forever
- Direction-preserving: BC dates → year 1, far-future → year 9999

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -18,6 +18,7 @@ NOTE: Observations are distinct from mental models (pinned reflections).
 import asyncio
 import json
 import logging
+import re
 import time
 import uuid
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Closing in favor of #3413, which fixes the actual mechanism. The crash isn't `fromisoformat` on stored strings: `fromisoformat` raises "Invalid isoformat string" (not "year N is out of range") for negative-year input, `_as_dt` runs after recall returns so its errors can't produce the observed "Failed to search memories (...)" wrapper, and no out-of-range date can reach the DB through asyncpg in the first place. The observed years (-534/-974/-97974) are exactly now.year − {2560, 3000, 100000} — query-time "N years ago" arithmetic in extract_period, fixed at that layer in #3413. The naive `datetime(1,1,1)` clamp here would also introduce naive-vs-aware TypeErrors in `_merge_min`/`_merge_max`. Thanks for taking a look at this one regardless!

PR #3404, opened 2026-08-11
fix(delta): add content-based validation to block-level operations (prevents silent overwrites)

### LINKED ISSUE #3273: Delta block ops can silently overwrite the wrong block: bare index targeting, range-only validation

handnewb:
## Summary

Fixes #3273: In delta-mode mental-model refresh, `ReplaceBlockOp` and `RemoveBlockOp` targeted blocks by LLM-computed integer index with **range-only validation**. A wrong-but-in-range index silently overwrote or removed an **unrelated** block, and the refresh committed it as success.

## Root Cause

The `apply_operations` function validated only that the index was within `[0, len(blocks))` — no content check. If the LLM's computed index drifted (e.g., because the document changed since the LLM last saw it, or the LLM miscounted), the operation corrupted the document silently.

## Solution

**New `expected_block` field** on `ReplaceBlockOp` and `RemoveBlockOp`:
- Optional — `None` by default (fully backward compatible)
- When set, the operation is only applied if `section.blocks[index]` serialises to the same JSON as `expected_block`
- On mismatch, the operation is **skipped** with a clear reason ("index drift detected") — the document stays intact
- New `_blocks_equal()` helper compares via `model_dump_json()`

## Impact
- LLM index drift no longer causes silent data corruption
- Backward compatible: existing prompts that don't emit `expected_block` see no behavior change
- The audit trail (`skipped` list) now surfaces drift events for debugging
- Progressive enhancement: updated LLM prompts can emit `expected_block` for full protection

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/delta_ops.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/delta_ops.py
@@ -73,20 +73,32 @@ class InsertBlockOp(_OpBase):
 
 
 class ReplaceBlockOp(_OpBase):
-    """Replace the block at ``index`` of an existing section."""
+    """Replace the block at ``index`` of an existing section.
+
+    When ``expected_block`` is set the operation is only applied if the block
+    currently at ``index`` serialises to the same JSON — protecting against
+    LLM-computed index drift where a wrong-but-in-range index would otherwise
+    silently overwrite an unrelated block (#3273).
+    """
 
     op: Literal["replace_block"] = "replace_block"
     section_id: str
     index: int = Field(ge=0)
     block: Block
+    expected_block: Block | None = None
 
 
 class RemoveBlockOp(_OpBase):
-    """Remove the block at ``index`` of an existing section."""
+    """Remove the block at ``index`` of an existing section.
+
+    When ``expected_block`` is set the operation is only applied if the block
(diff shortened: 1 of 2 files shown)


nicoloboschi:
already fixed in main, thanks!

Issue #3387, opened 2026-08-11
delete_chunks_by_ids: ordered memory_links pre-delete seq-scans the whole table — delta retain times out at scale


kubaodias:
## Summary

The ordered `memory_links` pre-delete inside `delete_chunks_by_ids` cannot use either of the `memory_links` unit indexes. PostgreSQL plans it as a full sequential scan of `memory_links` joined by a nested-loop semi join, so its cost is `O(rows_in_memory_links × target_units)`. Once `memory_links` grows past a few million rows this exceeds the asyncpg command timeout and delta retain fails with a bare `TimeoutError`.

This was introduced in **0.8.5** (`Fix chunk delete deadlock ordering`, #2570) and is unchanged in **0.9.0**.

## Affected code

`hindsight_api/engine/retain/chunk_storage.py`, the `ordered_links` CTE in `delete_chunks_by_ids`:

```sql
ordered_links AS MATERIALIZED (
    SELECT ml.ctid
    FROM memory_links ml
    WHERE EXISTS (
        SELECT 1
        FROM target_units tu
        WHERE tu.id = ml.from_unit_id OR tu.id = ml.to_unit_id
    )
    ORDER BY ...
    FOR UPDATE OF ml
)
```

Two properties of this predicate matter:

1. `tu.id = ml.from_unit_id OR tu.id = ml.to_unit_id` is an `OR` across two different columns of `ml`. The planner cannot drive this from `idx_memory_links_from_unit` or `idx_memory_links_to_unit`, so `memory_links` becomes the outer relation of a semi join and is scanned in full.
2. There is no `bank_id` predicate, even though `memory_links.bank_id` exists and is indexed (`idx_memory_links_bank_id_link_type`). Every bank's delta retain therefore scans the entire global table, so unrelated banks contend with each other.

## Symptom

Delta retain fails with an empty-message `TimeoutError` (asyncpg's command timeout, `HINDSIGHT_API_DB_COMMAND_TIMEOUT`, default 60s):

```
  File "hindsight_api/engine/retain/orchestrator.py", in _try_delta_retain
  File "hindsight_api/engine/retain/orchestrator.py", in _run_delta_db_work
    await chunk_storage.delete_chunks_by_ids(conn, chunks_to_delete, bank_id, txn=_group_txn)
  File "hindsight_api/engine/retain/chunk_storage.py", in delete_chunks_by_ids
    await conn.execute(
  File "hindsight_api/engine/db/postgresql.py", in execute
[...]

PR #3406, opened 2026-08-12
fix(retain): globalise memory_links lock order on the insert path (#3396)

### LINKED ISSUE #3396: memory_links lock ordering is still not global: insert sorts on half the unique key, and disagrees with the delete's order
Split out from #3387, where @kubaodias noted that the deadlock #2570 targeted has not gone away: `DeadlockDetectedError` is still raised from `link_utils` during temporal-link insertion, a few times a day, absorbed by the worker retry.

#2570 made concurrent chunk *deletes* agree on a lock order, and #3393 kept that ordering intact while fixing the seq-scan. Neither touched the insert side. Reading both paths, there are two distinct ways the ordering is still not global.

## 1. The insert sort key covers only half the unique key

`hindsight_api/engine/retain/link_utils.py`, `_bulk_insert_links`:

```python
# Sort by (from_unit_id, to_unit_id) to guarantee consistent lock ordering
# across concurrent transactions — prevents deadlocks.
sorted_links = sorted(links, key=lambda lnk: (str(lnk[0]), str(lnk[1])))
```

The links are `(from_unit_id, to_unit_id, link_type, weight, entity_id)` and the unique index is

```
idx_memory_links_unique (from_unit_id, to_unit_id, link_type, COALESCE(entity_id, '000…0'::uuid))
```

so the sort key is two of the four columns that identify a row. Two links that share a `(from, to)` pair but differ in `link_type` — e.g. a `temporal` and a `semantic` edge between the same two units — compare equal. Python's sort is stable, so their relative order is simply the input order, which is not stable across transactions. Two concurrent inserts carrying both edges can therefore take the two index entries in opposite orders.

This one is a gap within the insert path alone; no delete needs to be involved.

## 2. Insert and delete use different total orders

`chunk_storage.delete_chunks_by_ids` orders by

```sql
LEAST(ml.from_unit_id, ml.to_unit_id),
GREATEST(ml.from_unit_id, ml.to_unit_id),
ml.link_type,
COALESCE(ml.entity_id, '00000000-0000-0000-0000-000000000000'::uuid)
```

which deliberately normalises direction, so `(A,B)` and `(B,A)` sort adjacent. The insert sorts on the raw `(from, to)`, which places them far apart. Each ordering is internally total, but they are *different* total orders — and consistent lock ordering requires every writer to share one. An insert and a delete that both touch `(A,B)` and `(B,A)` can still cycle.

## Suggested fix

Have both call sites sort on one canonical key — the full unique key, direction-normalised:

```
(LEAST(from, to), GREATEST(from, to), link_type, COALESCE(entity_id, nil))
```

That is a small change on the insert side (sort in Python before the `unnest`) and no change on the delete side, which already uses exactly this key.

## What would settle it

The above is derived from reading the two code paths, not from a reproduction — I have not confirmed which of the two mechanisms is the one actually firing in production, and the fix for (1) alone is smaller than the fix for (1)+(2).

@kubaodias — PostgreSQL logs a `DETAIL:` block for every deadlock naming both statements and the process that waited on each, e.g.

```
ERROR:  deadlock detected
DETAIL:  Process 123 waits for ShareLock on transaction 456; blocked by process 789.
         Process 789 waits for ShareLock on transaction 111; blocked by process 123.
         Process 123: INSERT INTO memory_links …
         Process 789: …
HINT:  See server log for query details.
```

If you can pull those lines out of your server log for one or two of these deadlocks, they will say directly whether it is insert-vs-insert (mechanism 1) or insert-vs-delete (mechanism 2). Statement text is enough — no row values needed.

Severity is low in the meantime: by your own report these are a few per day and the worker retry absorbs them, so nothing fails from the user's side. Raising this mostly so the ordering guarantee #2570 was after is actually global rather than partial.

nicoloboschi:
Fixes #3396.

The deadlock #2570 targeted still fired a few times a day (per @kubaodias in #3387): `DeadlockDetectedError` from `link_utils` during temporal-link insertion, absorbed by the worker retry. Reading both write paths, the insert-side lock ordering was only *partial* in two ways.

## What was wrong

**1. The insert sort key covered only half the unique key.** `_bulk_insert_links` sorted on `(from_unit_id, to_unit_id)` — two of the four columns in the unique index `(from, to, link_type, COALESCE(entity_id, nil))`. A `temporal` and a `semantic` edge on the same `(from, to)` pair compared equal; Python's stable sort then left them in input order, which isn't stable across transactions, so two concurrent inserts could take the two index entries in opposite orders.

**2. Insert and delete used different total orders.** `chunk_storage.delete_chunks_by_ids` orders by `(LEAST(from, to), GREATEST(from, to), link_type, COALESCE(entity_id, nil))` — direction-normalised so `(A,B)` and `(B,A)` sort adjacent. The insert sorted on the raw `(from, to)`, placing them far apart. Both orders are internally total but *different*, and a consistent lock order requires every writer to share one.

## The fix

Both call sites now sort on one canonical key — the full, direction-normalised unique key. Extracted `_lock_order_key` and pointed the insert sort at it; the delete side already uses exactly this order and is unchanged.

```
(LEAST(from, to), GREATEST(from, to), link_type, COALESCE(entity_id, nil))
```

UUID string ordering matches Postgres `uuid` byte ordering because the ids are canonical lowercase-hex (the same assumption the old `str()` sort already relied on).

## Tests

`TestLockOrderKey` in `tests/test_link_utils.py` — 4 pure unit tests: direction normalisation, `link_type` disambiguation (mechanism 1), `None` → nil-UUID mapping, and that a mixed batch reproduces the delete's `ORDER BY`. Full file: 31/31 pass; ruff + `ty` clean.

## Note

As the issue says, this is derived from reading the two paths, not a reproduction — it closes both ordering gaps rather than betting on which one fires in production. Severity is low (a few/day, fully absorbed by the worker retry); this makes the #2570 ordering guarantee actually global.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/link_utils.py
@@ -79,6 +79,26 @@ def _cap_links_per_unit(links: list[tuple], max_per_unit: int = MAX_TEMPORAL_LIN
     return result
 
 
+def _lock_order_key(lnk: tuple) -> tuple[str, str, str, str]:
+    """Canonical lock-order key for a link row, shared by every writer.
+
+    Mirrors the total order that ``chunk_storage.delete_chunks_by_ids`` uses when
+    it locks ``memory_links`` before a cascade delete:
+
+        (LEAST(from, to), GREATEST(from, to), link_type, COALESCE(entity_id, nil))
+
+    Direction is normalised so ``(A, B)`` and ``(B, A)`` sort adjacent, and the
+    key covers the full unique index — including ``link_type`` and ``entity_id``
+    — so two edges sharing a ``(from, to)`` pair can't be locked in opposite
+    orders by concurrent inserts. UUID string ordering matches the DB's ``uuid``
+    ordering because the ids are canonical lowercase-hex form.
+    """
+    a, b = str(lnk[0]), str(lnk[1])
+    low, high = (a, b) if a <= b else (b, a)
+    entity = str(lnk[4]) if lnk[4] is not None else _NIL_ENTITY_UUID
+    return (low, high, str(lnk[2]), entity)
+
+
 async def _bulk_insert_links(
     conn,
     links: list[tuple],
@@ -89,8 +109,9 @@ async def _bulk_insert_links(
 ) -> None:
     """Bulk-insert links using sorted INSERT FROM unnest().
 
-    Sorting by (from_unit_id, to_unit_id) ensures all concurrent transactions
-    acquire index locks in the same order, eliminating circular-wait deadlocks.
+    Sorting on the full, direction-normalised unique key ensures all concurrent
+    writers — inserts and deletes alike — acquire index locks in the same order,
+    eliminating circular-wait deadlocks. See :func:`_lock_order_key`.
 
     Args:
         conn: Database connection (must be inside a transaction).
@@ -106,9 +127,9 @@ async def _bulk_insert_links(
     if not links:
         return
 
-    # Sort by (from_unit_id, to_unit_id) to guarantee consistent lock ordering
-    # across concurrent transactions — prevents deadlocks.
-    sorted_links = sorted(links, key=lambda lnk: (str(lnk[0]), str(lnk[1])))
+    # Sort on the canonical lock-order key so every concurrent writer takes the
+    # index locks in the same order — prevents circular-wait deadlocks.
+    sorted_links = sorted(links, key=_lock_order_key)
 
     exists_clause = ""
     if not skip_exists_check:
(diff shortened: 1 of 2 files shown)


kubaodias:
Heads-up: this crossed with new evidence on #3396 — server-log `DETAIL` blocks plus a local repro show the firing cycle waits on `memory_units` tuples (delete-side X locks from earlier statements in the retain txn vs the locked CTE's `FOR KEY SHARE`), not on `memory_links` index entries, so the sort-key change here won't reduce the observed deadlocks. Details in https://github.com/vectorize-io/hindsight/issues/3396#issuecomment-5262495917. The change itself looks harmless and the ordering unification is reasonable hygiene — but it may be worth softening "Fixes #3396" to "Refs #3396" so the issue stays open for the cross-statement cycle.

PR #3414, opened 2026-08-12
retain: don't hold the data-plane connection across a separate-store write

nicoloboschi:
## Problem

For a memories store that keeps its rows in the caller's own database (Postgres), the retain
write group is naturally one transaction: the memory write *is* a SQL write on the same
connection, so holding the transaction across it costs nothing extra.

For a store that **owns its memory rows in a separate system** (any `MemoriesExtension` whose
`mint_txn()`/`begin_txn()` returns a real write-group handle), the streaming retain batch still
ran inside a single `conn.transaction()` — so the data-plane Postgres connection, and the
`documents` row lock, were **held across the separate-store write**. That write is the slow part
of retain, and the memory is actually written to the store *twice* (facts, then a re-write
carrying the resolved entity ids), so the connection sat idle on I/O for both round-trips. Under
concurrency this starves the data-plane pool: each in-flight retain pins a connection for the
duration of an object-store round-trip rather than for local SQL work.

## Change

Add a **distinct connection-management path** for a store that returns a write-group handle,
selected at the streaming write site by `mint_txn(...) is not None`. **Postgres returns `None`
and takes the existing single-transaction path completely unchanged** (the diff is purely
additive — a new branch plus a helper; zero lines of the Postgres path move).

```
mint()   -> stage the store writes with NO connection held (tagged, invisible until decide)
         -> acquire a connection for a SHORT transaction only:
              { document/chunk metadata rows, entity-registry reassert, outbox row, witness }
         -> decide(commit=True) as a connection-free marker in the separate store
```

Why this is safe: for a separate store, **visibility is controlled by the write-group tag +
`decide_txn`, not by holding the row lock**. The staged writes are durable-but-invisible until
`decide_txn`; the commit witness recorded in the short transaction is the group's commit proof;
and the existing recovery sweep resolves a crash between the writes and the witness (witness
present → commit, absent after grace → abort). So the connection is only needed for the fast
local rows and the one-row witness INSERT.

The Postgres link writers (temporal / semantic / causal) are **skipped for such a store**:
temporal/semantic touch zero rows there (no `memory_units`), and causal edges already travel on
the memory record — writing them to `memory_links` would violate its `DEFERRABLE` FK to
`memory_units` (a latent failure this path avoids).

Adds `EntityResolver.record_unit_entity_postings` — the store-owned entity posting plus the
in-memory co-occurrence accumulation, neither of which needs a connection.

## Tests

`tests/test_retain_ext_writegroup.py` pins the contract with a fake store + a connection
tracker:
- `insert_facts_batch` and the entity re-posting run with **no connection checked out**;
- the witness is written **inside** the short transaction, and `decide(commit=True)` fires only
  **after** the connection is released;
- a later-batch takeover discards the staged write via `decide(commit=False)` — still
  connection-free;
- the outbox row rides the transaction;
- none of the Postgres link writers are invoked.

Existing `test_retain_orchestrator_mapping.py` (Postgres path) stays green. The DB-backed
concurrency/delta/append suites need the `pg0` embedded-Postgres fixture and run in CI.

## Scope / follow-ups (not in this PR)

- The **delta re-retain** write site (`_run_delta_db_work`) has the same conn-hold and should get
  the same treatment; it also adds tombstone store-deletes worth hoisting. Deferred so the pattern
  can be reviewed first (the delta path falls back to streaming on conflict).
- **Collapse the double store-write** for a store-owned backend into a single write carrying the
  entity ids (the currently-unused `index_facts` single-write path), halving the object-store
  writes per retain.

Draft: opening for review of the connection-management pattern before replicating to the delta site.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
+++ b/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
@@ -1242,6 +1242,31 @@ class EntityResolver:
         else:
             return await self._link_units_to_entities_batch_impl(conn, normalized, bank_id)
 
+    async def record_unit_entity_postings(
+        self,
+        unit_entity_pairs: list[tuple[str, str]] | list[tuple[str, str, datetime | None]],
+        bank_id: str | None = None,
+    ):
+        """Store-owned variant of :meth:`link_units_to_entities_batch` that touches NO
+        Postgres connection.
+
+        For a memories store that OWNS its memory rows (an external backend), the unit→entity
+        posting is recorded by the store — ``record_unit_entities`` ignores the ``conn`` — and
+        the co-occurrence update only accumulates in memory for the post-transaction flush.
+        Neither needs a database transaction, so the retain orchestrator can run the posting in
+        its connection-free store phase and never hold the data-plane connection across the
+        object-store write. NOT for the Postgres store, whose posting is a real ``unit_entities``
+        INSERT that requires the connection.
+        """
+        if not unit_entity_pairs:
+            return
+
+        normalized: list[tuple[str, str, datetime | None]] = [
+            (t[0], t[1], t[2] if len(t) >= 3 else None)  # type: ignore[misc]
+            for t in unit_entity_pairs
+        ]
+        return await self._link_units_to_entities_batch_impl(None, normalized, bank_id)
+
     async def _link_units_to_entities_batch_impl(
         self, conn, unit_entity_pairs: list[tuple[str, str, datetime | None]], bank_id: str | None = None
     ):
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -643,7 +643,9 @@ async def _streaming_batch_write_ext(
     # so facts can be tagged with document_id + chunk_id before the metadata rows are written.
     chunk_id_by_index = {}
     if batch_chunk_meta:
-        chunk_id_by_index = {cm.chunk_index: f"{bank_id}_{effective_doc_id}_{cm.chunk_index}" for cm in batch_chunk_meta}
+        chunk_id_by_index = {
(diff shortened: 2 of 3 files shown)


nicoloboschi:
## Dev measurement (single api pod, ext-routed org, chunks-mode retain)

Built an overlay image on the running dev tag and A/B'd old vs. the fix. Closed-loop concurrent
retain, spread across 32 banks (so memlake per-namespace write serialization isn't the ceiling and
the **data-plane PG pool** — `max_size=28` — is the thing under pressure).

| workers | old (conn held across store write) | fix (conn released) |
|--:|---|---|
| 12 | 4.4 rps · p99 6.4s · 0 slow-acq · stable | 4.8 rps · p99 7.4s · 0 slow-acq · stable |
| 20 | 5.4 rps · p99 10s · 0 slow-acq · stable | 5.7 rps · p99 9.4s · 1 slow-acq · stable |
| 32 | **0 served — every request HTTP 500, p99 59s (pool exhausted)** | 5.0 rps · p99 14s · 0 slow-acq · **stable** |
| 48 | **0 served — pod crashed / restarted** | 7.0 rps · p99 8s · 8 slow-acq · **stable** |

Single-bank C=64 baseline on the old image: **58 `[DB POOL] Slow acquire` warnings**, 1.1 rps,
retain p99 62s — the original pool-starvation signature.

**Read:** below the pool size the two are identical (nothing to fix when the pool isn't stressed).
**At/above the pool size the old path hits a hard pool-exhaustion cliff** — 28 concurrent retains
each pin a connection for the whole ~0.8s object-store write, the pool empties, and further
requests 500 (C=32) or the pod dies (C=48). The fix holds a connection only for the short witness
transaction, so the pool never empties and the pod serves through it. The change converts a
catastrophic failure at concurrency ≥ pool size into graceful service. (Per-pod retain throughput
on one dev pod is compute-bound ~5–7 rps and unchanged below the cliff — as expected; the win is
removing the cliff, which is what lets pods scale horizontally without pool starvation.)

PR #3418, opened 2026-08-12
feat(coding-agents): apply retain attribution to all ingestion paths

voarsh2:
Extends the coding-agents integration's existing `retainTags` and `retainMetadata` support to every document it writes, rather than only live session write-back.

This includes:

- Imported conversations
- Aggregated Git history and full commit documents
- Codebase survey lifecycle documents
- Initiative markers
- Documents saved through `hindsight_ingest_document`

Configured attribution is merged with each document's built-in tags and metadata, with built-in source identity remaining authoritative.

## Why

In shared-bank deployments, consistent attribution is needed to identify and filter documents by project or other configured provenance. Partial attribution leaves Git, survey, and manually ingested documents indistinguishable across projects.

##  Notes

This is additive and disabled by default. Existing configurations without `retainTags` or `retainMetadata` retain their current behavior.

## Verification

- 441 coding-agents tests passed
- TypeScript and package build passed
- Declaration generation passed
- Documentation sync passed

### DIFF
--- a/hindsight-docs/docs-integrations/coding-agents.md
+++ b/hindsight-docs/docs-integrations/coding-agents.md
@@ -310,8 +310,8 @@ hook by Codex...), so one shared config serves several agents side by side:
 | `bankIdTemplate`        | `"coding-agent::{gitProject}"`       | dynamic bank id format; the default makes every agent share one bank per repo                                                                                                                                                       |
 | `mapPathToBank`         | —                                    | absolute path → bank; **longest prefix wins**; overrides everything                                                                                                                                                                 |
 | `resolveWorktrees`      | `true`                               | `{gitProject}`: linked worktrees share the main repo's bank                                                                                                                                                                         |
-| `retainTags`            | —                                    | extra tags on every session write-back, e.g. `["project:{gitProject}"]` — see **Recording where a memory came from** below                                                                                                          |
-| `retainMetadata`        | —                                    | extra metadata on every session write-back, e.g. `{"repo": "{gitProject}"}`                                                                                                                                                         |
+| `retainTags`            | —                                    | extra tags on every document written by the integration, e.g. `["project:{gitProject}"]` — see **Recording where a memory came from** below                                                                                         |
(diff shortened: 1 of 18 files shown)


nicoloboschi:
Reviewed and tested locally — rebased onto current main (it was 13 commits behind, predating the cursor fix, Prime Agent and the `retainEveryTurns` removal) and pushed one cleanup. **462 tests pass**, tsc, lint and docs-sync clean.

## What's right

The merge order is correct in every path — configured first, built-ins last, deduped:

```ts
[...new Set([...(stamp?.tags ?? []), "source:git"])]
{ ...stamp?.metadata, source: "git", repo: repoName, … }
```

So source identity stays authoritative, and because every path builds its stamp through `buildRetainStamp`, the reserved-namespace filter comes along for free — a user can't tag a git document `harness:codex`. Making `sessionId` optional and resolving it to `"unknown"` is the right call for documents that have no session.

`runtime.toolSpecs()` passing `repoDir: this.projectDir` also quietly fixes a real inconsistency: `hindsight_diagnose` was reporting `process.cwd()` rather than the resolved workspace.

## What I changed

Both survey-baseline writes branched into two complete `client.retain(...)` calls differing only in whether `{ metadata }` was passed — six duplicated argument lists, which is where a later edit updates one and misses the other. `retain` only assigns metadata when truthy, so `{ metadata: undefined }` already omits it and one call covers both cases; that's what `knowledge-tools.ts` does with `Object.keys(metadata).length ? { metadata } : {}`. The marker assertion in `session-start.test.ts` gains the opts argument to match.

I assume the branching existed to keep that assertion passing unchanged — reasonable, but the test was the thing to update.

## One thing worth knowing

With `retainTags: ["session:{sessionId}"]`, git and survey documents now carry `session:unknown`. Documented and honest, but it is new user-visible output; if you'd rather an unresolvable placeholder dropped the tag instead, say so and I'll follow up separately.

Merging once CI is green. Thanks — partial attribution really was the gap, and this closes it cleanly.

Issue #3222, opened 2026-08-06
graph_maintenance: never completes on multi-million-edge banks — silent TimeoutError, permanent retry churn


nicoloboschi:
## Problem

On a bank whose `memory_links` graph has grown to ~10M edges, every `graph_maintenance` run fails with a bare timeout — logged as an **empty error string** (`Task execution failed: graph_maintenance, error: `, the signature of `asyncio.TimeoutError`'s empty `str()`).

Observed pattern over ~28h on one bank: **78 consecutive failed ops** (each exhausting its retry budget), re-enqueued indefinitely, zero successes. Effects:

- the bank's graph maintenance never completes — relink/cooccurrence hygiene stops permanently once the graph crosses the size threshold;
- each attempt occupies a worker slot for the full timeout duration, several times per hour, forever — a standing tax on fleet capacity;
- the empty error message makes it invisible without correlating op-status history (compounds the traceback gap in #3218).

The existing iteration cap in `run_graph_maintenance_job` bounds work per pass, but the overall run evidently still exceeds whatever task/socket timeout cancels it, and cancellation is treated as a retryable failure rather than 'ran out of budget — persist progress and continue next run'.

## Proposed fix

1. Make graph maintenance **incremental and resumable**: treat the time budget as a first-class limit — on budget exhaustion, checkpoint progress (the queue/cursor already exists) and complete the op successfully, resuming next cycle. A large graph then converges over multiple runs instead of failing forever.
2. Distinguish cancellation/timeout from genuine errors: log it as budget exhaustion (with graph size and progress made), not an empty ERROR.
3. Backstop: after N consecutive timeouts for the same bank, back off the re-enqueue cadence and surface a WARNING naming the bank, instead of hot-looping the failure.

Related: #3218 (empty task-failure logging).

nicoloboschi:
Additional observation on the same bank: the empty-TimeoutError signature now also appears intermittently on `batch_retain` tasks (several per hour, absorbed by task-level retries — retains do eventually complete, ~460/day). Consistent with the graph size taxing the write path generally (temporal/semantic link maintenance against the ~10M-edge table), not just the maintenance job. The retry-absorbed churn burns worker slots and inflates latency; one more reason for the time-budget/incremental approach over hard timeouts.

PR #3420, opened 2026-08-12
fix(retain): enqueue relink victims before delta chunk deletion

koriyoshi2041:
## Problem

Delta retain deletes changed and removed chunks directly. The cascade drops their memory links, but unlike full document replacement, this path does not enqueue surviving units whose outgoing temporal or semantic links pointed at the deleted facts. Graph maintenance runs afterwards, but by then the links needed to discover those victims are gone.

This is the relink-victim residue called out as out of scope in #3384.

## Fix

Reuse the outgoing fact IDs already collected for observation invalidation and enqueue relink victims before the chunk/link cascade. The store abstraction keeps this a no-op for stores that own links inline.

The existing real-Postgres endpoint-deletion test now also asserts that the surviving source whose outgoing link targeted the deleted chunk is present in `graph_maintenance_queue`.

## Test

- `uv run pytest tests/test_chunk_storage_delete_ordering.py::test_delete_chunks_by_ids_sweeps_links_on_both_endpoints -q -n 0` (1 passed)
- `uv run ruff check hindsight_api/engine/retain/chunk_storage.py tests/test_chunk_storage_delete_ordering.py`
- `uv run ruff format --check hindsight_api/engine/retain/chunk_storage.py tests/test_chunk_storage_delete_ordering.py`
- `git diff --check`

Fixes the relink-victim follow-up noted in #3384.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/chunk_storage.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/chunk_storage.py
@@ -147,6 +147,16 @@ async def delete_chunks_by_ids(conn, chunk_ids: list[str], bank_id: str | None =
             # (the full-replace path enqueues in ``handle_document_tracking``).
             await enqueue_entity_prune_candidates(conn, bank_id, outgoing_unit_ids)
 
+            # Capture surviving units whose temporal/semantic links point at
+            # the outgoing facts before the link/chunk cascade below removes
+            # the evidence needed to find them. Full document replacement does
+            # the same in ``handle_document_tracking``; without it, a delta
+            # edit leaves survivors permanently below their configured link
+            # caps even though retain submits graph maintenance afterwards.
+            from ..graph_maintenance import enqueue_relink_victims
+
+            await enqueue_relink_victims(conn, bank_id, outgoing_unit_ids)
+
     # The chunks->memory_units FK cascade below does not reach a store that keeps memories
     # outside SQL (its memory_units is empty), so drop the memories carrying each deleted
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Verified on current `main` (`e32e6f246`).

Confirmed the bug still reproduces: with only the new assertion applied and not the fix, `graph_maintenance_queue` comes back empty after a delta chunk delete — the surviving unit whose outgoing link targeted the deleted fact is never enqueued, and the relink pass is purely queue-driven (`store.relink_pass` → `claim_graph_maintenance_batch`, no bank-wide sweep since #3222/#3409), so nothing self-heals it. v0.9.1 shipped the observation sweep (#3384) and the entity-prune enqueue (#3409) into this same function but not the relink half, exactly as #3384 called out.

The placement is right: same txn as the cascade and before it, which is mandatory — the victim query reads the `memory_links` rows the delete is about to drop. Delegating through the store keeps it a no-op for link-inline backends.

Note that `test-api` is skipped on fork PRs (no secrets), so the new test never ran in CI. Ran it locally on a freshly migrated DB instead: 147 passed, 1 skipped across test_chunk_storage_delete_ordering, test_delta_retain{,_duplicates,_orphan_observations}, test_graph_maintenance{,_deadlock,_queue_race}, test_enqueue_graph_maintenance_ordered, test_memory_curation, test_link_utils, test_chunk_storage_upsert, test_document_chunks_and_reprocess. ruff check / format --check / ty all clean.

Nit, not blocking: the new `from ..graph_maintenance import enqueue_relink_victims` could fold into the `enqueue_entity_prune_candidates` import three lines above — that is what every other call site does.

PR #3430, opened 2026-08-12
fix(engine): scope document statements by bank_id (#3429)

### LINKED ISSUE #3429: update_document re-tags memory units in other banks that share the same document_id
### Bug Description

In `MemoryEngine.delete_document` and `MemoryEngine.update_document`, four statements filter `memory_units` on `document_id` alone. `bank_id` is a parameter of both methods and is used by neighbouring statements in the same block.

`memory_engine.py` on `main` (64ede361):

- `:7159` — `SELECT id ... WHERE document_id = $1 AND fact_type IN (...)`
- `:7164` — `SELECT COUNT(*) ... WHERE document_id = $1`
- `:7348` — `SELECT id ... WHERE document_id = $1 AND fact_type IN (...)`
- `:7354` — `UPDATE ... SET tags = $1 WHERE document_id = $2`

The `UPDATE` is a cross-bank write: re-tagging a document rewrites `tags` on the memory units of every bank holding a document with that id.

Beyond correctness, the two `SELECT`s make the cost of both endpoints scale with the number of banks rather than with the caller's data. With N banks holding a document under the same id, `:7159` and `:7348` return N times the caller's rows, and that id list is then passed as a bind parameter into two further statements — the observation overlap check (`source_memory_ids && $2::uuid[]`) and the relink enqueue (`to_unit_id = ANY($1::uuid[])`). On a bank-per-user deployment, where a role-named document exists in every bank, this grows linearly with signups, on a write path.

Deletion itself is not affected: the `DELETE` carries `bank_id`, and `memory_units_document_fkey` is composite, `(document_id, bank_id) ON DELETE CASCADE`.

Two notes on why this is reachable rather than theoretical. `documents` is keyed on `(id, bank_id)`, so the same `document_id` in several banks is a supported state — and under the bank-per-user layout the docs recommend, naming a document by its role (`onboarding`, `profile`) gives every bank the same id. And in both methods, the branch of the same `if` that goes through the non-SQL memories store already passes `bank_id` (`:7167`, `:7330`); only the SQL branch omits it.

### Steps to Reproduce

1. Retain a document under the same `document_id` in two banks, `bank_a` and `bank_b`.
2. `PATCH /v1/default/banks/bank_a/documents/<document_id>` with `{"tags": ["rewritten"]}`.
3. List `bank_b`'s memory units for that document.

For the count: call `DELETE /v1/default/banks/bank_a/documents/<document_id>` on the same setup and compare `memory_units_deleted` with the number of units `bank_a` actually held.

### Expected Behavior

Only the caller's bank is read and written. `bank_b`'s memory units keep their tags, and `memory_units_deleted` reports the units deleted in `bank_a`.

### Actual Behavior

`bank_b`'s memory units carry `rewritten`. `memory_units_deleted` returns 8 where `bank_a` held 4 — the `SELECT` returned both banks' units, and the same list was passed to the downstream cleanup statements.

### Version

0.8.6, and present on `main` at 64ede361.

raf52:
Closes #3429.

`delete_document` and `update_document` filter `memory_units` on `document_id` alone in four statements, although `bank_id` is a parameter of both methods and is used by neighbouring statements:

- `:7159`, `:7164` — read and count the units of every bank sharing the id, so `memory_units_deleted` over-reports and the id list handed to the downstream cleanup carries other banks' ids.
- `:7348` — the same read, on the update path.
- `:7354` — `UPDATE ... SET tags` with no bank predicate: re-tags the memory units of other banks.

Beyond isolation, scoping the reads makes the cost of both endpoints depend on the caller's data rather than on the number of banks sharing the document id — on a bank-per-user deployment, where a role-named document exists in every bank, that cost currently grows with signups.

In both methods, the branch of the same `if` that goes through the non-SQL memories store already passes `bank_id` (`:7167`, `:7330`). This aligns the SQL branch with it.

Deletion itself was not affected: the `DELETE` carries `bank_id`, and `memory_units_document_fkey` is composite `(document_id, bank_id) ON DELETE CASCADE`.

## Tests

Two tests added to `tests/test_document_tracking.py`, each with two banks sharing one `document_id`:

- `test_document_deletion_counts_only_its_own_bank_units`
- `test_document_tag_update_is_scoped_to_its_bank`

Without the change:

```
E   AssertionError: assert 'rewritten' not in (['rewritten'])
E   assert 8 == 4
2 failed
```

With the change: `2 passed`.

## CI note

The Python test jobs are gated on secrets and do not run on fork PRs, so the suite was run locally: 5348 passed, 21 failed. Seven are pre-existing and caused by the absence of an LLM API key (`test_fact_extraction_output_ratio`, `test_retain`) — they fail identically on `main` without this change. The other fourteen only appear under full-suite parallel execution and pass in isolation (`18 passed`); they come from a mock shared across tests and an asyncio lock bound to a different event loop, unrelated to this change.

`ruff check hindsight_api` and `ruff format --check` both pass.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -7156,12 +7156,15 @@ class MemoryEngine(MemoryEngineInterface):
                 _store = get_memories()
                 if _store.writes_memory_rows_in_sql_for(bank_id):
                     unit_rows = await conn.fetch(
-                        f"SELECT id FROM {fq_table('memory_units')} WHERE document_id = $1 AND fact_type IN ('experience', 'world')",
+                        f"SELECT id FROM {fq_table('memory_units')} WHERE document_id = $1 AND bank_id = $2 AND fact_type IN ('experience', 'world')",
                         document_id,
+                        bank_id,
                     )
                     unit_ids = [str(row["id"]) for row in unit_rows]
                     units_count = await conn.fetchval(
-                        f"SELECT COUNT(*) FROM {fq_table('memory_units')} WHERE document_id = $1", document_id
+                        f"SELECT COUNT(*) FROM {fq_table('memory_units')} WHERE document_id = $1 AND bank_id = $2",
+                        document_id,
+                        bank_id,
                     )
                 else:
                     src_page = await _store.scan_memories(
@@ -7345,15 +7348,17 @@ class MemoryEngine(MemoryEngineInterface):
                         )
                 elif tags is not None:
                     unit_rows = await conn.fetch(
-                        f"SELECT id FROM {fq_table('memory_units')} WHERE document_id = $1 AND fact_type IN ('experience', 'world')",
+                        f"SELECT id FROM {fq_table('memory_units')} WHERE document_id = $1 AND bank_id = $2 AND fact_type IN ('experience', 'world')",
                         document_id,
+                        bank_id,
                     )
                     unit_ids = [str(row["id"]) for row in unit_rows]
 
                     await conn.execute(
-                        f"UPDATE {fq_table('memory_units')} SET tags = $1 WHERE document_id = $2",
+                        f"UPDATE {fq_table('memory_units')} SET tags = $1 WHERE document_id = $2 AND bank_id = $3",
                         tags,
                         document_id,
+                        bank_id,
                     )
 
                     if unit_ids:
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Thanks for catching this @raf52 — it's a genuine cross-bank isolation defect (including a cross-bank `UPDATE ... SET tags` write), and your fix and tests are spot on.

I've moved this to #3435, which carries your commit **verbatim with your authorship preserved**, and adds a full bank/tenant scoping audit of the surrounding engine query surface (confirming these four statements were the only instances of the pattern) plus a couple of hardening follow-ups. Doing it from a branch on the main repo also lets CI actually run the test suite, which it can't on a fork PR.

Closing in favour of #3435. Credit is all yours — thank you for the report and the fix. 🙏

PR #3439, opened 2026-08-12
feat(coding-agents): add opt-in project scoping within shared banks

voarsh2:
Adds an opt-in tag-scoped mode for installations that deliberately keep multiple projects in one Hindsight bank.

This complements the new `optInOnly` / per-project routing support rather than replacing it:

- `optInOnly` controls which projects may use memory at all.
- Separate dynamic banks remain the default and provide strong isolation.
- This PR supports the different deployment shape where related opted-in projects intentionally share one established bank, while ordinary retrieval should stay project-local.

It builds on #3418, which supplies consistent ingestion attribution across coding-agent write paths.

## Configuration

```json
{
  "bankId": "team::shared",
  "projectScope": "tags",
  "projectTagTemplate": "project:{gitProject}",
  "globalTags": ["scope:global"]
}


nicoloboschi (changes requested):
I'd prefer to not complicate things here. the bank mapping to a repository is pretty clear to everyone and it works out of the box. adding tags is powerful but also carries complexity and it's error prone 

so unless there's a real need, I'd decline this feature, thanks

PR #3442, opened 2026-08-12
fix(files): add OCR fallback for image-only PDFs

### LINKED ISSUE #3255: file conversion: image-only (scanned) PDFs fail with 'no content extracted' — no OCR fallback
## Problem

PDFs without a text layer (scans, image-exported documents) fail file conversion outright:

```
Failed to parse file '<name>.pdf': Parser 'markitdown' returned no content for '<name>.pdf'
```

Observed on multiple tenants uploading real-world documents (book scans, whitepapers exported as images). The failure is correct given markitdown has no OCR path, and the error message is clear — but from the customer's perspective their document simply cannot be ingested, with no recourse inside the product.

Secondary nit: the failure is deterministic per file yet consumes the full task retry budget before going terminal (same classification gap as other deterministic parse failures) — 'parser produced no content' should be non-retryable on first occurrence.

## Proposed direction

1. Detect the no-text-layer case and route to an OCR-capable fallback where the deployment has one configured (extraction-service backends may already support it), before failing.
2. If no OCR path exists in the deployment, fail fast (non-retryable) with an error that tells the user *why* ('this PDF appears to be scanned images without selectable text') rather than the generic no-content message.
3. Classification fix: mark deterministic parser failures non-retryable.

Sanderhoff-alt:
## Summary

Image-only and scanned PDFs currently fail when MarkItDown cannot extract a text layer, even when MarkItDown OCR is configured. This change adds a PDF-specific OCR fallback while preserving the existing parser fallback chain.

## Problem

MarkItDown's configured vision model is used for image inputs, but its PDF converter only attempts text-layer extraction. A scanned PDF can therefore return no content or raise during conversion without any OCR attempt.

## Design

When MarkItDown conversion of a PDF produces no usable text or raises an exception, the parser renders the PDF one page at a time at 200 DPI using pypdfium2. Each page is written to a temporary PNG, passed through the existing MarkItDown image OCR path, and deleted immediately after processing. Successful page results are combined in document order. A failed page is logged and does not discard successful results from other pages.

If OCR is disabled, unavailable, or produces no content, the parser raises an actionable error. The existing parser registry then continues to the next configured parser, so configurations such as `markitdown,llama_parse` retain their expected fallback behavior.

## Dependency and Licensing

PDF rendering uses the explicitly declared `pypdfium2>=5.4.0` dependency. This avoids adding PyMuPDF's AGPL/commercial dual-licensing requirements to the MIT-licensed package. The lockfile is updated accordingly.

## Testing

Regression tests cover empty PDF extraction, PDF converter exceptions, successful per-page OCR, partial page failures, OCR-disabled errors, OCR failure without duplicate attempts, and advancement to the next parser in the configured fallback chain. The focused test suite, project lint, type checks, and lockfile validation pass.

Fixes #3255

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/parsers/markitdown.py
+++ b/hindsight-api-slim/hindsight_api/engine/parsers/markitdown.py
@@ -33,6 +33,8 @@ _TEXT_EXTENSIONS = {
     ".html",
     ".htm",
 }
+_PDF_EXTENSION = ".pdf"
+_PDF_OCR_DPI = 200
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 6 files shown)


nicoloboschi:
I'd prefer to create another provider that is more complete, markitdown is very simple but also not vllm optimized at all, closing for now

PR #3444, opened 2026-08-12
Dedupe graph maintenance submits against running jobs, not just pending ones

cdbartholomew:
## Problem

`run_graph_maintenance_job` documents that it relies on submit-time dedup to keep at most one job per bank in flight, and skips `SKIP LOCKED` on that basis. That guarantee didn't hold.

Dedup matched only `status = 'pending'`. Under load a pending row is claimed within milliseconds, so the next submit sees no pending row and inserts another job. Every subsequent trigger does the same. A bank taking sustained writes accumulates roughly one maintenance job per triggering write — many concurrent jobs against a single bank, each doing the same bank-wide sweeps, while the job body is written assuming it is the only one running.

## Fix

Widen the dedup predicate to `status IN ('pending', 'processing')`, behind an opt-in flag (`dedupe_by_bank_includes_processing`).

Deduping against a running job means a submit made *while* the job runs is suppressed, so whatever it queued must be picked up by that job or it strands. The job therefore hands off at the end: if the queue is non-empty when it finishes, it submits a successor. That hand-off is gated on `made_progress` — without the gate, a run that drains nothing and still sees queued work would submit a successor that repeats the outcome, an endless per-bank chain.

### The hand-off defeated itself (second commit)

The first version of this branch shipped a hand-off that never ran. The submitting job is still `processing` while its body executes — the worker marks the operation completed only after the body returns — so the hand-off matched **its own row** and was deduplicated away. It was dead code, and the gap it exists to close stayed open.

The original tests could not see this: they monkeypatch `submit_async_graph_maintenance` and assert the *call* happens, not that an operation results.

Fix: the caller can name an operation the dedup scan must ignore (`dedupe_excludes_operation_id`), and the hand-off passes the running job's own id. `run_graph_maintenance_job` already receives it from `_handle_graph_maintenance`, so nothing new is plumbed to the worker. The comparison is done in Python alongside the existing scope check, because the `operation_id` cast is not portable across the Postgres and Oracle backends.

## Why opt-in rather than global

Consolidation self-resubmits when it hits `max_memories_per_round`, and is safe today **only because** it does not match `processing` — its self-resubmit would otherwise be suppressed by its own row, the identical defect. Now that the exclusion mechanism exists, consolidation could adopt the widened predicate with the same one-line change, but that is a behaviour change to a second subsystem and belongs in its own PR.

(An earlier draft of this description justified the flag by claiming consolidation "fixes a watermark at start". That was wrong — the code explicitly picks up memories retained mid-run. The real reason is the self-resubmit above.)

## Implementation note

The status filter is a literal `IN (...)` rather than `= ANY($n)`. The array form is PostgreSQL-only and this path also targets Oracle.

## Validation

`tests/test_graph_maintenance_submit_dedup.py` — 10 tests against a real PostgreSQL instance, not mocks, since the change is a SQL predicate.

**Behaviour covered**
- a pending job still dedupes a second submit (pre-existing behaviour, pinned)
- a **processing** job dedupes a new submit (the fix)
- sustained triggers against a claimed job produce one job, not one per trigger
- a completed job does **not** dedupe, so later work still drains
- consolidation still ignores `processing` (the opt-in must not leak)
- hand-off fires when work lands mid-run; not when the queue is empty; not when the run drained nothing

**The hand-off, proven end-to-end through the real worker path**

The direct-call tests set `processing` with a hand-written `UPDATE`, which leaves the integration unproven — and the integration is where this fix lives. `test_handoff_survives_the_real_worker_path` drives the real task payload the backend receives, the real `ops.mark_operations_processing` claim, and the real `execute_task` router:

| | status after real claim | graph_maintenance operations |
|---|---|---|
| with self-exclusion | `processing` | **1 → 2** |
| without self-exclusion | `processing` | **1 → 1** |

This confirms empirically, rather than by reading the code, that the claim marks the row `processing` before the body runs and that `execute_task` threads `operation_id` down to the hand-off. Both are load-bearing: if either were false the exclusion would never apply and the hand-off would be silently dead again.

**Mutation testing** — each mutant fails the tests that should catch it, and the suite returns to green when reverted:

| mutant | result |
|---|---|
| narrow predicate back to `status = 'pending'` | kills `test_processing_job_dedupes_a_new_submit`, `test_sustained_triggers_do_not_stack_jobs` |
| remove the hand-off self-exclusion | kills both hand-off tests (`1 → 1`) |

**Wider regression** — 245 tests pass across the graph-maintenance, consolidation, operation-status/completion, async-retain and worker test files. `test_consolidation_round_limit` and `test_consolidation_reschedule_after_round` specifically exercise consolidation's self-resubmit path and are unaffected.

**Call-site audit** — both new parameters are keyword-only with `None` defaults. All six `_submit_async_operation` callers and the five non-hand-off `submit_async_graph_maintenance` callers are unchanged; the only behavioural change is the hand-off itself.

## Not verified

- **No deployment.** This has never run against real traffic in any environment.
- **Concurrency untested.** Simultaneous submits from two workers go through a `serialize` path that was read but not exercised. Since the defect being fixed is about concurrent job creation, this is the most significant remaining gap.
- **Oracle untested.** The comparison is Python-side and should be portable, but has not been executed against that backend.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/graph_maintenance.py
+++ b/hindsight-api-slim/hindsight_api/engine/graph_maintenance.py
@@ -246,73 +246,98 @@ async def run_graph_maintenance_job(
     result.stale_cooccurrences_pruned = prune.stale_cooccurrences_pruned
     result.queues_drained = relink.queue_exhausted and prune.queue_exhausted
 
-    # Close the hand-off gap. Submit-time dedup treats a *running* job as
-    # covering this bank (see _submit_async_operation's
-    # dedupe_by_bank_includes_processing), which is what stops one job being
-    # queued per triggering operation. The cost is a narrow window: anything
-    # enqueued after Pass 1's final claim — including during the sweeps above,
-    # which are not instant — was deduped against this job and would otherwise
-    # sit in graph_maintenance_queue until some unrelated future operation
-    # happened to trigger another submit. For the last write of an ingest batch
-    # that may be never.
-    #
-    # So re-check before finishing and hand off to a fresh job if anything
-    # landed.
-    #
-    # Gated on this run having drained something. A run that processed nothing
-    # and still sees a non-empty queue has not made progress, and a successor
-    # would repeat that exact outcome — an endless self-resubmitting chain for
-    # the bank. Requiring progress means the chain can only continue while it
-    # is actually consuming the queue, so it terminates. (The normal case
-    # always qualifies: submit_async_graph_maintenance short-circuits on an
-    # empty queue, so a job only starts when there is work to drain.)
-    made_progress = result.relink_units_processed > 0
-    try:
-        backend_check = await memory_engine._get_backend()
-        async with acquire_with_retry(backend_check) as conn:
-            leftover = await conn.fetchval(
-                f"SELECT 1 FROM {fq_table('graph_maintenance_queue')} WHERE bank_id = $1 LIMIT 1",
-                bank_id,
-            )
-        if leftover and not made_progress:
-            logger.warning(
-                f"[GRAPH_MAINT] bank={bank_id} queue still non-empty after a run that drained "
-                f"nothing; not chaining a successor (it would repeat this outcome)"
-            )
-        elif leftover:
-            logger.info(f"[GRAPH_MAINT] bank={bank_id} work arrived during the run; submitting a follow-up job")
-            await memory_engine.submit_async_graph_maintenance(
-                bank_id=bank_id,
-                request_context=request_context,
-                dedupe_excludes_operation_id=operation_id,
-            )
-    except Exception:
-        # Never fail a completed maintenance run over the hand-off. The work is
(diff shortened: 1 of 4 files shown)


koriyoshi2041:
I reproduced the deterministic `verify-generated-files` failure at `7dd0c071e`. Running `./scripts/hooks/lint.sh` rewrites only the `_record_submit` signature in `tests/test_graph_maintenance.py` for Python formatting:

```python
async def _record_submit(
    *, bank_id: str, request_context, dedupe_excludes_operation_id=None, force_sweep: bool = False
):
```

`git diff --check` is clean after that rewrite. The local all-repo lint run then stopped at ESLint because this disposable worktree did not have the control-plane Node dependencies installed, so I did not use it as evidence for the unrelated frontend files.

The `test_completed_refresh_enriches_result_metadata` shard failure does not touch this PR's files and its log shows the operation disappearing before `mark-completed`; I would treat that separately from the formatting fix.

Issue #3443, opened 2026-08-12
coding-agents: bounded hook reflect silently degrades to a memoryless turn when it overruns HOOK_REFLECT_CAP_MS


arturbasinki:
### Bug Description

The once-per-session hook reflect in `hindsight-integrations/coding-agents` regularly exceeds `HOOK_REFLECT_CAP_MS` on a real bank, and when it does the failure is **completely silent**: `buildHookOutput` catches the abort, sets `reflectAnswer = ""`, and the turn proceeds with no memory injection and nothing visible to the user.

To be clear up front: I am **not** asking you to raise the cap. The invariant documented at `src/core/hook.ts:65-72` is correct — the cap must stay below the harness hook timeout, or the host kills the hook before the result is cached and reflect re-fires uncached every turn. This report is about two other things:

1. the failure being invisible to the user, and
2. **why** the bounded call overruns in the first place — which turns out to have a cheap fix that keeps your invariant intact.

### Steps to Reproduce

1. Self-hosted Hindsight, bank seeded normally from a repo (~150 nodes / ~2300 edges here).
2. Global LLM chain: `deepseek` (`deepseek-v4-flash`) → local llama.cpp as failover.
3. Start a Claude Code session and send a second prompt (the first is deferred via `deferInitialReflect`).
4. Watch `/tmp/hindsight-plugin.log`.

### Expected Behavior

Either the bounded reflect completes within the cap, or — when it cannot — the user gets some signal that this turn is running without memory. The `notice` channel already exists in `HookOutput` for user-facing lines.

### Actual Behavior

Both automatic reflects in my logs aborted at the cap, and nothing surfaced:

```
{"event":"reflect_failed","ms":25016,"error":"This operation was aborted","query":"..."}
{"event":"reflect_failed","ms":25009,"error":"This operation was aborted","query":"..."}
```

Two sessions, two failures, zero user-visible signal. I only found it because I went looking in the diag log for an unrelated reason. The second one ate the prompt in which I was asking about this very limit.

Measured latency of `POST /memories/reflect` at `budget: "low"` (the same budget the hook uses), on the bank above:

| query | wall time |
|---|---|
| narrow ("which model handles retain and why") | 20.5 / 21.8 / 23.8 / 32.5 s |
| broader ("how are changes deployed, what are the pitfalls") | 78.5 s, >120 s, >120 s |

[...]

PR #3445, opened 2026-08-13
fix(coding-agents): expose automatic reflect failures

Sanderhoff-alt:
## Summary

The coding-agents prompt hook can time out during its once-per-session automatic reflect. The current fallback caches an empty string and emits no user-visible signal, so the agent continues without synthesized memory while the user cannot distinguish that from “no relevant memory exists”.

This PR makes that fallback explicit without changing the bounded-hook design.

## Design

- Persist a separate `reflectStatus` in the session cache with `success`, `timeout`, or `error`.
- Keep `reflectAnswer` for the answer payload and retain compatibility with caches written before `reflectStatus` existed.
- Treat any recorded outcome as an attempted automatic reflect, so a failed request is still not retried on every subsequent prompt.
- Emit a notice on the turn that automatic reflect fails, distinguishing timeout from other errors and pointing the agent/user to manual `hindsight_reflect`.
- Classify timeout using the complete provider error, while limiting the logged/diagnostic error text to 200 characters.

## Scope and non-goals

The 25-second cap is unchanged because it must remain below the host hook timeout; increasing it alone could cause uncached re-fires. This PR does not add automatic retry/backoff. It also does not force a generic `thinking: disabled` request parameter, since reasoning controls are provider/model-specific and can affect quality or API compatibility. Provider-specific tuning remains available through `HINDSIGHT_API_REFLECT_LLM_EXTRA_BODY`. Retrieval/reranker performance is a separate optimization.

## Testing

- Added coverage for success, timeout, error, cache state, no-retry behavior, and long error messages.
- `npm test`: 494 tests passed.
- `npm run build`: passed.

Partially addresses #3443

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/hook.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/hook.test.ts
@@ -67,6 +67,7 @@ describe("buildHookOutput", () => {
     const cached = JSON.parse(readFileSync(cacheFile, "utf8"));
     expect(cached.turns).toBe(1);
     expect(cached.reflectAnswer).toBe("REFLECT_ANSWER");
+    expect(cached.reflectStatus).toBe("success");
   });
 
   it("reflect runs ONCE per session: cached on turn 1, not called again on turn 2", async () => {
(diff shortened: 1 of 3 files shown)


nicoloboschi:
it's already visible in the logs, that is enough to troubleshoot

PR #3465, opened 2026-08-13
docs(examples): add TEI embeddings + reranker docker-compose example

nicoloboschi:
## What

Adds `docker/docker-compose/tei/` — a runnable Docker Compose example that serves **embeddings and reranking from two HuggingFace Text Embeddings Inference (TEI) sidecars**, with the **slim** Hindsight image talking to them over HTTP.

Previously there was no Compose example for TEI: it was only wired up in the Helm chart (`helm/hindsight/templates/tei-*`). This closes that gap for local/self-hosted Docker users.

## Contents

- `docker-compose.yaml` — `tei-embedding` + `tei-reranker` (both `ghcr.io/huggingface/text-embeddings-inference:cpu-1.8.3`) + `hindsight:latest-slim`, wired via `HINDSIGHT_API_{EMBEDDINGS,RERANKER}_PROVIDER=tei` and the `*_TEI_URL` service DNS names. `depends_on: service_healthy` gates the API on both TEI servers being ready (curl `/health` healthcheck).
- `README.md` — quick start, what-it-runs table, using-your-own-models, verification curls, and an Apple Silicon (`linux/amd64` emulation) note.
- `models.mdx` — links the runnable example from the TEI embeddings section (mirrors the existing `local-llm` pointer).

Serves Hindsight's **default** models (`BAAI/bge-small-en-v1.5` 384-dim + `cross-encoder/ms-marco-MiniLM-L-6-v2`), so it's a drop-in "move embeddings/reranking onto a dedicated inference server" demo. Embedding dimension is auto-detected from the server.

## Testing

Brought the full stack up and verified end-to-end:

- Both TEI servers reach `healthy`; `/embed` returns a **384**-length vector.
- `POST .../memories` (retain) → `200 {"success": true}`.
- `POST .../memories/recall` for *"where does the user live?"* returns the expected memory *"I live in Paris and work on databases."*, with the `scores` object populated by **both** TEI services — `semantic: 0.637` (embedding server) **and** `reranker: 0.0081` (reranker server).

(Smoke test used the `mock` LLM provider so retain/recall could complete without a real LLM key — the LLM path is orthogonal to what this example changes.)

### DIFF
--- a/docker/docker-compose/tei/README.md
+++ b/docker/docker-compose/tei/README.md
@@ -23,15 +23,18 @@ sidecars required.
 
 ## What it runs
 
-| Service         | Image                                                  | Model (default)                          |
+| Service         | Image                                                  | Model                                    |
 | --------------- | ------------------------------------------------------ | ---------------------------------------- |
 | `tei-embedding` | `ghcr.io/huggingface/text-embeddings-inference:cpu-1.8.3` | `BAAI/bge-small-en-v1.5` (384-dim)       |
-| `tei-reranker`  | `ghcr.io/huggingface/text-embeddings-inference:cpu-1.8.3` | `cross-encoder/ms-marco-MiniLM-L-6-v2`   |
+| `tei-reranker`  | `ghcr.io/huggingface/text-embeddings-inference:cpu-1.8.3` | `BAAI/bge-reranker-base`                 |
 | `hindsight`     | `ghcr.io/vectorize-io/hindsight:latest-slim`           | — (slim; talks to the sidecars)          |
 
-These are Hindsight's default embedding/reranker models, so this is a drop-in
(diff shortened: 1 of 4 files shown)


nicoloboschi:
Follow-up commit d1b34ea: switched the example to a prod-like configuration — reranker is now `BAAI/bge-reranker-base` (the cross-encoder commonly paired with bge-small embeddings on dedicated inference servers), and both TEI services now carry explicit throughput flags (`--max-concurrent-requests`, `--max-batch-tokens`, `--max-client-batch-size`) instead of bare defaults, so the example doubles as a starting point for real deployments. Note the end-to-end test numbers in the description (reranker score 0.0081) were captured with the previous `ms-marco` reranker; `bge-reranker-base` produces a different score scale.

PR #3472, opened 2026-08-13
fix(coding-agents): honor retainSessions in stop hooks

altaywtf:
## Summary

Honor `retainSessions: false` in Stop-hook harnesses as persistent plugin harnesses already do. The fully resolved global, harness, and bank configuration is checked before daemon startup or API activity.

## Risk

The default remains `true`; only explicit retention opt-outs change behavior.

## Verification

- 535 tests passed
- package build passed
- generated coding-agent docs match
- changed TypeScript passes Prettier
- rebase range-diff is patch-equivalent


koriyoshi2041 (approved):
Independently reproduced the missing Stop-hook gate and reviewed the resolved-config placement. The check runs after harness/bank overrides and before daemon/client activity, so it preserves recall/tools while honoring global, harness, and bank-specific opt-outs. The integration coverage exercises both Codex and Cursor stop entrypoints plus the daemon-start early exit. I also verified the coding-agents package locally: build succeeded and all 571 tests passed.

nicoloboschi:
duplicated, already fixed in main

nicoloboschi:
Superseded by #3607, which landed the same fix on `main` on 15 Aug (for #3596).

`retain-hook.ts` already has the `if (!cfg.retainSessions) return` guard in the same place — after bank resolution, so a `banks.<id>` section can re-enable write-back under a global opt-out. That version also emits a `retain_disabled` diagnostic, and its test is a family-wide structural guard ("every session write-back path honors retainSessions") that enumerates the source files and asserts each write-back path consults the flag — so the next harness to forget it fails a test rather than shipping silently.

Closing as already fixed. Thanks for catching it.

PR #3478, opened 2026-08-13
retain: let a store that owns its vector index skip the Postgres ANN pass

nicoloboschi:
## What

Adds a backend-agnostic capability flag, `derives_semantic_links_internally` (with a per-bank `derives_semantic_links_internally_for`), that lets a store which owns its own vector index skip the retain-time Postgres kNN work that populates `memory_links`.

## Why

The default SQL stores have no standing kNN index, so retain finds each unit's semantic neighbours with a pgvector ANN probe and materializes them as `semantic` `memory_links` rows, which the SQL graph arm reads back at recall time.

A store that keeps its own vector index can answer the same question from that index (at read time, or from links it derived at fold time), so the Postgres work is redundant for it — and for a store whose memories don't live in `memory_units` at all it is worse than redundant: it queries a table holding none of its rows and writes `memory_links` rows nothing on its read path consults.

## How

The flag is honoured at three choke points, so **every** retain path — streaming (full), delta, and bank import — inherits the skip and no caller can reintroduce the query by forgetting to pass a flag down:

- `engine/retain/orchestrator.py`
  - `_pre_resolve_phase1` — skips the ANN probe (this is the probe delta retain and bank import run).
  - `_insert_facts_and_links` — skips the `memory_links` write. Necessary on its own: the within-batch similarities are computed in Python, so skipping the probe alone would still leave rows behind.
  - `_run_final_semantic_ann` — skips streaming retain's post-commit pass, which reaches pgvector directly rather than through Phase 1.
- `engine/memories/base.py` — adds the class attr `derives_semantic_links_internally: bool = False` and `derives_semantic_links_internally_for(bank_id)`, mirroring the existing `writes_memory_rows_in_sql_for` / `owns_document_store_for` pattern.

Setting the flag True is a promise, not just an optimization, and the docstring says so: the store's `graph_retriever` / `graph_direct_links` must serve the semantic neighbours itself, because afterwards there are no `semantic` rows in `memory_links` for the SQL graph arm to find.

## Tests

`tests/test_retain_orchestrator_mapping.py` drives all three gates through the orchestrator — each test fails if its gate is removed — and covers the per-bank answer (one store, one bank it owns the index for and one it leaves in SQL). `tests/test_memories_extension.py` keeps the default-False assertion.

Default is `False`, so there is **no behaviour change** for existing stores.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memories/base.py
+++ b/hindsight-api-slim/hindsight_api/engine/memories/base.py
@@ -452,6 +452,22 @@ class MemoriesExtension(Extension, ABC):
         that keeps some banks in a separate backend overrides it. See :meth:`writes_memory_rows_in_sql_for`."""
         return self.owns_document_store
 
+    #: Whether this store derives the semantic (kNN) link graph itself, so retain must NOT run the
+    #: Postgres ANN pass that populates ``memory_links``. Default False: the SQL stores have no
+    #: standing kNN index, so retain computes each committed unit's semantic neighbours with a
+    #: pgvector ANN pass and writes ``memory_links`` rows, which the recall graph arm then reads.
+    #: A store that keeps its own vector index can serve the same neighbours from it (at read time,
+    #: or from links it derived at fold time) — for such a store the Postgres pass is pure waste: it
+    #: scans a ``memory_units`` table that is empty for its banks and writes ``memory_links`` rows
+    #: nothing on its read path consults. Setting this True makes retain skip that pass entirely.
+    derives_semantic_links_internally: bool = False
+
+    def derives_semantic_links_internally_for(self, bank_id: str) -> bool:
+        """Per-bank form of :attr:`derives_semantic_links_internally`. Defaults to the class
+        attribute; a store that owns its vector index for only some banks overrides it. See
+        :meth:`writes_memory_rows_in_sql_for`."""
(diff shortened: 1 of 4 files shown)


nicoloboschi:
Closing — the effect this was after is already there, and the branch now conflicts with #3560.

**Why it is no longer load-bearing.** For a store-owned bank the Postgres ANN pass this skips does
not do any work anyway: Phase 1's ANN is already skipped unconditionally in streaming mode
(`skip_semantic_ann=True`, deferred to Phase 3), and Phase 3 builds `ann_unit_ids` from
`memory_units`, which such a bank leaves empty — so it returns at `if not ann_unit_ids` before
querying pgvector or writing a single `memory_links` row. such a store derives semantic links from its
own clusters at read time and stores none, so nothing is lost.

**What is genuinely left of it** is smaller than the PR: one wasted embeddings load per retain, and
a `[streaming] Final ANN: no units found in DB (unexpected)` log line that says "unexpected" about
the expected state. Both are worth fixing, neither needs a capability flag on the interface — and
if a flag does turn out to be the right shape, it should be added on top of #3560 rather than in a
branch that predates it.

**Practical reason too:** this touches `memories/base.py` and `retain/orchestrator.py`, both of
which #3560 rewrites substantially, so keeping it open means resolving that conflict for an
optimization that currently changes no work done.

PR #3481, opened 2026-08-14
fix(curation): add exact entity resolution for memory edits

### LINKED ISSUE #3479: [BUG] update_memory silently mis-resolves the entities field on ~5% of edits (text field unaffected) — no error, deterministic, not fixed by retry
## Summary

`update_memory` silently mis-resolves the `entities` field on a small but consistent fraction of edits, while the `text` field is always written correctly. The wrong entity is not random noise — it's an existing, valid entity already in the bank, sometimes a completely unrelated real person, so the result reads as a normal, well-formed record. There is no error, no partial-failure indicator, and `200 OK` is returned on the corrupted writes exactly as on the correct ones.

## Environment

- Package: `hindsight-all` 0.9.0 (self-hosted, pip install, embedded Postgres backend)
- Called via the Hindsight MCP server, from Claude Code
- macOS 25.3.0 (Darwin arm64)

## Repro

1. Have a bank with several hundred existing memories and an entity graph that includes near-duplicate/related names (e.g. `FirstName` / `FirstName LastName`, or multiple people who share a surname).
2. Run a bulk `update_memory` correction across many records — in this case, 657 calls each replacing a generic placeholder entity (`User`) with the correct person's real name in a record's `entities` list (and updating `text` to match).
3. After the batch, re-fetch every edited record (`list_memories` or per-id fetch) and diff the *actual* stored `entities` array against the `entities` you submitted for that call.

## Observed

- **`text` field: 657/657 correct.** No mismatches anywhere.
- **`entities` field: 25/523 entity-bearing edits (~4.8%) came back different from what was submitted**, and none of the 25 are cosmetic — e.g. (names/labels below are placeholders standing in for real values, structure and specificity preserved):
  - Submitted `["topic A", "Person B", "Person C", "Person C Full Name"]` → stored `["topic A", "Person B", "Person C", "Person D Full Name"]` — an existing, unrelated real person (`Person D`, who shares a surname with the intended target) was swapped in on a high-value financial fact.
  - Submitted `["Medication X", "Person C Full Name"]` (a medication name + a person) → stored `["Person E", "Person C Full Name"]` — the medication name silently became a different existing entity, an unrelated person, even though nothing about this edit touched that entity.
  - Submitted `["Person C Full Name", "Account Label 1", "topic"]` → stored `["Person C Full Name", "Account Label 2", "topic"]` — an untouched account-name entity was swapped for a different existing account entity belonging to the same person.
  - An entity *not targeted by the edit at all* (present unchanged in both old and new lists) came back as a different real contact already in the bank.
  - 11 of the 25 simply reverted the `entities` field to the pre-edit value despite `text` updating correctly and the new entities being submitted.
- **Deterministic, not a race condition.** A single isolated sequential retry (zero concurrency) reproduced the exact same wrong output. A batch of 8 parallel retries reproduced the identical wrong output 8/8 times. Same input → same wrong output, every time.
- **Not fixable by resubmission**, for 24 of the 25 — retrying with the same or a cleaned-up entity list changed nothing.
- **One partial mechanism found:** 1 of the 25 had a near-duplicate short/long-form pair in the submitted list (e.g. `"Person C"` and `"Person C Full Name"` together). Dropping the short form and resubmitting the long form alone fixed that one record. This does not explain the other 24 — several of those have no duplicate-form pair anywhere in the submitted list (e.g. the medication and account-label cases above).
- **Clusters by source document, not randomly distributed.** Grouping the 25 mismatches by the source document the memory was originally extracted from: 17 of 25 trace back to just 4 source documents (5, 5, 4, and 3 mismatches respectively), with 8 further isolated singletons across 8 other documents. Fields failing within the same source document fail in unrelated ways (different entities, different substitution targets) — it isn't one bad value propagating, it looks scoped to something about the document/chunk rather than the specific entity string.

## Expected

Either the stored `entities` field matches what was submitted, or the call fails/reports a conflict — not a silent, confidently-wrong substitution returned as a successful write.

## Why this matters more than a normal bug

`text` was correct on every single edit, so a spot-check of a few records looks completely clean. The only way this surfaced was a deliberate full post-hoc diff (re-fetch all 657 edited records, compare `entities` arrays field-by-field against what was sent). Without that audit, a sensitive financial fact would have sat silently attached to the wrong real person (an actual family member of the intended person, already an entity in the bank) indefinitely — discoverable only if someone later asked a `reflect`/graph-scoped question about that wrong person and got an answer partly built from a fact that was never theirs.

## What I haven't verified

- Root cause is unconfirmed — the document-clustering pattern suggests something cache- or index-scoped in the entity-resolution path of `update_memory`'s write, but this is inference from black-box I/O only; I didn't have access to server internals/logs.
- Not tested against the Docker distribution, only the pip (`hindsight-all`) self-hosted install with the embedded Postgres backend — unknown whether this is install-specific.
- Not tested against a fresh/small bank — this bank had ~2,000 pre-existing facts and an entity graph with several genuinely similar/overlapping names, which may or may not be a precondition.

## Data

I have the full input/output diff for all 25 mismatched records (record id, expected `entities`, actual `entities`) and the complete 657-record batch that was submitted. The real values are personal data, so I've kept this report to structurally-accurate placeholders — happy to provide an anonymized version of the diff (ids + expected/actual shape, values replaced with consistent placeholders) if that would help debugging.

Sanderhoff-alt:
## Problem

`update_memory` uses fuzzy entity matching by default. During a manual or bulk correction, the entity name supplied by the caller can be matched to a different existing entity with a similar name. The request still succeeds, but the corrected memory is attached to the wrong entity. This is the behavior reported in #3479.

## Change

Add an optional `entity_resolution_mode` parameter to `update_memory`. When the parameter is omitted or set to `fuzzy`, the existing fuzzy matching behavior is preserved. When it is set to `exact`, the resolver reuses an existing entity only when its name matches case-insensitively. A name with no exact match creates a new entity. Exact mode also disables fuzzy deduplication between entity names in the same request.

Exact mode is intended for user-authored corrections and other operations where the caller already knows which entity name should be updated. Keeping fuzzy mode as the default preserves compatibility for existing callers.

## Scope

The parameter is supported by the HTTP API and MCP tool and is passed through the memory engine and entity resolver. The OpenAPI specification, generated Python, Go, and TypeScript client models, and current and 0.9 documentation have been updated. Regression tests cover similar existing entities and the curation API.

## Verification

The resolver and curation tests pass (43 tests), as do the HTTP curation tests (4 tests), Python Ruff check and format check, Python `ty check`, OpenAPI compatibility validation, TypeScript client build, generated Python model validation, and `git diff --check`.

Full linting could not run because the Control Plane environment is missing `@eslint/js`. The documentation build could not run because its environment is missing `ajv`. Go client root integration tests require an API server on `localhost:8888`; package compilation and generated tests pass without that server. An unrelated existing client coverage check still reports the missing `reflect.apply_all_directives` wrapper field in Python and TypeScript.

Fixes #3479.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -1793,8 +1793,15 @@ class UpdateMemoryRequest(BaseModel):
     )
     entities: list[str] | None = Field(
         default=None,
-        description="Replace the fact's entities. Names are resolved/find-or-created "
-        "the same way retain does; '[]' detaches all entities. Omit to leave unchanged.",
+        description="Replace the fact's entities. By default names use the same fuzzy "
+        "resolution as retain; set entity_resolution_mode to 'exact' to only reuse "
+        "case-insensitively identical names. '[]' detaches all entities. Omit to leave unchanged.",
+    )
+    entity_resolution_mode: Literal["fuzzy", "exact"] = Field(
+        default="fuzzy",
+        description="How to resolve submitted entity names. 'fuzzy' preserves the legacy "
+        "disambiguation behavior; 'exact' only reuses case-insensitively identical names "
+        "and creates unmatched entities.",
     )
     state: str | None = Field(
         default=None,
@@ -4450,6 +4457,7 @@ def _register_routes(app: FastAPI):
                 occurred_end=occurred_end,
                 new_fact_type=request.fact_type,
                 entities=request.entities,
+                entity_resolution_mode=request.entity_resolution_mode,
                 state=request.state,
                 reason=request.reason,
                 request_context=request_context,
--- a/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
+++ b/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
@@ -15,7 +15,7 @@ from collections.abc import Iterator
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from difflib import SequenceMatcher
-from typing import Any, Final, cast
+from typing import Any, Final, Literal, cast
 
 from .db_utils import acquire_with_retry
 from .memory_engine import fq_table
@@ -414,6 +414,7 @@ class EntityResolver:
         unit_event_date,
         conn=None,
         entity_labels: list | None = None,
+        resolution_mode: Literal["fuzzy", "exact"] = "fuzzy",
     ) -> list[ResolvedEntity]:
         """
         Resolve multiple entities in batch (MUCH faster than sequential).
@@ -427,6 +428,9 @@ class EntityResolver:
             context: Context where entities appear
             unit_event_date: When this unit was created
             conn: Optional connection to use (if None, acquires from pool)
+            resolution_mode: ``fuzzy`` preserves the normal retain behavior;
+                ``exact`` reuses only case-insensitively identical names and
+                creates a new entity for every unmatched name.
 
         Returns:
             Resolved entity identities (id + stored canonical name) in the same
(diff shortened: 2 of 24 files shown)


Sanderhoff-alt:
Closing this PR because its entity-resolution fix has been superseded by #3576, which covers the same update-memory use case and extends the behavior to explicit entities in retain, mixed extracted/literal entities, client wrappers, and broader regression coverage.

PR #3482, opened 2026-08-14
fix(consolidation): preserve temporal fields across dedup merges

### LINKED ISSUE #3477: bug(consolidation): dated source facts can produce observations with null occurred dates
## Bug Description

Derived observations can have `occurred_start = null` and `occurred_end = null` even when every source world/experience fact has non-null occurred dates. The observation still has a populated `date`/`mentioned_at`, which makes the result appear dated while temporal filtering and date-range consumers lose the source event interval.

This is distinct from the already-merged #501 behavior for source memories with no temporal data. In this case, the source facts are dated, but the derived observation does not inherit those dates.

## Steps to Reproduce

1. Run Hindsight with observations/consolidation enabled and create a temporary bank.
2. Retain several related facts using explicit event dates or `occurred_start`/`occurred_end` values, for example:
   - source A: `occurred_start=2026-08-13T00:00:00Z`, `occurred_end=2026-08-13T00:00:00Z`
   - source B: `occurred_start=2026-08-13T00:00:00Z`, `occurred_end=2026-08-13T00:00:00Z`
3. Wait for the automatic consolidation operation to complete.
4. List or recall observations and inspect the observation plus its source memory IDs.
5. Compare the observation temporal fields with the source facts.

## Expected Behavior

For an observation derived only from dated source facts, `occurred_start` and `occurred_end` should be populated from the source temporal bounds, consistent with the documented meaning of these fields and the existing temporal handling in #501. For multiple sources, the expected interval should be the minimum source start and maximum source end, or another documented deterministic rule.

If source dates cannot be established, the observation should remain explicitly undated with no misleading substitute, but dated source evidence should not be discarded.

## Actual Behavior

In a self-hosted Hindsight deployment, all raw source facts had non-null `occurred_start` and `occurred_end`, but two derived observations linked to those sources returned:

```json
{
  "fact_type": "observation",
  "occurred_start": null,
  "occurred_end": null,
  "date": "2026-08-13T00:00:00Z",
  "mentioned_at": "2026-08-13T00:00:00Z",
  "source_memory_ids": ["<dated source memory id>"]
}
```

The populated `date`/`mentioned_at` appears to represent observation/consolidation time, not the event interval, and does not fix the missing occurred fields.

The affected observations were later regenerated after their dated source facts were edited to include a leading date marker, but this is only a client-side workaround and does not establish that consolidation correctly propagates source dates in the general case.

## Version

Hosted/self-managed deployment version was not exposed by the MCP response. The behavior was observed on 2026-08-13/14. Please advise which release or commit contains the relevant consolidation implementation and whether this is already fixed on `main`.

Relevant upstream context:
- PR #501: https://github.com/vectorize-io/hindsight/pull/501
- Consolidation implementation: `hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py`

## LLM Provider

MiniMax-compatible provider in the deployment; exact provider configuration can be supplied privately if needed. The issue is about persisted temporal fields after consolidation, not the provider API response format.

## Additional Notes

- Raw world/experience records were verified before and after consolidation.
- The issue is observable through the public memory fields and source-memory linkage; no credentials or private memory content are included here.
- A regression test should cover: dated source facts -> consolidation -> derived observation has non-null occurred bounds matching the source range; undated source facts should retain the behavior fixed by #501.

Sanderhoff-alt:
## Summary

Issue #3477 reports that dated source facts can produce observations with null `occurred_start` and `occurred_end`. The normal CREATE path already aggregates source dates, but dedup folds bypassed that writer and updated observations did not pass `event_date` through the PostgreSQL SQL path. As a result, merging observations could silently discard part of their temporal information.

This change makes all observation merge paths use the same null-safe rules: `event_date` and `occurred_start` keep the earliest non-null value, while `occurred_end` and `mentioned_at` keep the latest non-null value. CREATE dedup carries source-derived fields into the survivor, UPDATE dedup merges both rows before deleting the rewritten row, and external memory stores follow the same rules.

As a related fix, the ordinary observation UPDATE path now propagates `event_date` from the source aggregation instead of leaving it unchanged. Its PostgreSQL SQL update and external-store upsert now apply the same temporal merge rules, so updates cannot lose an earlier event date while extending an observation.

## Verification

The deterministic regression tests cover CREATE and UPDATE dedup SQL, the ordinary UPDATE parameter contract, external-store folds, and min/max/null behavior. 45 dedup tests, 15 UPDATE/integrity tests, and 7 source aggregation tests pass. The existing real-LLM temporal consolidation integration test was also run with the configured test runtime; this PR does not add a new LLM integration test.

Fixes #3477

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -237,6 +237,16 @@ class _DedupOutcome:
     best_text: str = ""
 
 
+@dataclass(frozen=True)
+class _TemporalFields:
+    """The four temporal columns that must survive an observation merge."""
+
+    event_date: datetime | None
+    occurred_start: datetime | None
+    occurred_end: datetime | None
+    mentioned_at: datetime | None
+
+
 async def _dedup_adjudicate(
     pool: DatabaseBackend,
     memory_engine: "MemoryEngine",
@@ -322,6 +332,7 @@ async def _dedup_reconcile_create(
     create_text: str,
     create_source_ids: list[uuid.UUID],
     tags: list[str] | None,
+    source_temporal_fields: _TemporalFields,
     txn=None,
 ) -> str | None:
     """Semantic dedup for a single CREATE (create-time, focused 1-by-1).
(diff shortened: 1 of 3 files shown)


Sanderhoff-alt:
Closing this PR because #3500 provides a better implementation.

PR #3483, opened 2026-08-14
fix(coding-agents): keep knowledge pages current

### LINKED ISSUE #3475: Knowledge Pages restate retired facts as current — a reversed setting, and a dependency abandoned a month ago that four pages named as current for 26 hours (superseded ≠ incorrect, so neither invalidation nor the #3248 supersession rule applies)
## Summary

Knowledge Pages state facts that were true when written and have since been retired, in the present tense, with no indication that they expired. Two instances from the same bank are below: a configuration setting that was reversed the following day, and a memory system the project migrated away from a month ago which four of five pages still named as current.

Neither existing mechanism covers this:

- **`state=invalidated`** (#1391, #2696) retires a fact judged *wrong*. This fact was never wrong — it was true on 2026-07-16 and stopped being true on 2026-07-17. There is nothing to invalidate.
- **The `observations_mission` supersession rule** (#3248, shipped in coding-agents 0.1.1) fires when *"a new fact contradicts or supersedes an existing observation"*. These two facts do not contradict each other: one enables a setting, the other removes it, and both were correct when written. Read as standalone historical statements they are perfectly consistent.

The missing step is recognising that two facts are *about the same subject* — the same config key, the same dependency — and that the later one retires the earlier. That is a stronger relation than contradiction, and nothing currently establishes it.

## Environment

| | |
|---|---|
| hindsight-embed / hindsight-api | 0.9.0 (self-hosted, pg0, port 9077) |
| @vectorize-io/hindsight-coding-agents | 0.3.3 (staged runtime and bun global both) |
| Harness | Claude Code, per-repo bank `claude-code::<gitProject>` |
| Pages | `Conventions and patterns`, `Component map`, `Core concepts`, `Key decisions and rationale`, `Initiatives and enhancements` — all auto-seeded, none hand-authored |
| Trigger config | `{"mode":"full","refresh_after_consolidation":true}` on every page; `auto_refresh` unset — refresh is consolidation-driven, never clock-driven |

Two instances follow. The first is a reversed decision; the second is an abandoned dependency. Same shape: a fact that was true, stopped being true, and is still stated in the present tense.

## Instance 1 — a reversed decision, stated as a current convention

From the `Conventions and patterns` page, under *Change-management Practices*:

> **Disable automatic behaviours that hinder review** – "Autocommit was disabled in chezmoi (`autocommit=false`) to enable hunk-level review."

Presented as a current convention of the project, alongside items that genuinely are current.

## What the repository actually contains

```
06190bd  2026-07-16  chezmoi: disable autocommit, capture drifted files, revert macroscope PATH
6b6f254  2026-07-17  Remove dead autocommit=false, enable hunk agent_notes + transparent_bg, automate hunk review
```

`06190bd` — the decision the page reports:

```diff
-    autocommit = true
+    # autocommit is disabled so changes can be reviewed with hunk before
+    autocommit = false
```

`6b6f254`, one day later — the reversal:

```diff
-    # autocommit is disabled so changes can be reviewed with hunk before
-    autocommit = false
+    # autopush implies autocommit (per chezmoi docs: "autoPush implies
```

The live config today has `autopush = true` and no `autocommit` key at all, so autocommit is **on** — the opposite of what the page states. Both commits are in the bank's ingested history; the page surfaced the older one.

## Instance 2 — an abandoned dependency, named as the project's memory system for 26 hours

This repository migrated from `agentmemory` to Hindsight in mid-July 2026. `agentmemory` has not been installed, configured or referenced by any live file since. Four of the five pages nevertheless described it as the project's memory system, in the present tense, and the fifth described the migration as in progress.

`GET /v1/default/banks/{bank}/mental-models/{id}/history` stores each version's prior content, which dates the window exactly. Counting lines matching `agentmemory` in the content each refresh *replaced*:

| Page | agentmemory lines | Present from | Removed at |
|---|---|---|---|
| `Component map` | 8 → 6 | 2026-08-12T18:23Z | 2026-08-13T20:24Z |
| `Core concepts` | 3 → 2 | 2026-08-12T18:23Z | 2026-08-13T20:24Z |
| `Initiatives and enhancements` | 3 → 1 | 2026-08-13T09:49Z | 2026-08-13T11:03Z |
| `Key decisions and rationale` | 1 | 2026-08-12T15:26Z | 2026-08-12T18:23Z |

The `Component map` and `Core concepts` figures are the ones to look at: **26 hours**, across at least three intervening refreshes that each rewrote the page and each preserved the claim.

### What eventually removed it, and why that is the concerning part

Not a staleness check. The bank had been seeded from `git log` subject lines only; on 2026-08-13 I began ingesting **full commit diffs** (`gitDiffDocs` climbing toward a 300-document target). The resulting observations triggered consolidation, consolidation triggered `refresh_after_consolidation`, and all six models were rewritten between 20:23Z and 20:35Z. Every `agentmemory` mention disappeared in that burst.

So the page self-corrected only because unrelated work happened to push new observations through the pipeline. Three properties follow, and I think they are the substance of this report:

1. **Healing is a side effect of ingestion, not a property of the page.** With `refresh_after_consolidation: true` and no `auto_refresh` schedule, a page on a repository that goes quiet is frozen at whatever it last said. Staleness scales with *quiet*, not with age — and a dotfiles repo, a finished service, or an archived library are all exactly the repositories that go quiet.
2. **Repeated refreshes do not converge on truth.** The claim survived three rewrites of the same page. Each refresh re-derived the same present-tense statement from the same observations, because nothing in those observations says the fact expired.
3. **Seed depth silently sets page quality.** A bank seeded from commit subjects can only produce subject-line-grade pages, and "this project uses agentmemory" is a defensible summary of a July history. There is no signal to the reader that a page is subject-line-grade rather than diff-grade.

Neither instance is an *incorrect* fact. `agentmemory` was the memory system, and `autocommit=false` was set. Both entered the bank as true statements and neither was ever contradicted — the repository simply moved on, and no mechanism exists to notice that.

## Why this is worse than a stale fact in recall

A stale fact returned by `recall` arrives as one result among several, with a timestamp, and can be weighed. A knowledge page is different in two ways:

1. **It is injected at session start**, before the agent knows what it is working on, and it is framed as the project's curated conventions rather than as retrieved evidence.
2. **It is prose, not a fact list.** The reversal is not present as a competing entry the reader can compare against — it is simply absent. There is nothing on the page to weigh the claim against.

In this instance an agent read the page, noticed the claim conflicted with the repo's own `AGENTS.md`, and discarded it. That is luck: the contradiction happened to be documented elsewhere in the same repo. A convention with no second source would have been followed.

This matters more given the direction discussed in Slack — that the knowledge pages are the answer to auto-maintained agent instruction files, rather than writing back into `CLAUDE.md`. If pages are to serve that role, a page stating a reversed decision in the present tense is not a cosmetic defect: it is the failure mode that makes the whole approach unsafe, since the reader has no way to tell a current convention from a retired one.

## What does not already cover it

- **#3279** gave the reflect/refresh agent a notion of "now", which fixed *"is this event past or upcoming"*. This is the neighbouring gap: not when a fact happened, but whether a later fact retired it.
- **Refresh frequency is not the variable.** Instance 2 survived three consecutive refreshes of the same page. Each re-derived the same present-tense statement from the same observations; the derivation is stable because the underlying evidence is.
- **Nor is seed depth, on its own.** Deeper ingestion is what eventually displaced instance 2, but only incidentally — it added observations that outweighed the old ones, rather than marking the old ones retired. Instance 1's reversal was in the bank the whole time, in a diff that removed the exact lines an earlier diff added, and the page still reported only the earlier one.
- **The two instances differ in subject but not in shape.** A config key set then removed, and a dependency adopted then abandoned, are both "the project moved on" — and in both cases the later evidence is present in the bank and contradicts nothing.

## Reproduction

Instance 1 needs a repo whose history contains a decision and its reversal, so it is not reproducible on demand. Instance 2 is closer to a recipe:

1. Take a repo that migrated off a dependency at least a few weeks ago, with both the adoption and the removal in its history.
2. Seed a bank from `git log` subject lines only — the default depth — and let the pages generate.
3. `Component map` and `Core concepts` describe the abandoned dependency as current. The removal is absent, not contradicted.
4. `GET /mental-models/{id}/history` then dates the window precisely: compare `agentmemory`-matching lines in each version's `previous_content` against the commit that removed it.

Happy to run any diagnostic against this bank, or to share any page body, the version history, or the two commits in full.

Sanderhoff-alt:
## Summary

Knowledge pages can keep presenting facts that were true in the past but were later replaced, removed, or superseded. This is especially risky for coding agents because the pages are injected as current repository guidance.

This change addresses the issue by making coding-agent knowledge page refreshes explicitly use full mode and by adding a shared current-state policy to the page queries. The policy asks synthesis to prefer later evidence and label historical or superseded facts instead of presenting them as current.

## Changes

- Set `mode: "full"` for the five seeded knowledge pages and dynamically created initiative pages.
- Apply the same current-state policy to all seeded page queries and initiative page queries.
- Extend page seeding tests to verify the refresh mode and policy are sent to the API.

This is a focused mitigation. It does not introduce fact validity windows or a new supersession data model.

## Verification

- `npm test` in `hindsight-integrations/coding-agents`: 499 tests passed.
- `npm run build` in `hindsight-integrations/coding-agents`: passed.
- `git diff --check`: passed.

Fixes #3475

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/hindsight.pages.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/hindsight.pages.test.ts
@@ -175,7 +175,9 @@ describe("HindsightClient.seedPages", () => {
         /^knowledge:(feature-work|decision|convention|component|concept)$/
       );
       expect(post.body.max_tokens).toBe(PAGE_MAX_TOKENS);
+      expect(post.body.trigger.mode).toBe("full");
       expect(post.body.trigger.refresh_after_consolidation).toBe(true);
+      expect(post.body.source_query).toContain("Current-state policy:");
       expect(post.body.parent_id).toBeUndefined(); // seeded at the tree root
     }
     // Nothing on the mental-models surface.
@@ -346,6 +348,8 @@ describe("HindsightClient.captureInitiative", () => {
     expect(pagePost.body.name).toBe("Retry backoff for the uploader");
     expect(pagePost.body.parent_id).toBe("folder-abc");
     expect(pagePost.body.tags).toEqual(["knowledge:feature-work"]);
+    expect(pagePost.body.trigger.mode).toBe("full");
+    expect(pagePost.body.source_query).toContain("Current-state policy:");
 
     // Marker retain POST to /memories: the ONLY tag is relatedPageId, pointing at the REAL
     // server-assigned page id ("pg").
--- a/hindsight-integrations/coding-agents/src/core/hindsight.ts
+++ b/hindsight-integrations/coding-agents/src/core/hindsight.ts
@@ -11,6 +11,7 @@ import {
   PAGE_MAX_TOKENS,
   PAGE_TRIGGER,
   PAGES,
+  withCurrentStatePolicy,
 } from "./missions";
 import { pool, semverGte, sleep } from "./util";
 import type { RetainStamp } from "./retain-stamp";
@@ -570,13 +571,12 @@ export class HindsightClient {
       const folderId = await this.ensureFolder("Initiatives");
       const r = await this.req("POST", this.bankUrl("/knowledge-base/pages"), {
         name: args.title,
-        source_query: `Summarize the "${args.title}" initiative: what is being built or changed and why, and its current state — drawn from the project's memory.`,
+        source_query: withCurrentStatePolicy(
+          `Summarize the "${args.title}" initiative: what is being built or changed and why, and its current state — drawn from the project's memory.`
+        ),
         parent_id: folderId,
         tags: ["knowledge:feature-work"],
-        trigger: {
-          fact_types: ["world", "experience", "observation"],
-          refresh_after_consolidation: true,
-        },
+        trigger: PAGE_TRIGGER,
       });
       try {
         const j = (await r.json()) as { page_id?: string; id?: string };
--- a/hindsight-integrations/coding-agents/src/core/missions.ts
+++ b/hindsight-integrations/coding-agents/src/core/missions.ts
@@ -192,46 +192,65 @@ export interface KnowledgePage {
   tags: string[];
 }
 
+const CURRENT_STATE_POLICY = [
+  "Current-state policy:",
+  "- Describe the repository's current state, not only its history.",
+  "- Prefer later relevant evidence over earlier evidence when they concern the same subject.",
+  "- When later evidence updates, removes, replaces, or supersedes an earlier fact, " +
+    "do not present the earlier fact as current.",
+  "- Mention superseded, removed, or historical facts only with an explicit status label.",
+  "- If the evidence does not establish the current state, say that it is uncertain.",
+].join("\n");
+
+export function withCurrentStatePolicy(query: string): string {
+  return `${query}\n\n${CURRENT_STATE_POLICY}`;
+}
+
 export const PAGES: KnowledgePage[] = [
   {
     name: "Component map",
-    source_query:
+    source_query: withCurrentStatePolicy(
       "From this project's commit history and past discussions, what are the main " +
-      "components/modules/subsystems, what is each responsible for, and how do they relate to or " +
-      "depend on one another? Describe the structure and responsibilities.",
+        "components/modules/subsystems, what is each responsible for, and how do they relate to or " +
+        "depend on one another? Describe the structure and responsibilities."
+    ),
     tags: ["knowledge:component"],
   },
   {
     name: "Core concepts",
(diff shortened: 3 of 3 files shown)


Sanderhoff-alt:
Closing this PR for now. The problem described in #3475 needs a more nuanced design than adding broader current-state prompt instructions. We should revisit the underlying freshness, retrieval, and refresh semantics before settling on an implementation.

PR #3484, opened 2026-08-14
fix(coding-agents): exclude external facts from knowledge pages

### LINKED ISSUE #3476: Knowledge Pages present facts about a discussed dependency as decisions of the project — document attribution records where a fact arrived, never what it is about
## Summary

The `Key decisions and rationale` page for a dotfiles repository lists, among that project's own significant technical decisions, two decisions belonging to Hindsight itself — worker configuration and an environment-variable rename — citing an upstream commit SHA that does not exist in the repository.

The mechanism is not cross-bank leakage. The bank is correct: these facts were extracted from conversations and documents that genuinely took place *in* that repository, because that repository is where Hindsight is configured, operated and discussed. The error is one level up: the page builder treats "a fact in this project's bank" as "a fact about this project", and a repository that talks about its dependencies has no way to say otherwise.

This is a sibling of the retirement gap in #3475, with the opposite cause. There, both facts were about the project and the later one retired the earlier. Here, the fact is current, accurate, and about somebody else's codebase. Ordering cannot fix it, and — unlike #3475 — **further ingestion cannot either**: no volume of additional observations teaches the builder that a fact is about a dependency.

## Environment

| | |
|---|---|
| hindsight-embed / hindsight-api | 0.9.0 (self-hosted, pg0, port 9077) |
| @vectorize-io/hindsight-coding-agents | 0.3.3 (staged runtime and bun global both) |
| Harness | Claude Code, per-repo bank `claude-code::<gitProject>` |
| Bank | `claude-code::dotfiles` — a chezmoi-managed dotfiles repository |
| Page | `Key decisions and rationale`, auto-seeded, refreshed 2026-08-13T20:35Z |

## What the page says

Verbatim rows from the page's table, headed *"Key Technical Decisions & Their Rationale"*:

> | **Helm/Kubernetes workers** (K8s manifests) | "configure workers to use the correct **API port override** to avoid service-discovery failures." – *bf0ab6ca* |

> | **Worker slot env-var renaming** (runtime) | "Renamed per-type worker slot settings and environment variables from the legacy `*_MAX_SLOTS` names to **`HINDSIGHT_API_WORKER_<TYPE>_RESERVED_SLOTS`**, preserving the old names with a deprecation warning…" |

Both are presented in the same table, in the same voice, as genuine decisions of this repository such as "commit directly to `main` rather than using stacked branches" and "keep secrets in a 1Password → keychain pipeline".

## What the repository actually contains

1. `git cat-file -t bf0ab6ca` → `fatal: Not a valid object name`. The cited commit is not among the repository's 988 commits on any branch.
2. `find . -iname '*helm*' -o -iname '*.yaml' -path '*k8s*'` → nothing outside `.git`. There are no Helm charts, no Kubernetes manifests, no deployments.
3. `HINDSIGHT_API_WORKER_<TYPE>_RESERVED_SLOTS` is a Hindsight API server setting. This repository *runs* `hindsight-embed` as a local daemon on port 9077 and configures it through `~/.hindsight/`. It does not develop the API server, define its worker slots, or deploy it anywhere.

The repository is a chezmoi dotfiles tree for two machine profiles. It has no services, no deployments, and no runtime workers of its own.

## Why it happens

This repository is where Hindsight is documented and debugged: several `docs/hindsight-*.md` references, upstream issue drafts, and long sessions spent reading upstream source. Facts extracted from that material are *about* upstream, but they land in this repository's bank because that is where the conversation happened — and the page builder's question is "what are this project's key decisions?", answered over everything the bank holds.

Any repository that discusses its dependencies in depth will hit this: a repo whose issues are mostly about a framework, a service whose sessions are mostly about its database, an integration repo whose whole purpose is another product.

## Why the existing attribution work does not cover it

#3285 and #3418 add configurable document attribution — templated tags and metadata such as `project:{gitProject}`, `harness:{harness}`, `workspace:{cwd}`, applied consistently across transcripts, imported conversations, git history, diffs and survey findings.

That machinery answers **where a fact arrived from**. Every fact above is already correctly tagged `project:dotfiles`, because it did arrive from a session in that project. What is missing is **what the fact is about**, which no arrival-time tag can express. Stamping the container more consistently cannot separate two facts that arrived in the same container and differ only in subject.

Similarly #3343 (route repository surveys explicitly) addresses which bank a survey lands in, not which subject a fact within it concerns.

## Why this is worse than the retirement gap

- **It cannot self-heal.** The staleness in #3475 was eventually corrected when deeper ingestion produced new observations. Here, more observations about upstream produce *more* upstream decisions on the page, not fewer.
- **It is invisible to a reader who does not know both codebases.** A retired fact usually contradicts something the reader can check. An accurate fact about a dependency reads as authoritative — it is well-written, specific, and cites a SHA. Only someone who tries `git show bf0ab6ca` discovers it is not this project's.
- **It scales with how well the memory works.** The more thoroughly a project's sessions are captured, the more of its dependency discussion becomes "its decisions".

## Reproduction

1. Use a repository whose sessions frequently discuss an external dependency's internals — its source, its configuration, its upstream issues.
2. Let the bank ingest those sessions normally and let `Key decisions and rationale` generate.
3. Decisions belonging to the dependency appear in the table as decisions of the repository, sometimes citing commit SHAs from the dependency's history.
4. `git cat-file -t <sha>` in the repository confirms the SHA does not resolve there.

Happy to share the full page body, the version history, or run any diagnostic against this bank.

Sanderhoff-alt:
## Summary

Knowledge Pages can currently summarize a dependency's implementation or decisions as if they belonged to the repository being worked on. This happens because document attribution records where a fact was captured, but does not identify what project the fact is about. This PR adds an explicit external-knowledge classification and prevents those facts from entering the repository's standard Knowledge Pages.

Fixes #3476

## What changed

The coding-agents plugin adds `knowledge:external` to the entity-label vocabulary. The extraction guidance says to use it only when the fact is about another project's own implementation, configuration, or decision. A decision about how the current repository uses or integrates a dependency remains normal project knowledge. The external label is explicitly exclusive with the other `knowledge:*` categories.

Each seeded Knowledge Page now refreshes from its normal category tag and excludes `knowledge:external` with a deterministic compound tag filter. The exclusion is therefore enforced by the refresh query instead of relying on wording in the page's natural-language source query.

Existing pages are upgraded in place. On startup the plugin reads the Knowledge Base tree, matches the fixed root pages by name, and patches a page only when its trigger is missing or different. Page IDs, existing content, and synthesized history are preserved. New pages receive the trigger when they are created, and repeated startup runs are idempotent.

The Knowledge Base API now exposes the page trigger in tree responses and accepts an optional `trigger` field in node PATCH requests. The route forwards that field to the existing backing mental-model update logic. The field is optional, so older clients and older page behavior remain compatible.

The CLI now exposes the same capability as `knowledge-base update BANK NODE --trigger <JSON>`. The argument accepts the complete trigger object, including nested `tag_groups`, and is included in the CLI OpenAPI coverage check.

The Rust SDK does not commit generated source files. Its `build.rs` generates the models from `hindsight-docs/static/openapi.json` at build time, so `KnowledgeNode.trigger`, `UpdateNodeRequest.trigger`, and nested trigger tag groups are available after rebuilding. The service's native MCP has no Knowledge Base CRUD operation, so it has no corresponding parameter to add. The coding-agents MCP also keeps page CRUD internal to the plugin.

The OpenAPI specification and generated Python, TypeScript, and Go clients are updated together with the control-plane types and wrappers.

## Compatibility

The deprecated per-harness plugins continue to use their existing mental-model endpoints and are not changed by this PR. They remain operational, but they do not automatically gain the new external classification. The current unified plugin upgrades pages created by earlier versions of that same plugin when the server exposes the trigger field. Historical facts are not reclassified automatically; newly extracted facts use the new label.

When an older server does not expose the Knowledge Base surface, the plugin treats 404, 405, and 501 responses as an unavailable optional capability. It skips page creation or upgrade and continues with bank configuration and memory operations. Older servers that return a tree without `trigger` are not sent a trigger PATCH, so their existing page behavior is preserved.

## Verification

The coding-agents test suite passed with 504 tests, including 31 Knowledge Page tests. The CLI trigger parsing test passed, the CLI compiled successfully, and the OpenAPI coverage check reported all 89 operations and 115 request parameters covered. Client and documentation generation, repository lint hooks, and whitespace checks passed. The server Knowledge Base integration tests were not run because this temporary worktree does not include `pg0-embedded`, so its test database cannot start.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2325,7 +2325,7 @@ class MentalModelListResponse(BaseModel):
 class KnowledgeNode(BaseModel):
     """A node in the knowledge-base tree — a folder or a page.
 
-    Pages carry ``description``/``tags`` from their backing mental model. The
+    Pages carry ``description``/``tags``/``trigger`` from their backing mental model. The
     knowledge base is client-managed (CRUD); ``managed`` lets a client tag a node
     as system-owned vs. hand-authored.
     """
@@ -2338,6 +2338,10 @@ class KnowledgeNode(BaseModel):
     managed: bool = Field(default=False, description="Client-set flag: true = system-owned, false = hand-authored.")
     description: str | None = Field(default=None, description="Page source query (the page's `description`).")
     tags: list[str] = FieldWithDefault(list)
+    trigger: MentalModelTrigger | None = Field(
+        default=None,
+        description="Page refresh trigger, including optional compound tag filters.",
+    )
     timestamp: str | None = Field(default=None, description="Last refresh (page) or last update (folder).")
     is_stale: bool | None = Field(
         default=None,
@@ -2380,10 +2384,15 @@ class UpdateNodeRequest(BaseModel):
     name: str | None = None
     parent_id: str | None = None
     # Page-only options (updated on the backing mental model). Changing
-    # source_query schedules an async refresh so the page rebuilds.
+    # source_query schedules an async refresh so the page rebuilds; trigger changes only
+    # change the refresh scope and do not rebuild content by themselves.
     source_query: str | None = None
     tags: list[str] | None = None
     max_tokens: int | None = None
+    trigger: MentalModelTrigger | None = Field(
+        default=None,
+        description="Page refresh trigger. Omit to leave it unchanged; explicit null is rejected.",
+    )
 
 
 class CreateKnowledgePageResponse(BaseModel):
@@ -2450,6 +2459,7 @@ def _knowledge_node_model(node: dict[str, Any]) -> KnowledgeNode:
         managed=bool(node.get("managed")),
         description=node.get("source_query") if is_page else None,
         tags=list(node.get("tags") or []) if is_page else [],
+        trigger=node.get("trigger") if is_page else None,
         timestamp=(node.get("last_refreshed_at") if is_page else node.get("updated_at")),
         is_stale=node.get("is_stale") if is_page else None,
     )
@@ -5787,7 +5797,7 @@ def _register_routes(app: FastAPI):
         response_model=KnowledgeNode,
         summary="Rename/move a knowledge-base node or update a page's options",
         description="Rename a node (set `name`), move it under another folder (set `parent_id`, null "
-        "for the root), and/or update a page's options (`source_query`, `tags`, `max_tokens`). "
+        "for the root), and/or update a page's options (`source_query`, `tags`, `max_tokens`, `trigger`). "
         "Changing `source_query` schedules an async refresh so the page rebuilds against the new question.",
         operation_id="update_knowledge_node",
         tags=["Knowledge Base"],
@@ -5802,6 +5812,11 @@ def _register_routes(app: FastAPI):
         try:
             updated: dict[str, Any] | None = None
             did_change = False
+            if "trigger" in body.model_fields_set and body.trigger is None:
+                raise HTTPException(
+                    status_code=400,
+                    detail="trigger cannot be null; omit it to leave the trigger unchanged",
+                )
             if body.name is not None:
                 did_change = True
                 updated = await app.state.memory.rename_knowledge_node(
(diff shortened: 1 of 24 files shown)


Sanderhoff-alt:
Closing as superseded by #3571, which has been merged and fixes #3476 by scoping each seeded Knowledge Page query to the repository it is about. That approach addresses the issue without introducing a new `knowledge:external` classification or expanding the Knowledge Base trigger API and generated clients. Thanks!

PR #3492, opened 2026-08-14
feat(cursor): add prompt-aware recall and resilient transcript handling

voarsh2:
- Add automatic `beforeSubmitPrompt` recall for every substantive user prompt.
- Build prompt-aware recall queries from recent transcript turns, with configurable roles, tool markers, query length, score floors, and timeout.
- Refresh the workspace rules-file context after successful prompt recall while returning Cursor’s documented `{"continue": true}` response.
- Improve session-start recall handling with `CURSOR_PROJECT_DIR`, UTF-8 BOM-safe hook input, configurable API timeout, and the documented `additional_context` field.
- Fix transcript retention for Cursor 3.x role-nested JSONL payloads and environment-provided transcript/session paths.
- Keep retained content focused on conversation by removing injected memories, `<external_links>` previews, internal reasoning, and tool traffic by default.
- Preserve structured tool-call data only when explicitly enabled, while retaining channel-message text and avoiding Hindsight feedback loops.
- Ensure repeated retains receive distinct conversation-scoped document IDs and support configurable retained roles.
- Make installation portable across platforms with `py -3` on Windows, `python3` elsewhere, plus `--python-command` and `--hook-timeout` overrides.
- Add hook timeouts of 15s for session/stop hooks and 45s for prompt recall, while preserving stop-hook timing when overrides are used.
- Expand unit, hook, CLI, rules-file, transcript, configuration, and integration coverage.
- Synchronize the package README, integration docs, catalog metadata, Cursor skill guidance, and SDK reference documentation.

## Note

Cursor uses `additional_context` for `sessionStart`; `beforeSubmitPrompt` only supports `continue`, so prompt recall refreshes the workspace rules file instead. The removed `additionalContext` references reflect the previous Cursor implementation and are not emitted by this branch.

## Test plan

- [x] `py -3 -m pytest tests -q` — 101 passed, 4 skipped
- [x] Covered prompt recall, multi-turn transcript context, score filtering, and timeout configuration
- [x] Covered Cursor 3.x transcript formats, UTF-8 content, BOM input, environment transcript paths, and retention cleanup
- [x] Covered Windows/non-Windows launcher rendering and explicit launcher/timeout overrides
- [x] Covered rules-file generation, `.gitignore` handling, session context delivery, and end-to-end hook behavior
- [x] Ran Python compilation checks


nicoloboschi:
Checking in on this one — it's been in draft for 10 days and `verify-generated-files` is currently failing.

The failing check is usually a generated-docs drift: this PR edits `hindsight-docs/docs-integrations/cursor.md` and `skills/hindsight-docs/references/sdks/integrations/cursor.md`, which are produced from source rather than hand-edited. Running `./scripts/hooks/lint.sh` first, then the docs sync, should clear it.

The bigger issue is scope. At 25 files this bundles several independent changes:

1. `beforeSubmitPrompt` recall + prompt-aware query building (the actual feature)
2. Cursor 3.x role-nested JSONL transcript handling (a bug fix, and one we've hit before)
3. retained-content filtering — dropping injected memories, `<external_links>`, reasoning, tool traffic
4. Windows/portability of the installer (`py -3`, `--python-command`)

Any one of those is reviewable on its own; together they're hard to evaluate or to revert independently. Could you split them, starting with (2)? The transcript fix stands alone, is the lowest-risk, and unblocks users on Cursor 3.x immediately.

Leaving this open — just flagging that it needs the CI fix and ideally a split before it can move.

hindsight-integrations/cursor/scripts/retain.py
```diff
@@ -227,19 +237,18 @@ def _dbg(*a):
...
-    # so back-to-back retains in the same wall-clock millisecond stay distinct
-    # via session_id.
-    document_id = f"{session_id}-{int(time.time() * 1000)}"
+    # Full-session retains resend the cumulative transcript, so reuse the
+    # session ID and let Hindsight update one document. Chunked retains need
+    # distinct IDs because each request represents a separate transcript window.
+    if retain_mode == "chunked" and retain_every_n > 1:
+        document_id = f"{session_id}-{int(time.time() * 1000)}"
+    else:
+        document_id = session_id
```
Sanderhoff-alt, line 246: This reverts a deliberate fix. On `main` every retain gets its own `document_id` precisely because reusing `session_id` in full-session mode lost earlier turns — the comment being removed here describes it as the "5-turn -> 1 topic" failure mode. Re-ingesting an existing `document_id` deletes that document's existing memories and re-extracts from the new content (see `hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py:3841`), so with the default `full-session` mode each retain discards what the previous one stored and relies on re-extraction of an ever-longer transcript reproducing it.

Two related points: `tests/test_daemon.py::TestDocumentIdIsUniquePerRetain`, which pinned the previous behaviour, is deleted in the same diff, and the PR description states the opposite ("Ensure repeated retains receive distinct conversation-scoped document IDs") while the code makes them identical for the default mode. If the upsert is intentional, please say why the earlier failure mode no longer applies.

hindsight-integrations/cursor/scripts/recall.py
```diff
@@ -0,0 +1,276 @@
...
+    json.dump({"continue": True, "additional_context": context_message}, sys.stdout)
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    except Exception as error:
+        print(f"[Hindsight] Unexpected error in recall: {error}", file=sys.stderr)
+        try:
+            sys.exit(2 if load_config().get("debug") else 0)
```
Sanderhoff-alt, line 274: Cursor treats exit code 2 from a command hook as a block, and for `beforeSubmitPrompt` that means the user's prompt is not submitted. Since `debug: true` is the documented troubleshooting switch, any unexpected error in this hook would silently stop prompts from being sent while debugging. A memory hook should never be able to block submission — exit 0 here and keep the diagnostics on stderr.

hindsight-integrations/cursor/README.md
```diff
@@ -67,27 +68,28 @@ The plugin uses two complementary mechanisms:
...
+- **Native prompt recall is preferred.** `beforeSubmitPrompt` emits `additional_context` on Cursor builds that support it.
+- **The rules file is regenerated after successful prompt recalls and at the top of every `sessionStart`.** This is a compatibility snapshot, not a guaranteed live context channel.
 - **The rules file is auto-`.gitignore`'d** in git workspaces. It's safe to delete by hand; it'll be regenerated.
-- **`additionalContext` is still emitted to stdout** for forward-compat. If Cursor restores the native channel, the same plugin keeps working without code changes.
+- **`additional_context` is emitted by both recall hooks.** The rules-file fallback remains enabled by default for older or unreliable Cursor builds.
 
-You can disable the rules-file write entirely (`useRulesFileFallback: false`) — then the plugin relies on `additionalContext`, which means no memory delivery until Cursor fixes the upstream bug. Useful only if you'd rather see the bug bite than have the plugin touch your workspace.
+You can disable the rules-file write entirely (`useRulesFileFallback: false`) — then the plugin relies on native `additional_context` support in the installed Cursor build.
```
Sanderhoff-alt, line 92: This, the "Native prompt recall is preferred" bullet (line 87) and line 81 all state that `beforeSubmitPrompt` delivers memories via `additional_context`. Cursor's hook reference documents `additional_context` only for `sessionStart` and `postToolUse`; `beforeSubmitPrompt` output is `continue` / `user_message` — which is also what the Note in this PR's description says. If that is right, prompt recall has no native delivery path and works only through the rules file, so this sentence is actively harmful: setting `useRulesFileFallback: false` would disable prompt recall entirely rather than fall back to a native channel. Please reconcile the docs with the actual behaviour here, in `hindsight-docs/docs-integrations/cursor.md` and in the generated `skills/hindsight-docs/references/sdks/integrations/cursor.md`.

hindsight-integrations/cursor/scripts/recall.py
```diff
@@ -0,0 +1,276 @@
...
+    except Exception as error:
+        print(f"[Hindsight] Recall failed: {error}", file=sys.stderr)
+        _write_recall_status("error", reason=str(error)[:200], bank_id=bank_id)
+        return
+
+    results = filter_by_min_scores(response.get("results", []), config.get("recallMinScores") or {}, config)
+    if not results:
+        debug_log(config, "No memories found")
+        _write_recall_status("empty", bank_id=bank_id, query_length=len(query))
+        return
```
Sanderhoff-alt, line 236: When a prompt recall returns nothing (or fails earlier) the hook returns without touching the rules file, so `.cursor/rules/hindsight-session.mdc` keeps the memories recalled for a previous prompt and, with `alwaysApply: true`, they stay in the agent's context for unrelated prompts. `session_start.py` calls `rotate_session_rules` for exactly this reason; the same rotation (or writing an empty block) is needed here, otherwise the delivery path the README calls best-effort also serves stale results.

Sanderhoff-alt (commented):
Draft, but a few things look like they need resolving before this can land: the retain `document_id` change reverts a fix that was added deliberately (and deletes its regression test), the new prompt hook can block prompt submission when `debug` is on, and the docs promise a Cursor context channel that the PR description itself says does not exist.

nicoloboschi:
closing as stale

nicoloboschi:
closing as stale

Issue #3485, opened 2026-08-14
Per-bank partial vector indexes exhaust the Postgres lock table at ~2,100 banks — recall AND bank deletion both fail cluster-wide


TaraTritt:
## Summary

With the pgvector backend (also pgvectorscale/vchord — anything where `uses_per_bank_vector_indexes()` is true), every bank creates 3 partial HNSW indexes on the shared `memory_units` table. PostgreSQL takes an `AccessShareLock` on **every index of a table during query planning**, so total index count across *all* banks is a hard, cluster-wide ceiling:

- At **10,000 banks** (30,021 indexes, v0.9.0, Postgres 17.10, default `max_locks_per_transaction=64` × `max_connections=100` = 6,400 lock slots): **every recall fails** with `out of shared memory / HINT: increase max_locks_per_transaction`, and **every `DELETE /v1/default/banks/{id}` fails identically** (it plans DML on `memory_units`) — so once a deployment crosses the wall, the API cannot delete banks to recover. We had to `DROP INDEX` via SQL directly.
- The wall is ≈ `lock_slots / 3` banks (~2,100 at defaults), confirmed bidirectionally: dead at 10k banks, recall revived after dropping back to 2k banks' indexes.
- Degradation starts an order of magnitude earlier. Recall p50 on a healthy bank, measured after `ANALYZE` (which changed nothing — this is plan-time overhead, not statistics):

  | banks | memory_units indexes | recall p50 | planning time, single trivial query |
  |---|---|---|---|
  | 10 | 51 | 154 ms | 0.2 ms |
  | 100 | 321 | 184 ms | 0.9 ms |
  | 500 | 1,521 | 2.3 s | 11 ms |
  | 2,000 | 6,021 | ~11 s | 51 ms |
  | 10,000 | 30,021 | dead | (lock exhaustion) |

- Bank *contents* are not the variable: a bank holding 421 memory units recalled at 190 ms p50 vs 161 ms empty, at the 10-bank scale.
- Bank **creation** stays flat throughout (mean 89 ms across all 10,000 creates) — so the failure is invisible until reads die.
- Managed Postgres offerings commonly refuse `ALTER SYSTEM` (verified on Fly MPG), and raising `max_locks_per_transaction` only moves the cliff — N concurrent queries each need the full per-query lock count, and plan-time cost keeps scaling with index count regardless.

## Method

[...]

PR #3493, opened 2026-08-14
fix(engine): drop per-bank vector indexes before bank delete data (#3485)

xiaoxiaoHe-E:
## Summary

Short-term fix for #3485 (per-bank partial vector indexes exhaust the Postgres lock table at ~2,100 banks — recall AND bank deletion both fail cluster-wide).

`delete_bank` previously dropped a bank's 3 per-bank vector indexes **only after** its delete transaction committed. With per-bank backends (pgvector/pgvectorscale/vchord), every bank adds 3 partial indexes to the shared `memory_units` table, and Postgres plans any statement against that table by taking a lock on **all** of its indexes. Past a few thousand banks the lock-manager shared-memory pool is exhausted, so every statement against `memory_units` fails — including the delete DML. Bank deletion was therefore part of the failure, not the recovery: the API could not delete banks to escape the wall.

This PR changes the order: the bank's per-bank vector indexes are dropped **before** the delete transaction starts.

## Why this works

- `DROP INDEX (CONCURRENTLY)` is a utility statement that only locks its own index plus the table — it still succeeds when the lock pool is exhausted, while the delete DML cannot plan.
- Dropping first keeps bank deletion usable at any index count, and a deletion attempt whose data delete cannot plan yet still removes 3 indexes on every call, restoring an API-driven recovery path.
- Still `CONCURRENTLY` on an autocommit connection (never inside a transaction), still wrapped in `retry_with_backoff` for the transient deadlock with concurrent index builds.
- Behavior preserved: `fact_type`-scoped deletes and `delete_bank_profile=False` keep the indexes (only the full-delete path drops them).

## Changes

- `hindsight_api/engine/memory_engine.py`: fetch `internal_id`, drop the 3 per-bank vector indexes on an autocommit connection, then run the delete transaction; removed the post-commit drop.
- `hindsight_api/engine/db/ops_postgresql.py` / `hindsight_api/engine/retain/bank_utils.py`: comments/docstrings updated to the new order.
- `tests/test_hnsw_indexes.py`: new regression test `test_delete_bank_drops_indexes_before_data` asserting the indexes are dropped while the bank's rows and profile still exist.
- `tests/test_bank_lifecycle_deadlock_retry.py`: docstring updated.

## Notes

- This is the short-term fix; the durable direction (global/partitioned indexes so index count is independent of bank count, plus a catalog-only recovery path) is discussed in the issue.
- Verified: 38 tests green (HNSW lifecycle incl. the new regression test, bank-lifecycle deadlock retry, non-SQL store deletion, extension bank tables, bank stats cache), ruff + `./scripts/hooks/lint.sh` clean.

Refs #3485


koriyoshi2041 (approved):
Tested the exact head 968010774 locally on macOS/Python 3.11 with embedded PostgreSQL: both HNSW delete-order tests passed, including the new before-data regression, and both bank lifecycle deadlock retry tests passed. The full-delete-only guard also preserves the scoped-delete behavior. The recovery ordering looks sound to me.

xiaoxiaoHe-E:
@nicoloboschi Thanks for the detailed write-up on #3561 — the threshold approach is a solid fix, and I think it addresses the root cause well.

One point I'd push back on gently: the claim that #3493 becomes unnecessary. I may be missing something, but it seems to me the two fixes are complementary rather than substitutes:

1. **The default threshold is 0.** For deployments that don't set `HINDSIGHT_API_VECTOR_INDEX_MIN_ROWS` (likely the majority), every bank still has its indexes, and the #3485 failure mode is unchanged there. In that configuration, drop-first in `delete_bank` is the only recovery path — worth keeping just for that.
2. **#3561 keeps the delete-path drop but doesn't change its timing.** At lock-table exhaustion, the delete DML itself can't be planned, so a post-commit drop never runs. #3493 is specifically about moving that drop *before* the delete so the delete stays possible. It feels like #3561 relies on that ordering rather than replacing it.
3. **The orphan sweep is a backstop, not a replacement.** It cleans up indexes left behind when `delete_bank` couldn't run — exactly the state #3493 prevents. Without #3493, wedged instances would produce those orphans at scale.

Happy to close it if you still think it's redundant after this — just wanted to make the case for keeping both, since they solve different halves of the same problem (#3485). Either way, thanks for driving the fix forward!

nicoloboschi:
Thanks for the lock-table analysis on #3485 — it was the right read of the problem. #3485 has since been fixed at the source by #3561: banks no longer get three partial vector indexes each, they *earn* an index by size, so the per-bank index count no longer scales with bank count and the lock table stops being the binding constraint.

On the ordering change specifically: `delete_bank` drops the indexes after the transaction commits because the drop runs `CONCURRENTLY`, which cannot run inside a transaction block. The residual cross-process deadlock is absorbed by `retry_with_backoff(max_retries=7)` from #3245.

Closing as superseded by #3561.

Issue #2229, opened 2026-06-16
Flaky: test_llm_trace::test_disabled_writes_no_rows — leaked global LLM-trace recorder records a disabled test's calls


nicoloboschi:
## Summary

`tests/test_llm_trace.py::test_disabled_writes_no_rows` is intermittently flaky: it disables LLM tracing and asserts the bank has **0** recorded LLM requests, but occasionally sees a non-zero count (observed `assert 6 == 0` in CI). It's not deterministic — it depends on test ordering across xdist workers — and it's unrelated to the code under test (it surfaced on an unrelated PR's CI, see #2225, where it was explicitly deferred).

```
FAILED tests/test_llm_trace.py::test_disabled_writes_no_rows - assert 6 == 0
```

## Root cause

LLM-trace recorders are stored in a **process-global registry**, not per-engine:

- `MemoryEngine.__init__` registers its recorder: `register_span_recorder(self._llm_recorder)` (`memory_engine.py:1063`).
- The provider/LLM-wrapper layer fans every call out to **all** registered recorders, regardless of which engine issued the call.
- The only thing that removes a recorder is `MemoryEngine.close()` → `unregister_span_recorder(self._llm_recorder)` (`memory_engine.py:2883`).

The `memory` fixture is **function-scoped** (`tests/conftest.py:440`), so every test builds its own engine and registers a fresh recorder. But teardown only conditionally closes it:

```python
yield mem
try:
    if mem._pool and not mem._pool._closing:
        await mem.close()
except Exception:
    pass
```

If `close()` is skipped (pool falsy / already `_closing`) **or** `close()` raises before line 2883 (e.g. `_maintenance_loop.stop()` throws — unregister happens *after* it, and the bare `except` swallows it), the recorder is **never unregistered** and leaks into the global registry while still `_enabled=True`.

`test_disabled_writes_no_rows` then disables only *its own* engine's recorder. Tests run with `SyncTaskBackend`, so its retain LLM calls execute synchronously during the POST — and a **leaked, still-enabled recorder from an earlier test** (e.g. any `trace_api_client` test, which sets `_enabled=True`) captures those calls and writes them to the shared pg0 DB under this test's `bank_id` (the bank comes from the trace context at call time). The endpoint then returns N>0 and the assert fails. The `6` is just the number of LLM calls in that retain.

This is why it's intermittent: it only fires when a prior enabled-recorder test ran on the same worker **and** its engine teardown didn't reach `unregister_span_recorder`.

[...]

PR #3497, opened 2026-08-14
fix(retain): entity postings and witness coverage inside a write-group

nicoloboschi:
Two silent defects on the external-store (write-group) retain path. Neither surfaces as an error, a
latency change, or a moved count.

## 1. Entity postings were dropped entirely

Retain resolves entities only after the memory ids exist, so `record_unit_entities` is a **second
write over rows the same write-group just wrote**. A store with deferred visibility keeps those rows
invisible until the group is decided — so a store implementation that reads them back before
rewriting gets nothing, and every unit→entity posting is lost.

The memories themselves land, so the bank looks healthy. Only the entity graph is quietly empty.

`docs/` for the cross-store protocol already warned about exactly this ("inside steps 3–5 the
provider cannot query the memories it just wrote... do not introduce it"). It was introduced anyway,
which is the argument for a regression test rather than another sentence in the doc.

**Fix:** thread the caller's `txn` through `record_unit_entity_postings` → `record_unit_entities`,
so the store can recognise the write as part of the group and rebuild the rows from what it already
staged instead of querying for rows it has hidden. The fix is deliberately *not* "make the read
work" — that is the thing the caveat forbids.

Inert for the Postgres store: its posting is an ordinary INSERT in the caller's own transaction,
which is already the unit of atomicity. `txn` is accepted and ignored there.

## 2. Seven `begin_txn` sites never re-recorded their witness

`begin_txn` writes the witness row *before any write has happened*, so a store that records what
each group wrote sees an empty group. Retain's two main paths (`_streaming_batch_write_ext`,
`_delta_batch_write_ext`) already re-recorded it before commit. These did not:

| site | path |
|---|---|
| `orchestrator` streaming group txn | per-batch streaming write |
| `orchestrator` delta group txn | delta re-ingest |
| `orchestrator` ×2 document-tracking | 0-fact / no-fact re-ingest replace-delete |
| `memory_engine` document delete | standalone `delete_document` |
| `memory_engine` unit delete | standalone delete-unit |
| `memory_engine` curation | edit / invalidate / revert |

Each now calls `write_txn_witness` as the last statement inside its transaction. That call is an
idempotent widening upsert — calling it twice is the intended usage, which is why this is additive
and safe.

## Verification

Placement was checked **structurally rather than by eye** — an AST pass asserting that every
`begin_txn` and `write_txn_witness` is lexically inside its `conn.transaction()`, and that every
commit-decide is outside it. (The three `decide_txn` calls that remain inside a transaction are all
`commit=False` abort paths, which is correct.)

- `tests/test_retain_ext_writegroup.py` now asserts the posting rides the same write-group as the
  fact write, on both the streaming and delta paths.
- `tests/` slice for retain/entity/curation/memories: 513 passed. The 3 failures and 335 errors in
  that slice are environmental (missing `sentence-transformers`, `pg0-embedded`, LLM API keys) and
  reproduce identically on an unmodified checkout.
- `ruff check`, `ruff format --check`, `ty check hindsight_api`: all clean.

## Deploy ordering

`hindsight_api` now calls `record_unit_entities(..., txn=...)`. The base class and the Postgres store
both accept it, but **an out-of-tree store extension that overrides `record_unit_entities` without a
`txn` parameter will raise `TypeError` on every retain.** Ship the extension update first, or
together.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
+++ b/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
@@ -1246,6 +1246,7 @@ class EntityResolver:
         self,
         unit_entity_pairs: list[tuple[str, str]] | list[tuple[str, str, datetime | None]],
         bank_id: str | None = None,
+        txn=None,
     ):
         """Store-owned variant of :meth:`link_units_to_entities_batch` that touches NO
         Postgres connection.
@@ -1257,6 +1258,10 @@ class EntityResolver:
         its connection-free store phase and never hold the data-plane connection across the
         object-store write. NOT for the Postgres store, whose posting is a real ``unit_entities``
         INSERT that requires the connection.
+
+        ``txn`` is the caller's write-group handle. For a store that keeps the posting on the
+        memory, this re-writes rows the same write-group just created, so it belongs to that
+        group — see :meth:`MemoriesExtension.record_unit_entities`.
         """
         if not unit_entity_pairs:
             return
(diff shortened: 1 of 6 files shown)


nicoloboschi:
`test-api (2/3)` failed on `test_llm_trace.py::test_disabled_writes_no_rows — assert 3 == 0`. That is the known flake tracked in #2229 ("leaked global LLM-trace recorder records a disabled test's calls"), fixed once in #2231 and touched again by #2491 — the same test, the same signature. It is unrelated to this PR, which touches only entity postings and write-group witness calls and cannot produce LLM-trace rows.

Everything else in that shard passed: 1763 passed, 186 skipped, 1 failed.

I could not re-run the job from the CLI (`cannot be rerun`), so it needs a maintainer re-run or a push to retrigger.

PR #3508, opened 2026-08-14
fix(llm): honor OpenAI's x-ratelimit-reset-* headers on 429s

romanbsd:
## Summary
- `_rate_limit_retry_at` only parsed the `Retry-After` header and the free-text error body, so a 429 without `Retry-After` (e.g. OpenAI's daily RPD cap) fell back to blind exponential backoff instead of the wait OpenAI actually specified.
- Parse OpenAI's `x-ratelimit-reset-requests` / `x-ratelimit-reset-tokens` headers (Go duration format, e.g. `"6m0s"`) and wait for whichever budget clears later. Requests/tokens can reset at different times.
- Existing `Retry-After` and free-text body parsing paths are unchanged.

## Test plan
- [x] `uv run pytest tests/test_openai_compatible_response_hardening.py -q` (11 passed)
- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run ty check` on the changed file
- [x] Verified the patch applies cleanly and tests pass against the `v0.9.1` tag

https://claude.ai/code/session_0184wdx18AyKn3amtvcqYRkU

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -496,10 +496,12 @@ def _rate_limit_retry_at(e: APIStatusError) -> datetime | None:
     retry_candidates: list[datetime] = []
     response = getattr(e, "response", None)
     headers = getattr(response, "headers", None)
+    has_future_retry_after = False
     if headers is not None:
         retry_at = _parse_retry_after_header(headers.get("retry-after") or headers.get("Retry-After"), now)
         if retry_at is not None and retry_at > now:
             retry_candidates.append(retry_at)
+            has_future_retry_after = True
 
         # Requests, tokens, and longer body-reported quotas are independent
         # budgets. Collect every future reset so a full/near-full header budget
@@ -538,7 +540,10 @@ def _rate_limit_retry_at(e: APIStatusError) -> datetime | None:
             if retry_at > now:
                 body_retry_at = retry_at
 
-    if body_retry_at is not None:
+    # Retry-After is an explicit server instruction and must not be extended
+    # by a less reliable, free-text error message. Body hints remain useful
+    # alongside reset headers when Retry-After is absent.
+    if body_retry_at is not None and not has_future_retry_after:
         retry_candidates.append(body_retry_at)
     return max(retry_candidates, default=None)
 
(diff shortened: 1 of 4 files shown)


Sanderhoff-alt (commented):
The Go-duration parsing itself looks right, including the contiguous-match handling. One issue with where the new branch returns.

hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
```diff
@@ -479,6 +500,17 @@ def _rate_limit_retry_at(e: APIStatusError) -> datetime | None:
...
+        # Requests and tokens are independent budgets that can reset at
+        # different times; wait for whichever one clears later.
+        reset_seconds = []
+        for key in ("x-ratelimit-reset-requests", "x-ratelimit-reset-tokens"):
+            if key in headers:
+                seconds = _parse_go_duration_seconds(headers[key])
+                if seconds is not None:
+                    reset_seconds.append(seconds)
+        if reset_seconds:
+            return now + timedelta(seconds=max(reset_seconds))
```
Sanderhoff-alt: The two sibling paths only return when the computed time is in the future (`retry_at is not None and retry_at > now`); this one returns unconditionally. OpenAI reports `0s` for a budget that is already full, which is the common case on a 429 triggered by the other budget or by a daily cap, so `reset_seconds == [0.0]` makes the function return `now` and skips the `reset at ...` / `try again in N hours` parsing below. A 429 whose body does carry the real wait (an RPD cap, for instance) then stops being deferred and goes back to blind exponential backoff, which is the behaviour this change sets out to remove.

Minimal fix — ignore non-positive resets so the body parsing is still reached:

```suggestion
        reset_seconds = []
        for key in ("x-ratelimit-reset-requests", "x-ratelimit-reset-tokens"):
            if key in headers:
                seconds = _parse_go_duration_seconds(headers[key])
                if seconds is not None and seconds > 0:
                    reset_seconds.append(seconds)
        if reset_seconds:
            return now + timedelta(seconds=max(reset_seconds))
```

A small but positive header (`233ms`) can still mask a much longer body window, so taking the later of the header-derived and body-derived times would be more robust than returning early here.

romanbsd: addressed that in the commit below

PR #3511, opened 2026-08-15
fix(llm): forward consolidation temperature

lusoris:
## Summary

Forward `llm_temperature_consolidation` to both observation-consolidation LLM calls:

- the batch synthesis call (`scope="consolidation"`)
- the semantic dedup adjudication call (`scope="consolidation_dedup"`)

The configuration field and its environment resolution were added by #2469 for #2459. The current call sites do not pass the resolved value, so `HINDSIGHT_API_LLM_TEMPERATURE_CONSOLIDATION` has no effect on these two operations. This also means the global `none` setting cannot omit temperature consistently across all consolidation calls.

## Changes

- Pass the resolved consolidation temperature to both LLM call sites.
- Add a focused batch-call regression test with a non-default temperature.
- Assert that dedup adjudication receives the configured default.
- Complete the minimal test config objects used by existing consolidation tests.

## Verification

- `pytest tests/test_consolidation_retry_budget.py tests/test_consolidation_dedup.py -q` — 52 passed.
- `pytest tests/test_consolidation.py::TestBuildResponseModel -q` — 8 passed.
- `pytest tests/test_llm_temperature_pipeline.py -q` — 3 passed.
- Repository pre-commit hooks — passed. The existing vulture and knip findings remained advisory and do not involve this diff.

## Risk

Low. This uses the existing resolved configuration and provider behavior. The default remains `0.0`. A configured `None` continues through `LLMConfig.call()` so providers can omit the parameter for models that reject explicit temperatures.

## Related

- Refs #2459
- Completes the consolidation wiring introduced in #2469

## AI assistance

OpenAI Codex with the `gpt-5` model, high reasoning, tool use, and code execution. The client does not expose the context-window size.


koriyoshi2041:
Checked exact head `467e0987` locally. The consolidation wiring and focused regressions behave as intended: 60 tests passed across retry-budget, dedup, temperature-pipeline, and response-model coverage.

The remaining 3 temperature-pipeline cases stopped during fixture setup before exercising this patch because the local environment points at a SOCKS proxy without `socksio`; this is an environment dependency error, not an assertion failure. In particular, the new `0.65` batch assertion and the dedup temperature assertion both passed.

nicoloboschi (approved):
Verified the gap: these were the only two `.call()` sites in `engine/consolidation/` and neither passed a temperature, so `HINDSIGHT_API_LLM_TEMPERATURE_CONSOLIDATION` only ever reached the mental-model delta call. `config` at both sites comes from `resolve_full_config()` (a full `HindsightConfig`), so the field is always present, and `None` still means omit. Ran the deterministic consolidation tests against the branch merged with current main: 52 passed, including the new `test_temperature_threaded_to_call`.

PR #3512, opened 2026-08-15
perf(postgres): preserve hash join in observation scoring

### LINKED ISSUE #3510: PostgreSQL planner turns set-wise observation scoring into an 8.5M-comparison nested loop
## Summary

The PostgreSQL observation graph query added by #3451 can still choose a pathological nested-loop plan on an aged bank, even though the scoring SQL is set-wise. The query is fully cached and CPU-bound.

`connected_sources` is estimated as one row but produces about 1,934 rows. PostgreSQL consequently makes it the outer side of the join to unnested candidate source IDs, rejects about 8.56 million row pairs, and spends about 4.4 seconds in that join alone.

A relationally equivalent `EXCEPT` formulation for `connected_sources` raises the estimate enough for PostgreSQL to choose the intended hash join. It reduced the representative SQL execution from 4,400ms to 39ms. Three additional seed sets returned identical `(id, score)` mappings and ran 63–70x faster.

This appears to be a follow-on planner/cardinality issue in the same path addressed by #3085 / #3451, rather than the original per-candidate correlated scoring problem. The affected code is `hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py::PostgreSQLOps.expand_observations()`.

## Environment

- Hindsight: `0.9.1-slim`
- PostgreSQL: `18.4`
- Upstream `main` checked at `396f63aafc9b618f04d446e2465cac95aa1cb426`; the observation expansion SQL is unchanged there
- Bank shape:
  - approximately 11,493 observations
  - 52,055 `unit_entities` rows
  - 8,872 distinct entities
  - source IDs per observation: mean 1.7, p95 4, p99 8, max 61
  - entity degree: mean 5.9, p95 15, p99 59, max 5,289
- Relevant data was warm in PostgreSQL buffers; this was not physical-I/O latency

## Steps to reproduce

1. Use a PostgreSQL-backed bank with a sufficiently high-fanout entity graph, such as the shape above.
2. Recall all three fact types at `budget=mid`, including observation graph retrieval.
3. Capture the `expand_observations()` entity-query plan using `EXPLAIN (ANALYZE, BUFFERS, FORMAT JSON)` with the ten observation semantic seeds.
4. Inspect the join inside the `scored` CTE between unnested candidate `source_memory_ids` and `connected_sources`.

## Actual behavior

A representative warm-cache plan reported:

```text
connected_sources: estimated 1 row, actual 1,934 rows
candidates:        estimated 280 rows, actual 1,589 rows
scored join:       Nested Loop
join output:       1,939 rows
rows removed:      8,556,011
planning time:     2.157ms
execution time:    4,400.057ms
```

Read-only sampling during an API recall showed the observation graph query dominating the request:

```text
observation graph query max age: 8.558s, CPU-active
semantic/BM25 query max age:     0.362s, CPU-active
world/experience graph max age:  0.005s, CPU-active
```

As a diagnostic only, `SET LOCAL enable_nestloop=off` changed the bad join to a hash join and reduced the same SQL to 53.777ms. Disabling nested loops globally is not proposed because it is too broad.

## Expected behavior

The set-wise scoring introduced by #3451 should retain a plan whose work is approximately linear in the unnested candidate and connected-source row counts. It should not regress into millions of rejected nested-loop comparisons because an upstream CTE is estimated at one row.

## Semantically equivalent rewrite tested

The current `connected_sources` CTE uses `DISTINCT` plus a correlated anti-join:

```sql
connected_sources AS (
    SELECT DISTINCT t.unit_id AS source_id
    FROM source_entities se
    CROSS JOIN LATERAL (
        SELECT ue_target.unit_id
        FROM unit_entities ue_target
        WHERE ue_target.entity_id = se.entity_id
        ORDER BY ue_target.unit_id DESC
        LIMIT 200
    ) t
    WHERE NOT EXISTS (
        SELECT 1
        FROM seed_sources ss
        WHERE ss.source_id = t.unit_id
    )
)
```

This equivalent set-difference form was tested:

```sql
connected_sources AS (
    (
        SELECT DISTINCT t.unit_id AS source_id
        FROM source_entities se
        CROSS JOIN LATERAL (
            SELECT ue_target.unit_id
            FROM unit_entities ue_target
            WHERE ue_target.entity_id = se.entity_id
            ORDER BY ue_target.unit_id DESC
            LIMIT 200
        ) t
    )
    EXCEPT
    (
        SELECT source_id
        FROM seed_sources
    )
)
```

`EXCEPT` provides the same distinct set as `SELECT DISTINCT ... WHERE NOT EXISTS (...)`. On the representative plan it changed the connected-source estimate from 1 to 3 rows—still low, but enough for the downstream score join to become a hash join:

```text
execution time: 39.274ms
rows removed by bad scored nested-loop join: 0
```

The candidate retains the logically redundant left-side `DISTINCT` for the smallest reviewable source diff. A fresh read-only plan of that exact form completed in 20.056ms.

## Equivalence and timing checks

The original and rewritten entity-query SQL were run side-by-side on three additional recall seed sets. Each query returned the top 300 rows.

| Case | Original | `EXCEPT` rewrite | Speedup | Same IDs and scores |
|---|---:|---:|---:|---|
| 1 | 1.753s | 0.026s | 67.1x | yes |
| 2 | 1.299s | 0.021s | 62.8x | yes |
| 3 | 2.069s | 0.030s | 69.8x | yes |

Median: 1.753s → 0.026s (67.4x).

These checks compared the full returned `id -> score` mapping, not only row counts. No Hindsight production code or retrieval settings were changed while gathering this evidence.

The candidate was also applied in an isolated worktree at upstream `main`. Three focused tests passed: a SQL-shape regression check plus PostgreSQL integration coverage, including an assertion that a seed's own source still does not contribute to candidate scores. The broader selected non-LLM link-expansion run passed 8 tests total; Ruff lint/format and the repository's `ty` type check also passed.

## Additional context

- Related: #3085
- The set-wise scoring fix in #3451 is present and does remove the original correlated `COUNT(DISTINCT ...)` subplan.
- This report concerns a different adverse plan inside that rewritten query: PostgreSQL underestimates `connected_sources` after the anti-join and then chooses a nested loop for the set-wise score join.
- Raising PostgreSQL `shared_buffers` from 128MiB to 1GiB eliminated almost all PostgreSQL buffer reads after warm-up but improved end-to-end retrieval p95 by only about 4%, which is consistent with this remaining bottleneck being CPU/query-plan work rather than storage latency.

stepanov1975:
## Summary

- replace the planner-sensitive correlated seed-source anti-join in PostgreSQL observation expansion with an equivalent `EXCEPT` set difference
- preserve the existing per-entity fanout limit, seed-source exclusion, retrieval breadth, and scoring behavior
- add a regression test for the SQL shape and strengthen seed-source scoring coverage

Fixes #3510.

## Problem

On a sufficiently connected PostgreSQL-backed observation graph, `connected_sources` can be underestimated as one row. PostgreSQL then chooses it as the outer side of a nested-loop join and performs millions of rejected comparisons while scoring otherwise set-wise candidates.

A representative warm-cache plan performed about 8.56 million rejected comparisons and took about 4.4 seconds in this path.

## Approach

The previous form used `SELECT DISTINCT ... WHERE NOT EXISTS (...)`. This change expresses the same distinct seed-source exclusion as `EXCEPT`:

```sql
SELECT DISTINCT t.unit_id AS source_id
...
EXCEPT
SELECT source_id FROM seed_sources
```

The left-side `DISTINCT` is intentionally retained to keep the source change minimal. The per-entity lateral `ORDER BY ... LIMIT` remains in the same position, before seed-source removal.

## Results

Across three additional realistic seed sets, the original and rewritten SQL returned identical complete top-300 `id -> score` mappings:

| Case | Original | `EXCEPT` | Speedup |
|---|---:|---:|---:|
| 1 | 1.753s | 0.026s | 67.1x |
| 2 | 1.299s | 0.021s | 62.8x |
| 3 | 2.069s | 0.030s | 69.8x |

Median: **1.753s -> 0.026s (67.4x)**.

A packaged PostgreSQL full-stack A/B on the same synthetic graph-linked database also returned identical ordered result IDs, entities, chunks, source facts, and non-final score inputs. Tiny final-score drift was reproduced on repeated requests against the same image and came from request-time freshness, not the SQL rewrite. Restart persistence and rollback to the unmodified 0.9.1 image were also exercised successfully.

## Testing

- `uv run pytest tests/test_observation_expansion_scoring.py -q -n0` — 3 passed
- selected non-LLM observation/link-expansion suite — 8 passed
- repository lint hook: `./scripts/hooks/lint.sh`
- Ruff formatting check
- `uv run ty check hindsight_api`
- packaged Hindsight 0.9.1 + PostgreSQL retain/recall smoke test
- real client/plugin retain and recall through the packaged image
- clean restart and rollback/restore smoke tests

The complete external-LLM test suite was not run locally; the focused and selected suites above exclude tests requiring an external LLM provider.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -938,9 +938,11 @@ class PostgreSQLOps(DataAccessOps):
                     ORDER BY ue_target.unit_id DESC
                     LIMIT {per_entity_limit}
                 ) t
-                WHERE NOT EXISTS (
-                    SELECT 1 FROM seed_sources ss WHERE ss.source_id = t.unit_id
-                )
+                -- Keep this as a set difference rather than a NOT EXISTS anti-join.
+                -- PostgreSQL can estimate the anti-join at one row and then turn the
+                -- set-wise scoring join below back into a quadratic nested loop.
+                EXCEPT
+                SELECT source_id FROM seed_sources
             ),
             connected_array AS (
                 SELECT array_agg(source_id) AS source_ids FROM connected_sources
--- a/hindsight-api-slim/tests/test_observation_expansion_scoring.py
+++ b/hindsight-api-slim/tests/test_observation_expansion_scoring.py
@@ -11,6 +11,7 @@ shared source facts, and it orders the results.
 
 import uuid
 from datetime import datetime, timezone
+from unittest.mock import AsyncMock
 
 import pytest
 
@@ -45,6 +46,34 @@ async def _insert_unit(
         datetime.now(timezone.utc),
     )
     return unit_id
+
+
+@pytest.mark.asyncio
+async def test_connected_sources_uses_set_difference():
+    """Keep the planner-safe set difference in the observation entity query."""
+    from hindsight_api.engine.db.ops import UpdatedWindow
+    from hindsight_api.engine.db.ops_postgresql import PostgreSQLOps
+
+    conn = AsyncMock()
+    conn.fetch.side_effect = [[], []]
+
+    await PostgreSQLOps().expand_observations(
+        conn,
+        "memory_units",
+        "unit_entities",
+        "memory_links",
+        [uuid.uuid4()],
+        300,
+        200,
+        UpdatedWindow(after=None, before=None, first_param_index=3),
+    )
+
+    entity_query = conn.fetch.await_args_list[0].args[0]
+    connected_sources = entity_query.partition("connected_sources AS (")[2].partition(
+        "),\n            connected_array AS"
+    )[0]
+    assert "EXCEPT" in connected_sources
+    assert "WHERE NOT EXISTS" not in connected_sources
 
 
 @pytest.mark.asyncio
@@ -83,6 +112,9 @@ async def test_score_counts_distinct_shared_sources(memory, request_context):
             seed = await _insert_unit(conn, mu, bank_id, "seed obs", "observation", [facts[0]])
             three = await _insert_unit(conn, mu, bank_id, "shares three", "observation", facts[1:4])
             one = await _insert_unit(conn, mu, bank_id, "shares one", "observation", [facts[1]])
+            includes_seed_source = await _insert_unit(
+                conn, mu, bank_id, "shares one plus the seed source", "observation", facts[:2]
+            )
             # Duplicate ids in the array must count once — the old query used
             # COUNT(DISTINCT ...) and the rewrite must not turn that into COUNT(*).
             dupes = await _insert_unit(
@@ -103,6 +135,7 @@ async def test_score_counts_distinct_shared_sources(memory, request_context):
         scores = {r["id"]: r["score"] for r in rows.entity}
         assert scores[three] == 3.0
         assert scores[one] == 1.0
+        assert scores[includes_seed_source] == 1.0, "the seed's own source must not contribute to the score"
         assert scores[dupes] == 1.0, "repeated source ids must count once"
         assert seed not in scores, "the seed itself must not come back as a result"
 


koriyoshi2041 (approved):
Verified exact head `b7989ac721c25fdf878225539545149ec987d03d`. The `EXCEPT` keeps the seed-source exclusion set-wise while preserving the per-entity limit, and the added case confirms a seed source cannot inflate the score. The focused PostgreSQL scoring suite passes 3/3 with the embedded DB/local-ML extras; targeted Ruff, formatting, `ty`, and diff checks also pass.

nicoloboschi:
Thanks — the analysis here (the anti-join estimate collapsing the scoring join back into a nested loop) was correct, and #3588 fixed it for #3510.

It took a different route: `connected_sources` now caps each entity with `row_number()` instead of a LATERAL + LIMIT, so the capped column stays traceable to `unit_entities.unit_id` and the planner gets real statistics (207–3,449 estimated against 2,242–4,193 actual) rather than an estimate of 1. `ops_postgresql.py` now carries a comment pinning that shape against future simplification.

Since the CTE no longer has the shape this patch targets, closing as superseded by #3588.

Issue #3506, opened 2026-08-14
default non-configurable new page triggers for coding agents integrations


mbravorus:
### Use Case

Allow users to override expensive defaults in coding-agent integrations.

### Problem Statement

There are hardcoded settings for all coding agents (27 as of now) that case all new pages/mental models to be created with trigger "on new memories", refresh_after_consolidation set to true and defaulting to full refresh from scratch. This causes pretty hefty token burn, especially if a few heavy repos get auto-surveyed/ingested. 

### How This Feature Would Help

Adding these config knobs would allow users to change the currently hardcoded defaults and potentially achieve significant token savings (or at least unexpected token spend)

### Proposed Solution

I suggest adding config knobs/fields to allow to change the defaults as hardcoded in PAGE_TRIGGER constant (with the addition of trigger mode which is now defaulting to full, and cron string if the trigger type is changed to cron)

specifically, add new config types to src/core/config.ts i resolveConfig() return object like this:

```
  pageTriggerType: raw.pageTriggerType === "cron" || raw.pageTriggerType === "manual"
     ? raw.pageTriggerType
     : "reactive",  // preserves today's behavior as the default
   pageTriggerCron: typeof raw.pageTriggerCron === "string" && raw.pageTriggerCron.trim()
     ? raw.pageTriggerCron.trim() : null,
```

and then in src/core/missions.ts convert PAGE_TRIGGER from static constant to a function like this:

```
   export function buildPageTrigger(cfg: ResolvedConfig): MentalModelTrigger {
     const base = { fact_types: ["world", "experience", "observation"], mode: cfg.pageTriggerMode };
     switch (cfg.pageTriggerType) {
       case "cron": return { ...base, refresh_cron: cfg.pageTriggerCron! };
       case "manual": return { ...base, refresh_after_consolidation: false };
       default: return { ...base, refresh_after_consolidation: true };
     }
   }
```

### Alternatives Considered

unfortunately, the only workaround is to manually patch under ~/.hindsight/code-agents/dist which is fragile and will be silently overwritten on upgrade. or running a cron-based script fixing the newly created pages/mental-models post factum, which is also kinda sad

### Priority

Important - affects my workflow

### Checklist

- [x] I would be willing to contribute this feature

PR #3513, opened 2026-08-15
[CLI] Add a `--trigger-mode full|delta` flag to `mental-models update`

mdbenito:
This PR adds `--trigger-mode full|delta` to `hindsight mental-model update` (and now `create`), propagating it to the `trigger.mode` field of the `PATCH` endpoint.

The new flag composes with the existing `--trigger-refresh-after-consolidation`.

This PR also adds some test coverage for trigger override construction and serialization, a live round-trip integration test, and updates the OpenAPI coverage note.

---

**Update (maintainer, rebased onto main):** the original note here said that passing a trigger flag replaces the stored trigger configuration. That is no longer true of the API — `PATCH /mental-models/{id}` patches a supplied trigger over the stored one, and reads which fields were "named" from the ones the request actually carried (#3506/#3549).

The CLI still reset them, for a reason specific to the generated Rust client: `MentalModelTriggerInput` serializes its four non-`Option` fields unconditionally (`mode`, `refresh_after_consolidation`, `exclude_mental_models`, `keep_trace`), so a freshly-built trigger looked to the server like an explicit request for this type's defaults. Switching a model to delta mode also turned off its trace capture and its mental-model exclusion; toggling refresh-after-consolidation reset it to full mode.

Two commits on top of the original:

- **`fix(cli): merge mental-model update trigger flags over the stored trigger`** — the update reads the model's current trigger and applies only the flags the user passed on top of it, so every setting they did not name survives. The `Option`-typed fields (recall budgets, `fact_types`, `tag_groups`, `response_schema`) were already safe via `skip_serializing_if`; now the rest are too. Setting one refresh trigger drops an unstated other, mirroring the server's own `_merge_trigger`, so naming only a cron does not produce a combination the API rejects; naming both in one command is the one case the user has to settle.

  Also exposes the remaining scalar trigger settings: `--trigger-refresh-cron`, `--trigger-min-refresh-interval-seconds`, `--trigger-tags-match`, `--trigger-keep-trace`, `--trigger-exclude-mental-models`. An empty string clears the cron schedule or the tags-match override.

- **`feat(cli): add --trigger-mode to mental-model create`** — the flag existed only on `update`, so a delta model had to be created full and switched afterwards, with the first refresh in between running as a from-scratch build.

The same bug class in the Python SDK wrapper (where the generated model's defaults ride along on every request) is fixed separately in #3913.

https://claude.ai/code/session_01WbaYxouCERUv9djo3GnnA2

### DIFF
--- a/hindsight-cli/src/commands/mental_model.rs
+++ b/hindsight-cli/src/commands/mental_model.rs
@@ -23,6 +23,47 @@ fn parse_tags_match(value: &str) -> Result<types::TagsMatch> {
     })
 }
 
+/// Parse a --trigger-mode value into the generated refresh-mode enum
+fn parse_trigger_mode(value: &str) -> Result<types::Mode> {
+    Ok(match value.to_lowercase().as_str() {
+        "full" => types::Mode::Full,
+        "delta" => types::Mode::Delta,
+        other => anyhow::bail!("invalid --trigger-mode '{other}': expected one of full, delta"),
+    })
+}
+
+/// Build the trigger override sent on update when at least one trigger flag
+/// was passed. `None` (no trigger flags) leaves the stored trigger config
+/// untouched on the server. When an override IS sent, the server replaces the
+/// whole trigger, so fields the user did not specify fall back to their
+/// defaults.
+fn build_trigger_override(
+    trigger_mode: Option<&str>,
+    trigger_refresh_after_consolidation: Option<bool>,
+) -> Result<Option<types::MentalModelTriggerInput>> {
+    if trigger_mode.is_none() && trigger_refresh_after_consolidation.is_none() {
+        return Ok(None);
+    }
+    Ok(Some(types::MentalModelTriggerInput {
+        mode: trigger_mode
+            .map(parse_trigger_mode)
+            .transpose()?
+            .unwrap_or(types::Mode::Full),
+        refresh_after_consolidation: trigger_refresh_after_consolidation.unwrap_or(false),
+        refresh_cron: None,
+        exclude_mental_models: false,
+        exclude_mental_model_ids: None,
+        fact_types: None,
+        tag_groups: None,
+        tags_match: None,
+        include_chunks: None,
+        recall_max_tokens: None,
+        recall_chunks_max_tokens: None,
+        response_schema: None,
+        keep_trace: false,
+    }))
+}
+
 /// List mental models for a bank
 pub fn list(
     client: &ApiClient,
@@ -194,6 +235,7 @@ pub fn update(
     max_tokens: Option<i64>,
     tags: Option<Vec<String>>,
     trigger_refresh_after_consolidation: Option<bool>,
+    trigger_mode: Option<&str>,
     verbose: bool,
     output_format: OutputFormat,
 ) -> Result<()> {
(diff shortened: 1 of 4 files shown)


Sanderhoff-alt (commented):
The flag itself looks right, but the branch has been overtaken by the trigger-merge change on main, which makes the request body this builds overwrite settings the user did not pass.

hindsight-cli/src/commands/mental_model.rs
```diff
@@ -23,6 +23,47 @@ fn parse_tags_match(value: &str) -> Result<types::TagsMatch> {
...
+        exclude_mental_models: false,
+        exclude_mental_model_ids: None,
+        fact_types: None,
+        tag_groups: None,
+        tags_match: None,
+        include_chunks: None,
+        recall_max_tokens: None,
+        recall_chunks_max_tokens: None,
+        response_schema: None,
+        keep_trace: false,
```
Sanderhoff-alt: `mode`, `refresh_after_consolidation`, `exclude_mental_models` and `keep_trace` are non-`Option` in the generated `MentalModelTriggerInput`, so they are always serialized. Since #3755 the PATCH route no longer replaces the trigger: `api/http.py` passes `body.trigger.model_dump(exclude_unset=True)` into `MemoryEngine._merge_trigger`, which layers only the keys the client actually sent over the stored trigger. Every key filled in here therefore counts as explicitly set.

Effect: `hindsight mental-model update <id> --trigger-mode delta` on a model with `refresh_after_consolidation: true` silently turns auto-refresh off (and clears `keep_trace`), and symmetrically `--trigger-refresh-after-consolidation true` resets a delta model back to `full`.

To send only the flags the user passed, either read the stored trigger first via `get_mental_model` and carry its values into this literal, or PATCH a hand-built JSON body the way `import_bank_template` does in `hindsight-cli/src/api.rs`. Note that `build_trigger_override_refresh_only_defaults_to_full_mode` currently pins the resetting behaviour, so it needs updating alongside the fix.

hindsight-cli/src/commands/mental_model.rs
```diff
@@ -23,6 +23,47 @@ fn parse_tags_match(value: &str) -> Result<types::TagsMatch> {
...
+    if trigger_mode.is_none() && trigger_refresh_after_consolidation.is_none() {
+        return Ok(None);
+    }
+    Ok(Some(types::MentalModelTriggerInput {
+        mode: trigger_mode
+            .map(parse_trigger_mode)
+            .transpose()?
+            .unwrap_or(types::Mode::Full),
+        refresh_after_consolidation: trigger_refresh_after_consolidation.unwrap_or(false),
+        refresh_cron: None,
```
Sanderhoff-alt, line 71: This branch predates `min_refresh_interval_seconds` on `MentalModelTrigger`; the equivalent literal on `main` now sets it. After rebasing, this initializer needs `min_refresh_interval_seconds: None` or it will not compile.

hindsight-cli/src/main.rs
```diff
@@ -1073,6 +1073,12 @@ enum MentalModelCommands {
         /// Enable/disable automatic refresh after observations consolidation
         #[arg(long)]
         trigger_refresh_after_consolidation: Option<bool>,
+
+        /// Refresh mode: full or delta. Passing a trigger flag replaces the
+        /// stored trigger configuration: trigger settings you do not pass
+        /// reset to their API defaults.
```
Sanderhoff-alt: This help text (and the same claim in the doc comment on `build_trigger_override`, and in the PR description) no longer matches the server. Since #3755 `PATCH /mental-models/{id}` merges the supplied trigger field by field over the stored one, so unstated settings such as `refresh_cron`, `tags_match`, `fact_types` and `min_refresh_interval_seconds` are preserved, while only the always-serialized fields are overwritten. Please reword once the request body carries just the flags that were passed.

PR #3515, opened 2026-08-15
fix: preserve Voyage asymmetric embedding semantics

Tosko4:
## Summary

- send `input_type="document"` when initializing Voyage embeddings and embedding stored content
- send `input_type="query"` when embedding recall queries
- preserve provider-default behavior for `encode()` and every non-Voyage LiteLLM provider
- add regression coverage for initialization, query/document routing, and non-Voyage compatibility

## Why

Hindsight already exposes Voyage models through `litellm-sdk`, but the adapter currently omits Voyage's asymmetric retrieval `input_type`.

Voyage's retrieval contract embeds corpus content as `document` and search text as `query`. Passing those roles through Hindsight keeps stored vectors and recall queries in the intended retrieval spaces instead of relying on the provider's general-purpose default.

## Validation

- `uv run python -m pytest -n 0 tests/test_litellm_sdk_embeddings.py -q` (`29 passed, 3 skipped`)
- `uv run ruff check hindsight_api/engine/embeddings.py tests/test_litellm_sdk_embeddings.py`
- `uv run ruff format --check hindsight_api/engine/embeddings.py tests/test_litellm_sdk_embeddings.py`
- `./scripts/hooks/lint.sh`
- end-to-end retain, recall, worker, and reranking validation in a 2048-dimensional Voyage deployment

## Risk and impact

The behavioral change is restricted to LiteLLM model names beginning with `voyage/`. Other providers continue to omit `input_type`, and direct `encode()` callers retain provider-default behavior.

Deployments that already indexed content with a Voyage model while `input_type` was omitted should rebuild those stored embeddings when adopting this change, so document and query vectors consistently use the asymmetric retrieval contract.

## Scope

Two files only:

- `hindsight-api-slim/hindsight_api/engine/embeddings.py`
- `hindsight-api-slim/tests/test_litellm_sdk_embeddings.py`

No schema, configuration, dependency, generated-client, or public API changes.

## Rollout

No database migration is included. New Voyage embeddings use the correct role automatically after upgrade. Existing Voyage-backed indexes should be re-embedded as noted above.

## Reviewer focus

Please verify that:

- initialization and stored content use `document`
- recall queries use `query`
- `encode()` and non-Voyage providers stay unchanged

PR #3090 also touches these two files for retry and timeout behavior. It is not a duplicate, but whichever change lands second may need a focused rebase.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/embeddings.py
+++ b/hindsight-api-slim/hindsight_api/engine/embeddings.py
@@ -1312,6 +1312,8 @@ class LiteLLMSDKEmbeddings(Embeddings):
                 embed_kwargs["dimensions"] = self.output_dimensions
                 if self.model.startswith("openai/"):
                     embed_kwargs["allowed_openai_params"] = ["dimensions"]
+            if self.model.startswith("voyage/"):
+                embed_kwargs["input_type"] = "document"
 
             # Use async embedding method (standard in litellm)
             response = await self._litellm.aembedding(**embed_kwargs)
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
lgtm

PR #3521, opened 2026-08-16
fix: disable automatic cache affinity for Azure OpenAI

### LINKED ISSUE #3518: azure-openai base_url triggers prompt_cache_key, returns 400 on Azure Foundry
## Bug description

When `provider=openai` and `base_url` points to an Azure OpenAI / Azure Foundry endpoint (`*.openai.azure.com`), the `auto` cache-affinity mode resolves to `OPENAI_PROMPT_CACHE_KEY`. This adds a `prompt_cache_key` field to every chat completion request.

Azure Foundry (at least when serving DeepSeek models) does **not** support this parameter and returns:

```json
{
  "error": {
    "code": "unrecognized_request_argument",
    "message": "Unrecognized request argument supplied: prompt_cache_key"
  }
}
```

This breaks all LLM-dependent operations (retain, reflect, consolidation).

## Where

`hindsight_api/engine/cache_affinity.py`, line 45:

```python
_OPENAI_DOMAINS = ("openai.com", "openai.azure.com")
```

The `"openai.azure.com"` entry causes `resolve_cache_affinity()` to return `OPENAI_PROMPT_CACHE_KEY` for any Azure OpenAI host.

## Reproduction

1. Set `HINDSIGHT_API_LLM_PROVIDER=openai`
2. Set `HINDSIGHT_API_LLM_BASE_URL=https://<resource>.openai.azure.com/openai/v1`
3. Set a valid Azure Foundry key
4. Start hindsight-api → connection verification passes (it uses `/models`)
5. Call retain → 400 error with `unrecognized_request_argument: prompt_cache_key`

```python
# Manual repro:
import httpx
resp = httpx.post(
    "https://<resource>.openai.azure.com/openai/v1/chat/completions",
    json={"model": "deepseek-v4-pro", "messages": [{"role": "user", "content": "hi"}],
           "prompt_cache_key": "test"},  # ← Azure doesn't know this field
    headers={"api-key": "..."}
)
# → 400
```

## Workaround

Set `HINDSIGHT_API_LLM_CACHE_AFFINITY=none` in environment. This disables the feature globally, which is acceptable but means prompt caching is also disabled for actual OpenAI hosts.

## Suggested fix

Either:
1. Remove `"openai.azure.com"` from `_OPENAI_DOMAINS` — Azure OpenAI does not document `prompt_cache_key` support
2. Or make the domain list configurable so operators can override without a code patch
3. Or add a capability check (e.g. probe the endpoint) before emitting the parameter

## Environment

- hindsight-api version: 0.9.1
- Azure Foundry with deepseek-v4-pro model
- Python 3.11, Windows 11

koriyoshi2041:
## Problem

With the `openai` provider and an `*.openai.azure.com` base URL, automatic cache affinity currently adds OpenAI's `prompt_cache_key` field. Azure Foundry rejects that field with `unrecognized_request_argument`, breaking LLM-backed operations.

Fixes #3518.

## Fix

Remove Azure OpenAI hosts from the automatic `prompt_cache_key` allowlist. Native OpenAI and `openai.com` hosts keep the existing behavior, while Azure hosts now resolve `auto` to `none`.

The configuration and Azure setup docs now describe the same behavior.

## Test

- `uv run pytest tests/test_cache_affinity.py -q` — 67 passed
- Ruff check and format check on the changed Python files — passed
- `uv run ty check hindsight_api/engine/cache_affinity.py hindsight_api/config.py` — passed
- `git diff --check` — passed

The repository-wide lint script reached the TypeScript lint step but could not load `@eslint/js` in this fresh worktree. No TypeScript files are changed by this PR.

## Risk

Low. This only narrows the `auto` host allowlist. Operators who explicitly select `openai_prompt_cache_key` retain that override, and non-Azure host behavior is unchanged.


nicoloboschi (approved):

nicoloboschi:
Approved — the code change is right, but the **docs wording is factually wrong** and should be fixed before merge.

`hindsight-docs/docs/developer/models.mdx` (and the mirrored `skills/hindsight-docs/references/developer/models.md`) now say:

> "Azure does not support OpenAI's `prompt_cache_key` field"

Azure OpenAI *does* accept it. Microsoft documents it as a supported request field with no api-version requirement:

> "On GPT-5.6 models and later model families, set the `prompt_cache_key` parameter and reuse the same key for requests that share long, common prompt prefixes… You don't need a specific API version to use `prompt_cache_key`."

and is explicit about which cache fields *do* 400 — a list `prompt_cache_key` is deliberately not on:

> "Models before the GPT-5.6 family don't support `prompt_cache_options` or `prompt_cache_breakpoint`. Requests that include these parameters return a `400` error."

The 400 in #3518 came from `deepseek-v4-pro` — a **non-OpenAI Foundry model** served over the *same* `<resource>.openai.azure.com/openai/v1` endpoint as GPT deployments. That passthrough validates against the upstream model's parameter set, so it rejects an OpenAI-only field.

So removing Azure from `_OPENAI_DOMAINS` is still correct, just for a different reason: the host fronts both deployments that accept the field and deployments that reject it, and an allowlist keyed on hostname can't tell them apart. That reason is worth stating, because as written the docs steer Azure GPT operators away from an option that would actually work for them.

Suggested replacement for the `models.mdx` bullet (and its `skills/` mirror):

```
- Azure OpenAI accepts `prompt_cache_key` on GPT deployments (it improves cache
  routing on GPT-5.6 and later), but the same `*.openai.azure.com` endpoint also
  serves non-OpenAI Foundry models (DeepSeek, Llama, Mistral) that reject it with
  `unrecognized_request_argument`. The host alone can't tell the two apart, so
  [`HINDSIGHT_API_LLM_CACHE_AFFINITY`](./configuration#llm-provider) resolves
  `auto` to `none` on Azure hosts. If your deployment serves an OpenAI model, set
  it explicitly to `openai_prompt_cache_key`.
```

The `configuration.md` table line only describes the resolution behaviour ("anything else, including Azure OpenAI, → `none`"), so it's accurate as-is — though the explicit-override hint would help Azure GPT operators there too.

Sources: [Prompt caching with Azure OpenAI — Microsoft Learn](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/prompt-caching), [MS Q&A: GPT-5.6 implicit prompt caching](https://learn.microsoft.com/en-us/answers/questions/5942997/gpt-5-6-implicit-prompt-caching-and-explicit-promp)

Issue #3517, opened 2026-08-15
`_find_pid_on_port` fails on Linux without `lsof`, producing "Could not find PID for port" warnings on every daemon stop


PaNe0302:
# `_find_pid_on_port` fails on Linux without `lsof`, producing "Could not find PID for port" warnings on every daemon stop

## Summary

On Linux systems where the `lsof` binary is not installed (e.g. minimal containers, some Arch-based distros like CachyOS by default), `DaemonEmbedManager._find_pid_on_port()` silently returns `None`, so every call to `stop()` logs `WARNING hindsight_embed.daemon_embed_manager: Could not find PID for port 9177` and the daemon can never be stopped programmatically (e.g. during Hermes Desktop backend restarts).

## Environment

- Hindsight plugin `hindsight-embed` (installed in Hermes Agent venv, v0.8.6 era code)
- OS: Linux (CachyOS, Arch-based) — `lsof` not installed by default
- `ss` (from `iproute2`) IS present and is the standard tool for this on modern Linux
- `python-psutil` is NOT a dependency, so no pure-Python fallback exists today

## Root cause

`daemon_embed_manager.py`, `_find_pid_on_port()` (around line 287):

```python
else:
    # Use lsof on macOS/Linux
    result = subprocess.run(
        ["lsof", "-ti", f":{port}", "-sTCP:LISTEN"],
        capture_output=True,
        text=True,
        timeout=5,
    )
    if result.returncode == 0 and result.stdout.strip():
        return int(result.stdout.strip().split()[0])
```

When `lsof` is missing, `subprocess.run` raises `FileNotFoundError`, which is caught by the broad `except (…, FileNotFoundError): pass` — so the function returns `None` **without any fallback**. The warning at line 967 (`Could not find PID for port {port}`) then fires, and `_kill_process(pid)` is skipped.

Note the Windows branch already has a fallback (`netstat`), but the macOS/Linux branch relies on a single external binary that is not a declared dependency and is not guaranteed on Linux.

## Impact

- Log noise on every daemon stop attempt on affected systems
- `stop()` cannot actually stop the daemon (returns after the health-check wait loop)
- Matters in embedded usage (Hermes plugin) where `hindsight-embed`'s lifecycle management is exercised on every Desktop restart

## Suggested fix

Fall back to `ss` (iproute2, present on essentially every modern Linux) before giving up, or prefer it outright:

```python
else:
    # Prefer ss (iproute2) — present on essentially all modern Linux.
[...]

PR #3523, opened 2026-08-16
fix Linux daemon PID lookup without lsof

koriyoshi2041:
Complements #3585 for minimal Linux images where neither `lsof` nor `ss` is installed.

PID discovery now falls back to procfs: it finds listening TCP/TCP6 socket inodes for the target port in `/proc/net`, then resolves the owning process through `/proc/<pid>/fd` symlinks. Existing `lsof` and `ss` paths remain preferred.

Validation after rebasing onto current `main`:

- `uv run pytest tests/test_embed_manager.py -q` — 28 passed
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy NO_PROXY=127.0.0.1,localhost uv run pytest tests/ -q` — 187 passed
- `uv run ruff check hindsight_embed tests` — passed
- `uv run ruff format --check hindsight_embed tests` — 22 files already formatted
- `git diff --check` — passed

The proxy variables are removed for the full suite because the local environment points HTTPX at a SOCKS proxy without its optional `socksio` dependency; the first run otherwise produced 18 control-center transport setup failures unrelated to this patch.

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -43,6 +43,43 @@ def _parse_float_env(name: str, default: float) -> float:
         return default
 
 
+def _find_linux_pid_via_proc(port: int, proc_root: Path = Path("/proc")) -> int | None:
+    """Resolve a listening TCP socket to its owner without external tools."""
+    socket_inodes: set[str] = set()
+    for table in ("tcp", "tcp6"):
+        try:
+            lines = (proc_root / "net" / table).read_text().splitlines()[1:]
+        except OSError:
+            continue
+
+        for line in lines:
+            fields = line.split()
+            if len(fields) > 9 and fields[3] == "0A":
+                try:
+                    local_port = int(fields[1].rsplit(":", 1)[1], 16)
+                except (IndexError, ValueError):
+                    continue
+                if local_port == port:
+                    socket_inodes.add(fields[9])
+
+    if not socket_inodes:
+        return None
+
+    for process_dir in proc_root.iterdir():
+        if not process_dir.name.isdigit():
+            continue
+        try:
+            descriptors = (process_dir / "fd").iterdir()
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
Ran this against a real listener rather than the mocked `subprocess.run`, since CI's `test-embed` job is skipped on this PR (it is gated on `has_secrets`, which is false for a fork), so the new test has not executed upstream.

python:3.12-slim plus iproute2, `lsof` absent, a real socket bound to 127.0.0.1:9177, the package pip-installed from each side:

- base `205e47b4`: `_find_pid_on_port(9177)` returns `None`
- head `b69873f3`: returns 342, which is the listener's own pid
- `tests/test_embed_manager.py` at head: 22 passed

So the `ss -H -ltnp "sport = :9177"` argv is valid as written (iproute2 6.15.0), which the mocked test cannot show on its own.

One gap, not blocking. #3517 lists minimal containers next to CachyOS, and those often have neither binary. On stock python:3.12-slim (no `lsof`, no `ss`) head still returns `None`, and since #3171 a missing pid makes `stop()` return False rather than only log a warning. A stdlib fallback would cover that case with no new dependency: read `/proc/net/tcp` and `/proc/net/tcp6` for the listening socket's inode (state `0A`), then match it against the `/proc/[0-9]*/fd/*` symlinks. I tried that in a container with neither binary and it resolved the correct pid. Fine as a follow-up if you would rather keep this PR to the environment that was reported.

koriyoshi2041:
Added a stdlib `/proc` fallback in 80d75f5e for Linux systems with neither `lsof` nor `ss`. It matches LISTEN entries in `/proc/net/tcp{,6}` to `/proc/<pid>/fd` socket inodes; focused embed-manager coverage is now 24/24, with source lint, format, and diff checks clean.

nicoloboschi:
already fixed in main

PR #3525, opened 2026-08-16
fix(coding-agents): start daemon for dsh adapter

### LINKED ISSUE #3524: dsh adapter: daemon mode never starts the daemon (ensureDaemon parity missing)
## Summary

Hook-based integrations (claude-code, codex, cursor-cli, copilot-cli, antigravity, grok-build, ...)
call ensureDaemon() from their session-start/stop hooks, so the local daemon is lazily started or
adopted whenever a session needs it. The dsh adapter does not — dist/dsh.js contains no daemon
management code — so with serverMode: "daemon", every hindsight_* tool fails with
"connect ECONNREFUSED 127.0.0.1:9077" until the user manually runs
node dist/daemon-start.js --harness dsh. After the daemon idles out (daemonIdleTimeout, default
300s) nothing starts it again, so memory silently breaks until the next manual start.

## Repro

1. dsh plugin --profile web add @vectorize-io/hindsight-coding-agents
   (same result with npx @vectorize-io/hindsight-coding-agents install dsh)
2. ~/.hindsight/coding-agent.json = {"serverMode": "daemon"}
3. Restart dsh web
4. Call any hindsight_* tool -> {"error":"fetch failed: connect ECONNREFUSED 127.0.0.1:9077"}
5. GET http://127.0.0.1:9077/health -> connection refused (no daemon process)
6. Manually run node <pkg>/dist/daemon-start.js --harness dsh -> tools start working

## Evidence (v0.3.4, current latest on npm)

- dist/dsh.js has no references to ensureDaemon / HindsightServer / daemon-start / uvx;
  src/core/daemon.ts is tree-shaken out of the dsh bundle, while e.g.
  dist/claude-sessionstart-hook.js and dist/codex-sessionstart-hook.js include ensureDaemon +
  startDaemonDetached and call them at session start/retain.
- dist/installer.js dsh wiring ("install dsh") only writes the ~/.dsh/cordis.patch.yml row and
  installs the skill under ~/.agents/skills — no daemon start, no service registration.
- README's "The plugin starts hindsight-embed on demand at 127.0.0.1:9077" therefore holds for the
  hook harnesses but not for dsh.

## Environment

Windows 11 (zh-CN), Node 22.x, DeepSeek Harness profile "web",
@vectorize-io/hindsight-coding-agents@0.3.4, daemon mode.

## Suggested fix

Give the dsh adapter parity with the hook harnesses: call ensureDaemon(cfg, "dsh")
(adopt-if-healthy, detached spawn + health wait) when the per-workspace client is created in
workspaceFor(), or lazily before the first tool call of a session. At minimum, document the manual
daemon-start.js step in the dsh section of the README.

## Related environment notes (hit while bringing daemon mode up on Windows)

- On zh-CN Windows, "profile create" crashes with "'gbk' codec can't decode byte ..." unless
  PYTHONUTF8=1 is exported for the spawned process.
- Behind a proxy that maps public hosts into reserved ranges, the SentenceTransformer download dies
  with httpx.RemoteProtocolError: Server disconnected without sending a response;
  HF_ENDPOINT=https://hf-mirror.com works around it.

koriyoshi2041:
Closes #3524.

The dsh adapter now routes daemon-mode workspaces through the shared daemon lifecycle:

- session start starts or adopts the local daemon with the normal startup budget
- resumed sessions retry at pre-step before recall
- turn stopping gets the longer retain budget before write-back
- Cloud and self-hosted modes remain no-ops through the existing ensureDaemon guard

A regression covers daemon startup before workspace seeding.

Validation:
- focused dsh + daemon tests: 24 passed
- package suite: 48 files / 518 tests passed; 3 unrelated suites could not import locally because the existing install lacks @modelcontextprotocol/sdk and @opencode-ai/plugin
- repository pre-commit lint hooks passed
- Prettier and diff checks passed

### DIFF
--- a/hindsight-integrations/coding-agents/src/dsh.test.ts
+++ b/hindsight-integrations/coding-agents/src/dsh.test.ts
@@ -135,6 +135,28 @@ describe("dsh write-back", () => {
 });
 
 describe("dsh session start", () => {
+  it("starts or adopts the local daemon before seeding the workspace", async () => {
+    const calls: string[] = [];
+    const core = {
+      seedIfCold: vi.fn(async () => {
+        calls.push("seed");
+      }),
+    };
+    const workspace = {
+      core: core as never,
+      root: "/repo",
+      ensureDaemon: vi.fn(async () => {
+        calls.push("daemon");
+      }),
+    } as Workspace;
+    const hooks = createDshHooks(() => workspace);
+
+    hooks.sessionStart({ agent: agentWith("s-1") });
+    await vi.waitFor(() => expect(workspace.ensureDaemon).toHaveBeenCalledWith(12_000));
+
+    expect(calls).toEqual(["daemon", "seed"]);
+  });
+
   it("starts the repo seed exactly once, however many sessions open", () => {
     const core = { seedIfCold: vi.fn(async () => {}) };
     const workspace = { core: core as never, root: "/repo" } as Workspace;
--- a/hindsight-integrations/coding-agents/src/dsh.ts
+++ b/hindsight-integrations/coding-agents/src/dsh.ts
@@ -25,6 +25,7 @@
 import { randomUUID } from "node:crypto";
 import { deriveBankId } from "./core/bank";
 import { applyBankConfig, loadConfig } from "./core/config";
+import { DAEMON_WAIT_RETAIN_MS, DAEMON_WAIT_SESSION_START_MS, ensureDaemon } from "./core/daemon";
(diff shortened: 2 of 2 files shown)


koriyoshi2041:
Closing because #3542 fixed the same daemon-start gap at the shared `RuntimeCore` boundary and covered all persistent-plugin harnesses, not only dsh. That is the better shape for this fix; there is nothing left from this branch that should be merged separately.

PR #3526, opened 2026-08-16
fix(embed): verify process ownership before termination

### LINKED ISSUE #3520: hindsight-embed: _clear_port kills any PID holding the port with no ownership check, terminating unrelated processes
**Component:** `hindsight-embed` (`daemon_embed_manager.py`)
**Version:** hindsight-embed 0.8.6 / hindsight-api 0.8.6 (PyPI)
**Platform:** Linux (Ubuntu), Python 3.11

## Summary

`_clear_port()` reclaims a busy port by killing **whatever process is listening on it**, with no verification that the process is a Hindsight daemon. When the port is held by an unrelated process, that process is SIGTERMed.

This is not hypothetical — on our host it terminated a separate, legitimately-owned `hindsight-api` service (systemd-managed, its own database) because the embedded daemon manager wanted the same port. The service had no `Restart=`, so it stayed down, and the memory backend was unavailable until someone noticed.

## The code

`daemon_embed_manager.py:398-413` — the victim is selected purely by "who holds the port":

```python
# Unhealthy after grace window — attempt to reclaim by killing the listener.
if not self._is_port_in_use(port):
    return True

pid = self._find_pid_on_port(port)
if pid is None:
    logger.warning(f"Port {port} is in use by another process")
    return False

logger.info(f"Clearing unhealthy process on port {port} (PID {pid})")
if self._kill_process(pid):
```

`_kill_process` (`:321-335`) then SIGTERMs and polls to death:

```python
os.kill(pid, signal.SIGTERM)
for _ in range(50):
    time.sleep(0.1)
    try:
        os.kill(pid, 0)
    except OSError:
        return True
```

The docstring at `:371-386` states the intent plainly — and names the problem:

> Port occupied but /health is unreachable, non-200, or does not return Hindsight's initialized health payload → treat as a stale daemon (**or foreign process**) and attempt to reclaim by killing the PID listening on the port.

"Or foreign process" is doing a lot of work there. A foreign process is precisely the case that should **not** be killed.

Note this is not limited to `_clear_port`. The same `_find_pid_on_port` → `_kill_process` pattern, with no ownership check, is also used at `:912` (`stop_ui`) and `:965` (daemon stop).

## Why the health probe fails is a separate question

`_port_health_ok` requires `status == "healthy"` **and** `database == "connected"`. Any listener that isn't an initialized Hindsight daemon fails that check by definition — including a perfectly healthy *different* Hindsight deployment that simply isn't the one this manager started.

#3099 documents a way the probe returns a false negative for the manager's *own* daemon (event loop blocked by a slow LLM call). That issue is about the trigger; this one is about the blast radius. They compound: #3099 makes a false negative more likely, and the missing ownership check means a false negative can take out an unrelated process. Fixing #3099 alone still leaves an unverified kill on any genuinely foreign listener.

Related: #3517 notes `_find_pid_on_port` breaks without `lsof`. Separately, it takes `result.stdout.strip().split()[0]` — the *first* PID — which is arbitrary when more than one socket is bound to the port.

## Suggested fix

Verify the listener belongs to this manager before signalling it. Options, roughly in order of robustness:

1. **Ownership handshake** — have the manager record the daemon's PID (pidfile or instance metadata it wrote itself) and only signal a PID it started.
2. **Identity on `/health`** — include an instance/profile identifier and, ideally, a token the manager generated; only reclaim when the listener identifies as *this* manager's daemon. This also distinguishes "another Hindsight deployment" from "my stale daemon".
3. **Command-line check** — inspect the candidate PID's cmdline and refuse to signal anything that isn't recognizably a Hindsight daemon. Cheapest to implement, and enough to stop the foreign-process case.

If the listener can't be positively identified as this manager's daemon, the safe behaviour is the one already used when `pid is None`: log and return `False`, letting startup fail with "port in use" rather than killing an unknown process. A failed start is recoverable; killing someone else's service is not.

As prior art for option 3, a downstream consumer of this library solves the identical "reclaim my port from a stale sidecar" problem by verifying the candidate's command line before signalling, and refusing to signal anything it cannot confirm is its own process. The reclaim logic is otherwise the same shape as `_clear_port`; the only difference is the guard.

## Impact

Any host where a Hindsight embedded daemon shares a port with another service. The failure is silent from the embed side (it logs "Clearing unhealthy process ... Stale process stopped" and proceeds), so the operator sees an unrelated service die with no indication of what killed it. In our case it took an audit trail (`auditd` signal auditing) to attribute the SIGTERM, because the kill is an in-process `os.kill()` — no subprocess is executed, so there is no command to find in process accounting.

Beverly621:
Closes #3520

## Summary

Prevent `hindsight-embed` from terminating an arbitrary process merely because it listens on the configured daemon or UI port.

The embedded API and control-plane server now write PID ownership receipts from the actual listener processes. `_clear_port()`, `stop()`, and `stop_ui()` only signal a listener when its PID matches the receipt for that profile. Missing or mismatched ownership fails safely with the port left untouched.

## Why

The previous unhealthy-port recovery path selected a victim solely through `_find_pid_on_port()`. A failed health probe could therefore SIGTERM an unrelated service, including a separately managed `hindsight-api` deployment.

This keeps stale-process recovery for daemons started by the embed manager while changing the ambiguous case from destructive recovery to a recoverable “port in use” failure.

## Implementation

- pass a profile-specific PID receipt path to the embedded API daemon
- write and conditionally clean up the receipt from the actual API process after daemonization
- write the actual Next.js server PID from control-plane instrumentation
- require listener PID and receipt PID to match before every termination path
- retain the UI receipt if termination fails, so a later stop can retry safely
- cover owned, foreign, missing, and failed-termination cases with regression tests

## Validation

```text
53 passed: hindsight-embed/tests/test_daemon_client.py + test_embed_manager.py
12 relevant profile daemon tests passed (3 unrelated local-ML command-selection tests require optional sentence-transformers)
ruff check: passed for all changed Python source files
ruff format --check: passed for all changed Python files
python compileall: passed
git diff --check: passed
```

The repository-wide lint hook could not complete in this checkout because its absolute directory name contains `:`, which `uv` rejects as a path separator. The changed Python files were checked directly with the same Ruff rules; CI runs from a normal checkout path.

## AI assistance

AI model: GPT-5.6Sol  
Coding agent: Codex  
The contributor reviewed and accepted the changes.


koriyoshi2041 (commented):
The fail-closed behavior for missing/mismatched receipts is a good improvement. One ownership edge remains: a receipt containing only a PID can outlive the daemon, and after PID reuse it can authorize signaling an unrelated listener that happens to acquire both that PID and the configured port. The cleanup handler covers orderly exit, but SIGKILL/crash/power loss leaves the receipt behind. Could the receipt also include a process birth marker and have _owned_pid_on_port compare it (for example Linux /proc/<pid>/stat starttime, with a platform fallback), or could startup invalidate stale receipts before treating them as authority? A focused regression could write a matching PID with a mismatched birth marker and assert _kill_process is not called. I checked the current stop and unhealthy-port paths; both currently rely only on listener_pid == owned_pid.

Beverly621:
Thanks for catching this ownership edge. I addressed it in `498a78ad` and the follow-up `7b733572`.

Ownership receipts are now versioned JSON records containing both the PID and a process birth marker. Linux uses `/proc/<pid>/stat` starttime, Windows uses the process creation date from CIM, and other POSIX platforms fall back to `ps lstart`. `_owned_pid_on_port()` now requires both the listener PID and the current birth marker to match the receipt; if either is missing or mismatched, it fails closed.

Both the unhealthy-port cleanup path and `stop()` go through this shared ownership check. I added focused regressions for a matching PID with a mismatched birth marker in both paths, asserting that `_kill_process` is not called.

I also replaced the receipt tuple with a frozen dataclass to comply with the project's typed-data conventions.

Validation completed:

- `hindsight-embed`: 56 tests passed
- daemon receipt lifecycle: 5 tests passed
- control-plane instrumentation receipt test: passed
- Ruff, formatting, ty, Prettier, ESLint, and `git diff --check`: passed

The latest CI run is currently awaiting maintainer approval. Please take another look when convenient.

koriyoshi2041 (approved):
Re-reviewed at `7b733572`. The receipt now binds both the listener PID and an OS process-birth marker, and both `_clear_port()` and `stop()` fail closed when either value differs. The Linux `/proc/<pid>/stat` parsing reads field 22 correctly even when `comm` contains spaces, and the receipt cleanup avoids removing a replacement record.

Local validation passed: 56/56 embed manager/client tests, 5/5 daemon receipt lifecycle tests, the control-plane instrumentation test, and `git diff --check`. The focused PID-reuse regressions cover both termination paths and assert `_kill_process` is not called. This resolves my earlier ownership concern.

nicoloboschi:
Thanks for this — refusing to kill a process just because it holds the port is the right instinct, and #3520 was fixed as part of #3585.

`main` gets there by a different mechanism: `_owned_pid_on_port(port, PROCESS_MARKERS, ...)` verifies the listener is actually ours before signalling it, rather than writing PID ownership receipts. Same guarantee — a foreign process on the port is left untouched.

Closing as already fixed.

Beverly621:
👍🏻Thanks for the clarification and for pointing me to #3585. The PROCESS_MARKERS approach makes sense, and I agree that it provides the same fail-closed guarantee without introducing a separate ownership receipt. I appreciate the thoughtful reviews and the explanation of the final design. Glad to see #3520 resolved!

Issue #3411, opened 2026-08-12
Mental models with refresh_after_consolidation are not refreshed when consolidation spans multiple rounds


nicoloboschi:
## Summary

Mental models configured with `trigger.refresh_after_consolidation: true` are silently **not** refreshed when a bank's consolidation backlog spans more than one round. Only the models associated with the *final* consolidation round get refreshed; everything consolidated in the preceding rounds is dropped and the models stay stale indefinitely.

This makes auto-refresh effectively non-functional for any bank that ingests documents in bulk.

## Environment

- Bank with ~110 mental models, one per entity, each tagged `entity:<id>` and each with:
  ```json
  {"mode": "full", "refresh_after_consolidation": true, "exclude_mental_models": true, "tags_match": "all_strict"}
  ```
- `consolidation_max_memories_per_round: 100` (default)
- `enable_auto_consolidation: true`, `enable_observations: true`
- ~3,150 observations / ~6,700 facts in the bank

## Reproduction

1. Create a bank with N entity-scoped mental models (N ≈ 110), each tagged `entity:<id>`, all with `refresh_after_consolidation: true`.
2. Bulk-ingest one document per entity (112 documents over ~6 minutes), enough to produce a consolidation backlog larger than `consolidation_max_memories_per_round`.
3. Wait for consolidation to fully drain.
4. List mental models and compare `last_refreshed_at` against the bank's `last_consolidated_at`.

## Observed

From `GET /v1/{ns}/banks/{bank}/operations`:

| Time | Event |
|---|---|
| T+00:00 – T+06:14 | 112 documents ingested |
| T+00:49 – T+28:10 | **36 chained `consolidation` rounds** drain the backlog |
| T+28:10 | Final consolidation completes → **2** `refresh_mental_model` tasks enqueued |

- Rounds 1–35 enqueued **zero** `refresh_mental_model` operations.
- The final round enqueued exactly 2.
- The two refreshed models correspond precisely to the **last two documents ingested** (ingested at T+06:08 and T+06:14). Every other entity's document was consolidated in an earlier round and its model was never refreshed.
- Net result: **103 of 110** mental models have `last_refreshed_at < last_consolidated_at`.

All operations report `status: completed` — there are no failed or retried tasks, and `failed_consolidation` is 0. The refreshes are never enqueued in the first place, so nothing surfaces as an error.

[...]

Issue #3210, opened 2026-08-06
maintenance: scheduled mental-model refresh enqueues duplicates when completions stall (no existing-op check, no single-flight)


nicoloboschi:
## Problem

The scheduled mental-model refresh in `maintenance.py` decides a model is *due* from `refresh_cron` vs `last_refreshed_at`, re-checks staleness, and then calls `submit_async_refresh_mental_model` — but it **never checks whether a pending/processing `refresh_mental_model` operation for that (bank, model) already exists**.

`last_refreshed_at` only advances when a refresh **completes**. Whenever completion stalls fleet-wide (worker crash loop, slow LLM provider, saturated queue), every subsequent maintenance tick re-judges the same models due-and-stale and enqueues **another full wave** of refresh ops. The queue amplifies in proportion to the outage duration.

Compounding it, the maintenance loop runs **independently in every worker process** — there is no leader election or single-flight, so N workers = N redundant schedulers making the same due-judgment each interval. (Startup logs also show the same scan running twice within seconds in one process — worth checking whether the loop is registered twice per engine instance.)

Observed in a multi-tenant deployment during a worker outage: ~8× duplication — thousands of queued refresh ops for a few hundred distinct models. The duplicates are not free even though execution re-checks staleness: they occupy claim slots, inflate queue-depth metrics (autoscaler input), and delay unrelated tenants' work behind them.

## Proposed fix

1. **Dedup at enqueue**: before submitting, skip if a pending or processing `refresh_mental_model` op already exists for the same (bank_id, mental_model_id) — one indexed `EXISTS` against `async_operations` (payload or dedicated columns).
2. **Single-flight the scheduler across workers**: an atomic interval-bucketed claim row (`INSERT … ON CONFLICT DO NOTHING` into a maintenance-run marker table; the winner runs the scan, others skip). Advisory locks are not an option per project policy (poolers).
3. Regression test: block completion (mock executor), tick the scheduler twice, assert exactly one queued op per due model.

Either fix alone bounds the damage; both together make refresh scheduling idempotent and O(models) regardless of fleet size or outage duration.

PR #3528, opened 2026-08-16
fix(engine): make the in-flight refresh dedupe work on Oracle (#3487)

### LINKED ISSUE #3487: refresh_mental_model submit dedupe misses pending duplicates — slow-draining banks accumulate ~45 copies per model
## Summary

`submit_async_refresh_mental_model` with `skip_if_in_flight=True` does not prevent duplicate **pending** refresh operations from accumulating. On a bank whose refresh ops drain slowly (e.g. constrained worker capacity), every consolidation round's refresh flush enqueues a fresh copy per mental model, unbounded.

## Observed (multi-tenant deployment, one slow-draining tenant)

- ~510 `refresh_mental_model` ops enqueued per hour by routine consolidation chains, ~0 draining.
- Pile reached **12,461 pending ops covering only 259 distinct mental models — ~45 duplicates per model** — and grew monotonically until manually deduplicated (`DELETE` keeping newest per `(bank_id, mental_model_id)`).
- Each duplicate is a full recall+LLM refresh when it eventually runs, so the duplication multiplies both drain time and metered cost by ~45× for zero freshness benefit (the model is refreshed to the same state each time).

## Expected

At most one *pending* refresh per `(bank_id, mental_model_id)`: a submit that finds an existing pending (not just claimed/processing) op for the same model should no-op, or fold into it — mirroring what #3416's dedupe does for consolidation re-queues.

## Suggested fix

Extend the in-flight check to cover `status='pending'` rows for the same model (or add a partial unique index on `(bank_id, (task_payload->>'mental_model_id')) WHERE status='pending' AND operation_type='refresh_mental_model'` with `ON CONFLICT DO NOTHING` semantics at submit). The index form is race-proof under concurrent flushes, which is exactly the multi-chain scenario #3416 describes.

chethanuk:
## Problem

The sub-bank dedupe in `_submit_async_operation` binds its JSON key:

```sql
AND task_payload->>$7 = $8
```

Oracle never sees that as JSON. `_JSON_ARROW_TEXT_RE` (`db/oracle.py:82`) only matches a **quoted literal** key, so `->>$7` survives the rewrite as a raw PG operator and the statement raises. Both callers swallow it, so nothing surfaces.

```mermaid
flowchart LR
    C["consolidator.py:1944<br/>skip_if_in_flight=True"] --> S["_submit_async_operation"]
    M["maintenance.py:505<br/>cron refresh"] --> S
    S --> D{"dedupe_in_flight<br/>_payload_key"}
    D -->|None| U["INSERT ... VALUES<br/>fine on both dialects"]
    D -->|set| P{"backend"}
    P -->|postgresql| OK["INSERT ... WHERE NOT EXISTS<br/>fine"]
    P -->|oracle| X["task_payload->>$7 raises,<br/>caught at consolidator.py:1955"]
    X --> L["'Failed to trigger refresh',<br/>model never refreshed"]
```

The comment on this branch claimed it was PG-only, and that was true when #3210 wrote it: the maintenance loop is PG-gated at `maintenance.py:96`. Then #3411 added the post-consolidation flush, and consolidation does run on Oracle. Every consolidation-triggered mental-model refresh on Oracle has been dropped silently since.

Fixes vectorize-io/hindsight#3487

## Fix

Inline the key so the rewrite fires:

```
AND task_payload->>'mental_model_id' = $7
  → JSON_VALUE(task_payload, '$.mental_model_id') = :7
```

It is an internal constant, but the code now rejects anything that is not a plain identifier rather than trusting the caller.

Oracle still needs its own write path. `INSERT ... SELECT` requires a `FROM DUAL` there and cannot carry a `RETURNING` clause at all, so the check cannot ride inside the INSERT the way it does on PG. Splitting them is safe because `serialize` already holds `FOR NO KEY UPDATE` (`FOR UPDATE` on Oracle) on the bank row for the whole transaction, so no concurrent submit for the same bank can interleave. PG keeps the single-statement form, where the check holds even without that lock.

No behaviour change on PostgreSQL.

## Note on the issue

#3487 reports a PostgreSQL symptom that main already handles. The dedupe has covered `status='pending'` since #3210, the consolidation flush opted in via #3411, and `test_concurrent_scheduled_submits_queue_one_refresh` pins it. The reporter's deployment most likely predates those.

I still have this closing #3487, because the issue asks for "at most one pending refresh per `(bank_id, mental_model_id)`" and that guarantee did not hold on Oracle. Say the word if you would rather it stayed open for the reporter to confirm their version.

## Test evidence

| Check | Result |
|---|---|
| `tests/test_db_abstraction.py` | 125 passed |
| `tests/test_mental_model_scheduled_refresh.py` | 8 passed |
| `ruff check` / `ruff format` / `ty check` | clean |

One new case, `test_jsonb_arrow_rewrite_needs_a_literal_key`, asserts the emitted predicate rewrites to `JSON_VALUE` with a literal path, and that a bound `->>$N` does not rewrite. The second assertion pins the bug, not just the fix.

It calls the real rewriter instead of mocking a connection. A mock would have passed on the broken SQL too, since no Oracle parser is involved either way.

## Limitation

Not introduced here and not addressed: on Oracle, `_convert_arg` (`db/oracle.py:107`) coerces any dash-formatted UUID string bind to `RAW(16)`, so a client-supplied UUID-shaped `mental_model_id` binds as bytes and will not match the `VARCHAR2` result of `JSON_VALUE`. Server-generated ids are `mm-<hex>` and unaffected, and `bank_id` in the same query carries the identical exposure.


nicoloboschi:
superseded by #3550

PR #3529, opened 2026-08-16
fix(embed): use family-agnostic localhost probe and port helpers (#3527)

### LINKED ISSUE #3527: hindsight-embed: is_ui_running() hardcodes 127.0.0.1 so ui start/status always fail when the UI binds [::1] (localhost)
**Bug Description**

`hindsight-embed ui start` (and `ui status`) consistently report the Control Plane UI as *not running / failed to start*, even though the UI process is up and serving requests. The root cause is in `hindsight_embed/daemon_embed_manager.py` → `is_ui_running()`: it health-checks the UI on a hardcoded IPv4 loopback address (`http://127.0.0.1:{ui_port}/api/health`) **regardless of the hostname the UI actually bound to**.

When the Control Plane starts with `--hostname localhost` (the documented workaround for #1926) — and in many environments even with `0.0.0.0` — Next.js binds to the **IPv6** loopback `[::1]` only, so `127.0.0.1:{port}` refuses the connection, the health check throws, and `is_ui_running()` returns `False`. The UI is genuinely reachable via `http://localhost:{port}` (which resolves to `::1`), but every CLI command that depends on the health check reports failure.

This also means `ui start` **always times out after 30s** ("UI failed to start (timeout)") even on success, `ui status` reports "UI is not running" while it is, and `hindsight-embed ui stop` can leave an orphaned UI process behind.

**Steps to Reproduce**

1. Start the daemon: `hindsight-embed -p hermes daemon start` (Linux)
2. Start the UI with the documented workaround from #1926: `hindsight-embed -p hermes ui start --hostname localhost`
3. Observe: CLI prints `⏳ Waiting for UI... (29s elapsed)` then `✗ UI Failed (Timeout)`
4. Meanwhile the UI is actually listening and healthy:
   ```
   $ ss -tlnp | grep 19177
   LISTEN 0 511 [::1]:19177 [::]:* users:(("next-server (v1",pid=16076,fd=21))
   $ curl -s http://localhost:19177/api/health
   {"status":"ok","service":"hindsight-control-plane","dataplane":{"status":"connected",...}}
   ```
5. `hindsight-embed -p hermes ui status` → `UI is not running` (wrong)
6. `curl -s http://127.0.0.1:19177/api/health` → connection refused (the CLI's probe address)

**Expected Behavior**

`ui start` should detect the UI as running and return success, and `ui status` should report the truth. The health check should query the **same hostname the UI was started with** (or try both `::1` and `127.0.0.1`, or use `localhost`), rather than hardcoding `127.0.0.1`.

**Actual Behavior**

CLI reports failure/timeout while the Control Plane is fully functional on `[::1]:19177`. `ui status` reports "not running". The only way to manage the UI in this state is to bypass the CLI and spawn `npx @vectorize-io/hindsight-control-plane` directly.

**Version**

- hindsight-embed 0.8.6 (PyPI) — also confirmed present in `main` and in v0.9.1 (`hindsight-embed/hindsight_embed/daemon_embed_manager.py`):
  ```python
  def is_ui_running(self, profile: str, ui_port: int | None = None) -> bool:
      # Always health-check on 127.0.0.1 regardless of bind hostname
      ui_url = self.get_ui_url(profile, ui_port, hostname="127.0.0.1")
  ```
- hindsight-control-plane 0.8.6 / 0.9.1 (npx), Next.js 16.2.11
- OS: Linux (CachyOS), Python 3.11

**LLM Provider**

Other

chethanuk:
## Description

`hindsight-embed ui start` and `ui status` report failure when the control plane binds IPv6 (`[::1]`) under `--hostname localhost` or in dual-stack environments. The UI health probe previously hardcoded IPv4 `127.0.0.1`, which is refused when Next.js binds IPv6-only `::1`.

## Root Cause

- `daemon_embed_manager.py:is_ui_running` hardcoded `127.0.0.1` for `/api/health`.
- `_is_port_in_use` used an `AF_INET`-only socket, treating live IPv6 listeners as free ports.
- `_find_pid_on_port` filtered Windows `netstat` output for `127.0.0.1:{port}` substring.

## Fix

1. `get_ui_url`: Default host fallback to `localhost` instead of `0.0.0.0`.
2. `is_ui_running`: Remove hardcoded `127.0.0.1` override; probe `localhost` so `httpx` resolves both IPv4 and IPv6 families via `getaddrinfo`.
3. `_is_port_in_use`: Use stdlib `socket.create_connection(("localhost", port), timeout=1)` for dual-stack support.
4. `_find_pid_on_port`: Replace substring check with bounds-checked netstat column parsing.
5. Printed URLs: Update CLI messages to `http://localhost:{port}`.

## Testing

Added `hindsight-embed/tests/test_ui_health_probe.py` covering:
- IPv6-only (`::1`) health probe response (failing before fix, passing now)
- IPv4-only (`127.0.0.1`) health probe response
- Closed port fail-fast (<5s)
- Windows netstat parsing for `[::1]:port` and `127.0.0.1:port`

All 169 unit & integration tests pass cleanly:
```
============================= 169 passed in 22.22s =============================
```

Fixes #3527

### DIFF
--- a/hindsight-embed/hindsight_embed/cli.py
+++ b/hindsight-embed/hindsight_embed/cli.py
@@ -785,7 +785,7 @@ def do_ui(args, config: dict, logger):
             status_text = Text()
             status_text.append("UI is running\n\n", style="green bold")
             status_text.append("  URL: ", style="dim")
-            status_text.append(f"http://127.0.0.1:{effective_port}\n", style="cyan")
+            status_text.append(f"http://localhost:{effective_port}\n", style="cyan")
             status_text.append("  Logs: ", style="dim")
             status_text.append(f"{paths.ui_log}", style="")
 
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -320,9 +320,11 @@ class DaemonEmbedManager(EmbedManager):
         """Check if a port is in use using a socket connection (cross-platform)."""
         import socket
 
-        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
-            sock.settimeout(1)
-            return sock.connect_ex(("127.0.0.1", port)) == 0
+        try:
+            with socket.create_connection(("localhost", port), timeout=1):
+                return True
+        except OSError:
+            return False
 
     @staticmethod
     def _find_pid_on_port(port: int) -> int | None:
@@ -342,8 +344,9 @@ class DaemonEmbedManager(EmbedManager):
                 )
                 if result.returncode == 0:
                     for line in result.stdout.splitlines():
-                        if f"127.0.0.1:{port}" in line and "LISTENING" in line:
-                            return int(line.strip().split()[-1])
+                        parts = line.split()
+                        if len(parts) >= 5 and parts[3] == "LISTENING" and parts[1].rsplit(":", 1)[-1] == str(port):
+                            return int(parts[4])
             else:
                 # Use lsof on macOS/Linux
                 result = subprocess.run(
@@ -779,13 +782,12 @@ class DaemonEmbedManager(EmbedManager):
         if ui_port is None:
             paths = self._profile_manager.resolve_profile_paths(profile)
             ui_port = paths.ui_port
-        host = hostname or "0.0.0.0"
+        host = hostname or "localhost"
         return f"http://{host}:{ui_port}"
 
     def is_ui_running(self, profile: str, ui_port: int | None = None) -> bool:
         """Check if the UI is running and responsive."""
-        # Always health-check on 127.0.0.1 regardless of bind hostname
-        ui_url = self.get_ui_url(profile, ui_port, hostname="127.0.0.1")
+        ui_url = self.get_ui_url(profile, ui_port)
         try:
             with httpx.Client(timeout=2) as client:
                 response = client.get(f"{ui_url}/api/health")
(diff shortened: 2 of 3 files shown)


hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -16129,51 +16129,75 @@ async def _submit_async_operation(
                         *insert_args,
                     )
                 else:
-                    # Sub-bank dedup: the INSERT itself only materialises a row when no
-                    # operation of this type is already queued or running for the same
-                    # payload subject (e.g. one mental model), so the check cannot be
-                    # separated from the write (#3210).
+                    # Sub-bank dedup: skip the INSERT when an operation of this type is
```
nicoloboschi: unrelated change, please remove

nicoloboschi (changes requested):

hindsight-api-slim/hindsight_api/engine/memory_engine.py
chethanuk: updated

nicoloboschi:
Superseded by #3585 (merged as d98990b46), which fixed #3527 along with #3099, #3100, #3517 and #3520 in the same two files.

That change covers the same ground as this PR — `is_ui_running`, `_is_port_in_use` and the Windows `netstat` parse now probe both loopback families (`127.0.0.1` and `::1`), and user-facing URLs print `localhost` — plus PID-ownership checks before signalling a listener and an `ss` fallback for hosts without `lsof`.

Thanks for the report and the fix here — closing as already resolved on main.

Issue #1926, opened 2026-06-02
Control plane: every page route 307-redirect-loops under Next 16 (next-intl rewrite leaked as redirect)


nicoloboschi:
## Summary

Every **page** route in the control plane returns a `307 Temporary Redirect` to **itself**, producing an infinite redirect loop (`ERR_TOO_MANY_REDIRECTS` in a browser). API routes (`/api/*`) are unaffected and work fine. This makes the CP UI unusable when run locally.

## Environment

- `hindsight-control-plane`, Next.js **16.2.6**, next-intl **4.12.0** (both as pinned in `package.json` / lockfile)
- Reproduced in: `next dev` (Turbopack **and** non-Turbopack), `next start`, and the **standalone production server** (`node standalone/server.js`)
- Reproduced via both `http://127.0.0.1:<port>` and `http://localhost:<port>`
- Authentication **disabled** (no `HINDSIGHT_CP_ACCESS_KEY`; logs show `No API key configured (public access)`)

## Symptom / evidence

Raw response for any page (e.g. `GET /login`):

```
HTTP/1.1 307 Temporary Redirect
x-middleware-rewrite: http://localhost:<port>/en/login    ← next-intl requests an INTERNAL rewrite
location: /login                                           ← …but it is emitted as a REDIRECT to the original path
set-cookie: NEXT_LOCALE=en; Path=/; SameSite=lax
link: <…/login>; rel="alternate"; hreflang="en", …(all locales)…
```

next-intl's middleware issues a rewrite (`/login` → `/en/login`) for the default locale (`localePrefix: "as-needed"`), but Next 16 surfaces it to the client as a **307 redirect to the original path** instead of serving the rewrite internally → loop. The `The "middleware" file convention is deprecated. Please use "proxy" instead` warning is logged on every start.

`/`, `/banks/[bankId]`, `/login` all loop identically.

## Ruled out

- **Auth** — loops with auth fully disabled; the `middleware.ts` auth block is skipped entirely when no access key is set, and control falls straight through to `intlMiddleware`.
- **Turbopack** — loops with and without `--turbopack`.
- **Locale detection** — loops with `localeDetection: false`.
- **Naive `middleware.ts` → `proxy.ts` rename** (function renamed to `proxy`) — still loops (likely needs the correct Next 16 export/signature, not just a rename).
[...]

PR #3530, opened 2026-08-16
fix(embed): probe both loopback families in UI health check

### LINKED ISSUE #3527: hindsight-embed: is_ui_running() hardcodes 127.0.0.1 so ui start/status always fail when the UI binds [::1] (localhost)

NovaLux12:
Closes #3527

## Problem

`hindsight-embed ui start` / `ui status` report the Control Plane UI as not
running even when it is up: `is_ui_running()` health-checks a hardcoded
`http://127.0.0.1:<port>/api/health` URL, but when the UI binds only the IPv6
loopback (e.g. Next.js started with `--hostname localhost`, the documented
#1926 workaround) the probe is refused. Every command that depends on the
check — `ui start`'s ready-wait, `ui status` — reports failure, and `ui start`
times out after 30s while the UI is actually serving on `[::1]`.

## Change

- `DaemonEmbedManager.is_ui_running()` now probes both loopback families,
  `127.0.0.1` then `::1`, returning `True` on the first healthy response.
- `get_ui_url()` brackets IPv6 literals (`::1` → `[::1]`) so the probe URL —
  and any display of an IPv6 hostname — is a valid URL.

## Verification

- New regression tests in `hindsight-embed/tests/test_embed_manager.py`:
  - UI bound to `[::1]` only is detected as running (asserts the exact probe
    URLs, including the bracketed IPv6 form);
  - IPv4-loopback success short-circuits without probing IPv6;
  - both loopback probes refused → `False`;
  - `get_ui_url("::1")` → `http://[::1]:<port>`.
- `uv run pytest tests/` in `hindsight-embed`: **162 passed**; the 3 failures
  (`test_profile_daemon_config.py`) are pre-existing and environmental —
  they fail identically on a clean `main` checkout because the venv lacks
  `sentence_transformers`, which flips `_find_api_command` to the `uvx`
  fallback. Unrelated to this change.
- `ruff check` and `ruff format --check` pass on both changed files.

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -780,18 +780,31 @@ class DaemonEmbedManager(EmbedManager):
             paths = self._profile_manager.resolve_profile_paths(profile)
             ui_port = paths.ui_port
         host = hostname or "0.0.0.0"
+        # IPv6 literals must be bracketed in the URL authority component.
+        if ":" in host:
+            host = f"[{host}]"
         return f"http://{host}:{ui_port}"
 
     def is_ui_running(self, profile: str, ui_port: int | None = None) -> bool:
-        """Check if the UI is running and responsive."""
-        # Always health-check on 127.0.0.1 regardless of bind hostname
-        ui_url = self.get_ui_url(profile, ui_port, hostname="127.0.0.1")
-        try:
-            with httpx.Client(timeout=2) as client:
-                response = client.get(f"{ui_url}/api/health")
-                return response.status_code == 200
-        except Exception:
-            return False
+        """Check if the UI is running and responsive.
+
+        Probes both loopback families: the Control Plane UI can bind only the
+        IPv6 loopback (e.g. Next.js started with ``--hostname localhost``, the
+        documented workaround for #1926), so a 127.0.0.1-only probe reports it
+        as down even while it is serving (issue #3527).
+        """
+        if ui_port is None:
+            paths = self._profile_manager.resolve_profile_paths(profile)
+            ui_port = paths.ui_port
+        for host in ("127.0.0.1", "::1"):
+            try:
+                with httpx.Client(timeout=2) as client:
+                    response = client.get(f"{self.get_ui_url(profile, ui_port, hostname=host)}/api/health")
+                    if response.status_code == 200:
+                        return True
+            except Exception:
+                continue
+        return False
 
     @staticmethod
     def _ui_port_file(paths) -> Path:
--- a/hindsight-embed/tests/test_embed_manager.py
+++ b/hindsight-embed/tests/test_embed_manager.py
@@ -10,6 +10,27 @@ def _mock_sentence_transformers_present(monkeypatch):
     monkeypatch.setattr("hindsight_embed.daemon_embed_manager.find_spec", lambda name: object())
 
 
+class _FakeUiHealthClient:
+    """httpx.Client stand-in that records requested URLs and returns a canned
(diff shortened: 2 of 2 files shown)


NovaLux12:
Duplicate of #3529 (opened 4 minutes earlier by @chethanuk) — closing mine to keep the issue to a single PR. Mine took a narrower approach (explicit 127.0.0.1 then [::1] probes with bracketed IPv6 URLs in is_ui_running, +98/-9 over 2 files, 162 embed tests passing); #3529 is broader (family-agnostic localhost probe, netstat parsing, _is_port_in_use). Happy to defer to #3529 or help consolidate if the maintainers prefer either variant.

PR #3531, opened 2026-08-16
fix(engine): drop null metadata values at retain and recall (#3209)

### LINKED ISSUE #3209: recall: MemoryFact rejects stored metadata containing null values (retain accepts, recall fails)
## Problem

`MemoryFact.metadata` is typed `dict[str, str] | None`. Retain accepts arbitrary user-supplied JSON metadata and stores it, but if any value is `null` (e.g. `{"ocr_engine": null}`), every subsequent recall/search that returns the affected rows fails Pydantic validation:

```
Failed to search memories (ValidationError): validation errors for MemoryFact
metadata.ocr_engine
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

Consequences: the rows become unreadable through the search path, and background operations that recall internally (consolidation, mental-model refresh) fail repeatedly for that bank — an ingest-time input shape silently poisons later reads. Observed in a multi-tenant deployment with client-submitted metadata.

## Root cause

Validation asymmetry: the retain path does not enforce `dict[str, str]` on metadata values, but the read-path response model does.

## Proposed fix

Pick one consistency direction, applied at both ends:

- **Tolerant read (preferred, fixes existing data):** widen the model to `dict[str, str | None]` (or drop null-valued keys in the existing `parse_metadata` before-validator). No data migration needed; already-stored rows become readable again.
- Optionally also normalize at retain time (drop or stringify nulls) so stored data is canonical going forward.

A regression test should retain a memory with `{"key": null}` metadata and assert recall returns it.

NovaLux12:
Closes #3209

## Problem

`MemoryFact.metadata` is typed `dict[str, str] | None`, but retain accepts
arbitrary user-supplied JSON metadata. A null value (e.g.
`{"ocr_engine": null}`) was stored verbatim, and every read path that returns
`MemoryFact` — recall, consolidation, mental-model refresh — then failed
Pydantic validation (`Input should be a valid string`), permanently wedging
the affected bank.

## Change

Normalize on both ends, per the plan in the issue:

- **Write path** — `RetainContent.__post_init__` drops null-valued metadata
  keys at construction, so facts extracted from it (`metadata=content.metadata`
  in `fact_extraction`) are stored canonically. Single chokepoint: covers the
  main retain path, delta retain, recovery, and transfer import. Document
  metadata in `retain_params` is untouched (built from the raw request dicts).
- **Read path** — `MemoryFact.parse_metadata` drops null-valued keys instead
  of coercing them to the string `"None"`, so already-stored legacy rows
  become readable again with no data migration. Existing string coercion for
  other non-string values is preserved.

## Verification

- New tests:
  - `tests/test_response_models.py` (read path): null metadata value omitted,
    null inside a JSONB-string is omitted, all-null becomes `{}`.
  - `tests/test_retain_metadata_normalization.py` (write path): `RetainContent`
    drops nulls, keeps non-string values as-is, and `_build_contents` (the API
    ingestion funnel) normalizes a null-metadata request.
- `uv run pytest tests/test_response_models.py tests/test_retain_metadata_normalization.py tests/test_retain_orchestrator_mapping.py -q` → **27 passed**.
- Consumer slice (`test_large_document_replacement.py`,
  `test_retain_entity_prune_race.py`, `test_validation_result_enrichment.py`,
  `test_response_exclude_none.py`): **16 passed**; the 9 errors are all the
  same pre-existing environmental `ImportError: pg0-embedded is required for
  embedded PostgreSQL` (optional `embedded-db` extra not installed in this
  venv), unrelated to this change.
- `ruff check`, `ruff format --check`, and `ty check` pass on all changed files.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/metadata_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/metadata_utils.py
@@ -1,0 +1,35 @@
+"""Canonical handling of user-supplied memory metadata (issue #3209).
+
+Metadata is accepted as arbitrary JSON at ingest (file retain, the MCP tools and
+direct engine calls all take ``dict[str, Any]``) but is stored in a JSONB column
+and read back through models that declare ``dict[str, str]``. A JSON ``null``
+value therefore sailed through the write path and then failed validation on
+every read that returned the affected rows.
+
+Both ends normalize through the helpers here so the rule lives in one place:
+
+* ``drop_null_values`` — the write contract. Null-valued keys are dropped;
+  every other value is stored as given (the read contract stringifies).
+* ``as_string_metadata`` — the read contract. Null-valued keys are dropped and
+  the rest are coerced to strings, so rows written before this normalization
+  existed stay readable without a data migration.
+"""
+
+from collections.abc import Mapping
+from typing import Any
+
+
+def drop_null_values(metadata: Mapping[str, Any] | None) -> dict[str, Any]:
+    """Return ``metadata`` without its null-valued keys (empty dict for no metadata)."""
+    if not metadata:
+        return {}
+    return {k: v for k, v in metadata.items() if v is not None}
+
+
+def as_string_metadata(metadata: Mapping[str, Any] | None) -> dict[str, str]:
+    """Coerce a stored metadata bag to the ``dict[str, str]`` read contract.
+
+    Drops null-valued keys and stringifies the rest (JSONB round-trips integers
+    as integers, e.g. ``{"original_id": 348}``).
+    """
+    return {str(k): str(v) for k, v in drop_null_values(metadata).items()}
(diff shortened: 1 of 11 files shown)


chaoyuan-guo:
One write-side gap remains: delta updates bypass RetainContent normalization.
Both _delta_metadata_only and the partial-delta survivor sync pass raw retain_params.metadata to update_memory_units_metadata_and_tags. As a result, new or changed facts omit null-valued keys, while existing SQL-backed memory_units can receive them during re-retain. MemoryFact filters them on read, so recall hides the inconsistency.
Could these two paths use normalized RetainContent.metadata, while documents.retain_params.metadata continues to preserve the raw input?
The current _build_contents test only checks construction. Regression coverage should exercise:
- unchanged content, reaching _delta_metadata_only;
- a partial delta with unchanged survivors.

In both cases, documents.retain_params.metadata should preserve the null value, while memory_units.metadata should omit the null-valued key.

PR #3533, opened 2026-08-16
feat(mental-models): expose refresh operation model id and list total

benfrank241:
Two small observability improvements around mental models. Independent of each other; no schema changes.

## 1. `mental_model_id` on operation records
`refresh_mental_model` operations return `document_id: null` and no model identifier, so a client can't tell which mental model an operation refreshed (diagnosing refresh behaviour from the operations log is impossible). The id already exists in the operation row's `result_metadata` — this surfaces it as a first-class `mental_model_id` field on `OperationResponse` and `OperationStatusResponse` (null for other task types).

## 2. `total` on `GET /banks/{bank}/mental-models`
The list envelope was `{"items": [...]}` only and silently capped at `limit` (default 100), so a bank with more models than the page size looked fully-listed. Adds `total` (matching-row count, ignoring limit/offset) plus `limit`/`offset` echoes, backed by a new `count_mental_models()` engine method that applies the same tag filter as `list_mental_models`.

## Tests
- A `refresh_mental_model` operation surfaces the correct `mental_model_id`; non-refresh ops don't carry it.
- Listing with a smaller `limit` still reports the full `total`, including a tag-filtered total.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -2315,6 +2315,15 @@ class MentalModelListResponse(BaseModel):
     """Response model for listing mental models."""
 
     items: list[MentalModelResponse]
+    total: int = Field(
+        description="Total number of mental models matching the filter, ignoring limit/offset.",
+    )
+    limit: int = Field(
+        description="Maximum number of items returned in this page (echoes the request).",
+    )
+    offset: int = Field(
+        description="Number of items skipped before this page (echoes the request).",
+    )
 
 
 # =========================================================================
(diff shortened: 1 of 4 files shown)


koriyoshi2041 (changes requested):
The API shape and focused tests look consistent, but this head cannot merge yet because the checked-in generated surface is stale. `verify-generated-files` regenerated 10 files (347 insertions / 9 deletions), including the Go/Python/TypeScript response models plus both OpenAPI copies. Please run the generation commands printed by the job and commit those outputs.

The separate API shard failure looks unrelated to this diff: 1,940 tests passed and the only failure was a PostgreSQL deadlock in `test_import_bank_creates_per_bank_indexes`. I would treat that as a rerun candidate after the generated files are updated.

nicoloboschi:
superseded by #3548 including same fix for directives

PR #3537, opened 2026-08-17
fix(coding-agents): answer tools/list when disabled

altaywtf:
## Summary

Keep opted-out Hindsight MCP servers inert while making their tool-discovery contract valid. This prevents clients that probe `tools/list` from treating an intentionally disabled project as an MCP startup failure.

## Changed

- Build the MCP server through one registration helper.
- When the resolved project exposes no tools, advertise an empty tool capability and answer `tools/list` with `{ "tools": [] }`.
- Add an in-memory MCP client/server regression test for the zero-tool path.

## Risks

Low. Normal tool registration is unchanged; only the existing zero-tool path gains an explicit discovery response.

## Verification

- `npm test` — 530 tests passed.
- `npm run build` — passed.
- `git diff --check upstream/main...HEAD` — passed.
- Project code review — no findings.

## Complexity

The zero-tool path is constant-time. No API, persistence, or network behavior changes.

### DIFF
--- a/hindsight-integrations/coding-agents/src/mcp-server.test.ts
+++ b/hindsight-integrations/coding-agents/src/mcp-server.test.ts
@@ -1,5 +1,7 @@
 import { describe, expect, it, vi } from "vitest";
-import { selectTools } from "./mcp-server";
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
+import { buildMcpServer, selectTools } from "./mcp-server";
 import { resolveConfig } from "./core/config";
 import type { HindsightClient } from "./core/hindsight";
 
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Rebased onto current `main` and continued in #3626 — GitHub rejects pushes to this fork branch with a 403 despite `maintainer_can_modify: true`, so the commit could not be updated in place. Your commit and authorship are preserved there.

The conflicts with `main` were textual only: #3603 refactored `main()` onto `resolveHostMemory` and added `resolveHarness` after this branch was cut. `buildMcpServer` and its `InMemoryTransport` test needed no changes — full suite is 619 passing on the rebase.

Closing in favour of #3626.

altaywtf:
thanks @nicoloboschi 🙇 
think it's because I forked it to org repo rather than personal. 
will keep in mind next time. 👍

Issue #3499, opened 2026-08-14
Worker claim fabric pays ~12 statements per schema-visit (setup SETs + resets per acquire) — ~10x transaction amplification behind transaction-mode poolers


nicoloboschi:
## Summary

On a multi-tenant PostgreSQL deployment behind a transaction-mode pooler, the worker claim fabric now costs ~10× more database transactions than useful queries: each poll cycle acquires a pooled connection **per active schema**, and every acquire/release pays the full session ceremony. Database CPU shows a sustained multi-core burn that scales with `active schemas × workers × poll rate` while row throughput stays flat.

## Measured (12 workers, ~22 schemas flagged with pending work, poll_interval 500ms)

45-second `pg_stat_statements` window, calls aggregated by tenant-scrubbed shape:

| statement | calls | role |
|---|---|---|
| `SELECT DISTINCT bank_id FROM <schema>.async_operations WHERE operation_type…` | 25,350 | busy-banks pre-query, per schema-visit |
| claim `SELECT operation_id, operation_type, task_payload…` | 25,349 | the claim, paired 1:1 |
| `SET hnsw.ef_search` / `SET statement_timeout` / `SET pg_trgm.similarity_threshold` / `SET max_parallel_workers_per_gather` | ~20,880 **each** | pool `setup=` callback, runs on **every acquire** |
| `pg_advisory_unlock_all()` / `RESET ALL` / `CLOSE ALL` / `UNLISTEN *` | ~20,847 each | asyncpg release reset |
| `BEGIN` / `COMMIT` | 16,850 | explicit pairs |

Net: ~463 acquire/release cycles/s; **~12 statements per schema-visit, 2 of them useful**. Server-side commit rate ~4,300 tx/s (baseline before the `setup=`-on-every-acquire behavior: ~300–450 tx/s with the same worker count and poll cadence); tuples read/written per second are flat-to-lower, so this is ceremony, not throughput. Under a transaction-mode pooler each statement is its own transaction, so the ceremony multiplies straight into commit rate and CPU (~4 cores sustained on a 6-core primary at 22 active schemas — and it scales linearly with flagged schemas).

The `setup=` callback exists for a good reason (session GUCs must survive connection reuse through a pooler that resets state); the cost problem is its multiplication by per-schema acquires.

## Suggested fixes (either alone removes most of the cost; both are small)

[...]

PR #3541, opened 2026-08-17
fix(recall): make the recall budget reach the vector index

nicoloboschi:
## The bug

The semantic arm asks the index for `max(limit * 5, 100)` rows, but every pool connection runs with a fixed `hnsw.ef_search = 200`, applied at init time (`memory_engine._init_connection` → `_ANN_TUNING_HIGH_RECALL`).

In pgvector the candidate list **is** the result set. `GetScanItems` runs **once** with `ef = hnsw_ef_search`, `HnswSearchLayer` caps that list at `ef`, and when it drains the scan ends:

```c
if (list_length(so->w) == 0)
{
    if (hnsw_iterative_scan == HNSW_ITERATIVE_SCAN_OFF)
        break;      /* scan over — a larger LIMIT is unreachable through the index */
```

Iterative scans (0.8.0+) are what would resume it, and they are off by default. pgvector documents the consequence under *"Why are there less results for a query after adding an HNSW index?"*: *"Results are limited by the size of the dynamic candidate list (`hnsw.ef_search`)."*

So the recall budget moved the SQL `LIMIT` and nothing else:

| Budget | `thinking_budget` | SQL LIMIT | rows the scan can return |
|---|---|---|---|
| LOW | 100 | 500 | ~200 |
| MID (default) | 300 | 1500 | ~200 |
| HIGH | 1000 | 5000 | ~200 |

At MID and HIGH the arm could not even fill the rows it then trimmed to.

## The fix

Enable `hnsw.iterative_scan = strict_order` on recall connections, beside the `ef_search` value already there. The drained list is refilled in `ef_search`-sized rounds until the query's `LIMIT` is met, so depth follows each query's budget **with no per-query statement**.

That last part is why it is done this way rather than by setting `ef_search` per query: behind a transaction-mode pooler, a session GUC issued between statements can bind to a different backend than the query — the failure mode `link_utils.py:559-566` already documents from a production incident. `SET LOCAL` in a transaction would be correct but gives recall transaction semantics it never had. Riding the connection-init path has exactly the exposure that exists today and no more.

`strict_order`, not `relaxed_order`: the arms are trimmed in Python assuming rows arrive ordered by distance. Retain-side link probing pins it **off** — it is tuned for latency, not depth.

**The row over-fetch is deleted rather than tuned.** It never did anything: each arm's rows arrive already distance-ordered, so keeping the first N of 5N returns precisely what `LIMIT N` would have — the surplus was fetched, decoded and dropped unread. Invisible on pgvector (`ef_search` truncated first) but real on backends with no such bound, which fetched 5× the rows they used. The `LIMIT` is now `max(limit, GRAPH_SEED_LIMIT)`, because the graph arm reads its entry points from these same rows and a budget below 20 would otherwise starve it.

No arm caps: truncating an arm removes the corroboration signal RRF uses to promote a memory one arm ranked highly and another did not.

## Measurements

**40k-row bank, `EXPLAIN` confirming an ANN index scan, median of 7:**

| case | iterative off | iterative on | delta |
|---|---|---|---|
| MID unfiltered | **200 rows**, 5.0ms | **300 rows**, 4.9ms | −3% |
| HIGH unfiltered | **200 rows**, 4.9ms | **1000 rows**, 6.8ms | +39% |
| HIGH + selective filter | **200 rows**, 5.9ms | **324 rows**, 14.2ms | +142% |

`off` returns exactly 200 in every case regardless of `LIMIT` — the bug reproduces at scale. The filtered row is the cost: post-scan filters make the scan resume and discard most of what it finds, then stop at the ceiling. **+8ms against a ~2,600ms recall.**

**`run-perf-test.sh --suite recall --scale large`** (5k bank, `Budget.HIGH`, concurrency 16, 100 iterations), identical code both runs with only the tuning profile reverted:

| metric | before | after | delta |
|---|---|---|---|
| latency p50 | 2.717s | 2.771s | +2.0% |
| latency p95 | 3.750s | 3.780s | +0.8% |
| throughput | 6.19 q/s | 6.11 q/s | −1.3% |

**Memory is not the constraint it looks like.** pgvector caps a resumed scan at `work_mem × hnsw.scan_mem_multiplier` (`hnswscan.c:155`), but `max_scan_tuples` binds first — squeezing `work_mem` from 64MB to 256kB leaves both rows and latency unchanged, so the scan's footprint is under a quarter megabyte.

## Operational controls

```bash
HINDSIGHT_API_ANN_ITERATIVE_SCAN=false     # revert to current behaviour
HINDSIGHT_API_ANN_MAX_SCAN_TUPLES=1500     # or dial the cost down without reverting
```

The kill switch **drops** the resume GUCs rather than sending `iterative_scan=off`, so a connection is left exactly as it was before this existed — the revert lands on the behaviour already running in production, not a third state. No deploy needed: the pool re-applies session settings on every acquire.

The ceiling is the softer lever. pgvector does not count the initial scan toward it, so it interpolates between the old behaviour and full budget depth rather than switching; even `1` leaves today's depth intact.

## Also fixed

pgvector reserves the `hnsw.` GUC prefix, so a server **older than 0.8.0 rejects** `hnsw.iterative_scan` rather than accepting a placeholder. That fails the batched session setup and drops `apply_session_settings` into its per-setting fallback — 4 round trips per acquire instead of 1, forever, which behind a transaction-mode pooler is 4 server-side transactions each (the burn #3499 removed). Rejected GUCs are now remembered and dropped from later batches. Also covers `pg_trgm` on a cluster without the extension, which that function's docstring already anticipated.

## Tests

`tests/test_ann_iterative_scan.py` — both tuning profiles, the flags, and a DB-backed test that drives the env flag through the real pool and asserts **both** directions on one fixture: on, a 400-row budget reaches the index; off, it caps at ~200. It builds its own bank and index and asserts the plan is an ANN index scan (`Index Scan` present, `Sort` absent) — without that check a btree scan plus a sort satisfies it while returning every row, and the test measures nothing.

## Not measured

Whether the extra candidates produce better answers. Everything above is cost. `run-longmemeval.sh` / `run-locomo.sh` need credentials that were not available here — worth running before treating the trade as settled.


nicoloboschi:
## Benchmark results

Two measurements, because the perf suite alone can't answer the question.

### 1. `run-perf-test.sh --suite recall --scale large`

5,000-memory bank, `Budget.HIGH`, concurrency 16, 100 iterations. Identical code both runs — only the tuning profile reverted, so the scan mode is the sole variable.

| metric | before | after | delta |
|---|---|---|---|
| recall latency p50 | 2.717s | 2.771s | +2.0% |
| recall latency p95 | 3.750s | 3.780s | +0.8% |
| recall latency p99 | 3.879s | 3.859s | −0.5% |
| throughput | 6.19 q/s | 6.11 q/s | −1.3% |
| parallel_retrieval p50 | 1.931s | 2.018s | +4.5% |

**But this run cannot be trusted on its own.** The harness deletes its bank afterwards, so I could not confirm the planner used the HNSW index — at 5k rows with `LIMIT 1000` a sequential scan is entirely plausible, and then the numbers say nothing about a change to ANN scanning.

### 2. Direct measurement on a 40,000-row bank

Same partial HNSW index recall uses (`WHERE fact_type = … AND bank_id = …`), `EXPLAIN` confirming an index scan in every case, median of 7 runs:

| case | plan | iterative off | iterative on | delta |
|---|---|---|---|---|
| MID unfiltered (`LIMIT 300`) | index scan | **200 rows**, 5.0ms | **300 rows**, 4.9ms | −3% |
| HIGH unfiltered (`LIMIT 1000`) | index scan | **200 rows**, 4.9ms | **1000 rows**, 6.8ms | +39% |
| HIGH + selective filter | index scan | **200 rows**, 5.9ms | **324 rows**, 14.2ms | +142% |

Three things this settles:

1. **The bug reproduces at scale.** `off` returns exactly 200 rows in every case regardless of `LIMIT` — the recall budget genuinely does not reach the index today.
2. **MID is free.** 50% more candidates at the same ~5ms.
3. **The worst case is +8ms** against a recall whose end-to-end latency is ~2,600ms — the ANN scan is well under 1% of a recall. That is why the suite above shows ~1%.

The filtered row shows `hnsw.max_scan_tuples = 4000` doing its job: the scan stopped at 324 of the 1,000 rows requested rather than running away. Uncapped it would have kept resuming to pgvector's default of 20,000 tuples.

### Caveats

- Measured on a 40k bank on a dev box. HNSW cost grows with `log(N)`, so a multi-million-row bank will cost more per resume round — still bounded by `max_scan_tuples`.
- The perf suite runs **unfiltered** recalls only, so row 3 above is the only evidence for the filtered path.
- `4000` is chosen, not derived. It is the knob to reach for if a p99 regression shows up.

### One more fix this surfaced

pgvector reserves the `hnsw.` GUC prefix, so a server **older than 0.8.0 rejects** `hnsw.iterative_scan` rather than accepting it as a placeholder. That fails the batched session setup and drops `apply_session_settings` into its per-setting fallback — 4 round trips per acquire instead of 1, forever, which behind a transaction-mode pooler is 4 server-side transactions each (the burn #3499 removed).

Fixed in `88c776ddb`: a GUC the server rejects with `UndefinedObjectError` is remembered and dropped from later batches, so the cost is paid once per process. This also covers `pg_trgm.similarity_threshold` on a cluster without pg_trgm, which that function's docstring already anticipated.

PR #3547, opened 2026-08-17
docs(api): correct recall fact type descriptions

Sanderhoff-alt:
## Summary

- Correct the HTTP recall documentation to state that omitted `types` recalls all fact types.
- Document `observation` in the recall endpoint description.
- Regenerate the two OpenAPI references and synchronize Python, TypeScript, and Go client comments.

## Validation

- OpenAPI JSON parses successfully and both copies are identical.
- `ruff check hindsight_api/api/http.py`
- `npm run build` in `hindsight-docs`

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -298,7 +298,7 @@ class RecallRequest(BaseModel):
     query: str
     types: list[str] | None = Field(
         default=None,
-        description="List of fact types to recall: 'world', 'experience', 'observation'. Defaults to world and experience if not specified.",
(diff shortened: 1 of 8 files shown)


koriyoshi2041 (approved):
Verified at `53c26cd`: the endpoint already defaults absent/empty `types` to all three values in `VALID_RECALL_FACT_TYPES`, so the corrected description matches runtime behavior. The two OpenAPI copies are byte-identical and parse cleanly; the generated Go/Python/TypeScript comments carry the same contract. Local checks: Ruff on `http.py`, `git diff --check`, and 201 focused recall/MCP tests passed. Five DB-backed recall cases could not start locally because the optional `pg0-embedded` extra is not installed; hosted API/client/docs/generated-file checks are green.

nicoloboschi (approved):

PR #3560, opened 2026-08-17
engine: serve the addressed document/chunk/entity reads from the store too

nicoloboschi:
Draft — syncing as work proceeds. Follows #3557 (now merged into this base).

Store-generic: nothing here names a particular backend; everything keys off the capability probes already on `MemoriesExtension`.

## The problem

The **list** routes already consult `owns_document_store_for` and read the store. The **addressed** routes did not. So for a store that owns its document metadata and entity registry, the API handed out ids from a list route that its own detail routes then 404'd:

```
GET /banks/{b}/documents             -> 200, 50 items
GET /banks/{b}/documents/{that id}   -> 404 "Document not found"
GET /banks/{b}/entities              -> 200, 31 items
GET /banks/{b}/entities/{that id}    -> 404
GET /banks/{b}/documents/{id}/chunks -> 404  (so no chunk id is discoverable at all)
```

A list route and its detail route disagreeing about whether a record exists is the kind of thing a client cannot work around.

## The change

Four reads, each of which could only ever miss for such a store:

| method | was | now |
|---|---|---|
| `get_document` | both branches read the row `FROM documents`; the else branch's comment even asserted *"the documents row is still SQL"* | reads the record from the store when it owns the document store |
| `list_document_chunks` | verified the document against SQL and paged SQL `chunks`, overlaying only the *text* from the store — both empty | served from the store, rebuilding chunk ids in the same `{bank_id}_{document_id}_{index}` shape retain writes, so an id from this route is accepted by the addressed route |
| `get_chunk` | looked the id up in SQL `chunks` | parses the self-describing id and serves from the store |
| `get_entity` | read only SQL `entities` | resolves the one id against the store's registry — an addressed lookup, not a page-and-scan, so it stays O(1) in registry size |

Two small module helpers rather than inline logic:

* `_parse_chunk_id` — both splits are from the **right**: the index is the final segment and the document id before it is a UUID (no underscore), so a bank id that itself contains underscores still parses. Returns `None` on anything not in that shape, so the caller falls through to SQL rather than guessing.
* `_epoch_ms_to_datetime` — a store returns epoch millis where SQL returns datetimes, and the response builders call `.isoformat()`.

## Known gap, surfaced rather than hidden

`retain_params` is not carried in the store's document record, so `retain_params`, `document_metadata` and `observation_scopes` come back **null** from `get_document` for such a bank. That is a gap in what the *write* path persists — returning null beats 404-ing the whole document, and closing it properly is a separate write-path change.

## Why none of this was caught before

These paths are only reachable once a deployment actually answers `store_owned_retain_for`. The deployment layer I was testing against never did, so the store-owned branch had never executed anywhere. Every one of these was found within an hour of turning it on, by a per-endpoint benchmark that flags an endpoint returning zero records rather than reporting its (fast) latency as a good number.


nicoloboschi:
### Checkpoint — companion store-side fix found and landed

While verifying this PR end to end I found that the addressed-read 404s were only half the story. The other half was in the store, and it explains the `"entities": []` symptom that looked like a curation bug:

**The cheap fold dropped the entity registry.** `fold()` takes the flush path when there is no mergeable run, and that path is a separate implementation from the streaming fold — it materialized document records but not entity records, so the published cursor advanced past every registration and GC then took the only remaining copy. Registrations resolved fine from the un-folded tail and vanished at the next fold.

That is fixed store-side, with the regression test restored. Nothing in this PR changes as a result — the four addressed reads were independently wrong — but reviewers should know the two were being diagnosed together, and that `get_entity` here now has a registry that actually survives to answer it.

One thing this PR deliberately does **not** touch: `list_entities` is registry-backed, while the legacy posting path attaches entity ids without registering records, so a bank written that way lists zero entities. Whether that listing should also fall back to SQL is a behavioural decision rather than a bug fix, so it is called out rather than quietly changed.

nicoloboschi:
### Added: recall chunk hydration was 28x plain recall

Pushed a third commit, from chasing the biggest number the per-endpoint benchmark turned up.

`recall` with `include.chunks` measured **20.6s p50** against **735ms** for the same recall without it — same 30 results. Entity hydration on the same corpus is free by comparison, so the cost was specific to chunks.

Two causes, and counting round-trips alone hides both:

* the per-document fetches were `await`ed **in a loop**, serialising one store round-trip per document — ~690ms each across ~30 documents, which is essentially the entire 20.6s;
* `list_chunk_texts` downloads a document's **whole packed chunk blob**, and recall hits are spread across documents, so the usual case is one wanted chunk per document with the rest of the blob wasted.

Fix: fan the documents out concurrently (bounded), and choose the call shape per document — `get_chunk_text` when only one chunk of that document was hit, `list_chunk_texts` when several were and the blob amortises. A failed document now degrades to empty text with a warning instead of failing the whole recall.

Worth noting for review: the comment being replaced justified the loop as *"O(documents) round-trips, not O(chunks)"*. That was accurate and still far too slow — the round-trip **count** was never the problem; the serialisation and the per-call payload were. Measured numbers to follow once it is deployed.

nicoloboschi:
### Verified on a live deployment

Deployed and benchmarked on a fresh bank (5000 documents, concurrency 8). **0% error rate on every endpoint** — the first run in which none of the store-owned reads failed:

| endpoint | before | after |
|---|---|---|
| `recall+chunks` | 20613 ms | **3926 ms** — 5.25x, 0.37 → 1.92 qps |
| `document-get` | **100% errors** | **1019 ms** |
| `document-chunks` | **100% errors** | **117 ms** |
| `chunk-get` | **no id was discoverable** | **60 ms**, 127 qps |
| `entity-get` | **100% errors** | **85 ms**, 92 qps |
| `documents-list` | 4273 ms | 3253 ms |
| `recall` | 735 ms | 761 ms — control, unchanged |
| `recall+entities` | 716 ms | 785 ms — control, unchanged |

Id discovery over identical corpora went from **0 chunks / 0 entities** to **200 / 32**.

The two `recall` rows are the controls, and both are unchanged — that is what says the rest of the table is these fixes rather than a quieter cluster.

### One number this does NOT close

`recall+chunks` is still **5.2x plain recall** (3926 ms vs 761 ms). The fix removed the serialisation, not the per-document cost: at ~30 documents against a concurrency bound of 16 that is roughly two waves, which puts a single chunk fetch near 2 s. That points at the store side — whether `GetChunk` reads the whole packed blob to slice one chunk — rather than at more orchestration. Raising the bound would halve wall-clock by making it one wave, but that trades a real fan-out increase for a number I have not measured, so I have left it rather than tuning blind.

nicoloboschi:
### Two more store-owned read gaps, and a correction

**`get_observation_history` (new commit).** It probed SQL `memory_units` to decide whether the memory exists and whether it is an observation. That table is empty for a store-owned bank, so it 404'd a memory `memories/list` had just returned — the same list-says-yes/detail-says-no shape as the document, chunk and entity reads already in this PR. Only the existence + fact_type probe moves; the history rows themselves stay in SQL, which is right — they are curation metadata, not memory content.

Worth stating plainly because **I got this one wrong first**: I originally recorded that 404 as expected, on the grounds that the bench bank runs with `enable_observations=false`. It is not. The failure is in the existence check, before history is consulted at all, so it would 404 with observations enabled too.

**`retain_params` (earlier commit).** Now carried into the store's document record as one JSON value and parsed back through the same `parse_json` the SQL column uses, so both paths produce identical output for identical input. This closes the gap the PR description previously only surfaced. Two unit tests pin it and I verified the first fails without the fix. Documents written before this still read back null — a data gap, not a code path.

Also corrected `_store_document_bodies`'s docstring, which asserted the SQL `documents`/`chunks` rows still carry the small metadata. True when only the bulky text moved to the store; not true for a store-owned retain, where there is no SQL row at all — which is exactly how those three fields went missing.

nicoloboschi:
Force-pushed back to `9655b72`: the two per-bank counter commits are dropped, so this PR is once again only the addressed-read work (documents, chunks, entities, observation history) plus the concurrent chunk hydration and `retain_params` carry-through.

The counters were the wrong shape. They maintained per-bank counts in a Postgres table, bumped by a `touch()` on the retain path — which puts a database round-trip back on the write path that store-owned mode exists to keep off the database, and duplicates numbers the store already owns into every tenant schema.

The counts belong to the store, which already computes them; the replacement is a cheaper read on the store side, and it depends on this endpoint becoming paginated so that fetching per page is bounded.

Nothing here was deployed anywhere; the schema change was never applied.

PR #3561, opened 2026-08-17
fix(db): earn per-bank vector indexes by size instead of creating three per bank (#3485)

### LINKED ISSUE #3485: Per-bank partial vector indexes exhaust the Postgres lock table at ~2,100 banks — recall AND bank deletion both fail cluster-wide

nicoloboschi:
## The problem

Every bank got three partial vector indexes on the shared `memory_units` table at creation time. PostgreSQL takes a lock and builds an `IndexOptInfo` for **every** index on a relation at plan time, and `ExecOpenIndices` opens every one of them for each DML statement — so an index created for one bank is a cost paid by queries belonging to every *other* bank.

Past a few thousand banks that exhausts the lock-manager pool and both recall and bank deletion fail cluster-wide. Deletion failing is what makes it a trap: the API can no longer delete banks to recover.

Closes #3485.

## Why size, and not global indexes

The obvious fix — three global fact-type indexes — was rejected on the issue's own evidence. In both benchmarks offered as proof it works, the global index **is never chosen**: the planner takes `idx_memory_units_bank_fact_type` + a top-N sort for every bank up to ~10k units. The measured win (1,500 ms → 23 ms) comes entirely from *deleting* 6,000 indexes and none from adding 3 — which would then be maintained on every write to serve a minority of queries, and served worst exactly when used, since filtered ANN over a whole-corpus graph is weakest when the bank is a small fraction of the table.

The real variable is bank size, and the old code never looked at it.

## What changes

`HINDSIGHT_API_VECTOR_INDEX_MIN_ROWS` sets the rows a `(bank, fact_type)` partition needs before it gets its own index.

**It defaults to 0 — no minimum — so nothing changes for an existing deployment.** Every bank holding memories is indexed exactly as before. A deployment with thousands of banks sets `10000` and gets:

```
10k banks × 421 units  →  0 vector indexes   (was 30,000)
1 bank × 500k units    →  3 vector indexes   (unchanged)
```

Below the threshold the planner answers the same query from the B-tree plus a sort — **exact** rather than approximate, and faster — so small banks lose nothing. Index count becomes proportional to the number of *large* banks, and bank count stops being a ceiling.

An empty partition is excluded explicitly rather than by arithmetic. At a threshold of 0 the bare `row_count >= minimum` is true for zero rows, which would entitle every bank to three indexes over nothing the moment it was created — the index explosion the threshold exists to prevent, reintroduced by its own default.

## No index DDL on any request path

Bank creation, retain and import all drop their `CREATE INDEX`. That takes `CREATE INDEX`'s `ShareLock` on the shared table out of the retain transaction, where it deadlocked against concurrent writers.

Coverage is maintained by a `vector_index_maintenance` async operation instead. **Only a write moves a bank across the threshold**, so there is nothing a periodic sweep could discover that the writer does not already know — it hangs off `_submit_post_insert_maintenance` (retain and import) and `_handle_consolidation`, the other writer of `memory_units` rows, since observations are their own indexed partition.

It follows `submit_async_graph_maintenance` exactly: plan first and short-circuit with `no_work=True`, so an unconditional caller pays only a cheap bank-scoped pre-check; dedupe by bank including *processing*, since the job re-plans from live counts when it starts. A failed build is **logged, never raised** — it leaves the bank on the exact B-tree path (slower on a large bank, never wrong), the usual cause is transient DDL contention, and the next write re-queues it. Failing the operation would surface routine contention as a broken async op.

Bank deletion keeps its drop: it is the only place that still knows the `internal_id` the index names derive from.

## Recovery for an instance already at the wall

An index whose bank row is gone cannot be planned from — there is no `internal_id` left to derive its name — so no bank-scoped path can reach it. That is exactly the state a wedged deployment is in, because it could not run `delete_bank` at all. `hindsight-admin repair-bank --all` now also runs a catalog-only orphan sweep, keyed off the live `internal_id` set so it cannot mistake a live bank's index for a leftover.

`DROP INDEX` is a utility statement that locks its own index plus the table, rather than planning against all of them the way any DML must — which is why this path still works while everything else on that relation is failing.

## Bugs caught before merge, worth naming

- **A bank-scoped reconcile would have dropped every other bank's indexes.** An earlier cut derived the drop set by subtraction — "every index this schema owns that the partition list does not account for" — and with one bank's partitions that covers everyone else's. `repair-bank --bank X` would have stripped the schema. Now structurally impossible (planning starts from one bank's `internal_id`), with a bystander test kept as a guard.
- **A schema discovery *skipped* looked identical to "scanned, nothing qualifies"**, so a transient lock would have dropped all of that schema's indexes. Moot now that discovery is gone, but it is why the design moved from a sweep to per-bank planning.
- **The handler raised on a failed build**, turning routine DDL contention into failed operations. Now advisory.
- **A guard test caught a missing registry entry:** a new operation type needs `WORKER_SLOT_TYPE_DEFAULTS` (or no `RESERVED_SLOTS` env var exists and capacity cannot be reserved for it) and the control plane's `OPERATION_TYPE_VALUES` + label map across all ten locales (or the operations list shows a raw snake_case string and cannot filter on it).

## Verification

Full `api-slim` suite: **5695 passed**, with zero failures in any file this branch touches. `lint.sh`, `ty` and the migration-shape lint clean.

The reconcile and DDL-lock suites are pinned to one xdist worker: they issue `CREATE/DROP INDEX CONCURRENTLY` against a single shared `memory_units`, and eight workers doing that deadlock by design (`CONCURRENTLY` holds `ShareUpdateExclusive` while waiting out every session whose snapshot could still see the index — including other sessions' queued index DDL). Test staging builds concurrently with retry for the same reason: a plain `CREATE INDEX` takes `ShareLock` and closes a three-way cycle.

Tests also raise the threshold out of reach suite-wide. At the shipped default every throwaway bank the suite creates would queue a build; tests that exercise coverage patch the threshold themselves.

Remaining failures on this branch are pre-existing, reproduced on unmodified pre-change source against a clean database: `test_recall_min_score` ×2 (tie-sensitive threshold arithmetic — `finals[-1] + (finals[-2] - finals[-1]) / 2` drops nothing when the two lowest scores are equal), three live-LLM tests, and pg0 port collisions in the migration suites.

## Open questions

- **`10000` as the suggested value is reasoned, not measured.** It matches `SCANN_MIN_ROWS_FOR_AUTO_INDEX` and sits at the low end of the 10k–100k crossover the issue's probes bracket; the real crossover on production-shaped data is one `run-perf-test.sh` pass away. Erring low is safe — below the threshold, search is exact.
- **The default means #3485 is opt-in.** A deployment hitting the wall gets no relief until it sets the flag. That is the price of not changing behaviour on upgrade; the docs lead with "Limiting vector indexes on large deployments" so the knob is findable.
- **Many short-lived banks now cause a burst of concurrent index builds** at the default, for banks that never grow. Long-lived banks build once and go quiet, so this only bites a create-heavy workload.

## Relationship to the other open PRs

Supersedes #3535 (global fact-type indexes) and makes #3493 unnecessary — deletion no longer needs indexes dropped first, because at a raised threshold almost no bank has any. #3541 is orthogonal and still stands.


xiaoxiaoHe-E:
@nicoloboschi Thanks for the detailed write-up on #3561 — the threshold approach is a solid fix, and I think it addresses the root cause well.

One point I'd push back on gently: the claim that #3493 becomes unnecessary. I may be missing something, but it seems to me the two fixes are complementary rather than substitutes:

1. **The default threshold is 0.** For deployments that don't set `HINDSIGHT_API_VECTOR_INDEX_MIN_ROWS` (likely the majority), every bank still has its indexes, and the #3485 failure mode is unchanged there. In that configuration, drop-first in `delete_bank` is the only recovery path — worth keeping just for that.
2. **#3561 keeps the delete-path drop but doesn't change its timing.** At lock-table exhaustion, the delete DML itself can't be planned, so a post-commit drop never runs. #3493 is specifically about moving that drop *before* the delete so the delete stays possible. It feels like #3561 relies on that ordering rather than replacing it.
3. **The orphan sweep is a backstop, not a replacement.** It cleans up indexes left behind when `delete_bank` couldn't run — exactly the state #3493 prevents. Without #3493, wedged instances would produce those orphans at scale.

Happy to close it if you still think it's redundant after this — just wanted to make the case for keeping both, since they solve different halves of the same problem (#3485). Either way, thanks for driving the fix forward!

PR #3565, opened 2026-08-18
feat(maintenance): support cron trigger timezones

### LINKED ISSUE #3553: [Feature Request] Timezone support for mental model refresh_cron (currently hardcoded to UTC)
## Problem

`refresh_cron` on mental model triggers is always evaluated in **UTC**, with no way to specify a timezone:

- `MentalModelTrigger` has no `timezone` field (see OpenAPI `MentalModelTrigger-Input`).
- The scheduled-refresh routine in `hindsight_api/engine/maintenance.py` computes due-ness against `datetime.now(timezone.utc)` and passes it to `croniter` directly, so cron hours are interpreted as UTC hours.
- The field description does document "Cron expression (UTC, ...)", so this is by design today — but it's a recurring trap for users outside UTC.

## User impact

I set `refresh_cron: "0 21 * * *"` on 37 mental models intending **21:00 local time (Asia/Shanghai)**, and only noticed when the control-plane UI showed "next run in 11 hours" instead of ~3. The refreshes actually fire at 21:00 UTC = 05:00 local the next day. Anyone configuring "run at night" schedules in a non-UTC timezone silently gets the opposite — a morning run.

Related past issues show timezone pain in other parts of the pipeline (#2033 extractor tz context, #1232 recalled memory dates) — mental model scheduling is another instance of the same gap.

## Requested behavior

1. Add an optional `timezone` field (IANA name, e.g. `Asia/Shanghai`) to `MentalModelTrigger` / the mental-models PATCH API.
2. Default it to **UTC** for backward compatibility.
3. In the maintenance routine, evaluate the cron against `now` converted to the trigger's timezone (e.g. `datetime.now(ZoneInfo(tz))`), rather than raw UTC.
4. Surface it in the control-plane UI next to the schedule so the effective local time is visible.

An environment-level fallback (`HINDSIGHT_API_CRON_TIMEZONE`) would also work if a per-trigger field is too granular.

## Current workaround

Manually convert the cron to UTC hours (e.g. `0 13 * * *` for 21:00 Asia/Shanghai). This breaks if the operator's timezone has DST, and is undiscoverable unless you read the API field description.

## Environment

- hindsight-api 0.9.0 (venv install, macOS, launchd-managed)
- PostgreSQL backend

Sanderhoff-alt:
## Summary

Add per-trigger IANA timezone support for mental-model scheduled refreshes. Existing triggers remain UTC-compatible, while users can write a cron schedule in the timezone where it should run.

### API and storage

`MentalModelTrigger` exposes `timezone: string = "UTC"`. Valid IANA names such as `Asia/Shanghai` and `Europe/Berlin` are accepted through Python `zoneinfo`. The validator runs in `mode="before"` so omitted, blank, and legacy explicit `null` values normalize to `UTC`; non-string JSON values are rejected with a normal validation error (HTTP 422), and invalid IANA names are also rejected.

The timezone is stored inside the existing `trigger` JSON document. No new mental-model table column is introduced. Existing trigger documents without this key remain valid and are interpreted as UTC. `tzdata` is an explicit, unpinned dependency; `zoneinfo` still prefers system IANA data and uses the package as a fallback. The lockfile currently resolves `tzdata` 2026.3.

### Schedule evaluation

The maintenance loop evaluates each trigger's five-field cron expression in its configured timezone using `ZoneInfo` and `croniter`. It compares scheduled instants against the existing UTC `last_refreshed_at`, preserving the existing stale-model check, mutual exclusion with refresh-after-consolidation, and maintenance cadence. DST-aware local wall-clock schedules therefore remain correct; missing timezones continue to behave exactly like the former UTC implementation.

The due/not-due decision is isolated in a deterministic helper and covered with fixed UTC instants, including a non-UTC timezone case.

### Database discovery and migration safety

PostgreSQL `mental_models_with_cron()` now returns `trigger->>'timezone'` in addition to the existing fields. Missing JSON keys return `NULL`; the application treats that as UTC.

Migration `d4e5f6a7b8c9` is PostgreSQL-only and uses the established install-run gate because the routine is database-global: only the base/config-schema migration run rebuilds it, while per-tenant runs skip it. Its function body is based on `c8b4e2a71f95`, including the 250ms `lock_timeout` set/restore and the `lock_not_available` / `deadlock_detected` skip arms. Upgrade and downgrade both preserve those protections. `DROP + CREATE` is required because adding a column changes the `RETURNS TABLE` signature and PostgreSQL cannot replace that signature with `CREATE OR REPLACE`.

### Frontend/backend tradeoffs

The backend accepts every valid IANA key available to its `zoneinfo`/tzdata environment. This keeps the API broad for integrations, aliases, and advanced values such as `Etc/*`; backend validation is the final authority.

The control plane intentionally offers a friendlier searchable combobox containing `UTC` plus the browser's `Intl.supportedValuesOf("timeZone")` city list. New mental models default to the browser's local timezone. Each option shows its current UTC offset as a helpful reference; the value is explicitly current and may change with DST.

When editing an existing valid timezone that is outside the browser list, the combobox injects and displays that current value without rewriting it. Leaving it untouched preserves the API value; changing it uses the recommended UTC/city options. Browser/server tzdata differences therefore do not silently rewrite existing configuration, and the server remains authoritative.

### API clients and control plane

- OpenAPI, Go, Python, and TypeScript generated clients expose the non-null timezone field.
- Hand-written TypeScript wrappers and control-plane API types use `timezone?: string`; they no longer advertise `null`, while the backend still accepts legacy JSON `null` and normalizes it to UTC.
- Mental-model create/update and knowledge-page wrappers forward `refreshCron` and `timezone`.
- Mapping regression tests cover those forwarding paths.
- The schedule editor uses existing shadcn `Command` + `Popover` primitives with no new dependency.
- Cron previews and next-refresh timestamps use the selected timezone, and localized copy/docs describe timezone-aware scheduling.

## Compatibility

- Existing API clients and trigger JSON documents remain valid.
- Omitted, blank, or legacy `null` timezone values run in UTC.
- Existing models retain their configured timezone when edited.
- Cron syntax remains the existing five-field format; the timezone changes where it is evaluated, not the syntax or refresh eligibility rules.
- Invalid timezone names and non-string values receive API validation errors rather than runtime failures.

## Validation

- OpenAPI, client, bank-schema, and docs generation scripts pass and generated files are synchronized.
- Targeted backend timezone and deterministic due/not-due tests pass.
- TypeScript trigger-mapping tests, timezone utility tests, TypeScript checks, and the control-plane production build pass.
- Ruff and repository lint hooks pass.
- Full database integration tests were not run locally because `pg0-embedded` is unavailable.

Fixes https://github.com/vectorize-io/hindsight/issues/3553


nicoloboschi:
this adds unnecessary complexity

PR #3577, opened 2026-08-18
feat(memory-defense): detect international and Chinese PII

### LINKED ISSUE #3574: Feature: context-aware PII detection for Mainland China and international data
### Use Case

We are deploying Hindsight for Chinese users. Retained conversations may contain
personal information that must be redacted before it is persisted.

The current Memory Defense implementation does not cover several common PII
categories used in our environment, especially Mainland China and
international formats.

### Problem Statement

The default regex Memory Defense extension only implements the
`sensitive_data` detector.

In `hindsight-api-slim/hindsight_api/extensions/memory_defense.py`, the current
PII patterns are primarily US-centric (`credit_card` and `ssn_us`). It does not
currently detect:

- Mainland China mobile phone numbers
- International phone numbers
- Email addresses
- Mainland China national ID numbers
- Mainland China bank card numbers
- Person names
- Postal addresses

The first five categories may be detectable with normalized, checksum-aware
rules. Person names and addresses are context-dependent and generally require
NER, an optional LLM detector, or another context-aware implementation.

With the following policy enabled:

{
  "enabled": true,
  "rules": [
    { "on": "sensitive_data", "action": "redact" }
  ]
}

these values are currently retained unchanged instead of being redacted.

This would make Memory Defense usable for Chinese and international
deployments, reduce the risk of storing personal information in memory, and
provide a consistent redaction path for retain, recall, export, and reflect
operations.

### Proposed Solution

Possible approaches:

1. Add high-confidence, locale-aware detectors for:
   - `phone_cn`
   - `phone_international`
   - `email`
   - `id_cn`
   - `bank_card_cn`

2. Normalize spaces, hyphens, country prefixes, and full-width characters before
   matching. Use checksum validation where applicable to reduce false positives.

3. Add an optional context-aware PII detector for:
   - `person_name`
   - `address`

   This could be implemented as an optional extension or configurable detector,
   since broad regexes for names and addresses would have a high false-positive
   rate.

4. Keep the default path local and deterministic if possible. Any LLM or
   external NER-based detector should be explicitly opt-in.

5. Add tests for both `redact` and `block`, including positive and
   false-positive cases, and update the Memory Defense documentation.

6. Ensure raw PII is never included in logs, webhook payloads, audit metadata, or
   hit previews.

### Alternatives Considered

We can implement a separate application-level PII scanner before calling
Hindsight, but this duplicates the retain protection logic and does not protect
other write paths consistently.

We could also provide a custom Memory Defense extension, but built-in support
for common locale-specific formats would be useful for the open-source default.

### Priority

Critical - blocking my use case

### Additional Context

Relevant implementation:

- `hindsight-api-slim/hindsight_api/extensions/memory_defense.py`
- `hindsight-api-slim/hindsight_api/extensions/builtin/memory_defense_regex.py`
- `hindsight-api-slim/tests/test_memory_defense.py`

The current implementation intentionally favors high-confidence regexes and
documents that context-dependent matches are outside the pure-regex scope.

Synthetic test examples should be used only; no real phone numbers, identity
numbers, bank cards, email addresses, API keys, or `.env` contents should be
included in the issue.

### Checklist

oldnicke:
Fixes #3574.
## Summary

- rebase the fix onto the latest upstream `main`
- add redaction patterns for email, Mainland China mobile numbers, international phone numbers, PRC identity numbers, and Mainland China bank cards
- add context-bound Chinese and English person-name and postal-address detection to limit false positives
- cover comma-delimited English and Chinese addresses without truncating persisted PII
- prevent invalid 16-digit `62` candidates from falling through to generic credit-card redaction
- validate PRC birth dates and checksums, validate bank cards with Luhn, and fully mask new PII hit previews
- synchronize all locale catalogs and generated coding-agent documentation

## Testing

- `uv run --no-sync pytest -p no:xdist -o addopts="" tests/test_memory_defense.py -q -k "redaction_samples or redaction_patterns or each_redaction or apply_redaction"` (241 passed, 56 deselected)
- `./scripts/hooks/lint.sh`
- pre-commit hooks
- `node hindsight-docs/scripts/sync-coding-agents-doc.mjs --check`

The DB-backed Memory Defense cases still require a local test database with the current `entities.entity_kind` migration.

### DIFF
--- a/hindsight-api-slim/hindsight_api/extensions/memory_defense.py
+++ b/hindsight-api-slim/hindsight_api/extensions/memory_defense.py
@@ -13,7 +13,9 @@ from __future__ import annotations
 import logging
 import re
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from dataclasses import dataclass, field
+from datetime import date
 from enum import Enum
 
 from hindsight_api.extensions.base import Extension
@@ -219,7 +221,53 @@ _REDACTION_PATTERNS: list[tuple[str, str]] = [
     # --- Private keys & generic credentials ---
     ("private_key_pem", r"-----BEGIN (?:RSA |EC |DSA |OPENSSH |PGP )?PRIVATE KEY( BLOCK)?-----"),
     ("jwt", _ascii_token_pattern(r"eyJ[A-Za-z0-9_-]{10,}\.eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}")),
-    # --- PII (US-centric defaults; can be tuned per deployment) ---
+    # --- PII ---
+    # These patterns use explicit ASCII lookarounds instead of ``\b`` so a
+    # value immediately next to CJK text is still detected. Name and address
+    # matching is intentionally context-bound to reduce false positives.
+    (
+        "email",
+        r"(?<![A-Za-z0-9._%+-])[A-Za-z0-9.!#$%&'*+/=?^_`{|}~-]+@"
+        r"[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?"
+        r"(?:\.[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?)+"
+        r"(?![A-Za-z0-9._%+-])",
+    ),
+    (
+        "phone_cn",
+        _ascii_token_pattern(r"(?:(?:\+|00)86[ -]?)?1[3-9](?:[ -]?\d){9}"),
+    ),
+    (
+        "phone_international",
+        _ascii_token_pattern(r"(?:\+[1-9]\d{0,2}|00[1-9]\d{0,2})(?:[ .()-]*\d){6,14}"),
+    ),
+    (
+        "id_cn",
+        _ascii_token_pattern(
+            r"[1-9]\d{5}(?:18|19|20)\d{2}"
+            r"(?:0[1-9]|1[0-2])(?:0[1-9]|[12]\d|3[01])\d{3}[\dXx]"
+        ),
+    ),
+    (
+        "bank_card_cn",
+        _ascii_token_pattern(r"62(?:[ -]?\d){14,17}"),
+    ),
+    (
+        "person_name",
+        _ascii_token_pattern(
+            r"(?i:(?:full[ -]?name|name|contact|recipient|"
+            r"客户姓名|姓名|联系人|收件人)[ \t:：]+"
+            r"(?:[\u4e00-\u9fff]{2,4}|[A-Z][a-z]+(?:[ \t]+[A-Z][a-z]+){0,2}))"
+        ),
+    ),
+    (
+        "address",
+        _ascii_token_pattern(
+            r"(?im:(?:(?:^|(?<=[\r\n]))[ \t]*address[ \t:：]+|"
+            r"(?:home[ -]?address|shipping[ -]?address|postal[ -]?address|"
+            r"收件地址|收货地址|联系地址|(?<![\u4e00-\u9fff])(?:地址|住址))[ \t:：]+)"
+            r"[^\r\n;；。.!！？]{4,160}(?<![ \t,，]))"
+        ),
+    ),
     # NOTE: credit_card regex is intentionally narrowed to 13-19 digits with
     # exact separators to reduce false positives on long product IDs.
     ("credit_card", _ascii_token_pattern(r"(?:\d{4}[ -]?){3}\d{1,4}")),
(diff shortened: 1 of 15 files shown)


r266-tech (commented):
Reviewed PR #3577 at head `f9a5f89bd8ebe9627fb428fb478ef8538bd0f67a` from snapshot `snapshot_42730654edccdd2d9eaf5a3ca26f16e72`; the supplied verification confirms checkout matches the authority-bound external PR head and `git diff --check` exited 0.

The PR description reports 233 focused Memory Defense tests passing plus lint and pre-commit checks; DB-backed Memory Defense cases were not run against a current test database.

**High:** The postal-address detector stops at the first comma/semicolon/sentence punctuation and caps the match at 120 characters, so `Shipping address: 123 Main St, Apt 4, Springfield, IL 62704` is only partially redacted, leaving the apartment, city, state, and postal code in persisted content. Extend address matching to cover comma-delimited components while bounding unrelated prose, and add English and Chinese multi-component regression tests.

**Medium:** The contextual person-name and address alternatives have no leading label boundary and do not constrain the address value, so `username: Alice` can match the embedded `name:` substring and `server address: localhost` can be redacted as an address. Require a separator/word boundary before context labels, add stronger value/context constraints, and cover `username`, `server address`, and `IP address` with negative tests.

**Medium:** A 16-digit value prefixed with `62` that fails the new bank-card Luhn validator still falls through to the older generic `credit_card` pattern, which accepts 13-16 digits, so the invalid candidate is redacted anyway. Define precedence for invalid 62-prefixed 16-digit candidates and add a regression test for that shape.

**Low:** Only the English control-plane catalog was updated to report 52 Memory Defense patterns; the nine other actively loaded locale catalogs still report 45. Synchronize the 52-pattern text across all active locale catalogs or derive the displayed count from a single source.

oldnicke:
@r266-tech Thanks for the detailed review. I addressed all of the requested changes in the latest commit (`7ae13198b18fe491570dbfcd273ab91e1980bb4a`).

Addressed items:
- Extended the postal-address detector to preserve comma-delimited English and Chinese address components while keeping length and statement-boundary limits.
- Added label boundaries and negative coverage for `username: Alice`, `server address: localhost`, and `IP address: 192.0.2.1`.
- Prevented invalid 16-digit `62`-prefixed values from falling through to the generic `credit_card` detector.
- Updated all active locale catalogs from 45 to 52 Memory Defense patterns.
- Synchronized the generated coding-agent documentation.
- Applied `_ascii_token_pattern` to the locale-aware detectors where the generic ASCII boundary is safe. The email detector keeps a dedicated boundary expression because valid email local-parts may contain `.`, `+`, `%`, and related characters.

Validation:
- Memory Defense focused tests: 265 passed
- lint: passed
- pre-commit hooks: passed
- generated documentation check: passed
- all current CI checks: passed
- branch rebased onto the latest `main`

The DB-backed Memory Defense cases still require a current local test database and were not included in the focused local run. Could you please take another look and re-review the updated commit?

oldnicke:
@r266-tech @nicoloboschi 

Thanks for the review. I’ve addressed all of the requested changes in the latest commit. All tests, lint, pre-commit, documentation checks, and CI are passing, and the PR is mergeable.

Could you please re-review it when you have a chance? If everything looks good, would you mind merging it?

Thanks!

hindsight-api-slim/hindsight_api/extensions/memory_defense.py
```diff
@@ -219,7 +221,53 @@ def _ascii_token_pattern(body: str) -> str:
...
+    (
+        "bank_card_cn",
+        _ascii_token_pattern(r"62(?:[ -]?\d){14,17}"),
+    ),
+    (
+        "person_name",
+        _ascii_token_pattern(
+            r"(?i:(?:full[ -]?name|name|contact|recipient|"
+            r"客户姓名|姓名|联系人|收件人)[ \t:：]+"
+            r"(?:[\u4e00-\u9fff]{2,4}|[A-Z][a-z]+(?:[ \t]+[A-Z][a-z]+){0,2}))"
```
Sanderhoff-alt, line 259: The `(?i:` group spans the value alternation as well as the labels, so `[A-Z][a-z]+` is effectively case-insensitive and any lowercase word after a `name:`/`contact:` label is treated as a name:

- `apply_redaction("name: my-service")` -> `"[REDACTED:person_name]-service"`
- `"Contact: sales team"` -> `"[REDACTED:person_name]"`
- `"The user said their name: Alice lives there"` -> the match extends over `Alice lives there`, since the two trailing lowercase words satisfy `[ \t]+[A-Z][a-z]+` too.

Retained YAML/config/log lines (`name: hindsight-api`) are common, so this mangles legitimate content. Scoping the flag to the labels only fixes it, e.g. `(?:(?i:full[ -]?name|name|contact)|客户姓名|姓名|联系人|收件人)[ \t:：]+` with the value alternation left case-sensitive.

hindsight-api-slim/hindsight_api/extensions/memory_defense.py
```diff
@@ -219,7 +221,53 @@ def _ascii_token_pattern(body: str) -> str:
...
+        r"(?:\.[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?)+"
+        r"(?![A-Za-z0-9._%+-])",
+    ),
+    (
+        "phone_cn",
+        _ascii_token_pattern(r"(?:(?:\+|00)86[ -]?)?1[3-9](?:[ -]?\d){9}"),
+    ),
+    (
+        "phone_international",
+        _ascii_token_pattern(r"(?:\+[1-9]\d{0,2}|00[1-9]\d{0,2})(?:[ .()-]*\d){6,14}"),
```
Sanderhoff-alt, line 241: The `00` branch matches any ASCII token that starts with `00` and has 6-14 further digits, so zero-padded identifiers are redacted as phone numbers: `"Order 00123456789 shipped"` -> `"Order [REDACTED:phone_international] shipped"`. The `+` branch has the same problem for numeric quantities: `"Revenue grew by +100 000 000 last year"` redacts `+100 000 000`.

Consider requiring a separator after the country code (or a minimum national-number length) on the `00` branch so plain digit runs are excluded, and add these two strings to the false-positive parametrization.

hindsight-api-slim/hindsight_api/extensions/memory_defense.py
```diff
@@ -219,7 +221,53 @@ def _ascii_token_pattern(body: str) -> str:
...
+            r"(?:[\u4e00-\u9fff]{2,4}|[A-Z][a-z]+(?:[ \t]+[A-Z][a-z]+){0,2}))"
+        ),
+    ),
+    (
+        "address",
+        _ascii_token_pattern(
+            r"(?im:(?:(?:^|(?<=[\r\n]))[ \t]*address[ \t:：]+|"
+            r"(?:home[ -]?address|shipping[ -]?address|postal[ -]?address|"
+            r"收件地址|收货地址|联系地址|(?<![\u4e00-\u9fff])(?:地址|住址))[ \t:：]+)"
+            r"[^\r\n;；。.!！？]{4,160}(?<![ \t,，]))"
```
Sanderhoff-alt, line 268: The body `[^\r\n;；。.!！？]{4,160}` is greedy to the end of the line, so everything after the label goes with the address even when it is unrelated prose: `"Address: 123 Main St and we then agreed to ship the Q3 revenue plan with Bob"` is replaced in full by `[REDACTED:address]`. Retained conversation lines frequently have no terminating `.`/`。`, so this drops content the user expects to keep. Bounding the span (a tighter length, or stopping at the first token that cannot be part of an address) would keep the multi-component address cases the tests cover without swallowing the rest of the line.

Sanderhoff-alt (commented):
The three new context-bound detectors over-match on ordinary retained content and silently delete it. Each case below is reproducible with `apply_redaction` alone, and none is covered by the added false-positive parametrization.

nicoloboschi:
for now we're not going to support this in the open source version as it's not that simple as this pr would imply

PR #3578, opened 2026-08-18
fix(engine): give each reranker worker its own CrossEncoder

JoshFunnell:
fix(engine): give each reranker worker its own CrossEncoder

## What

`LocalSTCrossEncoder` shared one SentenceTransformers `CrossEncoder`
(and its HuggingFace fast tokenizer) across a class-level
`ThreadPoolExecutor`. Concurrent `predict()` races
`set_truncation_and_padding` + `encode_batch`.

Observed live on 2026-08-15 15:33:08Z during a 3-bank recall fan-out
as `ValueError: Unable to create tensor` wrapping
`TypeError: 'int' object is not callable` (1 hit / 72 h on that
deployment). This is an engine bug. Multi-bank made the race more
likely (N parallel CE windows); a single-bank host with
`max_concurrent>1` has the same shape.

Each executor thread now owns its own `CrossEncoder`, loaded under a
lock. `to(device)` / `eval()` run once at load. Load failures raise;
there is no fallback to a shared instance. `initialize()` warms every
pool worker so a load failure is a startup error and the first recall
does not pay N cold loads.

`max_workers` is still `HINDSIGHT_API_RERANKER_LOCAL_MAX_CONCURRENT`
(default 4). Memory cost was ~+290 MiB private for the 3 extra MiniLM
copies at `max_concurrent=4` (measured on that deployment).

Rebased onto current upstream main. The engine/test tree is the same as
the earlier per-thread fix; the only local correction since then was
dropping an uncommitted private PR-body file before this rebase.

## This does not fix

A later process death (2026-08-16 10:58Z) landed immediately after a
17-candidate CE rerank and is **correlated with, not proven to be
caused by**, long-input CE behaviour (ST 5.2.0 `predict` omits
`max_length`). That is a separate change (tokenizer `max_length` +
pair clip) and should be its own PR. Bundling it here would mix two
failure classes and two review surfaces.

Not a multi-bank change. File separately from the multi-bank recall PR.
Does not change the public reranker provider interface. Does not
change FlashRank batching (#3355 / #3441).

## Bounded CrossEncoder inputs

Local CrossEncoder scoring tokenises every query/passage pair with
explicit `truncation=True` and `max_length` from the loaded tokenizer
(`tokenizer.model_max_length`). 512 is only the fallback when a
stub/model exposes no usable length. Values below 8 are skipped; values
above 131072 clamp (and log).

- First oversized **query** in a 60s window logs WARNING; later ones DEBUG
  so a busy reranker cannot flood logs.
- Known overflow `ValueError` (`Unable to create tensor`, `expected
  sequence of length`) retries one pair at a time. Other ValueErrors
  (shape, dtype, config) propagate. CUDA OOM / TypeError still raise.
- A pair that still overflows scores `-inf` so it cannot outrank a real
  score.
- If the tokenizer wrapper cannot bind (read-only tokenizer), a warning
  is logged and predict still runs.

Single-bank and multi-bank (union CE) share this path. Observed crash
class: `ValueError: Unable to create tensor` on a ~7.5 KB query.

## Tests

`hindsight-api-slim/tests/test_local_cross_encoder.py`:

- Existing mocked predict / bucket-batching / FlashRank tests.
- Isolation: exclusivity vs a legacy shared-model subclass (the
  attacking RED on the old design), score identity, both predict
  arms, load failure, warmup vs a smaller existing pool.
- Optional real-MiniLM soak: `HS_RERANKER_STRESS=1`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -9,6 +9,7 @@ Configuration via environment variables - see hindsight_api.config for all env v
 import asyncio
 import logging
 import threading
+import time
 import warnings
 from abc import ABC, abstractmethod
 from concurrent.futures import Future, ThreadPoolExecutor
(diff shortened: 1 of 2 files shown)


JoshFunnell:
Withdrawing to keep this in our local overlay alongside the multi-bank recall work; thanks.

PR #3580, opened 2026-08-18
fix(mental-models): refuse to overwrite real content with a failure placeholder

JoshFunnell:
fix(mental-models): refuse to overwrite real content with a failure placeholder

## What

A mental-model refresh whose candidate is a failure placeholder
(`NO_ANSWER_TEXT`, or an iteration-limit render) or whose retrieval
returned nothing this run will overwrite a document built over months.
Neither case is empty, so the existing emptiness guard does not see it.

Two checks on the write path, both scoped to **existing real** content
so a bootstrap write onto an empty/`Generating content...` model still
lands:

- Placeholder candidates are refused.
- This run's retrieval emptiness is captured **before** the delta-mode
  `based_on` carry-over merge, which would otherwise make an empty
  retrieval look populated.

Reuses the existing outcome `refresh_failed_empty_candidate` so the
caller's preserve-and-fail path needs no new outcome value. The warning
records `placeholder_candidate` or `empty_fresh_retrieval`.

Related: #2959, #2894.

## Tests

`hindsight-api-slim/tests/test_mental_model_placeholder_guard.py`:

- `NO_ANSWER_TEXT` and the iteration-limit prefix are placeholders.
- A real document is not.
- Refuse placeholder or empty-retrieval over a real baseline.
- Allow bootstrap on empty/PENDING.
- Allow a real refresh that used facts.

The dry-run refresh tests in
`hindsight-api-slim/tests/test_mental_model_dry_run_refresh.py` cover
the same outcomes through the write path.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -13066,9 +13066,7 @@ class MemoryEngine(MemoryEngineInterface):
         # Capture whether THIS reflect used any facts BEFORE the delta-mode
         # merge below folds in facts from previous refreshes. Post-merge
         # emptiness is fail-open under carry-over (#2894).
-        fresh_retrieval_empty = not any(
-            payload for payload in based_on_serialized_payload.values() if payload
-        )
+        fresh_retrieval_empty = not any(payload for payload in based_on_serialized_payload.values() if payload)
 
         # In delta mode, based_on must accumulate: the mental model is
         # grounded on ALL facts ever used, not just the latest delta's new
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -134,6 +134,26 @@ def _authorize_nested_operations() -> "Iterator[None]":
 
 MENTAL_MODEL_PENDING_CONTENT = "Generating content..."
 
+_ITERATION_LIMIT_PREFIX = "I was unable to formulate a complete answer"
+
+
+def is_placeholder_refresh_candidate(text: str) -> bool:
+    """True when the reflect candidate is a known failure placeholder, not a document."""
+    from .reflect.agent import NO_ANSWER_TEXT
+
+    candidate = text.strip()
+    return candidate == NO_ANSWER_TEXT.strip() or candidate.startswith(_ITERATION_LIMIT_PREFIX)
+
+
+def should_refuse_failed_refresh_overwrite(
+    *,
+    has_delta_baseline: bool,
+    is_placeholder: bool,
+    fresh_retrieval_empty: bool,
+) -> bool:
+    """Refuse a failed render only when it would overwrite existing real content."""
+    return has_delta_baseline and (is_placeholder or fresh_retrieval_empty)
+
 
 def get_current_schema() -> str:
     """Get the current schema from context (falls back to config default)."""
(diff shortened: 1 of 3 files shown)


koriyoshi2041 (commented):
I checked this at exact head `04150aeb912e`. The six model-free guard tests pass locally, and the three write-path cases reach only the optional embedded-Postgres setup gate on my machine (`pg0-embedded` is not installed), so I did not count those as patch failures. `git diff --check` is clean.

The current red `verify-generated-files` job is actionable formatting drift: `./scripts/hooks/lint.sh` rewrites `memory_engine.py` and `test_mental_model_placeholder_guard.py` (2 insertions / 6 deletions). The docs red looks base-related rather than caused by this diff: the coding-agents README/doc sync changed on main after this PR's base. Rebasing onto current main, running the lint hook, and committing its two rewrites should separate the patch signal from that drift.

One naming/detail worth tightening while doing that: `fresh_retrieval_empty` is computed from `reflect_result.based_on`, so it means “no facts used/grounded,” not necessarily “retrieval returned nothing” (the existing `facts.retrieved` count can be nonzero while `based_on` is empty). The conservative guard still looks intentional for #2894, but the variable/warning currently collapses those two diagnostic states.

koriyoshi2041 (approved):
Rechecked at exact head `ad8d1b342d`. The follow-up cleanly separates `retrieval_empty` from `no_grounded_facts` while preserving the conservative overwrite guard when recall returned facts but the render grounded none of them.

The focused placeholder/evidence-state suite passes 8/8 locally, and `git diff --check 04150aeb912e..HEAD` is clean. The earlier lint drift is resolved. The remaining docs failure is unchanged base-era README/docs sync drift rather than a file touched by this PR.

JoshFunnell:
Thanks for re-checking at `ad8d1b34`.

One correction to my earlier read of the docs failure: it is not base-era drift that a rebase clears. I ran CI's own check — `node hindsight-docs/scripts/sync-coding-agents-doc.mjs --check` — against **current `origin/main` (`edd0d0c5`)** as well as this PR head, and both report `[coding-agents] docs page is out of date with the README`. In the `build-docs` job, steps 1-5 (code parity, integration SEO, integrations, deps) pass; only the generated-doc check fails, and it fails on main itself.

The delta is one line: the generator emits `https://hindsight.vectorize.io/developer/observations` where the checked-in `docs-integrations/coding-agents.md` has the site-relative `/developer/observations`. Regenerating on `edd0d0c5` makes `--check` pass.

So this PR's red will not clear by rebasing. I have opened that one-file regeneration separately so the signal is separable for this PR and the others sitting on the same step.

nicoloboschi:
Superseded by #3761 (merged as f4a93695), which removes the placeholder rather than guarding against it.

Two things pushed the merged fix upstream of the write rather than at it:

- **Placement.** The guard here sits after the delta ops are applied and the document is re-rendered, so in delta mode `final_content` is the re-rendered document and `is_placeholder_refresh_candidate` no longer recognises it. In the reported production case `based_on` was full (the agent used 133 and 266 facts and still returned a blank `done.answer`), so the `no_grounded_facts` leg does not fire either — the run reports `content_unchanged` and advances the watermark past facts that were never incorporated.
- **Text matching.** `startswith("I was unable to formulate a complete answer")` breaks as soon as the copy is edited, and can false-positive on a document that legitimately opens that way.

#3761 raises `ReflectNoAnswerError` where each terminal path already knows it produced nothing, and deletes both placeholder strings, so nothing downstream needs to recognise them.

The bootstrap-safety reasoning here (`has_delta_baseline`, so a first write onto empty/`Generating content...` still lands) and the dry-run tests through the write path were useful in reviewing the merged version — thank you.

The empty-retrieval half of this PR is #2894 and is still open; #3135 covers it separately and is untouched by the merge.

PR #3579, opened 2026-08-18
feat(recall): multi-bank recall with score/interleave merge

JoshFunnell:
feat(recall): multi-bank recall with score/interleave merge

## What

Adds server-side multi-bank recall so a client that splits memory across
banks does not have to merge rank-relative per-bank scores itself.

- `MemoryEngine.recall_multi_async(bank_ids, query, *, merge=...)` fans out
  one unchanged `recall_async` per bank (`asyncio` task per bank, own
  `@_bind_bank_id` ContextVar).
- `merge="score"` (default) sorts the union by `scores.reranker`.
  Falls back to `merge="interleave"` (round-robin by per-bank rank) when
  CE is not comparable: `enable_reranking=false`, caller asked for
  `rrf`/`interleave`, or returned facts have no usable `scores.reranker`.
- After gather: per-bank cap 50 → merge → exact/normalized text dedup
  (`casefold` + Unicode whitespace collapse, mode `exact_normalized`) →
  token cut. Dedup is not embedding/near-duplicate collapse.
- Additive HTTP route `POST /v1/{tenant}/memories/recall`. Existing
  per-bank route is unchanged. MCP `recall` gains optional `bank_ids` /
  `merge` (2+ = multi; length-1 selects that bank; omit = session/`bank_id`).

`prefer_observations` defaults True on this orchestrator only. HTTP/MCP
request models still default False and pass the caller's value through.
Single-bank `recall_async` is unchanged.

MCP `bank_ids` of length 1 selects that bank. A singular `bank_id` is
also explicit (it is not replaced by the session bank). Tenant
`AuthenticationError` is authenticated once before fan-out and re-raised
as HTTP 401. Per-bank status lives in `metadata.multi_bank`.

## Single-bank vs multi-bank: what runs, and the speed consequence

| | single bank (`bank_ids` omitted, or length 1 / singular `bank_id`) | multi-bank (`bank_ids` length 2..10) |
|---|---|---|
| code path | `recall_async` exactly as before this PR — no new hop | `recall_multi_async`: one unchanged `recall_async` per bank, run concurrently via `asyncio.gather` |
| retrieval | one bank's candidates | N banks' candidates in parallel; each bank capped at `per_bank_cap` (50) before merge |
[...]

PR #3581, opened 2026-08-18
fix(recall): one union CrossEncoder pass for multi-bank recall

JoshFunnell:
fix(recall): one union CrossEncoder pass for multi-bank recall

> Stacked on #3579 (multi-bank recall): this branch carries that PR's commit; this PR's own change is the top commit. It will be rebased onto main once #3579 lands.

## What

Single-bank recall is untouched by this PR (no CE change, no new hop). Multi-bank recall changes only in HOW MANY times
the CrossEncoder runs: N per-bank passes (one inside each sub-call) become one pass over a capped union, which is what
makes CE scores comparable across banks and removes the N× rerank cost. See the "Single-bank vs multi-bank" table in #3579
for the full path split.

`recall_multi_async` already fans out one `recall_async` per bank. This
PR stops paying the local CrossEncoder N times.

When union CE is eligible (caller asked for `cross_encoder`, every bank
has `enable_reranking`, engine has a non-passthrough CE):

1. Each subcall uses `reranking="rrf"` (retrieval only).
2. After gather: per-bank pre-cap (default 50) + floor reservation +
   interleave fill up to `union_cap` (default **200**).
3. One `CrossEncoderReranker.rerank` over that union.
4. `apply_per_bank_floor` so a CE-weak bank still keeps
   `min(k_floor, its count)` after the CE cut (default floor 1).
5. Score/interleave merge as today (CE scores are now comparable across
   banks). The caller `max_tokens` is applied **after** that CE pass.

If CE is disabled / null / rrf-passthrough, or the caller requested
`rrf`/`interleave`, today's per-bank path is unchanged.

Per-bank subcalls that will be union-CE scored use a wide token budget
(`max(caller, 100000)`) so a typical caller budget of ~2000 cannot drop
deep RRF ranks before the one CE pass. The caller budget still cuts the
merged list afterwards.

HTTP/MCP request schema is unchanged.

Depends on the multi-bank recall PR. Does **not** change the per-thread
CrossEncoder ownership (file that first; this PR does not touch
`cross_encoder.py`). Does not pin torch intra-op threads.

## Why (measured, with configuration)

Config: local MiniLM, `budget=low`/`mid`, `HINDSIGHT_API_MULTI_UNION_CAP=200`,
no `HINDSIGHT_API_RERANKER_TORCH_THREADS` (unset = no pin), golds kept
in the CE batch, 5 reps, idle host, 2026-08-17.

| arm | before (per-bank CE) | after (union CE, U=200, no pin) |
|---|---|---|
| single-bank p50 | baseline | 0.90–0.94× |
| 3-bank p50 low / mid | 4.27 s / 5.34 s | 3.87 s / 3.79 s |
| 5-bank p50 low / mid | 6.29 s / 7.51 s | 4.66 s / 4.48 s |
| quality (top-5 overlap) | — | PASS (median 5/5, min 3/5) |

A 5-bank fan-out at U=200 keeps per-bank RRF ranks 0–39 in the CE pool.
A caller `max_tokens=2000` applied *before* CE in RRF order dropped
those deeper ranks (the earlier quality drop was absent golds, not a
different score on the same pair).

## Config knobs (static)

| Variable | Default |
|---|---|
| `HINDSIGHT_API_MULTI_UNION_CAP` | 200 |
| `HINDSIGHT_API_MULTI_PER_BANK_PRE_CAP` | 50 |
| `HINDSIGHT_API_MULTI_PER_BANK_FLOOR` | 1 |

Values below 1 (or non-integers) fall back to the default, not to 1.

## Tests

- Union CE: rrf-per-bank + one `predict`; floor keeps a CE-weak bank;
  CE-off / caller-`rrf` fallback.
- U=200 keeps RRF ranks 36/37 at 5 banks; U=150 drops them.
- Subcall budget: caller 2000 drops a rank-36 fact; the union budget
  keeps it.
- `_parse_iso_datetime` accepts `datetime`/`date` (a TypeError there
  would abort the whole union CE).
- Env `0` for `MULTI_UNION_CAP` uses the default, not 1.
- `hindsight-api-slim/tests/bench_multi_bank_union.py`: quality overlap
  >= 4/5; `HS_BENCH=1` for the synthetic wall-clock bench.

## Surfaces not updated

HTTP/MCP/OpenAPI/control-plane request models. `MAX_MULTI_BANK_RECALL_BANKS`
is still 10.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -333,6 +333,14 @@ HINDSIGHT_API_LOG_LEVEL=info
 # HINDSIGHT_API_RERANKER_MAX_CANDIDATES_LOW=0
 # HINDSIGHT_API_RERANKER_MAX_CANDIDATES_MID=0
 # HINDSIGHT_API_RERANKER_MAX_CANDIDATES_HIGH=0
+# Multi-bank union CE: one CrossEncoder pass over the merged pool (not N per-bank
+# CE jobs). Cost is O(union_cap), not O(N_banks). Default 200 keeps per-bank
+# RRF ranks 0-39 at a 5-bank fan-out in the CE batch.
+# HINDSIGHT_API_MULTI_UNION_CAP=200
+# Per-bank RRF head admitted into the union pool before the cap:
+# HINDSIGHT_API_MULTI_PER_BANK_PRE_CAP=50
+# After the CE cut, each bank keeps at least min(floor, its count):
+# HINDSIGHT_API_MULTI_PER_BANK_FLOOR=1
 
 # Reranker failover chain: extra rerankers tried, in order, when the one above
 # fails. Members are numbered from 1 (indices must be contiguous) and every
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -184,6 +184,7 @@ from hindsight_api.engine.response_models import (
     RecallScores,
     TokenUsage,
 )
+from hindsight_api.engine.response_models import RecallResult as CoreRecallResult
 from hindsight_api.engine.search.tags import TagGroup, TagsMatch
 from hindsight_api.engine.structured_output import validate_response_schema
 from hindsight_api.extensions import HttpExtension, OperationValidationError, load_extension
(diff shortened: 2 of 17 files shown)


JoshFunnell:
Folded into #3579 (commit 0e55cf10 and the section "Union CrossEncoder pass (folded from #3581)" in its description) so the multi-bank recall change and its CE cost fix review as one PR. Closing this one.

PR #3592, opened 2026-08-18
fix(coding-agents): let hindsight_reflect wait as long as it is configured to (#3590)

### LINKED ISSUE #3590: coding-agents: hindsight_reflect MCP tool ignores reflectTimeoutMs — always aborts at hardcoded 120s
### Bug Description

The `hindsight_reflect` MCP tool in `@vectorize-io/hindsight-coding-agents` always fails after
~120s with `This operation was aborted (20)`, no matter what `reflectTimeoutMs` is set to in
`~/.hindsight/coding-agent.json` (or via `HINDSIGHT_REFLECT_TIMEOUT_MS`).

The config value is parsed but never reaches the reflect call. On current `main`:

- `src/core/knowledge-tools.ts` — the `hindsight_reflect` handler calls
  `client.reflect(query, { budget: "high" })` without `timeoutMs`
- `src/mcp-server.ts` — `selectTools()` does not forward `cfg.reflectTimeoutMs` into
  `buildKnowledgeTools()`
- `src/core/hindsight.ts` — `reflect()` falls back to `opts.timeoutMs ?? 120000`

Impact: against a self-hosted server, `budget: "high"` synthesis on a populated bank routinely
takes >120s (ours: ~216s; the bank grows over time and reflect gets slower), so the MCP tool
is unusable — while the identical direct API call
`POST /v1/default/banks/{bank}/reflect` succeeds.

A secondary, related limitation: `budget` is also hardcoded to `"high"` in the same handler with
no config escape hatch. On large banks even `mid` can exceed the server's default
`HINDSIGHT_API_REFLECT_WALL_TIMEOUT` (300s), so making budget configurable would help too.

### Steps to Reproduce

1. Point `~/.hindsight/coding-agent.json` at a self-hosted Hindsight server and set
   `"reflectTimeoutMs": 660000`
2. In Claude Code, call the `hindsight_reflect` MCP tool with any query on a populated bank
3. Observe failure after ~120s: `{"error":"This operation was aborted (20)"}`
4. For contrast, `curl -X POST http://<server>/v1/default/banks/<bank>/reflect
   -d '{"query": ..., "budget": "high"}'` succeeds in ~216s

### Expected Behavior

The tool waits up to the configured reflectTimeoutMs and returns the synthesis.

### Actual Behavior

Aborts client-side at exactly ~120s regardless of configuration; the configured value is dead.

Verified local fix (patching installed dist):
- `selectTools()`: pass `reflectTimeoutMs: cfg.reflectTimeoutMs` into `buildKnowledgeTools()`
- handler: `client.reflect(query, { budget: "high", timeoutMs: opts.reflectTimeoutMs })`
After this, the tool waits for the server and returns successfully.

### Version

@vectorize-io/hindsight-coding-agents 0.3.4, server 0.8.4 (self-hosted)

### LLM Provider

Other

nicoloboschi:
Fixes #3590.

## The bug

`hindsight_reflect` aborted every call at a hardcoded 120s with
`This operation was aborted (20)`, whatever `reflectTimeoutMs` was set to. The handler passed no
`timeoutMs`, so `HindsightClient.reflect()` fell back to its own `?? 120000`. On a populated bank,
`budget: "high"` synthesis routinely runs longer than that — the reporter's identical direct
`POST /reflect` took ~216s and succeeded — so the tool was unusable and the config field was dead.

**Both** paths that build the tools dropped the setting, not just the one the issue names:

| call site | harnesses affected |
| --- | --- |
| `selectTools()` (`src/mcp-server.ts`) | the MCP server — Claude Code, Codex, … |
| `RuntimeCore.toolSpecs()` (`src/core/runtime.ts`) | the persistent plugin harnesses — opencode, Kilo, Cline, dsh, Prime Agent |

A fix to `selectTools()` alone would have left every plugin harness still aborting at 120s.

## The fix

- **`reflectToolTimeoutMs`** (default **330000**) — the agent-invoked tool's own window, set
  deliberately *above* the server's `HINDSIGHT_API_REFLECT_WALL_TIMEOUT` (300s) so the server
  decides when to give up rather than an arbitrary client deadline. Unset, it inherits an
  explicitly raised `reflectTimeoutMs` (the field users already reach for, as the reporter did),
  but a **short** one never lowers it: that value bounds an automatic hook which must fit the
  host's 25s window, not a call the agent is sitting and waiting on.
- **`reflectBudget`** (`"low" | "mid" | "high"`, default `"high"`) — the secondary ask in the
  issue. On a large bank even `mid` can be what fits the server's wall timeout. An unrecognized
  value warns and falls back to `"high"` rather than travelling to an API that rejects it. The
  automatic session-start reflect is unaffected: it always uses `"low"` to fit its hook window.
- Both are forwarded from **both** call sites, and env-settable
  (`HINDSIGHT_REFLECT_TOOL_TIMEOUT_MS`, `HINDSIGHT_REFLECT_BUDGET`).

Note the default alone fixes the reported symptom for users who never touch config — 120s was
below the server's own wall timeout, so unconfigured installs were aborting mid-synthesis too.

## Not recurring

`HindsightClient.reflect()`'s `timeoutMs` is now **required**. The right deadline differs by an
order of magnitude between the hook (25s) and the tool (minutes), so there is no sensible value to
fall back to silently — which is exactly how the 120s got applied to a call it was never meant for.
`buildKnowledgeTools` also defaults the tool's window itself, so a future third call site cannot
reintroduce a short deadline by omission.

## Tests

- `knowledge-tools.test.ts` — the configured timeout/budget reach the client; the timeout is
  never left unset (asserted `> 300_000`, i.e. above the server's wall timeout).
- `mcp-server.test.ts` + `runtime.test.ts` — both call sites forward the resolved config.
- `config.test.ts` — the default, inheritance from a raised `reflectTimeoutMs`, that a short one
  does not lower it, explicit override, env fallbacks, unknown-budget fallback.

580 pass, `tsc --noEmit` clean, `./scripts/hooks/lint.sh` clean.

## Docs

README reference table, `skill/SKILL.md`, and the generated `docs-integrations` + docs-skill
mirrors. `reflectTimeoutMs`'s row now says **automatic** session-reflect, which is all it ever
controlled.


jonrzhang (commented):
LGTM

PR #3598, opened 2026-08-18
fix(cli): suppress ANSI color and spinner when output is not a TTY

yossiovadia:
## Problem

The `hindsight` CLI writes truecolor gradient escapes and an animated spinner (80ms/frame) to **stdout** with no terminal detection, and no `NO_COLOR` / `TERM=dumb` handling. Any piped or redirected invocation gets flooded with escape codes — and the README documents coding agents as the primary consumers, which read piped output by definition.

Default output mode is `pretty` (so this hits the default `hindsight <cmd>` an agent runs, not just interactive use).

## Root cause

`hindsight-cli/src/ui.rs` has no `is_terminal()` check anywhere. `gradient*/dim` build raw `\x1b[38;2;…m` strings, `GradientSpinner` unconditionally spawns a thread that `print!`s frames to stdout, and the `--help` logo is raw ANSI art.

## Fix

One central predicate — `ansi_enabled()` = `stdout().is_terminal() && NO_COLOR unset && TERM != "dumb"` (stdlib `IsTerminal`, no new dependency) — applied to the color helpers, the spinner (now a no-op: no thread, no output when disabled), and the `--help` logo. Auto-detection only: no new flags, no `-o` output-format changes. The decision layer is split into pure, injectable functions so it's unit tested without depending on the harness's terminal.

## Real before/after

**Live `recall` piped to a file** (the result text itself is gradient-colored):

```
BEFORE:  58,833 bytes, 2,254 ANSI escape sequences
         "  \e[38;2;128;128;128mResult #1\e[0m"
AFTER:   23,141 bytes,     0 ANSI escape sequences
         "  Result #1"
```

**Spinner animation flood** (a 5s call, stdout piped):

```
BEFORE:  21,960 bytes / 1,260 escape sequences
AFTER:        0 bytes /     0 escape sequences
```

**`NO_COLOR=1` and `TERM=dumb`**: ignored before (byte-identical flood), respected after (0 bytes).

**Interactive terminal unchanged** — under a real PTY the output is byte-identical before and after (spinner still animates, results still colored):

```
PTY (TTY):  BEFORE 13,542 bytes / 777 ESC   ==   AFTER 13,542 bytes / 777 ESC
```

## Tests

Adds unit tests in `ui.rs` (matching the existing `#[cfg(test)]` style) covering the predicate's four env/TTY combinations, the color helpers in both states, and the spinner no-op vs animate paths. The decision logic is injected, so the tests are deterministic under both `cargo test | cat` and a real terminal. Full CLI suite green.

## Scope

Single concern, `hindsight-cli/` only. Not touched: `print_error` (already goes through the `colored` crate's own NO_COLOR/TTY detection) and the `indicatif` progress bar (already auto-hides on non-TTY). The spinner also writes to stdout rather than stderr since a past refactor (`fa554b898`, indicatif → `GradientSpinner`); happy to move it back to stderr in a follow-up if preferred, but it's out of scope here.


koriyoshi2041 (approved):
Verified at `93c47efb63`. The stdout TTY/`NO_COLOR`/`TERM=dumb` gate is applied consistently to the raw gradient helpers, help logo, and spinner thread, while leaving interactive behavior intact. The CLI test suite passes locally (126/126), redirected `--help` contains 0 ANSI escapes, and the PR diff is whitespace-clean.

Issue #3599, opened 2026-08-19
ollama: HINDSIGHT_API_LLM_OLLAMA_NUM_CTX is ignored on unstructured calls — verify_connection reloads the model at the server default (0.9.1)


thomasthinks:
Running `0.9.1` against a self-hosted Ollama on a **shared, pinned** GPU rig. `HINDSIGHT_API_LLM_OLLAMA_NUM_CTX` is honored for extraction/consolidation but **silently ignored for every unstructured call**, including the startup `verify_connection()` probe. On a shared Ollama that is not cosmetic: because Ollama keys a loaded model instance by context size, the ignored call **reloads the model at the server default**, re-tuning it for every other consumer of that host.

## Root cause

`hindsight_api/engine/providers/openai_compatible_llm.py`

The native path is the **only** place `num_ctx` is attached (~line 1571):

```python
options: dict[str, Any] = {
    "num_batch": 512,
}
if self.ollama_num_ctx is not None:
    options["num_ctx"] = self.ollama_num_ctx
```

…but that path is entered only when a `response_format` is supplied (~line 884):

```python
# Handle Ollama with native API for structured output (better schema enforcement)
if self.provider == "ollama" and response_format is not None:
    return await self._call_ollama_native(...)
```

`verify_connection()` calls `self.call(...)` with **no** `response_format`:

```python
await self.call(
    messages=[{"role": "user", "content": "Say 'ok'"}],
    max_completion_tokens=self._verification_max_completion_tokens(),
    ...
    scope="verification",
)
```

so it falls through to the OpenAI-compatible `/v1/chat/completions` request, where `num_ctx` has no representation and Ollama applies its server default.

## Steps to reproduce

1. Point Hindsight at an Ollama host: `HINDSIGHT_API_LLM_PROVIDER=ollama`, `HINDSIGHT_API_LLM_BASE_URL=http://<host>:11434/v1`, `HINDSIGHT_API_LLM_MODEL=qwen3:14b`.
2. Set `HINDSIGHT_API_LLM_OLLAMA_NUM_CTX=24576` (any value differing from the Ollama server default).
3. Warm the model at the configured size, e.g. `curl .../api/generate -d '{"model":"qwen3:14b","prompt":"hi","options":{"num_ctx":24576}}'`. Confirm `GET /api/ps` reports `context_length: 24576`.
4. Start (or restart) the Hindsight container.
5. `GET /api/ps` again.

[...]

PR #3601, opened 2026-08-19
Propagate Ollama num_ctx to unstructured requests

r266-tech:
The OpenAI-compatible Ollama path omitted the configured `ollama_num_ctx` from unstructured requests, including `verify_connection()`, allowing those calls to use Ollama's server-default context instead of the configured value.

## Changes

- Add the configured value to `extra_body.options.num_ctx` for unstructured Ollama calls.
- Preserve an explicit caller-provided `options.num_ctx`, leave unset behavior unchanged, and avoid mutating nested configured options.
- Add regression coverage for ordinary unstructured calls, `verify_connection()`, caller precedence, unset behavior, and the existing native path.

## Verification

- Targeted Ollama regression tests: 3 passed.
- Ruff lint check on both changed files: passed.
- Ruff format check on both changed files: passed.
- `ty` check on the changed production module: passed.
- `git diff --check`: passed.

Addresses #3599.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -839,6 +839,13 @@ class OpenAICompatibleLLM(LLMInterface):
         """Apply provider-specific extra_body defaults while preserving user overrides."""
         if self.provider == "minimax":
             extra_body.setdefault("thinking", {"type": "disabled"})
+        if self.provider == "ollama" and self.ollama_num_ctx is not None:
+            # ``extra_body`` is shallow-copied per call, so copy nested options
+            # before adding the default to avoid mutating the configured payload.
+            options = extra_body.get("options")
+            options = dict(options) if isinstance(options, dict) else {}
+            options.setdefault("num_ctx", self.ollama_num_ctx)
+            extra_body["options"] = options
 
     async def call(
         self,
--- a/hindsight-api-slim/tests/test_llm_wrapper.py
+++ b/hindsight-api-slim/tests/test_llm_wrapper.py
@@ -1,3 +1,6 @@
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
+
 import pytest
 
 from hindsight_api.engine.llm_wrapper import create_llm_provider, sanitize_llm_output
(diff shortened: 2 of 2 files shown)


koriyoshi2041 (changes requested):
The implementation and focused tests look sound: I ran all 49 `test_llm_wrapper.py` cases at `fb97d8d`, and the configured/default/caller-override paths passed.

One user-facing contract needs to move with the fix. Both `hindsight-docs/docs/developer/configuration.md` and the version-0.9 copy still describe `HINDSIGHT_API_LLM_OLLAMA_NUM_CTX` as an override "for structured-output calls." After this change it also governs unstructured calls and startup verification, which is the central behavior fixed here. Could you update the current and versioned descriptions so operators no longer infer the old scope?

nicoloboschi:
Closing this — the change is clean, but it doesn't reach Ollama, so #3599 stays open.

`extra_body["options"]["num_ctx"]` is serialized into the `/v1/chat/completions` body, and Ollama's OpenAI-compat handler decodes that body into `ChatCompletionRequest`, which has no `options` field. Its fields are `Model, Messages, Stream, StreamOptions, MaxTokens, Seed, Stop, Temperature, FrequencyPenalty, PresencePenalty, TopP, ResponseFormat, Tools, Reasoning, ReasoningEffort, Logprobs, TopLogprobs, DebugRenderOnly`, and `FromChatRequest()` maps only `MaxTokens→num_predict`, `Temperature`, `Seed`, `FrequencyPenalty`, `PresencePenalty`, `TopP` and `Stop` into the native options map — there is no `num_ctx` mapping. Go drops unknown JSON keys silently, so the new field produces no error and no effect: `verify_connection()` still reloads the model at the server default. See ollama/ollama#6544 and ollama/ollama#6137 (unmerged).

Our own docs already say this — `hindsight-docs/docs/developer/configuration.md`:

> Generation parameters (`seed`, `top_p`, `top_k`, `num_ctx`, ...) live under Ollama's `options` object, so nest them ... On the OpenAI-compatible endpoints these same params are top-level instead.

The tests pass because they assert on what was handed to the SDK (`create.call_args.kwargs["extra_body"]`), which is a mock-shape assertion standing in for a server-contract claim.

The endpoint that does honour `num_ctx` is native `/api/chat`, which `_call_ollama_native` already builds correctly. The fix is to give that path a text mode and route unstructured Ollama calls (including the `verify_connection()` probe) through it. Doing that in a follow-up.

PR #3605, opened 2026-08-19
fix(coding-agents): timestamp the aggregated git-log document so its facts get occurrence dates (#3602)

### LINKED ISSUE #3602: feat(coding-agents): pass commit timestamp on git history ingestion so memories get occurred_start dates
## Problem

With `gitIngest: "message"` (the default), the `ingestGitLog` function retains an aggregated commit-message document **without a `timestamp`** parameter. Since the retain API defaults `timestamp` to "now" at ingestion, the LLM fact-extraction prompt receives `Event Date: Unknown` and extracts no temporal references from the content. This results in memories with `occurred_start = null` and `occurred_end = null`.

On a bank with ~590 memories, **none** have occurrence dates. This means:

- **Temporal search** ("what happened last week?") cannot find these memories at all
- **Recency scoring** gives all memories a neutral 0.5 boost (no recency advantage for newer facts)
- **Observation freshness** — derived observations can't inherit temporal bounds from undated sources (related: #3477)

## Root Cause

In `hindsight-integrations/coding-agents/src/core/git.ts`, `ingestGitLog()` calls `client.retain()` without passing `timestamp`:

```typescript
// git.ts — ingestGitLog (default path, gitIngest: "message")
await client.retain(
  text,
  `git commit-message history (last ${n}) for ${repoName}`,
  `gitlog:${repoName}`,
  tags,
  "gitlog",
  { metadata: stamp?.metadata }
  // <-- no timestamp parameter
);
```

Compare with `retainCommit()` (full diff path, `gitIngest: "full"`) which correctly passes the author date:

```typescript
// git.ts — retainCommit (opt-in path, gitIngest: "full")
await client.retain(
  content,
  `git commit in ${repoName}`,
  `git:${sha}`,
  [...],
  "git",
  {
    timestamp: aISO,  // <-- author date (correct)
    metadata: { ... }
  }
);
```

## Proposed Fix

In `ingestGitLog()`, derive a timestamp from the most recent commit's author date and pass it to `client.retain()`. The HEAD commit's author date is the natural anchor — it represents "when this history ends":

```typescript
// Get HEAD author date for the timestamp
const headSha = gitHeadSha(repo);
let headAuthorDate: string | undefined;
if (headSha) {
  const parts = git(repo, "show", "-s", `--format=%aI`, headSha).trim();
  headAuthorDate = parts || undefined;
}

await client.retain(
  text,
  `git commit-message history (last ${n}) for ${repoName}`,
  `gitlog:${repoName}`,
  tags,
  "gitlog",
  {
    timestamp: headAuthorDate,  // anchor the LLM prompt
    metadata: stamp?.metadata,
  }
);
```

This gives the LLM an `Event Date:` anchor so it can resolve relative temporal references in commit messages (e.g., "yesterday", "last week") and populate `occurred_start`/`occurred_end` on extracted memories.

## Impact

- Affects **all coding-agent integrations** (OpenCode, Claude Code, Codex, Cursor, etc.) — they all share this git ingestion code
- Default path only (`gitIngest: "message"`) — the `full` path already passes timestamps correctly
- Low risk: the `timestamp` parameter is purely additive to the LLM extraction prompt; existing behavior is unchanged for content without temporal references

## Version

Observed on `@vectorize-io/hindsight-coding-agents` installed via `npx`, backed by self-hosted Hindsight 0.9.x.

nicoloboschi:
Fixes #3602.

## Problem

`ingestGitLog` — the **default** git-ingestion path (`gitIngest: "message"`) shared by every
coding-agent integration — retained the aggregated commit-message document with no `timestamp`.
Retain then stamped "now", the fact-extraction prompt received `Event Date: Unknown`, and every
fact extracted from those commit messages landed with `occurred_start = null` /
`occurred_end = null`: invisible to temporal search, and neutral (0.5) for recency scoring.

The `gitIngest: "full"` path (`retainCommit`) already passed the commit's author date.

## Fix

`ingestGitLog` now anchors the document on the author date of the newest commit it contains.

One deviation from the issue's proposed patch: it suggested `gitHeadSha` + `git show -s --format=%aI`.
That misdates the document whenever HEAD is a **merge** commit, because `gitLogText` builds the
content with `--no-merges` — the merge isn't in the document at all. The new
`gitLogNewestAuthorDate()` helper uses the same `--no-merges` traversal
(`git log -n 1 --no-merges --format=%aI`), so the timestamp is always the date of a commit the
document actually contains. It returns `null` on an empty repo / non-repo / any git error, in which
case the timestamp is omitted exactly as before.

The two duplicated `client.retain(...)` branches (metadata vs. no metadata) collapse into one call
using the `metadata: Object.keys(...).length ? … : undefined` idiom already used in `deepen.ts`.

## Tests

`src/core/git.gitlog.test.ts` (10 passing):

- timestamp equals the newest commit's author date — two commits with pinned `GIT_AUTHOR_DATE`,
  asserts the newer one wins
- `gitLogNewestAuthorDate` returns `null` for a repo with no commits
- the existing attribution assertion relaxed from `toEqual` to `toMatchObject`, since retain opts
  now also carry `timestamp`

## Parity check

Every other retain call site that ingests dated content already passes a timestamp (`retainCommit`,
the chat/transcript paths); the survey-baseline marker deliberately doesn't, since no memories are
extracted from it. `ingestGitLog` was the only gap.

## Risk

Low and additive — `timestamp` only feeds the extraction prompt's `Event Date:` anchor. The
document id is unchanged (`gitlog:<repoName>`), so the next deepen re-upsert replaces the existing
document rather than duplicating it, and picks up the anchor.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/git.gitlog.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/git.gitlog.test.ts
@@ -4,7 +4,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { HindsightClient } from "./hindsight";
-import { gitLogText, ingestGitLog, repoNameOf, retainCommit } from "./git";
+import { gitLogNewestAuthorDate, gitLogText, ingestGitLog, repoNameOf, retainCommit } from "./git";
 
 let dir: string;
 
@@ -38,6 +38,10 @@ describe("gitLogText", () => {
 
   it("returns empty string for a repo with no commits", () => {
     expect(gitLogText(dir, 10)).toBe("");
+  });
+
+  it("gitLogNewestAuthorDate returns null for a repo with no commits", () => {
+    expect(gitLogNewestAuthorDate(dir)).toBeNull();
   });
 });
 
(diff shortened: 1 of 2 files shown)


seppaleinen:
Great! Thanks for this! Do you have some idea on how to make this change work for already ingested data?

PR #3615, opened 2026-08-19
fix(knowledge): cast the mental_models.name bind for vchord writes

valda:
## Problem

With `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=vchord`, every write that sets `mental_models.name` fails with HTTP 500:

```
$ curl -sX POST .../v1/default/banks/$BANK/knowledge-base/pages \
    -H 'content-type: application/json' \
    -d '{"name":"Rejected options","source_query":"..."}'
{"detail":"inconsistent types deduced for parameter $3\nDETAIL:  text versus character varying"}
```

Affected endpoints:

| Endpoint | Engine method |
|---|---|
| `POST /knowledge-base/pages` | `_insert_pinned_mental_model` |
| `POST /mental-models` | `_insert_pinned_mental_model` |
| `PATCH /mental-models/{id}` (with `name`) | `update_mental_model` |
| `PATCH /knowledge-base/nodes/{id}` (page rename) | `rename_knowledge_node` |

In practice this means **knowledge pages cannot be created at all on a vchord deployment**.

## Cause

`mental_models.name` is `VARCHAR`, but the same bind parameter is reused inside the `search_vector` expression that `pg_search_vector_expr` builds, where `tokenize()` takes `TEXT`:

```sql
INSERT INTO mental_models (..., name, ..., search_vector)
VALUES (..., $3, ..., tokenize(COALESCE($3, '') || ' ' || COALESCE($5, ''), 'llmlingua2')::bm25_catalog.bm25vector)
--            ^^ varchar          ^^ text
```

PostgreSQL infers a single type per parameter, so the statement never reaches execution — it is rejected at prepare time. Minimal repro, no Hindsight involved:

```sql
CREATE TEMP TABLE t (a varchar, b text);
PREPARE p AS INSERT INTO t (a, b) VALUES ($1, upper($1));
-- ERROR:  inconsistent types deduced for parameter $1
-- DETAIL:  text versus character varying

PREPARE p AS INSERT INTO t (a, b) VALUES ($1::text, upper(COALESCE($1, '')));
-- PREPARE
```

Only vchord is affected: it is the one backend that writes `search_vector` inline, so it is the only one that binds the name twice. `native` has a `GENERATED` column and `pgroonga` / `pg_search` / `pg_textsearch` index the base columns, so `pg_search_vector_expr(..., native_inline=False)` returns `None` for all of them and the second reference never appears.

## Fix

Cast the column assignment to `::text`. That pins the parameter's type, and `VARCHAR` accepts a text value on assignment. Applied to the three writes that feed `mental_models.name` into `pg_search_vector_expr`:

- `_insert_pinned_mental_model` — `VALUES (..., $3::text, ...)`
- `update_mental_model` — `name = $N::text`
- `rename_knowledge_node` — `SET name = $3::text`

`memory_units.text` / `.context` (`writes.py:apply_edit`) and `knowledge_pages.name` are already `TEXT`, so those statements need no change and none was made.

An alternative fix is a migration widening `mental_models.name` to `TEXT`, which would remove the footgun rather than work around it. Happy to switch if you prefer that — the cast is the smaller, backport-friendly change.

## Tests

`tests/test_knowledge_bm25_dispatch.py` gains two cases that assert on the generated SQL, matching the module's existing approach ("They need no live extension because they assert on the generated SQL"). Reproducing the actual failure needs a live vchord Postgres, which CI does not provision.

Both new tests fail on `main` and pass with the fix:

```
tests/test_knowledge_bm25_dispatch.py::test_pinned_insert_casts_the_name_bind PASSED
tests/test_knowledge_bm25_dispatch.py::test_pinned_insert_keeps_the_cast_without_inline_tokenization PASSED
11 passed
```

`ruff check` / `ruff format --check` / `ty check` clean.

## Verification against a live vchord deployment

Verified on 0.9.1 against `tensorchord/vchord-suite` with `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=vchord`. Before the patch all four endpoints above 500'd; after it, page create returned `201` and completed its background refresh, and both rename paths returned `200`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -12749,6 +12749,15 @@ class MemoryEngine(MemoryEngineInterface):
         """Insert a pinned model using the caller's transaction."""
         # VectorChord needs mental_models.search_vector tokenized on write; every
         # other backend either generates it or indexes the source columns.
+        #
+        # $3 is bound twice: to mental_models.name (VARCHAR) and, inside sv_expr,
+        # to tokenize() (TEXT). PostgreSQL infers one type per parameter, so the
+        # two sites make the statement fail at prepare time with "inconsistent
+        # types deduced for parameter $3: text versus character varying". Casting
+        # the column assignment pins the parameter to text; VARCHAR accepts a text
+        # value on assignment. Every write that feeds mental_models.name into
+        # pg_search_vector_expr needs the same cast (see update_mental_model and
+        # rename_knowledge_node).
         sv_expr = pg_search_vector_expr(
             get_config(), text_col="$3", context_col="$5", signals_col=None, native_inline=False
         )
@@ -12758,7 +12767,7 @@ class MemoryEngine(MemoryEngineInterface):
             f"""
             INSERT INTO {fq_table("mental_models")}
             (id, bank_id, subtype, name, description, source_query, content, embedding, tags, max_tokens, trigger{sv_col})
-            VALUES ($1, $2, 'pinned', $3, ' ', $4, $5, $6, $7, COALESCE($8, 2048), COALESCE($9, '{{"refresh_after_consolidation": false}}'::jsonb){sv_val})
+            VALUES ($1, $2, 'pinned', $3::text, ' ', $4, $5, $6, $7, COALESCE($8, 2048), COALESCE($9, '{{"refresh_after_consolidation": false}}'::jsonb){sv_val})
             RETURNING id, bank_id, name, source_query, content, tags,
                       last_refreshed_at, last_memory_seen_at, created_at, reflect_response,
                       max_tokens, trigger, structured_content
@@ -13824,7 +13833,11 @@ class MemoryEngine(MemoryEngineInterface):
             content_sql = "content"
 
             if name is not None:
-                updates.append(f"name = ${param_idx}")
+                # ::text pins the parameter's type — it also feeds the tokenize()
+                # call in sv_expr below, and a bind deduced as both VARCHAR (the
+                # column) and TEXT (the function argument) is rejected at prepare
+                # time. See _insert_pinned_mental_model for the full rationale.
+                updates.append(f"name = ${param_idx}::text")
                 params.append(name)
                 name_sql = f"${param_idx}"
                 param_idx += 1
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
Verified the diagnosis rather than taking it on trust — reproduced the failure against a live Postgres 16 with asyncpg using the same statement shapes (a bind landing in a `varchar` column and inside a text-arg function): both the INSERT and UPDATE forms raise `AmbiguousParameterError: inconsistent types deduced for parameter $2 (text versus character varying)`, and both are fixed by the `::text` cast on the column assignment.

Checks that back the change:

- Column types are as described: `mental_models.name` is `VARCHAR(256)` (`h3c4d5e6f7g8_mental_models_v4.py:75`), while `content`, `memory_units.text`/`.context` and `knowledge_pages.name` are all `TEXT` — so `name` is genuinely the only double-bound non-TEXT column.
- Write-site coverage is complete: of the `pg_search_vector_expr` callers, only these three bind `name`. `clear_mental_model`, `importer._apply_mental_model_derived_state` and both `memory_units` paths reference columns, not binds, and correctly needed no change.
- The cast doesn't widen the field: a 300-char value passed as `$N::text` still raises `StringDataRightTruncationError`, so the `VARCHAR(256)` cap is preserved.
- Both new tests fail when the insert cast is reverted and pass with it; `ruff check` / `ruff format --check` clean.

Agree with keeping the cast over the migration alternative: widening to `TEXT` would drop the 256-char cap that nothing else enforces, and would need an Oracle half in the same migration. Casting at the assignment is also the only placement that works — pushing it into the tokenize side (`text_col="$3::text"`) would still conflict with the varchar column context.

Two non-blocking notes:

1. Only the insert cast is regression-guarded; the `update_mental_model` and `rename_knowledge_node` casts can be silently dropped later. A line in `pg_search_vector_expr`'s docstring — a bind passed as `text_col` must be cast at its column assignment when the target column isn't `TEXT` — would put the warning where the next person adding a write path will look.
2. No CI job exercises vchord, which is why this shipped broken. Out of scope here, but the class of bug will recur.

Note the API test shards were skipped on this PR (they're gated on `has_secrets`, unavailable to fork PRs). I ran `tests/test_knowledge_bm25_dispatch.py` locally: 11 passed.

PR #3634, opened 2026-08-19
fix(retain): bound structured extraction schemas

devinkuhn:
## Summary
- bound retain fact arrays and nested arrays/strings in strict JSON schemas
- preserve tighter existing constraints without mutating Pydantic schemas
- apply the bounded retain schema to interactive and batch extraction paths only

## Production evidence
Unbounded strict schemas let vLLM/Nemotron generate until the completion cap and truncate invalid JSON. The same three historical pathological prompts passed 3/3 with bounded schemas: valid JSON, finish_reason=stop, 1/5/5 facts, 9.7–63.9 seconds.

## Verification
- 15 focused regression tests passed
- ruff check and format passed
- py_compile passed
- git diff --check passed

No deployment is included.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -6,6 +6,7 @@ Uses the LLMConfig wrapper for all LLM calls.
 """
 
 import asyncio
+import functools
 import json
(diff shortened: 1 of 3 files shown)


nicoloboschi:
this is problematic on some llms and actually adding more rules is generally bad for mid models, closing

PR #3639, opened 2026-08-19
fix(config): validate and persist bank config as one serialized unit (#3037)

### LINKED ISSUE #3037: Make bank config validation and persistence atomic
## Problem

Bank config validation and persistence are separate operations. `validate_bank_config_updates()` loads the current bank overrides and validates the projected final configuration, then `_persist_bank_config()` later opens another connection and merges the normalized JSON into `banks.config`.

Two requests can therefore read the same old overrides, each validate successfully in isolation, and then both persist. The final merged configuration may violate a cross-field constraint even though neither request validated that final state. The existing `projected_bank_overrides` support only accounts for default-template config that the same request is about to apply to a missing bank; it does not protect against concurrent requests.

## Affected code

- `hindsight-api-slim/hindsight_api/config_resolver.py`: `validate_bank_config_updates()`, `_persist_bank_config()`
- `hindsight-api-slim/hindsight_api/engine/memory_engine.py`: `_validate_bank_config_updates()`, `update_bank()`, `update_bank_config()`

## Impact

Concurrent updates to the same bank can persist a configuration combination that was never validated. This weakens every cross-field invariant enforced against the resolved configuration and makes the result depend on request interleaving.

## Design decision required

Choose and document one concurrency model for config writes to the same bank:

1. Serialize them by locking the bank row and performing load, validation, and persistence in one transaction; or
2. Add an optimistic revision/version and use a conditional update, returning 409 when the revision changes.

Last-write-wins is only acceptable if cross-field constraints are revalidated against the state that is actually committed.

## Requirements

- Support both PostgreSQL and Oracle.
- Do not use advisory locks.
- Never commit a final override combination that has not passed validation.
- Define whether conflicts return 409 or are retried.
- Call authorization/config hooks exactly once; an optimistic retry must not silently repeat quota-consuming or time-sensitive hooks.
- Preserve the current missing-bank default-template projection semantics.

## Suggested tests

- Use events or a barrier, not sleeps, to interleave two updates that are each valid against the old state but invalid when combined.
- Assert that the final overrides remain valid.
- Assert the documented conflict response and exact hook counts.
- Cover PostgreSQL and Oracle behavior.

nicoloboschi:
Closes #3037.

## The problem is real, and reachable through the public API

`validate_bank_config_updates()` loaded the current overrides on one connection and `_persist_bank_config()` merged the update on another, with nothing in between. Two requests could each validate against the same old state and both commit a union neither had validated.

Reproduced end to end before writing any fix (barrier, no sleeps):

- start: `retain_chunk_size=3000`, `retain_strategies={"s": {"retain_chunk_size": 1000}}`
- request A sets `retain_chunk_size=100000` — valid, because strategy `s` pins its own chunk size
- request B sets `retain_strategies={"s": {"retain_extraction_mode": "detailed"}}` — valid against the old base of 3000

Both succeeded, and `retain_batch_async(..., strategy="s")` on that bank then raised `retain_max_completion_tokens (64000) must be greater than retain_chunk_size (100000)`. Applied **sequentially**, the second update is correctly rejected — the interleaving is what defeated the check.

## The fix

Persistence opens a transaction, takes `SELECT … FOR UPDATE` on the bank row, re-validates the update against the overrides actually committed at that moment, and only then merges. Writers to one bank serialize on that row.

Design choices worth reviewing:

- **Pessimistic lock, not an optimistic revision.** An optimistic retry would have to re-run validation, and validation calls the tenant permission hook, which may reserve quota or make a time-sensitive decision. The issue requires those hooks fire exactly once — here they run once, before the transaction opens. `ValidatedBankConfigUpdate` carries the already-resolved global+tenant base into the locked path so nothing is re-resolved under the lock.
- **No 409, no retry.** The loser of a race gets the same `400` it would have got had the two updates arrived in order. Documented in `configuration.md`.
- **Only cross-field-constrained updates re-check.** An update touching none of `retain_chunk_size` / `retain_structured_chunk_size` / `retain_strategies` / `recall_budget_min` / `recall_budget_max` cannot change either side of any constraint, so it merges as before and independent concurrent field writes both survive.
- **Missing-bank default-template projection is preserved** — and strengthened: the template is applied to the row before persistence runs, so the re-check validates against the real committed state rather than a projection.

`update_bank`'s `except ValueError → 404` is narrowed to `BankConfigPersistenceConflictError`, so a genuine write-time validation failure stays a 400 instead of being reported as "bank not found".

## Also fixed: a hole locking alone would not close

`recall_budget_min <= recall_budget_max` was only checked when both bounds arrived in the same update. Sequentially setting `recall_budget_max=100` and then `recall_budget_min=500` was accepted, no concurrency needed. Both bounds are now validated against the projected final configuration.

## Tests

`tests/test_bank_config_atomicity.py` — barrier-forced interleavings for both constraint families, a test that unconstrained concurrent writes still both land, and a hook-count test asserting each tenant hook fires exactly once across validate + persist. The three behavioral tests were confirmed to fail on the unfixed code and pass on the fix.

The same three scenarios run against Oracle 23ai via `TestBankConfigAtomicity` in `test_oracle_integration.py` (sharing the helpers, so both dialects run identical interleavings). Adding the `oracle-tests` label so that job runs.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config_resolver.py
+++ b/hindsight-api-slim/hindsight_api/config_resolver.py
@@ -11,7 +11,7 @@ multiple API servers.
 import asyncio
 import json
 import logging
-from dataclasses import asdict, fields, replace
+from dataclasses import asdict, dataclass, fields, replace
 from functools import lru_cache
 from types import UnionType
 from typing import TYPE_CHECKING, Any, Union, get_args, get_origin
@@ -42,6 +42,44 @@ class BankConfigPersistenceConflictError(ValueError):
         super().__init__(f"Cannot update config for bank '{bank_id}': the bank does not exist")
 
 
+# Fields that participate in a constraint spanning more than one config field.
+# An update touching any of them is re-checked against the committed overrides
+# while the bank row is locked (see _persist_bank_config), because two updates
+# that are each valid against the old state can combine into an invalid one.
+#
+# An update touching none of them cannot change either side of any of these
+# constraints, so the constrained state it leaves behind is exactly the state it
+# found, and it needs no re-check.
+_CROSS_FIELD_CONSTRAINED_FIELDS = frozenset(
+    {
+        # retain_max_completion_tokens (server-level) > retain_chunk_size, for the
+        # bank config and for every retain strategy spliced onto it.
+        "retain_chunk_size",
+        "retain_structured_chunk_size",
+        "retain_strategies",
+        # recall_budget_min <= recall_budget_max
+        "recall_budget_min",
+        "recall_budget_max",
+    }
+)
+
+
+@dataclass(frozen=True)
+class ValidatedBankConfigUpdate:
+    """Normalized bank overrides plus what is needed to re-validate them at write time.
+
+    ``updates`` is ready to merge into ``banks.config``. ``parent_config`` is the
+    resolved global + tenant base the projected configuration is rebuilt from
+    while the bank row is locked; it is None when the update touches no
+    cross-field constrained field and therefore needs no re-check. Resolving it
+    during validation (rather than during persistence) keeps the tenant config
+    hook on the once-per-request path and off the locked path.
+    """
+
+    updates: dict[str, Any]
+    parent_config: dict[str, Any] | None = None
+
+
 def _validate_retain_strategy_chunking(base_config: HindsightConfig, strategies: Any) -> None:
     """Validate retain strategy chunking with the same semantics as apply_strategy()."""
     if not isinstance(strategies, dict):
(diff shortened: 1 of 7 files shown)


nicoloboschi:
### Oracle run surfaced a pre-existing dialect divergence (not from this PR)

With the atomicity tests moved early in `test_oracle_integration.py` (the Oracle Free container reliably drops every connection partway through Tier 6 — same `DPY-4011` crash as #3500's run on Aug 18), the Oracle job gave a real verdict:

- ✅ `test_concurrent_updates_cannot_persist_invalid_recall_budget_bounds` — the row lock serializes writers on Oracle
- ✅ `test_one_sided_recall_budget_update_is_validated_against_stored_state`
- ❌ `test_concurrent_updates_cannot_persist_invalid_retain_chunking` — *expected*, for a reason worth recording

**`banks.config` merges differently on the two dialects.** `_persist_bank_config` writes `config = COALESCE(config, '{}'::jsonb) || $1::jsonb`, and `_rewrite_pg_to_oracle` (`engine/db/oracle.py:324`) rewrites that to `JSON_MERGEPATCH(...)`:

| | `retain_strategies = {"s": {"retain_extraction_mode": "detailed"}}` |
|---|---|
| PostgreSQL (`\|\|`) | shallow, top-level — **replaces** the whole map, so a nested `retain_chunk_size` pin disappears |
| Oracle (`JSON_MERGEPATCH`, RFC 7396) | deep merge — **keeps** the nested pin |

So the wedge this scenario constructs cannot exist on Oracle, and the test correctly reported that neither update needed rejecting. More generally: on Oracle you cannot remove a nested config key by writing the parent object. That predates this PR and is orthogonal to #3037 — flagging it rather than fixing it here.

The scenario is now PG-only with that reasoning in a comment; Oracle keeps the two recall-budget scenarios, which use top-level scalars that both dialects store identically, so the atomicity property is still covered on both.

PR #3644, opened 2026-08-19
fix(recall): build ANN index before scan depth assertion

### LINKED ISSUE #3628: [BUG] test_the_kill_switch_flips_real_retrieval_depth asserts pre-#3485 index behaviour — test-api (1/3) red on every PR
## Summary

`tests/test_ann_iterative_scan.py::test_the_kill_switch_flips_real_retrieval_depth` fails deterministically on every PR. It asserts that creating a bank builds that bank's per-`(bank, fact_type)` partial vector index — which stopped being true one day before the test was written.

This is not a flake: it reproduces on every run, and it will keep `test-api (1/3)` red on `main` and on every open PR until it is fixed.

## Why it fails

Two changes landed a day apart with contradictory assumptions:

- **edd0d0c5b** (2026-08-18) — *"fix(db): earn per-bank vector indexes by size instead of creating three per bank (#3485) (#3561)"* removed the vector-index DDL from `get_or_create_bank_profile`. The comment it left is explicit:

  > No vector-index DDL here. A fresh bank holds no rows, so it cannot meet the size threshold that earns a per-(bank, fact_type) partial index; the maintenance sweep builds one if and when the bank grows into it.

- **31c1aaf21** (2026-08-19) — *"fix(recall): make the recall budget reach the vector index (#3541)"* added this test, which says the opposite:

  > Creating the bank also builds its per-(bank, fact_type) partial vector index — the same one recall uses — so this exercises the production index, not a stand-in.

The test creates an empty bank, inserts 600 rows with raw SQL, and never runs the maintenance sweep. So no HNSW index is ever built, and the planner falls back to the btree index plus a sort — which is exactly what the assertion is designed to reject:

```
assert "Index Scan" in plan and "Sort" not in plan, f"expected an ANN scan, got:\n{plan}"
E   AssertionError: expected an ANN scan, got:
E     Limit  (cost=214.22..215.22 rows=400 width=24)
E       ->  Sort  (cost=214.22..215.70 rows=590 width=24)
E             Disabled: true
E             ->  Index Scan using idx_memory_units_fact_type on memory_units  (...)
E                   Index Cond: (fact_type = 'world'::text)
```

## Why it was not caught at merge time

`main`'s most recent `test.yml` run is [32142821174](https://github.com/vectorize-io/hindsight/actions/runs/32142821174) at `19318ac0`, which **predates the test file** — `git cat-file -e 19318ac0:hindsight-api-slim/tests/test_ann_iterative_scan.py` reports it does not exist there. So `main` has never actually run this test; the first CI runs to include it are the PRs opened since.

## Suggested fix

Build the index the test is measuring, after the rows exist:

```python
from hindsight_api.engine.vector_index_health import reconcile_bank_vector_indexes

# ... after the INSERTs and ANALYZE
await reconcile_bank_vector_indexes(conn, schema, bank_id)
```

That is the sweep #3485 made responsible for building the index once a bank grows into it, so the test would then exercise the production path as its docstring intends, rather than a path that no longer exists.

Alternatively, drop the two `enable_seqscan`/`enable_sort` GUC pokes and assert on the row counts only — but the plan assertion is there deliberately (to stop the test quietly measuring nothing), so building the index is the truer fix.

Found while working on #3584 / #3610, where it is the sole remaining CI failure.

green3sf:
## Summary

- restore the shipped per-bank vector-index threshold for the ANN depth integration test
- reconcile the bank's real partial vector index after seeding the test rows
- serialize this concurrent-index-DDL test with the existing `vector_index_reconcile` xdist group

## Root cause

#3485 moved per-bank vector-index creation out of bank creation and into the maintenance reconcile path. This test still assumed that creating an empty bank built its partial HNSW index. The test suite also raises the index threshold globally to keep incidental DDL out of parallel tests, so merely invoking the reconcile without restoring the shipped default would still leave the seeded 600-row bank unindexed.

The updated setup restores the production default policy, loads and analyzes the rows, then calls `reconcile_bank_vector_indexes` with the backend's real index clause. The existing plan assertion therefore measures the intended ANN path instead of the fallback B-tree scan plus sort.

## Validation

- `python3 -m py_compile hindsight-api-slim/tests/test_ann_iterative_scan.py`
- `uvx 'ruff==0.14.9' format --check hindsight-api-slim/tests/test_ann_iterative_scan.py`

The full pgvector integration test was not run locally because it requires the repository's database and embedding test environment; GitHub CI will exercise it.

Fixes #3628

### DIFF
--- a/hindsight-api-slim/tests/test_ann_iterative_scan.py
+++ b/hindsight-api-slim/tests/test_ann_iterative_scan.py
@@ -185,8 +185,9 @@ def _near_query_vector(seed: int) -> str:
     return "[" + ",".join(f"{v / norm:.5f}" for v in values) + "]"
 
 
+@pytest.mark.xdist_group("vector_index_reconcile")
 @pytest.mark.asyncio
-async def test_the_kill_switch_flips_real_retrieval_depth(memory, request_context, ann_config):
+async def test_the_kill_switch_flips_real_retrieval_depth(memory, request_context, ann_config, monkeypatch):
     """End to end, through the pool: on, the budget reaches the index; off, it does not.
 
     Both halves matter. On is the fix — with iterative scans off the ground-layer search
@@ -200,14 +201,20 @@ async def test_the_kill_switch_flips_real_retrieval_depth(memory, request_contex
     index are built directly: the property belongs to the index scan, and going through
     retain would drag in extraction and consolidation.
     """
+    from hindsight_api.engine import vector_index_health
+    from hindsight_api.engine.memory_engine import get_current_schema
+    from hindsight_api.engine.retain.bank_utils import _vector_index_clause, get_or_create_bank_profile
     from hindsight_api.engine.search.retrieval import retrieve_semantic_bm25_combined_sql
-    from hindsight_api.engine.retain.bank_utils import get_or_create_bank_profile
     from hindsight_api.engine.task_backend import fq_table
+    from hindsight_api.engine.vector_index_health import reconcile_bank_vector_indexes
 
     bank_id = f"test_iter_scan_{uuid.uuid4().hex[:8]}"
     budget = 400  # deliberately above the standing ef_search of 200
-    # Creating the bank also builds its per-(bank, fact_type) partial vector index —
-    # the same one recall uses — so this exercises the production index, not a stand-in.
+    # The suite raises this threshold out of reach to stop incidental index DDL.
+    # Restore the shipped default here because this test deliberately exercises
+    # the production per-(bank, fact_type) index lifecycle.
+    monkeypatch.setattr(vector_index_health, "qualifies_for_per_bank_index", lambda rows: rows > 0)
+    monkeypatch.setattr(vector_index_health, "should_keep_per_bank_index", lambda rows: rows > 0)
     await get_or_create_bank_profile(memory._backend, bank_id)
     pool = await memory._get_pool()
     probe = _near_query_vector(0)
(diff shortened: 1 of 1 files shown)


nicoloboschi:
Closing as overtaken by #3638, which landed as `94815f5d0`.

The diagnosis here was exactly right, and it named the real conflict between `edd0d0c5b` and `31c1aaf21`. #3638 resolved it from the other side: instead of teaching the test to build the index itself, it made `HINDSIGHT_API_VECTOR_INDEX_MIN_ROWS=0` mean the threshold is *off* rather than "no minimum". At the default a bank gets its three partial indexes in the bank-create transaction again, exactly as before #3561, and the suite-wide threshold override in `conftest.py` is gone.

So the comment this PR replaced is true again:

> Creating the bank also builds its per-(bank, fact_type) partial vector index — the same one recall uses — so this exercises the production index, not a stand-in.

`test_the_kill_switch_flips_real_retrieval_depth` passes on current `main` unmodified — I ran it before closing this.

The patch would also no longer apply cleanly: #3638 removed `qualifies_for_per_bank_index` and `should_keep_per_bank_index` in favour of `per_bank_index_build_bound` / `per_bank_index_keep_bound` / `per_bank_indexes_are_eager`, so both `monkeypatch.setattr` calls would raise `AttributeError` under monkeypatch's default `raising=True`. And `assert index_result.created == 1` would fail too, since bank creation has already built all three indexes by then and the reconcile creates none.

One idea from here that did carry over: `@pytest.mark.xdist_group("vector_index_reconcile")` for tests issuing concurrent index DDL. #3638 relies on that same group, and separately fixed a latent trap in the reconcile suite's own setup helper — a `CREATE INDEX CONCURRENTLY` chosen as a deadlock victim leaves an INVALID index that `IF NOT EXISTS` then skips on retry, so the helper reported success having left the index unusable.

Thanks for the writeup on #3628 — the root-cause analysis is what made #3638 straightforward to scope.

green3sf:
Thanks for the detailed explanation and for validating this against current `main`. The #3638 approach makes sense—restoring eager per-bank indexes when `min_rows=0` fixes the regression at the policy layer and makes the test-side reconcile unnecessary.

I'm glad the root-cause analysis and the `vector_index_reconcile` xdist grouping were useful. I also appreciate you explaining why this patch no longer applies cleanly and closing the loop so thoroughly. Thanks again!

Issue #3647, opened 2026-08-19
[Docs] Add a production 'LLM routing & failover' recipe: cheap memory-processing model distinct from the agent, with per-operation chains


handnewb:
## Summary

Hindsight is designed to run its **own** LLM accounts and to be pointed at a cheap, high-throughput model for background memory processing — independent of whatever model the "main" agent uses for reasoning. The multi-LLM chain (`HINDSIGHT_API_LLM_{1..N}_*` + `HINDSIGHT_API_LLM_STRATEGY`) and the per-operation overrides (`HINDSIGHT_API_RETAIN_LLM_*`, `_REFLECT_*`, `_CONSOLIDATION_*`) already make this fully configurable, but there is no official guide showing how to compose them into a supported deployment. A short **"LLM routing & failover"** recipe section would close that gap.

## Proposed docs addition

A section (e.g. under operations/configuration) covering:

1. **Concept**: why you'd assign a dedicated, cheaper memory-processing model that differs from your main agent, and how Hindsight keeps the two decoupled.
2. **A working chain example** with placeholders — a cheap workhorse primary + mid-tier + top-tier fallbacks and `HINDSIGHT_API_LLM_STRATEGY='{"mode":"failover"}'` (and note `round-robin` + `weights`).
3. **Per-operation override example**: e.g. pin `batch_retain` to the cheap model while letting consolidation use a stronger one.
4. **Gotchas** (currently only discoverable by reading the source):
   - members must be **contiguous from index 1** — scanning stops at the first unset `..._1_PROVIDER`;
   - `batch_retain` is **primary-member-only** (no failover) — see the related feature issue; pin it to a reliable provider;
   - `HINDSIGHT_API_SKIP_LLM_VERIFICATION=true` to skip eager startup verification;
   - per-operation chains fall back to the global chain when unset.

All example values illustrative; API keys shown only as `<placeholder>`.

## Related

- Feature issue: batch_retain has no multi-LLM failover (the pair of this documentation request)
- #1778 (fallback auto-discovery)

handnewb:
Cross-reference: **#3645** tracks the underlying limitation this recipe's gotcha documents — `batch_retain` is bound to the primary member via `MultiLLMProvider.__getattr__` passthrough, so it gets no multi-LLM failover today. If/when #3645 lands, the "pin `batch_retain` to a reliable provider" note here can be relaxed to "failover applies to the batch path too".

[...]

Issue #3645, opened 2026-08-19
MultiLLMProvider attribute passthrough binds _provider_impl / OpenAI Batch-API paths to the primary member only — secondary members can't supply batch capacity


handnewb:
## Summary

`MultiLLMProvider` (hindsight-api 0.8.4) routes LLM calls through `.call()` / `.call_with_tools()` with failover/round-robin, **but the `_provider_impl` attribute and the OpenAI Batch-API path are bound to the primary member only** via attribute passthrough. A secondary member cannot supply batch capacity, and `HINDSIGHT_API_RETAIN_BATCH_ENABLED=true` can hard-fail startup based on the primary's capability alone.

## Root cause

`engine/multi_llm.py:200-204`:

```python
def __getattr__(self, name: str) -> Any:
    # Anything not defined here (provider, model, api_key, base_url,
    # _provider_impl, mock helpers, batch helpers, ...) delegates to the
    # primary member so existing call sites keep working unchanged.
    return getattr(object.__getattribute__(self, "_members")[0], name)
```

`MultiLLMProvider` only defines `call`, `call_with_tools`, `with_config`, `cleanup`, `members`. Everything else — notably `_provider_impl`, `provider`, `model`, `service_tier`, and batch helpers — resolves to `_members[0]` (the **primary**). Confirmed refs:

- `engine/memory_engine.py:2896` — `await self._retain_llm_config._provider_impl.supports_batch_api()`.
- `engine/retain/fact_extraction.py:1292` — `llm_config._provider_impl.openai_service_tier`.
- The OpenAI Batch-API submission path itself goes through `_provider_impl` (primary-only).

By contrast, the per-chunk extraction path uses `.call()` (`engine/retain/fact_extraction.py:1415`), which **does** route through `_dispatch` (`multi_llm.py:106-129`) and fail over across members.

## Consequence

1. **Batch capacity can't fail over / can't come from a secondary member.** If the primary (e.g. a cheap model routed through a provider without batch support) is paired with a secondary that supports batch (e.g. `openai`, `groq`, `gemini`), the secondary is never consulted for batch.
2. **Startup hard-fail mismatch.** With `HINDSIGHT_API_RETAIN_BATCH_ENABLED=true` and a primary that lacks batch, `memory_engine.py:2896` raises `RuntimeError` ("the retain LLM provider … does not support the batch API") even when a secondary member does support it — the config knob is evaluated against the primary only.
[...]

PR #3649, opened 2026-08-19
feat(llm): route batch retain through a batch-capable multi-LLM member

handnewb:
## Summary

Fixes the gap in #3645: `MultiLLMProvider.__getattr__` bound `_provider_impl` (and therefore the OpenAI/Groq Batch-API path) to the **primary member only**, so:

- a secondary member could never supply batch capacity, and
- `HINDSIGHT_API_RETAIN_BATCH_ENABLED=true` hard-failed at startup (`memory_engine.py` raises `RuntimeError`) based on the *primary's* capability alone — even when a secondary member supported batch.

## Changes

- **`engine/llm_wrapper.py`** — add `supports_batch_api()` and `batch_provider_impl()` to `LLMProvider`, delegating to `_provider_impl` (single-provider path unchanged).
- **`engine/multi_llm.py`** — add the same two methods to `MultiLLMProvider`:
  - `supports_batch_api()` returns `True` if *any* member supports batch.
  - `batch_provider_impl()` returns the impl of the **first batch-capable member** (deterministic by declared order; falls back to the primary when none support batch). Deterministic selection keeps submit and crash-recovery resume on the same provider account.
- **`engine/memory_engine.py`** — startup batch validation now calls `_retain_llm_config.supports_batch_api()` (whole chain) instead of `_provider_impl.supports_batch_api()` (primary only).
- **`engine/retain/fact_extraction.py`** — `extract_facts_from_contents_batch_api` resolves the serving impl once and targets it for the whole lifecycle (submit → poll → retrieve); `_build_request_body` now carries the batch member's `model`/`provider`/`service_tier` (not the chain primary's).

## Behavior

- Single-provider deployments: unchanged behavior.
- Multi-LLM chains: batch retain can now run on a secondary batch-capable member (e.g. an `openai`/`groq`/`gemini` fallback behind a non-batch primary), and the startup knob no longer false-negatives.

## Testing

- New `tests/test_multi_llm_batch.py` (5 tests) covering chain batch capability + deterministic member selection.
- Updated `tests/test_batch_api.py` `mock_llm_config` fixture to expose `batch_provider_impl` (the batch path now resolves through it).
- `ruff check` and `ruff format --check` clean on all touched files.
- Ran and passing: `test_multi_llm_batch.py`, `test_multi_llm_provider.py`, `test_multi_llm_config.py`, `test_batch_api_validation.py`, and the mock-based `test_batch_api.py` cases (53 tests green).

Note: the DB-backed `test_batch_api.py` crash-recovery cases error with `initdb: cannot be run as root` in my container (embedded PostgreSQL refuses to run as root) — environmental, unrelated to this change.

## Related

- #3645 (root cause)
- #3647 (docs recipe — the "batch retain is primary-only" gotcha can be relaxed once this lands)

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
+++ b/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
@@ -32,6 +32,7 @@ from ..config import (
 from .cache_affinity import parse_cache_affinity
 from .llm_interface import (
     LLM_TOOL_CHOICE_AUTO,
+    LLMInterface,
     LLMToolChoice,
     LLMToolChoiceMode,
 )
@@ -962,14 +963,18 @@ class LLMProvider:
         """Whether the underlying provider supports the OpenAI/Groq Batch API."""
         return await self._provider_impl.supports_batch_api()
 
-    async def batch_provider_impl(self) -> Any:
-        """Return the implementation used for batch operations.
+    async def batch_provider_impl(self) -> LLMInterface | None:
+        """The implementation serving batch, or ``None`` when it cannot serve one.
 
         Exists so the batch path (``extract_facts_from_contents_batch_api``) can
         target the implementation through the same interface as a multi-LLM chain:
         ``MultiLLMProvider`` returns the first batch-capable member's impl here,
-        while a single provider returns its own.
+        while a single provider returns its own. Returning ``None`` rather than a
+        provider that would reject every batch call keeps the "can it serve one?"
+        answer in one place — the caller raises on ``None``.
         """
+        if not await self._provider_impl.supports_batch_api():
+            return None
         return self._provider_impl
 
     async def call(
--- a/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
+++ b/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
@@ -958,6 +958,20 @@ class LLMProvider:
         """
         await self._provider_impl.verify_connection()
 
+    async def supports_batch_api(self) -> bool:
+        """Whether the underlying provider supports the OpenAI/Groq Batch API."""
+        return await self._provider_impl.supports_batch_api()
+
+    async def batch_provider_impl(self) -> Any:
+        """Return the implementation used for batch operations.
+
+        Exists so the batch path (``extract_facts_from_contents_batch_api``) can
+        target the implementation through the same interface as a multi-LLM chain:
+        ``MultiLLMProvider`` returns the first batch-capable member's impl here,
+        while a single provider returns its own.
+        """
+        return self._provider_impl
+
     async def call(
         self,
         messages: list[dict[str, str]],
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -663,6 +663,34 @@ def _build_llm(
     return MultiLLMProvider([base, *extra], strategy)
 
 
+async def validate_retain_batch_support(
+    retain_llm_config: "LLMConfig | MultiLLMProvider", config: HindsightConfig
+) -> None:
+    """Fail startup when batch retain is enabled but nothing configured can serve it.
+
+    Otherwise the server would silently fall back to sync mode on every retain,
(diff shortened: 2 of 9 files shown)


koriyoshi2041 (changes requested):
The chain-capability and first-capable selection tests pass locally (53/53 focused tests with proxy variables removed), but crash recovery can resume against the wrong provider account. The code resolves batch_provider_impl() before reading result_metadata, then uses that fresh selection to poll an existing batch_id; the stored batch_provider is never consulted. If member order/capability changes across a worker restart, the batch may have been submitted by member B but get polled through member A. Two members using the same provider with different credentials have the same problem, so provider name alone is not a sufficient key. Please persist a stable member/account selector with the batch state and use it on resume (with a clear error if that member is no longer configured), plus a recovery test that submits on a secondary and resumes after the preferred selection changes.

nicoloboschi:
Reviewed and pushed a follow-up commit to this branch — the routing itself was right, and I confirmed it end to end: with a `[deepseek (no batch), openai (batch)]` chain, submit, poll and retrieve all land on the openai member and the body carries that member's model and service tier. What I changed:

**Docs and comments still said the opposite.** `hindsight-docs/docs/developer/configuration.md` stated "**Batch retain** runs on the primary member only", and `multi_llm.py`'s module docstring and `__getattr__` comment said the same. All three now describe the new behaviour, including that batch does not fail over once a member is selected. (The docs edit also has to be mirrored into `skills/hindsight-docs/` via `./scripts/generate-docs-skill.sh`, or `verify-generated-files` fails.)

**Contract.** `batch_provider_impl()` fell back to the primary's impl when nothing could serve a batch, which is why the call site had to re-ask `supports_batch_api()` immediately after resolving. It now returns `None` as the single "cannot serve one" answer, typed `LLMInterface | None` rather than `Any`, and walks members through their public batch surface instead of `member._provider_impl`.

**`_build_request_body`.** Back to direct attribute access: `provider` is set by every `LLMInterface`, and the `provider == "openai"` short-circuit already keeps gemini/anthropic/fireworks impls from touching `openai_service_tier`. The `getattr(..., None)` form would silently drop flex pricing if the field were ever renamed — that's the pattern cc2dd394 removed elsewhere.

**Resume guard.** "Crash-recovery resume re-resolves to the same member" holds only while the chain configuration is unchanged. Add a member (or give the primary batch capacity) between submit and resume and the old code polls an account that has never heard of that `batch_id` — hanging until the wall clock runs out, then reporting a provider error nobody can act on. `result_metadata.batch_provider` was already recorded at submit and only ever logged; a resume now compares it against the member selected and fails loudly on a mismatch, naming both sides.

**Tests.** This is the part that needed the most work:

- Every test in `test_batch_api_validation.py` re-implemented the `if` inline and asserted against its own copy, so the startup change — the headline fix for #3645, consequence 2 — was untested. I moved the check out of the `verify_llm` closure into `validate_retain_batch_support()` and the tests now drive the real function with real providers, including a chain whose only batch-capable member is a secondary.
- Added coverage for the lifecycle running entirely on the secondary, the request body carrying that member's `model`/`service_tier`, and both resume paths (matching provider resumes; mismatched provider raises).
- The `mock_llm_config` fixture's `batch_provider_impl` was a static `AsyncMock`, so it returned an impl even when the test declared the provider batch-incapable. That made `test_batch_api_raises_for_unsupported_provider` pass vacuously — it stopped raising. The fixture now honours the same `None` contract as the real wrapper.

That last one is worth flagging: this is a fork PR, so `test-api (1/3..3/3)` and `test-api-oracle` are **skipped** — the green rollup here never ran any of these tests. I ran the full set locally against an isolated pg0 instance (including the crash-recovery cases that hit `initdb: cannot be run as root` in your container): **69 passed**, plus `ruff check`, `ruff format --check` and `ty check` clean.

koriyoshi2041 (changes requested):
The new mismatch guard correctly stops cross-provider recovery, but the same-provider/different-account case from the review is still unresolved. result_metadata stores only batch_impl.provider, so two OpenAI members with different credentials both serialize as openai; after reordering, recovery accepts the newly selected account and polls the old batch_id through it. Please persist and resolve a stable chain-member/account identity (not the provider type), and add that reorder regression. The docs, capability contract, and real startup tests look good.

PR #3650, opened 2026-08-19
retain: run delta for a store-owned bank, fenced by the store's own compare-and-set

nicoloboschi:
Fixes the store-owned half of delta retain.

## What was wrong

Delta retain was disabled for a store-owned bank, so **every re-ingest took the full-replace path**:
the document's facts were deleted and rebuilt, every chunk re-extracted, and every observation
standing on those facts orphaned — even when the submitted content was byte-identical to what was
stored. Re-ingesting the same document is the common path for a conversation or a file that keeps
being appended to, so this was paid constantly.

## Why the previous attempt was reverted, and what is different

Enabling it before **lost concurrent appends**. `_delta_batch_write_ext` serializes writers on
`SELECT content_hash FROM documents ... FOR UPDATE`; a store-owned bank has no such row, so the
recheck was skipped entirely and parallel appends each planned against the same base and overwrote
each other, with every call returning success.

The fence is now the store's own compare-and-set: the plan reads the document record once — hash,
text and **watermark** together, since the CAS depends on that pairing — and the write commits only
if the document is still at that watermark.

Two things had to be right, and neither was obvious from the outside:

* **The CAS must be the batch's FIRST store write.** `expect_watermark` guards on the namespace's
  WAL head, and the batch's own fact writes move it — putting the guard after them fences the batch
  against *itself*, and a plain **sequential** append fails with
  `required WAL head < 10, but head was 12`. `_store_document_bodies` already runs before the fact
  writes and already accepts `expect_watermark`, so the fence belongs there rather than in a new
  write.
* **It raises `ConcurrentAppendConflict`, not `StoreWriteConflict`.** `_store_document_bodies`
  already translates the store exception into the retain-level one; catching the store type let it
  escape to the caller and turned a losable race into a failed request.

## Verification

`test_concurrent_appends_keep_every_turn` — the test that caught the regression last time — was run
**first**, and passes.

* Store-owned backend: the orphan-observation cluster (3), the same-document concurrency pair (2) and the
  within-budget replacement now pass.
* Postgres baseline: 33/33 unchanged.
* 236 OSS tests across the extension, consolidation and transfer suites pass.

## Not fixed here

`test_oversized_replacement_still_skips_unchanged_chunks` still fails, and it is a different gap:
the oversized (multi-slice) replacement **never enters delta at all** — no delta log is emitted, so
the CAS is not involved. Tracked as a follow-up.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -1033,6 +1033,7 @@ async def _delta_batch_write_ext(
     document_tags,
     document_body_override,
     doc_hash_at_load,
+    doc_watermark_at_load=None,
     new_chunk_metadata: list,
     delta_chunk_map: dict,
     new_chunks_with_contents: dict,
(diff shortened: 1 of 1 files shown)


nicoloboschi:
CI's two failures are both unrelated to this change:

- `test_ann_iterative_scan.py::test_the_kill_switch_flips_real_retrieval_depth` — fails identically on plain `main` (verified in a worktree before merging #3560).
- `test_quality_integration.py::TestDispositionInfluence` — a non-deterministic LLM-quality assertion ("disposition should make response A more skeptical"); nothing here touches dispositions or prompts.

The gate for this PR is `test_concurrent_appends_keep_every_turn`, which is what caught the previous attempt's data loss. It passes, and so does the Postgres baseline at 33/33.

PR #3651, opened 2026-08-19
retain: an oversized replacement diffs the whole document, and every slice may delta

nicoloboschi:
Follow-up to #3650, completing the oversized case.

Re-ingesting an oversized document re-extracted the parts that had not changed, and tombstoned their memories on the way.

**Cause 1 — the diff was taken against the slice.** An oversized retain is split, so `contents` is one slice while `existing_chunks` covers the whole stored document; every chunk merely absent from this slice was classified as REMOVED. Probed on the test's 20-chunk document: `unchanged=1 changed=0 new=0 removed=19`. The complete body is already to hand as `document_body_override`, and diffing against that gives `unchanged=19 changed=1 new=6 removed=0`.

**Cause 2 — and why cause 1 cannot be fixed alone.** Delta was gated on `is_first_batch`, so the later slices did a full extraction of their own content regardless — after fixing the diff they would write facts for chunks the first slice had already handled. Every slice of an oversized item now gets to try: each diffs the same complete body, the first does the real work, and the rest find nothing left to change.

I raised and reverted the half-fix on the issue for exactly this reason before completing it.

## Measured

* Store-owned backend, same nine modules: **6 failed before, 5 after** — the one that moved is exactly `test_oversized_replacement_still_skips_unchanged_chunks`. The other 5 fail identically without this change (they assert through raw SQL on tables a store-owned bank leaves empty).
* Postgres baseline: **75/75**.
* The concurrency gates — `test_concurrent_appends_keep_every_turn` and the same-document pair — stay green, which is what makes changing how many slices write to one document safe at all.
* OSS suites: 266 passed across retain, transfer and extension (one `test_document_transfer` failure is the known cross-module pollution: 43/43 when that file runs alone).

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -1739,7 +1739,13 @@ async def retain_batch(
             return [[] for _ in contents], TokenUsage(), 0
 
     # --- Delta retain: check if we can skip unchanged chunks ---
-    if is_first_batch:
+    #
+    # Every slice of an OVERSIZED item gets to try, not just the first. Each one diffs the same
+    # complete body against what is stored, so the first slice does the real work and the rest find
+    # nothing left to change and fall through to the metadata-only path. Gating on the first slice
+    # alone left slices 2..N doing a full extraction of their own content regardless, which is what
+    # made an oversized replacement re-extract a document it had just diffed correctly.
+    if is_first_batch or document_body_override is not None:
         delta_result = await _try_delta_retain(
             pool,
             embeddings_model,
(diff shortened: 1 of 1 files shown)


r266-tech (commented):
High-severity correctness issue: oversized `update_mode=append` retains still diff against the append payload only, not the complete post-append document. `_split_contents_into_sub_batches` stores the submitted payload as `document_body_override` for every slice, while append handling prepends the stored text only inside the first slice (`is_first_batch`). The delta path then chunks `document_body_override`; for the first slice this omits the historical prefix, and for later slices it is the only body used. Existing prefix chunks can therefore be classified as removed (or the document body can be persisted as only the new payload), causing data loss/tombstoning when callers append to an oversized document.

Required change: build the override from the actual complete post-append body (stored prefix plus submitted content) before running oversized slicing/delta, and pass that same body to every slice; alternatively bypass delta for oversized append slices until that invariant is guaranteed. Add a regression that appends an oversized payload and verifies the original prefix chunks/memories and the combined document text remain present.

The focused regression tests do not verify persisted chunk indices, removed-chunk tombstones, memory-unit counts, or duplicate facts across all slices. `git diff --check` passed, and head binding matched the authority-bound head SHA `c89ae80b2684a544e5bb26f1baebb1719aa7ed26`.

nicoloboschi:

PR #3652, opened 2026-08-19
fix(retain): preserve nested conversation chunks during output retry

kamilfurtak:
## Summary

Defensively unwrap a single-element list containing a list of turn dictionaries before the existing retain output-retry splitter logic.

## Reproduction

On Hindsight 0.9.1, an integration can provide a conversation shaped like:

```json
[[{"role": "user", "content": "..."}]]
```

This is parsed as a one-element list whose element is another list. `_split_chunk_for_output_retry` only handled a standard list of turns, so it returned `None` for this nested/wrapped shape. The caller then logged that it was dropping the sub-chunk.

The fix normalizes only the concrete `list -> list -> dict` shape to the standard turn-list shape, allowing the existing split-and-retry path to process it. It does not flatten arbitrary JSON values.

## Impact

This can cause silent data loss: the retain operation may report `completed` even though the oversized sub-chunk was dropped and its facts were never stored.

## Scope

- Fixes only nested conversation-array handling in the retain splitter.
- Does not change behavior for standard lists of turns.
- Does not change behavior for plain-text chunks.
- Does not include the `maxItems`/saturation guard or strict-schema flag fixes.

## Testing

- `tests/test_fact_extraction_retry.py`: 27 passed
- Regression coverage includes both `[[{turn}, ...]]` list splitting and the exact `[[{single turn}]]` path that unwraps and then splits one turn's content.
- Ruff check and format check passed.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -258,8 +258,8 @@ def _split_chunk_for_output_retry(chunk: str) -> tuple[str, str] | None:
         parsed = None
 
     if isinstance(parsed, list):
-        # Some OpenAI-compatible providers wrap a conversation array once more;
-        # normalize that shape before applying the existing array split logic.
+        # Some integrations wrap a conversation array once more; normalize that
+        # shape before applying the existing array split logic.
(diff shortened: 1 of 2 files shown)


kamilfurtak:
Superseded by #3656, which combines this with the other output-retry fix into a single reviewable change. No content was dropped.

PR #3653, opened 2026-08-19
fix(retain): honor retain strict-schema flag

kamilfurtak:
## Summary

Use the retain-scoped strict-schema flag consistently when building the Batch API request body.

## Reproduction

Hindsight 0.9.1 can resolve:

```text
llm_strict_schema        = false
llm_strict_schema_retain = true
```

The batch retain path used the global flag to choose between `strict_json_schema()` and the ordinary Pydantic schema, but used the retain flag for the request's `strict` field. That sent `strict: true` with a schema built through the non-strict path, so retain-specific strict-subset constraints were absent from the final request body.

## Impact

Retain requests could advertise strict structured output while omitting constraints that were present in the retain strict schema, including array bounds used by downstream safeguards. The batch path could therefore disagree with the streaming retain path.

## Fix

Resolve `llm_strict_schema_retain` once and use it for both:

- selecting `strict_json_schema()` versus the regular schema;
- the JSON-schema request body's `strict` field.

This preserves the intended per-operation override and does not change reflect or consolidation behavior.

## Testing

- `tests/test_fact_extraction_retry.py` and `tests/test_strict_json_schema.py`: 29 passed.
- Regression coverage includes both override directions:
  - `global=false / retain=true` selects the strict schema and sends `strict: true`;
  - `global=true / retain=false` does not call `strict_json_schema()` and sends `strict: false`.
- Ruff check and format check passed.

Found while diagnosing Hindsight 0.9.1 retain behavior. This PR is intentionally separate from the nested-conversation splitter fix and the maxItems/saturation changes.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -1293,12 +1293,11 @@ def _build_request_body(llm_config, config, prompt: str, user_message: str, resp
     # retain-scoped field, which already folds in the global HINDSIGHT_API_LLM_STRICT_SCHEMA
     # fallback, so the batch and streaming paths can't disagree.
     if hasattr(response_schema, "model_json_schema"):
-        schema = (
-            strict_json_schema(response_schema) if config.llm_strict_schema else response_schema.model_json_schema()
-        )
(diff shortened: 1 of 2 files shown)


kamilfurtak:
Superseded by #3656, which combines this with the other output-retry fix into a single reviewable change. No content was dropped.

PR #3657, opened 2026-08-19
fix(retain): bound and recover degenerate fact-extraction output

kamilfurtak:
## Summary

Bounds and recovery for degenerate fact-extraction output. Depends on #3673 — see
below.

### The failure

Against a local OpenAI-compatible backend, the same 2376-token prompt produced
16 384 output tokens with `finish_reason=length` on five attempts and completed
normally in 554 tokens on the sixth. Input size, sampling parameters and chunk
size were all ruled out: the model stochastically degenerates into repeating
structurally valid fact objects, and nothing in the schema tells the grammar to
stop. This repo already documents the same conclusion for the split floor:

> if a chunk this small still overflows the model's output cap, the cause is
> degenerate or looping model output rather than genuinely dense input

### 1. Bound the facts array, and never accept the bound as success

The response schema gains a `maxItems` bound so a capable structured-output
backend stops the array itself. Reaching the bound is **not** treated as a
result — it raises `OutputTooLongError` and enters the existing split path, so the
bound can never silently truncate a genuinely dense chunk. In the worst case a
chunk is split unnecessarily; no facts are lost.

The bound is derived from `retain_max_completion_tokens` rather than fixed, and it
has to land inside a window:

- **above real density**, or dense chunks are split for nothing. Field data from
  168 single-chunk extractions: median 2 facts, p95 10, **max 29**.
- **below the output-token ceiling**, or the schema bound is unreachable and the
  array runs to `finish_reason=length` instead — the case this exists to prevent.

Three quarters of the token-limited count clears both:

```
budget  4096 -> 36 facts   (ceiling ~48, field max 29)
budget  8192 -> 72 facts   (ceiling ~96)
budget 65536 -> 578 facts  (ceiling ~771)
```

Two earlier revisions got this wrong in both directions. One hard-coded 16. The
next halved the budget, giving 24 at a 4096-token ceiling — below the 29 already
reported above, so it would have split honest input. Regression tests now pin the
window from both sides.

The per-fact cost is measured, not assumed. Against a strict `json_schema` — the
shape retain actually sends — four runs across two local models gave 80.1, 82.1,
89.5 and 101.5 output tokens per fact, so the estimate of 85 holds and only the
halving was wrong.

The floor is additionally capped by what the budget can reach: below roughly 2.7k
output tokens a floor of 32 would advertise a `maxItems` the response can never
hit, so such budgets get the token-limited count instead.

**`maxItems` support is backend-dependent and hard to verify from the outside,
which is why the Python-side check stays authoritative.**

xgrammar tracks `minItems`/`maxItems` as a known gap (mlc-ai/xgrammar#160,
vllm-project/vllm#12131, and vllm-project/vllm#12201, where the constraint fails
without falling back to outlines), and xgrammar 0.2.3's compiled binding carries
no array-length keyword among its schema strings. Yet measured end to end against
oMLX 0.6.3rc2 with that same xgrammar build and no fallback logged, a schema with
`maxItems: 8` returned exactly 8 facts where the uncapped schema returned 22.

I cannot tell from outside whether the bound was enforced by the grammar or merely
obeyed by the model, because stating the same cap in the prompt alone also produced
exactly 8. That ambiguity is the point: a caller cannot know whether `maxItems`
binds, so treating it as a guarantee would be unsafe. The saturation check in
`_extract_facts_from_chunk` counts the parsed facts and is provider-independent;
`maxItems` is an optimisation that lets a backend which honours it stop before
spending the tokens.

`retain_max_completion_tokens` and `llm_supports_max_items` are read by direct
attribute access. An earlier revision used `getattr(config, ..., default)`, which
#3610 had just removed everywhere in the engine: `StaticConfigProxy` signals
"bank-configurable" by raising `ConfigFieldAccessError`, a subclass of
`AttributeError`, so a default silently substitutes the global value for the
bank's resolved one. Both fields are required, and
`validate_retain_completion_token_budget` already guarantees the budget is
positive, so there is nothing legitimate for a default to cover.

### 2. Fail instead of dropping

When a chunk cannot be reduced further, the splitter previously returned an empty
result that the caller counted as a successful extraction — the operation reported
`completed` with the sub-chunk's data gone. It now raises.

A failing half propagates rather than being salvaged around. An earlier revision
gathered sub-chunks with `return_exceptions=True` and kept the successful half's
facts, but that contradicts both this PR and the caller: `extract_facts_from_text`
fails the retain if *any* chunk could not be extracted, on the stated grounds that
"partial extraction is not acceptable". Keeping half the facts would have reported
the operation complete with the rest silently missing — the exact failure this
change exists to prevent.

### Behaviour change, stated plainly

This resolves a contradiction already in the tree rather than adding a policy. The
`_MIN_SPLIT_CHUNK_CHARS` comment documents the sub-chunk drop as a deliberate
runaway bound, while `extract_facts_from_text` documents partial extraction as
unacceptable. Both cannot hold. This PR keeps the runaway bound — the recursion
still terminates in the same number of calls — and changes only what happens at
the bound: a loud failure instead of a silent hole in the stored memory.

The cost is real and worth naming: a deployment that currently absorbs these
events silently will start seeing retain operations fail. In the deployment that
motivated this PR that is roughly two per day, against 10 confirmed silent drops
over five days.

## Depends on #3673

That PR is what makes splits succeed for conversation payloads wrapped in an extra
array. This PR increases how often the split path is entered and makes the
unsplittable case fatal, so merging it alone would turn exactly those recoverable
cases into hard failures.

(The dependency was previously written as #3656. That was a copy-paste error from
when both fixes shared a branch — #3656 is the strict-schema flag, already merged.
The nested-array fix is #3673.)

## Tests

```
46 passed
```

Verified end-to-end on a document that previously lost its content: saturation at
the bound → `OutputTooLongError` → recursive split (2747 → 1824 + 922 → …) →
8/2/7/11 facts extracted, every call `finish_reason=stop`, no dropped sub-chunk.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -469,12 +469,27 @@ class VerbatimFactExtractionResponse(BaseModel):
 # this bound is sent through the existing split/retry path, so it can never
 # silently truncate a degenerate or excessively dense chunk.
 #
-# The bound is derived from the configured output budget rather than fixed, so a
-# deployment with a large ceiling is not forced into constant splitting. A fact
-# object costs roughly this many output tokens; halving the budget keeps the item
-# bound biting before the token limit rather than at the same point.
-RETAIN_FACTS_SATURATION_FLOOR = 16
+# The bound has to sit in a window with a floor and a ceiling:
+#
+#   above legitimate density, or dense chunks are split for no reason. Field data
+#   from 168 single-chunk extractions: median 2 facts, p95 10, max 29. A measured
+#   worst case on a deliberately dense ~2 kB chunk produced 21-22 facts.
+#
+#   below the output-token ceiling, or the schema bound never gets to stop the
+#   array and the response comes back truncated at finish_reason=length instead.
+#   At a 4096-token budget that ceiling is around 48 facts.
+#
+# So the bound is three quarters of the token-limited count, not half the budget:
+# an earlier revision halved it and produced 24 at a 4096-token budget, below the
+# observed maximum of 29, so it would have forced splits on honest input.
+#
+# The per-fact cost is measured, not assumed. Against a strict json_schema — the
+# shape retain actually sends — four runs across two models gave 80.1, 82.1, 89.5
+# and 101.5 output tokens per fact, so 85 is a central estimate rather than a
+# guess. Modes that emit fewer fields are cheaper and simply bind later.
+RETAIN_FACTS_SATURATION_FLOOR = 32
 _APPROX_OUTPUT_TOKENS_PER_FACT = 85
+_SATURATION_BUDGET_SHARE = 3, 4  # numerator, denominator
 
 
 def resolve_facts_saturation_limit(config) -> int:
@@ -484,8 +499,15 @@ def resolve_facts_saturation_limit(config) -> int:
     # which subclasses AttributeError, so a default would silently substitute the
     # global value for the bank's resolved one (#3610). The field is required and
     # validated positive by validate_retain_completion_token_budget.
-    derived = config.retain_max_completion_tokens // (2 * _APPROX_OUTPUT_TOKENS_PER_FACT)
-    return max(RETAIN_FACTS_SATURATION_FLOOR, derived)
+    budget = config.retain_max_completion_tokens
+    numerator, denominator = _SATURATION_BUDGET_SHARE
+    derived = budget * numerator // (denominator * _APPROX_OUTPUT_TOKENS_PER_FACT)
+    # Never advertise a bound the budget cannot reach. Below roughly 2.7k output
+    # tokens the floor would exceed the token-limited count, putting a maxItems
+    # in the schema that the response can never hit — the array would run to
+    # finish_reason=length instead, which is the path this bound exists to avoid.
+    reachable_floor = min(RETAIN_FACTS_SATURATION_FLOOR, budget // _APPROX_OUTPUT_TOKENS_PER_FACT)
+    return max(reachable_floor, derived)
 
 
 # Separators for sentence-aware recursive text splitting, ordered most- to
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -465,6 +465,11 @@ class VerbatimFactExtractionResponse(BaseModel):
     facts: list[VerbatimExtractedFact] = Field(description="List of metadata entries (one per chunk)")
 
 
+# A response at this cap is treated as saturated and sent through the existing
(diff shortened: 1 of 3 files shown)


kamilfurtak:
Rebased onto current `main` (post-#3656). The only conflict was in `tests/test_fact_extraction_retry.py`, where both branches append new tests at the end of the file — resolved by keeping both sets. No changes to the fix itself.

`tests/test_fact_extraction_retry.py`: 32 passed, 1 failed; `tests/test_fact_extraction_entities_schema.py`: 12 passed. The single failure is `test_output_too_long_error_is_a_single_class_across_modules`, which fails identically on a clean checkout of `main` in my environment (an optional-dependency import), so it is unrelated to this branch.

kamilfurtak:
Pushed two corrections and rewrote the description; ready for review.

- Config reads go back to direct attribute access. The previous revision used `getattr(config, ..., default)`, which is exactly what #3610 removed from the engine a day earlier — `StaticConfigProxy` raises `ConfigFieldAccessError` (an `AttributeError` subclass) to signal a bank-configurable field, so a default silently substitutes the global value. Both fields are required and the budget is already validated positive, so the fallback covered nothing.
- The sibling-salvage path is gone. Keeping one half's facts when the other failed contradicted `extract_facts_from_text`, which fails the retain if any chunk could not be extracted because "partial extraction is not acceptable" — salvaging would have reported the operation complete with data silently missing, which is the failure this PR exists to prevent. The test that asserted salvage now asserts propagation.

Net effect is 20 fewer lines than the previous revision. 44 passed.

Also corrected a stale reference: the dependency is **#3673**, not #3656. #3656 (strict-schema flag) is merged; #3673 is the nested-conversation-array unwrap. That ordering matters — this PR makes the unsplittable case fatal, so merging it before #3673 would turn the wrapped-array payloads that #3673 recovers into hard retain failures.

Happy to split the "fail instead of drop" half into its own PR if the behaviour change is the sticking point — the `maxItems` bound stands on its own.

kamilfurtak:
Recalibrated the saturation bound; this was wrong and would have hurt.

The bound was half the budget — 24 facts at a 4096-token ceiling — while the field data in the description already reported a maximum of 29 from a single chunk. It sat *below* observed honest output, so dense input would have been split for nothing.

It needs to land in a window: above real density, and below the output-token ceiling (otherwise the schema bound is unreachable and the array runs to `finish_reason=length`, which is the case this exists to prevent). Three quarters of the token-limited count clears both — 36 at a 4096-token budget, against ~48 the budget can pay for and 29 seen in the field. Regression tests pin both edges.

The 85-tokens-per-fact estimate turned out to be sound: measured against a strict `json_schema` across two local models, four runs gave 80.1, 82.1, 89.5 and 101.5. Only the halving was wrong.

Also capped the floor by what the budget can reach — below ~2.7k output tokens a floor of 32 advertises a `maxItems` the response can never hit.

46 passed.

nicoloboschi:
this is actually not solving the problem and potentially breaking sane and healthy setups, closing

PR #3658, opened 2026-08-19
bank list: take last_write_at and last_document_at from the store

nicoloboschi:
The engine half of taking these two times from the store. The interface defaults return `{}`, so a store that does not report them is unaffected and the Postgres path is untouched.

## The gap

A store-owned bank's row had neither `last_write_at` nor `last_document_at`: both come from `MAX` over SQL tables such a bank leaves empty, and the response model drops null keys, so the fields vanished from the JSON rather than reading null (clients saw a `KeyError`, not a `None`). **The list is ordered by last write**, so that ordering silently degenerated to created-at.

## Two decisions worth reviewing

**Applied before the sort, over every bank — not in the page-level overlay.** The Counters RPC is batched, so all banks cost one call. Doing it per page would fix the field while leaving the order wrong *and inconsistent across pages*, which is worse than uniformly wrong.

**The two times are asked for separately, because they are different questions.** `last_document_at` is INGESTION time and must not move when an existing document is re-retained; `last_write_at` must. Reporting one clock under both names looked harmless and is not — it makes a rewrite read as a new document, which is exactly what `test_last_write_at_advances_when_existing_document_is_rewritten` pins. (The store preserves a document's original `created_at` across re-ingest deliberately, so the ingestion counter is stable by construction.)

A bank the store cannot answer for keeps whatever SQL gave it: **absent means "unknown", never "the epoch"**, so an un-folded bank is not pushed to the bottom of the order on a guess.

## Verification

* Against a store-owned backend: `test_bank_last_write_at` **3/3**, `test_list_banks_non_sql_store` and `test_http_api_integration` green.
* Postgres: unaffected (4/4 on its own bank-list tests).

One caveat, stated rather than hidden: `test_last_write_at_advances_when_existing_document_is_rewritten` fails when run alongside several other modules and passes 3/3 alone. The mechanism that could move that field is a missed read of the prior document record under concurrency, not this code — but I have not proven that, so I am flagging it rather than calling it clean.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memories/base.py
+++ b/hindsight-api-slim/hindsight_api/engine/memories/base.py
@@ -855,6 +855,28 @@ class MemoriesExtension(Extension, ABC):
                 out[bank_id] = counts
         return out
 
+    async def last_write_at_many(self, *, bank_ids: "list[str]") -> "dict[str, datetime]":
+        """When each bank was last written — ``{bank_id: datetime}``, for many banks at once.
+
+        The bank list is ORDERED by this. Postgres derives it with ``MAX`` over live
+        ``documents`` / ``memory_units`` rows, which a store owning those rows leaves empty, so
+        without an answer here such a bank's ordering silently degenerates to created-at.
+
+        Empty by default: the SQL path already has the columns and does not need this, and a store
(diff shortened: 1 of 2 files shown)


nicoloboschi:

PR #3659, opened 2026-08-19
comments: keep the store-extension comments backend-generic

nicoloboschi:
Four comments added in #3560 named a specific downstream store by name.

The seam they describe is the generic store-owned one (`store_owned_retain_for` /
`writes_memory_rows_in_sql_for`), and the reasoning holds for any store that keeps memories outside
SQL — so the comments now say that instead of naming one implementation.

One of them also said "no write-group (X-only)", which read as a property of that store; the reason
is actually that there is no SQL witness to decide, so it now says that.

Comment-only — no behaviour change, no test change. `ruff check` and `ruff format --check` clean.


nicoloboschi:

PR #3660, opened 2026-08-19
Memory backend: store-owned delta retain, bank-list times, and generic comments

nicoloboschi:
One PR for the store-backend work. Everything here came out of running the whole suite against a
store-owned backend and diffing it against the Postgres baseline — a failure counts only if it
**passes** on Postgres.

## The bugs — eleven, all the same mistake

A read or a write that assumes Postgres holds the memories, inside code already made store-aware
everywhere else. **Every one of them failed silently**: no error, no warning, a wrong answer.

The three worth reading first:

1. **A streaming retain kept only its LAST batch.** Each consumer batch re-issued the document
   replace, tombstoning what the previous batch wrote. Measured on a ten-batch document: the retain
   returned **100 unit ids** and the bank held **10**. Data loss on any document large enough to
   split, and the retain reports every id it created, so the caller sees success. The gate was
   `is_first_batch` — a parameter of the whole `retain_batch` call, true for every batch it
   produces. `doc_tracking_done`, the latch that can express "the first batch of THIS document",
   was set and never read.
2. **`export_bank` produced a valid, EMPTY archive** with a success status — `document_count=0,
   fact_count=0`. It takes `memories=None` and treated "no store supplied" as "SQL-backed". This had
   already happened once: it was fixed at the two call sites and the default that causes it was left
   in place.
3. **The importer wrote documents nowhere.** `_store_document_bodies` is the retain path's document
   write; the importer called nothing equivalent, so a restore produced a bank listing no documents.
   With (2): backup and restore were both broken.

And the rest:

4. `on_conflict` was inert — `skip` re-imported the document it was told to leave alone, `new-id`
   kept the original id, `replace` became a plain insert. The existence check was SQL-only.
5. `expand` returned neither chunk nor document: its memories read was routed to the store, its
   chunk and document reads were not.
6. A metadata-only retain (tags or metadata changed, no chunk changed) relabelled the DOCUMENT and
   left every unit on the old labels — the store branch re-put the record and stopped, where the SQL
   branch propagates onto the units.
7. A best-effort txn reap crashed the whole maintenance tick on a loop with no engine, because the
   engine deref sat outside the `try` its own docstring implies.
8. Surviving units kept the PREVIOUS retain's metadata — the store branch patched tags only, where
   the SQL branch sets tags and metadata.
9. User metadata was written where no reader looks: flat keys merged into the record's own bag
   instead of the one JSON value every read surface reconstructs from.
10. Delta ran only on the first sub-batch while the whole-body diff it needs was computed from a
    single slice, so an oversized replacement classified every chunk absent from that slice as
    REMOVED — `unchanged=1 changed=0 new=0 removed=19` on a 20-chunk document.
11. (Store-side, in the paired store PR — NOT in this diff) gRPC channels reused across event
    loops, and `entity_kind` missing from `list_entities`.

## Three new tests, and why they exist

`test_every_streaming_batch_survives_the_next_one`, `test_append_coalesces` and
`test_reflect_expand_store_reads` all assert through the **public read API**. The tests that should
have caught these bugs seed Postgres-internal state a store-owned bank never reads — so they were
already failing for their own reasons, and the real defect underneath stayed invisible.

The streaming test checks the surviving **ids**, not just the count: a replace leaving an equal
number of DIFFERENT memories would satisfy a count check.

## Sixteen tests marked `memory_backend_incompatible`

Each carries the specific statement that cannot apply — a raw `INSERT INTO memory_units`, a
`SELECT ... FROM documents`, a race forced by writing `documents.updated_at`, Postgres's in-txn
`FOR SHARE` liveness skip. Where the property still matters the covering test is named.

**The rule that found bug 1:** before writing a failure off as storage-layout, check whether the
store path is EQUIVALENT, not merely PRESENT. `test_retain.py` had two failures that genuinely were
storage-bound and one that asserted through the public API. Marking the file wholesale — the move
that drops a failure count fastest — would have buried the data loss indefinitely.

One of these also removes ~10 minutes from every run: `test_consumer_failure_cancels_in_flight_extractions`
injects its failure through an exploding Postgres pool that the PG-free store path never acquires
from, so the failure never fired, the deliberately-hanging extraction was never cancelled, and it
timed out at 600s.

## Verification

* Full paired sweep: **5882 passed** against the store-owned backend in 8:56 (Postgres baseline
  5894). The stable failure set — failing in BOTH of the last two sweeps and passing on Postgres in
  both — is **10**, and every one is accounted for: 3 the webhook design gap below, 1 classified
  after that sweep ran, 6 LLM-config tests needing provider credentials.
* Several fixes were proven load-bearing by reverting them and reproducing the exact symptom, e.g.
  `expand returned no document for <id>` and `100 returned / 10 held`.
* Raw failure counts are not quoted on purpose: ~50 failures flip buckets between identical runs, so
  only the stable set means anything.

## One thing deliberately NOT fixed

`retain.completed` webhooks do not fire on the store-owned retain path. That path skips the Postgres
transactional-outbox row **by design** — `_streaming_store_owned_retain` says so in its docstring,
the intended replacement being at-least-once emission from the store. That replacement is not built,
so the delivery is silently dropped: a two-document async retain queues 0 rows where Postgres queues
2. Wiring the outbox back in would reverse a recorded design decision, so it needs a call: implement
store-side emission, or accept the gap and document it in the API docs.

The three tests are left failing rather than marked, because they assert product behaviour and
deselecting them would turn a visible gap into a silent one.

---

## The two riskiest changes, described

Neither was covered above, and both are the ones to read closely.

### The delta full-body diff (and what it deliberately does NOT change)

For an oversized item `contents` is one slice while the stored document is whole, so diffing those
two marks every chunk merely absent from the slice as REMOVED and tombstones its memories. Slice 1
now diffs `document_body_override` — the complete body — instead.

An **append** is excluded: there the override holds only the new tail (the stored body is prepended
onto slice 1), so diffing the whole document against it dropped every pre-existing chunk — measured,
chunks covered 4,348 of 18,538 chars. Appends keep diffing the already-prepended `contents`.

The **gate stays `is_first_batch`**. An earlier revision of this PR widened it to every slice; that
changes the Postgres path, and three things assume the narrow gate: the caller keeps one result list
per sub-batch item (`sub_origins` is length 1 for an oversized slice, so a multi-chunk delta's extra
unit ids are dropped), `chunk_index_offset` advances by the splitter's per-slice count rather than by
what a delta wrote, and a brand-new oversized document would extract its whole tail in one step —
re-creating the unbounded step the sub-batch splitting exists to prevent (#1571, an orchestrator
memory bound).

Consequence, tracked in the store repo rather than papered over: an oversized RE-retain still
re-extracts unchanged chunks on a store-owned bank, as it does on main. The cross-slice skip
Postgres relies on has no equivalent there, and routing it at the store does not work — measured,
`_store_document_bodies` uploads the full chunk-text list before extraction, so the check matches its
own write on slice 1 (`stored=492a32c9 new=492a32c9`).

### The bank-list write times

A store-owned bank's row had neither `last_write_at` nor `last_document_at`: both come from `MAX`
over SQL tables such a bank leaves empty, and the response model drops null keys, so the fields
vanished from the JSON rather than reading null. **The list is ordered by last write**, so that
ordering silently degenerated to created-at.

Two decisions worth the review: the times are asked **separately**, because ingestion time must not
move when a document is rewritten and reporting one under both names makes a rewrite read as a new
document; and they are applied **before** the sort over all banks, because fixing the field in the
page-level overlay would leave the order wrong and inconsistent across pages.

The two calls are gathered and run **outside** the pool connection — they are round trips to another
service and nothing after them needs the connection. `return_exceptions` keeps one failing from
discarding the other's answer, and the whole thing stays inside a `try`, because building the gather
arguments already calls into the store: a store that does not implement these raises before
`return_exceptions` can see anything. They are optional — the interface defaults return `{}`.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/maintenance.py
+++ b/hindsight-api-slim/hindsight_api/engine/maintenance.py
@@ -113,7 +113,9 @@ _RETENTION_BATCH_PAUSE_SECONDS = 0.25
 class MaintenanceLoop:
     """Owns the single periodic maintenance task for a :class:`MemoryEngine`."""
 
-    def __init__(self, engine: "MemoryEngine") -> None:
+    def __init__(self, engine: "MemoryEngine | None") -> None:
+        """``engine`` is optional so the loop can be exercised for its scheduling alone — the
+        cadence tests construct it that way. Every job that needs the engine checks first."""
         self._engine = engine
         self._task: asyncio.Task | None = None
         self._stop = asyncio.Event()
@@ -454,11 +456,12 @@ class MaintenanceLoop:
         store = get_memories()
         if store.writes_memory_rows_in_sql:
             return
-        # Best-effort includes not having an engine to ask: the loop can be constructed without one,
-        # and this reap is explicitly allowed to be skipped (the docstring above). The SQL stores
-        # never get here — they return on the line above — so a store-owned deployment was the only
-        # one that could lose an entire maintenance tick to an AttributeError over a reap that is
-        # permitted to wait for the next one.
+        # Best-effort includes not having an engine to ask: the loop can be constructed without one
+        # (see `__init__`), and this reap is explicitly allowed to be skipped. Only a store-owned
+        # deployment reaches this line — the SQL stores return above — where it would otherwise
+        # raise AttributeError. `_run` catches per tick and this reap is the tick's last job, so the
+        # cost was a logged traceback rather than lost work; it is still noise standing in for a
+        # condition the code should simply state.
         engine = self._engine
         if engine is None:
             return
--- a/hindsight-api-slim/hindsight_api/engine/maintenance.py
+++ b/hindsight-api-slim/hindsight_api/engine/maintenance.py
@@ -454,7 +454,15 @@ class MaintenanceLoop:
         store = get_memories()
         if store.writes_memory_rows_in_sql:
             return
-        backend = self._engine._backend
+        # Best-effort includes not having an engine to ask: the loop can be constructed without one,
+        # and this reap is explicitly allowed to be skipped (the docstring above). The SQL stores
+        # never get here — they return on the line above — so a store-owned deployment was the only
+        # one that could lose an entire maintenance tick to an AttributeError over a reap that is
+        # permitted to wait for the next one.
+        engine = self._engine
+        if engine is None:
+            return
+        backend = engine._backend
         try:
             async with acquire_with_retry(backend, max_retries=1) as conn:
                 bank_ids = [r[0] for r in await conn.fetch(f"SELECT bank_id FROM {fq_table('banks')}")]
--- a/hindsight-api-slim/hindsight_api/engine/memories/base.py
+++ b/hindsight-api-slim/hindsight_api/engine/memories/base.py
@@ -855,6 +855,28 @@ class MemoriesExtension(Extension, ABC):
                 out[bank_id] = counts
         return out
 
+    async def last_write_at_many(self, *, bank_ids: "list[str]") -> "dict[str, datetime]":
+        """When each bank was last written — ``{bank_id: datetime}``, for many banks at once.
+
+        The bank list is ORDERED by this. Postgres derives it with ``MAX`` over live
+        ``documents`` / ``memory_units`` rows, which a store owning those rows leaves empty, so
+        without an answer here such a bank's ordering silently degenerates to created-at.
+
+        Empty by default: the SQL path already has the columns and does not need this, and a store
+        that cannot answer should say nothing rather than guess. **A bank absent from the result
+        keeps whatever the caller already had — absent means "unknown", never "the epoch".**
+        """
+        return {}
+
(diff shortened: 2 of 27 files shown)


r266-tech (commented):
Reviewing PR #3660 at head `3b4c797efd7c6dad7b50076bcc13432bcef9e771`, the oversized-replacement fallback returns `_delta_metadata_only` whenever there are no unchanged chunks. A strict extension can reach this path when the existing chunk changes at its boundary while the replacement adds a large tail. `_delta_metadata_only` updates the document and existing-unit metadata only; it does not create or extract the tail chunks, so a successful retain can expose the complete new `original_text` while omitting memories for the appended content.

Please route replacements with uncovered added content through extraction for the added range, and add a regression with a one-chunk stored document whose boundary changes plus a tail large enough to trigger sub-batching; assert that tail chunks and facts are present through the engine API.

The authority-bound head is `3b4c797efd7c6dad7b50076bcc13432bcef9e771`; head binding passed and `git diff --check` exited 0. No tests were executed during the read-only inspection.

PR #3662, opened 2026-08-20
fix(engine): drop per-bank vector indexes before the delete transaction (#3485)

xiaoxiaoHe-E:
## Problem

At the shipped default (`HINDSIGHT_API_VECTOR_INDEX_MIN_ROWS=0`), every bank is still given its three per-(bank, fact_type) partial vector indexes at creation — the behaviour #3638 deliberately restored ("exactly as before #3561"). So the cluster-wide index count still scales with bank count, and the #3485 lock-table exhaustion failure mode remains in the default configuration; only a deployment that sets the threshold (e.g. `10000`) escapes it.

In that failure mode `delete_bank` is the only recovery path — but today it drops the bank's indexes **after** the delete transaction commits. At lock-table exhaustion, planning the delete DML itself fails, so the transaction never commits and the drop never runs. The one tool that still works there — `DROP INDEX` is a utility statement that locks only its own index plus the table — is placed where it can never be reached.

## Change

Drop the bank's per-bank vector indexes **before** the delete transaction starts, on the autocommit connection (where `DROP INDEX CONCURRENTLY` already has to run). The bank row and its rows still exist at that point, so the `internal_id` the index names derive from is read with a plain `SELECT` instead of `DELETE ... RETURNING`. The delete transaction then runs against a `memory_units` with three fewer indexes, freeing lock-table space as it goes.

- Only the full-delete path drops indexes; `fact_type`-scoped deletes and `delete_bank_profile=False` are unchanged.
- Same `DROP INDEX CONCURRENTLY` + `retry_with_backoff` as before; Oracle remains a no-op.
- New regression test asserts the drop happens while the bank's rows and profile still exist.

## On "CONCURRENTLY cannot run inside a transaction block"

That constraint is exactly why the drop has to run *outside* the delete transaction — but "outside" has two sides: before it or after it. The current code chose after; this PR chooses before, on the same autocommit connection `DROP INDEX CONCURRENTLY` already requires. The constraint does not argue for after-transaction in particular, and the failure mode does: when the lock table is exhausted the delete transaction never commits, so an after-transaction drop never runs, while a before-transaction drop always can.

Nor do retries cover this: `retry_with_backoff` absorbs *transient deadlocks* (40P01) between concurrent DDL, but lock-table exhaustion fails every DML at plan time — the same failure repeats on every retry, while `DROP INDEX` never plans against the relation's other indexes.

## Failure semantics

- **Drop fails (after retries):** `delete_bank` fails and the bank's data is untouched — the whole operation can simply be retried.
- **Drop succeeds, delete transaction later fails:** the bank keeps its data without its indexes and recall serves through the exact B-tree path — slower, never wrong.
- Both are safer than the status quo, where the data is already gone when the post-commit drop fails and the leftover indexes become orphans that only a catalog sweep can find.

## Why not "just set the threshold"

The threshold is opt-in and defaults to 0, so the default deployment's index count still grows with bank count. Until `HINDSIGHT_API_VECTOR_INDEX_MIN_ROWS` ships with a non-zero default (e.g. `10000`), this drop-first ordering is the only thing keeping `delete_bank` usable as a recovery path in the default configuration — and even at a raised threshold, banks above it still have indexes and need the same ordering.

#3561's own orphan sweep (`repair-bank --all`) exists precisely because a deployment at the #3485 wall could not run `delete_bank` at all — its test simulates a bank whose rows were deleted manually with the indexes left behind. That is exactly the state an after-transaction drop produces by construction at lock-table exhaustion; this PR makes `delete_bank` the one that drops them instead.

This fixes the ordering half of #3485; #3561/#3638 address the count half. Both are needed while the default stays 0.

## Validation

- `pytest tests/test_hnsw_indexes.py` — 11 passed (incl. new `test_delete_bank_drops_indexes_before_data`)
- `pytest tests/test_bank_lifecycle_deadlock_retry.py` — 6 passed
- `pytest tests/test_bank_vector_index_ddl_lock.py` — 4 passed
- `ruff check` / `ruff format --check` clean

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -8586,10 +8586,36 @@ class MemoryEngine(MemoryEngineInterface):
         backend = await self._get_backend()
         invalidated_obs = 0
         result: dict[str, int] = {}
-        bank_internal_id: str | None = None
         async with acquire_with_retry(backend) as conn:
             # Ensure connection is not in read-only mode (can happen with connection poolers)
             await conn.execute("SET SESSION CHARACTERISTICS AS TRANSACTION READ WRITE")
+
+            # Drop this bank's per-bank vector indexes BEFORE the delete
+            # transaction (issue #3485): when the cluster-wide index count
+            # approaches the lock-table budget, planning any statement against
+            # memory_units fails — including the delete DML — while DROP INDEX
+            # (a utility statement) only locks its own index and the table, so
+            # it still works. Dropping first keeps bank deletion usable as a
+            # recovery path and frees lock-table space for the delete that
+            # follows. The drop runs CONCURRENTLY (see
+            # ops.drop_bank_vector_indexes), which cannot run inside a
+            # transaction block — this autocommit connection is exactly where
+            # it belongs. Only the full-delete path drops indexes:
+            # fact_type-scoped deletes and delete_bank_profile=False keep them.
+            if not fact_type and delete_bank_profile:
+                internal_id = await conn.fetchval(
+                    f"SELECT internal_id FROM {fq_table('banks')} WHERE bank_id = $1", bank_id
+                )
+                if internal_id:
+                    # Retry the transient deadlock a concurrent index
+                    # build/drop on the shared memory_units table can still
+                    # trigger (sqlstate 40P01 / ORA-00060).
+                    await retry_with_backoff(
+                        lambda: bank_utils.drop_bank_vector_indexes(conn, str(internal_id), ops=self._backend.ops),
+                        max_retries=7,
+                        max_delay=10.0,
+                    )
+
             async with conn.transaction():
                 try:
                     if fact_type:
@@ -8740,35 +8766,13 @@ class MemoryEngine(MemoryEngineInterface):
                         }
 
                         if delete_bank_profile:
-                            # Delete the bank profile and retrieve internal_id for HNSW index cleanup
-                            internal_id = await conn.fetchval(
-                                f"DELETE FROM {fq_table('banks')} WHERE bank_id = $1 RETURNING internal_id", bank_id
-                            )
-                            if internal_id:
-                                bank_internal_id = str(internal_id)
+                            # Delete the bank profile (its vector indexes were
+                            # already dropped above, before the delete tx).
+                            await conn.execute(f"DELETE FROM {fq_table('banks')} WHERE bank_id = $1", bank_id)
                             result["bank_deleted"] = True
 
                 except Exception as e:
                     raise Exception(f"Failed to delete agent data: {str(e)}")
-
-            # Drop per-bank vector indexes AFTER the transaction commits: the
-            # drop runs CONCURRENTLY (see ops.drop_bank_vector_indexes), which
-            # cannot run inside a transaction block. Same-process drops are
-            # serialized by the ops-level DDL lock; retry_with_backoff absorbs
-            # the residual cross-process deadlock a concurrent index build/drop
-            # on the shared memory_units table can still trigger (sqlstate
-            # 40P01 / ORA-00060) so a delete is never lost to a transient lock
-            # cycle. Sized well above the defaults: a many-process delete storm
-            # (CI teardown ran 8 workers' drops at once) drains at roughly one
-            # deadlock victim per deadlock_timeout (1s), so the default ~2.4s
-            # of backoff lost every retry; ~30s of jittered backoff outlasts
-            # any realistic pile-up.
-            if bank_internal_id:
-                await retry_with_backoff(
-                    lambda: bank_utils.drop_bank_vector_indexes(conn, bank_internal_id, ops=self._backend.ops),
-                    max_retries=7,
-                    max_delay=10.0,
-                )
 
         # A store that keeps memories outside SQL leaves memory_units empty, so every DELETE
         # above was a no-op on its data — it must be told to drop the bank's memories too, or
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -8586,32 +8586,30 @@ class MemoryEngine(MemoryEngineInterface):
         backend = await self._get_backend()
         invalidated_obs = 0
         result: dict[str, int] = {}
+        bank_internal_id: str | None = None
         async with acquire_with_retry(backend) as conn:
             # Ensure connection is not in read-only mode (can happen with connection poolers)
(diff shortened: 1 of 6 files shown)


r266-tech (commented):
High-severity correctness issue: moving the drop before the delete leaves an unprotected window for vector-index maintenance to recreate an index. A worker that has observed the bank can reach CREATE INDEX CONCURRENTLY after the pre-drop but before the delete commits; that creation can finish after the bank and its data are deleted, and because this path no longer has a post-commit drop, the recreated partial index remains orphaned.

Please serialize vector-index creation with a full bank deletion, or add an equivalent deletion marker/exclusion that covers the maintenance existence check through its DDL. Add a regression test that forces maintenance creation to overlap the pre-drop and delete transaction and asserts no bank index remains.

Also update the PostgreSQL ops comment at ops_postgresql.py:1106-1111, which still says delete_bank invokes drop_bank_vector_indexes after its delete transaction, although the caller now invokes it before the transaction.

The bound head is d7701378cadeec6890194489ad213d990f86b3d5; verification reports that the checkout matches this head and that git diff --check exited 0.

koriyoshi2041 (approved):
Re-checked exact head e92b22ae. The update closes both sides of the race: delete re-drops the deleted internal_id after commit, while index maintenance re-checks bank identity after each concurrent build and removes an index built across deletion/recreation.

Local exact-head validation passed:
- `pytest -n 0 tests/test_hnsw_indexes.py::test_delete_bank_drops_indexes_rebuilt_in_window tests/test_hnsw_indexes.py::test_apply_index_plan_drops_orphan_built_during_delete tests/test_bank_lifecycle_deadlock_retry.py -q` — 4 passed
- `git diff --check d7701378..e92b22ae` — clean

The new two-sided regressions cover the pre-drop/post-commit window and a build that completes after deletion. No remaining correctness issue found in the update.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -8590,6 +8590,30 @@ async def delete_bank(
...
+                if internal_id:
+                    bank_internal_id = str(internal_id)
+                    # Retry the transient deadlock a concurrent index
+                    # build/drop on the shared memory_units table can still
+                    # trigger (sqlstate 40P01 / ORA-00060).
+                    await retry_with_backoff(
+                        lambda: bank_utils.drop_bank_vector_indexes(conn, bank_internal_id, ops=self._backend.ops),
+                        max_retries=7,
+                        max_delay=10.0,
+                    )
```
Sanderhoff-alt, line 8615: This drop is unguarded and now runs *before* the delete transaction, so any failure that outlasts the 7 retries aborts `delete_bank` with nothing deleted. The comment removed further down documented exactly that condition: a multi-process drop storm ("CI teardown ran 8 workers' drops at once") drains at roughly one deadlock victim per `deadlock_timeout`, which is why the backoff had to be sized up in the first place. Under that storm the previous ordering lost only the index cleanup (recoverable by `repair-bank --all`); with drop-first, every bank delete in the storm fails and no data is removed at all, which is the opposite of making `delete_bank` a usable recovery path.

Making the pre-drop best-effort keeps the #3485 benefit (the indexes are gone for the common case) without letting cleanup failure block deletion:

```python
try:
    await retry_with_backoff(...)
except Exception as exc:
    logger.warning("Pre-delete vector index drop for bank %s failed: %s; continuing - the post-commit drop and repair-bank will retry.", bank_id, exc)
```

hindsight-api-slim/hindsight_api/engine/vector_index_health.py
```diff
@@ -402,6 +405,32 @@ async def _rebuild(qindex: str = qindex, qualified: str = qualified, fact_type:
...
             except Exception as cleanup_exc:  # noqa: BLE001
                 logger.warning("Cleanup DROP INDEX for %s also failed: %s", qualified, cleanup_exc)
+        else:
+            # Re-check the bank after the build: if it was deleted (or
+            # recreated under a fresh internal_id) while CREATE INDEX
+            # CONCURRENTLY was running, the index just built is an orphan that
+            # delete_bank's post-commit drop could not name. Drop it, don't
+            # count it, and stop — the remaining fact_types would only build
+            # more orphans.
+            still = await conn.fetchval(
```
Sanderhoff-alt, line 415: This re-check sits in the `else:` clause, outside the `try`, so if the `fetchval` fails (connection blip, statement timeout) the exception escapes `apply_bank_index_plan` and aborts the whole run, dropping the remaining `plan.to_build` entries and discarding the result. That contradicts the invariant the sibling handler states one line above ("one failed index must not abort the rest"). Wrapping the re-check and its drop in their own `try/except` that logs and continues would keep the behaviour consistent.

Sanderhoff-alt (commented):
One blocking issue: the new pre-transaction drop is unguarded, so an index-drop failure now aborts the delete before any data is removed. Separately, the Validation section reports 11 tests in `test_hnsw_indexes.py`, but the file has 18 after this change, so the three additional tests may not have been run.

nicoloboschi:
Thanks for coming back to this with an actual argument rather than just reopening it — and you're right on both of the reasons I closed #3493 with.

"`CONCURRENTLY` cannot run inside a transaction block" is true but doesn't justify picking *after* over *before*; that comment of mine didn't actually make the case. And #3561 no longer settles it either: #3638 made `min_rows=0` mean "off" and that is the shipped default, so every bank still gets its three partial indexes at creation and #3485's failure mode is reachable in the default configuration. Both fair points.

What doesn't hold is that this restores the recovery path. At the wall the delete plans against every index on `memory_units` — at the 10k-bank scale from #3485 that's ~30,021. Dropping this bank's three leaves ~30,018, still far past the lock budget, so the delete DML fails exactly as before. The difference is where it leaves you: today the delete aborts cleanly and the bank is untouched; with this change it aborts after the bank has lost its vector indexes while keeping all of its data. And that state doesn't heal — at `min_rows=0`, `_handle_vector_index_maintenance` returns early and `create_bank_vector_indexes` only runs at bank creation and import, so nothing rebuilds them short of `repair-bank`. A permanently un-indexed bank is the difference between the 154 ms and multi-second recall in your own table.

The genuine gap you've surfaced is the one under all of this: `HINDSIGHT_API_VECTOR_INDEX_MIN_ROWS` still defaults to 0, so the count fix from #3561 is opt-in. That's the thing worth fixing, and it's on us — I'll track it on #3485 rather than here.

One piece of this is worth keeping on its own: `test_delete_bank_fact_type_scoped_keeps_indexes` covers a real gap (that a `fact_type`-scoped delete must not drop the bank's indexes) and is independent of the ordering change. If you want to send that as a standalone PR I'll take it.

Closing this one — not as superseded this time, but because the ordering doesn't get the delete under the lock budget. The analysis in #3485 remains the best account of this failure we have, and it's appreciated.

PR #3663, opened 2026-08-20
fix(kb): execute knowledge node updates as a single atomic transaction

Sanderhoff-alt:
## Summary

A `PATCH` request to `/knowledge-base/nodes/{node_id}` or the `update_knowledge_node` MCP tool previously applied rename, move, and page options sequentially across separate database connections and transactions. A failure on a subsequent step (such as moving to a non-existent folder or creating a hierarchy cycle) left earlier changes (such as rename) persisted, causing partial writes and state drift on client retries. Additionally, folder nodes could accept page-only options and fail midway after renaming.

## Changes

- **Single Atomic Transaction**: Added `MemoryEngine.update_knowledge_node()` to validate node kind compatibility, check parent hierarchy loops, and execute all updates (`knowledge_pages` and backing `mental_models`) in a single database transaction.
- **Refactoring**: Refactored `rename_knowledge_node()`, `move_knowledge_node()`, and `update_knowledge_page()` to delegate to `update_knowledge_node()`.
- **API & MCP Tooling**: Updated `api_update_knowledge_node()` and MCP tool handler `_do_update_knowledge_node()` to invoke `update_knowledge_node()`.
- **Error Handling**: Handled asynchronous mental model refresh submission errors gracefully so that submission failures do not mask a successful database update with a 500 response.
- **Testing**: Added regression tests in `test_knowledge_base.py` for atomic rollback on invalid parent, folder field validation, and combined multi-field updates. Verified all test suites pass.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -6064,51 +6064,44 @@ def _register_routes(app: FastAPI):
     ):
         """Rename/move a node and/or update a page's options."""
         try:
-            updated: dict[str, Any] | None = None
-            did_change = False
-            if body.name is not None:
-                did_change = True
-                updated = await app.state.memory.rename_knowledge_node(
-                    bank_id=bank_id, node_id=node_id, name=body.name, request_context=request_context
-                )
-            # parent_id is applied only when present in the body, so passing null
-            # moves the node to the root (distinct from "not provided").
-            if "parent_id" in body.model_fields_set:
-                did_change = True
-                updated = await app.state.memory.move_knowledge_node(
-                    bank_id=bank_id, node_id=node_id, new_parent_id=body.parent_id, request_context=request_context
-                )
-            # Page options live on the backing mental model; each applies only when
-            # present in the body (so tags=[] clears, distinct from "not provided").
-            page_fields = {"source_query", "tags", "max_tokens", "trigger"} & body.model_fields_set
-            if page_fields:
-                did_change = True
-                updated = await app.state.memory.update_knowledge_page(
-                    bank_id=bank_id,
-                    page_id=node_id,
-                    source_query=body.source_query if "source_query" in page_fields else None,
-                    tags=body.tags if "tags" in page_fields else None,
-                    max_tokens=body.max_tokens if "max_tokens" in page_fields else None,
-                    # Only the trigger fields the client stated: the engine patches them over
-                    # the page's current trigger, and a full dump would carry this model's own
-                    # defaults (mode="full", exclude_mental_models=False) into every update.
-                    trigger=(body.trigger.model_dump(exclude_unset=True) if body.trigger else None),
-                    request_context=request_context,
-                )
-                # A new source query means the content is stale — rebuild it.
-                if updated is not None and "source_query" in page_fields and updated.get("mental_model_id"):
-                    await app.state.memory.submit_async_refresh_mental_model(
-                        bank_id=bank_id,
-                        mental_model_id=updated["mental_model_id"],
-                        request_context=request_context,
-                    )
-            if not did_change:
+            has_name = "name" in body.model_fields_set
+            has_parent = "parent_id" in body.model_fields_set
+            has_source_query = "source_query" in body.model_fields_set
+            has_tags = "tags" in body.model_fields_set
+            has_max_tokens = "max_tokens" in body.model_fields_set
+            has_trigger = "trigger" in body.model_fields_set
+
+            if not (has_name or has_parent or has_source_query or has_tags or has_max_tokens or has_trigger):
                 raise HTTPException(
                     status_code=400,
                     detail="Provide name, parent_id, source_query, tags, max_tokens, and/or trigger to update",
                 )
+
+            updated = await app.state.memory.update_knowledge_node(
+                bank_id=bank_id,
+                node_id=node_id,
+                name=body.name if has_name else MemoryEngine.KP_UNSET,
+                parent_id=body.parent_id if has_parent else MemoryEngine.KP_UNSET,
+                source_query=body.source_query if has_source_query else MemoryEngine.KP_UNSET,
+                tags=body.tags if has_tags else MemoryEngine.KP_UNSET,
+                max_tokens=body.max_tokens if has_max_tokens else MemoryEngine.KP_UNSET,
+                trigger=(
+                    (body.trigger.model_dump(exclude_unset=True) if body.trigger is not None else None)
+                    if has_trigger
+                    else MemoryEngine.KP_UNSET
+                ),
+                request_context=request_context,
+            )
             if updated is None:
                 raise HTTPException(status_code=404, detail=f"Knowledge node '{node_id}' not found")
+
+            # A new source query means the content is stale — schedule refresh.
+            if has_source_query and body.source_query is not None and updated.get("mental_model_id"):
+                await app.state.memory.submit_async_refresh_mental_model(
+                    bank_id=bank_id,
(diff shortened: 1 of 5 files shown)


oldnicke:
This PR addresses an important consistency issue in the knowledge-base update workflow and provides a comprehensive fix across the engine, HTTP API, MCP tooling, and regression tests.

The main contributions are:

1. Atomic knowledge-node updates

Previously, rename, move, and page-option updates could be executed through separate database operations. If a later operation failed, earlier changes could remain persisted, resulting in partial updates and inconsistent state.

This PR introduces `MemoryEngine.update_knowledge_node()` and applies all requested changes within a single database transaction. As a result:

- A failed parent validation or hierarchy-cycle check rolls back the entire update.
- A node rename and its backing mental-model rename remain consistent.
- Combined updates such as rename + move + page-option changes either succeed completely or fail without leaving partial state.
- Client retries no longer encounter state drift caused by partially applied requests.

2. Unified update path across the codebase

The HTTP PATCH endpoint and the MCP `update_knowledge_node` tool now use the same atomic engine method. The existing `rename_knowledge_node()`, `move_knowledge_node()`, and `update_knowledge_page()` methods delegate to this implementation, which keeps the behavior consistent while preserving the existing public method interfaces.

This significantly reduces duplicated update logic and ensures that API and MCP callers receive the same validation and transaction guarantees.

3. Stronger validation and safer error handling

The new implementation validates the complete request before writing any data:

- Folder nodes cannot receive page-only options such as `tags`, `source_query`, `max_tokens`, or `trigger`.
- Invalid null values and parameter types are rejected explicitly.
- Self-parenting is rejected.
- Moving a node into its own subtree is rejected.
- The target parent must be a valid folder.
- Field-presence semantics are preserved, so omitted fields are not accidentally overwritten and explicit values such as `tags=[]` continue to work correctly.
- Trigger updates continue to merge supplied fields with the existing trigger configuration.

These checks prevent invalid combinations from causing partial writes or confusing downstream failures.

4. Consistency of backing mental-model data

For page nodes, the implementation updates both the `knowledge_pages` record and its backing `mental_models` record in the same transaction. Renaming a page also updates the mental-model name and its search vector, keeping the displayed knowledge node, generated content, and search index aligned.

5. Resilient asynchronous refresh handling

When `source_query` changes, the PR still schedules an asynchronous mental-model refresh. However, a failure while submitting that background refresh no longer masks a successful database update with an HTTP 500 response.

The persisted update is returned successfully, while the refresh-submission failure is logged for follow-up. This is a useful separation between the durable update and best-effort background processing.

6. Broad regression coverage

The added tests cover the key failure and success paths, including:

- Atomic rollback when a rename is combined with an invalid parent.
- Rejection of page-only fields on folders.
- Rejection of explicit null values for fields that cannot be cleared with `null`.
- Successful combined rename, move, and page-option updates.
- Verification that both the knowledge page and backing mental model are updated.
- Successful API behavior when asynchronous refresh submission fails.
- MCP dispatch behavior and refresh semantics.

Overall, this PR improves transactional correctness, prevents partial knowledge-base writes, unifies update behavior across interfaces, and adds strong regression protection for the affected workflows.

oldnicke (commented):
This PR addresses an important consistency issue in the knowledge-base update workflow and provides a comprehensive fix across the engine, HTTP API, MCP tooling, and regression tests.

The main contributions are:

1. Atomic knowledge-node updates

Previously, rename, move, and page-option updates could be executed through separate database operations. If a later operation failed, earlier changes could remain persisted, resulting in partial updates and inconsistent state.

This PR introduces `MemoryEngine.update_knowledge_node()` and applies all requested changes within a single database transaction. As a result:

- A failed parent validation or hierarchy-cycle check rolls back the entire update.
- A node rename and its backing mental-model rename remain consistent.
- Combined updates such as rename + move + page-option changes either succeed completely or fail without leaving partial state.
- Client retries no longer encounter state drift caused by partially applied requests.

2. Unified update path across the codebase

The HTTP PATCH endpoint and the MCP `update_knowledge_node` tool now use the same atomic engine method. The existing `rename_knowledge_node()`, `move_knowledge_node()`, and `update_knowledge_page()` methods delegate to this implementation, which keeps the behavior consistent while preserving the existing public method interfaces.

This significantly reduces duplicated update logic and ensures that API and MCP callers receive the same validation and transaction guarantees.

3. Stronger validation and safer error handling

The new implementation validates the complete request before writing any data:

- Folder nodes cannot receive page-only options such as `tags`, `source_query`, `max_tokens`, or `trigger`.
- Invalid null values and parameter types are rejected explicitly.
- Self-parenting is rejected.
- Moving a node into its own subtree is rejected.
- The target parent must be a valid folder.
- Field-presence semantics are preserved, so omitted fields are not accidentally overwritten and explicit values such as `tags=[]` continue to work correctly.
- Trigger updates continue to merge supplied fields with the existing trigger configuration.

These checks prevent invalid combinations from causing partial writes or confusing downstream failures.

4. Consistency of backing mental-model data

For page nodes, the implementation updates both the `knowledge_pages` record and its backing `mental_models` record in the same transaction. Renaming a page also updates the mental-model name and its search vector, keeping the displayed knowledge node, generated content, and search index aligned.

5. Resilient asynchronous refresh handling

When `source_query` changes, the PR still schedules an asynchronous mental-model refresh. However, a failure while submitting that background refresh no longer masks a successful database update with an HTTP 500 response.

The persisted update is returned successfully, while the refresh-submission failure is logged for follow-up. This is a useful separation between the durable update and best-effort background processing.

6. Broad regression coverage

The added tests cover the key failure and success paths, including:

- Atomic rollback when a rename is combined with an invalid parent.
- Rejection of page-only fields on folders.
- Rejection of explicit null values for fields that cannot be cleared with `null`.
- Successful combined rename, move, and page-option updates.
- Verification that both the knowledge page and backing mental model are updated.
- Successful API behavior when asynchronous refresh submission fails.
- MCP dispatch behavior and refresh semantics.

Overall, this PR improves transactional correctness, prevents partial knowledge-base writes, unifies update behavior across interfaces, and adds strong regression protection for the affected workflows.

nicoloboschi:
Thanks for this — the diagnosis is exactly right, and the transaction boundary you drew is the correct one. I'm closing this in favour of #3914, which fixes the same bug with the same approach but keeps the request contract unchanged.

The blocker was `null` handling. Every field on `UpdateNodeRequest` is declared `X | None = None`, so a null is valid input that has always meant "not changing this field". With the new per-field checks, several previously-working requests now fail:

| PATCH body (on a page) | before | this PR |
|---|---|---|
| `{"name": "X", "tags": null}` | 200, renamed | 400 `Page tags cannot be null` |
| `{"name": "X", "trigger": null, "max_tokens": null}` | 200, renamed | 400 `Page max_tokens cannot be null` |
| `{"source_query": ""}` | 200 | 400 |
| `{"name": "   "}` | 200 | 400 |

Any client that serializes its whole model with nulls — a common SDK and UI shape — gets a 400 on a rename that used to work. `check-openapi-compatibility` stayed green because the schema still declares those fields nullable, so the break is invisible to CI.

Two smaller things also fed into the decision: the mental-model column writes were restated rather than reused, leaving two copies of the same SQL (including the `search_vector` re-tokenization and the trigger merge) free to drift; and the bank tree lock was taken for every patch, including renames and tag edits that change no parent link.

Worth flagging for anyone reading this later: because this is a fork PR, `test-api` was skipped — only 12 of 94 checks actually ran, and none of them execute the tests added here. I ran them locally and they do pass; the compatibility break is not something they cover.

The replacement keeps your transaction boundary and your validate-before-write ordering, reuses `update_mental_model` on the caller's connection instead of duplicating it, and carries regression tests for each rollback path. Thanks again for finding and pinning down the bug.

PR #3666, opened 2026-08-20
fix(retain): replace the previous version when an oversized re-retain rewrites a chunk

nicoloboschi:
## The problem

A replacement body over `retain_batch_tokens` is sliced into sub-batches before any retain logic
runs, and each slice is its own `retain_batch` call. Only slice 1 carries `is_first_batch` — the
flag that makes document tracking cascade-delete the outgoing version — and slice 1 normally
resolves through delta retain, which deletes nothing outside its own slice.

So for an oversized re-retain, **nothing ever removed the previous version's facts for the chunks
the later slices rewrote.** Measured on a 10-section document (20 chunks, 10 facts each):

| | before | after |
|---|---|---|
| Edit one section, re-retain 4x | 200 → 210 → 220 → 230 units; the edited chunk holds 10 → 20 → 30 → 40 facts | stays 200, edited chunk stays at 10 |
| Re-retain 10 sections as 6 | 200 units, all 20 chunk indices populated, sections 7-10 still recallable | the dropped tail is gone |
| Edit the *first* chunk | 8,322 tokens through the LLM for a one-chunk edit of an 8,320-token document | only the edited chunk |

The first two are correctness: recall returns facts from a version of the document that no longer
exists, and it compounds with every re-ingest. The third is cost — exactly the cost the split path
exists to avoid.

The same document inside `retain_batch_tokens` (the plain delta path) was correct throughout, which
is what made this invisible: it only shows up once transport splits the body.

## The fix

`is_first_batch` keeps meaning what it did for the document row and its content hash. What changes
is that replacement becomes **positional** where the whole-document cascade cannot reach:

* **`_supersede_rewritten_chunks`** (`retain/orchestrator.py`) — a `chunk_id` is
  `{bank}_{document}_{index}`, so writing an index replaces its occupant. Before storing its chunk
  rows, a batch now takes the outgoing facts at those indices — with the observations, entity-prune
  and relink bookkeeping `delete_chunks_by_ids` already does for delta. Skipped when that batch's
  own tracking just cascade-deleted the version. Called from both SQL write paths (Postgres and
  Protocol-B); a store-owned bank has no chunk rows and replaces the document wholesale on its
  first batch, so it needs nothing — noted in `_streaming_store_owned_retain`'s docstring.
* **`MemoryEngine._truncate_replaced_document_tails`** + `chunk_storage.delete_chunks_from_index` —
  drops whatever sits past the new last chunk, once every sub-batch has landed and the document's
  real chunk count is known. Takes the document row lock and only acts while the stored hash is
  still the body this retain wrote, so a concurrent takeover's (possibly longer) version is never
  truncated; skipped when the run was cancelled part-way.
* **Delta retain stops drawing document-wide conclusions from one slice** — stored chunks outside
  the slice are no longer counted as "removed" (they are the other slices'), and a slice whose own
  chunks have all changed now runs delta on itself instead of falling back to the streaming path,
  whose `is_first_batch` cascade wiped the version the other slices were about to skip.

## Tests

`tests/test_delta_oversized_stale_facts.py` — five tests, three of which fail on `main`:

* no stale-fact accumulation across repeated edits of an oversized document
* the removed tail of a shrinking oversized replacement is deleted
* a first-chunk edit no longer re-extracts the rest of the document
* control: the same edit within the transport budget replaces rather than accumulates
* an edit landing in a *late* sub-batch still reaches extraction (the skip is keyed on the chunk
  hash, so a changed chunk is never skipped — pinning that the fix does not overshoot)

`tests/test_retain_ext_writegroup.py` calls `_streaming_batch_write_ext` directly and gains the two
new arguments.

## Verification

* Full `hindsight-api-slim` suite: 6,167 passed. The 29 failures / 21 errors are pre-existing and
  environmental — LLM-provider, tracing, S3 and real-LLM tests, migration tests that need their own
  pg0 instance, plus a handful of parallel-unsafe tests; every retain/delta one of them passes when
  re-run on this branch.
* Targeted run of every retain / delta / chunk / document / append / observation test file: 506
  passed.
* `./scripts/hooks/lint.sh` and `uv run ty check hindsight_api/` clean; `check-unused.sh` reports
  nothing new.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -14,6 +14,7 @@ import contextvars
 import copy
 import difflib
 import functools
+import hashlib
 import inspect
 import json
 import logging
@@ -5226,6 +5227,22 @@ class MemoryEngine(MemoryEngineInterface):
 
             chunk_offsets: dict[str, int] = {}
 
+            # The content hash of the COMPLETE body each split document is being replaced with —
+            # the same value the retain writes to documents.content_hash. Kept so the tail
+            # truncation after the loop can prove this retain still owns the document before it
+            # deletes anything (a concurrent takeover writes a different hash, and its version
+            # may legitimately have more chunks than ours).
+            split_document_hashes: dict[str, str] = {}
+            for _sub_batch, _override in zip(sub_batches, document_body_overrides):
+                _override_doc_id = document_id or _sub_batch[0].get("document_id")
+                # Append grows a document, so it has no tail to drop — and its stored body is the
+                # prepend, not this override, so the hash below would not be the one written.
+                if _override is None or not _override_doc_id or _sub_batch[0].get("update_mode") == "append":
+                    continue
+                split_document_hashes[_override_doc_id] = hashlib.sha256(
+                    (fact_extraction._sanitize_text(_override) or "").encode()
+                ).hexdigest()
+
             # In update_mode="append", retain_batch prepends the existing document
             # body to the FIRST sub-batch as an extra content item before chunking
             # (see orchestrator.retain_batch), consuming chunks(existing_body)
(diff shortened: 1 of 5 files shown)


r266-tech (commented):
Please address the cross-store atomicity issue in tail truncation before merge. `_truncate_replaced_document_tails` opens a SQL transaction, but its call to `delete_chunks_from_index` does not pass that transaction through to `delete_chunks_by_ids`. For Protocol-B banks, external-store tombstones are therefore issued with `txn=None` before the SQL chunk deletion commits. A later SQL failure or rollback leaves deleted external memories behind while their SQL metadata remains, causing cross-store divergence and recall failures.

Thread the active transaction through the tail-deletion path for Protocol-B tombstones, or defer external deletion until the SQL transaction has committed with an equivalent recovery strategy. Add a failure-boundary regression that forces the SQL cleanup transaction to roll back and verifies both stores remain aligned.

This review targets PR #3666 at head `b0f55dc4ec76c93df3e69c9f154b013935013c9f`, matching the supplied source observation, repository snapshot, and head-binding verification. The recorded head-binding check passed, and `git diff --check` exited 0.

PR #3667, opened 2026-08-20
fix(webhooks): omit Content-Type and signature on GET deliveries

Sanderhoff-alt:
## Context & Relationship to #3368

In #3368, we identified that GET webhook deliveries compute `X-Hindsight-Signature` and set `Content-Type: application/json` over the serialized event payload, but dispatched the request without a body via `.get()`, leaving receivers with an empty body and an unverifiable signature.

#3368 was closed to avoid the breaking changes and network interoperability issues associated with adding request bodies to GET requests (e.g., intermediary proxies/gateways stripping bodies or servers rejecting GET requests with `Content-Length`, per RFC 9110 §9.3.1).

As suggested in the [review on #3368](https://github.com/vectorize-io/hindsight/pull/3368#issuecomment-5353617125) (Option 1), this PR makes GET deliveries honest:
- Since GET requests carry no body, we do not send `Content-Type: application/json` or `X-Hindsight-Signature`.
- GET requests continue to send `X-Hindsight-Event` and any caller-configured `http_config.headers`.
- POST delivery semantics and signature generation remain completely untouched.

## Changes
- `hindsight_api/engine/memory_engine.py`: Conditionally assemble headers based on whether `http_config.method` is `GET` or `POST`.
- `tests/test_webhooks.py`: Add `TestGetWebhookHeaders` test class covering:
  - GET delivery omits `Content-Type` and `X-Hindsight-Signature` while preserving `X-Hindsight-Event`.
  - GET delivery without secret also omits both headers.
  - POST delivery continues to send `Content-Type: application/json` and a valid HMAC signature.

## Verification
- Ran webhook delivery and header tests in `tests/test_webhooks.py`: all passed.
- Ran `./scripts/hooks/lint.sh` (`ruff check`, `ruff format`, vulture, knip): all passed cleanly.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -3384,13 +3384,23 @@ class MemoryEngine(MemoryEngineInterface):
         else:
             payload_bytes = str(raw_payload).encode()
 
-        headers: dict[str, str] = {
-            "Content-Type": "application/json",
-            "X-Hindsight-Event": event_type,
-            **http_config.headers,
-        }
-        if secret and self._webhook_manager:
-            headers["X-Hindsight-Signature"] = self._webhook_manager._sign_payload(secret, payload_bytes)
+        is_get = http_config.method.upper() == "GET"
+        if is_get:
+            # GET has no body, so Content-Type and X-Hindsight-Signature are
+            # meaningless (the signature would be over bytes the receiver never
+            # gets). Send only the event header plus any caller-supplied headers.
+            headers: dict[str, str] = {
+                "X-Hindsight-Event": event_type,
+                **http_config.headers,
+            }
+        else:
+            headers = {
+                "Content-Type": "application/json",
+                "X-Hindsight-Event": event_type,
+                **http_config.headers,
+            }
+            if secret and self._webhook_manager:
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Thanks for picking this up, and for the write-up — the reasoning is sound and it is exactly the option we pointed at in #3368.

We're going to pass on it, though, and that's about the direction rather than the patch. In practice the two headers are inert: a receiver that can't verify the signature just ignores it, the HMAC is keyed so it discloses nothing, and `Content-Type` on a body-less GET is meaningless but harmless. So the change is honesty rather than a fix — and the one thing it *does* change is that any receiver treating the mere presence of `X-Hindsight-Signature` as a weak auth signal would start failing after a version bump. Small, but it's a real downside against no user-visible upside.

The part actually worth fixing is the one underneath: GET deliveries carry no event data at all. `params` are static caller-configured strings and the payload is serialized and then dropped, so a GET receiver only learns that *some* event of type X fired and has to call back to find out what happened. The misleading headers are a symptom of that, not the defect. When we do touch this, it will be to either give GET a real payload (event data in the query string, signature over the URL) or drop GET support — and we'd want to hear from someone actually running GET webhooks before choosing.

Sorry to send you away empty-handed after the effort — the analysis in the description was genuinely useful in getting us to that conclusion. If you do have a GET webhook use case of your own, an issue describing it would be the most useful next step.

PR #3669, opened 2026-08-20
store-owned: finish Protocol-B removal (PG-free zero-batch tracking)

nicoloboschi:
## What

Completes the store-owned / PG-free write path so a store that owns retain, entities, and document
metadata (`store_owned_retain_for(bank_id)` / `owns_document_store_for(bank_id)`) no longer goes
through cross-store write-group (Protocol-B) transactions for **any** operation.

Consolidation, curation edits, unit deletes, delete-document/where, delta re-ingest, the main
streaming write, and the per-batch 0-fact path were already converted in the preceding commits. This
adds the **one remaining gap**:

### The fix — zero-batch document tracking

`_streaming_retain_batch`'s post-loop "no batch processed" finalizer created a Postgres `documents`
row and opened a `begin_txn` / `decide_txn` write-group **even for a store-owned retain**. It is
reached when a retain yields *zero* batches:

- empty or gibberish content (no facts extracted at all), or
- a recovery where every chunk was already committed on a prior attempt.

In that window the per-batch 0-fact branch's store-owned path never runs, so `doc_tracking_done`
stays False and the finalizer falls through to the legacy Postgres path. That reintroduces both a
Postgres write and an undecided write-group on the supposedly PG-free path.

The finalizer now takes the same store-owned branch the per-batch 0-fact path uses: for a
store-owned retain, drop the document's prior memories with one plain store-side
delete-by-document — no Postgres row, no write-group, nothing left undecided to stall the store's
indexer.

## Also

A comments-only commit describes the store-owned delete generically (drops an internal backend name
from three comments); no behavioural change.

## Status

Draft — not for merge yet. Opened to make the completed store-owned series reviewable against
`main` and to continue the remaining work here.

## Test notes

- `orchestrator.py` and `memory_engine.py` parse clean.
- The finalizer change mirrors the reviewed per-batch 0-fact store-owned branch (same guard, same
  `delete_document(conn=None, ...)` shape).
- Still to run here: the retain zero-batch / recovery paths and the store-owned e2e suite.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -1395,8 +1395,19 @@ async def _run_consolidation_job(
             # they are durable-but-invisible in the external store while this batch runs its LLM work. The
             # witness row + decide happen in ONE short transaction at the end (below) — we must not
             # hold a Postgres transaction across the LLM calls in the sub-batch loop.
+            #
+            # A store that owns the whole retain (store_owned_retain) keeps ALL of this batch's memory
+            # writes — observation upserts/deletes and the mark_consolidated stamps — in ITS store, not
+            # Postgres, so there is nothing to make atomic with a Postgres witness. Skip the write-group
+            # entirely (``_batch_txn = None`` → the writes below are plain, immediately-visible writes).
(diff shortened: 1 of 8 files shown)


nicoloboschi:
Superseded by #3696, which consolidates the store-owned writes, the delta re-ingest CAS, and the PG-free zero-batch finalizer into one branch so the store interface and its implementers version together.

PR #3672, opened 2026-08-20
refactor(bank): eliminate redundant existence queries in _ensure_bank_exists

Sanderhoff-alt:
## Problem

`_ensure_bank_exists` is the lazy bank-create every write path goes through. For a bank that already exists — the overwhelmingly common case — it did more work than it needed to:

- **Without a validator:** it went straight to `get_or_create_bank_profile`, which acquires a connection, opens a transaction, runs `SELECT name, disposition, mission`, parses the disposition JSON, and throws the profile away. Three round trips (`BEGIN`, `SELECT`, `COMMIT`) plus a JSON decode to answer a yes/no question.
- **With a validator:** it ran its own `SELECT 1` existence check to decide whether to call `validate_create_bank`, and then called `get_or_create_bank_profile*()` anyway — which repeated the SELECT. Two existence queries per call.

On the missing-bank path the second SELECT inside `get_or_create_bank_profile*()` was likewise redundant: the caller had just established the row wasn't there.

## Change

Hoist a single existence probe to the top of `_ensure_bank_exists`, shared by both the validator and non-validator paths, and return immediately when it says the bank is there. Everything below the probe now only runs on the create path.

New primitives in `bank_utils.py`, all of which existing code already contained in some form:

- `bank_exists_on_conn` / `bank_exists` — narrow `SELECT 1` probes. `bank_exists` deliberately runs no transaction; it is the read-only fast path.
- `create_bank_row_on_conn` — the `INSERT ... ON CONFLICT (bank_id) DO NOTHING RETURNING bank_id` plus per-bank vector index creation, lifted unchanged out of `get_or_create_bank_profile_on_conn`.
- `create_bank_if_missing` — the dedicated-connection variant, keeping the whole-transaction deadlock retry that `get_or_create_bank_profile` has.
- `get_bank_profile_if_exists_on_conn` — the connection-bound profile read, so `get_bank_profile_if_exists` and `get_or_create_bank_profile_on_conn` stop carrying two copies of it.

`get_or_create_bank_profile*` keep their exact previous behaviour, now composed from those pieces. `retain`'s `fact_storage.ensure_bank_exists` was a third hand-rolled copy of the same INSERT; it now delegates to `create_bank_row_on_conn`, so a bank born through retain is identical to one born through the profile helpers.

### Net effect per call

| case | before | after |
|---|---|---|
| existing bank, no validator | `BEGIN` + profile `SELECT` + `COMMIT` | one `SELECT 1` |
| existing bank, validator | `SELECT 1` + `BEGIN` + profile `SELECT` + `COMMIT` | one `SELECT 1` |
| missing bank, validator | `SELECT 1` + `BEGIN` + `SELECT` + `INSERT` + `COMMIT` | `SELECT 1` + `BEGIN` + `INSERT` + `COMMIT` |
| missing bank, no validator | one connection: `SELECT` + `INSERT` | two connections: `SELECT`, then `INSERT` |

The last row is the one regression, and it is deliberate: the validator hook between the probe and the insert may do arbitrary I/O, so a connection must not be held across it. It costs one extra acquisition exactly once per bank, on the call that creates it.

## Concurrency

The probe, the validator hook and the insert are **not** serialised, and cannot be — `validate_create_bank` is an extension hook, so neither an advisory lock nor a held connection may span it. Concurrent callers can therefore all observe the bank as missing and all run the validator; the unique constraint arbitrates, giving exactly one caller `created=True`, and only that caller applies the default bank template.

This is unchanged from before: on `main`, a caller that saw a missing bank already validated and *then* re-entered `get_or_create_bank_profile`. Hoisting the probe made the contract visible, it did not introduce it. It is now written down on `_ensure_bank_exists` and pinned by tests.

## Tests

`tests/test_ensure_bank_exists_optimization.py`:

- existing bank issues no insert and never reaches `validate_create_bank`, with and without a validator, on both the `conn=` and own-connection paths
- missing bank probes once, validates once, inserts once — and the follow-up call returns on the probe
- a rejected `validate_create_bank` leaves no row
- concurrent creation: exactly one `created=True`, exactly one default-template application
- barrier-driven creation race — caller 2 probes as missing, stalls inside its validator until caller 1 commits, then reaches its own insert — asserting `created=False` for the loser, and (in the rejected variant) that caller 1's bank survives caller 2's rejection
- the extracted primitives standalone, including the `ON CONFLICT` arm

`ruff check`, `ruff format`, `ty check` clean. `test_bank_utils`, `test_bank_lifecycle_deadlock_retry`, `test_repair_bank_vector_indexes`, `test_extensions` and the new file pass together (107 tests) — `test_repair_bank_vector_indexes` is the one that guards #2645, i.e. that every bank-create path still builds its per-bank vector indexes.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -11444,32 +11444,46 @@ class MemoryEngine(MemoryEngineInterface):
             if created:
                 await self._apply_default_bank_template(bank_id, rc)
 
+        Concurrency & Validation:
+          * Lockless Optimistic Concurrency: To avoid advisory lock deadlocks
+            and holding transactions across async validator hooks, all callers
+            observing an uncreated bank state independently evaluate
+            `validate_create_bank` before attempting database insertion.
+          * Atomic Arbitration: The database unique constraint arbitrates
+            competing inserts (via `INSERT ... ON CONFLICT DO NOTHING`). Only
+            the winning caller receives `created=True` and applies the default
+            bank template; concurrent losers safely receive `created=False`.
+          * Existing Bank Bypass: Once a bank is committed, subsequent callers
+            observe `exists=True` on their initial probe, returning `False`
+            immediately and bypassing `validate_create_bank`.
+
         Returns:
             True if the bank was freshly created on this call.
         """
         backend = await self._get_backend()
-        if self._operation_validator:
-            if conn is not None:
-                exists = await conn.fetchval(f"SELECT 1 FROM {fq_table('banks')} WHERE bank_id = $1", bank_id)
-            else:
-                exists = await bank_utils.get_bank_profile_if_exists(backend, bank_id)
-            if not exists:
-                from hindsight_api.extensions import CreateBankContext
+        if conn is not None:
+            exists = bool(await conn.fetchval(f"SELECT 1 FROM {fq_table('banks')} WHERE bank_id = $1", bank_id))
+        else:
+            exists = bool(await bank_utils.get_bank_profile_if_exists(backend, bank_id))
+        if exists:
+            return False
 
-                ctx = CreateBankContext(
-                    bank_id=bank_id,
-                    request_context=request_context,
-                )
-                await self._validate_operation(self._operation_validator.validate_create_bank(ctx))
+        if self._operation_validator:
+            from hindsight_api.extensions import CreateBankContext
+
+            ctx = CreateBankContext(
+                bank_id=bank_id,
+                request_context=request_context,
+            )
+            await self._validate_operation(self._operation_validator.validate_create_bank(ctx))
 
         if conn is not None:
-            result = await bank_utils.get_or_create_bank_profile_on_conn(conn, bank_id, ops=backend.ops)
-            return result.created
+            return await bank_utils.create_bank_row_on_conn(conn, bank_id, ops=backend.ops)
 
-        result = await bank_utils.get_or_create_bank_profile(backend, bank_id)
-        if result.created:
+        created = await bank_utils.create_bank_if_missing(backend, bank_id)
+        if created:
             await self._apply_default_bank_template(bank_id, request_context)
-        return result.created
+        return created
 
     async def get_bank_config(
         self,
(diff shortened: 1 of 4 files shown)


r266-tech (commented):
The existence probe, create-bank validation, and idempotent insert are separate steps. Two concurrent calls can both observe a missing bank; after one creates it, the other can still run (and be rejected by) `validate_create_bank`, even though the bank is now existing and would normally bypass that validation. Both calls can also invoke a stateful validator despite only one insert winning, so the stated existing-bank protection does not hold under a creation race.

Please define the intended concurrent validator semantics and coordinate validation with the winning create operation, or explicitly make duplicate validation an accepted contract. Add a barrier-based race test that records validator calls and covers a second caller reaching validation after the first has committed.

This review is for PR #3672 at head `24d550c98a156611e88dc7a3df835d21584ffbc4`. The bound checkout matched the authority-bound head, and `git diff --check` passed. The PR's reported test, lint, and type-check results were not independently reproduced here; local git/test execution was unavailable in the read-only environment.

Sanderhoff-alt:
Thanks for the thorough review and for highlighting the concurrent validator race interleaving!

### Concurrent Validator Semantics & Design Rationale
In accordance with the project's lockless architecture (advisory locks are avoided to prevent cross-transaction deadlocks and holding database connections across async validator hooks), `_ensure_bank_exists` adheres to optimistic concurrency semantics:
1. **Pre-write Authorization**: Callers observing an uncreated bank state act as prospective creators and must pass `validate_create_bank` before attempting database writes.
2. **Creation Race Contract**: When concurrent callers race to lazily initialize the same missing bank, each competing caller independently evaluates `validate_create_bank`. The database constraint (`INSERT ... ON CONFLICT DO NOTHING`) deterministically arbitrates the winning insert, ensuring `_apply_default_bank_template` executes strictly once (`created=True` for the winner, `created=False` for subsequent callers).
3. **Existing Bank Fast-Path**: Once a bank is committed, subsequent calls observe `exists=True` on their initial probe, returning `False` immediately and bypassing `validate_create_bank`.

### Barrier-Based Race Tests Added
We have added two deterministic barrier tests in `tests/test_ensure_bank_exists_optimization.py`:
1. `test_ensure_bank_exists_creation_race_barrier_delayed_validation`: Caller 2 observes the missing bank, Caller 1 commits its insert, and Caller 2 subsequently validates and executes its idempotent insert. Asserts both validation calls are recorded, the template is applied exactly once, Caller 1 reports `created=True`, and Caller 2 reports `created=False`.
2. `test_ensure_bank_exists_creation_race_barrier_delayed_validation_rejected`: Same race interleaving where Caller 2's delayed validation rejects with `OperationValidationError`, asserting Caller 1's committed bank remains intact and template execution is unimpacted.

Both the PR description and `_ensure_bank_exists` docstring have been updated to explicitly document these concurrent validator semantics.

Issue #4, opened 2025-12-01
500 error when writing memory


cdbartholomew:
Got an error when trying to write to hindsight:

```
ERROR:hindsight_api.engine.llm_wrapper:Unexpected error during LLM call: UnicodeEncodeError: 'utf-8' codec can't encode character '\ud83d' in position 29085: surrogates not allowed
ERROR:hindsight_api.api.http:Error in /v1/default/banks/cc-mem-rider/memories (retain): 'utf-8' codec can't encode character '\ud83d' in position 29085: surrogates not allowed

Traceback:
Traceback (most recent call last):
  File "/Users/chrisbartholomew/git/hindsight/hindsight-api/hindsight_api/api/http.py", line 1772, in api_retain
    result = await app.state.memory.retain_batch_async(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chrisbartholomew/git/hindsight/hindsight-api/hindsight_api/engine/memory_engine.py", line 723, in retain_batch_async
    return await self._retain_batch_async_internal(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chrisbartholomew/git/hindsight/hindsight-api/hindsight_api/engine/memory_engine.py", line 763, in _retain_batch_async_internal
    return await orchestrator.retain_batch(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chrisbartholomew/git/hindsight/hindsight-api/hindsight_api/engine/retain/orchestrator.py", line 99, in retain_batch
    extracted_facts, chunks = await fact_extraction.extract_facts_from_contents(
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chrisbartholomew/git/hindsight/hindsight-api/hindsight_api/engine/retain/fact_extraction.py", line 964, in extract_facts_from_contents
    all_fact_results = await asyncio.gather(*fact_extraction_tasks)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chrisbartholomew/git/hindsight/hindsight-api/hindsight_api/engine/retain/fact_extraction.py", line 898, in extract_facts_from_text
    chunk_results = await asyncio.gather(*tasks)
[...]

PR #3, opened 2025-11-26
Ai-fitness-coach

benfrank241:

Issue #2932, opened 2026-07-23
Feature: per-request LLM credentials for multi-user / enterprise deployments


yossiovadia:
## Summary

In enterprise / team deployments, users often have individual cloud credentials — GCP per-user service accounts, AWS IAM roles, Azure managed identities — rather than a shared API key. Today Hindsight requires LLM configuration at server startup (`HINDSIGHT_API_LLM_PROVIDER`, `HINDSIGHT_API_LLM_API_KEY`, etc.) and that config is fixed for the process lifetime. There is no way to pass per-request LLM credentials.

This blocks several legitimate deployment patterns:
- **Per-user GCP Vertex AI service accounts** — each team member authenticates with their own service account; a shared key is either impossible (per-user OAuth tokens) or a policy violation
- **Per-user billing attribution** — route each user's retain/recall calls through their own API account so cloud billing lands on the right cost center
- **Orgs that prohibit shared API keys** — security policies require per-identity credentials; a server-wide shared key fails compliance audits
- **Scales naturally** — new users bring their own credentials; no server reconfiguration or redeployment needed

**Concrete example**: My team (two engineers at Red Hat) runs a shared Hindsight server on GCP. Each of us has our own Vertex AI service account. We want to retain memories under our individual accounts so GCP billing is attributed correctly and we're not violating the "no shared service accounts" policy. Today this requires running one Hindsight server per user.

## Current Architecture

`MemoryEngine.__init__` constructs `LLMConfig` once at startup, which calls `create_llm_provider()` and instantiates the SDK client:

```python
self._retain_llm_config = LLMConfig(provider=retain_provider, api_key=..., model=..., ...)
```

The per-request path (`with_config(resolved_config)`) applies only `llm_gemini_safety_settings` — provider, model, API key, and base URL are excluded from `_CONFIGURABLE_FIELDS` (and most live in `_CREDENTIAL_FIELDS`). `RetainRequest` has no `llm_config` field today.

## Proposed API Change

Add an optional `llm_config` field to `RetainRequest` (and analogously to any future consolidation endpoints):

```json
POST /v1/default/banks/{bank_id}/memories
{
  "items": [...],
  "llm_config": {
    "provider": "vertexai",
    "model": "gemini-2.5-flash",
    "vertexai_project_id": "my-gcp-project",
    "vertexai_region": "us-central1",
    "vertexai_service_account_key": { ... }
  }
}
```

[...]

Issue #2235, opened 2026-06-16
Feature Request: Team, agent, and memory-level access control


Sanderhoff-alt:
## Summary

I would like to request a discussion around a more detailed permission and sharing model for the open-source Hindsight server.

The current OSS authorization model seems intentionally simple:

- `ApiKeyTenantExtension`: one shared API key mapped to the configured schema.
- `SupabaseTenantExtension`: one Supabase user JWT mapped to a schema derived from that user's `sub` claim.
- Banks are the main application-level resource boundary exposed by the API, but the built-in OSS tenant extensions do not appear to provide organization/team membership, scoped API keys, or memory-level authorization.

This works well for simple self-hosted or per-user-isolated deployments, but it does not fully cover collaborative/team memory use cases.

## Use Cases

I would like Hindsight OSS to better support collaborative memory deployments such as:

- A single tenant / organization with multiple users.
- Users with different roles inside the same tenant, such as Owner, Admin, Member, service account, or agent.
- A deployment where each user can have private memory while also participating in shared organization memory.
- Multiple agents working under the same tenant, where each agent may need private working memory and access to selected shared memory.
- Scoped API access for different applications, agents, services, or environments.
- Read-only, write-only, or limited-access integrations.
- Controlled sharing of selected memories between users or agents.
- Optional sharing across tenant or organization boundaries, if the project decides this is an important supported model.

## Why This Matters

For collaborative AI agent systems, memory is not always either fully private or fully shared.

Common patterns may include:

- A user has private memories.
- A team has shared memories.
- An agent has its own working memory.
- Some memories should be visible to selected users or agents.
- Some API keys should only access specific banks, agents, sessions, or memory subsets.
- Some users should read shared memories but not modify them.
- Some agents should write memories but not delete or share them.

The current `SupabaseTenantExtension` provides strong isolation, but each Supabase user effectively becomes their own tenant/schema. That makes organization-level sharing difficult.

Hindsight Cloud appears to support a richer organization model with team roles and bank-scoped API keys. It would be useful to discuss whether some version of that permission model should exist in the OSS server as well, probably as an optional capability rather than a requirement for simple deployments.

[...]

PR #3675, opened 2026-08-20
feat(extensions): add StaticKeysTenantExtension — env-configured per-user API keys with per-schema isolation

### LINKED ISSUE #3674: Feature Request: StaticKeysTenantExtension — env-configured per-user API keys with per-schema isolation
## Use Case

I run a self-hosted Hindsight deployment for a small team / homelab (single node, Docker Compose, external Postgres). I need multiple users — each with their own private memory — without standing up an identity provider (no Supabase, no OAuth), and without running a users database.

## Problem Statement

The built-in tenant extensions cover two ends:
- `ApiKeyTenantExtension` — one shared key, one schema: no per-user isolation.
- `SupabaseTenantExtension` — per-user schema isolation, but requires Supabase Auth (external IdP + network dependency), which a self-hosted deployment may not want.

There is no built-in middle option: static, operator-managed per-user keys with per-user schema isolation, fully self-hosted.

## How This Feature Would Help

With this extension, each user gets their own API key and their own PostgreSQL schema (`user_<id>`), provisioned lazily on first access — memory isolation at the database level, worker consolidation per user, no external services. Key rotation = edit env + restart.

## Proposed Solution

A new built-in `TenantExtension` in `hindsight_api.extensions.builtin.multi_key_tenant` (`StaticKeysTenantExtension`):

- Config:
  - `HINDSIGHT_API_TENANT_USERS="user1:key1,user1:key2,user2:key3"` (required; multiple keys may map to one user)
  - `HINDSIGHT_API_TENANT_SCHEMA_PREFIX=user` (optional, default `user` → `user_<user_id>` schemas)
  - `HINDSIGHT_API_TENANT_MCP_AUTH_DISABLED=true` (optional, parity with ApiKeyTenantExtension)
- `Authorization: Bearer <key>` → per-user schema, provisioned lazily + cached (`run_migration`); unknown/missing key → 401
- Constant-time key comparison (`hmac.compare_digest`)
- Sets `RequestContext.tenant_id` / `api_key_id` for usage metering
- `list_tenants()` returns all configured users → worker/maintenance per schema
- `/health` stays public

I already have an implementation + tests (30) + docs on a branch and would be happy to open the PR.

## Alternatives Considered

- Supabase Auth + `SupabaseTenantExtension` — works but adds an external IdP dependency.
- A custom extension mounted only in my stack — works but isn't reusable for others.
- Custom middleware/proxy for authz — maintainers prefer the extension mechanism (per #2235 discussion).

## Priority

## Additional Context

This is the deliberately small, env-only slice of #2235's authorization discussion (which also spawned closed draft PR #2262 — a larger Supabase-org approach). It also complements #2932 (per-request LLM credentials): this gives the per-identity anchor.

Happy to adjust naming/behavior per maintainer feedback — open questions:
1. `mcp_auth_disabled=true` currently falls back to the base schema (ApiKey parity); is that acceptable for multi-user, or should it refuse to start?
2. Plaintext keys in env — acceptable for v1, or is a hashed variant preferred?
3. Is `StaticKeysTenantExtension` the right name?

## Checklist

rafaelkallis:
## Summary

Adds a `TenantExtension` mapping static API keys (from env) to per-user PostgreSQL schemas — **shipped in the extensions registry** (`hindsight-extensions/static-keys-tenant/`), not the server, per review: it follows the `supabase-tenant` layout (package, test-only `pyproject.toml`, Dockerfile, README, tests, CI job). Closes #3674.

Fully self-hosted multi-user memory isolation with no users table and no new services: a bridge between `ApiKeyTenantExtension` (single shared key) and `SupabaseTenantExtension` (external IdP).

## Configuration

```bash
HINDSIGHT_API_TENANT_EXTENSION=hindsight_ext_static_keys_tenant:StaticKeysTenantExtension
HINDSIGHT_API_TENANT_USERS=user1:key1,user1:key2,user2:key3   # required; multiple keys may map to one user
HINDSIGHT_API_TENANT_SCHEMA_PREFIX=user                       # optional, default "user" → user_<user_id> schemas
```

`HINDSIGHT_API_TENANT_MCP_AUTH_DISABLED` is deliberately **not supported** — setting it fails at startup (MCP clients always authenticate with a user's key; no isolation bypass).

## Behavior

- `Authorization: Bearer <key>` → authenticated as the mapped user
- Per-user schema `{prefix}_{user_id}` provisioned lazily on first access, serialized per schema with an `asyncio.Lock` (no concurrent double-migration); provisioning failure raises `AuthenticationError` (not cached)
- User ids normalized (lowercased, dashes → underscores) so case variants collapse onto one canonical schema, consistent with Postgres identifier folding; collisions (dash normalization, >63-byte identifiers) rejected at startup
- Keys must be **ASCII and comma-free** (documented); constant-time comparison on bytes via `hmac.compare_digest`
- Unknown/missing key → 401 (`AuthenticationError`); non-ASCII bearer tokens → 401, never a 500
- `authenticate()` sets `context.tenant_id` and a stable non-secret `api_key_id` (truncated sha256 of the key) — metering can distinguish a user's multiple keys
- Config error messages never echo key material (entry index / user id / derived key_id only)
- `list_tenants()` returns all configured users so the worker and maintenance sweep per-user schemas; `hindsight-admin run-db-migration` pre-provisions them (documented)
- `/health` stays public

## Testing

- 49 unit tests (`uv run pytest` in the package, no DB): init validation (missing/invalid config, SQL-injection ids, schema collisions, duplicate keys, key-format edge cases), auth (valid/invalid/missing/non-ASCII, metering fields, per-key ids, schema provision-once + concurrency + failure-not-cached), MCP (auth required, flag refused), `list_tenants`, loader integration, package entrypoint
- CI: `test-extension-static-keys-tenant` runs the suite and builds the image on the `latest-slim` base (detect-changes filter + gate wired)
- `ruff` / `ruff format --check` / `ty` clean

## Compatibility

Fully opt-in; no changes to the server package. Only activates when `HINDSIGHT_API_TENANT_EXTENSION` points at it.

## Related

- Closes #3674 (feature request)
- Ref #2235 (authorization discussion — deliberately small env-only slice), #2932 (per-request LLM credentials — this provides the per-identity anchor)

### DIFF
--- a/hindsight-api-slim/hindsight_api/extensions/builtin/multi_key_tenant.py
+++ b/hindsight-api-slim/hindsight_api/extensions/builtin/multi_key_tenant.py
@@ -1,0 +1,226 @@
+"""Multi-key tenant extension: env-configured static API keys with per-user schema isolation.
+
+Bridges ApiKeyTenantExtension (one shared key) and SupabaseTenantExtension (external
+IdP): a fully self-hosted, single-node multi-user mode where users and their API keys
+are declared in environment variables. Each user maps to their own PostgreSQL schema
+({prefix}_{user_id}), provisioned lazily on first access — memory isolation at the
+database level, with the worker processing every tenant via list_tenants().
+
+Features:
+    - In-memory user store from env vars (no users table, no external IdP)
+    - Multiple API keys may map to the same user (same isolated schema)
+    - Per-user schema isolation with lazy provisioning + caching
+    - Constant-time key comparison (hmac.compare_digest)
+    - Works for HTTP and MCP auth
+    - Usage-metering: sets RequestContext.tenant_id / api_key_id after auth
+    - Fail-fast on misconfiguration
+
+Configuration via environment variables:
+    HINDSIGHT_API_TENANT_EXTENSION=hindsight_api.extensions.builtin.multi_key_tenant:StaticKeysTenantExtension
+    HINDSIGHT_API_TENANT_USERS=user1:key1,user1:key2,user2:key3   # required, comma-separated user:key pairs
+    HINDSIGHT_API_TENANT_SCHEMA_PREFIX=user            # optional, default: "user" (creates user_<user_id> schemas)
+    HINDSIGHT_API_TENANT_MCP_AUTH_DISABLED=true        # optional, disable auth for MCP endpoints
+
+Usage:
+    Clients pass their API key in the Authorization header:
+
+    curl -H "Authorization: Bearer <api_key>" \\
+        http://localhost:8888/v1/default/banks
+
+Authors: Rafael Kallis
+License: MIT
+"""
+
+from __future__ import annotations
+
+import hmac
+import logging
+import re
+
+from hindsight_api.config import get_config
+from hindsight_api.extensions.tenant import AuthenticationError, Tenant, TenantContext, TenantExtension
+from hindsight_api.models import RequestContext
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["StaticKeysTenantExtension"]
+
+# Schema prefix must be a valid Postgres identifier component (letters, digits, underscores)
+_SCHEMA_PREFIX_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
+
+# User IDs appear in schema names, so they must be safe as a Postgres identifier
+# component: start with a letter/underscore, then letters/digits/underscore/dash.
+# Dashes are normalized to underscores before building the schema name.
+_USER_ID_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_-]*$")
+
+
+class StaticKeysTenantExtension(TenantExtension):
+    """
+    TenantExtension mapping env-configured static API keys to per-user schemas.
+
+    Each entry in ``HINDSIGHT_API_TENANT_USERS`` is a ``user_id:api_key`` pair.
+    Multiple keys may map to the same user. Authenticated requests are mapped to
+    schema ``{prefix}_{user_id}`` (dashes in the user id normalized to underscores),
+    provisioned lazily on first access.
+
+    Example:
+        HINDSIGHT_API_TENANT_USERS=rafael:key-a,sophie:key-b
+        HINDSIGHT_API_TENANT_SCHEMA_PREFIX=user
+
+        User "rafael" with key "key-a" gets schema "user_rafael";
+        user "sophie" with key "key-b" gets schema "user_sophie".
+    """
+
+    def __init__(self, config: dict[str, str]) -> None:
+        """
+        Initialize with configuration from environment variables.
+
+        Config keys are derived from HINDSIGHT_API_TENANT_* env vars:
+        - HINDSIGHT_API_TENANT_USERS -> config["users"] (required)
+        - HINDSIGHT_API_TENANT_SCHEMA_PREFIX -> config["schema_prefix"] (optional, default "user")
+        - HINDSIGHT_API_TENANT_MCP_AUTH_DISABLED -> config["mcp_auth_disabled"] (optional)
+
+        Args:
+            config: Dictionary of configuration values from environment.
+
+        Raises:
+            ValueError: If required configuration is missing or invalid.
+        """
+        super().__init__(config)
+
+        users_raw = config.get("users", "")
+        self.schema_prefix = config.get("schema_prefix", "user")
+
+        if not users_raw.strip():
+            raise ValueError(
+                "HINDSIGHT_API_TENANT_USERS is required when using StaticKeysTenantExtension. "
+                'Format: "user1:key1,user2:key2"'
+            )
+
+        if not _SCHEMA_PREFIX_RE.match(self.schema_prefix):
+            raise ValueError(
+                f"Invalid schema_prefix '{self.schema_prefix}'. "
+                "Must be a valid Postgres identifier (letters, digits, underscores, starting with a letter or underscore)."
+            )
+
+        # Parse users into an API-key -> (user_id, schema_name) map.
+        # Multiple keys may map to the same user. Also keep a per-user map so
+        # list_tenants() can return every configured tenant, even before any
+        # authentication has happened.
+        self._key_to_user: dict[str, tuple[str, str]] = {}
+        self._users: dict[str, str] = {}  # user_id -> schema_name
+
+        for entry in users_raw.split(","):
+            entry = entry.strip()
+            if not entry:
+                continue
+            if ":" not in entry:
+                raise ValueError(
+                    f"Invalid HINDSIGHT_API_TENANT_USERS entry '{entry}'. "
+                    'Expected format "user_id:api_key".'
+                )
+            user_id, api_key = entry.split(":", 1)
+            user_id = user_id.strip()
+            api_key = api_key.strip()
+            if not user_id or not api_key:
+                raise ValueError(
+                    f"Invalid HINDSIGHT_API_TENANT_USERS entry '{entry}'. "
+                    "user_id and api_key must be non-empty."
+                )
+            if not _USER_ID_RE.match(user_id):
+                raise ValueError(
+                    f"Invalid user_id '{user_id}' in HINDSIGHT_API_TENANT_USERS. "
+                    "Must start with a letter or underscore, then letters, digits, underscores or dashes."
+                )
+
+            safe_user_id = user_id.replace("-", "_")
+            schema_name = f"{self.schema_prefix}_{safe_user_id}"
+            # Multiple keys may map to the same user (and thus the same schema).
+            self._key_to_user[api_key] = (user_id, schema_name)
+            self._users[user_id] = schema_name
+
+        # Track initialized schemas to avoid redundant migrations
+        self._initialized_schemas: set[str] = set()
+
+        self.mcp_auth_disabled = config.get("mcp_auth_disabled", "").lower() in ("true", "1", "yes")
+
+    # ------------------------------------------------------------------
+    # Authentication
+    # ------------------------------------------------------------------
+
+    async def authenticate(self, context: RequestContext) -> TenantContext:
+        """
+        Validate the API key and return tenant context.
+
+        Args:
+            context: Request context containing the API key (the Authorization header).
+
+        Returns:
+            TenantContext with schema_name set to ``{prefix}_{user_id}``.
+
+        Raises:
+            AuthenticationError: If the key is missing or unknown.
+        """
+        key = context.api_key
+        if not key:
+            raise AuthenticationError("Missing Authorization header. Expected: Bearer <api_key>")
+
+        match = self._key_to_user.get(key)
+        if match is None:
+            # Constant-time comparison against each configured key to avoid
+            # timing side-channels on the key value itself.
+            for configured_key, (user_id, schema_name) in self._key_to_user.items():
+                if hmac.compare_digest(key, configured_key):
+                    match = (user_id, schema_name)
+                    break
+
+        if match is None:
+            raise AuthenticationError("Invalid API key")
+
+        user_id, schema_name = match
+
+        # Initialize schema on first access
+        if schema_name not in self._initialized_schemas:
+            await self._initialize_schema(schema_name)
+
+        # Usage metering: the HTTP/MCP layers read these fields back after auth
+        # to attribute operations to a tenant / API key.
+        context.tenant_id = user_id
+        context.api_key_id = user_id
+
+        return TenantContext(schema_name=schema_name)
+
+    async def authenticate_mcp(self, context: RequestContext) -> TenantContext:
+        """
+        Authenticate MCP requests.
+
+        If mcp_auth_disabled is set, skip authentication and land in the base
+        schema (parity with ApiKeyTenantExtension). Otherwise, delegate to
+        authenticate().
+        """
+        if self.mcp_auth_disabled:
+            return TenantContext(schema_name=get_config().database_schema)
+        return await self.authenticate(context)
+
+    # ------------------------------------------------------------------
+    # Schema management
+    # ------------------------------------------------------------------
+
+    async def _initialize_schema(self, schema_name: str) -> None:
+        """Run migrations for a new tenant schema and cache the result."""
+        logger.info("Initializing schema: %s", schema_name)
+        try:
+            await self.context.run_migration(schema_name)
+            self._initialized_schemas.add(schema_name)
+            logger.info("Schema ready: %s", schema_name)
+        except Exception as e:
+            logger.error("Schema initialization failed for %s: %s", schema_name, e)
+            raise AuthenticationError(f"Failed to initialize tenant: {e!s}")
+
+    # ------------------------------------------------------------------
+    # Worker discovery
+    # ------------------------------------------------------------------
+
+    async def list_tenants(self) -> list[Tenant]:
+        """Return all configured tenants for worker processing."""
+        return [Tenant(schema=schema, tenant_id=user_id) for user_id, schema in self._users.items()]
--- a/hindsight-api-slim/hindsight_api/extensions/builtin/multi_key_tenant.py
+++ b/hindsight-api-slim/hindsight_api/extensions/builtin/multi_key_tenant.py
@@ -116,16 +116,14 @@ class StaticKeysTenantExtension(TenantExtension):
                 continue
             if ":" not in entry:
                 raise ValueError(
-                    f"Invalid HINDSIGHT_API_TENANT_USERS entry '{entry}'. "
-                    'Expected format "user_id:api_key".'
+                    f"Invalid HINDSIGHT_API_TENANT_USERS entry '{entry}'. Expected format \"user_id:api_key\"."
                 )
             user_id, api_key = entry.split(":", 1)
             user_id = user_id.strip()
             api_key = api_key.strip()
             if not user_id or not api_key:
                 raise ValueError(
-                    f"Invalid HINDSIGHT_API_TENANT_USERS entry '{entry}'. "
-                    "user_id and api_key must be non-empty."
+                    f"Invalid HINDSIGHT_API_TENANT_USERS entry '{entry}'. user_id and api_key must be non-empty."
                 )
             if not _USER_ID_RE.match(user_id):
                 raise ValueError(
--- a/hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
+++ b/hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
@@ -37,6 +37,7 @@ License: MIT
 
 from __future__ import annotations
 
+import hashlib
 import hmac
 import logging
 import re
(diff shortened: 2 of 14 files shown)


hindsight-api-slim/hindsight_api/extensions/builtin/multi_key_tenant.py
```diff
@@ -0,0 +1,224 @@
...
+        if match is None:
+            # Constant-time comparison against each configured key to avoid
+            # timing side-channels on the key value itself.
+            for configured_key, (user_id, schema_name) in self._key_to_user.items():
+                if hmac.compare_digest(key, configured_key):
+                    match = (user_id, schema_name)
+                    break
+
+        if match is None:
+            raise AuthenticationError("Invalid API key")
```
Sanderhoff-alt: This fallback can never produce a match, and it can raise instead of rejecting.

`self._key_to_user.get(key)` already performs an exact equality lookup, so the loop only runs when no configured key equals `key` — `hmac.compare_digest` is therefore guaranteed to return False for every entry. The advertised constant-time comparison never covers a real authentication.

Worse, `hmac.compare_digest` raises `TypeError` when either `str` argument is not ASCII-only, and header values arrive latin-1-decoded, so `Authorization: Bearer <byte >= 0x80>` raises `TypeError` out of `authenticate()` instead of `AuthenticationError`, i.e. a 500 rather than a 401 for an unauthenticated request.

Suggested fix: drop the dict fast path and compare bytes over all configured keys.

```python
match: tuple[str, str] | None = None
key_bytes = key.encode("utf-8", "surrogateescape")
for configured_key, entry in self._key_to_user.items():
    if hmac.compare_digest(key_bytes, configured_key.encode("utf-8", "surrogateescape")):
        match = entry
```

A test for a non-ASCII bearer token and one for the fallback path would pin both.

hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
```diff
@@ -0,0 +1,224 @@
...
+                raise ValueError(
+                    f"Invalid user_id '{user_id}' in HINDSIGHT_API_TENANT_USERS. "
+                    "Must start with a letter or underscore, then letters, digits, underscores or dashes."
+                )
+
+            safe_user_id = user_id.replace("-", "_")
+            schema_name = f"{self.schema_prefix}_{safe_user_id}"
+            # Multiple keys may map to the same user (and thus the same schema).
+            self._key_to_user[api_key] = (user_id, schema_name)
+            self._users[user_id] = schema_name
```
Sanderhoff-alt, line 237: Two distinct users can silently share one schema, which defeats the isolation this extension exists to provide.

`HINDSIGHT_API_TENANT_USERS=jane-doe:k1,jane_doe:k2` maps both users to `user_jane_doe`, so each reads and writes the other's memories. The same happens for user ids that differ only past the 63-byte PostgreSQL identifier limit, and a repeated api_key (`alice:k1,bob:k1`) silently overwrites the earlier entry so `alice`'s key authenticates as `bob`.

Since every other malformed input here fails loudly at init, please reject these too: raise `ValueError` when a computed `schema_name` is already claimed by a different user id, when an api_key appears twice, or when `schema_name` exceeds 63 characters.

hindsight-api-slim/hindsight_api/extensions/builtin/multi_key_tenant.py
```diff
@@ -0,0 +1,224 @@
...
+            raise ValueError(
+                f"Invalid schema_prefix '{self.schema_prefix}'. "
+                "Must be a valid Postgres identifier (letters, digits, underscores, starting with a letter or underscore)."
+            )
+
+        # Parse users into an API-key -> (user_id, schema_name) map.
+        # Multiple keys may map to the same user. Also keep a per-user map so
+        # list_tenants() can return every configured tenant, even before any
+        # authentication has happened.
+        self._key_to_user: dict[str, tuple[str, str]] = {}
```
Sanderhoff-alt: `dict[str, tuple[str, str]]` holds structured data with known fields, which the project standards ask to model as a named type (see `.claude/skills/code-review/SKILL.md`, Type Safety). A small `@dataclass(frozen=True)` with `user_id` and `schema_name` removes the positional unpacking at line 178 and makes the map self-describing; `supabase_tenant.py` predates the rule rather than sanctioning it.

Sanderhoff-alt (commented):
The authentication path does not behave as the description states: the constant-time fallback can never match, and it turns a bad key into a 500 for some inputs. Two other points on tenant-id collisions and the untyped key map.

hindsight-api-slim/hindsight_api/extensions/builtin/multi_key_tenant.py
rafaelkallis: Good catch — you are right on both counts. I have fixed this in `f7c98536`:

- Dropped the dict fast-path entirely; `authenticate()` now always iterates every configured key and compares **bytes** via `hmac.compare_digest` (encoded with `"surrogateescape"`) — so the constant-time comparison actually runs for every attempt, and non-ASCII bearer tokens (latin-1-decoded bytes ≥ 0x80) are now rejected with `AuthenticationError` (401) instead of raising `TypeError` (500).

Implementation in `authenticate()`:
```python
key_bytes = key.encode("utf-8", "surrogateescape")
match: _KeyEntry | None = None
for configured_key, entry in self._key_to_user.items():
    if hmac.compare_digest(key_bytes, configured_key.encode("utf-8", "surrogateescape")):
        match = entry
        break
if match is None:
    raise AuthenticationError("Invalid API key")
```

Tests added as you suggested:
- `test_authenticate_non_ascii_key_is_401_not_500` — non-ASCII bearer → `AuthenticationError`, not `TypeError`.
- `test_authenticate_compares_all_keys_in_constant_time` — spies on `hmac.compare_digest` to pin that an unknown key is compared against **every** configured key as bytes (no fast path).

`ruff`, `ty`, and the full `tests/test_multi_key_tenant.py` suite (36 tests) are green.

hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
rafaelkallis: Agreed — silent schema collisions defeat the whole point of the extension. I have added fail-fast validation in `f7c98536`:

- `ValueError` when a computed `schema_name` is already claimed by a **different** user id (e.g. `jane-doe:k1,jane_doe:k2` → both normalize to `user_jane_doe`).
- `ValueError` when a `schema_name` exceeds the 63-character PostgreSQL identifier limit (`_MAX_SCHEMA_LENGTH = 63`).
- `ValueError` when an `api_key` appears twice (`alice:k1,bob:k1`), instead of silently overwriting the earlier entry.
- The legitimate multi-key-per-user case (`alice:k1,alice:k2` → same user, same schema) still works and is covered by a test that asserts it does **not** raise.

New tests: `test_init_rejects_schema_collision_after_dash_normalization`, `test_init_rejects_schema_name_over_63_chars`, `test_init_rejects_duplicate_api_key`, `test_init_multiple_keys_same_user_does_not_collide`.

hindsight-api-slim/hindsight_api/extensions/builtin/multi_key_tenant.py
rafaelkallis: Fair point — done in `f7c98536`. The key map now uses a small frozen dataclass instead of the tuple:

```python
@dataclass(frozen=True)
class _KeyEntry:
    """A configured API key's mapping: the owning user and its isolated schema."""

    user_id: str
    schema_name: str
```

`self._key_to_user` is now `dict[str, _KeyEntry]`, which removes the positional unpacking in `authenticate()` (now `match.user_id, match.schema_name`) and makes the map self-describing.

rafaelkallis:
**Fixing a tenant-isolation edge case found in review: mixed-case user IDs**

During code review we found that `StaticKeysTenantExtension` allowed mixed-case user IDs (e.g. `Rafael`), which produced mixed-case schema names (`user_Rafael`). That is unsafe:

- The migration path creates the schema **quoted** (`CREATE SCHEMA IF NOT EXISTS "user_Rafael"`), preserving case.
- But the runtime's `fq_table()` does **not** quote the schema (`f"{schema}.{table}"`), so Postgres folds it to lowercase.
- Result: either every query for such a user fails ("schema does not exist"), or — worse — two users like `Rafael` and `rafael` silently collapse onto one schema, breaking the extension's core isolation guarantee.

**Fix (commit `4f0f820e`):** normalize user IDs to lowercase at `__init__` before building the schema name. Case variants (`Rafael` / `rafael` / `RAFAEL`) now resolve to a single canonical tenant and schema, consistent with Postgres identifier folding. Distinct users whose ids collide in non-case ways (e.g. `jane-doe` vs `jane_doe`, or ids colliding past the 63-byte identifier limit) are still rejected loudly, as before.

Added regression tests covering: lowercase normalization, mixed-case + dash normalization, case-insensitive duplicate-user acceptance, and mixed-case dash-collision rejection, plus an auth-level test asserting a mixed-case key resolves onto the lowercased schema and a `list_tenants()` test for the lowercased tenant id.

nicoloboschi (changes requested):
please move this work to https://github.com/vectorize-io/hindsight/tree/main/hindsight-extensions --- thanks!

rafaelkallis:
Done — the work has been moved to the extensions registry as requested.

**What changed (PR now shows only these files):**
- New standalone package: `hindsight-extensions/static-keys-tenant/` with `hindsight_ext_static_keys_tenant/` (package), `pyproject.toml` (test harness only), `Dockerfile` (image packaging), `README.md`, and `tests/` — mirroring `supabase-tenant`.
- Registry entry added in `hindsight-extensions/README.md`.
- Developer docs updated: `hindsight-docs/docs/developer/extensions.md` now points at the registry.

**Nothing remains in `hindsight-api-slim`** — the builtin `multi_key_tenant.py` and its exports are gone, since the extension is no longer bundled with the server (consistent with the `supabase-tenant` precedent).

The extension is imported via `HINDSIGHT_API_TENANT_EXTENSION=hindsight_ext_static_keys_tenant:StaticKeysTenantExtension`, has no third-party dependencies beyond the server, and its 40 unit tests pass (`uv run pytest`, no DB needed — matching the registry dev workflow).

rafaelkallis:
@nicoloboschi thanks for the pointer to the registry — the move is fine and the package now follows the `supabase-tenant` layout. Just want to put the reasoning behind this extension on the record, since I would argue it sits on the "broadly useful" side of the slim boundary:

- **Zero third-party dependencies.** It imports only stdlib plus the server's own `hindsight_api.*` interfaces — the Dockerfile has no `pip install` step, and it runs on the `latest-slim` base unchanged. It would add nothing to the slim image.
- **Primary motivation: one user, many harnesses.** A single-user deployment can mint multiple access tokens for the same user so each agent harness (Claude Code, Codex, Cursor, …) authenticates with its own key while sharing one memory schema — with per-key attribution in usage metering. That is a common self-hosted setup, not a niche one.
- **Secondary: self-hosted multi-user isolation.** Per-profile schema isolation without an IdP or users table — e.g. running multiple Hermes profiles, each with its own isolated memory banks under its own key, all on one deployment. Exactly the gap between the shared-key builtin and the vendor-bound Supabase extension.

If you would prefer this bundled in `hindsight-all-slim` after all, say the word — it is a minimal move on my side. Otherwise I am happy to merge as is, in the registry.

hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
```diff
@@ -0,0 +1,288 @@
...
+                )
+            claimed_by = next((uid for uid, s in self._users.items() if uid != user_id and s == schema_name), None)
+            if claimed_by is not None:
+                raise ValueError(
+                    f"Schema name '{schema_name}' for user_id '{user_id}' is already claimed by "
+                    f"user_id '{claimed_by}'. Two distinct users cannot share one schema."
+                )
+            if api_key in self._key_to_user:
+                raise ValueError(
+                    f"Duplicate API key '{api_key}' in HINDSIGHT_API_TENANT_USERS. Each key must be unique."
```
nicoloboschi: This writes a live credential to stdout / the container log on a misconfigured start.

Same problem on lines 139 and 146: `entry` is the whole `user_id:api_key` pair, so `HINDSIGHT_API_TENANT_USERS=rafael:` produces an error message containing the key of whichever entry was malformed.

An operator who hits any of these three pastes the message into an issue or a chat and has just disclosed a key. Report the user id and the entry's index instead, never the key itself.

hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
```diff
@@ -0,0 +1,288 @@
...
+        user_id, schema_name = match.user_id, match.schema_name
+
+        # Initialize schema on first access
+        if schema_name not in self._initialized_schemas:
+            await self._initialize_schema(schema_name)
+
+        # Usage metering: the HTTP/MCP layers read these fields back after auth
+        # to attribute operations to a tenant / API key.
+        context.tenant_id = user_id
+        context.api_key_id = user_id
```
nicoloboschi: `models.py:26` documents `api_key_id` as identifying **the key**, and multiple keys per user is this extension's selling point — so as written, metering cannot tell which of a user's keys was used, and `api_key_id` is just a duplicate of `tenant_id` set one line above.

Give each `_KeyEntry` a stable derived id at init (e.g. `sha256(key).hexdigest()[:16]`) and set that here. That also gives you a non-secret way to name a key in the error messages flagged above.

hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
```diff
@@ -0,0 +1,288 @@
...
+        # key's position/size via timing AND mean unknown keys never ran the
+        # constant-time loop — compare_digest must always run over all entries.
+        # Keys (and header values) arrive latin-1-decoded, so encode with
+        # "surrogateescape" so any byte sequence round-trips losslessly
+        # instead of raising TypeError (which would surface as a 500, not a 401)
+        # for non-ASCII bearer tokens.
+        key_bytes = key.encode("utf-8", "surrogateescape")
+        match: _KeyEntry | None = None
+        for configured_key, entry in self._key_to_user.items():
+            if hmac.compare_digest(key_bytes, configured_key.encode("utf-8", "surrogateescape")):
```
nicoloboschi: Nit: `configured_key.encode(...)` re-encodes every configured key on every request. Store the bytes on `_KeyEntry` at init and compare against those.

Also, the comment above promises more than the code delivers: the `break` on a match means the number of comparisons still depends on the matching key's position. That is fine in practice (an attacker holding a valid key learns nothing useful), but the comment reads as if the loop is position-independent.

nicoloboschi (changes requested):
Thanks — the move to `hindsight-extensions/` is done properly: the layout, `pyproject.toml`, `Dockerfile` (with `ARG HINDSIGHT_IMAGE`), entrypoint test, registry row and docs blurb all mirror `supabase-tenant`. I checked the branch out locally: **40 tests pass, `ruff check` and `ruff format --check` are clean.**

Two things must change before merge, plus a few smaller ones.

## Must fix

**1. No CI job — none of this ever runs in CI.**
`supabase-tenant` has `test-extension-supabase-tenant` (`.github/workflows/test.yml:3928`), a `detect-changes` filter (`:180`) and a gate entry (`:5603`), and that job also builds its image. This PR touches no workflow file, so its 40 tests and its `Dockerfile` are never exercised on any future change. Please add the same three pieces.

**2. API keys leak into startup logs.** See inline comments — the plaintext key ends up in `ValueError` messages on misconfiguration.

## Should fix

**3. Undocumented key-format constraints.** A key cannot contain a comma (it is the entry separator), and a non-ASCII key can never authenticate: header values arrive latin-1-decoded while env values are utf-8-decoded, so `encode("utf-8", "surrogateescape")` on the two sides does not produce the same bytes. It fails closed, which is fine, but a permanent silent 401 is a miserable debugging session — please state "keys must be ASCII and must not contain a comma" in the README.

**4. `mcp_auth_disabled` on a multi-user extension.** You raised this yourself; my vote is refuse at init rather than ship parity. On `ApiKeyTenantExtension` the flag downgrades one shared key to none. Here it hands anyone unauthenticated MCP access to the base schema, in a deployment whose entire purpose is per-user isolation.

## Nits

- `list_tenants()` returning every configured user rather than only the provisioned ones is the right call for `run-db-migration`, and better than what `supabase-tenant` does. It does mean the poller's fallback path issues one EXISTS probe per configured user on every idle cycle, against schemas that may not exist yet (`worker/poller.py:440`; the exception is swallowed, so it is harmless). Worth a README line pointing at `hindsight-admin run-db-migration` to pre-provision.
- Two concurrent first requests from the same user both call `run_migration`. Inherited from `supabase-tenant`, so not a blocker, but a per-schema `asyncio.Lock` would close it.
- The branch is behind `main` (still merges cleanly), and the PR description is stale: it still describes `hindsight_api.extensions.builtin.multi_key_tenant` and "30 tests".

hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
rafaelkallis: Fixed in `1b11e1a1`. All three sites now report the entry **index** (and the user id once parsed and validated) — never `entry` or the key. The duplicate-key error additionally names the two conflicting user ids via the sha256-derived `key_id` (see the comment on the metering line). Regression tests pin the non-disclosure: the malformed-entry and duplicate-key messages are asserted not to contain key material.

hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
rafaelkallis: Fixed in `1b11e1a1`. Each `_KeyEntry` now carries a stable, non-secret `key_id` — `sha256(key).hexdigest()[:16]`, derived once at init — and `context.api_key_id` is set to it. As you said, `models.py:26` documents `api_key_id` as identifying *the key*, and with multiple keys per user the old code could not tell which one authenticated. Now: same tenant_id, different `api_key_id` per key (test: `test_authenticate_metering_distinguishes_keys_for_same_user`). The digest is also what the duplicate-key error uses to name a key without disclosing it.

hindsight-extensions/static-keys-tenant/hindsight_ext_static_keys_tenant/extension.py
rafaelkallis: Both points addressed in `f112daf9` and `1b11e1a1`:

- The comparison loop now reads pre-encoded bytes off `_KeyEntry`-adjacent state — configured keys are encoded once at init and stored, so the per-request `encode` is gone (only the incoming key is encoded).
- The comment no longer overpromises: it now states the loop stops at the first match, so comparison count still depends on the matching key's position, and why that is harmless (invalid keys traverse the whole list; an attacker holding a valid key already knows where it sits).

rafaelkallis:
All points addressed — one commit per item, pushed to the PR branch:

**Must fix #1 — CI job** (`ec27d176`): `test-extension-static-keys-tenant` mirrors the supabase wiring — `detect-changes` filter + output mapping for `hindsight-extensions/static-keys-tenant/**`, a job that runs `uv sync` + `pytest tests -v` and builds the image on the `latest-slim` base, and the job added to the `report-pr-status` gate.

**Must fix #2 — key leaks** (`1b11e1a1` + inline replies): config errors now report the entry index / validated user id / derived `key_id` — never `entry` or the key. Regression tests assert the messages contain no key material.

**Should fix #3 — key constraints documented** (`57bc49d0`): README states keys must be **ASCII and comma-free**, with the latin-1/utf-8 explanation for why a non-ASCII key would otherwise fail closed with a permanent silent 401. Same commit documents `hindsight-admin run-db-migration` for pre-provisioning (the poller EXISTS-probe nit).

**Should fix #4 — `mcp_auth_disabled`** (`a64e31a7`): adopting your vote — the variable now raises `ValueError` at init (documented in the variable table), and `authenticate_mcp` always delegates to `authenticate()`. The parity behavior is gone.

**Nits**:
- Concurrent first provision per schema now serialized with a per-schema `asyncio.Lock` + re-check (`f112daf9`).
- Configured keys are pre-encoded to `compare_digest`-ready bytes at init; per-request encoding is only the incoming key (`6fe2a989`), and the constant-time comment now states honestly that the loop stops at the first match.

**PR description** rewritten (the old one still said `builtin.multi_key_tenant` and "30 tests").

Checks: 50 tests pass (`uv run pytest`), `ruff check` / `ruff format --check` / `ty check` all clean.

Branch is rebased onto current `main` tip (`7e17a120`) and merges cleanly. Re-requested your review.

nicoloboschi (approved):

PR #3681, opened 2026-08-20
fix(control-plane): reload memories when search is cleared

### LINKED ISSUE #3670: UI bug in Memories screen
### Bug Description

When searching for a term that yields zero results, and then backspacing to zero characters "closes" the search screen and indicates there are "no memories yet". Which is incorrect. 

(Going to the Home Screen and back to the Memories screen restores the search screen)

### Steps to Reproduce

See above

### Expected Behavior

I would expect the search screen to stay open and indicate XX total memories (just like in a new search screen)

### Actual Behavior

See above

### Version

0.9.1

### LLM Provider

Other

oldnicke:
Fixes #3670.

When a zero-result text search was cleared, the Memories view kept the stale zero-result response and incorrectly replaced the search UI with the unfiltered empty-bank state.

This change explicitly reloads the unfiltered graph when an active query is cleared, while preserving tag/scope filters. The empty-bank state is also deferred while that reload is in flight so the search controls remain visible.

Validation:
- `./scripts/hooks/lint.sh`
- `npx tsc --noEmit`
- `npm test -- --run tests/lib/search-query.test.ts`
- `./.githooks/pre-commit`

The full control-plane suite has one pre-existing date-sensitive failure in `tests/lib/relative-time.test.ts`; 143 tests pass.

### DIFF
--- a/hindsight-control-plane/src/components/data-view.tsx
+++ b/hindsight-control-plane/src/components/data-view.tsx
@@ -4,6 +4,7 @@ import { useState, useEffect, useRef, useMemo, useCallback } from "react";
 import { useTranslations } from "next-intl";
 import { client } from "@/lib/api";
 import { useBank } from "@/lib/bank-context";
+import { isSearchQueryCleared } from "@/lib/search-query";
 import { EntityChip, TagChip } from "@/components/ui/facet-chip";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
@@ -386,6 +387,18 @@ export function DataView({
     }
   };
 
+  const handleSearchQueryChange = (nextQuery: string) => {
+    const queryCleared = isSearchQueryCleared(searchQuery, nextQuery);
+    setSearchQuery(nextQuery);
+    if (!queryCleared || !currentBank) return;
+
+    setCurrentPage(1);
+    const { tags, match } = resolveTagQuery();
+    // The auto-loader intentionally waits for Enter for non-empty queries, so
+    // clearing an active query must explicitly restore the unfiltered dataset.
+    loadData(undefined, undefined, tags, match);
+  };
+
   // Single auto-loader for the graph data. This deliberately replaces what used
   // to be two effects (mount/context + filter change) that BOTH fired on mount,
   // doubling the initial /api/graph request (see issue #2158). When the context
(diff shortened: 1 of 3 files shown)


Sanderhoff-alt (commented):
Heads-up first: #3670 was fixed on main by #3697, which records the filters that produced the rows on screen (`appliedFilters` / `shouldShowEmptyBankState` in `hindsight-control-plane/src/lib/memory-filters.ts`) and gates the empty state on those. This branch shares the same base commit, so it both conflicts with that and largely duplicates it. Independently of that, the mechanism here leaves the reported dead end reachable - details inline.

hindsight-control-plane/src/components/data-view.tsx
```diff
@@ -386,6 +387,18 @@ export function DataView({
...
+  const handleSearchQueryChange = (nextQuery: string) => {
+    const queryCleared = isSearchQueryCleared(searchQuery, nextQuery);
+    setSearchQuery(nextQuery);
+    if (!queryCleared || !currentBank) return;
+
+    setCurrentPage(1);
+    const { tags, match } = resolveTagQuery();
+    // The auto-loader intentionally waits for Enter for non-empty queries, so
+    // clearing an active query must explicitly restore the unfiltered dataset.
+    loadData(undefined, undefined, tags, match);
```
Sanderhoff-alt, line 399: The empty-state decision still reads the live `searchQuery`, with `!loading` covering only the window while the restore request is in flight, so the dead end from #3670 is still reachable in two cases:

- The restore request fails. `loadData` swallows the error (the `catch` only lets the interceptor toast), so `data` keeps the zero-result payload while `loading` returns to false and `searchQuery` is now empty - the branch below replaces the whole view, search box included, with no way back except re-navigating.
- The user clears the box while the Enter-triggered search is still in flight. There is no request-sequence guard, so if the search response resolves after the unfiltered one it overwrites `data` with the zero-result payload under an empty `searchQuery`, producing the same screen.

Gating on the filters that actually produced `data` (captured when the request is issued) covers both, because a zero-result screen is then always attributed to the query that produced it.

hindsight-control-plane/src/components/data-view.tsx
```diff
@@ -386,6 +387,18 @@ export function DataView({
     }
   };
 
+  const handleSearchQueryChange = (nextQuery: string) => {
+    const queryCleared = isSearchQueryCleared(searchQuery, nextQuery);
+    setSearchQuery(nextQuery);
+    if (!queryCleared || !currentBank) return;
```
Sanderhoff-alt, line 393: `isSearchQueryCleared` only compares the input's previous and next value, so emptying a box whose query was never submitted also triggers a reload: type a few characters, change your mind, erase them, and a full `/api/graph` fetch is issued even though the data on screen is already unfiltered. Comparing against the query that was last sent avoids the redundant request.

nicoloboschi:
already fixed, closing

PR #3685, opened 2026-08-20
fix(llm): recover malformed JSON in the openai-compatible provider (#3683)

### LINKED ISSUE #3683: openai_compatible_llm: bare json.loads + identical retries — parse_llm_json/json_repair never reached, deterministic failures burn all 4 attempts and drop the chunk
## Summary

`openai_compatible_llm.py` parses the model's structured output with bare `json.loads`, and on failure retries by re-sending the **byte-identical** request. It never reaches `parse_llm_json()` — the repository's own hardened parser in `engine/llm_wrapper.py`, which strips fences, replaces raw control characters and falls back to `repair_json` from the already-installed `json_repair` dependency.

The consequence is that a *recoverable* malformed response is never repaired, and a *deterministic* malformed response costs `max_retries + 1` full generations before the chunk is dropped. With `HINDSIGHT_API_RETAIN_MAX_COMPLETION_TOKENS=16000` and a model that leaves the JSON format (and therefore does not emit a stop token either), each of those attempts runs to the ceiling.

This is the same failure class as #2504 / #2911, which were fixed for `openai-codex` in 0.8.5. That hardening was not ported to the generic OpenAI-compatible provider — the one every self-hosted deployment behind a custom `HINDSIGHT_API_LLM_BASE_URL` uses.

## Who has the safe parser and who does not (0.9.1)

`parse_llm_json` is used by:

- `providers/gemini_llm.py`
- `providers/litellm_llm.py`
- `engine/retain/fact_extraction.py`, `engine/reflect/delta_ops.py`

Providers still on bare `json.loads`:

- **`providers/openai_compatible_llm.py`**
- `providers/anthropic_llm.py`, `providers/openai_responses_llm.py`, `providers/xai_oauth_llm.py`, `providers/fireworks_llm.py`, `providers/claude_code_llm.py`
- `providers/codex_llm.py` has its own narrower `_repair_invalid_json_escapes` (from #2513) rather than `parse_llm_json`

`fact_extraction.py` *does* use `parse_llm_json`, but it never gets the chance: the provider's own retry ladder exhausts and raises first, so the repair-capable layer never sees the string.

## Where

`hindsight_api/engine/providers/openai_compatible_llm.py`, ~L1022–1051:

```python
    clean_content = _strip_code_fences(content)
    try:
        json_data = json.loads(clean_content)
    except json.JSONDecodeError:
        # Fallback to parsing raw content in case stripping was wrong
        try:
            json_data = json.loads(content)
        except json.JSONDecodeError as json_err:
            logger.warning(f"JSON parse error from LLM response (attempt {attempt + 1}/{max_retries + 1}): {json_err} ...")
            # Retry on JSON parse errors
            if attempt < max_retries:
                backoff = min(initial_backoff * (2**attempt), max_backoff)
                await asyncio.sleep(backoff)
                last_exception = json_err
                continue
            else:
                logger.error(f"JSON parse error after {max_retries + 1} attempts, giving up")
```

`continue` re-enters the loop with an unchanged `payload`. Between attempts, only the backoff differs — the request, the temperature and the prompt are the same.

## Measured

Hindsight 0.9.1 (`e5b49eb`), self-hosted, `HINDSIGHT_API_LLM_PROVIDER=openai` against a custom base URL, `HINDSIGHT_API_LLM_STRICT_SCHEMA_RETAIN=false`, `retain_extraction_mode: verbose`, importing 20 documents (244 KB) into one bank.

Over 60 minutes:

| | |
|---|---|
| JSON parse errors logged | 17 |
| of those at the **identical** position `line 4 column 295 (char 317)` | **16** |
| chunks dropped after 4 attempts (`giving up`) | 3 |
| documents fully retained in the first 20 minutes | 1 of 20 |

The identical error position across every attempt is the point: this was never flaky. Four full generations, four identical failures, then the facts are gone.

Our own trigger was a proxy that drops `response_format` and appends its own trailing instruction, so the last line the model read contradicted the JSON schema pasted above it. That was our bug and we fixed it on our side. But the retry behaviour is what turned a roughly 1-in-3 failure rate into a stalled import, and it is generic — any deterministic parse failure behaves this way.

## Suggested fix

1. **Use `parse_llm_json()` in `openai_compatible_llm.py`** instead of bare `json.loads`. It already exists, it is already imported elsewhere, and `json_repair` is already a dependency. This alone would have recovered our responses without a single retry.
2. **Make the retries differ.** On a parse failure, append the invalid output plus the `json.JSONDecodeError` message as a repair turn ("your previous reply failed at `line 4 column 295: Expecting ',' delimiter` — return only the corrected JSON document"). Re-sending an identical request cannot fix a deterministic failure.
3. **Fail fast on identical failures.** If attempt *n* fails at the same `(msg, lineno, colno)` as attempt *n−1*, three more full generations are not worth their cost.

Independently: the `giving up` path drops a chunk's facts without the retain operation surfacing a partial result. A bank can look complete while individual chunks never made it in. A per-document dropped-chunk counter in the retain response would make that visible.

## Environment

- Hindsight 0.9.1, image `ghcr.io/vectorize-io/hindsight:latest` (`e5b49eb6729512bd9b103058daa93f701da25644`)
- `HINDSIGHT_API_LLM_PROVIDER=openai`, custom `HINDSIGHT_API_LLM_BASE_URL`
- `HINDSIGHT_API_LLM_STRICT_SCHEMA_RETAIN=false` (the proxy cannot honour `response_format`; in strict mode the schema is not put in the prompt at all, so the model would never see it)
- `HINDSIGHT_API_RETAIN_MAX_COMPLETION_TOKENS=16000`, `retain_extraction_mode: verbose`
- `json_repair` 0.61.7 present in the image

chiruu12:
Fixes #3683.

`openai_compatible_llm.py` parsed structured output with bare `json.loads` and, on failure, retried by re-sending a byte-identical request. `parse_llm_json` was never reached, so a recoverable response was never repaired and a deterministic one cost `max_retries + 1` full generations before the chunk was dropped.

## What changed

Both parse sites in the file now fall back to `parse_llm_json` once a fresh generation can no longer help: the `chat.completions` path and the native Ollama `/api/chat` path. That mirrors what `litellm_llm.py` has done since #2547/#2544, including its preference for a clean re-roll before repair.

The retry ladder also stops early when the same response content comes back twice. Every attempt re-sends a byte-identical request, so identical content means more generations cannot help. Two attempts establish that, and the rest of the budget goes to repair. Different content still earns a fresh generation, including two malformed outputs that fail at the same line and column, which is why the check is on content and not on the decode position.

A token-capped response is reported rather than repaired. `chat.completions.create` returns truncated content with `finish_reason: "length"` instead of raising, so without a guard it reached `json_repair`, the open braces were closed, and a shorter list validated and returned as a success. Both paths now raise `OutputTooLongError` first, keyed on `finish_reason` and on Ollama's `done_reason`. That is the recovery path rather than a failure: `fact_extraction` catches it and splits the chunk, so the whole input is still processed. The guard runs before the retry ladder because a cap is deterministic, matching `litellm_llm.py:343`.

## Tests

Added to `tests/test_openai_compatible_response_hardening.py`. All mocked, no network:

- a repairable response is recovered instead of dropped
- an identical repeat stops re-rolling after 2 attempts instead of 4
- a failure at a new position still re-rolls
- the same recovery on the native Ollama path
- a token-capped response raises instead of returning a shortened result
- the same on the native path, keyed on `done_reason`
- the cap is reported on the first response that shows one, not after the budget
- malformed output that was not truncated still gets repaired

Three of those fail on current `main`. The fourth pins the re-roll behaviour this change keeps.

`ruff check`, `ruff format --check` and `ty check` are clean. 295 passed and 5 skipped across every test file that touches the provider.

## Not here

The issue also suggests feeding the invalid output back as a repair turn, and a per-document dropped-chunk counter on the retain response. Both change behaviour past parsing, so I left them on the issue.


r266-tech (commented):
At head SHA `65161aa3e21fc5f3dba3d95b1bc8d3a67181a86a`, the retry short-circuit treats any two failures with the same `JSONDecodeError` message, line, and column as a deterministic repeat without checking that the response text is unchanged. Two different stochastic outputs can therefore be classified as a deterministic repeat and trigger `parse_llm_json` on the second attempt; if repair cannot recover that output, the exception exits the retry loop and skips the remaining generation budget. A later valid response is therefore dropped instead of being attempted.

Please only short-circuit when the failed response content is also identical, or otherwise retain a retry when content differs, and add a regression test with two distinct responses sharing the same `JSONDecodeError` signature followed by a valid response.

The bound verification observed head SHA `65161aa3e21fc5f3dba3d95b1bc8d3a67181a86a`, and `git diff --check` passed. The focused pytest command was not independently run because uv was denied permission to create its cache directory in the read-only environment.

chiruu12:
Right, that holds. Two different malformed responses can share a JSONDecodeError message, line and column, so the check called them a repeat and spent the rest of the retry budget on repair instead of on the generation that would have worked.

The tracker now holds the response content, and the short-circuit only fires when the text comes back unchanged. Identical content produces an identical decode error, so the position fields carried nothing the content did not already carry. The dataclass is gone with them.

Added `test_two_responses_failing_alike_are_not_treated_as_one`: `{"aa": 1,}` then `{"bb": 2,}`, both failing at line 1 column 10, then a valid response. Fails on 65161aa, passes now. 369 passed and 5 skipped across the 26 test files that touch the provider.

koriyoshi2041 (approved):
Verified the update at `2927e8e164aa6c33fcf06aefa1e80d4ffc0a1fc9`. The repeat check now keys on response content in both compatible and native Ollama paths, so distinct malformed outputs with the same decode position retain the remaining generation attempt. The focused response-hardening suite passes 13/13; Ruff check/format and the exact update diff check are clean.

koriyoshi2041:
Verified the update at `2927e8e164aa6c33fcf06aefa1e80d4ffc0a1fc9`. The repeat check now keys on response content in both compatible and native Ollama paths, so distinct malformed outputs with the same decode position retain the remaining generation attempt. The focused response-hardening suite passes 13/13; Ruff check/format and the exact update diff check are clean.

hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
```diff
@@ -1051,14 +1053,32 @@ async def call(
...
-                            else:
-                                logger.error(f"JSON parse error after {max_retries + 1} attempts, giving up")
+                            # Retry budget spent, or the same content twice on a
+                            # byte-identical request: another generation cannot
+                            # help. Repair structurally as a last resort, the way
+                            # litellm_llm.py has since #2547/#2544. json_repair is
+                            # already a dependency and raises again when the
+                            # content is genuinely unrecoverable.
+                            try:
+                                json_data = parse_llm_json(content)
```
Sanderhoff-alt, line 1148: A token-capped response reaches this fallback. `chat.completions.create` does not raise `LengthFinishReasonError` (only the SDK's parse helper does); it returns truncated content with `finish_reason: "length"`, which fails `json.loads` and lands here. `json_repair` then closes the open braces, `model_validate` accepts the shortened list, and the call returns success with part of the extracted facts silently missing. The `repeated` check makes it likely: a re-roll against the same cap truncates identically, so the repair is reached on the second attempt instead of after the full budget.

`litellm_llm.py:343` avoids exactly this by raising `OutputTooLongError` on `finish_reason == "length"` before parsing, which is what lets the caller split the input and retain the rest. `first_choice` is already in scope here (`_finish_reason_for_choice(first_choice)` is used in the warning above), so the same guard can be applied before falling back to `parse_llm_json`.

hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
```diff
@@ -1655,13 +1676,17 @@ async def _call_ollama_native(
...
                                     backoff = min(initial_backoff * (2**attempt), max_backoff)
                                     await asyncio.sleep(backoff)
                                     last_exception = json_err
                                     continue
-                                else:
-                                    raise
+                                # Same last-resort repair as the OpenAI-compatible
+                                # path above; this parses identically and had the
+                                # identical gap.
+                                json_data = parse_llm_json(content)
```
Sanderhoff-alt: Same case on the native path: Ollama reports a cap-truncated body as `done_reason: "length"` (already read from `result` further down for the trace span), so repairing here also returns a partial object as a success. The guard belongs here too.

Secondary: unlike the path above, this call is not wrapped, so an unrecoverable body raises out of the loop with no equivalent "giving up" log line.

Sanderhoff-alt (commented):
The repair fallback itself looks right, but both parse sites can now turn a token-truncated response into a successful partial result, and the new identical-content early stop makes that the common case rather than a corner case. One note on the description as well: it says the ladder stops early "when a parse failure repeats at the same `(msg, lineno, colno)`", while the code compares the whole content (and a test asserts the position check would be wrong) - worth aligning the text with the implementation.

chiruu12:
> The repair fallback itself looks right, but both parse sites can now turn a token-truncated response into a successful partial result, and the new identical-content early stop makes that the common case rather than a corner case. One note on the description as well: it says the ladder stops early "when a parse failure repeats at the same `(msg, lineno, colno)`", while the code compares the whole content (and a test asserts the position check would be wrong) - worth aligning the text with the implementation.

Thanks for the review! will work on it.

chiruu12:
Fixed. Please check the approach I took: both paths now raise `OutputTooLongError` on a token cap, before the retry ladder rather than before the repair, since a re-roll truncates identically and the caller can only split once it hears about the cap.

Description updated too.

r266-tech (commented):
Re-reviewed exact head `128be273507a966a5a35e978708102fc80807330`.

The malformed-truncation case is improved, but one blocking gap remains: both `finish_reason == "length"` / `done_reason == "length"` checks are inside the `except json.JSONDecodeError` path. A token-capped response can still be syntactically valid JSON—for example a shortened but closed `{"facts": [...]}` object—and then `json.loads` plus schema validation return it as a complete success without ever consulting the truncation sentinel. Please check the sentinel before accepting/parsing the body in both provider paths, and add valid-JSON-with-`length` regressions for each path.

There is also still a retry-contract concern in the new early stop. Two identical stochastic generations do not establish that later attempts will be identical; on the second repeat the code repairs immediately and discards the remaining configured retry budget. Please either exhaust the configured budget before repair, or gate the shortcut on an actual deterministic-generation contract, with a regression where two identical malformed responses are followed by a valid response.

chiruu12:
Both fixed.

The sentinel is checked before the parse now, on both paths. A cap that lands on a closing brace decodes and validates fine, so the decode handler never saw it. One regression per path with valid JSON and `length`.

I dropped the repeat shortcut instead of gating it. Temperature is set by the caller and never read here, so there is no determinism signal to gate on. Full budget, then repair. Test added for two identical malformed responses followed by a valid one.

r266-tech (commented):
Re-reviewed exact head `bf6807a1115cb9c4da073ff21388bde3e7c7df1f`. The two requested blockers are resolved: both provider paths now inspect the token-cap sentinel before parsing, and the unsupported “two equal outputs imply determinism” shortcut is gone.

One material safety gap remains in the new repair fallback. After the retry budget, both paths still pass any `JSONDecodeError` to `parse_llm_json`, whose documented structural repair can synthesize missing string/collection terminators. If an OpenAI-compatible proxy omits an accurate `finish_reason`, or native Ollama omits `done_reason`, a genuinely truncated body such as `{"facts": ["alpha", "beta", "gam` can therefore be repaired into schema-valid partial data. That changes this path from fail-loud to silent fact loss even though there is no positive evidence that generation completed.

Please make structural repair conditional on a positive non-truncated completion signal (or otherwise restrict it to repairs that cannot complete missing string/collection content), and add missing/unknown-sentinel truncated-list regressions for both paths.

There is also an accounting gap on the native Ollama path: the new `done_reason == "length"` raise happens before `prompt_eval_count` / `eval_count` extraction and metrics recording, so the most expensive capped responses disappear from usage accounting. Please record that usage before re-raising, with a focused assertion.

chiruu12:
Both fixed.

Repair is gated on a completion signal now, `stop`, `tool_calls`, `function_call` or `end_turn`. Anything else, missing or unrecognised included, re-raises the decode error instead of repairing. You were right that `parse_llm_json('{"facts": ["alpha", "beta", "gam')` returns `{'facts': ['alpha', 'beta', 'gam']}`, so without the gate a proxy that drops finish_reason turned a cap into a short answer. Regressions on both paths for missing and unknown sentinels.

Native path stashes usage right after `response.json()`, before the guards, with an assertion on the counts surviving the raise.

r266-tech (commented):
Re-reviewed exact head `e51cae0322b8532bc592f1c5a720fd09cd12883f`.

The two requested blockers are resolved. Structural repair is now limited to responses with a positive completion signal on both transports, so missing or unknown sentinels keep malformed/truncation-ambiguous bodies fail-loud. The native path also stashes usage before any cap/parse guard, preserving token accounting when `OutputTooLongError` is raised. The added missing/unknown-sentinel regressions and the capped-usage assertion cover those boundaries.

I also checked the remaining parse-valid ambiguity: a response with missing completion metadata that is already valid JSON is still accepted, while an explicitly `length`-capped valid body is rejected. That is the pre-existing compatibility boundary rather than a regression introduced by this repair fallback; requiring a closed allowlist before accepting every valid structured response would reject otherwise compatible providers. No further blocker from this review.

koriyoshi2041 (approved):
Reviewed exact head `e51cae0322b8532bc592f1c5a720fd09cd12883f`.

The changed completion boundary now looks sound across both transports: explicit token caps are rejected before parsing, structural repair requires a positive completed-generation signal, missing or unknown sentinels keep malformed bodies fail-loud, and the native path records usage before raising. This prevents repair from converting an ambiguous truncation into schema-valid partial memory while preserving compatibility for already-valid responses whose provider omits completion metadata.

Locally, the focused response-hardening file passes 28/28 with proxy environment variables cleared; Ruff check, Ruff format-check, and diff hygiene also pass. No further blocker from this review.

nicoloboschi:
Thank you for this — the diagnosis and the fix were both right, and the tests are what made the two quiet cases legible. Carrying it forward in #3935 with you as co-author on the commit.

Closing here only because `main` moved twice underneath this branch and the rebase was substantial enough to be a new PR rather than a force-push onto yours:

- #3827 landed the same truncation fix for the OpenAI-compatible path, which makes the `finish_reason == "length"` guard in `call()` unreachable (`_content_or_error` raises first) and duplicates three of the truncation tests.
- #3920 rewrote 119 lines of the same file.

Two things carried over unchanged because they were the substance and #3827 did not touch them: the `parse_llm_json` fallback on both parse sites, and the `done_reason` guard on the native Ollama path. I moved the latter above the free-form/structured split — a cap reached before the first visible token was still landing on the retryable `ProviderResponseError` and re-sending against the same limit, which is #3811's shape surviving on that path — and added two tests for it.

The `_COMPLETED_FINISH_REASONS` gate is kept as you wrote it, with a note at the definition recording why it is stricter than `litellm_llm.py:367`, since that asymmetry is invisible otherwise.

Appreciated — this was a real bug and a careful fix.

PR #3692, opened 2026-08-21
fix(knowledge-base): include global observations in page scope

### LINKED ISSUE #3687: Knowledge Pages: content generation returns empty on existing banks (graph retriever finds no seeds despite direct recall scoring 1.0+)
# Knowledge Pages: content generation always returns "no information" on an existing bank (graph retriever returns empty)

## Summary

Creating a knowledge page (0.9.1) on a bank that has substantial pre-existing memory always generates content stating "I don't have information about X" — even though direct recall on the same bank returns dozens of highly-relevant results for the same query.

## Environment

- Hindsight 0.9.1 (hindsight-api pip package, self-hosted, local_external mode)
- PostgreSQL backend (pg0 embedded)
- LLM provider: OpenAI-compatible endpoint
  - Main LLM: `xiaomi/mimo-v2.5` via opencode-go endpoint (works fine for retain/recall)
  - Reflect LLM: `zai-org/glm-5.2` via commandcode endpoint
- Bank: existing bank with 125 documents, 3,255 nodes, ~1,136 observations, pre-dating the 0.9.0 upgrade

## Steps to Reproduce

1. Have a bank with substantial pre-existing memory (data retained before the 0.9.0 upgrade).
2. Create a knowledge page:
   ```
   POST /v1/default/banks/{bank_id}/knowledge-base/pages
   {
     "name": "Homelab Infrastructure",
     "source_query": "Eddie homelab NAS ThinkPad docker containers jellyfin",
     "tags": ["type:runbook", "homelab", "infrastructure"]
   }
   ```
3. Wait for the async refresh (`REFRESH_MENTAL_MODEL_TASK`) to complete.
4. Fetch the page:
   ```
   GET /v1/default/banks/{bank_id}/knowledge-base/pages/{page_id}
   ```

## Expected Behavior

The page body should contain a generated runbook/wiki document synthesized from the bank's relevant memories.

## Actual Behavior

The page body contains:

> "I don't have information about Eddie's homelab infrastructure. All searches for details regarding NAS, ThinkPad hardware, Docker containers, Jellyfin, or related homelab/server setup returned no results. No observations or source facts were retrieved on these topics."

Meanwhile, a direct recall on the same bank with the same query returns 71–101 results, including a top hit scoring **1.087**:

```
POST /v1/default/banks/eddie/memories/recall
{"query":"Eddie homelab NAS ThinkPad docker containers jellyfin","types":["world","experience"]}
→ 81 results; top: "Eddie has a homelab with a ThinkPad server running 27 Docker containers, including Jellyfin, SABnzbd, Prowlarr..." (score 1.087)
```

## Diagnostic Evidence

Daemon log during page generation:

```
[REFLECT eddie-61499] Starting agentic reflect for query: Eddie homelab NAS Thi...
hindsight_api.engine.search.retrieval - Using LinkExpansion graph retriever
[REFLECT eddie-61555] forced | query='Eddie homelab NAS ThinkPad docker contain...
[REFLECT eddie-61499] Complete: 293 chars, 5 iterations, 10 tool calls | 17.761s
```

Key observations:

1. The reflect agent's internal recall uses the **LinkExpansion graph retriever** (`hindsight_api/engine/search/link_expansion_retrieval.py`).
2. The graph retriever's `_find_semantic_seeds()` applies `graph_seed_min_similarity` (default 0.3). If no seeds pass, it returns `([], timings)` — the agent then concludes there is no information.
3. The direct recall API path (used by the memory plugin) finds the same memories with top score 1.087 — well above 0.3 — so the seeds *should* exist. The two paths appear to score/retrieve differently (or the graph path fails to seed on banks whose data predates the graph/0.9 schema).
4. Also observed during earlier page generation (before reflect LLM was pinned separately):
   `Tool expand returned invalid memory ID formats` — the reflect agent's `tool_expand` received IDs that failed `uuid.UUID()` validation. This may be a related/secondary issue (LLM passing malformed IDs), but the primary issue here is the graph retriever returning zero results even when direct recall finds high-scoring matches.

## Hypothesis

The LinkExpansion graph retriever's seed-finding (semantic similarity threshold, or its interaction with pre-0.9 data / graph structure) returns no seeds for banks with substantial pre-upgrade memory, causing the reflect agent to conclude the bank has no relevant information. Direct (non-graph) recall on the same data works correctly.

## Impact

Knowledge Pages (a headline 0.9.0 feature) generate empty/useless content on any existing bank with pre-0.9 memory. Fresh banks (empty at upgrade) may not hit this because their data is written post-upgrade and indexes/structure match what the graph retriever expects.

## Suggested Areas to Investigate

- `_find_semantic_seeds()` in `engine/search/link_expansion_retrieval.py` — threshold vs. actual embedding similarity distribution on pre-existing data
- Whether pre-0.9 documents/chunks are missing graph links or an index that the graph retriever requires (data migration gap)
- The `Tool expand returned invalid memory ID formats` path — whether the reflect agent is fed non-UUID IDs from the graph retriever's seed results

## Environment Detail

- hindsight-api 0.9.1, installed Aug 17 2026 (upgrade from 0.8.x)
- Bank existed since July 2026 with data retained under 0.8.x
- Self-hosted: `hindsight-api --port 8888`, systemd user service

oldnicke:
Fixes  #3687
## Summary

Knowledge pages with tags inherited the generic `all_strict` refresh scope. Existing banks commonly contain untagged observations, so the graph seed query filtered every observation before synthesis and produced an empty page.

Knowledge-page defaults now use `tags_match: "all"`: tagged memories still need every page tag, while untagged observations remain available as bank-wide knowledge. The regression coverage checks page defaults, partial-trigger behavior, page staleness, and tagged-scope filtering. API documentation and its generated reference are updated.

## Verification

- `tests/test_knowledge_base.py`: 60 passed
- `tests/test_mental_model_consolidation_refresh_scope.py`: 6 passed
- `./scripts/hooks/lint.sh`: passed
- Documentation build: passed (with two pre-existing broken-anchor warnings)

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -15249,6 +15249,9 @@ class MemoryEngine(MemoryEngineInterface):
         "fact_types": ["observation"],
         "exclude_mental_models": True,
         "refresh_after_consolidation": True,
+        # Preserve AND filtering for tagged memories while keeping global (untagged)
+        # observations visible to pages.
+        "tags_match": "all",
     }
 
     def _merge_page_trigger(self, trigger: dict[str, Any] | None, base: dict[str, Any] | None = None) -> dict[str, Any]:
--- a/hindsight-docs/docs/developer/api/knowledge-pages.mdx
+++ b/hindsight-docs/docs/developer/api/knowledge-pages.mdx
@@ -145,7 +145,8 @@ When `trigger` is omitted, the page is created with a document-oriented configur
   "mode": "delta",
   "fact_types": ["observation"],
   "exclude_mental_models": true,
-  "refresh_after_consolidation": true
+  "refresh_after_consolidation": true,
+  "tags_match": "all"
 }
 ```
 
@@ -157,6 +158,7 @@ This makes the page a living document built from consolidated observations only,
 | `exclude_mental_models: true` | A page never reflects on sibling pages. Without this, pages would cite each other and drift into a feedback loop where one wrong claim propagates across the knowledge base. |
 | `mode: "delta"` | Each refresh edits the existing document with what is new since the last refresh instead of regenerating it, so hand-tuned structure and wording survive. See [Refresh Mode](./mental-models#refresh-mode). |
 | `refresh_after_consolidation: true` | The page rewrites itself whenever consolidation produces new knowledge in its scope — gated by the same [staleness check](./mental-models#staleness-gating) as any mental model, so unrelated bank activity doesn't trigger rebuilds. |
+| `tags_match: "all"` | Page tags remain an AND filter for tagged memories, while untagged observations remain visible as bank-wide knowledge. |
 
 ### Page Lifecycle
 
@@ -169,7 +171,10 @@ This makes the page a living document built from consolidated observations only,
 Observations are what a page is *built from*, but it can still inspect the evidence underneath them: the refresh agent can expand a memory to its original chunk or document (unless `store_document_text` is disabled for the bank), and if `HINDSIGHT_API_REFLECT_SOURCE_FACTS_MAX_TOKENS` is enabled — off by default — observation search also returns each observation's grounding facts.
 
 :::caution
-A supplied `trigger` **replaces** these defaults, it does not merge with them. Sending `{"trigger": {"mode": "full"}}` also resets `fact_types` to all types, `exclude_mental_models` to `false`, and `refresh_after_consolidation` to `false`. Repeat the fields you want to keep.
+A supplied `trigger` patches these defaults: fields you omit keep their page defaults. Sending
+`{"trigger": {"mode": "full"}}` changes only the refresh mode; it keeps observation-only
+retrieval, page isolation, automatic refresh, and the non-strict tag matching unless you override
+those fields explicitly.
 :::
 
 Every [mental model trigger setting](./mental-models#trigger-settings) is accepted here — including `refresh_cron` for scheduled rebuilds instead of consolidation-driven ones, and `tag_groups` for compound tag scoping.
--- a/hindsight-api-slim/tests/test_knowledge_base.py
+++ b/hindsight-api-slim/tests/test_knowledge_base.py
@@ -266,6 +266,7 @@ class TestTree:
         trigger = roots["Loose"]["trigger"]
         assert trigger["mode"] == "delta"
         assert trigger["fact_types"] == ["observation"]
+        assert trigger["tags_match"] == "all"
         assert trigger["exclude_mental_models"] is True
         assert trigger["refresh_after_consolidation"] is True
         # A folder has no backing mental model, so it has no refresh policy either —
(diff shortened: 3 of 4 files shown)


hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -15249,6 +15249,9 @@ def _build_mm_scope_filter(
         "fact_types": ["observation"],
         "exclude_mental_models": True,
         "refresh_after_consolidation": True,
+        # Preserve AND filtering for tagged memories while keeping global (untagged)
+        # observations visible to pages.
+        "tags_match": "all",
```
Sanderhoff-alt, line 15254: Flipping the page default to non-strict conflicts with text merged onto main after this branch's base (#3755), which this PR does not touch:

- `hindsight-api-slim/hindsight_api/api/http.py`, `CreatePageRequest.tags`: "By default a tagged page matches with `all_strict`: a memory must carry EVERY one of these tags, and untagged memories are excluded entirely ... set `trigger.tags_match` to 'all' to require the tags while still including untagged memories."
- `hindsight-docs/docs/developer/api/knowledge-pages.mdx`: the `tags` section and the options table repeat the same rule, including a row that presents `all_strict` as the strict per-user isolation choice.

After a rebase both statements are false, and the `http.py` one is a field description that ships in the OpenAPI spec and every generated SDK. Please rebase and update those descriptions alongside the default, then regenerate the spec/clients and the docs skill.

hindsight-docs/docs/developer/api/knowledge-pages.mdx
```diff
@@ -157,6 +158,7 @@ This makes the page a living document built from consolidated observations only,
 | `exclude_mental_models: true` | A page never reflects on sibling pages. Without this, pages would cite each other and drift into a feedback loop where one wrong claim propagates across the knowledge base. |
 | `mode: "delta"` | Each refresh edits the existing document with what is new since the last refresh instead of regenerating it, so hand-tuned structure and wording survive. See [Refresh Mode](./mental-models#refresh-mode). |
 | `refresh_after_consolidation: true` | The page rewrites itself whenever consolidation produces new knowledge in its scope — gated by the same [staleness check](./mental-models#staleness-gating) as any mental model, so unrelated bank activity doesn't trigger rebuilds. |
+| `tags_match: "all"` | Page tags remain an AND filter for tagged memories, while untagged observations remain visible as bank-wide knowledge. |
```
Sanderhoff-alt, line 161: This describes the default for pages created from now on, but the change does not reach pages that already exist: their merged trigger was persisted without `tags_match`, and `_resolve_refresh_tag_filtering` (`hindsight-api-slim/hindsight_api/engine/memory_engine.py:1457`) still resolves `all_strict` for a tagged model whose trigger omits it. The bank reported in #3687 is exactly that case, so it keeps producing empty pages until each page is patched with `{"trigger": {"tags_match": "all"}}`. Either say so here or backfill the stored page triggers.

Sanderhoff-alt (commented):
The branch is based on a commit from Aug 21; #3687 was closed on main by #3755, which took the opposite approach (keep the all_strict default, document `trigger.tags_match: "all"` as the override). A rebase is needed, and after it the new default contradicts what main documents in several places. Two points below.

nicoloboschi:
we can't change the default although I agree it would have been a better default.

PR #3693, opened 2026-08-21
fix: fingerprint LLM pool members by provider+model+api_key for batch…

faisalhussain-devs:
## Summary

Fixes the same-provider case flagged as remaining after #3649's cross-provider guard: crash recovery could resume a batch on the wrong account when two LLM pool members share the same `provider` string.

## Root cause

`fact_extraction.py`'s resume path selected `batch_impl` via `batch_provider_impl()`, which returns the **first batch-capable member unconditionally**, and validated only on `batch_provider` (e.g. `"openai" == "openai"`). Two members configured on the same provider were indistinguishable to this check, so a member reorder or insertion between batch submission and crash recovery could resume the batch using a different account's credentials than the one that submitted it.

## Reproduction

1. Configure two batch-capable members of the same provider with different API keys.
2. Submit a batch through account B; persisted metadata records only `batch_provider: "openai"`.
3. Restart with account A ordered before account B.
4. Resume the operation — account A's client polls/retrieves a batch ID that belongs to account B.

## What changed

- **`_generate_member_id(provider, model, api_key)`** (`config.py`) — an HMAC-SHA256 fingerprint, computed once per configured member at config-parse time. No provider subclass needed changes; the identity is stamped onto the `LLMInterface` instance post-construction.
- **`batch_provider_impl(selector=...)`** (`multi_llm.py`) — resolves the exact member by `member_id` when a selector is provided; falls back to first-capable-member (prior behavior) when `selector=None`, preserving backward compatibility.
- **`fact_extraction.py`** — persists `batch_member` in operation metadata at submission time, re-resolves against it on resume, and fails loudly with `RuntimeError` if the submitting member can't be found — rather than silently proceeding on a best-guess default.
- **Backward compatible** — pre-existing batch records with no `batch_member` key fall back to `selector=None`, i.e. the original first-match behavior. No migration needed.
- **Tests** (`test_multi_llm_batch.py`) — covers both the selector-resolution logic (reordered pool, hand-assigned member IDs) and the fingerprint function itself (two same-provider configs differing only by `api_key` produce distinct, deterministic IDs), plus an end-to-end variant using real fingerprints through the full reorder-and-resume path.

## Known limitations

I didn't try to resolve these unilaterally since they involve judgment calls beyond this specific bug fix — flagging both for your take:

1. **OAuth-based providers** (codex, claude-code, github-copilot) have no `api_key`, so `member_id` still collapses to `provider:model` for these — multiple accounts on the same OAuth-based provider remain indistinguishable. Would need a different discriminator (token/session identifier) for that provider family.
2. **The fingerprint is credential-derived.** It's not reversible, but a persisted digest isn't fully opaque against someone who already holds a candidate key for that provider — it functions as a verification oracle for that narrow case, not a way to recover the key itself.

A cleaner long-term fix for both would be an explicit, user-configured per-member account identifier instead of deriving identity from credentials — but that's a config-schema change spanning the indexed LLM chains (retain/reflect/consolidation), so I left it as your call rather than adding a new config field myself. Happy to implement that instead if you'd rather go that route.

## Testing
11 passed. Also ran `ruff check --fix`, `ruff format`, and `ty check` on all changed files — clean (pre-existing `ty` diagnostics elsewhere in the tree are unrelated platform-specific stdlib/optional-dependency noise, not touched by this change).


nicoloboschi:
Thanks for this — the root-cause analysis and the reproduction steps are both accurate, and the same-provider gap you identified was a real bug.

It's since been fixed on `main` by 6ff6dc692 ("fix(llm): bind batch recovery to the account that submitted the batch", #3699), which landed while this PR was open, so I'm closing this as superseded rather than because anything was wrong with the diagnosis.

The merged fix takes the same shape — a per-member identity persisted at submit time and re-resolved on resume — with a few differences worth noting:

- The identity is a `batch_account_key` property on `LLMInterface` (`provider|base_url|sha256(api_key)[:12]`), so every impl has one by construction and there's no `getattr` fallback to reason about.
- `model` is deliberately excluded: a batch belongs to an account, not to a model, so retargeting a member's model still resumes.
- `base_url` is included, so two members on the same provider/model/credential but different endpoints stay distinguishable.
- The `account_key` parameter is threaded through both `LLMProvider.batch_provider_impl` and `MultiLLMProvider.batch_provider_impl`, so the single-provider path is covered too.
- Records written before `batch_account` existed keep the coarse `batch_provider` guard.

One thing your "known limitations" section called out is still open: OAuth-based providers (codex, claude-code, github-copilot) have no `api_key`, so `batch_account_key` collapses to `provider|base_url|sha256("")[:12]` and multiple accounts on the same OAuth provider remain indistinguishable. If you'd like to pick that up as a follow-up, it's the piece with real value left in it — happy to look at a PR for it.

faisalhussain-devs:
Thanks for pointing this limitation out and for offering to look at a follow-up PR. I dug into the current implementations of Codex, Claude Code, and GitHub Copilot to see whether the issue is actionable.

Codex does expose a stable account_id, but it currently does not support the batch API in Hindsight, so the batch_account_key limitation does not affect its batch recovery path today. Claude Code and GitHub Copilot similarly do not implement batch support, so there is no active batch-recovery case to fix for them at the moment.

Given that, I don't think there's a meaningful follow-up PR to make right now. It seems more appropriate to revisit this if those providers gain batch support in the future.

Thanks again for the detailed explanation and for pointing out the remaining limitation. I also went through your #3699 implementation in detail, and it helped me understand the architectural considerations behind the fix, especially making account identity part of the LLMInterface and keeping the recovery logic account-based rather than member-based. I learned a lot from comparing the two approaches.

PR #3694, opened 2026-08-21
fix(openclaw): probe append support for local daemon

### LINKED ISSUE #3686: Local-daemon mode: per-turn document ids reset on host restart → silent same-id overwrites (data loss)
### Summary
In **local-daemon mode** (the default when the OpenClaw plugin spawns `hindsight-embed` itself), the append-capability probe never runs, so retains fall back to **per-turn document ids whose in-memory sequence resets on every host restart** — silently **overwriting the prior cycle's documents** server-side. Long-lived sessions lose nearly all history.

### Observed impact
On a production deployment (months of daily direct-message traffic on one session), only **29 documents** existed for that session — each restart's `…:turn:000001, 000002, …` ids replaced the previous cycle's docs. Bank-wide classification of 16.6K docs confirmed the per-turn id shape only on local-daemon-era writes.

### Mechanism (from the shipped plugin dist, @vectorize-io/hindsight-openclaw)
- `detectAppendCapability` is only invoked on the **external-API** code paths (dist ~:1466, :1551, :1594) — it is absent from the local-daemon branch (~:1490–1530)
- `supportsUpdateModeAppend` therefore stays `false` → retain uses per-turn document ids
- The sequence source `documentSequenceBySession` (~:170, :2222–2227) is **in-memory only** → resets to 1 on every host process restart
- The dist's own comment (~:2242–2244) notes same-id retains overwrite server-side

### Repro sketch
1. Run the OpenClaw plugin in local-daemon mode (no `hindsightApiUrl` set)
2. Retain a few turns on one session; restart the host process; retain more turns
3. List the bank's documents for that session: the second cycle's `turn:000001…` ids replaced the first cycle's

### Workaround we're using
Set `hindsightApiUrl` to the already-running daemon (external-API mode) with the daemon supervised separately — the probe runs, append mode activates, doc ids become session-stable, and the retain queue engages.

### Suggested fix
Run the append-capability probe in the local-daemon branch too (or persist the per-session sequence), so restarts can't recycle document ids.

**Env:** hindsight-embed / hindsight-api 0.9.x local daemon, plugin @vectorize-io/hindsight-openclaw (OpenClaw 2026.7.2-beta.7 host), macOS, node v26.5.0.

oldnicke:
Fixes #3686
## Summary

- probe the embedded daemon capability endpoint after startup so supported deployments use append retention
- repeat the probe on local-daemon recovery initialization
- cover initial startup and recovery with lifecycle regression tests that assert stable session document IDs

## Testing

- `npm test` (322 tests)
- `npm run build`
- `./scripts/hooks/lint.sh`

### DIFF
--- a/hindsight-integrations/openclaw/src/index.ts
+++ b/hindsight-integrations/openclaw/src/index.ts
@@ -92,6 +92,12 @@ export type AsyncRetainOperationIdCapability = "supported" | "unsupported" | "un
 let asyncRetainOperationIdCapability: AsyncRetainOperationIdCapability = "unknown";
 const MIN_VERSION_FOR_ASYNC_RETAIN_OPERATION_ID = "0.8.6";
 
+// The /version capability is deployment-wide, but store_document_text can be
+// overridden per bank. Cache the resolved value after the first retain for a
+// bank so dynamic bank routing does not add a config request to every turn.
+const bankAppendCapability = new Map<string, boolean>();
+const bankAppendCapabilityProbes = new Map<string, Promise<boolean | null>>();
+
 // Store the current plugin config for bank ID derivation
 let currentPluginConfig: PluginConfig | null = null;
 let serviceGeneration = 0;
@@ -127,6 +133,8 @@ export interface BankScopedClient {
     timeoutMs?: number
   ): Promise<RecallResponse>;
   setMissions(opts: BankMissionsUpdate): Promise<void>;
+  /** Return the bank's resolved config, or null when the config API is unavailable. */
+  getConfig(signal?: globalThis.AbortSignal): Promise<Record<string, unknown> | null>;
 }
 
 export interface BankMissionsUpdate {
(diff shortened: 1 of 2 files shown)


r266-tech (commented):
Please make append selection bank-aware or handle an append rejection by retrying that retain without append while preserving the existing retention path. The local-daemon capability probe uses the global `/version` value for `store_document_text` and then applies `update_mode=append` to every bank. Banks can override `store_document_text=false`; those banks reject append even when the global probe reports support. In local mode the retain error is logged and dropped because there is no retain queue, so conversation turns for such a bank can be lost rather than retained. Add a regression test for a local bank whose `store_document_text` override is false and verify that its retain is not dropped.

The inspected pull-request head is `ad54e85f1e04edee812296598beddcb9e8d5f197`; head binding passed and `git diff --check` exited 0. The requested focused test was not authority-approved, and the contributor's npm test, build, and lint claims are not independently evidenced.

oldnicke:
@r266-tech Thanks for catching this. I updated the OpenClaw retain path to resolve `store_document_text` per bank instead of relying only on the deployment-wide `/version` capability. Banks with document text disabled now use unique per-turn document IDs, and an unexpected append rejection also retries through that path so local retains are not dropped.

I added regression coverage for a local bank override and for append-rejection fallback. The updated single commit is `f42d3143a05d59526a0b30d2ba9b7d21c2480e43`. Verified with 324 OpenClaw tests, TypeScript build, lint, generated-file checks, and CI.

nicoloboschi:
fixed in https://github.com/vectorize-io/hindsight/pull/3706

PR #3713, opened 2026-08-22
fix(retain): retry facts missing required text

### LINKED ISSUE #3708: Retain completes with zero facts when extraction output fails schema validation
**Bug Description**
When the extraction model returns facts that fail schema validation (e.g. a fact object missing the `what` field), the affected facts are skipped with only a server-side log line (`skipping malformed fact: missing 'what' field`). The retain operation still completes with status `completed`, produces zero memory units for the document, and nothing in the operation result reports the loss. Verify-style tooling and callers see a successful ingestion.

At scale this loses real data silently. Measured on the same ~1,900-document corpus (technical notes, 300–5,000 chars each) via the ollama provider against Ollama Cloud (the ':cloud' model tags; no local GPU involved, so model size is not a factor in reproducing this):
- `deepseek-v4-pro`: 0 of 800 documents produced zero facts
- `glm-5.2` (JSON-mode only, no schema enforcement): 66 of 500 (13%)
- `deepseek-v4-flash`: 534 of 628 (85%)

The affected documents were substantive multi-paragraph notes, not empty content: re-running them through `/documents/{id}/reprocess` under a schema-compliant model extracted 4–16 facts each.

**Steps to Reproduce**
1. Configure an extraction model that frequently emits schema-nonconforming output (e.g. `glm-5.2` via Ollama, which supports JSON mode but not strict JSON schema).
2. Retain a document with several extractable facts.
3. Observe server logs show `skipping malformed fact: missing 'what' field` per dropped fact.
4. Poll the operation: status is `completed`. List memory units for the document: zero.

**Expected Behavior**
One or more of:
- Chunks whose extraction output fails validation are retried (the parse-error path already retries; the schema-validation path does not).
- The operation result carries the failure: a `failed`/`partial` status, or a per-chunk `skipped_facts`/`empty_chunks` counter, so callers can detect and re-drive.
- A completed retain whose document produced zero facts from non-empty content is distinguishable from "content genuinely contained nothing".

**Actual Behavior**
Operation `completed`, zero facts, loss visible only in server logs.

**Version**: ghcr.io/vectorize-io/hindsight:0.9.1
**LLM Provider**: ollama, models served on Ollama Cloud (glm-5.2, deepseek-v4-flash, deepseek-v4-pro compared)

koriyoshi2041:
## Problem

Facts without a usable `what`, `factual_core`, or `text` field are logged and skipped, but that branch does not mark the response as malformed. As a result, it bypasses the existing content retry and can let a non-empty retain finish with zero facts.

Fixes #3708.

## Fix

- Mark missing-text facts as malformed so the existing content retry budget is used.
- If every fact is still malformed after the retry budget is exhausted, raise instead of returning an empty successful result.
- Preserve the existing behavior for genuinely empty `facts` responses and for valid partial responses.

## Test

- `uv run pytest tests/test_fact_extraction_retry.py tests/test_content_policy_refusal_permanent.py -q` — 35 passed
- `uv run ruff check hindsight_api/engine/retain/fact_extraction.py tests/test_fact_extraction_retry.py`
- `uv run ruff format --check hindsight_api/engine/retain/fact_extraction.py tests/test_fact_extraction_retry.py`
- `git diff --check`

## Risk

Low and limited to malformed extraction output. Persistently malformed, non-empty fact arrays now fail loudly after the configured retries instead of being accepted as a zero-fact extraction. A valid retry or a valid partial response keeps the existing path.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -1492,6 +1492,7 @@ async def _extract_facts_from_chunk(
                     # In verbatim mode, 'what' is intentionally absent — text is backfilled from chunk
                     if extraction_mode != "verbatim":
                         logger.warning(f"Skipping fact {i}: missing 'what' field")
+                        has_malformed_facts = True
                         continue
 
                 # Critical field: fact_type — "assistant" maps to "experience", everything else is "world".
@@ -1660,6 +1661,9 @@ async def _extract_facts_from_chunk(
                     f"Got {len(raw_facts) - len(chunk_facts)} malformed facts out of {len(raw_facts)} on attempt {attempt + 1}/{outer_attempts}. Retrying..."
                 )
                 continue
+
+            if has_malformed_facts and not chunk_facts:
+                raise RuntimeError(f"Fact extraction failed: all facts were malformed after {outer_attempts} attempts")
 
             return chunk_facts, usage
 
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Superseded by #3760, which handles the same schema-drifted fact output upstream and is now merged. Closing this one to keep the queue clean.

Issue #3085, opened 2026-07-30
expand_observations semantic link CTE runs for minutes via prepared statements, completes in 1ms with inline literals


DiscountDarcy:
### Bug Description

Consolidation operations cause a single PostgreSQL backend process to peg at 90-95% CPU for 2-5+ minutes per memory unit. The bottleneck is the _expand_observations method's 4-hop graph traversal CTE, which lacks the timeout that the regular recall path has.

The official blog states consolidation is "LLM-bound — roughly 80%+ of wall-clock time" and that DB work is "comparatively negligible." Our experience is the opposite — the DB-side entity expansion dominates.

**Environment**

Hindsight: 0.8.5-slim (ghcr.io/vectorize-io/hindsight:0.8.5-slim)
PostgreSQL: 18.4 with pgvector 0.8.6
Bank: 7,465 nodes, 185,969 links, 334 unconsolidated observations (1/3)
Entity graph: 4,750 unique entities, 18,791 entity links, avg 3.04 entities/unit, max 38
LLM: Opencode-Go with Mimo-v2.5

**Configuration**

```
HINDSIGHT_API_ENABLE_AUTO_CONSOLIDATION=false
HINDSIGHT_API_CONSOLIDATION_BATCH_SIZE=1
HINDSIGHT_API_CONSOLIDATION_LLM_BATCH_SIZE=8
HINDSIGHT_API_CONSOLIDATION_LLM_PARALLELISM=4
HINDSIGHT_API_LINK_EXPANSION_TIMEOUT=10
HINDSIGHT_API_LINK_EXPANSION_PER_ENTITY_LIMIT=1  (tested: 1, 5, 30, 200)
HINDSIGHT_API_CONSOLIDATION_RECALL_BUDGET=low
HINDSIGHT_API_DB_POOL_MAX_SIZE=20
HINDSIGHT_API_READ_DB_POOL_MAX_SIZE=20
```

**Analysis:** 

The _expand_observations method (consolidation-specific) and the _expand_combined method (recall-path) use different timeout semantics:

Recall path (_expand_combined): Has graph_expansion_timeout (default 10s) that drops entity expansion entirely if the query exceeds the budget
Consolidation path (_expand_observations): Has no timeout — the 4-hop CTE runs to completion regardless of how long it takes

The 4-hop traversal is:
Observation → unnest source_memory_ids → source memories
Source memories → unit_entities → entities
[...]

Issue #1725, opened 2026-05-25
Observation updates append source_memory_ids without pruning inherited entities — entity set grows unboundedly while text narrows


eightHundreds:
## Summary

When an observation is updated during consolidation, `_execute_update_action` **appends** new `source_memory_ids` to the existing array without ever pruning stale entries. Since observations derive their entity set at read time via `source_memory_ids → unit_entities → entities`, this means the inherited entity set can only **grow**, even when the updated text describes fewer entities than before.

Over multiple consolidation rounds, an observation's text may narrow to describe 1–2 entities while its inherited entity set balloons to 15–20+, causing graph retrieval false positives and wasted recall tokens.

### 1. `source_memory_ids` is append-only on update

In `consolidator.py`, `_execute_update_action()` (line ~982):

```python
source_ids = list(model.source_fact_ids or []) + source_memory_ids
```

New source memory IDs are **appended**, never replaced or pruned. This is by design — it preserves the evidence chain so every source that ever contributed to an observation remains traceable.

### 2. Observations have no direct `unit_entities` rows

Observations never get `unit_entities` rows or `memory_links` rows with `link_type='entity'`. This is confirmed by `_create_memory_links()` being a no-op (`consolidator.py` line ~1100):

```python
async def _create_memory_links(conn, memory_id, observation_id):
    """
    Placeholder for observation link creation.
    Observations do NOT get any memory_links copied from their source facts.
    Instead, retrieval uses source_memory_ids to traverse:
    - Entity connections: observation -> source_memory_ids -> unit_entities
    """
    pass
```

### 3. Entity set is derived at read time from all source memories

In `ops_postgresql.py`, `expand_observations()` and the entity lookup in `memory_engine.py` traverse:

```
observation.source_memory_ids → unnest → unit_entities → entities
```

The observation inherits entities from **every** source memory in the array. Since the array only grows, the entity set only grows.

### 4. No entity re-extraction on update

When an observation's text is updated, `_execute_update_action` regenerates the embedding but does **not** re-extract entities. The consolidator has zero entity processing code — no import of `entity_processing`, no calls to `resolve_entities`. The `_UpdateAction` model only contains `text`, `observation_id`, and `source_fact_ids` — no entity list.

## Example Timeline

| Round | Text | source_memory_ids | Inherited entities |
|-------|------|-------------------|--------------------|
| 1 (create) | "Team A blocked by upstream X" | [fact_1, fact_2] | ["Team A", "X"] (2) |
[...]

PR #3715, opened 2026-08-22
perf(recall): score observation expansion without re-joining the connected_sources CTE

### LINKED ISSUE #3714: expand_observations: opaque CTE row estimate (1 vs 3,493) selects nested loop — 31M row comparisons, ~27.6s recall stall
# `expand_observations` scoring join: opaque CTE row estimate (1 vs 3,493) selects a nested loop, 31M row comparisons, ~27.6s

## Summary

The `scored` CTE in `expand_observations` (`hindsight_api/engine/db/ops_postgresql.py`) joins `candidates` against `connected_sources` — a CTE. PostgreSQL cannot see into a CTE scan to derive statistics, so it applies its hardcoded default estimate of **1 row**. The actual cardinality on an aged bank is **3,493 rows**.

Believing it is joining against a single row, the planner selects a nested loop. It then performs **31,003,646 row comparisons** to produce 3,715 output rows, taking **~27.6 seconds** of a saturated backend.

This is the same *class* of defect as #3085 (also `expand_observations`, also "fast in isolation / pathological in production"), but in a different join and with a different mechanism — that fix addressed the semantic-link CTE and prepared-statement plans; this is the `scored` shared-source counting join.

## Evidence

PostgreSQL 18.1, bank with 7,495 observations / 13,218 experiences / 3,064 world facts; `memory_links` ≈ 2.1M rows.

Actual CTE cardinalities with `GRAPH_SEED_LIMIT=20`, `per_entity_limit=200`:

```
seed_sources         50
source_entities      115
connected_sources    3493
candidate observations 2234
```

`EXPLAIN (ANALYZE, BUFFERS)` of the current query:

```
->  CTE Scan on connected_sources cs  (cost=0.00..0.02 rows=1 width=16)
                                      (actual time=0.001..2.120 rows=3493.00 loops=1)
->  Nested Loop  (cost=0.00..11.10 rows=211 width=32)
                 (actual time=19.493..27320.044 rows=3715.00 loops=1)
    Rows Removed by Join Filter: 31003646
Planning Time: 9.116 ms
Execution Time: 27633.224 ms
```

The `rows=1` estimate against `rows=3493` actual is the entire defect. Note `loops=3493` on the inner scan.

Confirming the planner (not the data) is the problem — forcing a hash join on the identical query and data:

```
SET enable_nestloop = off;
Execution Time: 372.385 ms          -- 74x faster, identical results
```

`WITH connected_sources AS MATERIALIZED (...)` alone does **not** help (`27,629ms`), which rules out repeated CTE evaluation as the cause and confirms it is join-shape selection driven by the bad estimate.

## Impact

This is not a micro-optimization. On the affected deployment, `expand_observations` dominates recall:

| recall stage | median (n=155 production recalls) | share |
|---|---:|---:|
| `[2]` parallel retrieval (contains `expand_observations`) | 24.12s | 54.4% |
| `[4]` reranking (cross-encoder) | 16.29s | 36.8% |
| `[1]` embedding | 0.056s | 0.1% |
| `[3]` RRF merge | 0.012s | 0.0% |
| `[6]` token filter | 0.008s | 0.0% |

Median total recall: **44.33s**. Recalls exceeding the service timeout fail outright with `ServiceException`; in a paired A/B one baseline arm took 64.55s and failed while every fixed arm completed.

Because `connected_sources` grows with entity fanout and bank age (per #1725, consolidation appends to `source_memory_ids` without pruning), the misestimate widens over time — the gap between the assumed 1 row and reality is unbounded. Young banks will not reproduce this; the query looks fine until it doesn't.

## Proposed fix

Rewrite `scored` to count shared sources by intersecting against the already-materialized `connected_array` rather than re-joining the opaque `connected_sources` CTE. `connected_array` is already computed in the same query for the `candidates` filter, so this adds no new work.

```sql
scored AS (
    SELECT c.id, (
        SELECT COUNT(DISTINCT s.source_id)::float
        FROM unnest(c.source_memory_ids) AS s(source_id)
        WHERE s.source_id = ANY(ca.source_ids)
    ) AS score
    FROM candidates c, connected_array ca
)
```

The `WHERE sc.score > 0` guard preserves the current inner-join semantics (rows with no shared source were previously dropped by the join). In practice it never fires — `&&` already guarantees at least one shared element, so no candidate reaching `scored` can score 0 — but it is kept so the rewrite is provably equivalent rather than equivalent-by-invariant.

Measured on the same data, same seeds:

| variant | execution time | result sha256 (first 16) |
|---|---:|---|
| current (opaque CTE join) | 27,733 ms | `c059b18224acb29d` |
| current + `MATERIALIZED` | 27,629 ms | `c059b18224acb29d` |
| **proposed (array intersect)** | **753 ms** | `c059b18224acb29d` |

**37x faster, byte-identical output** (300 rows, compared as an ordered `(id, score)` set).

Equivalence was checked across **6 independent randomized seed sets**, comparing the full ordered `(id, score)` result of the current and proposed queries by sha256 — all 6 matched exactly. The data also contains no empty `source_memory_ids` arrays and no arrays containing NULL, so those paths were verified as non-cases rather than assumed.

This keeps the work linear in the number of source ids — the same principle as the existing fix documented in the comment block above this query — and removes the planner's opportunity to choose a nested loop, rather than relying on it to estimate an opaque CTE correctly.

### Alternatives considered

- `SET enable_nestloop = off` — effective (372ms) but a session/database-wide sledgehammer that degrades unrelated queries. Not appropriate as a shipped fix.
- `MATERIALIZED` on the CTE — no effect, as shown above.
- Raising statistics targets — cannot help; the estimate is a hardcoded planner default for CTE scans, not a sampling problem.

## Environment

```
hindsight-api    0.9.1
PostgreSQL       18.1 (pgvector, embedded pg0 instance)
Python           3.14
platform         macOS 26.6.2 / Apple Silicon
graph retriever  link_expansion (default)
GRAPH_SEED_LIMIT 20
per_entity_limit 200 (DEFAULT_LINK_EXPANSION_PER_ENTITY_LIMIT)
```

Happy to open a PR with the query change plus a regression test asserting the plan does not select a nested loop against `connected_sources` at realistic cardinality.

trevornk:
## What

Scores observation graph expansion by array membership against the already-materialized `connected_array`, instead of re-joining the `connected_sources` CTE.

Fixes #3714.

## Why

PostgreSQL cannot derive statistics for a CTE scan, so it applies a hardcoded estimate of **1 row**. The `scored` CTE joined `candidates` against `connected_sources`, which on an aged bank actually holds thousands of rows — 3,493 in the reproduction.

Believing it was joining against one row, the planner chose a nested loop:

```
->  CTE Scan on connected_sources cs  (cost=0.00..0.02 rows=1 width=16)
                                      (actual time=0.001..2.120 rows=3493.00 loops=1)
->  Nested Loop  (cost=0.00..11.10 rows=211 width=32)
                 (actual time=19.493..27320.044 rows=3715.00 loops=1)
    Rows Removed by Join Filter: 31003646
Execution Time: 27633.224 ms
```

31M row comparisons to emit 3,715 rows. This is the same *class* of problem as #3085 (same function, "fast in isolation / pathological in production"), but a different join and a different mechanism — that fix addressed the semantic-link CTE and prepared-statement plans.

The misestimate widens as the bank ages: per #1725 consolidation appends to `source_memory_ids` without pruning, so `connected_sources` grows while the planner's assumption stays at 1. Young banks and test fixtures never reproduce it.

## Impact

On the affected deployment, this stage dominated recall — median across **155 production recalls**:

| stage | median | share |
|---|---:|---:|
| `[2]` parallel retrieval (contains `expand_observations`) | 24.12s | 54.4% |
| `[4]` reranking | 16.29s | 36.8% |
| `[1]` embedding | 0.056s | 0.1% |
| `[3]` RRF merge | 0.012s | 0.0% |

Recalls exceeding the service timeout failed outright with `ServiceException`; in a paired A/B one baseline arm took 64.55s and failed while every fixed arm completed.

## Results

Same bank (7,495 observations), same seeds:

| variant | execution | result sha256 |
|---|---:|---|
| current | 27,733 ms | `c059b18224acb29d` |
| current + `MATERIALIZED` | 27,629 ms | `c059b18224acb29d` |
| **this PR** | **753 ms** | `c059b18224acb29d` |

**37x faster, byte-identical output.**

`MATERIALIZED` alone changes nothing, which rules out repeated CTE evaluation and confirms the fault is join selection. `SET enable_nestloop=off` also fixes it (372ms) but is a session-wide sledgehammer, not a shipped fix.

## Correctness

Equivalence checked across **6 independent randomized seed sets**, comparing the full ordered `(id, score)` result by sha256 — all 6 matched exactly.

`WHERE sc.score > 0` preserves the previous inner-join semantics. In practice it never fires: `&&` in `candidates` already guarantees at least one shared source id (verified — zero score-0 rows across all 6 trials). It is kept so the rewrite is equivalent by construction rather than by invariant. The data also has no empty `source_memory_ids` arrays and no arrays containing NULL, verified rather than assumed.

## Tests

Two added to `test_observation_expansion_scoring.py`:

- `test_scoring_does_not_rejoin_the_connected_sources_cte` — pins the structural property that caused the blowup. Asserting wall-clock would be flaky on shared CI; asserting the scoring CTE contains no `JOIN connected_sources` is stable and catches a silent regression.
- `test_scoring_is_exact_across_wide_fanout` — guards against an off-by-one or double-count at the fan-out where the nested loop used to degrade.

The three pre-existing #3085 tests pass unchanged, including the duplicate-source-ids case that pins `COUNT(DISTINCT)` over `COUNT(*)`.

```
tests/test_observation_expansion_scoring.py  5 passed
related retrieval/graph/recall/observation suites: 527 passed, 28 skipped
```

The 13 errors in that broader run are `ValueError: LLM API key is required` (no `HINDSIGHT_API_LLM_API_KEY` in this environment) and are unrelated to this change — zero assertion failures. `ruff check` and `ruff format` clean.

## Note on the other half

This does not make recall fast on its own — it removes the largest stage and the timeout-driving work, taking median recall from ~44s to ~20s. The remaining floor is `[4]` reranking: the local `cross-encoder/ms-marco-MiniLM-L-6-v2` at 300 candidates on CPU, measured at 54.3 ms/candidate in production vs 23–25 ms/candidate for the same model standalone (contention; `RERANKER_LOCAL_MAX_CONCURRENT=4` × 4 torch threads on 10 cores). That is a separate concern with its own trade-offs — `DEFAULT_RERANKER_LOCAL_ALLOW_MPS = False` is deliberate given the documented allocator leak — so I have not touched it here. Happy to open a separate issue with those numbers if useful.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -915,6 +915,21 @@ class PostgreSQLOps(DataAccessOps):
         # Unnesting once and hash-joining connected_sources makes the work linear in
         # the number of source ids instead.
         #
+        # The score is computed against the materialized `connected_array` rather
+        # than by re-joining the `connected_sources` CTE. PostgreSQL cannot derive
+        # statistics for a CTE scan and falls back to a hardcoded estimate of 1 row;
+        # on an aged bank that CTE actually holds thousands (3,493 observed). Believing
+        # it was joining against a single row, the planner chose a nested loop and
+        # spent ~27.6s doing 31M row comparisons to emit 3,715 (issue #3714). Scoring
+        # by array membership against connected_array keeps the same set-wise semantics
+        # -- COUNT(DISTINCT) over shared source ids -- without exposing a join whose
+        # shape depends on an estimate the planner cannot make. MATERIALIZED does not
+        # help here: the problem is join selection, not repeated CTE evaluation.
+        #
+        # `score > 0` cannot filter anything the previous inner join kept: `&&` in
+        # `candidates` already guarantees at least one shared source id. It is retained
+        # so the rewrite is equivalent by construction rather than by invariant.
+        #
         # `connected_sources` caps each entity with row_number() rather than the
         # LATERAL + LIMIT that reads more naturally. Do not "simplify" it back
         # (issue #3510). The scoring join above is O(C + U) when planned as a hash
@@ -983,11 +998,12 @@ class PostgreSQLOps(DataAccessOps):
                   {window.clause("mu")}
             ),
             scored AS (
-                SELECT c.id, COUNT(DISTINCT cs.source_id)::float AS score
-                FROM candidates c
-                CROSS JOIN LATERAL unnest(c.source_memory_ids) AS s(source_id)
-                JOIN connected_sources cs ON cs.source_id = s.source_id
-                GROUP BY c.id
+                SELECT c.id, (
+                    SELECT COUNT(DISTINCT s.source_id)::float
+                    FROM unnest(c.source_memory_ids) AS s(source_id)
+                    WHERE s.source_id = ANY(ca.source_ids)
+                ) AS score
+                FROM candidates c, connected_array ca
             )
             SELECT
                 c.id, c.text, c.context, c.event_date, c.occurred_start,
@@ -996,6 +1012,7 @@ class PostgreSQLOps(DataAccessOps):
                 sc.score
             FROM candidates c
             JOIN scored sc ON sc.id = c.id
+            WHERE sc.score > 0
             ORDER BY sc.score DESC
             LIMIT $2
             """,
--- a/hindsight-api-slim/tests/test_observation_expansion_scoring.py
+++ b/hindsight-api-slim/tests/test_observation_expansion_scoring.py
@@ -233,3 +233,86 @@ async def test_per_entity_cap_bounds_hub_traversal(memory, request_context):
         assert outside not in scores, "the per-entity cap must exclude ids ranked below it"
     finally:
         await memory.delete_bank(bank_id=bank_id, request_context=request_context)
+
+
+@pytest.mark.asyncio
+async def test_scoring_does_not_rejoin_the_connected_sources_cte(memory, request_context):
+    """The scoring step must not re-join the opaque ``connected_sources`` CTE (#3714).
+
+    PostgreSQL cannot derive statistics for a CTE scan: it assumes 1 row. When the
+    score was computed by joining ``candidates`` against that CTE, an aged bank whose
+    CTE really held thousands of rows made the planner pick a nested loop and compare
+    tens of millions of rows to emit a few thousand -- ~27.6s for one recall arm, with
+    ``MATERIALIZED`` making no difference because the fault is join selection.
+
+    Asserting a wall-clock number would be flaky on shared CI, so this pins the
+    structural property that caused the blowup: the scoring CTE must not contain a
+    join against ``connected_sources``.
+    """
+    import inspect
+
+    from hindsight_api.engine.db.ops_postgresql import PostgreSQLOps
+
+    source = inspect.getsource(PostgreSQLOps.expand_observations)
+    scored = source[source.index("scored AS (") :]
+    scored = scored[: scored.index("LIMIT $2")]
+
+    assert "JOIN connected_sources" not in scored, (
+        "scoring must not join the connected_sources CTE -- PostgreSQL estimates a "
+        "CTE scan at 1 row and will choose a nested loop at real cardinality (#3714)"
+    )
+    assert "connected_array" in scored, "scoring should count shared sources against the materialized connected_array"
+
+
+@pytest.mark.asyncio
+async def test_scoring_is_exact_across_wide_fanout(memory, request_context):
+    """Shared-source counts stay exact when many candidates share many sources.
+
+    Guards the #3714 rewrite against an off-by-one or double-count at the fan-out
+    where the old nested loop used to blow up.
+    """
+    from hindsight_api.engine.db.ops import UpdatedWindow
+    from hindsight_api.engine.task_backend import fq_table
+
+    bank_id = f"test_obs_fanout_{uuid.uuid4().hex[:8]}"
+    try:
+        pool = await memory._get_pool()
+        backend = await memory._get_backend()
+        mu, ue, ml = fq_table("memory_units"), fq_table("unit_entities"), fq_table("memory_links")
+
+        async with pool.acquire() as conn:
+            await _ensure_bank(conn, bank_id)
+            entity_id = uuid.uuid4()
+            await conn.execute(
+                f"INSERT INTO {fq_table('entities')} (id, bank_id, canonical_name) VALUES ($1, $2, $3)",
+                entity_id,
+                bank_id,
+                "Hub",
+            )
+            facts = [await _insert_unit(conn, mu, bank_id, f"fact {i}", "world") for i in range(60)]
+            for fid in facts:
+                await conn.execute(f"INSERT INTO {ue} (unit_id, entity_id) VALUES ($1, $2)", fid, entity_id)
+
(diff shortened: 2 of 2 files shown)


koriyoshi2041:
Tested this at exact head `d6aba733e` on macOS / Python 3.11:

- `uv run pytest tests/test_observation_expansion_scoring.py -q` — 5 passed
- focused Ruff check and format check — passed
- `git diff --check` — passed

I also read the query rewrite against the existing overlap filter: the retained `score > 0` guard preserves the old inner-join behavior, while `COUNT(DISTINCT)` keeps duplicate source IDs from inflating scores. No issue found in this slice.

nicoloboschi:
Closing along with #3714, which this fixes — that issue is a duplicate of #3510, already fixed by #3588 (`5f137bf3`) on `main`. Full reasoning is in [my comment there](https://github.com/vectorize-io/hindsight/issues/3714#issuecomment-5394607217); the parts specific to this PR:

### The baseline in the results table isn't the code this PR changes

This branch is cut from `94017ede` (2026-08-21), which contains #3588. But the measurements are carried over verbatim from #3714 — same 27,733ms / 27,629ms / 753ms, same `c059b18224acb29d`, same 31,003,646 rows removed — and those were taken on `hindsight-api 0.9.1`, tagged 2026-08-14, which predates #3588 by four days.

So the "current (opaque CTE join): 27,733 ms" row describes the `LATERAL + LIMIT` form that no longer exists on this PR's own base. #3588's benchmark put the post-fix query at p50 217ms over 12 seed sets on a fixture built for this shape. The 37x is measured against something already removed.

### The rewrite reverts #3451

The new `scored` body is character-for-character the correlated subquery #3451 deleted:

```sql
(SELECT COUNT(DISTINCT s) FROM unnest(mu.source_memory_ids) s WHERE s = ANY(ca.source_ids))::float AS score
```

The comment block this PR adds sits directly above the one explaining why that shape was removed — it re-scans the connected-source array for every element of every candidate's array, and consolidation grows `source_memory_ids` without pruning (#1725), so the product widens with bank age: ~1.7B element comparisons at 5k observations averaging 113 sources (#3085). The new text and the retained text now contradict each other.

### The structural test would lock the regression in

`test_scoring_does_not_rejoin_the_connected_sources_cte` asserts `"JOIN connected_sources" not in scored` by `inspect.getsource`. That pins the absence of the shape #3451 introduced deliberately, so restoring the set-wise join later would have to delete this test. It also can't catch what it's named for: the join text is a proxy for a plan choice, and the plan depends on the estimate, which is what #3588 actually fixed.

`test_scoring_is_exact_across_wide_fanout` is worth keeping independently — exact shared-source counts at fan-out aren't covered today. Happy to take that one on its own if you want to open it separately against current `main`.

### If it still reproduces on `main`

Please reopen with a plan capture taken on `main` rather than 0.9.1. If the row_number() form still mis-plans on a bank shape #3588 didn't cover, that's a real and interesting finding — but it needs to be measured against the current query to tell.

Genuinely appreciate the rigor here: the 6-seed sha256 equivalence check, ruling out `MATERIALIZED`, and explicitly verifying the empty-array and NULL-element paths as non-cases rather than assuming them. That's the right method, aimed at a base that had already moved.

Issue #2550, opened 2026-07-03
Improve temporal ordering consideration for consolidation


noctuid:
### Use Case

I use hindsight to ingest jsonl agent sessions.

### Problem Statement

Reingesting older documents, or ingesting older documents for the first time can cause incorrect observations. If there is conflicting information in observations, newer information should always be prioritized regardless of ingestion order.

Some issues I see:

- the decision guide prompting should say to *only* consider what time information is from; the created at time is not the important thing and will be wrong for older documents; only when the actual content was was written matters
- "mentioned at" seems misleading since it is based on the document timestamp + synthetic amount of time that won't be correct when a session/document spans days or weeks; this is particularly an issue when documents overlap
- "occurred at" can sometimes be the actual time the source text was written (when you ingest messages with timestamps), but not if the discussed content takes place at some other time
- so neither of these can be reliably used to determine what memories should take priority

I have seen incorrect observations because of ingesting documents in close proximity or out of temporal order, e.g. something like "user used to like and now prefers" when it should be the opposite.

It seems to me that this could also be an issue for memories within the same session that conflict (the older memory may be preferred if a newer one does not clearly mention a change or correction).

### How This Feature Would Help

It would ensure accurate consolidation regardless of ingestion order and when documents span longer periods of time and overlap.

### Proposed Solution

There needs to be a timestamp that can be used to more accurately pinpoint when something was mentioned.

- "mentioned at" can keep the current implementation as a fallback, but maybe it can pull timestamps from json chunks (e.g. pull top level "timestamp" if it exists; related: https://github.com/vectorize-io/hindsight/issues/2548) or be optionally llm populated based on the chunk text (I think the latter makes sense/would be simpler especially if there are multiple messages in the chunk since they could have very different timestamps)
- order consolidation by "mentioned at" not "created at"; "mentioned at" affects observations, and the llm is instructed to specifically consider prioritizing the later "mentioned at" memories when processing conflicting information (e.g. so a newly processed memory "User's favorite frontend framework is react" from an *old* document does not incorrectly change an existing observation "User's favorite frontend framework is svelte"; the final observation should be "User's favorite frontend framework was react but now is svelte.")

### Priority

Important - affects my workflow

### Checklist

- [ ] I would be willing to contribute this feature

nicoloboschi:
each memory fact has an extracted (optional) time range that the consolidation is aware of so technically it's already aware of it. 

[...]

PR #3717, opened 2026-08-22
fix(retain): derive mentioned_at from per-fact source timestamps

chaoyuan-guo:
## Problem

Facts extracted from the same retain item currently inherit the item's `event_date` plus the existing synthetic offsets. For a multi-message session spanning several days, this can make `mentioned_at` inaccurate even when each source record contains its own timestamp.

## Changes

- During semantic fact extraction, ask the model for the timestamp of the source message or record that directly states each fact.
- Prefer timezone-aware ISO 8601 timestamps from source messages or records; fall back to the retain item's `event_date` when none is available.
- Preserve source timestamps without synthetic offsets; keep the existing offset behavior only for `event_date` fallbacks.
- Keep `occurred_start` / `occurred_end` as the time of the described event.
- Correct the `mentioned_at` definition in the recall documentation.

## Scope

PR #2952 documented how consolidation should interpret `mentioned_at`, but did not change how the field is derived. This Draft PR addresses a timestamp propagation gap related to Issue #2550: source-record timestamps do not reach a fact's `mentioned_at`.

This change only adjusts how `mentioned_at` is derived during retain. It does not change consolidation ordering or conflict resolution. The runtime check only verifies that a candidate timestamp appears in the current source chunk; the extraction model determines whether it belongs to that fact.

For documentation, this Draft updates only the central Recall field definition. Related wording in Reflect, MCP/API, generated clients, and examples is intentionally deferred until maintainers confirm the retain-side contract.

Before marking this PR ready for review, I would like to confirm whether deriving source timestamps during retain fact extraction is the intended approach.

## Tests

- Deterministic coverage for timestamp validation, `event_date` fallback, streaming and batch extraction, chunk isolation, existing fallback offset behavior, and the `mentioned_at` value returned by the public read API after a retain operation.
- A real-LLM judge test distinguishing source-message time from the time of the described event.

Related to Issue #2550.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
@@ -10,7 +10,8 @@ import json
 import logging
 import re
 from collections.abc import Iterator
-from datetime import datetime, timedelta
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
 from typing import Any, Literal, cast
 
 from pydantic import BaseModel, ConfigDict, Field, create_model, field_validator
@@ -160,6 +161,9 @@ class Fact(BaseModel):
     # Optional temporal fields
     occurred_start: str | None = None
     occurred_end: str | None = None
+    mentioned_at: str | None = None
+    # Pipeline-only provenance used by the offset step; never serialize it as fact data.
+    mentioned_at_from_source: bool = Field(default=False, exclude=True)
 
     # Optional location field
     where: str | None = Field(
@@ -198,6 +202,13 @@ class FactCausalRelation(BaseModel):
     )
 
 
+_MENTIONED_AT_FIELD_DESCRIPTION = (
+    "Timezone-aware ISO 8601 timestamp for when the source message or record directly stating this fact was written. "
+    "Prefer the extended form, such as YYYY-MM-DDTHH:MM:SSZ. This is not when the described event happened. "
+    "Return null when unavailable, invalid, or not uniquely attributable to this fact."
+)
+
+
 class ExtractedFact(BaseModel):
     """A single extracted fact."""
 
@@ -215,6 +226,7 @@ class ExtractedFact(BaseModel):
     fact_kind: str = Field(default="conversation", description="'event' or 'conversation'")
     occurred_start: str | None = Field(default=None, description="ISO timestamp for events")
     occurred_end: str | None = Field(default=None, description="ISO timestamp for event end")
+    mentioned_at: str | None = Field(default=None, description=_MENTIONED_AT_FIELD_DESCRIPTION)
     fact_type: Literal["world", "assistant"] = Field(
         description="'world' = objective/external facts, including user preferences, rules, corrections, and constraints even when stated during a conversation. 'assistant' = actions, experiences, or observations the assistant/agent actually performed."
     )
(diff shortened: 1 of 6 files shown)


hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py
```diff
@@ -2935,6 +2998,61 @@ def _parse_datetime(date_str: str):
...
+    """Accept a source time only when its instant is present in source text.
+
+    This membership guard blocks pure hallucinations; it does not prove that
+    the instant semantically belongs to this particular fact.
+    """
+    source_mentioned_at = _parse_source_mentioned_at(value)
+    if source_mentioned_at is not None:
+        if source_mentioned_at.astimezone(timezone.utc) in source_instants:
+            return _MentionedAtResolution(value=source_mentioned_at, from_source=True)
+    return _MentionedAtResolution(value=event_date, from_source=False)
```
Sanderhoff-alt, line 3053: A source instant is accepted regardless of `event_date`, which breaks the documented no-timestamp opt-out. A retain item with `timestamp: "unset"` is mapped to `event_date=None` (`hindsight-api-slim/hindsight_api/api/http.py:8186`) and is documented at `http.py:749-756` as "store without a timestamp"; the code this replaces returned `None` in that case, and the comment it removed said so explicitly ("or None when the caller opted into no timestamp", also noted at `engine/retain/types.py:321`).

After this change, any content that happens to carry an explicit-timezone ISO 8601 timestamp — JSON/JSONL logs, chat exports, a `created_at` field — will produce a persisted `mentioned_at` for a caller who asked for none, since the prompt section and the membership guard are both independent of `event_date`.

Suggested guard, if the opt-out is meant to hold:

```python
source_mentioned_at = _parse_source_mentioned_at(value) if event_date is not None else None
```

If accepting source times under `unset` is intentional instead, it needs to be stated in the PR description and the recall docs, since it changes a documented contract. Either way, `test_source_timestamp_resolution_contract` currently only covers `candidate=None` with `event_date=None`; a case with a valid, present candidate and `event_date=None` would pin whichever behaviour is chosen.

Sanderhoff-alt (commented):
The derivation looks sound for the normal case, but it overrides the explicit "store without a timestamp" opt-out. One inline note.

nicoloboschi:
mentioned_at is on purpose related to the document and deterministic. please explain your needs in a new issue and we can discuss what's the best way to fix it, thanks!

chaoyuan-guo:
> The derivation looks sound for the normal case, but it overrides the explicit "store without a timestamp" opt-out. One inline note.

This is indeed a problem, thanks for pointing it out. @Sanderhoff-alt

chaoyuan-guo:
> mentioned_at is on purpose related to the document and deterministic. please explain your needs in a new issue and we can discuss what's the best way to fix it, thanks!

@nicoloboschi Thank you for the response. This PR was a concrete attempt to address the inaccurate mentioned_at described in Issue #2550: messages within the same document or session may span multiple days and overlap with the time ranges of other documents, so deterministically tying mentioned_at to the document timestamp does not appear sufficient to represent when each fact was actually mentioned.
For example, document A has a timestamp of July 1 but contains a message written on July 10 saying, “I now prefer Svelte.” Document B has a timestamp of July 5 and says, “I prefer React.” Based on the document timestamps, React would appear newer, even though Svelte is actually the more recent information.
In my understanding, the requirements in #2550 conflict with the current document-level, deterministic design of mentioned_at. If you think this should be addressed differently, I’d be happy to discuss alternative approaches.

PR #3719, opened 2026-08-22
fix(db): exit when the connection pool has produced nothing for two minutes

cdbartholomew:
## Problem

A connection pool can reach a state it never leaves. Every `acquire` fails, the retries in `acquire_with_retry` exhaust, the caller gets an exception, and the next caller repeats it forever. The process is up, serving nothing, and will stay that way indefinitely.

Nothing recovers from this today. Liveness is deliberately database-free — a probe that queries the database restarts every replica at once whenever the database is merely slow, which is worse than the problem it detects — so the pod stays alive and permanently unready with no path back.

This is not hypothetical. It was observed in a live environment following a database failure: once the database came back and was healthy again, the API processes did not recover with it. Their pools stayed dead and every request kept failing until the processes were restarted by hand.

## Change

`acquire_with_retry` now records the last acquire that succeeded. When a call has spent its retries and *nothing* has succeeded anywhere for the whole window, it logs at `critical`, flushes the log handlers, and exits so the supervisor can replace the process.

`HINDSIGHT_API_DB_UNAVAILABLE_EXIT_SECONDS`, default `120`, `0` disables.

Design notes, all captured in comments at the relevant lines:

- **The window is measured from the last success, not a failure count.** A process serving little traffic may attempt an acquire only rarely, and a counter would never trip.
- **Any single success clears it.** Intermittent failures never accumulate toward the limit.
- **Non-retryable errors don't count.** A bug in our own code says nothing about whether the database is reachable.
- **`os._exit`, not an exception.** A raise would be caught by whatever request handler happened to trigger it and the process would carry on in the same unusable state.
- **Two minutes.** It has to clear the ordinary causes — a CNPG failover takes tens of seconds, a database restart a little longer — because exiting during those would turn a brief degradation into a restart loop across every replica at once. Beyond that, shorter is better: the window is time a wedged process spends serving nothing. This started at 600s, which was far longer than anything it needs to ride out.

Both acquire paths are wired: the `DatabaseBackend` path and the legacy raw `asyncpg.Pool` path.

## Also

Three acquire/retry log lines switch from `{e}` to `{e!r}`. Several driver exceptions stringify to the empty string, so the final error rendered as `Database acquire failed after N attempts: ` with the cause silently dropped — you couldn't tell auth failure from timeout from an exhausted pool. This cost me real debugging time.

## Tests

Six new cases in `tests/test_db_utils.py`, driving the real config path via env + `clear_config_cache` rather than patching a module constant:

| Case | Expected |
|---|---|
| Sustained total failure past the window | exits once |
| Failing acquires, still inside the window | does **not** exit |
| A success mid-run | resets the clock; a later failure starts a fresh window |
| Window set to `0` | never exits |
| Legacy raw-pool path | exits the same way |
| Non-retryable error | does not exit |

Verified as a real regression test, not just a green run: neutralizing both call sites makes the exit test fail and leaves the other five passing, so it is proving the fix rather than the harness.

`./scripts/hooks/lint.sh` clean. Config flag follows the house convention — `ENV_`/`DEFAULT_` constants, `HindsightConfig` field, `from_env` wiring, `.env.example` plus the byte-identical `hindsight-embed` copy, and a row in `docs/developer/configuration.md` next to the existing pool-observability flags.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -407,7 +407,7 @@ HINDSIGHT_API_LOG_LEVEL=info
 # seconds, so the supervisor can restart it. Liveness is deliberately DB-free, so
 # a pool that never recovers otherwise leaves the process up and permanently
 # unready. 0 disables it.
-# HINDSIGHT_API_DB_UNAVAILABLE_EXIT_SECONDS=600
+# HINDSIGHT_API_DB_UNAVAILABLE_EXIT_SECONDS=120
 
 # -----------------------------------------------------------------------------
 # Extensions (Optional)
--- a/.env.example
+++ b/.env.example
@@ -403,6 +403,12 @@ HINDSIGHT_API_LOG_LEVEL=info
 # HINDSIGHT_API_LOOP_WATCHDOG_POLL_INTERVAL_MS=250
 # HINDSIGHT_API_DB_ACQUIRE_WARN_THRESHOLD_MS=1000
 
+# Exit the process when no pooled DB connection has been acquired for this many
+# seconds, so the supervisor can restart it. Liveness is deliberately DB-free, so
+# a pool that never recovers otherwise leaves the process up and permanently
+# unready. 0 disables it.
+# HINDSIGHT_API_DB_UNAVAILABLE_EXIT_SECONDS=600
+
 # -----------------------------------------------------------------------------
 # Extensions (Optional)
 # -----------------------------------------------------------------------------
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -587,6 +587,7 @@ ENV_LOOP_WATCHDOG_ENABLED = "HINDSIGHT_API_LOOP_WATCHDOG_ENABLED"
 ENV_LOOP_WATCHDOG_STALL_THRESHOLD_MS = "HINDSIGHT_API_LOOP_WATCHDOG_STALL_THRESHOLD_MS"
 ENV_LOOP_WATCHDOG_POLL_INTERVAL_MS = "HINDSIGHT_API_LOOP_WATCHDOG_POLL_INTERVAL_MS"
 ENV_DB_ACQUIRE_WARN_THRESHOLD_MS = "HINDSIGHT_API_DB_ACQUIRE_WARN_THRESHOLD_MS"
+ENV_DB_UNAVAILABLE_EXIT_SECONDS = "HINDSIGHT_API_DB_UNAVAILABLE_EXIT_SECONDS"
 
 # Vertex AI configuration
 ENV_LLM_VERTEXAI_PROJECT_ID = "HINDSIGHT_API_LLM_VERTEXAI_PROJECT_ID"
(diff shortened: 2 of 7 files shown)


nicoloboschi:
Thanks for chasing this down — the incident is real and the debugging pain in the "Also" section is the most useful thing in the PR. We're taking that half now and holding the rest.

**Merging separately: #3762** — the `str(e)` → `repr(e)` change, plus the same fix in `health_check`'s readiness payload, with regression tests.

**Holding: the exit mechanism.** Two reasons.

### 1. The root cause isn't identified, and it probably isn't "the database was unavailable"

asyncpg self-heals across an outage by design. `PoolConnectionHolder.acquire()` (0.30, `pool.py:137-147`) drops a dead connection and dials a new one on every acquire:

```python
async def acquire(self) -> PoolConnectionProxy:
    if self._con is None or self._con.is_closed():
        self._con = None
        await self.connect()          # → _get_new_connection()
```

Nothing from the outage survives in the holder. Once the server accepts connections again, the next acquire succeeds. So "the database came back healthy and the pools stayed dead" is not something the pool does on its own — something else was true, and we don't know what. The two candidates that would produce exactly this symptom:

- **The pool was in a terminal state and the error was misclassified as transient.** asyncpg raises `InterfaceError` for `'pool is closed'` (`pool.py:986`), `'pool is closing'` (`:860`) and `'pool is not initialized'` (`:984`). `_is_retryable` matches on `type(exc).__name__`, and `InterfaceError` is in `_RETRYABLE_EXCEPTION_NAMES` — the same name as a dropped connection. A closed pool therefore retries forever, indifferent to database health. If this is what happened, the fix is a few lines of classification, not a process exit.
- **Every holder was checked out and never released.** With `max_size=100` and `acquire_timeout=30`, a leak eventually makes every acquire time out with `asyncio.TimeoutError` — also retryable, also forever, also independent of database health. Fix is the leak.

In both cases the new log line ("No database connection has been acquired for Ns… the pool is not recovering") sends the next on-call at the database, when the database was fine.

And this is circular: by your own account the incident logs rendered as `Database acquire failed after N attempts: ` with the cause blanked out. So the diagnosis behind the feature was inferred from logs that had the deciding field missing. #3762 is what supplies it. It should ship first and be in place for the next occurrence.

### 2. As written it exits healthy processes

`_note_acquire_success()` is only reachable from `acquire_with_retry`, but the two things that touch the database most reliably don't go through it:

- `MemoryEngine.health_check()` — `memory_engine.py:4051`, `async with backend.acquire()` — the readiness probe, every few seconds in every deployment.
- The worker claim loop — `worker/poller.py:467`, same — every 500 ms.

Meanwhile the shortest *guaranteed* `acquire_with_retry` caller is the maintenance loop, whose shortest job interval is 300s (`DEFAULT_CONSOLIDATION_RECONCILE_INTERVAL_SECONDS`) — more than double the 120s window. So on a low-traffic replica:

1. t=0 — maintenance tick succeeds, clock stamped.
2. t=0…300 — readiness probes and worker polls all succeed. Clock untouched.
3. t=300 — maintenance tick lands during a sub-second blip. `maintenance.py:294` passes `max_retries=1`, so that's two attempts.
4. `unavailable_for` = 300s > 120s → `os._exit(1)`.

Reproduced on this branch at the shipped default:

```
CRITICAL No database connection has been acquired for 300s (limit 120s).
         The pool is not recovering; exiting so the supervisor can replace this process.
```

on a backend that was never unavailable for more than one retry cycle. That's the restart-storm the PR description sets out to avoid — on idle replicas the window is spent before the outage starts.

Two smaller things in the same area: `_last_acquire_success` is seeded at *import*, so a slow cold start (model download, migrations) can burn the window before the first acquire is attempted, and one transient failure then crashloops the pod. And `maintenance.py:_purge_expired` deliberately catches acquire failures (`"Retention sweep discovery failed…"; return 0`) — the exit fires inside `acquire_with_retry` before the raise, so that handler no longer runs.

### What we'd want to revisit it

Not a no. Concretely:

1. #3762 in place, and a real error class captured from a recurrence.
2. Gate on the failure *streak*, not staleness — track first-failure-since-last-success and exit only when failures have been continuous for the window. That makes the condition match what the log line claims.
3. Record success at the single choke point — `pool_instrumentation._record_acquire_wait` (`engine/db/pool_instrumentation.py:106`) runs on every successful pooled acquire on both drivers, so it covers the probe, the worker poll, and the ~40 direct `backend.acquire()` sites.

If you want a safety net before all that, the narrow version is safe today and needs no window: when the backend reports the pool closed or uninitialised, that is *provably* unrecoverable — exit immediately, no false positives possible.

Closing this one; happy to reopen or review a follow-up built on #3762.

PR #3720, opened 2026-08-22
fix(worker): detect stuck tasks by stage progress

BrianHotopp:
Use the current stage age for worker stuck detection.

A large consolidation kept advancing through LLM batches but crossed the total-age threshold, producing false `STUCK` alerts and stack dumps.

- Advancing long-running tasks no longer emit `[STUCK?]` or stack dumps.
- An unchanged stage still emits escalating stack dumps after 5 minutes.
- Stack dumps retain total task age for context.

Test: `pytest tests/test_worker_stuck_watchdog.py -q`

### DIFF
--- a/hindsight-api-slim/hindsight_api/worker/poller.py
+++ b/hindsight-api-slim/hindsight_api/worker/poller.py
@@ -113,7 +113,7 @@ logger = logging.getLogger(__name__)
 PROGRESS_LOG_INTERVAL = 30
 
 # Stuck-task stack-dump thresholds (seconds). Each task gets one stack dump
-# per threshold it crosses (5min, 10min, 20min, 40min, 80min...).
+# per threshold its current stage crosses without making progress.
 STUCK_STACK_INITIAL_THRESHOLD_S = 300
 STUCK_STACK_MAX_THRESHOLD_S = 3600 * 6  # cap doubling at 6h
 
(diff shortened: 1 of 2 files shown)


hindsight-api-slim/hindsight_api/worker/poller.py
```diff
@@ -1851,30 +1852,36 @@ def _log_per_task_lines(self, active_tasks: dict[str, ActiveTaskInfo], now: floa
...
+        stalled stage remains unchanged. Total task age is retained in the log
+        for context but cannot distinguish a large, advancing job from a wedge.
         """
-        if age_s < STUCK_STACK_INITIAL_THRESHOLD_S:
+        stage = info.stage_holder.stage if info.stage_holder else "unknown"
+        if stage != info.last_stack_dump_stage:
+            info.last_stack_dump_stage = stage
+            info.last_stack_dump_threshold = 0
+
+        if stage_age_s < STUCK_STACK_INITIAL_THRESHOLD_S:
```
Sanderhoff-alt, line 1878: Consolidation processes tag groups concurrently under a single worker task (`consolidation_llm_parallelism` defaults to 4; `_gather_or_cancel([...])` in `hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py:1821`), and every one of those coroutines mutates the same `StageHolder` bound to the task. `set_stage` refreshes `updated_at` on each call, so `stage_age_s` measures "time since any branch last reported", not time spent in the current stage.

Consequence: if one group wedges (a hung LLM read, or a wait on `scope_locks`) while the other three keep committing batches, `stage_age_s` stays near zero and neither `[STUCK?]` nor `[STUCK_STACK]` ever fires — the partial-hang case the watchdog exists to diagnose becomes undetectable in the default config. The labels also alternate between `consolidation.llm_batch.N` values across branches, so the new stage comparison resets `last_stack_dump_threshold` on nearly every poll and the 5/10/20-minute escalation can never build up.

Suggest keeping a total-age fallback (one dump when total age crosses a much larger threshold, even if the stage advanced), or recording the stage per concurrent branch so a stalled branch is still visible.

Sanderhoff-alt (commented):
The stage-progress signal is not per-stage when a task runs work concurrently, which makes the watchdog silent for partially wedged consolidations under the default configuration. Details inline.

nicoloboschi (approved):
LGTM

PR #3724, opened 2026-08-22
fix(recall): add timeout/fallback guard to expand_observations entity CTE

### LINKED ISSUE #3723: recall() fails on TimeoutError in _expand_observations(); sibling _expand_combined() degrades gracefully
## Summary

PR #911 added a per-entity LATERAL cap and an `asyncio.wait_for`/fallback guard to `_expand_combined()`'s entity CTE (`link_expansion_retrieval.py`). Per that PR's own description, `_expand_observations()` got the cap too, but not the timeout+fallback guard — confirmed on current `main`:

- `_expand_combined()` (line ~330): `await asyncio.wait_for(conn.fetch(...), timeout=config.link_expansion_timeout)` wrapped in `try/except asyncio.TimeoutError:`, falling back to a semantic+causal-only query.
- `_expand_observations()` (line ~412): calls `ops.expand_observations()` with no wrapper at all.

## Impact

On a large/dense bank (~410K `entity_cooccurrences` rows in our case — comment above `expand_observations()` in `ops_postgresql.py` already documents this exact query's performance sensitivity, issues #1725/#3085/#3510), the entity CTE inside `expand_observations()` occasionally exceeds whatever timeout eventually fires below the application layer. Because there's no guard, the raw `TimeoutError` propagates and fails the *entire* `recall()` call for `fact_type=["observation"]`:

```
RuntimeError: Failed to search memories (TimeoutError): TimeoutError()
```

instead of degrading to semantic+causal results the way the non-observation path already does.

## Steps to reproduce

No large dataset needed — this simulates the real-world slow-query timeout deterministically via monkeypatch, so it runs in under a second on unpatched `main`.

1. Clone and set up the dev environment:
   ```bash
   git clone https://github.com/vectorize-io/hindsight.git
   cd hindsight/hindsight-api-slim
   uv sync --extra all --extra test --group dev
   ```

2. Save this as `tests/test_expand_observations_timeout_gap_repro.py`:
   ```python
   import uuid

   import pytest

   from hindsight_api.engine.db.ops_postgresql import PostgreSQLOps
   from hindsight_api.engine.retain import embedding_utils


   def _to_str(emb):
       return "[" + ",".join(str(v) for v in emb) + "]"


   @pytest.mark.asyncio
   async def test_expand_observations_has_no_timeout_guard(memory, request_context, monkeypatch):
       bank_id = f"test-expand-obs-timeout-gap-{uuid.uuid4().hex[:8]}"
       await memory.get_bank_profile(bank_id, request_context=request_context)

       try:
           fact_id, obs_id = str(uuid.uuid4()), str(uuid.uuid4())
           fact_text = "Alice migrated the billing service to the new cluster"
           obs_text = "Alice handles infrastructure migrations"
           embeddings = await embedding_utils.generate_embeddings_batch(memory.embeddings, [fact_text, obs_text])

           pool = await memory._get_pool()
           async with pool.acquire() as conn:
               ent_id = await conn.fetchval(
                   "INSERT INTO entities (bank_id, canonical_name, mention_count) VALUES ($1, $2, 1) RETURNING id",
                   bank_id, "billing service",
               )
               await conn.execute(
                   "INSERT INTO memory_units (id, bank_id, text, fact_type, embedding, event_date) "
                   "VALUES ($1, $2, $3, 'world', $4::vector, now())",
                   fact_id, bank_id, fact_text, _to_str(embeddings[0]),
               )
               await conn.execute("INSERT INTO unit_entities (unit_id, entity_id) VALUES ($1, $2)", fact_id, ent_id)
               await conn.execute(
                   "INSERT INTO memory_units (id, bank_id, text, fact_type, embedding, event_date, "
                   "source_memory_ids, proof_count) VALUES ($1, $2, $3, 'observation', $4::vector, now(), $5::uuid[], 1)",
                   obs_id, bank_id, obs_text, _to_str(embeddings[1]), [fact_id],
               )

           async def _boom(self, *args, **kwargs):
               # Simulates the real asyncpg TimeoutError a slow/dense bank produces.
               raise TimeoutError()

           monkeypatch.setattr(PostgreSQLOps, "expand_observations", _boom)

           from hindsight_api.engine.memory_engine import Budget

           # BUG: this raises RuntimeError("Failed to search memories (TimeoutError): TimeoutError()")
           # instead of degrading to semantic+causal results the way fact_type=["world"]
           # already does when _expand_combined()'s entity CTE times out.
           result = await memory.recall_async(
               bank_id=bank_id,
               query="Alice billing service migration",
               fact_type=["observation"],
               budget=Budget.MID,
               max_tokens=2048,
               request_context=request_context,
               _quiet=True,
           )
           assert result.results is not None
       finally:
           await memory.delete_bank(bank_id, request_context=request_context)
   ```

3. Run it:
   ```bash
   uv run pytest tests/test_expand_observations_timeout_gap_repro.py -v -n0
   ```

**Observed on unpatched `main`:**
```
RuntimeError: Failed to search memories (TimeoutError): TimeoutError()
```

**Expected:** recall degrades to semantic+causal results (still finds the seed observation via semantic search) instead of failing, matching how `fact_type=["world"]` already behaves when `_expand_combined()`'s entity CTE times out (see `test_graph_entity_fanout_cap.py::test_entity_expansion_timeout_fallback` for the equivalent passing case on the sibling path — same low-`link_expansion_timeout` trick, no monkeypatch needed there since the guard already exists).

## Fix

Opening a PR shortly: wraps just the `entity_rows` fetch inside `expand_observations()` in the same `asyncio.wait_for`/`except asyncio.TimeoutError` pattern, falling back to an empty entity result set while the semantic+causal query still runs — mirroring `_expand_combined()`'s existing behavior.

jordigilh:
Fixes #3723.

## Problem

`_expand_combined()` has an `asyncio.wait_for`/fallback guard around its entity CTE (added in #911); its sibling `_expand_observations()` only got that PR's per-entity LATERAL cap, not the timeout guard. On a large/dense bank the entity query inside `expand_observations()` can still exceed budget, and with no guard the raw `TimeoutError` fails the whole `recall()` call instead of degrading to semantic+causal results.

## Fix

Wraps just the `entity_rows` fetch in `PostgreSQLOps.expand_observations()` with the same `asyncio.wait_for(..., timeout=config.link_expansion_timeout)` / `except asyncio.TimeoutError` pattern already used by `_expand_combined()`, falling back to `entity_rows = []` (the semantic+causal query still runs independently, unaffected).

## Test plan

- Added `test_expand_observations_falls_back_gracefully_on_timeout`, mirroring the existing `test_entity_expansion_timeout_fallback` technique (forces the real query to exceed an artificially low `link_expansion_timeout`) for the observation path. Fails with the exact production error (`RuntimeError: Failed to search memories (TimeoutError): TimeoutError()`) without this fix, passes with it.
- Kept `test_expand_combined_gracefully_falls_back_on_timeout` alongside it as a contrast/sanity check that the sibling path's existing behavior is unchanged.
- `uv run pytest tests/test_expand_observations_timeout_gap.py tests/test_link_expansion_retrieval.py tests/test_observation_expansion_scoring.py tests/test_graph_entity_fanout_cap.py` all green locally (one unrelated pre-existing error in a test requiring a real LLM API key I don't have configured).
- `ruff check --fix` / `ruff format` clean.

Not touched: Oracle's `expand_observations()` (`ops_oracle.py`) has the same shape but I don't have an Oracle test environment to validate a change there — flagging in case a maintainer wants the same guard applied there.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -5,6 +5,7 @@ efficient batch operations.
 """
 
 import asyncio
+import logging
 from datetime import datetime
 
 from .base import DatabaseConnection
@@ -17,6 +18,8 @@ from .ops import (
     graph_maintenance_bank_serialization_sql,
 )
 from .result import ResultRow
+
+logger = logging.getLogger(__name__)
 
 
 def pg_search_vector_expr(
@@ -936,8 +939,22 @@ class PostgreSQLOps(DataAccessOps):
         # grow hubs far past that, re-measure before assuming this is still the
         # right shape.
 
-        entity_rows = await conn.fetch(
-            f"""
+        # Unlike _expand_combined()'s entity CTE (link_expansion_retrieval.py),
+        # this query has no caller-side asyncio.wait_for/fallback guard, so a
+        # slow run on a large/dense bank (very high entity fanout even under
+        # per_entity_limit, see issue #3510's notes above) propagates a raw
+        # TimeoutError all the way up and fails the entire recall() call
+        # instead of degrading to semantic+causal results the sibling path
+        # already falls back to. Apply the same guard here, scoped to just
+        # this query, so an observation-fact-type recall degrades the same
+        # way a non-observation one does.
+        from ...config import get_config
+
+        config = get_config()
+        try:
+            entity_rows = await asyncio.wait_for(
+                conn.fetch(
+                    f"""
             WITH seed_sources AS (
                 SELECT DISTINCT unnest(source_memory_ids) AS source_id
                 FROM {mu_table}
@@ -999,10 +1016,18 @@ class PostgreSQLOps(DataAccessOps):
             ORDER BY sc.score DESC
             LIMIT $2
             """,
-            seed_ids,
-            budget,
-            *window.params,
-        )
+                    seed_ids,
+                    budget,
+                    *window.params,
+                ),
+                timeout=config.link_expansion_timeout,
+            )
+        except asyncio.TimeoutError:
+            logger.warning(
+                f"[ExpandObservations] Entity expansion timed out after "
+                f"{config.link_expansion_timeout}s, falling back to semantic+causal only"
+            )
+            entity_rows = []
 
         # Exact v0.5.6 query shape: GROUP BY + MAX(weight) for semantic,
         # DISTINCT ON for causal, hardcoded to fact_type='observation'.
--- a/hindsight-api-slim/tests/test_expand_observations_timeout_gap.py
+++ b/hindsight-api-slim/tests/test_expand_observations_timeout_gap.py
@@ -1,0 +1,169 @@
+"""Regression test: `_expand_observations()`'s entity-expansion query now
+degrades gracefully on timeout, matching its sibling `_expand_combined()`.
+
+Background
+----------
+PR #911 ("fix(recall): cap entity fanout in graph expansion") added two
+protections to `LinkExpansionRetriever._expand_combined()` in
+`hindsight_api/engine/search/link_expansion_retrieval.py`:
+
+1. A LATERAL per-entity cap (`link_expansion_per_entity_limit`, default 200).
+2. An `asyncio.wait_for(..., timeout=config.link_expansion_timeout)` wrapper
+   (default 10s), with `except asyncio.TimeoutError:` falling back to
+   semantic+causal-only results instead of failing the whole recall.
+
+`_expand_observations()` (the sibling path for `fact_type="observation"`)
+received protection #1 but not #2 — its `ops.expand_observations()` call was
+never wrapped in a timeout/fallback, so a slow run on a large/dense bank
+(very high entity fanout even under the per-entity cap — see the "issue
+#3510" notes in `ops_postgresql.py::expand_observations`) propagated a raw
+`TimeoutError` all the way up and failed the entire recall call:
+`RuntimeError: Failed to search memories (TimeoutError): TimeoutError()`,
+observed in production on a bank with ~410K `entity_cooccurrences` rows.
+
+This test seeds directly via SQL (no LLM, no waiting on real consolidation)
+and forces the real entity-expansion query inside `expand_observations()` to
+exceed `link_expansion_timeout` — the same technique already established in
+`test_graph_entity_fanout_cap.py::test_entity_expansion_timeout_fallback` for
+the sibling path — and asserts recall now degrades instead of raising.
+"""
+
+import uuid
+
+import pytest
+
+from hindsight_api.engine.retain import embedding_utils
+
+
+def _to_str(emb: list[float]) -> str:
+    return "[" + ",".join(str(v) for v in emb) + "]"
+
+
+async def _seed_world_fact_and_observation(memory, bank_id: str) -> dict:
+    """Seeds directly via SQL -- same technique as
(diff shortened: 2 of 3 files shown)


hindsight-api-slim/tests/test_expand_observations_timeout_gap.py
```diff
@@ -0,0 +1,169 @@
...
+                budget=Budget.MID,
+                max_tokens=2048,
+                request_context=request_context,
+                _quiet=True,
+            )
+
+            # Recall succeeds even though entity expansion timed out; the
+            # seed observation itself is still found via semantic search.
+            assert result.results is not None
+            assert len(result.results) > 0
```
Sanderhoff-alt, line 131: This test passes on `main` as well, so it does not guard the fix. Before this change `link_expansion_timeout` is read in exactly one place (`hindsight_api/engine/search/link_expansion_retrieval.py:345`, inside `_expand_combined`), and `fact_type=["observation"]` routes to `_expand_observations`, which never consulted it — so lowering the timeout cannot raise `TimeoutError` on the old code path. With the guard in place the assertion here is satisfied by the semantic seed regardless of whether entity expansion ran. Please make the test discriminate, for example assert via `caplog` that the new `[ExpandObservations] Entity expansion timed out` warning fired and that no entity-expanded rows came back.

hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
```diff
@@ -936,8 +939,22 @@ async def expand_observations(
...
+        # TimeoutError all the way up and fails the entire recall() call
+        # instead of degrading to semantic+causal results the sibling path
+        # already falls back to. Apply the same guard here, scoped to just
+        # this query, so an observation-fact-type recall degrades the same
+        # way a non-observation one does.
+        from ...config import get_config
+
+        config = get_config()
+        try:
+            entity_rows = await asyncio.wait_for(
```
Sanderhoff-alt, line 955: The guard is added to the PostgreSQL ops class only. `hindsight_api/engine/db/ops_oracle.py:717` has the same unguarded `entity_rows = await conn.fetch(...)` inside its `expand_observations`, so on Oracle 23ai a slow entity CTE still propagates `TimeoutError` and fails the entire recall (`memory_engine.py:7850`). The wrapper is plain Python around the fetch with no dialect-specific SQL, so it can be mirrored in the Oracle implementation without an Oracle environment, or factored into a helper both ops classes call.

Sanderhoff-alt (commented):
Two items to address before merge: the new regression test also passes without the change, and the Oracle implementation of the same query is left unguarded, so Oracle deployments keep the failure this fixes.

hindsight-api-slim/tests/test_expand_observations_timeout_gap.py
jordigilh: Good catch, confirmed. Added a caplog check for the `[ExpandObservations] Entity expansion timed out` warning, which only fires once the guard exists and actually runs. Ran the test against `main` and this branch to confirm: fails on `main` (no warning), passes here. Pushed in 5403261.

hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
jordigilh: Confirmed, same gap. Mirrored the wait_for/except TimeoutError guard into ops_oracle.py's expand_observations, same log message format as the PG fix. Left the two PG-only optimizations you noted alone, they're already explained inline. Pushed in 5403261.

nicoloboschi:
it's already fixed in 0.9.2

PR #3730, opened 2026-08-23
fix(retain): sanitize model-derived fact text

### LINKED ISSUE #3729: Retain fails when structured LLM fact output contains a lone surrogate
> **Transparency note:** This issue was investigated and written up with an AI coding agent. A human account owner reviewed and sponsored the submission. The failure was observed in a real self-hosted Hindsight deployment.

## Bug Description

Retain can fail after successful fact extraction when otherwise-valid structured LLM output contains an unpaired UTF-16 surrogate in a fact field. The resulting `Fact.fact` is passed to the embedding backend and the local SentenceTransformers tokenizer rejects it:

```text
Exception: Failed to generate batch embeddings: TextEncodeInput must be Union[TextInputSequence, Tuple[InputSequence, InputSequence]]
```

This is a different boundary from #1875 / #1891. That fix sanitizes client-provided content and context at engine ingress. In this case, the original retained document is valid UTF-8; the LLM introduces a lone surrogate while generating structured fact JSON, after ingress sanitization has already run.

## Steps to Reproduce

1. Retain valid UTF-8 source text using fact extraction and a local SentenceTransformers embeddings backend.
2. Have the configured LLM return a structured fact containing a valid JSON surrogate escape without its matching half, for example:

   ```json
   {
     "facts": [{
       "what": "Alex laughed 😂",
       "when": "N/A",
       "where": "N/A",
       "who": "Alex",
       "why": "The joke was funny \ude02",
       "fact_type": "world",
       "fact_kind": "conversation"
     }]
   }
   ```

3. Structured parsing succeeds and produces a Python string containing the lone low surrogate.
4. Fact extraction builds `Alex laughed 😂 | Involving: Alex | The joke was funny <lone surrogate>`.
5. Retain reaches `generate_embeddings_batch`, where SentenceTransformers raises `TextEncodeInput` and the async operation eventually fails after retries.

This was reproduced from real model output and with a deterministic regression against `_extract_facts_from_chunk`: clean source input plus the structured response above produces a `Fact.fact` that cannot be UTF-8 encoded on current `main`.

## Expected Behavior

Model-derived fact text should be UTF-8-safe before it reaches storage or embedding. Valid Unicode, including complete emoji, should be preserved; impossible lone surrogate halves should be stripped in the same way as client input.

## Actual Behavior

The lone surrogate survives structured parsing and fact construction, causing embedding generation to fail and the retain operation to be retried and ultimately marked failed.

## Version

Observed on Hindsight 0.9.1. Reproduced on `main` at `da9a588`.

## LLM Provider

Other: a local Gemma model exposed through an OpenAI-compatible endpoint. The defect is provider-independent because any structured provider result containing a lone surrogate reaches the same fact construction path.

## Proposed Fix

Sanitize `Fact.fact` at the final extracted-fact model boundary. Both immediate and batch extraction paths construct this model before storage/embedding, so this is one narrow choke point that covers both without changing unrelated JSON parsing or duplicating cleanup logic.

BrianMcBrayer:
## Summary

Fixes #3729.

Retain can receive valid UTF-8 source text yet still fail during embedding when the LLM emits a lone UTF-16 surrogate in a structured fact field. This change sanitizes the final model-derived fact text before either the immediate or batch retain path can store or embed it.

## Root cause

#1891 sanitizes client-provided content and context at engine ingress. Structured providers parse the LLM response later, and JSON decoding can turn a model-generated `\udXXX` escape into a Python string containing a lone surrogate. Both extraction paths then construct `Fact.fact` from that parsed output without another UTF-8-safety boundary, so the invalid code point reaches SentenceTransformers and raises `TextEncodeInput`.

## Fix

Add a Pydantic field validator to the final extracted `Fact` model and reuse the existing `sanitize_llm_output` behavior. `Fact` is the shared convergence point for immediate and batch extraction, so the correction is applied once at the boundary that owns the derived text.

The sanitizer removes impossible surrogate halves while preserving valid Unicode, including complete emoji. It does not change prompts, model interpretation, provider parsing, or user content semantics.

## Impact

- Prevents retrying and ultimately failing retain operations when structured LLM output contains a lone surrogate.
- Covers immediate and batch extraction without duplicated path-specific cleanup.
- Leaves the existing ingress defense for client-provided text unchanged.

## Tests

Added a deterministic regression that supplies valid source text and a structured LLM response containing both a valid emoji and a lone low surrogate. It fails on `main` because the resulting fact is not UTF-8 encodable, and passes with this change while preserving the valid emoji.

Verified locally:

```text
./scripts/hooks/lint.sh
PYTHONIOENCODING=utf-8:backslashreplace uv run pytest -n 0 tests/test_input_robustness.py -k 'not retain_with and not recall_with' -q
uv run pytest -n 0 tests/test_fact_extraction_retry.py -q
./scripts/hooks/check-unused.sh  # advisory; only pre-existing findings outside this diff
```

Prepared with AI coding-agent assistance; the linked issue contains the production evidence and reproduction details reviewed by the account owner.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
+++ b/hindsight-api-slim/hindsight_api/engine/llm_wrapper.py
@@ -13,6 +13,7 @@ from contextlib import AsyncExitStack, asynccontextmanager
 from typing import TYPE_CHECKING, Any
 
 from json_repair import repair_json
+from pydantic import BaseModel
 
 # Vertex AI imports (conditional - for LLMProvider to pass credentials to GeminiLLM)
 try:
@@ -163,6 +164,9 @@ def _request_params(
     return params or None
 
 
+_UNSAFE_TEXT_RE = re.compile(r"[\x00-\x08\x0b\x0c\x0e-\x1f\x7f\ud800-\udfff]")
+
+
 def sanitize_text(text: str | None) -> str | None:
     """
     Sanitize text by removing characters that break downstream systems.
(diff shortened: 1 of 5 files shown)


BrianMcBrayer:
@strix-security please check again because I rebased

BrianMcBrayer:
Hi — I wanted to follow up when you have a chance. I’m using Hindsight in production as the memory layer for my personal assistant, and I’m currently having to work around this by stopping parts of my retain backlog. The source documents are valid UTF-8, but occasionally my local LLM produces a lone surrogate while extracting a fact. That value then reaches the embedder, causing the retain operation to retry and eventually fail.

I’ve rebased the PR onto the current `main`.

I know you have a lot of contributions coming in, and there’s no intention to rush you. I mainly wanted to provide some context on how this is affecting my installation. If you would prefer the sanitization at a different boundary, or want anything changed in the test or implementation, I’d be glad to revise it.

Thank you for your time, and for building and maintaining Hindsight. I’ve been getting a great deal of value from it.

BrianMcBrayer:
Hi @nicoloboschi — when you have a chance, would you mind taking a look at this PR and approving the pending fork CI workflow? Both runs are currently waiting at `action_required`, so I am unable to move the checks forward from my side. I would be glad to rebase or adjust the implementation or test if you would prefer a different approach.

Thank you very much for your time.

PR #3731, opened 2026-08-23
fix(embed): default macOS daemons to CPU inference

### LINKED ISSUE #3727: macOS: embedded daemon segfaults in torch MPS during consolidation — force_cpu should default ON for daemon mode
## Summary

On macOS (Apple Silicon), running the embedded daemon with the default local embeddings +
reranker providers segfaults in PyTorch's MPS backend during consolidation. The daemon comes
up cleanly, serves a recall, then dies within ~2 minutes. `force_cpu` fixes it — but it
defaults to `False`, so every macOS host hits this until someone finds the switch.

**Ask: default `embeddings_local_force_cpu` and `reranker_local_force_cpu` to `True` when
running in daemon mode on macOS** (or at minimum, detect MPS + daemon mode and warn).

## Crash

`torch 2.13.0`, `mps.is_available() == True`. macOS crash report, faulting thread:

```
exception  : EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS at 0x0000000000000000
termination: Segmentation fault: 11

libtorch_cpu.dylib    at::native::mps::MetalShaderLibrary::exec_unary_kernel(...)
libtorch_cpu.dylib    at::native::mps::copy_cast_kernel_mps(at::Tensor&, at::Tensor const&)
libtorch_cpu.dylib    at::native::mps::mps_copy_(at::Tensor&, at::Tensor const&, bool)
libtorch_cpu.dylib    at::native::copy_(at::Tensor&, at::Tensor const&, bool)
libtorch_cpu.dylib    at::_ops::to_device::call(...)
libtorch_python.dylib torch::autograd::THPVariable_to(...)
```

Observed cycle, repeated ~16 times in one afternoon:

```
migrations completed · pool created · Worker poller started · Uvicorn on 127.0.0.1:9177
claimed 2 tasks (2 consolidation) · [CONSOLIDATION] total_unconsolidated=12
resource_tracker: There appear to be 1 leaked semaphore
<process gone>
```

A second failure shape alternated with the segfault: the process surviving but pinned at
**~1200% CPU** and never completing the batch. Both end with the queue not draining.

An isolated `SentenceTransformer.encode()` on MPS runs clean, so this is specific to the
tensor traffic on the consolidation path rather than to MPS generally.

## The fix already exists and is off by default

`hindsight_api/engine/embeddings.py`:

```python
# Force CPU mode if configured (used in daemon mode to avoid MPS/XPC issues on macOS)
if self.force_cpu:
    device = "cpu"
else:
    ...
    if has_gpu:
        device = None      # let sentence-transformers auto-detect GPU/MPS/XPU
```

`cross_encoder.py` mirrors it. The comment shows the failure mode is known. But
`DEFAULT_EMBEDDINGS_LOCAL_FORCE_CPU = False` and `DEFAULT_RERANKER_LOCAL_FORCE_CPU = False`,
so the protection is opt-in and undiscoverable until you have a segfault to chase.

Setting both to `1` fixed it immediately and verifiably:

```
Embeddings: forcing CPU mode
Reranker: forcing CPU mode (HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU=1)

before: died every ~2 min, 1224% CPU, 16 crash reports
after:  15m48s uptime, 0.4% CPU, zero crash reports, health green
```

CPU consolidation is slower. Slower and alive is the right default for a background daemon.

## Why a default change rather than documentation

The daemon runs unattended. When it segfaults there is no Python traceback — the process
simply disappears, and the only real evidence is a macOS `.ips` crash report most operators
never look at. `fp16` is already opt-in for the reranker with the comment *"faster on
MPS/CUDA"*; GPU acceleration being opt-in and stability being the default is the safer
polarity for this component.

koriyoshi2041:
## Problem

Embedded daemons on Apple Silicon can still crash or hang in PyTorch MPS during consolidation. The daemon currently lets local embeddings and reranking auto-select MPS unless users discover and set both `force_cpu` flags themselves.

## Fix

Default both local inference paths to CPU when `DaemonEmbedManager` starts a child on macOS. Explicit environment or profile values remain authoritative, so users can still opt into MPS.

This restores the safety polarity that previously shipped in #1010, without changing Linux or Windows behavior.

## Test

- `uv run pytest tests/test_profile_daemon_config.py -q -k "macos_daemon or profile_env_propagates"` (3 passed)
- `uv run ruff check hindsight_embed/daemon_embed_manager.py`
- `uv run ruff format --check hindsight_embed/daemon_embed_manager.py tests/test_profile_daemon_config.py`
- `git diff --check`

The full file currently has three unrelated entrypoint-test failures in this slim local environment because `_find_api_command` detects missing local ML dependencies and falls back to `uvx`; all daemon-environment tests pass.

## Risk

macOS embedded consolidation is slower by default. Callers who prefer MPS can preserve it with explicit `HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU=0` and `HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU=0`.

Fixes #3727.

### DIFF
--- a/hindsight-embed/hindsight_embed/daemon_embed_manager.py
+++ b/hindsight-embed/hindsight_embed/daemon_embed_manager.py
@@ -765,6 +765,13 @@ class DaemonEmbedManager(EmbedManager):
         if "HINDSIGHT_EMBED_DAEMON_IDLE_TIMEOUT" not in env:
             env["HINDSIGHT_EMBED_DAEMON_IDLE_TIMEOUT"] = str(DEFAULT_DAEMON_IDLE_TIMEOUT)
 
+        # Local PyTorch inference can still crash in MPS during unattended
+        # daemon workloads on macOS. Prefer the stable CPU path by default,
+        # while preserving an explicit opt-in to MPS from the caller/profile.
+        if platform.system() == "Darwin":
+            env.setdefault("HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU", "1")
+            env.setdefault("HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU", "1")
+
         # Tell the daemon child it was already launched in a detached session
         # (via our Popen below) so daemonize() skips the redundant re-exec.
         env["_HINDSIGHT_DAEMON_CHILD"] = "1"
--- a/hindsight-embed/tests/test_profile_daemon_config.py
+++ b/hindsight-embed/tests/test_profile_daemon_config.py
@@ -304,6 +304,78 @@ def test_profile_env_propagates_arbitrary_hindsight_keys_to_daemon(temp_home, mo
     assert env.get("HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU") == "1"
 
 
+def test_macos_daemon_forces_local_inference_to_cpu_by_default(temp_home, monkeypatch):
+    """macOS daemon children should avoid MPS unless the user opts in."""
+    from unittest.mock import MagicMock, patch
+
+    from hindsight_embed.daemon_embed_manager import DaemonEmbedManager
+
+    monkeypatch.delenv("HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU", raising=False)
+    monkeypatch.delenv("HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU", raising=False)
+    manager = DaemonEmbedManager()
+    captured: dict[str, dict[str, str]] = {}
+    popen_called = [False]
+
+    def fake_popen(cmd, env, **kwargs):
+        captured["env"] = env
+        popen_called[0] = True
+        proc = MagicMock()
+        proc.pid = 12345
+        return proc
+
+    def fake_is_running(profile=""):
+        return popen_called[0]
+
+    with (
+        patch("hindsight_embed.daemon_embed_manager.subprocess.Popen", side_effect=fake_popen),
+        patch("hindsight_embed.daemon_embed_manager.time.sleep"),
+        patch.object(manager, "_clear_port", return_value=True),
+        patch.object(manager, "_find_api_command", return_value=["hindsight-api"]),
+        patch.object(manager, "is_running", side_effect=fake_is_running),
+        patch("hindsight_embed.daemon_embed_manager.platform.system", return_value="Darwin"),
+    ):
+        manager._start_daemon(config={}, profile="")
+
+    assert captured["env"]["HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU"] == "1"
+    assert captured["env"]["HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU"] == "1"
+
+
+def test_macos_daemon_preserves_explicit_mps_opt_in(temp_home, monkeypatch):
+    """Explicit force-CPU overrides must win over the macOS defaults."""
+    from unittest.mock import MagicMock, patch
+
+    from hindsight_embed.daemon_embed_manager import DaemonEmbedManager
+
+    monkeypatch.setenv("HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU", "0")
+    monkeypatch.setenv("HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU", "false")
+    manager = DaemonEmbedManager()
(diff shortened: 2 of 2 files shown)


hindsight-embed/hindsight_embed/daemon_embed_manager.py
```diff
@@ -765,6 +765,13 @@ def _start_daemon_locked(
         if "HINDSIGHT_EMBED_DAEMON_IDLE_TIMEOUT" not in env:
             env["HINDSIGHT_EMBED_DAEMON_IDLE_TIMEOUT"] = str(DEFAULT_DAEMON_IDLE_TIMEOUT)
 
+        # Local PyTorch inference can still crash in MPS during unattended
+        # daemon workloads on macOS. Prefer the stable CPU path by default,
+        # while preserving an explicit opt-in to MPS from the caller/profile.
+        if platform.system() == "Darwin":
+            env.setdefault("HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU", "1")
+            env.setdefault("HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU", "1")
```
Sanderhoff-alt, line 773: MPS is no longer auto-selected on main. `select_local_device` (`hindsight-api-slim/hindsight_api/engine/local_device.py:66-87`) returns `"cpu"` whenever MPS is the only accelerator and `allow_mps` is false, and both `DEFAULT_EMBEDDINGS_LOCAL_ALLOW_MPS` and `DEFAULT_RERANKER_LOCAL_ALLOW_MPS` are `False` (`config.py:981`, `config.py:1011`). So a macOS daemon started today already uses CPU for embeddings and reranking, and these two `setdefault` calls change nothing in the default path — the code quoted in #3727 predates that change.

The one case where the block does change behaviour is a regression: a user who opts into MPS the documented way, `HINDSIGHT_API_EMBEDDINGS_LOCAL_ALLOW_MPS=1` (`hindsight-docs/docs/developer/configuration.md:712`), now also gets `HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU=1` injected into the daemon env, and `force_cpu` short-circuits at `local_device.py:66-67` before `allow_mps` is ever consulted. The opt-in silently stops working, with no log line saying why.

Suggest dropping this block. If a daemon-specific guard is still wanted on top of the existing default, it must not be able to override `*_ALLOW_MPS` — e.g. only set `FORCE_CPU` when neither `*_ALLOW_MPS` var is present in `env`.

hindsight-embed/tests/test_profile_daemon_config.py
```diff
@@ -304,6 +304,78 @@ def fake_is_running(profile=""):
...
+        patch.object(manager, "is_running", side_effect=fake_is_running),
+        patch("hindsight_embed.daemon_embed_manager.platform.system", return_value="Darwin"),
+    ):
+        manager._start_daemon(config={}, profile="")
+
+    assert captured["env"]["HINDSIGHT_API_EMBEDDINGS_LOCAL_FORCE_CPU"] == "1"
+    assert captured["env"]["HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU"] == "1"
+
+
+def test_macos_daemon_preserves_explicit_mps_opt_in(temp_home, monkeypatch):
```
Sanderhoff-alt, line 343: This test does not cover the opt-in it is named after. `HINDSIGHT_API_*_LOCAL_FORCE_CPU=0` is not how MPS is opted into — `*_LOCAL_ALLOW_MPS=1` is (`local_device.py:79-80`), and with `ALLOW_MPS=1` and `FORCE_CPU` unset the new default injects `FORCE_CPU=1` and the assertion here would not catch it. If the guard is kept, please add a case that sets only `HINDSIGHT_API_EMBEDDINGS_LOCAL_ALLOW_MPS=1` / `HINDSIGHT_API_RERANKER_LOCAL_ALLOW_MPS=1` and asserts `FORCE_CPU` is not forced on.

Sanderhoff-alt (commented):
The premise this change is built on no longer holds on current main: local inference on Apple Silicon already runs on CPU unless the user explicitly opts into MPS. As written the block is a no-op in the default case and a regression in the opt-in case. Details inline.

koriyoshi2041:
You’re right—the existing device selector already defaults Apple Silicon to CPU, and this patch overrides the documented ALLOW_MPS opt-in. Closing it rather than preserving a redundant daemon-level guard. Thanks for tracing the current path.

PR #3732, opened 2026-08-23
fix(coding-agents): find Claude sessions for underscore paths

pstanton237:
## Problem

`--import-conversations` locates Claude Code transcripts by encoding the repository's absolute path into a directory name under `~/.claude/projects`.

Claude Code replaces `/`, `.`, and `_` with `-`, but `claudeProjectDir()` only replaced `/` and `.`. For a repository such as `/work/my_repo`, Claude writes sessions under:

```text
~/.claude/projects/-work-my-repo
```

The importer searched for:

```text
~/.claude/projects/-work-my_repo
```

Neither the exact-directory check nor its subdirectory-prefix check could match. The install completed successfully but reported `no past sessions found on disk`, even when Claude transcripts existed.

## Fix

Include `_` in the project-directory encoding:

```diff
-repoDir.replace(/[/.]/g, "-")
+repoDir.replace(/[/._]/g, "-")
```

The existing attribution guard remains unchanged. Directory names are only used to narrow the candidates; each transcript must still record a `cwd` within the requested repository before it is imported. This prevents the lossy encoding from pulling in a sibling repository whose path maps to the same directory name.

A regression test creates a transcript in Claude Code's underscore-to-hyphen directory and verifies that `importLocalHistory()` finds it.

## Verification

The regression failed against the previous implementation with `expected [] to have a length of 1`, then passed with the fix.

- `npm test -- --run src/core/history.test.ts` — 9 passed
- `npm test` — 629 passed across 53 test files
- `npm run build` — passed
- Prettier check for the two changed files — passed

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/history.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/history.test.ts
@@ -47,18 +47,50 @@ describe("local history import", () => {
     expect(JSON.stringify(r.sessions)).not.toContain("unrelated");
   });
 
-  it("reads Claude sessions when the repository path contains underscores", () => {
+  // Claude Code encodes EVERY non-alphanumeric character as `-`, not just `/` and `.`. Encoding
+  // only the separators left whole repositories invisible to `--import-conversations`: the install
+  // reported "no past sessions found on disk" while the transcripts sat there under the real name.
+  // A space is the common case in the field (`~/Documents/My Projects/...`), not just underscores.
+  it.each([
+    ["underscores", "/Users/x/dev/my_project", "-Users-x-dev-my-project"],
+    ["spaces", "/Users/x/My Projects/app", "-Users-x-My-Projects-app"],
+    ["mixed punctuation", "/Users/x/dev/hs+odd@repo v2", "-Users-x-dev-hs-odd-repo-v2"],
+    ["dots", "/Users/x/dev/repo.git", "-Users-x-dev-repo-git"],
+  ])("reads Claude sessions when the repository path contains %s", (_label, repo, encoded) => {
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Thanks for this — the diagnosis was right and the regression test was the right shape. I confirmed the premise empirically by running Claude Code 2.1.241 in a directory with an underscore:

```
cwd:  .../scratchpad/hs_under_test
dir:  ~/.claude/projects/...-scratchpad-hs-under-test
```

While verifying it I found the rule is broader than underscores. A second probe:

```
cwd basename:  hs+odd@repo v2
dir:           ...-hs-odd-repo-v2
```

`+`, `@` and **space** all encode to `-` as well — Claude Code replaces *every* non-alphanumeric character, preserving case (`crewAI` stays `crewAI`) and not collapsing runs (`claude-501/-Users` -> `claude-501--Users`). So `/[/._]/g` fixed the character you hit but left the same bug for the rest; a path like `~/Documents/My Projects/repo` was still invisible.

I pushed one commit on top of yours widening it to `/[^a-zA-Z0-9]/g`, and extended your test into a parameterized set (underscores, spaces, mixed punctuation, dots) plus two cases that pin the contract: case-preservation/no-collapsing, and that a sibling repo encoding to the same directory name is *not* imported. That last one is what makes widening safe — as your PR description notes, the directory name only narrows candidates and the recorded `cwd` still does the attributing.

Your commit and authorship are untouched; the push was a fast-forward, no rebase. I also approved the workflow run, which had been sitting at `action_required` since you opened this — that's why only the Strix check had reported. `test-coding-agents` now passes (646 tests).

Merging. Thanks again.

Issue #3718, opened 2026-08-22
coding-agents (Windows): detached background seed spawn and git execFileSync calls lack windowsHide — blank console windows on every session start


vkeezou:
## Environment

- OS: Windows (win32 x64)
- Package: `@vectorize-io/hindsight-coding-agents` 0.4.1 (loaded as a DeepSeek Harness profile plugin)
- Host: DSH Desktop 0.1.1 (Electron shell). The harness node is spawned hidden (`windowsHide: true`), so it has no visible console; any console-subsystem child it spawns without `windowsHide` gets its own visible console window.

## Affected code (hindsight-integrations/coding-agents)

- `src/core/seed.ts` — `startBackgroundSeed()` spawns the `deepen.js` worker with `{ detached: true, stdio: "ignore" }` and **no `windowsHide`**. On Windows a detached console child gets its own console, and with `stdio: "ignore"` that console is **blank** — a blank CMD window pops up and lingers while the seed worker ingests git history.
- `src/core/git.ts` — `git()` helper (`execFileSync("git", ["-C", repo, ...args], { encoding: "utf8", maxBuffer: 1 << 28 })`) — no `windowsHide`.
- `src/core/bank.ts` — `getProjectRootFromGit()` (`execFileSync("git", ["rev-parse", ...], { stdio: ["ignore","pipe","ignore"], timeout: 1e3 })`) — no `windowsHide`.
- `src/core/status.ts` — `commitCount()`, and `src/core/session-start.ts` — `gitSyncNote()` — no `windowsHide`.
- `src/deepen.ts` — `git()` helper and the `rev-list` call — no `windowsHide`.
- `startCodebaseSurvey()` (src/core/survey.ts) spawns agent CLIs detached without `windowsHide` (same pattern).

Note: `startDaemonDetached()` already sets `windowsHide: true` — the seed/survey/git paths simply missed it.

## Suggested fix

Add `windowsHide: true` to the spawn options in `startBackgroundSeed()` / `startCodebaseSurvey()`, and to the options objects of the `execFileSync("git", ...)` calls listed above. The option is ignored on POSIX.

### Steps to Reproduce

1. Install `@vectorize-io/hindsight-coding-agents` (0.4.1) as a DSH profile plugin on Windows (DSH Desktop 0.1.1, harness node started hidden with `windowsHide: true`).
2. Start/restart a session (happens on every app boot and every restart).
[...]

PR #3743, opened 2026-08-24
fix(windows): hide daemon subprocess windows

### LINKED ISSUE #3725: Annoying uv.exe command popups
### Bug Description

I understand this is from python but this makes coding sessions unusable after a while. There has to be a way to suppress the command window from showing up each and every prompt turn with claude and vs code

### Steps to Reproduce

Enable Hindsight in local docker
As claude code cli in vs code anything
Popups ov uv.exe happen relentlessly each turn.

### Expected Behavior

No popups should show. Make the calls headless silently in the background.

### Actual Behavior

Popups happening too frequent

### Version

latest

### LLM Provider

Anthropic

Sanderhoff-alt:
## Summary

Issue #3725 reports recurring visible `uv.exe` console windows during Claude Code sessions on Windows when Hindsight runs in local daemon mode.

## Root Cause

```mermaid
flowchart TD
    A[Coding-agent lifecycle] --> B[Daemon preflight]
    A --> C[Background seed]
    A --> D[Codebase survey]
    B --> E[uvx --version]
    B --> F[Detached daemon starter]
    F --> G[hindsight-all]
    G --> H[uvx profile and daemon commands]
    C --> I[Detached deepen engine]
    D --> J[Detached headless coding agent]
    E --> K[Visible Windows console]
    F --> K
    H --> K
    I --> K
    J --> K
```

Node.js child-process APIs default `windowsHide` to `false` on Windows. The daemon probes, detached daemon starter, background seed, codebase survey, and `hindsight-all` command runner therefore allowed console windows to appear during normal coding-agent lifecycle operations.

## Changes

| Component | Change | Effect |
| --- | --- | --- |
| Coding agents daemon probes | Add `windowsHide: true` to `cargo`, `uvx`, and PATH checks | Suppresses probe windows |
| Coding agents daemon starter | Add `windowsHide: true` to the detached Node process | Suppresses background starter windows |
| Background seed | Add `windowsHide: true` to the detached deepen engine | Suppresses session-start seed windows |
| Codebase survey | Add `windowsHide: true` to the detached headless agent | Suppresses cold-repository survey windows |
| `hindsight-all` | Add `windowsHide: true` to daemon `spawn()` calls | Suppresses `uvx` command windows |
| Regression tests | Assert hidden-window options across both packages | Prevents recurrence |

## Verification

- Coding agents daemon, seed, and survey tests: 38 passed.
- `hindsight-all` server tests: 10 passed.
- Repository lint passed.
- Prettier checks passed.
- `git diff --check` passed.

The fix is Windows-specific in effect and does not change process lifecycle timing, command arguments, or behavior on other platforms.
Fixes #3725

### DIFF
--- a/hindsight-all-npm/src/server.test.ts
+++ b/hindsight-all-npm/src/server.test.ts
@@ -1,5 +1,8 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import { HindsightServer } from "./server.js";
+
+const spawnMock = vi.hoisted(() => vi.fn());
+vi.mock("child_process", () => ({ spawn: spawnMock }));
 
 describe("HindsightServer construction", () => {
   it("defaults base URL to http://127.0.0.1:8888", () => {
@@ -31,5 +34,26 @@ describe("HindsightServer construction", () => {
     const server = new HindsightServer({ port: 1, readyTimeoutMs: 100 });
     const healthy = await server.checkHealth();
     expect(healthy).toBe(false);
+  });
+
+  it("hides daemon command windows on Windows", async () => {
+    const child = {
+      stdout: { on: vi.fn() },
+      stderr: { on: vi.fn() },
+      on: vi.fn((event: string, handler: (code?: number) => void) => {
+        if (event === "exit") handler(0);
+        return child;
+      }),
+    };
+    spawnMock.mockReturnValueOnce(child);
+
+    const server = new HindsightServer();
+    await server.stop();
+
+    expect(spawnMock).toHaveBeenCalledWith(
+      "uvx",
+      ["hindsight-embed@latest", "daemon", "--profile", "default", "stop"],
+      expect.objectContaining({ stdio: "pipe", windowsHide: true })
+    );
   });
 });
--- a/hindsight-all-npm/src/server.ts
+++ b/hindsight-all-npm/src/server.ts
@@ -102,7 +102,7 @@ export class HindsightServer {
     });
     const args = [...baseArgs, "daemon", "--profile", this.profile, "stop"];
 
-    const child = spawn(cmd, args, { stdio: "pipe" });
+    const child = spawn(cmd, args, { stdio: "pipe", windowsHide: true });
     this.pipeOutput(child, "daemon.stop");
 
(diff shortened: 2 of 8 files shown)


nicoloboschi:
Thanks — this is correct, and the `hindsight-all-npm` half is the piece that actually addresses #3725 (the `uvx` command runner is where the visible `uv.exe` windows come from). I'd like to land it.

One thing first: #3736 just merged (0429060be) for #3718, and the two PRs overlap. Comparing them line by line:

- `src/core/seed.ts` and `src/core/survey.ts` — **byte-identical** changes in both PRs
- `src/core/daemon.ts` — this PR is a **superset**: it has the same detached-spawn line *plus* the `cargo` / `uvx` / `which` probes, which #3736 did not touch
- `hindsight-all-npm/src/server.ts` — **only here**

So could you rebase onto current `main` and reduce this to just the parts that are still missing?

- keep `hindsight-all-npm/src/server.ts` (+ its test)
- keep the `cargo` / `uvx` / `which` `execFileSync` probes in `src/core/daemon.ts`
- drop the `seed.ts` / `survey.ts` / `daemon.ts` spawn changes and their test assertions — already on `main`

That leaves a clean, conflict-free diff that closes #3725 without re-landing what's already there. The approach itself is right; `windowsHide` is inert on POSIX, so the risk note in your description holds.

nicoloboschi:
Correction to my previous comment: **no rebase is needed.** I checked the merge preview against current `main`, and git already resolves the overlap on its own — the `seed.ts` / `survey.ts` / `daemon.test.ts` hunks are byte-identical to what #3736 landed, so they collapse to nothing. The net effect of merging this as-is is exactly the reduced diff I described:

```
hindsight-all-npm/src/server.ts                     |  4 ++--
hindsight-all-npm/src/server.test.ts                | 26 +++++++++++++++++-
hindsight-integrations/coding-agents/.../daemon.ts  | 18 ++++++++++---
```

`daemon.ts` ends up with four distinct `windowsHide` sites (the `cargo` / `uvx` / `which` probes from this PR plus the detached spawn already on `main`) — no duplication. Merging.

Thanks for this one — the `hindsight-all-npm` half is what actually closes #3725.

PR #3750, opened 2026-08-24
docs: explain how entity resolution decides to merge a name

nicoloboschi:
## Why

A user asked why a retain that correctly extracted a new person, `Tigran`, stored the fact
on a pre-existing country entity, `Iran`. They had done the obvious homework:
`similarity('tigran','iran') = 0.2`, below pg_trgm's `0.3` default, neither string a
substring of the other — so by every documented number, this merge should not happen.

Nothing in the docs explains why it does. The thresholds are documented one env var at a
time (`ENTITY_TRGM_SIMILARITY_THRESHOLD`, `ENTITY_INTRABATCH_MERGE_SIMILARITY`,
`RETAIN_ENTITY_LOOKUP`, `..._MAX_CANDIDATES`), and the natural reading of that list is that
trigram similarity is what decides a match. It isn't:

- the trigram threshold (default **0.15**, not pg_trgm's 0.3) only decides which existing
  entities are **considered**;
- the merge itself is a separate weighted score — `SequenceMatcher` name ratio (≤0.5),
  co-occurrence overlap (≤0.3), recency (≤0.2), merged above **0.6** — with no minimum name
  similarity of its own.

For `tigran`/`iran` the two measures disagree sharply: 0.20 by trigram, **0.80** by sequence
ratio, i.e. 0.40 of the 0.6 threshold before any other signal. Add co-occurrence with an
entity like `user` — which co-occurs with nearly every fact on a mature bank — and it merges.
That's also why it doesn't reproduce on a fresh bank: no co-occurrence history to add.

## What changed

- **`configuration.md`** — new *How entity resolution decides* section: the two stages, the
  three scoring signals and their weights, the 0.6 cutoff, the fact that no minimum name
  similarity rejects a candidate, the short-name divergence between the two measures, and
  what to change in either direction (with the caveat that the trigram threshold is a
  Postgres session setting, so it is server-wide and not per-bank).
- **`configuration.md`** — `RETAIN_ENTITY_LOOKUP` described `full` as "exact match". It is
  exact-**or-substring**, and both strategies feed the same scoring pass. Since `full` would
  never have admitted `Iran` for `Tigran` in the first place, the choice changes which names
  can merge, not just how fast the lookup is. Row rewritten.
- **`retain.md`** — one paragraph noting resolution can absorb a new name into an existing
  entity, linking to the reference section. Kept impact-only, no internals.
- Regenerated `skills/hindsight-docs/references/**`.

## Scope

Documentation only — it describes today's behaviour on 0.9.1. The behaviour itself is
separately questionable (admission and decision use different metrics, and only the
admission one is tunable); that is being tracked on its own, not fixed here.

## Verification

- `npm run build` in `hindsight-docs` passes; the only broken-anchor warnings are
  pre-existing, in the 2026-04-28 blog post.
- `./scripts/generate-docs-skill.sh` re-run; the skill diff is exactly these two files.

### DIFF
--- a/hindsight-docs/docs/developer/configuration.md
+++ b/hindsight-docs/docs/developer/configuration.md
@@ -1394,7 +1394,7 @@ Controls the retain (memory ingestion) pipeline.
 | `HINDSIGHT_API_RETAIN_WALL_TIMEOUT` | Wall-clock ceiling in seconds for one retain task in the worker. A retain that blocks indefinitely (lock contention, an unreachable LLM endpoint) is cancelled and marked `failed` instead of holding its worker slot until the process restarts, so it can be retried. Set well above your slowest healthy retain; `0` disables. | `3600` |
 | `HINDSIGHT_API_RETAIN_BATCH_TOKENS` | Max characters per sub-batch for async retain auto-splitting | `10000` |
 | `HINDSIGHT_API_RETAIN_CHUNK_BATCH_SIZE` | Max chunks per streaming batch when retain ingests long documents. Each chunk produces roughly 17 facts, so the default 100 chunks ≈ 1700 facts per batch. Lower to cap memory/LLM pressure on large documents; raise for smaller chunks. Configurable per bank. | `100` |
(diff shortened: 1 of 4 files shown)


nicoloboschi:
Folded into #3752 — same commit, one PR instead of a stack. Nothing dropped: the docs commit is the first of the three there.

PR #3769, opened 2026-08-24
feat(coding-agents): add qwen-code as a hook harness

malzzz:
Adds `qwen-code` to `HOOK_HARNESSES`, with its transcript reader, installer, entry points, tests, docs, control-plane logo entry and E2E setup.

Qwen Code's hook protocol is a deliberate Claude Code superset — same stdin envelope (`session_id` / `transcript_path` / `cwd` / `hook_event_name`), same `hookSpecificOutput.additionalContext` + `systemMessage` output, same exit semantics, and a `settings.json` shape byte-for-byte what `cmdHook()` already emits. The spec is therefore `claude-code`'s with three deltas. **All three are same-name/different-meaning, so none is visible from the diff** — that's the part worth reviewing.

### 1. `timeout` is milliseconds here, seconds everywhere else

Qwen passes a command hook's `timeout` straight to `setTimeout` — *"Timeout in milliseconds, default 60000"*, from its own bundled `docs/features/hooks.md`. Installed values are `30000/30000/60000`.

Writing `30`/`60` would register 30 ms hooks — and **that misconfiguration looks harmless in testing**, because Qwen spawns without `detached: true` and kills only the direct child, so the orphaned work still completes. You cannot validate the unit by observing that retain works.

Rather than leave that to a comment, `HookHarnessSpec` gains `timeoutUnit`, and two lifecycle tests normalise through it: the prompt budget must exceed `HOOK_REFLECT_CAP_MS` (25 000), and the installed stop timeout must equal `hostTimeoutSec * 1000`. Changing `30_000` → `30` fails the first; dropping `timeoutUnit` fails the second. A bare threshold cannot do this — `>= 25_000` passes vacuously for the seven seconds-based harnesses, `>= 25` passes vacuously for Qwen.

### 2. `parse` reads `submitted_prompt`, not `prompt`

`UserPromptSubmit` also fires on tool-result continuations, where `prompt` holds model-bound tool output — keying on it would recall roughly 20× per user turn against tool results. `submitted_prompt` is attached only when the turn is both the first and a genuine `userQuery`, and `runHook`'s `if (!prompt) return` then suppresses continuations with no core change.

**Accepted cost:** `submitted_prompt` is the interactive TUI's projection, so headless (`qwen -p`), `serve`, SDK and ACP sessions seed and retain but never recall. The E2E declares `injectsIntoModel: false` for that reason — a *different* reason from `grok-build`'s passive hook, and it is spelled out in the setup so it doesn't read as the same limitation.

### 3. `type: "user"` is not a user turn

`provenance` is the discriminator. Across a 22-transcript corpus, synthetic records (`notification` 285, `cron` 21, `goal_runtime` 1) outnumber `real_user` ones (69) by **4.4 : 1** — without the gate, background-task notifications and cron chatter are retained as things the user said.

`isRealUser` rejects a present-but-malformed `provenance` outright rather than falling through to a subtype heuristic, and requires a positive subagent marker (`agentId` / `isSidechain`) before accepting an absent one — subagent transcripts use a third envelope carrying neither `provenance` nor `subtype`, so a strict `real_user` test would read them as 100% synthetic and retain nothing. Verified against all 11 real subagent transcripts: 11/11 still accepted.

### The injected-context echo

Qwen echoes injected context back into its own transcript inside `<qwen:user-prompt-submit-context>`, with the inner tags HTML-escaped — so the shared `stripInjectedMemory`, which matches raw tags, cannot see it. Measured: 55 injections in the corpus, all escaped, 0 raw.

The reader removes it using the two forms of pairing evidence Qwen's contract defines: `systemPayload.hookContext` present → use `systemPayload.displayText` (the host's own pre-hook projection, no tag matching at all); otherwise a **complete** tagged context in the **final** part after at least one other part.

It deliberately does **not** match the tag as a substring. The contract is explicit that the tag is *"a provenance marker, not … a general trust boundary"* and that consumers *"must not infer that arbitrary tag-like user text is hook provenance"* — so a prompt merely quoting the tag is preserved intact, which there's a test for.

## Two things outside the harness itself

**`registry.test.ts` gains the reverse parity check.** The existing one is directional (installer → registry), so a harness present in the registry but missing from `INSTALLERS` passes it while `install <name>` returns `unknown harness`. That is exactly how this change was briefly broken during development — the README advertised a command the installer rejected, and the suite was green.

**`readQwenTranscript` catches lazy-read faults.** `readJsonlTail` guards `statSync`/`openSync`, but the generator reads at iteration, and `runRetainHook` calls the reader *outside* `buildRetain`'s catch. A directory passed as `transcript_path` passes both guards and then throws `EISDIR` on the first `readSync`, rejecting the whole Stop hook. I've handled it at this reader to keep the change additive — **the underlying gap is in the shared `jsonl.ts` and affects every harness**, and I think it deserves its own fix rather than riding along here. Happy to open that separately.

## Verification

- `npm test` — 675 pass
- `tsc --noEmit` clean, `npm run build` clean, `build-skill.mjs --check` in sync
- Live corpus regression: 22 transcripts → 80 user / 1587 assistant / 2234 action turns, **0 injected-memory leaks**

One note for anyone reproducing: `HINDSIGHT_BANK_ID` must be unset in the environment, or `config.test.ts`'s *"missing files yield defaults"* fails on the leaked value.

Logo asset is from [homarr-labs/dashboard-icons](https://github.com/homarr-labs/dashboard-icons) (`svg/qwen.svg`, Apache-2.0). Happy to swap it for an official mark if you'd prefer.

Kimi Code is the only other unsupported harness I looked at; I'm deliberately **not** proposing it here, since its injection contract isn't in its published docs and it would need a third `configStyle` for TOML.

### DIFF
--- a/hindsight-integrations/coding-agents/README.md
+++ b/hindsight-integrations/coding-agents/README.md
@@ -93,6 +93,20 @@ npx @vectorize-io/hindsight-coding-agents install grok-build
 
 Native hooks and MCP in `~/.grok/config.toml`, plus the companion skill.
 
+#### <img src="https://hindsight.vectorize.io/img/harness/qwen-code.svg" alt="" width="20" height="20" /> Qwen Code
+
+```bash
+npx @vectorize-io/hindsight-coding-agents install qwen-code
+```
+
+Native hooks in `~/.qwen/settings.json`, plus MCP and the companion skill.
+
+> Qwen's hook `timeout` is in **milliseconds** (its own docs: "Timeout in milliseconds, default
+> 60000"), unlike every other supported agent, so the installed values are `30000/30000/60000`.
+> Recall fires on genuine submissions only — `UserPromptSubmit` also fires on tool-result
+> continuations, so interactive sessions recall once per prompt while headless (`qwen -p`),
+> `serve`, SDK and ACP sessions seed and retain but do not recall.
+
 #### <img src="https://hindsight.vectorize.io/img/harness/antigravity-cli.png" alt="" width="20" height="20" /> Antigravity CLI
 
 ```bash
--- a/hindsight-integrations/coding-agents/src/core/transcript-qwen.ts
+++ b/hindsight-integrations/coding-agents/src/core/transcript-qwen.ts
@@ -1,0 +1,248 @@
+/**
+ * Qwen Code session transcript (JSONL) reader — a HYBRID of the two schemas this package already
+ * parses, which is why it needs its own reader instead of reusing either.
+ *
+ * The ENVELOPE is Claude Code's, field for field: one record per line carrying
+ * `uuid`/`parentUuid`/`sessionId`/`timestamp`/`type`/`cwd`/`version`/`gitBranch`, stored at
+ * `~/.qwen/projects/<slugified-cwd>/chats/<sessionId>.jsonl` and handed to the Stop hook as
+ * `transcript_path`. The BODY is Gemini's: `message.parts[]` rather than `message.content[]`, and
+ * an assistant record's `message.role` reads **"model"** — so, exactly as in transcript.ts, role is
+ * driven from the record's own `type` and never from the redundant `message.role`.
+ *
+ * What we keep, normalized to the SAME `TransportTurn[]` shape as readClaudeTranscript so the live
+ * write-back (retainLiveSession) renders every harness identically:
+ *   - real user prompts
+ *   - assistant prose
+ *   - each `functionCall` part → a compact `role:"action"` turn (name + primary target, no args)
+ *
+ * What we drop:
+ *   - `type:"system"` records — UI telemetry, attribution/file-history snapshots, slash-command
+ *     invocations. Not conversation, and the BULK of the file: 5,385 of 9,962 records (54%) across
+ *     the live transcripts this reader was measured against.
+ *   - `type:"tool_result"` records (`functionResponse` parts) — tool OUTPUT, the mechanical noise
+ *     the shared `actionLine` convention exists to keep out of the bank. The matching CALL is a
+ *     `functionCall` part on the assistant record, so nothing is lost by dropping these whole.
+ *   - parts flagged `thought: true` — the model's reasoning, like Claude `thinking` and Codex
+ *     `reasoning`.
+ *   - SYNTHETIC user records. `type:"user"` is not a user turn in Qwen: background-task
+ *     notifications, cron re-entries and goal-runtime traffic are all written as `type:"user"`, and
+ *     they outnumbered genuine prompts 307 to 69 in the measured corpus. `provenance` is the
+ *     discriminator (see `isRealUser`) — Qwen's analogue of Claude's `isMeta` and dsh's
+ *     `source.kind === "user"`. Retaining them files machine scaffolding as the user's own words.
+ *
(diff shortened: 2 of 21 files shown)


hindsight-integrations/coding-agents/README.md
```diff
@@ -93,6 +93,20 @@ npx @vectorize-io/hindsight-coding-agents install grok-build
 
 Native hooks and MCP in `~/.grok/config.toml`, plus the companion skill.
 
+#### <img src="https://hindsight.vectorize.io/img/harness/qwen-code.svg" alt="" width="20" height="20" /> Qwen Code
```
Sanderhoff-alt, line 96: This README is the source for two generated artifacts, and neither was regenerated here, so the jobs that check them fail:

- `hindsight-docs/docs-integrations/coding-agents.md` — `node hindsight-docs/scripts/sync-coding-agents-doc.mjs --check` runs as part of `npm run build --workspace=hindsight-docs` and reports "docs page is out of date with the README" on this head (the new section sits under `## Install`, which is not in `DROP_SECTIONS`).
- `skills/hindsight-docs/references/sdks/integrations/coding-agents.md`, which `./scripts/generate-docs-skill.sh` derives from that page; `verify-generated-files` runs it before the no-uncommitted-changes check.

Run `node hindsight-docs/scripts/sync-coding-agents-doc.mjs` and `./scripts/generate-docs-skill.sh` and commit both. `skill/SKILL.md` is genuinely unaffected — the new section falls outside the `skill:begin`/`skill:end` regions, and `build-skill.mjs --check` passes.

hindsight-integrations/coding-agents/src/core/transcript-qwen.ts
```diff
@@ -0,0 +1,248 @@
...
+    // `type` is validated as "user" | "assistant" above; drive role from it (Qwen's assistant
+    // records carry the Gemini `message.role: "model"`, so the nested role is worse than redundant
+    // here). One record can yield a prose turn plus one action turn per tool call.
+    for (const rendered of renderLine(line.message.parts, line.type)) {
+      const turn: TransportTurn = { role: rendered.role, content: rendered.content };
+      if (typeof line.timestamp === "string") turn.timestamp = line.timestamp;
+      turns.push(turn);
+    }
+  }
+
```
Sanderhoff-alt: `prettier --write` reformats four files on this branch, which fails `verify-generated-files` (it runs `./scripts/hooks/lint.sh`, then fails on any resulting diff). This blank line before the closing brace is removed; the rest:

- `src/core/transcript-qwen.test.ts` around lines 131 and 304 — quote style on two string literals
- `src/harness/hook-lifecycle.test.ts` around lines 102 and 116 — the `expect(...)` calls wrap differently at 100 columns
- `src/installer.test.ts` around line 184 — the `JSON.parse(...)` call fits on one line

Running `./scripts/hooks/lint.sh` and committing the result clears all of them.

Sanderhoff-alt (commented):
The harness wiring itself checks out (registry, HOOK_BINS, installer, control-plane logo entry and the reverse parity test are all present, and the ms-vs-s timeout handling is covered by the two new lifecycle tests). Two CI-blocking items remain: a stale generated docs page and a prettier diff.

nicoloboschi:
superseded by #3979 thanks for your work!

Issue #3771, opened 2026-08-24
coding-agents: --import-conversations dies on a stray file in ~/.claude/projects instead of skipping it


feniix:
## Bug Description

`importLocalHistory` documents that it never throws (`hindsight-integrations/coding-agents/src/core/history.ts`), but the Claude reader does. A single regular **file** sitting where a project *directory* is expected under `~/.claude/projects` aborts the entire `--import-conversations` run instead of costing that one entry.

`claudeHistory` prefilters candidate project directories by encoded name and hands each match to `jsonlFiles`, which does `existsSync(dir)` — true for a file — and then `readdirSync(dir)`, which throws `ENOTDIR`. The caller, `importConversations` in `installer.ts`, has no `try/catch` of its own, so the throw escapes the install.

## Steps to Reproduce

```bash
# a normal Claude session for the repo, plus one stray file named like a project dir
touch ~/.claude/projects/-Users-x-dev-myrepo-sub
cd /Users/x/dev/myrepo
npx @vectorize-io/hindsight-coding-agents install claude-code --import-conversations
```

Equivalent as a test, against `importLocalHistory` directly:

```ts
writeFileSync(claudeProjectDir("/Users/x/dev/myrepo/sub", home), "not a folder");
importLocalHistory("claude-code", "/Users/x/dev/myrepo", home);
```

## Expected Behavior

The stray entry is skipped and the real sessions still import — the same way `dshHistory` already handles it ("a stray file where a project directory was expected") and `codexHistory` wraps its walk.

## Actual Behavior

`Error: ENOTDIR: not a directory, scandir '…/.claude/projects/-Users-x-dev-myrepo-sub'`, and nothing is imported.

Measured across all five file-based readers, each seeded with one real session and one stray entry:

| reader | no stray | with stray |
| --- | --- | --- |
| claude-code | 1 session | **THREW ENOTDIR** |
| codex | 1 session | 1 session |
| dsh | 1 session | 1 session |
| pi | 1 session | 1 session (fixed, see below) |
| prime-agent | 1 session | 1 session |

## Suggested Fix

Wrap the per-directory listing, exactly as the pi reader now does:

```ts
for (const dir of candidateDirs) {
  try {
    files.push(...jsonlFiles(join(root, dir)));
  } catch {
    // stray file where a project folder was expected — skip the entry, not the run
  }
}
```

## Notes

[...]

PR #3775, opened 2026-08-24
feat(coding-agents): first-class pi support, sharing one extension adapter and installer with Prime Agent

### LINKED ISSUE #3772: coding-agents: Prime Agent never gets the companion skill, though it discovers the same skill roots pi does
## Summary

Prime Agent documents the same skill-discovery roots pi has, but never receives the packaged companion skill.

## Detail

`installSkill` runs for 9 of the 13 harnesses — claude-code, codex, antigravity-cli, cursor-cli, copilot-cli, grok-build, cline-cli, dsh, and now pi. It does not run for opencode, kilo, devin-cli or **prime-agent**. This issue is only about Prime Agent; I have not checked whether the other three can load skills at all.

For Prime Agent it looks like an oversight rather than a decision, because its own docs list the same discovery roots pi has —

- `~/.prime/agent/skills/`
- `~/.agents/skills/`

— so it can load the skill, and today it only ever does so by accident, when the user also runs Codex or dsh and picks ours up from the shared `~/.agents` root.

## Why it is nearly free

pi and Prime Agent already share one installer factory (`piFamilyInstaller` in `src/installer.ts`), which grew an optional `skillsDir` when pi was wired up. Prime Agent needs the argument:

```ts
const primeAgent = piFamilyInstaller("prime-agent", [".prime", "agent"], [".prime", "agent", "skills"]);
```

One caveat worth keeping: install into the host's **own** directory, not the shared `~/.agents/skills`. `uninstallSkill` removes by a fixed directory name, so writing to the shared root would make `uninstall prime-agent` delete Codex's and dsh's copy. That is why pi installs into `~/.pi/agent/skills`.

## Current state

`installer.test.ts` has a test asserting prime-agent installs no skill, added so the gap reads as a decision rather than an accident. It should be inverted along with the change.

feniix:
Gives **pi** (`@earendil-works/pi-coding-agent`) first-class harness support, and refactors the Prime Agent installer and extension into a shared pi-family adapter that both hosts use.

pi could already *run* the extension on `main`: the `pi` key in our `package.json` names `dist/prime-agent.js`, and pi and Prime Agent share the same extension API, so `pi install npm:@vectorize-io/hindsight-coding-agents` gave pi working retain/recall. But everything it did was attributed to `prime-agent` — wrong `harnesses.<name>` config section, wrong `metadata.harness` and `harness:` tag on every retained document, no logo — and there was no `install pi`, no companion skill in pi's own skills directory, and no import of pi's past sessions. This PR adds all of that.

> Rescoped after review: this PR originally carried the #3772 fix (Prime Agent's companion skill). That landed on `main` separately in #3812 (`e04d02ee6`), which this branch has merged in, so the net change here is the pi harness plus the refactor. #3772 is no longer linked.

## What changes

pi is the agent Prime Agent was forked from, and the two expose the identical extension API — same `extensions` array in `settings.json`, same `before_agent_start` / `agent_end` events, same `registerTool`, same `pkg.pi` manifest key, same stored-session schema. So the existing Prime Agent adapter fits pi unchanged; it just had no seam to plug into.

- **`createPiExtension(harness)`** (`src/harness/pi-extension.ts`) — the shared extension adapter, mirroring `createPluginEntry` for the opencode family. `src/pi.ts` and `src/prime-agent.ts` become thin entrypoints, each guarded so the bundle reports its own harness name.
- **`piFamilyInstaller(harness, configDir, skillsDir)`** (`src/installer.ts`) — one factory for both hosts' `install`/`uninstall`: extension entry in the host's own `settings.json`, companion skill in the host's **own** skills directory (not the shared `~/.agents/skills`, because `uninstallSkill` removes by fixed name and would take Codex's and dsh's copy). Tests assert the two hosts stay independent — uninstalling pi does not disturb Prime Agent and vice versa.
- **`--import-conversations` for pi and Prime Agent** (`src/core/history.ts`, `src/core/transcript-pi.ts`) — reads both hosts' stored sessions, attributing by the header `cwd`. pi keeps one folder per working directory, so a name prefilter (`--<repo>--` plus anything nested under `--<repo>-`) keeps the scan cheap on large trees; Prime Agent's store is flat and reads every header.
- Registry, control-plane logo, docs page, README and SKILL.md regenerated, e2e Dockerfile for pi.

Both hosts still share one bank per repo (the default `coding-agent::{gitProject}` is harness-neutral on purpose), so a separate entrypoint per host is about attribution — `harnesses.<name>` config section, `metadata.harness` on retained documents, control-plane logo — not about splitting memory.

## Notes for reviewers

- **The `pi` manifest key stays pointed at Prime Agent's bundle.** Both hosts read the same `pkg.pi.extensions`, so it can only ever name one. The package route therefore still works for pi exactly as it did on `main` — and still mis-attributes to `prime-agent`. The README tells pi users to run `install pi` instead, which registers the bundle that reports pi's own harness.
- **`piHistory` never throws** (`importLocalHistory`'s contract; its caller has no catch). Both the sessions root and each session folder now go through `listDir`, so a stray regular file at either level costs that one entry, not the run. Post-merge the inner `try`/`catch` had become redundant with `listDir` and the root `readdirSync` was the one call still exposed — fixed in `b7f3f6906`, with a regression test that fails on the old code with `ENOTDIR`.
- **On `a5fd6f17e`** (the original #3772 commit): its Prime Agent `installSkill` call is now a no-op against `main`, but the rest survives the merge — `skillsDir` becoming a required parameter of the factory, the installer tests generalised over both hosts plus the cross-host independence test, and the README/docs wording. Since the branch is a merge (not a rebase) of `upstream/main`, dropping the commit would mean rewriting the merge; left in place and described here instead.

## Verification

- 692 tests pass; `tsc`, `./scripts/hooks/lint.sh` clean.
- Regression tests confirmed to fail without their fix (`ENOTDIR` for both history guards; `expected 'prime-agent' to be 'pi'` for the entrypoint guard).
- History readers validated against real stores: 202 pi sessions attributed to one repo with zero unattributed, in 760ms across a 34k-folder tree.


hindsight-integrations/coding-agents/src/installer.ts
```diff
@@ -268,43 +268,67 @@ const opencode: HarnessInstaller = {
...
+      c.log?.(`${harness}: extension entry + skill removed`);
+    },
+  };
+}
+
+const pi = piFamilyInstaller("pi", [".pi", "agent"], [".pi", "agent", "skills"]);
+const primeAgent = piFamilyInstaller(
+  "prime-agent",
+  [".prime", "agent"],
+  [".prime", "agent", "skills"]
```
Sanderhoff-alt, line 330: The description presents this argument as the #3772 fix (`// <- was missing`), but the merge base already installs Prime Agent's skill here: e04d02ee6 (#3812) added `installSkill(c, "prime-agent", join(c.home, ".prime", "agent", "skills"))` plus the matching `uninstallSkill`, and this branch merged that in as de2477fed. The removed lines in this same hunk show both calls. So the net behaviour change for Prime Agent is zero, and `Closes #3772` points at an issue that is already fixed on main.

That also removes the stated reason for carrying the five pi commits — the fix and its prerequisite are no longer coupled, because there is no fix left to carry. Could you rescope the title/body to what the PR now does (add pi as a harness, refactor the shared installer), drop the `Closes #3772` link, and drop `a5fd6f17e` if nothing of it survives the merge?

Sanderhoff-alt (commented):
Two things to sort out before this merges. The Prime Agent skill install the PR is framed around already landed on main in #3812, which this branch has merged in, so the diff's net change is the pi harness plus a refactor rather than the #3772 fix. The new pi history reader also carries a guard that the same upstream commit made redundant, while leaving the one readdir that can still throw unguarded.

hindsight-integrations/coding-agents/src/core/history.ts
```diff
@@ -290,6 +303,76 @@ function readDshLog(file: string): string[] {
...
+  const files: string[] = [];
+  for (const dir of readdirSync(root).filter((d) => d === exact || d.startsWith(nested))) {
+    try {
+      files.push(...jsonlFiles(join(root, dir)));
+    } catch {
+      // A stray FILE named like a session folder: existsSync accepts it and readdirSync then throws
+      // ENOTDIR. That must cost this one entry, not the run — importLocalHistory promises never to
+      // throw and its caller (installer.ts importConversations) has no catch of its own, so the
+      // whole --import-conversations would die on one junk file. Same guard dshHistory applies.
+    }
```
Sanderhoff-alt: Two issues with this guard after the upstream merge.

`jsonlFiles` now routes through `listDir`, which already swallows ENOTDIR (added by the same e04d02ee6, for #3771), so the inner `try`/`catch` no longer guards anything — the regression test above it passes with the block deleted.

Meanwhile `readdirSync(root)` on line 354 is itself unguarded, and that is the call that can still throw: a regular file at `~/.pi/agent/sessions` passes the `existsSync` check on line 346 and raises ENOTDIR out of `importLocalHistory`, which documents that it never throws and whose caller (`importConversations`) has no catch. That is exactly the failure this comment describes, one level up. `primeAgentHistory` just below is safe only because it goes through `jsonlFiles`.

Suggest `for (const dir of listDir(root).filter((d) => d === exact || d.startsWith(nested)))` and removing the `try`/`catch`.

hindsight-integrations/coding-agents/src/installer.ts
feniix: Agreed — rescoped. Title and body now describe the PR as the pi harness plus the shared adapter/installer refactor, and the `Closes #3772` link is gone (fixed on main by #3812).

On `a5fd6f17e`: the `installSkill` call is a no-op against main now, but the rest of the commit survives the merge (the factory's `skillsDir` becoming required, the installer tests generalised over both hosts + the cross-host independence test, README/docs wording). Since the branch merges `upstream/main` rather than rebasing, dropping it would mean rewriting the merge commit, so I left it in and called that out in the body. Happy to rebase and squash it away if you'd prefer a clean history.

hindsight-integrations/coding-agents/src/core/history.ts
feniix: Good catch on both counts. Fixed in b7f3f6906: the root scan now goes through `listDir` exactly as you suggested, and the redundant inner `try`/`catch` is gone. Added a regression test for the root case (a regular file at `~/.pi/agent/sessions`) — confirmed it fails on the old code with `ENOTDIR: not a directory, scandir …/.pi/agent/sessions` and passes now. The existing session-folder test stays as the one-level-down case.

nicoloboschi:
superseded by https://github.com/vectorize-io/hindsight/pull/3993

Issue #3756, opened 2026-08-24
Retain holds a whole document in memory: embeddings as list[float] are 74% of peak, and nothing bounds the pipeline by chunk count


nicoloboschi:
## Summary

Retaining a single large document holds the document's **entire** derived state in memory at once, so peak RSS scales with document size rather than with a bounded working set. A ~47 M-character document costs about **+708 MB** of anonymous memory in a worker; a few of them concurrently is enough to OOM a 4 GiB worker.

Three quarters of that is not the text — it is the embeddings, held as `list[float]`.

## Measurements

Worker RSS sampled from `/proc/1/status` (`RssAnon`) while ingesting one document of 47,280,119 characters at `RETAIN_CHUNK_SIZE=1500` → **31,520 chunks**:

```
idle baseline      ~1,722 MB
peak during retain ~2,430 MB      (+708 MB for one document)
```

Per-fact cost, measured with `sys.getsizeof` over the real field set of `ProcessedFact`:

| component | per fact | × 31,520 |
|---|---:|---:|
| **`embedding: list[float]` (384-dim)** | 12,472 B | **393 MB (74%)** |
| `fact_text` (1500 chars) | 1,549 B | 49 MB |
| `context` (1500 chars) | 1,549 B | 49 MB |
| `metadata` dict | 607 B | 19 MB |
| timestamps, ids, empty lists | 510 B | 16 MB |
| | | **528 MB** |

The remainder is the raw document string held during chunking (~47 MB) plus a second full copy of the text produced by `augment_texts_with_dates`, plus request/serialisation buffers.

A 384-dim embedding as a Python `list[float]` measures **12,472 bytes** against **1,616 bytes** for `array('f')` — a **7.7×** overhead, because every element is a separate boxed `PyFloat`.

## Cause

`_extract_and_embed` (`hindsight_api/engine/retain/orchestrator.py`) materialises everything for a content item before writing any of it:

```python
extracted_facts, chunks, usage = await fact_extraction.extract_facts_from_contents(...)  # ALL chunks
augmented_texts = embedding_processing.augment_texts_with_dates(extracted_facts, ...)    # second copy of the text
embeddings      = await embedding_processing.generate_embeddings_batch(..., augmented_texts)  # ALL embeddings
```

Oversized submissions are already split into sub-batches, but that split is over content **items**. A single large document is one item, so the sub-batch machinery never engages inside it and nothing bounds the peak.

## Proposed fixes

[...]

Issue #181, opened 2026-01-21
[Bug] CONCISE mode outputs Japanese/Chinese for English/Italian input due to prompt priming


fabioscarsi:
## Bug Description

When using Hindsight with `CONCISE` extraction mode (default) via the MCP server, fact extraction **sometimes** generates memories in **Japanese or Chinese** also if the input content is entirely in **English or Italian**. This causes severe memory system reliability issues as the stored memories become unreadable in the wrong language.

### Environment

- **Hindsight version**: 0.3.0 (PyPI, installed without Docker)
- **Platform**: MacBook Pro (Apple Silicon M1 Max, 32GB RAM), macOS (latest version)
- **Interface**: Hindsight MCP Server
- **Python**: 3.11
- **LLM**: GPT-5.2 (via OpenAI)
- **Embeddings**: BAAI/bge-m3 (multilingual)
- **Reranking**: BAAI/bge-reranker-v2-m3
- **Extraction mode**: `concise` (default - `HINDSIGHT_API_RETAIN_EXTRACTION_MODE` not set)
- **Target languages**: Italian and English

### Example of Malformed Memory

**Input Document (English):**
```
Complete CONCISE prompt analyzed. Full text of problematic language rule identified at line 444:

**Critical Finding:**
"LANGUAGE RULE (CRITICAL): Output facts in the EXACT SAME language as the input text. 
If input is Japanese, output Japanese. If input is Chinese, output Chinese. 
NEVER translate to English. Preserve original language completely."

**Prompt Engineering Anti-Pattern Identified:**
This instruction creates unintended priming by:
1. Mentioning Japanese and Chinese as specific examples
2. Not mentioning English or Italian (the actual target languages)
3. Implicitly suggesting these Asian languages are primary use cases

**How This Causes the Bug:**
When GPT-5.2 processes this prompt:
- The explicit Japanese/Chinese examples prime the model to consider these languages
- The instruction to "NEVER translate to English" may be misinterpreted as "English input should become something else"
- Technical terminology or certain characters might trigger false language detection
- The model then dutifully outputs in Japanese/Chinese to comply with the "preserve original language" rule

**Why VERBOSE Mode Might Work Better:**
The VERBOSE prompt (lines 546-548) uses: "LANGUAGE REQUIREMENT: Detect the language of the input text. 
All extracted facts, entity names, descriptions, and other output MUST be in the SAME language as the input. 
Do not translate to English if the input is in another language."

This is better because:
[...]

PR #3777, opened 2026-08-24
fix(retain, reflect): drop the source-language rule when an output language is set

### LINKED ISSUE #3776: HINDSIGHT_API_LLM_OUTPUT_LANGUAGE is a silent no-op for retain: the extraction prompt contradicts itself
## Summary

`HINDSIGHT_API_LLM_OUTPUT_LANGUAGE` is silently a no-op for **retain (fact extraction)**. It works for consolidation and reflect.

Retain's extraction prompt hardcodes an emphatic "never translate" rule near the top, and then `output_language_directive()` appends the opposite instruction at the end. The prompt contradicts itself, and the earlier, more forceful rule wins — so facts keep getting stored in the source language while the operator believes they are being normalized.

Consolidation already solves this exact problem by **dropping** its default rule when an output language is configured. Retain never got the same treatment.

## Why this is a bug and not intended behaviour

The intended contract is stated in three places, all of which explicitly name retain facts:

1. **Public docs** — `hindsight-docs/docs/developer/configuration.md:209`:
   > `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE` | When set, forces every LLM-generated artifact (**retain facts**, consolidation observations, reflect responses) into this language.

2. **Config field docstring** — `hindsight_api/config.py:2271-2273`:
   > When set, every LLM-generated artifact (**retain facts**, consolidation observations, reflect responses) is forced into this language regardless of the source content.

3. **Retain's own inline comment** — `fact_extraction.py:1270-1276`:
   > Force the LLM to emit fact text in the configured language, regardless of the source content's language. Same directive is applied to consolidation and reflect so `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE` has **a uniform effect across the pipeline**.

And the invariant it violates is written down in `consolidation/prompts.py:22-27`:
> Retain's fact extraction carries the equivalent rule (see `_BASE_FACT_EXTRACTION_PROMPT`), so this makes "preserve the source language" the pipeline-wide default. When an output language IS configured, this section is omitted and `output_language_directive()` takes over — **the two must never both be present or they contradict each other**.

Consolidation even has a dedicated regression test for precisely this property, `test_consolidation_directive_replaces_source_language_rule` (`tests/test_multilingual_bm25.py:145`):
> An explicit output language wins outright: the source-language default is dropped rather than left to contradict "translate everything into X".

There is no retain equivalent. That is the gap that let this ship.

## Reproduction (deterministic, no LLM required)

Against `main` @ `21fbb9f93`:

```python
import sys; sys.path.insert(0, "hindsight-api-slim")
from hindsight_api.engine.retain.fact_extraction import CONCISE_FACT_EXTRACTION_PROMPT
from hindsight_api.engine.prompt_utils import output_language_directive

prompt = CONCISE_FACT_EXTRACTION_PROMPT.format(retain_mission_section="")
prompt = prompt + output_language_directive("English")   # exactly what fact_extraction.py:1278 does

hard = "STRICTLY FORBIDDEN from translating or switching to any other language"
soft = "Translate any source content into English"
print("never-translate rule present:", hard in prompt)   # True  (position 221)
print("translate directive present:", soft in prompt)    # True  (position 6595)
```

Both are present. The "never translate" rule sits at position 221 — near the top, high salience — while the directive lands at 6595, after ~6.4KB of guidelines and immediately before the source text.

**All four extraction modes are affected** (the base template feeds concise/custom/verbatim, and verbose carries its own copy of the same rule):

```
concise   never-translate=True  translate=True  CONTRADICTION=True
verbose   never-translate=True  translate=True  CONTRADICTION=True
verbatim  never-translate=True  translate=True  CONTRADICTION=True
custom    never-translate=True  translate=True  CONTRADICTION=True
```

## Runtime reproduction

Observed on `ghcr.io/vectorize-io/hindsight:0.9.1`, provider `openai-codex`, model `gpt-5.6-luna`, with `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE=English`:

Input (Spanish):
> La decisión final fue que el proveedor `openai-codex` no soporta la API de lotes. Por eso el equipo fijó la variable en falso...

Stored fact — still Spanish:
> El proveedor `openai-codex` no admite la API de lotes, por lo que el equipo fijó la variable correspondiente en falso; habilitarla rompería el proceso de retención...

The `llm_requests` audit table confirms both instructions were in the same prompt. The model is not at fault — given a short prompt containing only the directive, the same model translates correctly.

After patching the hardcoded rule to mandate English, the same Spanish input produced:
> The team determined that the `openai-codex` provider does not support the batch API and set the related feature flag to false; enabling it would break the retention process rather than accelerate it.

Russian and Chinese inputs likewise produced English only after the patch.

In our banks this had accumulated ~817 non-English units out of 4,426 (~18.5%) across Spanish, Russian and Chinese, while the setting was configured and appeared to be in effect.

## Root cause

- `hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py:822` — hardcoded rule in `_BASE_FACT_EXTRACTION_PROMPT` (feeds concise / custom / verbatim)
- `hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py:975` — same rule duplicated in `VERBOSE_FACT_EXTRACTION_PROMPT`
- `hindsight-api-slim/hindsight_api/engine/retain/fact_extraction.py:1278` — `output_language_directive()` appended unconditionally, with no code path removing the rule above

## Suggested fix

Mirror what consolidation does at `consolidation/prompts.py:177`: make the hardcoded LANGUAGE block a `{language_section}` placeholder that is emitted only when `llm_output_language` is unset.

**Important:** the rule must be kept when the setting is unset. It exists for #181 (CONCISE mode drifting to Japanese/Chinese on English/Italian input), so removing it outright would regress that. This is specifically about making the two mutually exclusive, exactly as consolidation already does.

A retain counterpart to `test_consolidation_directive_replaces_source_language_rule` would guard it. The existing retain tests in `test_multilingual_bm25.py:63-107` only assert the directive is *present*; none assert the contradicting rule is *absent*.

## Impact

Silent. Nothing errors, nothing warns, and `/config` reports the setting as active — the memories are simply stored in the wrong language, and the divergence is only visible by inspecting the bank. Operators who set this for a multilingual corpus get a partially-normalized bank: consolidation observations honour it, the underlying facts do not.

## Environment

- Reproduced deterministically against `main` @ `21fbb9f93`
- Runtime behaviour observed on `0.9.1` (`ghcr.io/vectorize-io/hindsight:0.9.1`)
- The relevant lines are unchanged between `v0.9.1` and `main` apart from `getattr(config, "llm_output_language", None)` → `config.llm_output_language`

feniix:
Fixes #3776.

> **Scope note.** #3776 reports retain only and says reflect works. Reflect turned out to have the same contradiction in three places (below), so this PR fixes both pipelines. It is one bug — a hardcoded "keep the source language" rule that out-ranks the appended `output_language_directive()` — applied at every site that had it, through one shared helper.

## Problem

`HINDSIGHT_API_LLM_OUTPUT_LANGUAGE` is silently a no-op for retain. The extraction prompt hardcodes a "detect the input language and never translate" rule near the top, and `output_language_directive()` appends the opposite instruction at the end — the prompt contradicts itself.

The model resolves that in favour of the first rule: it lands ~6KB earlier (position 221 vs 6595 in the concise prompt) and is phrased far more forcefully — "STRICTLY FORBIDDEN from translating" against "Respond exclusively in X". Facts keep getting stored in the source language while `/config` reports the setting as active.

Consolidation already solves exactly this. `build_consolidation_system_prompt()` omits its default rule when a language is configured, `consolidation/prompts.py:22-27` writes down the invariant ("the two must never both be present or they contradict each other"), and `test_consolidation_directive_replaces_source_language_rule` guards it. Retain never got the same treatment.

## Fix — retain

Hoist the rule into `_DEFAULT_LANGUAGE_RULE` and emit it through a `{language_section}` placeholder, deferred through the same two-stage `.format()` the mode templates already use for `{retain_mission_section}`. The selection lives in one shared helper, `prompt_utils.default_language_section()`, which retain, consolidation and reflect all call.

All four extraction modes carried the rule (the base template feeds concise/custom/verbatim; verbose had its own copy) and all four are fixed:

| mode | unset → rule / directive | set → rule / directive |
|---|---|---|
| concise | ✅ present / absent | ✅ absent / present |
| verbose | ✅ present / absent | ✅ absent / present |
| verbatim | ✅ present / absent | ✅ absent / present |
| custom | ✅ present / absent | ✅ absent / present |

**The rule is deliberately kept when no output language is set.** It exists for #181 (CONCISE mode drifting to Japanese/Chinese on English/Italian input), so removing it outright would regress that. This makes the two mutually exclusive rather than deleting either.

**The unset path is byte-identical to before this change** in all four modes — verified by diffing the built prompts against `main`.

On the cacheable prefix: `test_retain_cacheable_prefix_invariant_to_per_bank_freetext` already documents output language as a low-cardinality structural toggle that *may* change the prefix and is correctly keyed by the cache fingerprint, so gating on it is in line with the existing invariant. High-cardinality free-text (mission, custom instructions) still cannot leak in — that test passes unchanged.

### Verification against a live deployment

Reproduced on `0.9.1` with provider `openai-codex` / `gpt-5.6-luna` and `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE=English`.

Spanish input:
> La decisión final fue que el proveedor `openai-codex` no soporta la API de lotes...

Before — stored fact still Spanish:
> El proveedor `openai-codex` no admite la API de lotes, por lo que el equipo fijó la variable correspondiente en falso...

After:
> The team determined that the `openai-codex` provider does not support the batch API and set the related feature flag to false; enabling it would break the retention process rather than accelerate it.

Russian and Chinese inputs likewise produced English only after the change. The `llm_requests` audit table confirmed both contradicting instructions were present in the same prompt beforehand.

## Fix — reflect (three sites, found in two rounds)

The issue assumed reflect was fine. It was not. Every place reflect stated a source-language rule did so unconditionally and deferred to "a directive above" that nothing ever put there:

| site | what wrote the answer | fix |
|---|---|---|
| `build_final_system_prompt` | forced-synthesis fallback (iteration cap, no tool call, context overflow) | rule via `default_language_section`; directive moved to the end of the **user** message (`build_final_prompt` / `build_reduce_prompt`) — see measurements below |
| `build_system_prompt_for_tools` | the `done()` tool call — **most reflect answers** (review, @Sanderhoff-alt) | rule hoisted to `_TOOLS_LANGUAGE_RULE`, via `default_language_section`; directive on the opening user message (`build_agent_user_prompt`) |
| `done` tool schema, `answer` description | read at the moment the answer is written, after everything else | `get_reflect_tools(llm_output_language=…)` strips the "SAME language" sentence and puts the directive on the tool description, so all three `done` variants (markdown / directive-aware / document) carry it |

The first round fixed only the first row. The LLM test then showed that even with the rule gone, "last in the system prompt" is not last — the question and the retrieved data arrive after it:

| reflect final synthesis | gemini-2.5-flash-lite | gemini-2.5-flash |
|---|---|---|
| directive last in the **system** prompt | 0/12 English | 0/5 English |
| directive last in the **user** message | 12/12 English | 5/5 English |

Retain needed no such move — its directive is followed from the system prompt. Extraction rewrites its input, where the instruction is the whole job; answering a question carries a strong prior to mirror the asker's language.

The review pointed out the second row: `build_final_system_prompt`'s sole caller is `_forced_final_synthesis`, so a normal run — ending in `done()` under `build_system_prompt_for_tools` — never saw the fix. Threading the setting there surfaced the third row, the tool schema's own rule, which is the genuinely-last thing the model reads on that path.

`default_language_section()`'s docstring now records the invariant: every prompt carrying such a rule selects it there, the `done` schema applies the same exclusion by hand, and nothing may append a rule of its own.

## Tests

Deterministic (`test_multilingual_bm25.py`, `test_reflect_prompt_builder.py`, `test_reflect_agent.py`, `test_llm_tools.py`):
- retain: `test_retain_directive_replaces_source_language_rule` (fails on all 4 modes without the fix) and `test_retain_unset_requires_source_language` (guards #181 — the rule must survive when unset).
- reflect: the same pair for `build_final_system_prompt`, for `build_system_prompt_for_tools` + `build_agent_user_prompt`, and for every `done` tool variant; plus a MockLLM run of `run_reflect_agent` to `done()` asserting the system prompt lacks the rule and the user message and tool schema carry the directive.

The existing retain tests only asserted the directive was *present*, never that the contradicting rule was *absent* — which is how this shipped.

Behavioural, `hs_llm_core` (`tests/test_retain_reflect_output_language.py`), mirroring `test_consolidation_output_language.py`, which had no retain or reflect counterpart. Chinese source material so an untranslated sentence is unmistakable to the judge; the language judgement is the judge's alone (a script gate would reject the proper nouns the criteria allow — review):
- `test_retain_configured_output_language_overrides_source_language` — real `extract_facts_from_text`.
- `test_reflect_done_path_configured_output_language_overrides_question_language` — real `run_reflect_agent` with mocked retrieval, asserting the answer did **not** come from forced synthesis (no `"final"` scope in `llm_trace`). Fails on the code before the second-round fix, passes after.
- `test_reflect_forced_synthesis_configured_output_language_overrides_question_language` — the exact system/user prompts and tool-less call `_forced_final_synthesis` makes.

Verified against `gemini-2.5-flash-lite` (the Core LLM job's model) and `gpt-4.1-mini` with Gemini as judge.

## Scope

Narrowed at review request — this PR is only the language fix, all prompt assembly. The #3756 retain-memory follow-ups originally stacked here (content-hash derivation, the sub-batch splitter, the screening cache, `_SubBatch.index`, `_flush()`, the benchmark imports) moved to #3787 and land on their own merits.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
@@ -109,6 +109,35 @@ def build_directives_reminder(directives: list[dict[str, Any]]) -> str:
     return "\n".join(parts)
 
 
+# The reasoning-loop language rule. Emitted only when no output language is configured
+# — see default_language_section(). Like _FINAL_LANGUAGE_RULE it defers to a directive
+# "above", and nothing ever put one there for a configured language: the rule stayed,
+# out-ranked the directive, and the done() answer came back in the question's language
+# on every run that never fell through to forced synthesis (#3776, review).
+_TOOLS_LANGUAGE_RULE = (
+    "## LANGUAGE RULE (default - directives take precedence)\n"
+    "- By default, detect the language of the user's question and respond in that SAME language.\n"
+    "- If the question is in Chinese, respond in Chinese. If in Japanese, respond in Japanese.\n"
+    "- IMPORTANT: The DIRECTIVES section above has HIGHER PRIORITY than this rule.\n"
+    "  If a directive specifies a language (e.g. 'Always respond in French'), follow the directive."
+)
+
+
+def build_agent_user_prompt(query: str, llm_output_language: str | None = None) -> str:
+    """The reasoning loop's opening user message: the question, closed by the directive.
+
+    The ``done()`` answer is written by the tool-calling model, whose system prompt is
+    :func:`build_system_prompt_for_tools`. For the same reason :func:`build_final_prompt`
+    carries the directive instead of :func:`build_final_system_prompt`, it goes on the user
+    message here rather than at the end of the system prompt: the question arrives after
+    the system prompt and out-ranks it. Tool results still follow over later turns, so
+    this is "last" for the question the model is answering, not for the whole
+    conversation — the system prompt dropping its contradicting rule is what makes the
+    directive uncontested (#3776).
+    """
+    return query + output_language_directive(llm_output_language)
+
+
 def build_system_prompt_for_tools(
     bank_profile: dict[str, Any],
     context: str | None = None,
@@ -117,6 +146,7 @@ def build_system_prompt_for_tools(
     include_observations: bool = True,
     budget: str | None = None,
     answer_as_document: bool = False,
+    llm_output_language: str | None = None,
 ) -> str:
     """
     Build the system prompt for tool-calling reflect agent.
@@ -138,6 +168,9 @@ def build_system_prompt_for_tools(
         has_mental_models: Whether the bank has any mental models (skip if not)
         include_observations: Whether search_observations is in the tool list.
         budget: Search depth budget - "low", "mid", or "high". Controls exploration thoroughness.
+        answer_as_document: Whether done() takes a structured document instead of markdown.
+        llm_output_language: Configured output language; drops the default language rule
+            (the directive itself goes on the user message, see build_agent_user_prompt).
     """
     name = bank_profile.get("name", "Assistant")
     mission = bank_profile.get("mission", "")
@@ -165,14 +198,15 @@ def build_system_prompt_for_tools(
         ]
     )
 
+    # Mutually exclusive with the configured output language — the rule is dropped
+    # outright rather than left to out-rank the directive (#3776). The directive itself
+    # is not appended here: it rides on the user message, see build_agent_user_prompt().
+    tools_language_section = default_language_section(_TOOLS_LANGUAGE_RULE, llm_output_language)
+    if tools_language_section:
+        parts.extend([tools_language_section.rstrip("\n"), ""])
+
     parts.extend(
         [
-            "## LANGUAGE RULE (default - directives take precedence)",
-            "- By default, detect the language of the user's question and respond in that SAME language.",
-            "- If the question is in Chinese, respond in Chinese. If in Japanese, respond in Japanese.",
-            "- IMPORTANT: The DIRECTIVES section above has HIGHER PRIORITY than this rule.",
-            "  If a directive specifies a language (e.g. 'Always respond in French'), follow the directive.",
-            "",
             "## CRITICAL RULES",
             "- ONLY use information from tool results - no external knowledge or guessing",
             "- You SHOULD synthesize, infer, and reason from the retrieved memories",
--- a/hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
@@ -11,6 +11,7 @@ import json
 from datetime import datetime, timezone
 from typing import Any
 
+from ..prompt_utils import default_language_section, escape_for_prompt, output_language_directive
 from .tokenization import count_prompt_tokens
 
 # Fraction of max_context_tokens reserved for tool results in the final synthesis prompt.
@@ -666,10 +667,15 @@ def build_final_prompt(
     additional_context: str | None = None,
     max_context_tokens: int = 100_000,
     max_tokens: int | None = None,
+    llm_output_language: str | None = None,
 ) -> str:
     """Build the final prompt when forcing a text response (no tools).
 
     ``max_tokens`` is the soft visible-length target (see ``_length_directive``).
+
+    ``llm_output_language`` closes the prompt with :func:`output_language_directive`.
+    It rides on the USER message, not the system prompt, because it has to be the last
+    thing the model reads — see :func:`build_final_system_prompt` for the measurement.
 
     Callers overflow-proof this via ``split_context_history``: when the whole
     history fits one chunk this renders it directly, and the per-block budget
@@ -711,7 +717,7 @@ def build_final_prompt(
     if length_directive is not None:
         parts.append(length_directive)
 
-    return "\n".join(parts)
+    return "\n".join(parts) + output_language_directive(llm_output_language)
 
 
 #: System prompt for the intermediate (map) calls of split synthesis. They do
@@ -757,6 +763,7 @@ def build_reduce_prompt(
     bank_profile: dict,
     additional_context: str | None = None,
     max_tokens: int | None = None,
+    llm_output_language: str | None = None,
 ) -> str:
     """Build the final prompt that synthesizes the answer from per-chunk claims.
 
@@ -766,6 +773,9 @@ def build_reduce_prompt(
     in different sections — that is why the claims carry ``mentioned_at``: the
     supersession rule (latest statement wins) must be applied across sections,
     not within one.
+
+    Carries the output-language directive for the same reason, and in the same place, as
+    :func:`build_final_prompt` — this is the other prompt that writes a user-visible answer.
     """
     parts = _bank_identity_section(bank_profile, additional_context)
 
@@ -792,7 +802,7 @@ def build_reduce_prompt(
     if length_directive is not None:
         parts.append(length_directive)
 
-    return "\n".join(parts)
+    return "\n".join(parts) + output_language_directive(llm_output_language)
 
 
 _FINAL_SYSTEM_PROMPT_BASE = """CRITICAL: You MUST ONLY use information from retrieved tool results. NEVER make up names, people, events, or entities.
@@ -834,11 +844,10 @@ CRITICAL: This is a NON-CONVERSATIONAL system. NEVER ask follow-up questions, of
 # demands a specific one (the cause of flaky multilingual reflect tests).
 #
 # Emitted only when no output language is configured — see default_language_section().
-# The escape hatch below defers to a directive "above", but output_language_directive()
-# is appended at the very END of the prompt, so it never triggered: with a configured
-# language the model saw this rule first, phrased more forcefully, and answered in the
-# question's language instead. Retain and consolidation drop their rule for exactly this
-# reason (#3776); reflect now does too.
+# The escape hatch below defers to a directive "above", but nothing ever put one there:
+# with a configured language the model saw this rule, phrased more forcefully, and
+# answered in the question's language instead. Retain and consolidation drop their rule
+# for exactly this reason (#3776); reflect does too.
 _FINAL_LANGUAGE_RULE = (
     "## LANGUAGE\n"
     "- Respond in the SAME language as the user's question "
@@ -858,17 +867,20 @@ def build_final_system_prompt(
     ``directives`` are re-injected here (they live in the agent/reasoning prompt,
     but the final answer is a separate call) so output-constraining rules — most
     visibly response language — are honoured by the model that actually writes
-    the answer. When ``llm_output_language`` is set it forces that language
-    regardless of the query/source/directive language (config override wins), and
-    the answer-in-the-question's-language default is dropped rather than left to
-    contradict it.
-    """
-    from hindsight_api.engine.prompt_utils import (
-        default_language_section,
-        escape_for_prompt,
-        output_language_directive,
-    )
+    the answer.
 
+    ``llm_output_language`` drops the answer-in-the-question's-language default rather
+    than leaving it to contradict the directive. It does NOT add the directive here.
+    That is deliberate and measured: this prompt is the system message, and the question
+    and the retrieved data both arrive *after* it in the user message. Appending the
+    directive at the end of this string still leaves it out-ranked by everything the
+    model reads next — a Chinese question with the output language set to English came
+    back in Chinese 12 times out of 12 on gemini-2.5-flash-lite, and 0/5 on
+    gemini-2.5-flash, with no contradicting rule anywhere in the prompt. Moving the same
+    directive to the end of the user message is 12/12 and 5/5 English. So
+    :func:`build_final_prompt` and :func:`build_reduce_prompt` carry it instead, where it
+    is genuinely the last thing the model reads (#3776).
+    """
     role_section = escape_for_prompt(mission.strip()) if mission else _DEFAULT_FINAL_ROLE
 
     parts = [build_directives_section(directives) if directives else ""]
@@ -879,7 +891,7 @@ def build_final_system_prompt(
     # remain a cacheable prefix and only this timestamp falls outside the cache.
     parts.append(_current_datetime_section())
 
-    return "\n\n".join(p.strip() for p in parts if p.strip()) + output_language_directive(llm_output_language)
+    return "\n\n".join(p.strip() for p in parts if p.strip())
 
 
 # Backward-compatible constant for non-identity missions
--- a/hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
@@ -832,6 +832,13 @@ CRITICAL: This is a NON-CONVERSATIONAL system. NEVER ask follow-up questions, of
 # be repeated for the answer-writing model. Without it, weaker models drift to
 # English even when the question/facts are in another language or a directive
 # demands a specific one (the cause of flaky multilingual reflect tests).
+#
+# Emitted only when no output language is configured — see default_language_section().
+# The escape hatch below defers to a directive "above", but output_language_directive()
+# is appended at the very END of the prompt, so it never triggered: with a configured
+# language the model saw this rule first, phrased more forcefully, and answered in the
+# question's language instead. Retain and consolidation drop their rule for exactly this
+# reason (#3776); reflect now does too.
 _FINAL_LANGUAGE_RULE = (
     "## LANGUAGE\n"
     "- Respond in the SAME language as the user's question "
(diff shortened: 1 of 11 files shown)


nicoloboschi:
Thanks for this — I verified the bug and the fix independently, and the fix itself is correct. Running the real `_build_extraction_prompt_and_schema()` on current `main`, all four extraction modes carry both the "STRICTLY FORBIDDEN from translating" rule and `output_language_directive()` at once; on this branch the rule is dropped when a language is set and kept when it isn't. Same for reflect, which turned out to have the same contradiction the issue thought it didn't. Good catch on that.

**But could you please scope this PR down to only what's needed to fix #3776?**

Right now it's 14 commits spanning two unrelated workstreams. Only these are the language fix:

- `067ed1ef` fix(retain): drop the source-language rule when an output language is set
- `703eef9d` refactor(prompts): select the language rule in one shared helper
- `8ac098df` refactor(retain): fill the extraction template from a scalar, not a dict
- `f4fdce56` fix(reflect): drop the source-language rule when an output language is set
- `9af43568` test(reflect): update the language-override test to the corrected contract

The rest are #3756 retain-memory follow-ups — content-hash derivation, the sub-batch splitter, the screening cache, `_SubBatch.index`, `_flush()`, the benchmark imports — touching `memory_engine.py`, `token_encoding.py`, `fact_storage.py`, `orchestrator.py` and `hindsight-dev/benchmarks/`. That's ~160 lines of churn in the retain hot path that a reviewer of a prompt-assembly fix has no reason to be reading, and it makes the change much harder to reason about or revert on its own. Please move them to a separate PR; they can land on their own merits.

Two things I noticed that support splitting:

1. `sha256_windowed` (`token_encoding.py:79`) arrives with **no production caller** — only tests import it, since `derive_document_content_hash` rolls its own digest loop. `check-unused-code` won't flag it (vulture is advisory, and it is referenced from tests). Either wire it up or drop it, in whichever PR carries it.

2. `prompt_utils.py:53-56` now contradicts the PR's own fix: *"Reflect does NOT use this helper... so it still carries the contradiction described above."* `f4fdce56` made reflect use it (`reflect/prompts.py:876`). Worth correcting — it's exactly the comment that would send the next reader off to re-fix reflect.

One heads-up on CI: this is a fork PR, and `test-api` is gated on `has_secrets == 'true'` (`.github/workflows/test.yml:1708`), so all three shards plus Core LLM and LLM acceptance **skipped**. The green run only covers build-imports, `check-unused-code` and `verify-generated-files` — none of the new tests actually ran. We'll run the full suite on our side before merging.

feniix:
Scoped down: the five commits you listed, rebased on current `main`. 6 files, +173/−50, all prompt assembly (`prompt_utils.py`, `consolidation/prompts.py`, `reflect/prompts.py`, `retain/fact_extraction.py`, two test files). `memory_engine.py`, `token_encoding.py`, `fact_storage.py`, `orchestrator.py` and `hindsight-dev/` are out of the diff.

The #3756 follow-ups are #3787. Splitting them surfaced that `4ea0e005` straddled both workstreams, which is where the stale docstring came from.

`sha256_windowed` is fixed in #3787. It had no caller because `derive_document_content_hash` sanitizes each window before hashing it, so it couldn't call `sha256_windowed(text)` and ran its own digest loop instead. It now takes a per-window `transform`; retain passes `_sanitize_text` and drops the loop, other callers pass nothing. The transform has to be context-free or the digest becomes a function of `TEXT_WINDOW_CHARS` and stops matching rows already written, so there's a test hashing several windows through one against transforming the whole string first.

`prompt_utils.py:53-56`: that sentence came from `4ea0e005`, which landed before `f4fdce56` made reflect use the helper. `4ea0e005` went to #3787 with the rest of the hash work, and the docstring here now says all three pipelines call it. It's folded into the reflect commit rather than added on top.

Also expanded the PR description, which never covered the reflect half.

On CI: 55 passed locally here, 54 on #3787, `ruff` and `ty` clean. Not a substitute for the shards, so the full suite is still yours to run.

feniix:
Added the `hs_llm_core` test for retain and reflect, mirroring `test_consolidation_output_language.py`.

It found that the reflect fix didn't work. With the contradicting rule gone and nothing left in the prompt disagreeing with the directive, a Chinese question with the output language set to English still came back in Chinese:

| reflect final synthesis | gemini-2.5-flash-lite | gemini-2.5-flash |
|---|---|---|
| directive last in the system prompt (what I pushed) | 0/12 English | 0/5 English |
| directive last in the user message (now) | 12/12 English | 5/5 English |

`build_final_system_prompt` appended the directive to the end of the system message, but the question and the retrieved data arrive after it in the user message. `build_final_prompt` and `build_reduce_prompt` now carry it instead. Both of them, because `_forced_final_synthesis` picks between the two on whether the retrieved data fits one chunk, so fixing only the first would leave the setting working below that threshold and not above it. It also takes a per-bank toggle out of the cacheable system prefix.

Retain is unaffected. I ran it separately rather than assuming: its directive is followed from the system prompt and the extraction test passes unchanged. I don't have an explanation for the difference that I've actually tested, so I've left one out.

Every deterministic test passed throughout, including the two I'd added — the prompt was correct and only its position was wrong. Four prompt-assembly tests encoded the old layout and are updated to pin the directive to the user prompt.

Verified on `gemini-2.5-flash-lite`, which is what the Core LLM job runs: 2 passed over four consecutive runs, plus the samples above. 412 passed / 7 skipped across the non-LLM reflect and prompt tests. Still needs your run of the shards.

Retain follow-ups and `sha256_windowed` are in #3787.

hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
```diff
@@ -851,22 +867,31 @@ def build_final_system_prompt(
...
+    is genuinely the last thing the model reads (#3776).
     """
-    from hindsight_api.engine.prompt_utils import escape_for_prompt, output_language_directive
-
     role_section = escape_for_prompt(mission.strip()) if mission else _DEFAULT_FINAL_ROLE
 
     parts = [build_directives_section(directives) if directives else ""]
     parts.append(_FINAL_SYSTEM_PROMPT_BASE.format(role_section=role_section))
-    parts.append(_FINAL_LANGUAGE_RULE)
+    parts.append(default_language_section(_FINAL_LANGUAGE_RULE, llm_output_language))
```
Sanderhoff-alt, line 922: Inside reflect, `llm_output_language` only ever reaches `build_final_system_prompt`, and its sole caller is `agent.py:717`, inside `_forced_final_synthesis`. The path that produces most reflect answers is the `done` tool call handled by `_process_done_tool` (`agent.py:979`), and its system prompt comes from `build_system_prompt_for_tools`, which is never passed `llm_output_language` and still emits an unconditional "detect the language of the user's question and respond in that SAME language" block (`prompts.py:169`).

So with `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE=English` a Chinese question still comes back in Chinese on any run that completes normally; the fix only takes effect when the agent falls through to forced synthesis (iteration cap, no tool call, or a context-overflow error). That also makes the invariant recorded in `prompt_utils.py:56` ("All three call this - nothing may append a source-language rule of its own") untrue of that block.

Suggested fix: thread `llm_output_language` into `build_system_prompt_for_tools`, route its language rule through `default_language_section`, and put the directive where it is read last in that path as well. The new judge test assembles the final prompts by hand, so it cannot catch this; a test that drives `run_reflect_agent` through to a `done` call would.

hindsight-api-slim/tests/test_retain_reflect_output_language.py
```diff
@@ -0,0 +1,158 @@
...
+        agent_name="TestUser",
+        config=config,
+    )
+
+    assert facts, "Should extract at least one fact from the Chinese source"
+    extracted = "\n".join(f"- {fact.fact}" for fact in facts)
+
+    # Decisive and script-based: any Chinese sentence left in the output means the
+    # contradicting rule won again, whatever the judge makes of the phrasing.
+    assert not _has_cjk(extracted), f"Facts were left in the source language: {extracted!r}"
```
Sanderhoff-alt: This assert rejects any ideograph, but the judge criteria immediately below explicitly permits proper nouns "left in their original script", and the source names are 张伟 / 腾讯 / 阿里云. A correct English extraction that keeps one of them verbatim - which consolidation's own rule encourages ("Proper nouns, identifiers, and units stay verbatim") - fails here while passing the judge, so the test will flake across models and runs. Line 142 has the same problem for the reflect answer.

Per the testing rule in CLAUDE.md, leave the language judgement to `assert_meets_criteria` and drop the script gate, or narrow it to CJK outside a known proper-noun allowlist so the two checks agree.

Sanderhoff-alt (commented):
The retain half looks right, and keeping the two rules mutually exclusive rather than deleting either is the correct shape. Two issues: the reflect half misses the path that produces most reflect answers, and the new judge tests carry a script assert that contradicts their own criteria.

hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
feniix: Right — `build_final_system_prompt`'s only caller was the fallback, so a normal run never saw the fix. Addressed in b868e4b1e and e0850948b:

- `build_system_prompt_for_tools` takes `llm_output_language`; its rule is hoisted to `_TOOLS_LANGUAGE_RULE` and selected through `default_language_section`, so it's dropped when a language is set.
- The directive goes on the opening user message (`build_agent_user_prompt`), for the same reason `build_final_prompt` carries it rather than the system prompt.
- Threading it there turned up one more site on this path: the `done` tool schema's `answer` description had its own unconditional "write in the SAME language as the user's question" — read at the moment the answer is written, and pointing at "a language directive in the system prompt" where the directive doesn't live. `get_reflect_tools(llm_output_language=…)` now strips that and puts the directive on the tool description, for all three `done` variants (markdown / directive-aware / document). That's the genuinely-last read on the done path.
- `prompt_utils.py`'s invariant comment updated to name both reflect prompts and the schema.

Tests: a MockLLM `run_reflect_agent` run to `done()` asserting system prompt / user message / tool schema (`test_reflect_agent.py`), schema tests over every `done` variant (`test_llm_tools.py`), and a real-LLM `hs_llm_core` test that drives `run_reflect_agent` through `done()` with mocked retrieval and asserts no `"final"` scope in `llm_trace` — it fails on the previous head and passes now.

hindsight-api-slim/tests/test_retain_reflect_output_language.py
feniix: Agreed — the gate contradicted the criteria right under it. Dropped in b52323b39: `_has_cjk` is gone from both tests and the language judgement is the judge's alone, per the CLAUDE.md rule. The module docstring now says why (the proper nouns 张伟 / 腾讯 / 阿里云 are allowed verbatim). The remaining CJK checks in the suite (`test_consolidation_output_language.py`, `test_multilingual.py`) assert CJK is *present*, which a kept proper noun can't break.

nicoloboschi (approved):
Approving. I verified the fix locally, since this is a fork PR and `test-api`, `Core LLM tests` and `LLM acceptance` all skipped in CI.

Both review points are properly closed. The done() path now drops its rule in `build_system_prompt_for_tools`, carries the directive on the user message, and strips the `done` schema's own "SAME language" across all three variants — that third site was found by threading the fix, not by the report. And the script gate is gone from the judge tests, per the CLAUDE.md testing rule.

The best part is the mid-PR finding: with the contradicting rule removed, reflect still answered in Chinese 0/12 because the directive sat at the end of the *system* prompt and the question arrives after it. Moving it onto `build_final_prompt` and `build_reduce_prompt` — both, since `_forced_final_synthesis` picks between them on chunk count — is 12/12. Measured, and the measurement is written down in the docstrings where the next reader will find it.

Local runs, all green: 94 passed across `test_llm_tools` / `test_reflect_prompt_builder` / `test_reflect_agent`; 37 in `test_multilingual_bm25`; 111 across nine wider reflect/prompt/retain files including the real-LLM split-synthesis ones; the 3 `hs_llm_core` tests in `test_retain_reflect_output_language`; `ruff` and `ty` clean.

One note for later, not a blocker: `test_retain_configured_output_language_overrides_source_language` passes on the unfixed code — I reverted the engine to the merge-base and ran it twice on gemini-2.5-flash-lite, passing both times. The model translates the Chinese source anyway despite the contradicting prompt, so that test spends Core LLM budget without guarding the retain bug on the model CI actually runs. The retain half is really guarded by the deterministic `test_retain_directive_replaces_source_language_rule` / `test_retain_unset_requires_source_language` over all four modes, which do fail on base. The two reflect judge tests are the ones carrying signal.

Thanks for the scope-down and for chasing the reflect half past the point where the prompt merely looked right.

PR #3778, opened 2026-08-24
fix(coding-agents): don't let a stray file at the dsh sessions root abort the history import

### LINKED ISSUE #3771: coding-agents: --import-conversations dies on a stray file in ~/.claude/projects instead of skipping it

feniix:
## Summary

Rebased onto main after #3812, which landed the `listDir` guard for the Claude reader plus its regression tests. What this branch still contributes is the dsh half of #3771:

- `dshHistory` guarded its root with `existsSync` — true for a regular **file** — and then called `readdirSync(root)` unguarded. A stray file at `~/.dsh/sessions` threw `ENOTDIR` out of `importLocalHistory`, which documents that it never throws, and out of `importConversations` in `installer.ts`, which has no catch of its own: the exact escape #3771 reports, for dsh instead of Claude.
- The root listing now goes through the module's `listDir`, and the per-project loop uses it too, so `dshHistory` drops its private try/catch. Side effect: the `existsSync` precheck and its TOCTOU window go away.

Closes #3771.

## Type of change

- [ ] Feature
- [x] Bug fix
- [ ] Hotfix
- [ ] Spike / exploration
- [ ] Documentation
- [ ] Refactor

## Test plan

- `src/core/history.test.ts` — new regression test seeds a regular file **at the sessions root**. Verified it fails with the reported error when `history.ts` is reverted to main (`Error: ENOTDIR: not a directory, scandir '…/.dsh/sessions'`) and passes with the change.
- Second test covers the per-project stray-file case, which had no `importLocalHistory` coverage for dsh. Both stub `$DSH_HOME` so they can't read a developer's real sessions.
- `npx vitest run src/core src/installer.test.ts` → 582 passed. `./scripts/hooks/lint.sh` clean.

## Notes for the reviewer

- Small behaviour change: an existing-but-empty sessions root now returns `supported: true, sessions: []` before the zstd capability check runs, instead of reporting "unsupported" on Node < 22.15. With nothing to decompress, that seemed the more honest answer; easy to reorder if you'd rather keep the old precedence.
- Left alone as out of scope: `codexHistory`'s walk is guarded at the opposite altitude (one unstattable entry discards everything found so far), `importLocalHistory` still has no top-level backstop for its never-throws contract, and `installer.ts` creates its tmpfile outside the try whose comment says a failed backfill must not fail the install.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/history.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/history.test.ts
@@ -1,12 +1,13 @@
 import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { claudeProjectDir, importLocalHistory } from "./history";
 
 let home: string;
 afterEach(() => {
   if (home) rmSync(home, { recursive: true, force: true });
+  vi.unstubAllEnvs();
 });
 
 function newHome(): string {
@@ -125,6 +126,41 @@ describe("local history import", () => {
     const r = importLocalHistory("codex", "/repo/mine", h);
     expect(r.sessions).toHaveLength(1);
     expect(JSON.stringify(r.sessions)).toContain("found me");
+  });
+
+  // Regression: dshHistory guarded its root with existsSync — true for a regular file — and then
+  // called readdirSync on it unguarded, so a stray file at ~/.dsh/sessions threw ENOTDIR out of
+  // importLocalHistory, which documents that it never throws, and whose caller does not catch.
+  it("treats a regular file at the dsh sessions root as no sessions, not a crash", () => {
+    const h = newHome();
+    // The reader prefers $DSH_HOME over the home it is handed; a developer who runs dsh would
+    // otherwise have this test read their real sessions.
+    vi.stubEnv("DSH_HOME", join(h, ".dsh"));
+    mkdirSync(join(h, ".dsh"), { recursive: true });
+    writeFileSync(join(h, ".dsh", "sessions"), "not a folder");
+
+    const r = importLocalHistory("dsh", "/Users/x/dev/myrepo", h);
+    expect(r.supported).toBe(true);
+    expect(r.sessions).toEqual([]);
+  });
+
+  it("reads dsh sessions past a stray file where a project directory was expected", () => {
+    const h = newHome();
+    const repo = "/Users/x/dev/myrepo";
+    vi.stubEnv("DSH_HOME", join(h, ".dsh"));
+    const root = join(h, ".dsh", "sessions");
+    mkdirSync(join(root, "myrepo", "s1"), { recursive: true });
+    writeFileSync(
+      join(root, "myrepo", "s1", "session.jsonl"),
+      `${JSON.stringify({ id: "s1", cwd: repo })}\n` +
+        `${JSON.stringify({ type: "user/message", time: Date.parse("2026-08-14T10:00:00Z"), data: { role: "user", content: [{ type: "text", text: "real work" }], source: { kind: "user" } } })}\n`
+    );
+    writeFileSync(join(root, "stray"), "not a folder");
+
+    const r = importLocalHistory("dsh", repo, h);
+    expect(r.supported).toBe(true);
+    expect(r.sessions).toHaveLength(1);
+    expect(JSON.stringify(r.sessions)).toContain("real work");
   });
 
   it("reports SQLite-backed harnesses as unsupported with a reason, not an empty success", () => {
--- a/hindsight-integrations/coding-agents/src/core/history.ts
+++ b/hindsight-integrations/coding-agents/src/core/history.ts
@@ -234,7 +234,10 @@ function dshHistory(repoDir: string, home: string): HistoryImport {
   const root = process.env.DSH_HOME
     ? join(process.env.DSH_HOME, "sessions")
     : join(home, ".dsh", "sessions");
-  if (!existsSync(root)) return { supported: true, sessions: [] };
+  // listDir, not existsSync: a regular file at the sessions root passes existsSync and then makes
+  // readdirSync throw ENOTDIR out of importLocalHistory, which promises never to throw.
+  const projectDirs = listDir(root).map((project) => join(root, project));
+  if (projectDirs.length === 0) return { supported: true, sessions: [] };
   if (typeof zlib.zstdDecompressSync !== "function") {
     return {
       supported: false,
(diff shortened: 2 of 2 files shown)


hindsight-integrations/coding-agents/src/core/history.test.ts
```diff
@@ -93,6 +94,25 @@ describe("local history import", () => {
     expect(JSON.stringify(r.sessions)).not.toContain("the other one");
   });
 
+  // Regression: the prefilter hands each matching NAME to readdirSync, so a regular file named like
+  // a project directory threw ENOTDIR out of importLocalHistory — which documents that it never
+  // throws, and whose caller does not catch. One junk file killed the whole --import-conversations
+  // run instead of costing that entry.
+  it("skips a stray file where a Claude project directory was expected, keeping the real sessions", () => {
```
Sanderhoff-alt: After #3812 (`fix(coding-agents): don't let one stray file abort the Claude import ...`) main already carries this case as `skips a regular file sitting where a Claude project directory was expected`, in the `a junk entry costs itself, not the import` describe block, plus a second one for a regular file at `~/.claude/projects` itself. This adds a third assertion of the same scenario under a different name.

The production changes the description attributes to this PR — `listDir` holding the guard for the module, `dshHistory` dropping its own try/catch, the `existsSync` precheck going away — are also not in the diff, because they are already on main. Please rebase so the diff shows what this branch still contributes, and drop the duplicate.

hindsight-integrations/coding-agents/src/core/history.test.ts
```diff
@@ -127,6 +147,29 @@ describe("local history import", () => {
...
+    // otherwise have this test read their real sessions.
+    vi.stubEnv("DSH_HOME", join(h, ".dsh"));
+    const root = join(h, ".dsh", "sessions");
+    mkdirSync(join(root, "myrepo", "s1"), { recursive: true });
+    writeFileSync(
+      join(root, "myrepo", "s1", "session.jsonl"),
+      `${JSON.stringify({ id: "s1", cwd: repo })}\n` +
+        `${JSON.stringify({ type: "user/message", time: Date.parse("2026-08-14T10:00:00Z"), data: { role: "user", content: [{ type: "text", text: "real work" }], source: { kind: "user" } } })}\n`
+    );
+    writeFileSync(join(root, "stray"), "not a folder");
```
Sanderhoff-alt, line 158: This puts the stray file *inside* `root`, which is the case `dshHistory` already handled before this PR: the inner `try { readdirSync(dir) } catch { continue }` at `src/core/history.ts:250-255` swallows it. So the test passes on unmodified main, as the description acknowledges, and it does not exercise the hole the description says was closed "one level up".

That hole is still open. `dshHistory` guards its root with `existsSync(root)` (`src/core/history.ts:237`), which is true for a regular file, and then calls `readdirSync(root)` unguarded at `src/core/history.ts:249`. A regular file at `~/.dsh/sessions` therefore still throws `ENOTDIR` out of `importLocalHistory`, which has no top-level catch (`src/core/history.ts:298-320`), and out of `importConversations`, which does not catch either (`src/installer.ts:1249`) — the exact escape #3771 reports, for dsh instead of Claude.

Suggested change: replace lines 237 and 249 with `listDir(root)` (dropping the now-redundant `existsSync`), and move the stray file in this test to `root` itself so it fails without that change.

Sanderhoff-alt (commented):
As it stands the diff is tests only — the `listDir` change this describes landed on main in #3812, which also added an equivalent Claude regression test, so the branch needs a rebase before it is clear what is left. The one claim in the description that is still unimplemented anywhere is the dsh half, and the new dsh test does not cover it. Details inline.

hindsight-integrations/coding-agents/src/core/history.test.ts
feniix: Rebased onto main and dropped this test — #3812 covers the case. The branch is now only the dsh change.

hindsight-integrations/coding-agents/src/core/history.test.ts
feniix: Done: `dshHistory` now lists its root through `listDir` (the `existsSync` precheck and the inner try/catch are gone), and this test moved the stray file to the root itself — it fails with `ENOTDIR … scandir …/.dsh/sessions` against main and passes with the change. Kept the per-project case as a second test since dsh had no `importLocalHistory` coverage at all.

nicoloboschi:
fixed by #3812

Issue #3709, opened 2026-08-21
Async operations orphaned permanently after container recreation (worker id defaults to hostname)


oantoshchenko:
**Bug Description**
The worker id defaults to the container hostname (`worker_id = args.worker_id or socket.gethostname()`). Startup recovery only reclaims rows `WHERE status = 'processing' AND worker_id = <own id>`, and the HTTP API refuses to retry or cancel rows in `processing`. In Docker, every `compose down`/`up` (restart, upgrade, backup/restore flows) mints a new container id, so any operation in flight at that moment stays `processing` forever under a worker id that will never return.

Observed effects on a single-container deployment: four consolidation operations sat in `processing` for 71+ hours, each owned by a different dead container id; the live worker counted them as other workers' active jobs, which starved the claimable queue (`claimable=N, assigned=0` indefinitely). Only a manual SQL reset (`SET status='pending', worker_id=NULL, claimed_at=NULL`) recovered them.

Workaround that fully fixes it for single-worker deployments: pin `HINDSIGHT_API_WORKER_ID` to a fixed value. Verified by SIGKILLing the container mid-task, the recreated container logged `recovered 1 stale tasks from previous run` and the operation completed.

**Steps to Reproduce**
1. Run the standard docker-compose single-container setup without `HINDSIGHT_API_WORKER_ID`.
2. Start a long consolidation or retain.
3. `docker compose down && docker compose up -d` while it is `processing`.
4. The row remains `processing` forever; API retry/cancel refuse it; the pending queue can starve.

**Expected Behavior**
A lease/heartbeat timeout that returns `processing` rows to `pending` when their worker has not checked in, independent of worker-id stability. Short of that: document `HINDSIGHT_API_WORKER_ID` as strongly recommended for containerized single-worker deployments, and/or allow the operations API to cancel/retry a `processing` row whose worker is no longer registered.

**Actual Behavior**
Orphaned `processing` rows persist indefinitely, hold phantom slots, and require manual database surgery.

**Version**: ghcr.io/vectorize-io/hindsight:0.9.1
**LLM Provider**: n/a (worker lifecycle)

koriyoshi2041:
The automatic lease/heartbeat gap is still real, but 0.9.1 already has two operator-side mitigations that may help recover this deployment without manual SQL:

- `hindsight-admin worker-status` lists processing rows by worker and age.
[...]

Issue #3594, opened 2026-08-18
Operations stuck in 'processing' after max retries exhausted (zombie operations)


seppaleinen:
## Description

Operations (all types — `refresh_mental_model`, `consolidation`, `retain`, etc.) get permanently stuck in `processing` status after exhausting retries. The error is logged and `retry_count` increments correctly, but the status never transitions to `failed`.

## Steps to Reproduce

1. Have running operations (e.g. `refresh_mental_model`, `consolidation`)
2. Restart `hindsight-api` or the LLM backend (e.g. llama-swap) while operations are in-flight
3. The in-flight operations lose their backend connection and hit the 300s reflect timeout
4. Operations retry (up to `HINDSIGHT_API_LLM_MAX_RETRIES` = 2, so 3 total attempts), each timing out at 300s
5. After all retries are exhausted, the operation remains in `processing` status with `retry_count: 3` and an `error_message` set

## Expected Behavior

After max retries are exhausted, the operation status should transition to `failed` (or `pending` for retry), regardless of error type. The operation should not remain indefinitely in `processing`.

## Actual Behavior

The operation is stranded in `processing` forever, holding a worker slot. It cannot be cancelled via the API (no endpoint for `processing` operations). The only recovery is a direct database UPDATE:

```sql
UPDATE async_operations SET status = 'failed', updated_at = now()
WHERE status = 'processing' AND retry_count >= <max_retries>;
```

## Evidence

Sample from `async_operations` table:

| Field | Value |
|-------|-------|
| `status` | `processing` (should be `failed`) |
| `retry_count` | `3` (max retries exhausted) |
| `error_message` | `TimeoutError: Reflect operation timed out after 300 seconds.` |
| `next_retry_at` | Past timestamp (no more retries scheduled) |
| `updated_at` | Time of last retry (stale — no progress) |

Observed on multiple operation types: `refresh_mental_model`, `consolidation`, `retain`.

## Impact

- **Worker slot exhaustion**: Stuck operations hold worker slots (`WORKER_MAX_SLOTS`), preventing new operations from being picked up
- **Cascading failures**: Multiple timed-out operations compound the slot starvation
- **Manual intervention required**: No API endpoint to cancel `processing` operations; requires direct DB access or `hindsight-admin decommission-workers`

## Environment

- Hindsight version: 0.9.0
- Deployment: Kubernetes (Helm)
- LLM backend: LiteLLM proxy → llama-swap (local GPU)
[...]

Issue #3448, opened 2026-08-13
Operations orphaned by a worker restart are unrecoverable through the API


Linodex:
### Bug Description

When the container restarts while operations are in flight, those rows remain at
`status='processing'` with the `worker_id` of a worker that no longer exists. Nothing
reclaims them on startup. Because consolidation is serialized per bank, one orphaned
consolidation blocks every future consolidation for that bank indefinitely.

### Steps to Reproduce

1. Start a consolidation on a bank so it enters `processing`.
2. Restart the container mid-flight (`docker compose restart`).
3. The row keeps `status='processing'` and the dead `worker_id`.
4. All later consolidations for that bank stay `pending` and are never claimed.

### Expected Behavior

A mid-flight restart is a routine event — the container is an appliance, and every
configuration change requires restarting it. Recovering from one should not require
database access.

Specifically:

1. **On startup, the worker reclaims operations it cannot possibly own.** Any row in
   `processing` whose `worker_id` is not a live worker should be requeued to `pending`
   (or moved to `failed` so the existing retry path applies). A worker ID that no longer
   exists is unambiguous evidence the claim is dead.
2. **The per-bank consolidation lock is never held by a nonexistent worker.** A bank
   should not be blocked indefinitely by a claim nobody holds.
3. **`POST /banks/{id}/consolidation/recover` recovers stale claims.** Its name implies
   exactly this case; today it only considers `failed` rows and returns
   `{"retried_count": 0}` against the state that actually needs recovering.
4. **Failing automatic reclamation, one supported manual escape exists.** Either
   `cancel` or `retry` should accept a `processing` row whose owning worker is gone —
   currently all four endpoints reject that state, leaving no API path at all.
5. **The stall is visible.** The worker already logs `[STUCK?]` and `[STUCK_STACK]` for
   these rows but takes no action and surfaces nothing through the API; operation status
   still reads `processing`, which is indistinguishable from healthy progress.

### Actual Behavior

| Endpoint | Result |
|---|---|
| `DELETE /operations/{id}` | 409 — "only `pending` operations can be cancelled" |
| `POST /operations/{id}/retry` | 409 — "expected `failed` or `cancelled`" |
| `DELETE /operations/{id}/delete` | terminal states only |
| `POST /banks/{id}/consolidation/recover` | `{"retried_count": 0}` |

`POST /banks/{id}/consolidate` returns `{"deduplicated": true}` and hands back the
already-queued operation, so it cannot force a fresh run either.

[...]

PR #3779, opened 2026-08-24
fix(worker): reclaim expired 'processing' leases from departed workers

sanastasiou:
## Problem

`_reclaim_own_processing_tasks` matches `worker_id = <own id>`, so startup recovery can only rescue rows *this* worker abandoned. But `worker_id` defaults to the container hostname:

```python
# hindsight_api/worker/main.py:225
worker_id = args.worker_id or socket.gethostname()
```

In Docker every `compose down`/`up` mints a new container id, so anything in flight at that moment is stranded in `processing` under an id that will never return — and no future startup can match it. `claimed_at` is already recorded on claim but is never read, so nothing else notices either.

Consolidation turns that from lossy into permanent. It is serialised per bank at claim time (`busy_bank_ids` in `claim_tasks`), so a single corpse in `processing` marks its bank busy forever and **no consolidation for that bank can ever be claimed again**. With the default `{"consolidation": 2}` reservation, those slots then sit idle indefinitely.

## Evidence from production

Single-container deployment, v0.9.1. Two consolidations stranded on 2026-07-05 were still `processing` **50 days later**:

```
Worker: 5def1aaf9f8a (2 task(s))
  46e1a986  consolidation  bank=skynet-brain     running=50 days, 7:37:18
  260cf73c  consolidation  bank=stefanos-family  running=50 days, 7:32:12
```

The worker had been at half capacity that entire time — two slots reserved for work that could never be claimed, while retain queued behind the two shared slots:

```
slots=2/4 | reserved: [consolidation=0/2(avail=2)] | shared=2/2(avail=0)
```

`hindsight-admin worker-status` showed **six** departed container ids holding nine rows in total — one per historical redeploy. `hindsight-admin decommission-worker <id>` cleared them and slots went to `4/4` immediately:

```
slots=4/4 | reserved: [consolidation=2/2(avail=0)] | shared=2/2(avail=0)
```

So the recovery primitive already exists and works — it just isn't reachable automatically, and an operator has no reason to look until something is visibly broken. Nothing alerts: the API stays healthy, the queue keeps moving on the shared slots, and the only symptom is throughput quietly halving.

Worth noting the API can't help either — both affordances refuse `processing` rows:

```
409  cannot be retried: status is 'processing', expected 'failed' or 'cancelled'
409  cannot be deleted: status is 'processing', expected 'failed', 'cancelled' or 'completed'
```

## Change

A startup sweep alongside the existing own-rows recovery: reclaim `processing` rows whose `claimed_at` is older than `HINDSIGHT_API_WORKER_ORPHAN_LEASE_SECONDS` (default `86400`), regardless of owner.

Guards:

- **This worker's own rows are excluded** — the caller directly above already handles them, with retry/fail accounting.
- **Batch API rows are excluded**, for the same reason as the own-rows path: long-lived by design.
- **Only expired leases** — 24 h default is far beyond any real task, and the rows above were 50 days old. A live worker's fresh row is never touched.
- **`0` disables the sweep** entirely.

The log line names the departed owners and points at `HINDSIGHT_API_WORKER_ID`, so operators fix the root cause rather than lean on the sweep:

```
Worker live-worker reclaimed 9 operation(s) abandoned by 6 departed worker(s) ...
Set HINDSIGHT_API_WORKER_ID to a stable value so worker ids survive container
recreation and this is not needed.
```

## Relationship to existing issues

- **#3448** fixed this class for the case where the worker id is *stable*. This covers the case where it is not — which, with the hostname default, is every Docker restart.
- **#3709** describes this exact gap (closed `not_planned`).
- **#3594** (zombie operations after retries exhausted) is the same shape from a different entry point.
- **#3720** (detect stuck tasks by stage progress) is a finer-grained detector; a lease bound is the coarse backstop underneath it.

I'd also gently suggest `warn_if_container_default_worker_id` deserves to be louder. It fired on every one of our boots for 50 days and we sailed past it — a warning that predicts a 50 % capacity loss is arguably an error, or at least worth repeating periodically rather than once at startup.

## Tests

Three cases added to `TestWorkerRecovery`, following the existing `test_recover_own_tasks_does_not_affect_other_workers` pattern:

- `test_expired_lease_from_departed_worker_is_reclaimed`
- `test_fresh_lease_from_another_worker_is_left_alone`
- `test_orphan_sweep_is_disabled_when_lease_is_zero`

`tests/test_worker.py`: **110 passed**, `ruff check` clean. (2 unrelated setup errors locally — missing optional `sentence_transformers` in a minimal env.)

---

## Update — heartbeat-aware liveness (`7d96b38`)

The Strix review caught a real race in the first version, and I've taken the fix.

The sweep originally judged abandonment on `claimed_at` alone. That column is written once at claim time and never refreshed — but a long-running job *does* emit progress heartbeats, and those land on `updated_at` (`MemoryEngine._write_operation_progress`, `memory_engine.py:3579`). Combined with consolidation having no wall-clock ceiling (`_wall_timeout_for` returns `None` for everything except retain), a healthy consolidation can legitimately outlive a 24 h lease — and resetting it to `pending` mid-flight would let a second worker claim the same bank and run it concurrently.

The sweep now requires **both** timestamps to be stale. A job that is merely slow is never touched; only one that is old *and* silent.

Worth noting: this change made my own `test_expired_lease_from_departed_worker_is_reclaimed` fail, and that was the test being unfaithful rather than the fix being wrong. It set a stale `claimed_at` but let `updated_at` default to `now()`. Real abandoned rows have both stale — the two production corpses had `created == updated == 2026-07-05T14:31:37`, never touched again, because no heartbeat ever ran. The fixture now matches.

New test: `test_slow_but_live_job_with_fresh_heartbeat_is_not_reclaimed` — 50-day-old claim, heartbeat seconds ago, different worker → must not be reclaimed.

`tests/test_worker.py`: **111 passed**, ruff clean.

Two adjacent points from the same review — unfenced terminal writes (`_mark_completed` matches `operation_id` + `status` with no `worker_id` guard) and unbounded consolidation — I agree with, but both are pre-existing and independent of this change, so they're left out to keep the diff reviewable. Happy to fold either in if maintainers prefer.


sanastasiou:
Good catch — this is a real race and I've taken the suggested fix. Pushed in `7d96b38`.

I verified both premises in the source before accepting rather than taking them on trust:

- `MemoryEngine._write_operation_progress` (`memory_engine.py:3579`) does `SET result_metadata = ... , updated_at = now()`. Confirmed: the heartbeat lands on `updated_at`, never on `claimed_at`.
- `_wall_timeout_for` (`poller.py:65`) returns a timeout only for `_RETAIN_OP_TYPES` and `None` for everything else. Confirmed: consolidation is genuinely unbounded, so it can outlive a 24 h lease legitimately.

So the sweep now requires **both** timestamps to be stale — a job that is merely slow is never touched, only one that is old *and* silent.

One detail worth surfacing: my existing `test_expired_lease_from_departed_worker_is_reclaimed` **started failing** with this change, and that turned out to be the test being unfaithful rather than the fix being wrong. It inserted a stale `claimed_at` but let `updated_at` default to `now()` — so under the new predicate it correctly reclaimed nothing. Real abandoned rows have both stale: the two production corpses that motivated this PR had `created == updated == 2026-07-05T14:31:37`, never touched again, because no progress heartbeat ever ran. The fixture now sets both, which is what the wild actually looks like.

Added `test_slow_but_live_job_with_fresh_heartbeat_is_not_reclaimed` to pin the race directly: 50-day-old claim, heartbeat seconds ago, different worker → must not be reclaimed. `tests/test_worker.py`: 111 passed, ruff clean.

On the two adjacent points in the report — unfenced terminal writes (`_mark_completed` matching `operation_id` + `status` with no `worker_id` guard) and consolidation having no wall-clock ceiling — I agree with both, but they're pre-existing and independent of this change, so I've left them out to keep the diff reviewable. Happy to fold either into this PR or open separate ones, whichever maintainers prefer.

koriyoshi2041 (commented):
One small correctness issue in the diagnostic receipt: PostgreSQL `UPDATE ... RETURNING` exposes the updated tuple. Because this statement sets `worker_id = NULL` before `RETURNING operation_id, worker_id, operation_type`, every returned `worker_id` is `NULL`, so `owners = sorted({str(r["worker_id"]) ...})` will log `owners: None` rather than the departed container IDs.

The reclaim behavior itself still looks sound, but this loses the exact operator signal the warning promises. A regression could assert that `caplog` contains `dead-container-abc123`. One repair is a candidate CTE that selects/locks the stale rows and preserves their old owner IDs, followed by `UPDATE ... FROM candidates ... RETURNING candidates.worker_id`; alternatively, return only fields whose post-update values are intended.

sanastasiou:
Good catch — confirmed and fixed in `ced4ee0`.

You're right about the mechanism: `RETURNING` yields the post-update tuple, and the same statement sets `worker_id = NULL`, so the receipt could never have named an owner. Reproduced it against the old code before changing anything — the log line was literally:

```
Worker live-worker reclaimed 1 operation(s) abandoned by 1 departed worker(s)
in schema None (lease 86400s; owners: None). Set HINDSIGHT_API_WORKER_ID ...
```

which is the worst version of this bug: accurate, useless, and indistinguishable from the sweep having matched a row with a genuinely NULL owner.

Took the candidate-CTE option you suggested:

```sql
WITH candidates AS (
    SELECT operation_id, worker_id
    FROM {table}
    WHERE status = 'processing'
      ...
    FOR UPDATE SKIP LOCKED
)
UPDATE {table} AS t
SET status = 'pending', worker_id = NULL, claimed_at = NULL, updated_at = now()
FROM candidates c
WHERE t.operation_id = c.operation_id
RETURNING c.operation_id, c.worker_id AS previous_worker_id, t.operation_type
```

Predicates and reclaim behaviour are unchanged; only the reported owner moves.

One addition beyond the repair: the CTE takes `FOR UPDATE SKIP LOCKED`. This runs on the startup path, so a rolling restart can have two workers sweeping at once — without it one blocks on the other's row locks for the duration of that transaction. With `SKIP LOCKED` they reclaim disjoint sets and neither waits, and a row locked by a live worker is skipped, which is the reading we want anyway.

Regression test is the one you described — asserts `dead-container-abc123` appears in the warning and that `owners: None` does not. I mutation-checked it by reverting the SQL to the old `RETURNING worker_id`; the test fails there and passes on the fix, so it's actually pinning the behaviour rather than the implementation. Full worker suite: 112 passed (2 unrelated setup errors from `sentence-transformers` missing in my env).

koriyoshi2041 (commented):
Re-checked the diagnostic repair at `ced4ee0`: the candidate CTE preserves the pre-update owner, keeps the original stale-row predicates, and `FOR UPDATE SKIP LOCKED` makes concurrent startup sweeps select disjoint rows. The regression now observes the operator-facing contract directly (`dead-container-abc123` present, `owners: None` absent), so the issue I raised is resolved at this head. I did not re-evaluate the broader lease/fencing design in this follow-up.

Sanderhoff-alt (commented):
The heartbeat-aware liveness check reads correctly, and I confirmed `_write_operation_progress` does bump `updated_at`, so the "old AND silent" predicate is real. Three things I think need changing before merge: the statement is PostgreSQL-only in a file that is shared with Oracle, the reset bypasses the retry cap that the sibling reclaim path exists to enforce, and the new flag is missing from the env template and docs.

hindsight-api-slim/hindsight_api/worker/poller.py
```diff
@@ -1158,6 +1162,105 @@ async def _reclaim_own_processing_tasks(self, schema: str | None, *, operation_i
...
+            # The departed owner is read in a CTE, BEFORE the update. A plain
+            # `UPDATE ... SET worker_id = NULL ... RETURNING worker_id` returns
+            # the POST-update tuple, so every returned owner would be NULL and
+            # the warning below — whose whole purpose is naming the ids to pin
+            # via HINDSIGHT_API_WORKER_ID — would print "owners: None".
+            # SKIP LOCKED so two workers starting together each reclaim a
+            # disjoint set instead of one blocking on the other's row locks.
+            rows = await conn.fetch(
+                f"""
+                WITH candidates AS (
```
Sanderhoff-alt: This statement is PostgreSQL-only, but the poller is shared with Oracle (module docstring: "PostgreSQL via asyncpg, Oracle via oracledb"; the breakdown query further down deliberately uses `SUM(CASE WHEN ...)` instead of `COUNT(*) FILTER` for that reason). `_rewrite_pg_to_oracle` in `hindsight_api/engine/db/oracle.py` handles `now()`, `interval 'N unit'` literals and the `FOR UPDATE` variants, but it has no rule for `make_interval(secs => ...)`, no rule for `UPDATE ... FROM`, and Oracle does not accept a locking clause inside a `WITH` subquery.

So on Oracle this raises on every call. The failure is not visible either: the call sits inside the same per-schema `try` in `recover_own_tasks` and *before* `_reconcile_orphaned_parents(schema)`, so the exception is logged as a warning and orphaned batch_retain parents (#2985) silently stop being reconciled at every worker startup, for every schema.

Either route the statement through the dialect-specific ops layer the way `claim_tasks` does (`ops_postgresql.py` / `ops_oracle.py`), or restrict it to constructs the rewriter already covers. In both cases please also move the call after `_reconcile_orphaned_parents` so a failure in the new sweep cannot take that step down with it.

hindsight-api-slim/hindsight_api/worker/poller.py
```diff
@@ -1158,6 +1162,105 @@ async def _reclaim_own_processing_tasks(self, schema: str | None, *, operation_i
...
+                      AND worker_id IS NOT NULL
+                      AND worker_id <> $1
+                      AND claimed_at IS NOT NULL
+                      AND claimed_at < now() - make_interval(secs => $2)
+                      AND updated_at < now() - make_interval(secs => $2)
+                      AND result_metadata->>'batch_id' IS NULL
+                    FOR UPDATE SKIP LOCKED
+                )
+                UPDATE {table} AS t
+                SET status = 'pending', worker_id = NULL, claimed_at = NULL,
```
Sanderhoff-alt, line 1267: The reset leaves `retry_count` untouched and applies no ceiling, unlike `_reclaim_own_processing_tasks` directly above, whose two-statement shape exists specifically so that "a task that reliably kills its worker cannot be re-claimed forever (claim -> grind -> die -> reclaim -> ...)" (#2675/#2834).

That cap only matches `worker_id = <own id>`, so in the rotating-hostname deployments this PR targets it can never fire: a task that OOM-kills its container is stranded under an id that never returns, this sweep hands it back to `pending` a lease later with `retry_count` unchanged, it is claimed, it kills the worker again, and the cycle repeats with no bound and no terminal `failed` row. The 24 h lease slows the loop down; it does not bound it.

Please increment `retry_count` here and mirror the at/over-limit branch that moves the row to `failed`, so both reclaim paths share the same budget.

hindsight-api-slim/hindsight_api/config.py
```diff
@@ -741,6 +741,7 @@ def _parse_boolean_env(env_name: str, default: bool) -> bool:
 ENV_WORKER_TASK_RETRY_BACKOFF_SECONDS = "HINDSIGHT_API_WORKER_TASK_RETRY_BACKOFF_SECONDS"
 ENV_WORKER_HTTP_PORT = "HINDSIGHT_API_WORKER_HTTP_PORT"
 ENV_WORKER_MAX_SLOTS = "HINDSIGHT_API_WORKER_MAX_SLOTS"
+ENV_WORKER_ORPHAN_LEASE_SECONDS = "HINDSIGHT_API_WORKER_ORPHAN_LEASE_SECONDS"
```
Sanderhoff-alt, line 744: Per CLAUDE.md, "Adding New API Configuration Flags", a new env-backed field also needs an entry in `.env.example` (then re-copied to `hindsight-embed/hindsight_embed/env.example`, which the sync test compares byte-for-byte) and a row in `hindsight-docs/docs/developer/configuration.md`. Neither is in this diff, so the only description of a flag that decides when one worker may take over another's in-flight work lives in the source.

sanastasiou:
All three are real — I verified each against the source before changing anything, and two of them were bugs I would have shipped. Fixed in `ddc3f2d`.

**1. Oracle.** Confirmed: the module docstring does say "PostgreSQL via asyncpg, Oracle via oracledb", `_rewrite_pg_to_oracle` has no rule for `make_interval(secs => ...)` or `UPDATE ... FROM`, and the call sat at line 1296 inside the same per-schema `try`, *before* `_reconcile_orphaned_parents` at 1303. So on Oracle it raised every call, the `except` logged it as a warning, and #2985 reconciliation silently stopped running at every startup for every schema. That is worse than the sweep not working.

I took the second route you offered — restricting to constructs the rewriter already covers — rather than adding an ops-layer pair, because the sibling `_reclaim_own_processing_tasks` runs on Oracle today using exactly `->>`, `RETURNING`, `now()` and `COALESCE`, so mirroring it is dialect-safe by construction rather than by my reading of the rewriter. The lease cutoff is now computed in Python and bound as a parameter, so there is no interval syntax at all. The CTE and the lock are gone: with the two-statement shape the second `UPDATE` cannot match rows the first already moved (`status` is no longer `processing`), so concurrent workers converge without `SKIP LOCKED`. The call is also moved after `_reconcile_orphaned_parents`, so a failure in this newer optional sweep can never take that established step down again.

One thing your review made me catch in my own fix: I first wrote the failure reason as two adjacent SQL string literals across a newline. PostgreSQL concatenates those; Oracle does not — I would have reintroduced the same class of bug in the fix for it. It is a single literal now.

**2. Retry budget.** Confirmed and this was the sharpest one. The cap matches `worker_id = $1`, so in the rotating-hostname deployments this sweep exists for it is structurally unreachable — the lease slowed the crash loop but never bounded it, and no terminal `failed` row was ever written. Now mirrors the sibling exactly: increment `retry_count` under the cap, move to `failed` at/over it with a reason, and roll the terminal state up to the parent aggregator via `_maybe_update_parent_operation` (one transaction per child, same as the own-rows path).

**3. Config surface.** Added to `.env.example`, re-copied byte-for-byte to `hindsight-embed/hindsight_embed/env.example`, and added a row to `configuration.md`, per CLAUDE.md steps 5 and 6.

`ruff` is clean. I could not execute `test_bundled_template_matches_repo_root` here (the embed suite needs `rich`, which is not installable in this environment), so rather than claim it passes I verified the equality it asserts directly — both files 30990 bytes, identical, with a negative control to confirm the comparison can actually fail.

nicoloboschi:
set stable worker id via env. https://hindsight.vectorize.io/faq#what-are-zombie-operations-and-how-do-i-recover-them

PR #3781, opened 2026-08-25
fix(metrics): reset global metrics collector during teardown to avoid test pollution

### LINKED ISSUE #3780: Leaked global metrics collector makes provider tests fail depending on xdist worker assignment
## Summary

`create_metrics_collector()` permanently replaces a process-global collector and nothing ever restores it. Any test that starts the FastAPI app therefore poisons **every test that runs after it in the same xdist worker**, and a rotating subset of the LLM-provider tests fails with:

```
TypeError: '>' not supported between instances of 'MagicMock' and 'int'
```

Which tests fail is decided purely by worker assignment, so the count swings between runs of *identical* code and shifts whenever anyone adds or removes tests. That makes the suite unreliable as a signal for whether a change broke something — which is how I ran into it.

## Mechanism

`hindsight_api/metrics.py:1102` holds the collector as a module global, defaulting to the no-op:

```python
_metrics_collector: MetricsCollectorBase = NoOpMetricsCollector()

def create_metrics_collector() -> MetricsCollector:
    global _metrics_collector
    _metrics_collector = MetricsCollector()      # permanent, no teardown
    return _metrics_collector
```

`hindsight_api/api/http.py:3864` calls it unconditionally in the lifespan:

```python
prometheus_reader = initialize_metrics(service_name="hindsight-api", service_version="1.0.0")
create_metrics_collector()
```

`NoOpMetricsCollector` ignores its arguments. The real `MetricsCollector` does not — `metrics.py:662`:

```python
if cached_input_tokens > 0:
```

Provider tests build `usage` as a bare `MagicMock`, so `usage.prompt_tokens_details.cached_tokens` auto-creates a child `MagicMock`, and that comparison raises.

Note this is **only** reachable through a test mock. Production is already safe: `_usage_from_openai_response` (`engine/providers/openai_compatible_llm.py:351-353`) coerces with `getattr(usage.prompt_tokens_details, "cached_tokens", 0) or 0`, so a real provider response that omits or nulls the field yields `0`. `getattr` on a `MagicMock` returns a truthy mock, which is why only tests are affected. No production fix is needed here.

## Reproduction

Deterministic, in one process, with a plugin that does what the lifespan does:

```python
# poison_metrics.py
def pytest_configure(config):
    from hindsight_api.metrics import create_metrics_collector, initialize_metrics
    initialize_metrics(service_name="poison", service_version="1.0.0")
    create_metrics_collector()
```

```
$ uv run pytest tests/test_reasoning_effort_explicit_config.py -p no:randomly -n 0 -p poison_metrics
13 failed
```

Without `-p poison_metrics` the same file is 13 passed. Confirmed on `main` at 21fbb9f93.

## Observed impact

Two full runs, same flags (`-m 'not hs_llm_core and not hs_llm_mat' -p no:randomly`), one on `main` and one on a branch that only adds retain/prompt refactors and six tests:

| | failed | occurrences of the `TypeError` |
|---|---|---|
| `main` @ 21fbb9f93 | 12 | 7 |
| branch (+6 tests) | 43 | 39 |

Nothing in that branch touches metrics, the collector, or any provider — the six extra tests just reshuffled xdist's distribution. Files hit across runs: `test_reasoning_effort_explicit_config.py`, `test_tool_path_reasoning_effort.py`, `test_tool_choice_required_downgrade.py`, `test_llm_extra_body.py`, `test_lmstudio_tool_choice.py`, `test_tags_visibility.py`. All pass in isolation.

## Suggested fix

An autouse fixture in `tests/conftest.py` that snapshots and restores `metrics._metrics_collector` around each test would fix it globally and cheaply — `tests/test_metrics.py:276-283` already does exactly this save/restore by hand, so the pattern is established. Restoring the previous collector in the lifespan's shutdown would also stop the app from leaking process state, which seems worth doing on its own merits.

Fixing the mocks in each provider test would work too, but leaves the next test that starts the app free to poison the next unlucky file.

Sanderhoff-alt:
## Summary

Fixes #3780.

This PR resolves a process-global metrics collector leak that caused nondeterministic `TypeError` failures across LLM provider test suites during concurrent `pytest-xdist` execution.

---

### 1.1 State Leak Lifecycle

During `pytest-xdist` test execution, when a worker process executes a test that initializes the FastAPI application lifespan (e.g. integration or HTTP tests), `create_metrics_collector()` mutates module-level global state `_metrics_collector`. Because neither application shutdown nor pytest teardown restored this reference, the worker process remained "poisoned" with an active `MetricsCollector` instance for all subsequent tests scheduled on that worker.

```mermaid
sequenceDiagram
    autonumber
    participant Worker as xdist Worker Process
    participant TestApp as Test (Starts FastAPI App)
    participant Lifespan as FastAPI Lifespan
    participant Metrics as hindsight_api.metrics
    participant ProviderTest as Subsequent Provider Test

    Note over Worker,Metrics: Initial State: _metrics_collector = NoOpMetricsCollector()
    Worker->>TestApp: Execute test
    TestApp->>Lifespan: Startup
    Lifespan->>Metrics: create_metrics_collector()
    Metrics-->>Worker: Global _metrics_collector = MetricsCollector() (LIVE)
    TestApp->>Lifespan: Shutdown (Previously NO metrics cleanup)
    Note over Metrics: Leaked: _metrics_collector remains MetricsCollector()

    Worker->>ProviderTest: Execute provider test on same worker
    ProviderTest->>Metrics: get_metrics_collector()
    Metrics-->>ProviderTest: Returns active MetricsCollector() (POISONED)
    ProviderTest->>Metrics: record_llm_call(cached_input_tokens=MagicMock)
    Metrics-->>ProviderTest: TypeError: '>' not supported between 'MagicMock' and 'int'
```

### 1.2 Failure Mechanism in LLM Provider Mocks

Provider unit tests mock LLM responses with `MagicMock`. When `_usage_from_openai_response` extracts `cached_tokens`:
```python
getattr(usage.prompt_tokens_details, "cached_tokens", 0) or 0
```
On a `MagicMock`, attribute traversal auto-instantiates child `MagicMock` instances. When passed to the active `MetricsCollector`, `if cached_input_tokens > 0` raises a `TypeError`.

| Component | Default / Isolated Test State | Poisoned Worker State | Result |
| :--- | :--- | :--- | :--- |
| `_metrics_collector` | `NoOpMetricsCollector()` | `MetricsCollector()` | Global state leak across tests |
| Token evaluation | No-op (arguments ignored) | `cached_input_tokens > 0` | Comparison executed against mock |
| Mock response | `response.usage = MagicMock()` | `response.usage = MagicMock()` | `cached_input_tokens` is `MagicMock` |
| Test outcome | **PASSED** | **FAILED** (`TypeError`) | Flaky failures dependent on xdist worker assignment |

---

## 2. Technical Design & Solution

```mermaid
flowchart TD
    subgraph CoreModule ["Core Module (hindsight_api/metrics.py)"]
        GlobalVar["_metrics_collector<br/>(Process-Global State)"]
        ResetAPI["reset_metrics_collector()"]
        NoOpInst["NoOpMetricsCollector()"]
        ResetAPI -->|reassigns to| GlobalVar
        NoOpInst -.->|default value| GlobalVar
    end

    subgraph ProdLifecycle ["Production / App Lifecycle (api/http.py)"]
        ShutdownHook["FastAPI Lifespan Teardown<br/>(on app shutdown)"] -->|invokes| ResetAPI
    end

    subgraph TestIsolation ["Test Environment (tests/conftest.py)"]
        AutouseFixture["Autouse Fixture<br/>_cleanup_leaked_metrics_collector"]
        AutouseFixture -->|1. Setup: snapshot & reset to NoOp| GlobalVar
        AutouseFixture -->|2. Teardown: restore previous state| GlobalVar
    end
```

### 2.1 Changes Implemented

1. **`hindsight_api/metrics.py`**:
   - Implemented `reset_metrics_collector()` to explicitly restore `_metrics_collector = NoOpMetricsCollector()`.

2. **`hindsight_api/api/http.py`**:
   - Added `reset_metrics_collector()` to the FastAPI `lifespan` teardown stage after `memory.close()`.

3. **`tests/conftest.py`**:
   - Added `_cleanup_leaked_metrics_collector` as an autouse fixture (mirroring the established `_cleanup_leaked_span_recorders` pattern) to guarantee per-test isolation under xdist.

4. **`tests/test_metrics.py`**:
   - Added unit test coverage for `reset_metrics_collector()` and cross-test fixture isolation.

---

### 3.1 Test Execution Matrix

| Test Suite / Target | Concurrency | Status |
| :--- | :--- | :--- |
| `tests/test_metrics.py` | `-n 0` (Single-process sequential) | **36 passed** |
| `tests/test_reasoning_effort_explicit_config.py` | `-n 4` (xdist concurrent) | **13 passed** |
| Impacted provider suites (`test_reasoning_effort_*`, `test_tool_path_*`, `test_llm_extra_body`, `test_lmstudio_*`, `test_tags_visibility`, `test_metrics`) | `-n 4` (xdist concurrent) | **202 passed** |

### 3.2 Pre-commit Checks
- `lint.sh`: All lints passed.
- `vulture` & `knip`: Passed.

### DIFF
--- a/hindsight-api-slim/tests/test_metrics.py
+++ b/hindsight-api-slim/tests/test_metrics.py
@@ -12,6 +12,7 @@ from hindsight_api.metrics import (
     create_metrics_collector,
     initialize_metrics,
     normalize_http_endpoint,
+    reset_metrics_collector,
 )
 
 
@@ -265,22 +266,41 @@ class TestMetricsCollector:
         assert attributes["bank_id"] == "test_bank"
 
 
+@pytest.mark.xdist_group(name="metrics_collector_isolation")
 class TestGetMetricsCollector:
-    """Tests for the get_metrics_collector function."""
+    """Tests for the get_metrics_collector function.
+
+    Pinned to a single xdist worker group so the leak-then-verify pair runs
+    sequentially on the same worker, proving the autouse fixture cleans up
+    a leaked collector between tests (#3780).
+    """
 
     def test_returns_noop_by_default(self):
         """Test that get_metrics_collector returns NoOpMetricsCollector by default."""
-        # Reset global state
-        import hindsight_api.metrics as metrics_module
+        collector = get_metrics_collector()
+        assert isinstance(collector, NoOpMetricsCollector)
 
-        original_collector = metrics_module._metrics_collector
+    def test_create_and_reset_metrics_collector(self):
+        """Test creating and resetting the global metrics collector (#3780)."""
+        with patch("hindsight_api.metrics.get_meter") as mock_get_meter:
+            mock_get_meter.return_value = MagicMock()
+            collector = create_metrics_collector()
+            assert isinstance(collector, MetricsCollector)
+            assert get_metrics_collector() is collector
 
-        try:
-            metrics_module._metrics_collector = NoOpMetricsCollector()
-            collector = get_metrics_collector()
-            assert isinstance(collector, NoOpMetricsCollector)
-        finally:
-            metrics_module._metrics_collector = original_collector
+            reset_metrics_collector()
+            assert isinstance(get_metrics_collector(), NoOpMetricsCollector)
+
+    def test_creates_collector_without_explicit_reset(self):
+        """First of the ordered pair: creates a collector without explicit reset (#3780)."""
+        with patch("hindsight_api.metrics.get_meter") as mock_get_meter:
+            mock_get_meter.return_value = MagicMock()
+            create_metrics_collector()
+            assert isinstance(get_metrics_collector(), MetricsCollector)
+
+    def test_collector_is_noop_in_subsequent_test(self):
+        """Second of the ordered pair: proves the fixture cleaned up the leak on this worker (#3780)."""
+        assert isinstance(get_metrics_collector(), NoOpMetricsCollector)
 
 
 class TestMetricsCollectorBase:
--- a/hindsight-api-slim/tests/test_metrics.py
+++ b/hindsight-api-slim/tests/test_metrics.py
@@ -266,14 +266,8 @@ class TestMetricsCollector:
         assert attributes["bank_id"] == "test_bank"
 
 
-@pytest.mark.xdist_group(name="metrics_collector_isolation")
 class TestGetMetricsCollector:
-    """Tests for the get_metrics_collector function.
-
-    Pinned to a single xdist worker group so the leak-then-verify pair runs
-    sequentially on the same worker, proving the autouse fixture cleans up
-    a leaked collector between tests (#3780).
-    """
+    """Tests for the get_metrics_collector and reset_metrics_collector functions."""
 
     def test_returns_noop_by_default(self):
         """Test that get_metrics_collector returns NoOpMetricsCollector by default."""
@@ -290,17 +284,6 @@ class TestGetMetricsCollector:
 
             reset_metrics_collector()
             assert isinstance(get_metrics_collector(), NoOpMetricsCollector)
-
-    def test_creates_collector_without_explicit_reset(self):
-        """First of the ordered pair: creates a collector without explicit reset (#3780)."""
-        with patch("hindsight_api.metrics.get_meter") as mock_get_meter:
-            mock_get_meter.return_value = MagicMock()
-            create_metrics_collector()
-            assert isinstance(get_metrics_collector(), MetricsCollector)
-
-    def test_collector_is_noop_in_subsequent_test(self):
-        """Second of the ordered pair: proves the fixture cleaned up the leak on this worker (#3780)."""
-        assert isinstance(get_metrics_collector(), NoOpMetricsCollector)
 
 
 class TestMetricsCollectorBase:
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -198,6 +198,7 @@ from hindsight_api.metrics import (
     get_metrics_collector,
     initialize_metrics,
     normalize_http_endpoint,
+    reset_metrics_collector,
 )
 from hindsight_api.models import RequestContext
 
@@ -3969,6 +3970,9 @@ def create_app(
         from hindsight_api.tracing import shutdown_tracing
 
         shutdown_tracing()
+
+        # Prevent in-process app instances from leaking a live collector across tests (#3780).
+        reset_metrics_collector()
 
     from hindsight_api import __version__
     from hindsight_api.config import get_config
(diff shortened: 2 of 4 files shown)


hindsight-api-slim/tests/test_metrics.py
```diff
@@ -265,22 +266,41 @@ def test_record_operation_includes_bank_id_when_enabled(self):
...
 
-        try:
-            metrics_module._metrics_collector = NoOpMetricsCollector()
-            collector = get_metrics_collector()
-            assert isinstance(collector, NoOpMetricsCollector)
-        finally:
-            metrics_module._metrics_collector = original_collector
+    def test_collector_is_noop_in_subsequent_test(self):
+        """Second of the ordered pair: proves the fixture cleaned up the leak on this worker (#3780)."""
+        assert isinstance(get_metrics_collector(), NoOpMetricsCollector)
```
ebarkhordar: I ran the new ordered pair against `1d8bf11c` in a clean `python:3.11-slim` container. It passes, but it cannot fail for the reason its docstring gives: the fixture resets the global to `NoOpMetricsCollector` at setup, so this assert is satisfied by the setup half whether or not the teardown restore ever runs.

Four runs of `test_creates_collector_without_explicit_reset` plus this test, `-p no:randomly -n 0`, each editing only `tests/conftest.py`:

```
as shipped                                     2 passed
teardown restore replaced with `pass`          2 passed
setup reset line deleted, restore kept         2 passed
both halves removed                            1 failed, 1 passed
```

So either half alone keeps the pair green, and the `xdist_group` pinning is not buying the ordering guarantee it describes. An autouse fixture that also resets at setup cannot be observed by a sibling test, so nothing scheduled after the leaking test can pin the restore.

The smallest honest edit is to drop `test_creates_collector_without_explicit_reset`, `test_collector_is_noop_in_subsequent_test` and the `xdist_group` marker, and keep `test_create_and_reset_metrics_collector`, which does exercise `reset_metrics_collector()` directly. If you want the fixture itself pinned, `pytester` running a two-test session in a subprocess is the mechanism that can actually see the teardown.

Separately, #3800 is open against the same issue and adds its own autouse fixture to the same region of `tests/conftest.py`, so whichever lands second will conflict there.

Sanderhoff-alt: Good catch and great analysis on the setup fixture behavior! Updated to drop the redundant paired isolation tests and the `xdist_group` marker, keeping `test_create_and_reset_metrics_collector` in 749024c14. Thanks!

ebarkhordar: Thanks for the quick turnaround. 749024c14 looks right to me: `test_create_and_reset_metrics_collector` calls `reset_metrics_collector()` directly, which is the part a sibling test structurally could not observe.

nicoloboschi:
superseded by #3816

PR #3787, opened 2026-08-25
refactor(retain): follow-ups to the retain-memory bound (#3756)

feniix:
Split out of #3777 at review request — these are #3756 retain-memory follow-ups, unrelated to that PR's prompt-assembly fix.

Rebuilt on top of #3788 (quicktok) and #3789, which landed after this was opened and removed most of the token-counting half of the original branch. What's left is the part quicktok does not address.

## One derivation for `documents.content_hash`

Five spellings of the same rule wrote that column: `_screen_document_body` used a private `_sha256_windowed` in `memory_engine`, while `handle_document_tracking`, `upsert_document_metadata` and the orchestrator's two paths each inlined `hashlib.sha256(_sanitize_text(...).encode())`.

That is not tidiness. The streaming, delta and un-sliced paths compare these values against *each other* as an ownership check, so a spelling that drifts does not raise — it silently stops matching and unchanged history gets re-extracted (#3282). `derive_document_content_hash` is now the single definition, in `fact_extraction` beside `_sanitize_text`, and all five callers use it. `_sha256_windowed`, `_HASH_WINDOW_CHARS` and `hashlib` are gone from `memory_engine`.

Sanitizing now rides along with the windowing instead of happening first, so a 45 MB body costs a window rather than two more full copies of itself — the orchestrator's delta path was building a whole sanitized copy purely to hash it, then assigning `""` over it to get the memory back. The digest is unchanged for rows already written: `_sanitize_text` deletes a fixed character class with no cross-character context, so per-window sanitizing produces the same bytes as sanitizing the whole string.

## The screening cache never evicted

`iter_sub_batches` kept a one-entry cache of a document's redacted copy and never cleared it, so a finished document stayed pinned for the rest of the submission — up to ~45 MB with Memory Defense on. It is cleared when a sub-batch carries no body. Slices of one document are consecutive, so this cannot cause a miss, and a test covers the case where packed small items land either side of the slices.

## Splitter cleanups

`_SubBatch.index` replaced by `enumerate`; `_flush()` yields instead of returning an `Optional` the caller unwraps at three sites; the same-object contract stated on `_RawSubBatch.body_override`, where it is produced, since rebuilding that string per slice would silently turn the cache into a permanent miss.

## Benchmarks

`hindsight-dev` now type-checks its benchmark imports instead of ignoring them. #3789 repaired `retain_memory.py` by hand; this is the check that would have caught it, and it immediately surfaces a second one — locomo's `--api-url` branch imports `HindsightClientAdapter` from `benchmark_runner`, which has never defined it, so that flag has always raised `ImportError` the moment it was used. It now says so plainly instead.

## Testing

`test_document_body_hash_once.py` gains four boundary cases for the derivation (a control character every other byte; multi-byte text sized to put boundaries inside the non-ASCII run) plus the memory-bound assertion that lived in `test_token_counting_windowed.py` before #3788 deleted it. Token counting no longer needs windowing, but sanitizing still does — hashing is the last windowed pass over a document body, and now the only one whose bound is hand-rolled.

Local: 1500 passed / 41 skipped across the retain, batching, document and token tests. `lint.sh` clean, `ty check` on `hindsight-dev` clean with `unresolved-import` enabled. The remaining local failures are all `HINDSIGHT_API_LLM_API_KEY environment variable is required` — real-LLM tests with no key here.

Same fork-CI caveat as #3777: `test-api` is gated on `has_secrets == 'true'`, so the shards skip and the full suite needs a run on your side.


feniix:
@strix-security

PR #3797, opened 2026-08-25
fix(providers): native ollama path honours the configured LLM timeout

MasterST1337:
`_call_ollama_native` built its own `httpx.AsyncClient(timeout=300.0)`. It is the one request path that ignored `HINDSIGHT_API_LLM_TIMEOUT` — every other path threads `self.timeout`.

On a CPU ollama host a single fact-extraction prompt can need longer than 300 s just to be ingested. The call is aborted mid-prompt and surfaces as a bare `Ollama connection error` that raising the configured timeout cannot fix, because the literal is what is actually in force.

### Please read this bit before merging — it cuts both ways

`DEFAULT_LLM_TIMEOUT` is **120 s**, which is **lower** than the 300 s literal being removed. So for a deployment that never set `HINDSIGHT_API_LLM_TIMEOUT`, this change *shortens* the native-ollama timeout from 300 s to 120 s.

That is the intent — one knob, honoured everywhere — but it makes this a **behaviour change, not a pure bug fix**, and such a deployment would need to raise the env var. If you would rather not regress that case, the alternative is `max(self.timeout, 300.0)`, which keeps the old floor at the cost of the native path still not fully honouring config. I went with the strict reading; say the word and I will switch it. Both the code comment and the commit message spell this out so it is not a silent change.

### Commits

1. The fix itself.
2. Review follow-ups: dropped `self.timeout or 300.0` → `self.timeout` (`__init__` always assigns `timeout or float(getenv(ENV_LLM_TIMEOUT, DEFAULT_LLM_TIMEOUT))`, so the `or` branch was unreachable and the literal duplicated a named constant), and added `tests/test_ollama_native_timeout.py`.

The test asserts both directions: a configured 900 s survives (the bug), and an unset timeout resolves to `DEFAULT_LLM_TIMEOUT` rather than 300.0 (the regression guard, which would catch the literal being reintroduced).

`scripts/hooks/lint.sh` clean.

### DIFF
--- a/hindsight-api-slim/tests/test_ollama_native_timeout.py
+++ b/hindsight-api-slim/tests/test_ollama_native_timeout.py
@@ -1,0 +1,63 @@
+"""The native Ollama path must use the configured LLM timeout, not a literal.
+
+`_call_ollama_native` built its own `httpx.AsyncClient(timeout=300.0)`, which is
+the one request path that ignored `HINDSIGHT_API_LLM_TIMEOUT`. On a CPU ollama
+host a single fact-extraction prompt can need longer than 300 s just to be
+ingested, and the call was aborted mid-prompt with a bare "Ollama connection
+error" that raising the configured timeout could not fix.
+
+Note the fix cuts both ways: `self.timeout` falls back to DEFAULT_LLM_TIMEOUT
+(120 s), which is LOWER than the old literal, so a deployment relying on the
+implicit 300 s must now set ENV_LLM_TIMEOUT. Both directions are asserted here.
+"""
+
+import httpx
+import pytest
+
+from hindsight_api.config import DEFAULT_LLM_TIMEOUT, ENV_LLM_TIMEOUT
+from hindsight_api.engine.providers import openai_compatible_llm
+from hindsight_api.engine.providers.openai_compatible_llm import OpenAICompatibleLLM
+
+
+class _CapturingAsyncClient:
+    """Stand-in for httpx.AsyncClient that records the timeout it was built with."""
+
+    captured: list[float | None] = []
+
+    def __init__(self, *args, timeout=None, **kwargs):
+        type(self).captured.append(timeout)
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, *exc):
+        return False
+
+    async def post(self, *args, **kwargs):
+        # Fail the call immediately: the assertion is about client construction,
+        # so the request itself never needs to succeed.
+        raise httpx.ConnectError("no ollama in tests")
+
+
+def _timeout_used(monkeypatch, env_value: str | None) -> float | None:
+    if env_value is None:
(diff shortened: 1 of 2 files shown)


hindsight-api-slim/tests/test_ollama_native_timeout.py
```diff
@@ -0,0 +1,63 @@
...
+def _timeout_used(monkeypatch, env_value: str | None) -> float | None:
+    if env_value is None:
+        monkeypatch.delenv(ENV_LLM_TIMEOUT, raising=False)
+    else:
+        monkeypatch.setenv(ENV_LLM_TIMEOUT, env_value)
+    monkeypatch.setattr(openai_compatible_llm.httpx, "AsyncClient", _CapturingAsyncClient)
+    _CapturingAsyncClient.captured = []
+
+    llm = OpenAICompatibleLLM(provider="ollama", api_key="local", base_url="http://localhost:11434", model="qwen")
+    return llm.timeout
```
Sanderhoff-alt, line 51: This helper never calls `_call_ollama_native`, so neither test exercises the fixed path. It patches `httpx.AsyncClient`, resets `_CapturingAsyncClient.captured`, then returns `llm.timeout` — the constructor attribute, which was already correct before this change and is already covered by `tests/test_llm_timeout_propagation.py::test_openai_compatible_provider_impl_receives_timeout`. Reintroducing `httpx.AsyncClient(timeout=300.0)` in `_call_ollama_native` would leave both tests green, so the stated regression guard does not hold, and `_CapturingAsyncClient` (including `captured`) is dead code as written.

Driving the actual path makes the assertion real, e.g.:

```python
@pytest.mark.asyncio
async def test_native_path_uses_configured_timeout(monkeypatch):
    monkeypatch.setenv(ENV_LLM_TIMEOUT, "900")
    monkeypatch.setattr(openai_compatible_llm.httpx, "AsyncClient", _CapturingAsyncClient)
    _CapturingAsyncClient.captured = []
    llm = OpenAICompatibleLLM(provider="ollama", api_key="local", base_url="http://localhost:11434", model="qwen")
    with pytest.raises(Exception):
        await llm._call_ollama_native(
            messages=[{"role": "user", "content": "hi"}],
            response_format=None,
            max_completion_tokens=None,
            temperature=None,
            max_retries=0,
            initial_backoff=0.0,
            max_backoff=0.0,
            skip_validation=True,
        )
    assert _CapturingAsyncClient.captured == [pytest.approx(900.0)]
```

With the path actually invoked, the patch of `httpx.AsyncClient` also stops being a no-op that only widens the blast radius for anything else using httpx during the test.

Sanderhoff-alt (commented):
The production change looks right, and it lines up with the docs, which already state 120 s as the effective default and suggest raising `HINDSIGHT_API_LLM_TIMEOUT` for slow local generation. One problem with the new test: it never runs the code path it claims to guard.

nicoloboschi (approved):
lgtm

PR #3802, opened 2026-08-26
fix(search): score pg_textsearch results per row

### LINKED ISSUE #3794: recall: scores.keyword is a per-query constant under pg_textsearch — the BM25 score expression scores the query against itself
## Bug Description

With `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=pg_textsearch`, the `keyword` value in every recall result is a **per-query constant**: identical across every row in the response, and varying only when the query changes. It is not a per-document BM25 score.

The cause is in `hindsight_api/engine/sql/postgresql.py:311`, where the score expression puts the query bind parameter on **both** sides of `<@>`:

```python
elif text_search_extension == "pg_textsearch":
    bm25_score_expr = f"-({text_param} <@> to_bm25query({text_param}, 'idx_memory_units_text_search'))"
    bm25_order_by   = f"text <@> to_bm25query({text_param}, 'idx_memory_units_text_search') ASC"
```

The `ORDER BY` on the next line is correct — it uses the row's `text` column. Only the score expression substitutes `text_param` for it, so the operator scores the query against itself.

Two things confirm this is unintended rather than a deliberate normalization:

* Every other branch in the same function scores against a row column: `vchord` uses `search_vector <&> …`, `pgroonga` uses `pgroonga_score(tableoid, ctid)`, `pg_search` uses `paradedb.score(id)`, and the native tsvector fallback uses `ts_rank_cd(search_vector, …)`.
* The knowledge-base arm in the same file, for the same extension, does it correctly — `postgresql.py:89` uses `{a}.content <@> to_bm25query({p}, …)`.

The extension itself guards against this shape. Passing a literal is rejected outright:

```
ERROR:  left operand of <@> must reference a table column
HINT:   Use column_name <@> bm25query, not a constant.
```

It only slips through because the engine passes a bind parameter rather than a literal.

## Steps to Reproduce

1. Run the engine against a populated bank with `HINDSIGHT_API_TEXT_SEARCH_EXTENSION=pg_textsearch`.
2. `POST /v1/{tenant}/banks/{bank}/memories/recall` with any query, e.g. `{"query": "What's the Float API key path?", "budget": "mid", "max_tokens": 8192}`.
3. Read `scores.keyword` across the returned results.

Three queries against one bank of ~770 memories:

| Query | Results carrying a keyword score | Distinct keyword values |
| --- | --- | --- |
| a credential-path lookup | 249 | **1** (`15.721446`) |
| an approval lookup | 151 | **1** (`24.635809`) |
| a policy-decision lookup | 175 | **1** (`27.021330`) |

Three unrelated natural-language queries; the value is constant within each response and differs between them.

The same result reproduces directly in SQL. The engine's form, as a prepared statement so the operand is a bind parameter:

```sql
PREPARE buggy(text) AS
  SELECT substr(text,1,38) AS snippet,
         round((-($1 <@> to_bm25query($1, 'idx_memory_units_text_search')))::numeric, 6) AS score
  FROM memory_units WHERE text ILIKE '%API key%' LIMIT 4;
EXECUTE buggy('What''s the Float API key path?');

                snippet                 |   score
----------------------------------------+-----------
 Before each deploy, verify that the se | 15.721446
 Before deployment, the secrets path mu | 15.721446
 Every API key path is stored in the co | 15.721446
 API keys must be rotated during the we | 15.721446
```

The same statement with the row column on the left, which is what the neighbouring `ORDER BY` already uses:

```sql
PREPARE correct(text) AS
  SELECT substr(text,1,38) AS snippet,
         round((-(text <@> to_bm25query($1, 'idx_memory_units_text_search')))::numeric, 6) AS score
  FROM memory_units WHERE text ILIKE '%API key%' LIMIT 4;
EXECUTE correct('What''s the Float API key path?');

                snippet                 |  score
----------------------------------------+----------
 Before each deploy, verify that the se | 7.651417
 Before deployment, the secrets path mu | 6.098290
 Every API key path is stored in the co | 7.250359
 API keys must be rotated during the we | 6.142835
```

## Expected Behavior

`scores.keyword` reflects each document's own BM25 relevance to the query, so it varies per result and is usable for interpreting and filtering recall.

## Actual Behavior

`scores.keyword` is one value shared by every result in the response, determined solely by the query text.

Two consequences, one cosmetic and one functional:

* **Reporting** — `scores.keyword` cannot be used to interpret why a result ranked where it did. This is what led us to it: we were diagnosing a ranking question and read the keyword leg as contributing nothing, when in fact the reported number simply carries no per-document information.
* **Filtering** — `min_scores.keyword` filters on this value. `ArmScores` documents the field as existing so that "the recall response can report them (and `min_scores` can filter on them)" (`engine/search/types.py:102-112`), but filtering on a per-query constant is all-or-nothing: either every keyword-arm result passes the threshold or none does, regardless of individual relevance.

Ranking itself appears unaffected. The `ORDER BY` uses the row column, so the BM25 arm retrieves and orders correctly, and fusion is rank-based rather than score-based.

The fix looks like a one-token change, matching the branch's own `ORDER BY` and the knowledge-base arm at line 89:

```python
bm25_score_expr = f"-(text <@> to_bm25query({text_param}, 'idx_memory_units_text_search'))"
```

## Version

Reproduced on 0.9.0 and 0.9.2. `postgresql.py:311` is unchanged on `main` as of this report.

## LLM Provider

Not relevant — the defect is in the SQL scoring expression for the keyword retrieval arm. Reproduced with OpenAI (`gpt-5.6-luna`) for extraction and a local Ollama embedding model, on PostgreSQL 18.4 with `pg_textsearch` 1.3.1.

mameikagou:
Fixes #3794.

The `pg_textsearch` score expression used the query parameter on both sides of `<@>`, so every result received the same keyword score. Use the row `text` column on the left, matching the existing `ORDER BY` expression.

Adds a regression test that verifies the score and ordering both use the row text while preserving the query bind parameter.

Validation:
- `pytest -n 2 tests/test_db_abstraction.py::TestPostgreSQLDialect` (38 passed)
- Ruff check and format check
- `ty check hindsight_api/engine/sql/postgresql.py`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/sql/postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/sql/postgresql.py
@@ -308,7 +308,7 @@ class PostgreSQLDialect(SQLDialect):
             # score so only genuine term matches survive into fusion/reranking.
             bm25_where_filter = f"AND -(search_vector <&> to_bm25query('idx_memory_units_text_search', tokenize({text_param}, 'llmlingua2'))) > {bm25_min_score:g}"
         elif text_search_extension == "pg_textsearch":
-            bm25_score_expr = f"-({text_param} <@> to_bm25query({text_param}, 'idx_memory_units_text_search'))"
+            bm25_score_expr = f"-(text <@> to_bm25query({text_param}, 'idx_memory_units_text_search'))"
             bm25_order_by = f"text <@> to_bm25query({text_param}, 'idx_memory_units_text_search') ASC"
             bm25_where_filter = ""
         elif text_search_extension == "pgroonga":
--- a/hindsight-api-slim/tests/test_db_abstraction.py
+++ b/hindsight-api-slim/tests/test_db_abstraction.py
@@ -287,6 +287,21 @@ class TestPostgreSQLDialect:
         )
         assert "> 2.5" in arm
 
+    def test_build_bm25_arm_pg_textsearch_scores_each_row(self, d):
+        arm = d.build_bm25_arm(
+            table="schema.memory_units",
+            cols="id, text",
+            fact_type="world",
+            bank_id_param="$2",
+            limit_param="$3",
+            text_param="$4",
+            text_search_extension="pg_textsearch",
+        )
+        expected_distance = "text <@> to_bm25query($4, 'idx_memory_units_text_search')"
+        assert f"-({expected_distance}) AS bm25_score" in arm
+        assert f"ORDER BY {expected_distance} ASC" in arm
+        assert "$4 <@>" not in arm
+
     def test_build_bm25_arm_pgroonga(self, d):
         arm = d.build_bm25_arm(
             table="schema.memory_units",


koriyoshi2041 (approved):
Verified at `5a340d8b`: the score expression now uses the row text just like the existing ordering path, while preserving the query bind parameter. The full PostgreSQL dialect class passes locally (38 tests), and the diff check is clean.

nicoloboschi (approved):
nice catch!

PR #3804, opened 2026-08-26
fix: recover Codex OAuth refresh after shared auth rotation

dwayn:
## Summary

This hardens Codex OAuth refresh handling for long-lived Hindsight processes that share Codex CLI auth state with other processes (e.g. an interactive `codex` session rotating `auth.json` underneath the API server).

Two failure modes, both ending in a hard auth error on a perfectly recoverable state:

**1. Refresh token already rotated elsewhere (401 from the refresh endpoint).** Hindsight now re-reads `auth.json` under the shared lock and recovers when possible:

- If the disk access token is fresh, Hindsight adopts it and avoids an unnecessary second refresh.
- If the disk access token is stale but the disk refresh token changed, Hindsight retries refresh once with the newer disk refresh token.
- If disk credentials did not change or still cannot recover, Hindsight continues to surface the existing terminal auth error.

**2. `auth.json` already rotated *before* the lock was taken.** The pre-lock adopt path short-circuited on `force or _token_is_fresh_with_known_expiry()`, so a reactive caller returned any on-disk token that merely *differed* from the one in memory — including one already past its `exp`. Reactive callers get exactly one retry (`CodexOAuthEmbeddings.encode` retries `super().encode` once on `AuthenticationError`), so handing back a dead token spent that retry on a second 401 and failed the whole call. The short-circuit is now gated on the JWT clock alone.

Also drops a dead `response = None` initializer and hoists the "Refreshing…" log line out of the retry loop so it is not emitted twice per refresh.

## Test coverage

Regression coverage for both races, plus isolation so these tests never read a developer's or CI runner's real Codex credentials (`$CODEX_HOME` and `Path.home()` are both redirected at a tmp dir — without this, `test_codex_oauth_embeddings_picks_up_refreshed_token_on_encode` fails on any machine where Codex is actually logged in).

Each new test was confirmed to fail against the unpatched provider.

## Test Plan

- `uv run pytest -n0 tests/test_codex_oauth_refresh.py tests/test_codex_home_env.py tests/test_codex_request_headers.py tests/test_embeddings_openai_batch_size.py -q`
- Wider local sweep — all 8 `test_codex_*`, both `test_nous_*`, 3 embeddings suites: 107 passed
- `./scripts/hooks/lint.sh`
- `uv run ty check hindsight_api/engine/providers/codex_auth.py`
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/codex_auth.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/codex_auth.py
@@ -431,34 +431,46 @@ class CodexAuthManager:
                         "Run 'codex auth login' to re-authenticate."
                     )
 
-                log_reason = f" ({reason})" if reason else ""
-                logger.info(f"Refreshing Codex OAuth access_token{log_reason}")
+                response: httpx.Response | None = None
+                refresh_attempt = 0
+                while True:
+                    log_reason = f" ({reason})" if reason else ""
+                    logger.info(f"Refreshing Codex OAuth access_token{log_reason}")
 
-                request_access_token = self.access_token
-                request_refresh_token = self.refresh_token
-                request_body = {
-                    "client_id": _CODEX_CLIENT_ID,
-                    "grant_type": "refresh_token",
-                    "refresh_token": request_refresh_token,
-                }
-                try:
-                    response = self._http_client.post(
-                        _CODEX_REFRESH_TOKEN_URL,
-                        json=request_body,
-                        headers={"Content-Type": "application/json"},
-                        timeout=30.0,
-                    )
-                except httpx.RequestError as e:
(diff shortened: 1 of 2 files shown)


dwayn:
I ran into this race condition in hindsight where it got hung up and failed to auth due to consumed refresh token, so I tasked my agent with tracking down and fixing the cause. It found this race condition problem and patched it, so I thought I would share it.

It happened to find a couple of little regression issues in a couple of places where `CODEX_HOME` was not being cleared during tests and was causing it to misbehave because my agent has an explicit `CODEX_HOME` set in its environment.
 
My Hermes agent is sharing codex auth with hindsight by bind mounting .codex directory into hindsight container so that I don’t have to install and auth codex separately in the hindsight container. I think this is also why I had the fun of running into this race condition in the first place, but it had been working great for a while until I hit this snag, so this is not a particularly high probability race but can/does occur.

PR #3810, opened 2026-08-26
fix(control-plane): reset knowledge state on bank switch

### LINKED ISSUE #3807: Control Plane requests stale Knowledge Page and mental model IDs after switching banks
### Bug Description

Switching banks from the Control Plane header while the Knowledge view is open briefly combines the new bank ID with the previous bank's selected Knowledge Page and mental model IDs.

This sends two predictable 404 requests and displays these warning toasts:

- `Knowledge page 'kp-...' not found`
- `Failed to get mental model`

The normal bank-switch path recovers after route navigation remounts the view. A mismatched `?page=` deep link does not recover automatically and leaves the content pane empty.

I reproduced the bank-switch failure in 6 out of 6 attempts across different bank pairs.

Hindsight Control Plane showing stale Knowledge Page and mental model errors after switching banks

### Steps to Reproduce

1. Open `/banks/<bank-a>?view=knowledge`.
2. Wait for the first Knowledge Page to be selected and rendered.
3. Use the header bank selector to switch to Bank B.
4. Observe the warning toasts and network requests.

The request sequence includes the new bank ID with IDs owned by Bank A:

```text
GET /api/knowledge-base/tree?bank_id=<bank-b>                                      200
GET /api/banks/<bank-b>/mental-models/<bank-a-mental-model-id>                    404
GET /api/knowledge-base/pages/<bank-a-page-id>?bank_id=<bank-b>                   404
GET /api/knowledge-base/pages/<bank-b-page-id>?bank_id=<bank-b>                   200
GET /api/banks/<bank-b>/mental-models/<bank-b-mental-model-id>                    200
```

The deep-link path can be reproduced independently:

1. Take a Knowledge Page ID owned by Bank A.
2. Open `/banks/<bank-b>?view=knowledge&page=<bank-a-page-id>`.
3. The page request returns 404 and the content pane remains empty.

### Expected Behavior

Changing banks should clear or scope all Knowledge view state before any request is made for the new bank. A page ID that does not belong to the selected bank should fall back to a valid page or a clear empty/not-found state without issuing requests from stale state.

### Actual Behavior

The bank context changes before the Knowledge view's previous state is cleared. Effects then run with a mixed snapshot:

- `currentBank` belongs to Bank B.
- `roots`, `tabs`/`selected`, and the backing mental model ID still belong to Bank A.
- The auto-selection guard is reset while the previous tree is still present.

This causes stale page and mental model IDs to be requested against Bank B. State-reset calls in the bank-change effect do not prevent other effects in the same committed render from observing the previous values.

The relevant effects are in [`knowledge-base-view.tsx`](https://github.com/vectorize-io/hindsight/blob/main/hindsight-control-plane/src/components/knowledge-base-view.tsx):

- the selected page provenance effect requests the previous mental model with the new bank ID;
- the deep-link effect can request a `?page=` ID that belongs to another bank;
- the auto-selection effect can select the first page from the previous bank's tree.

### Impact

The observed failing requests were GET-only. During one complete reproduction session, all 299 requests were GET requests; no POST, PATCH, or DELETE was issued. The stale IDs were not persisted in local storage, session storage, or cookies.

The regular selector path is mainly a UI/error-notification defect because route navigation remounts the view and subsequent polling uses valid Bank B IDs. The mismatched deep-link path has a visible functional impact because the page content remains empty.

### Suggested Direction

Keying the Knowledge view by the current bank would remount it as soon as the bank changes and reset all bank-scoped state together. Another option is to track which bank produced the loaded tree, tabs, and selected model, then guard every dependent effect against `loadedBank !== currentBank`.

Clearing only the tree is not sufficient: it can prevent the auto-selection request while the selected-page provenance effect still holds the previous page and mental model ID.

### Version

- API/UI-reported version: `0.9.2`
- Control Plane image: `ghcr.io/vectorize-io/hindsight-control-plane:latest`
- Image ID: `sha256:4f83fa7ae5c55461bc55cd3da1056305df8b6b9325503e2875f069c735efafad`

### LLM Provider

Other

pstanton237:
## Background

The knowledge-base view kept page and mental-model state from the previous bank during a bank switch. Effects could then request those stale IDs against the newly selected bank, producing 404 notifications. A `?page=` link containing a missing, folder, or foreign-bank ID also prevented automatic selection and left the content pane empty.

## Changes

- Remount the knowledge-base view when the selected bank changes so page, tab, and provenance state cannot cross bank boundaries.
- Validate `?page=` against the loaded bank tree before requesting it.
- Fall back to the first page when the requested ID is invalid for the current bank.
- Add unit coverage for valid, missing, folder, and absent page parameters.

## Verification

- `TZ=UTC npm test` (158 tests)
- `npm run build`
- `./scripts/hooks/lint.sh`
- `./scripts/hooks/check-unused.sh` (advisory; no new findings)
- Headless browser network QA:
  - switching between populated banks produced no failed API requests
  - a foreign-bank `?page=` produced no 404 and opened the current bank's first page
  - a valid `?page=` requested and opened the specified page with HTTP 200

Fixes #3807

### DIFF
--- a/hindsight-control-plane/src/app/[locale]/banks/[bankId]/page.tsx
+++ b/hindsight-control-plane/src/app/[locale]/banks/[bankId]/page.tsx
@@ -718,7 +718,7 @@ export default function BankPage() {
                     </div>
                   </div>
 
-                  {knowledgeTab === "pages" && <KnowledgeBaseView />}
+                  {knowledgeTab === "pages" && <KnowledgeBaseView key={bankId} />}
                   {knowledgeTab === "models" && (
                     <div>
                       <p className="text-sm text-muted-foreground mb-4">
--- a/hindsight-control-plane/src/components/knowledge-base-view.tsx
+++ b/hindsight-control-plane/src/components/knowledge-base-view.tsx
@@ -50,6 +50,7 @@ import {
 } from "lucide-react";
 import { Spinner } from "@/components/ui/spinner";
 import { formatAbsoluteDateTime, formatRelativeTime } from "@/lib/relative-time";
+import { findRequestedKnowledgePageId } from "@/lib/knowledge-base-navigation";
 import { CompactMarkdown } from "./compact-markdown";
 import { StalenessBadge } from "./staleness-badge";
 import { FreshnessLine } from "./freshness-line";
@@ -201,6 +202,10 @@ export function KnowledgeBaseView() {
   }, [currentBank, loadTree]);
 
   const allNodes = useMemo(() => flatten(roots), [roots]);
+  const requestedPageId = useMemo(
+    () => findRequestedKnowledgePageId(roots, pageParam),
+    [roots, pageParam]
+  );
   // A synthetic top-level folder for the bank so the root itself is visible and
   // you can add folders/pages directly under it. Its "" id makes add-child create
   // at the root; it's never deletable (TreeRow hides delete for the root).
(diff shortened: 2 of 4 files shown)


pstanton237:
Closing because #3813 has merged the same fix with component-level regression coverage.

PR #3827, opened 2026-08-27
fix(llm): raise OutputTooLongError when a non-streaming completion is truncated

### LINKED ISSUE #3811: finish_reason: "length" on the non-streaming .create() path is treated as a parse error, not OutputTooLongError — the fact-extraction auto-split (#2579/#3174) is still dead code for the truncation subclass
## Bug Description

`openai_compatible_llm._content_or_error` does not raise when the provider returns `finish_reason == "length"`. The `except LengthFinishReasonError → raise OutputTooLongError` handler in the same file is only reachable via the SDK's `.parse()` method — but the structured-output retain path calls plain `.create()` (three call sites, e.g. `openai_compatible_llm.py:1013`), where the SDK never raises `LengthFinishReasonError`.

So when a dense chunk makes the model exceed the completion-token cap:

1. The provider returns a truncated JSON body with `finish_reason: "length"`.
2. `_content_or_error` returns the truncated content (no error).
3. The caller's `json.loads` fails with `Unterminated string ...`.
4. The provider's retry ladder re-sends the **byte-identical** request up to `max_retries` times (this is #3683's retry class — but the cause here is deterministic truncation, not a repairable parse defect).
5. The chunk is dropped, and the `except OutputTooLongError` auto-split in `_extract_facts_with_auto_split` (`fact_extraction.py`) never runs — for this failure subclass it remains dead code exactly as it was before #3174 unified the exception classes.

## Steps to Reproduce

No server, no database, no API key — against the published image:

```bash
docker run --rm --entrypoint /app/api/.venv/bin/python \
  ghcr.io/vectorize-io/hindsight:latest -c '
import types, json
from hindsight_api.engine.providers import openai_compatible_llm as P

truncated_json = """{"facts": [{"what": "user deployed a three-node cluster", "why": "infra decision"""
resp = types.SimpleNamespace(
    choices=[types.SimpleNamespace(
        finish_reason="length",
        message=types.SimpleNamespace(content=truncated_json, tool_calls=None, refusal=None),
    )],
    usage=types.SimpleNamespace(prompt_tokens=800, completion_tokens=4096, total_tokens=4896),
    model="test-model",
)

content, choice = P._content_or_error(resp, provider="openai", model="test-model", scope="retain_fact_extraction")
print("returned content ok:", content[:40])
try:
    json.loads(content)
except json.JSONDecodeError as e:
    print("caller json.loads fails:", e)
'
```

Output on 0.9.2:

```
returned content ok: {"facts": [{"what": "user deployed a thr
caller json.loads fails: Unterminated string starting at: line 1 column 66 (char 65)
```

`_content_or_error` returned truncated content with no error, and the unreachable handler (`except LengthFinishReasonError as e: raise OutputTooLongError(...)` at ~L1164) exists in the same file.

## Expected Behavior

`finish_reason == "length"` should be converted to `OutputTooLongError` (as the unreachable handler intends), so `_extract_facts_with_auto_split` splits the chunk recursively and the facts are recovered instead of burned.

## Actual Behavior

Each over-length chunk costs `(max_retries+1) × full-completion` inference and is then dropped — the session's facts never land in the bank. Measured on a self-hosted deployment over 7 days: 10 retain/consolidation operations terminally failed with `Unterminated string` JSONDecodeError; 5 at `retry_count=0` (fast give-up) and 5 at `retry_count=3` (full retry ladder burned first). On a capped local model this converts a ~6% daily fact-loss rate into a silent, deterministic leak.

## Suggested fix (10 lines)

In `_content_or_error`, raise the SDK exception manually so the existing handler converts it:

```python
if finish_reason == "length":
    raise LengthFinishReasonError(completion=response.model_copy())
```

Notes: in openai ≥2.x the constructor requires the `completion` kwarg (the ChatCompletion object, not a message string). `LengthFinishReasonError` is already imported at the top of the file (L36). We have this running in production behind a bind-mount overlay and it works as intended: the auto-split now fires and recovers dense chunks instead of burning the retry ladder.

## Related issues (same family, not duplicates)

- **#3172 / #3174** — unified the two `OutputTooLongError` classes so the auto-split could run. Necessary but not sufficient: this report is the complementary half — for the `finish_reason=length` subclass the exception is never **raised** on the `.create()` path, so the (now-unified) catch still never fires.
- **#3683** — the identical-retry ladder this defect falls into once truncation becomes a parse error. Repairing the JSON (its fix) cannot recover facts that were never generated; the length signal must be surfaced as `OutputTooLongError` instead.
- **#2571** — the original auto-split intent; #2579/#3172/#3174 lineage.

## Environment

- Hindsight 0.9.2, image `ghcr.io/vectorize-io/hindsight:latest`
- `HINDSIGHT_API_LLM_PROVIDER=openai`, custom `HINDSIGHT_API_LLM_BASE_URL` (llama.cpp llama-server), completion cap below dense-chunk extraction output
- openai SDK 2.24.0 (bundled in image)

ebarkhordar:
## Root cause

`chat.completions.create()` reports a token-limit truncation only through `finish_reason`. It never raises `LengthFinishReasonError`, which is what the `except LengthFinishReasonError` handler in `call()` converts into `OutputTooLongError`. Both structured-output and free-form calls go through `create()`, so that handler cannot fire for either of them, and `_content_or_error` returned the truncated body as though it were complete.

What that looked like downstream:

- **Structured output:** the caller's `json.loads` failed on the cut, so a truncation surfaced as a JSON parse error. The retry ladder treated it as transient and re-sent a byte-identical request against the same token limit.
- **Free-form output:** the truncated text was returned as the answer, with nothing to signal that it had been cut.

`_extract_facts_with_auto_split` recovers from truncation by catching `OutputTooLongError` and splitting the chunk, so on this path it could never fire.

## Fix

Detect the truncation where the signal actually arrives and raise there. This matches the two sibling providers: `litellm_llm` raises on `finish_reason == "length"`, and `openai_responses_llm` raises when the response status is incomplete with reason `max_output_tokens`.

The check sits ahead of the content read, so a response that is truncated before any visible token raises `OutputTooLongError` too. An earlier revision of this PR put it after the empty-content branch and described that as deliberate. That was wrong, and koriyoshi2041 caught it in review: the empty-content branch sets `retryable = finish_reason not in {"content_filter"}`, so `length` came back retryable and `call()` re-sent the same request against the same limit instead of letting the auto-split run.

## Verification

- New `tests/test_openai_compatible_truncation.py`, now 7 tests: 5 fail on `main` and all 7 pass on this branch. Two of the 7 are controls that pass on both, one for a normal `finish_reason: "stop"` response and one for an empty response with a non-truncation finish reason, which still raises the retryable `ProviderResponseError`.
- The two tests for the empty truncation also fail at this PR's own earlier ordering, not only on `main`: the `call()` one logs 4 attempts against the same limit and then raises `ProviderResponseError`.
- The two `call()` tests assert `create.call_count == 1`, pinning that a truncation is not retried against the same limit.
- Ran the provider and fact-extraction suites (`test_fact_extraction_retry`, `test_multi_llm_provider`, `test_openai_responses_provider`, `test_deepseek_tool_call_compat`, `test_consolidation_retry_budget`, `test_xai_oauth_llm` and 8 more) on both `main` and this branch: the set of failures is identical, and all of them are pre-existing in `test_xai_oauth_llm.py` on `main`. `ruff check`, `ruff format --check` and `ty check hindsight_api` are clean, with a `ty` diagnostic profile identical to `main`.
- Not verified: any behaviour against a live provider. The tests drive a mocked `chat.completions.create`, so what they pin is this repo's handling of a truncated response shape, not that a given provider emits that shape.
- Also not run end to end: a truncated provider response travelling all the way into the auto-split. The two halves are covered separately, by the tests here and by the existing `test_fact_extraction_retry.py`, and the catch site at `fact_extraction.py:1907` takes the same `OutputTooLongError` class this raises, but I did not exercise the whole path in one test.

## Interaction with #3685

No textual conflict: #3685 edits the `JSONDecodeError` handler around line 1053 and this changes `_content_or_error` around line 343. They are complementary rather than competing, but the order matters, so it is worth stating. A truncated response no longer reaches that handler at all, because it raises before returning content. If it did reach it, `parse_llm_json` would repair the cut into valid but silently partial JSON, whereas raising lets the auto-split re-extract the whole chunk. Non-truncation malformed JSON is untouched by this change and still reaches #3685's repair path.

Fixes #3811

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -317,7 +317,11 @@ def _first_choice_or_error(response: Any, *, provider: str, model: str, scope: s
 
 
 def _content_or_error(response: Any, *, provider: str, model: str, scope: str) -> tuple[str, Any]:
-    """Extract message.content while turning provider shape issues into useful errors."""
+    """Extract message.content, turning provider shape issues into useful errors.
+
+    Raises ``OutputTooLongError`` on a token-limit truncation and
+    ``ProviderResponseError`` on every other unusable success shape.
+    """
 
     choice = _first_choice_or_error(response, provider=provider, model=model, scope=scope)
     message = _message_for_choice(choice)
@@ -331,7 +335,18 @@ def _content_or_error(response: Any, *, provider: str, model: str, scope: str) -
 
     # chat.completions.create() signals truncation only through finish_reason; unlike
     # .parse(), it never raises LengthFinishReasonError for the handler in call() to convert.
-    # Checked before the content branch below: an empty response can still be a truncation.
+    # Checked before the content branch below: an empty response can still be a truncation,
+    # and reading it as empty content instead raises a *retryable* ProviderResponseError,
+    # which re-sends the same request against the same limit (#3811).
+    #
+    # Raised for every scope, not just the ones that recover from it. That matches the
+    # sibling OpenAI-shaped providers (litellm_llm, openai_responses_llm), and it means
+    # a truncated free-form answer — reflect synthesis, a mental-model page — now fails
+    # the call instead of being returned as if complete. Note this is the opposite of
+    # what gemini_llm does for the same signal (it logs a warning and returns the cut
+    # text, see #3365): there the truncation is common because reasoning tokens eat the
+    # visible budget, so failing every such call would be worse than surfacing it. The
+    # two can diverge only until one of them is shown wrong by a real workload.
     if finish_reason == "length":
         raise OutputTooLongError(
             f"LLM output exceeded token limits ({provider}/{model}, scope={scope}). "
@@ -349,7 +364,6 @@ def _content_or_error(response: Any, *, provider: str, model: str, scope: str) -
             f"refusal={bool(refusal)})",
             retryable=retryable,
         )
-
     return content, choice
 
 
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -340,6 +340,15 @@ def _content_or_error(response: Any, *, provider: str, model: str, scope: str) -
             f"refusal={bool(refusal)})",
             retryable=retryable,
         )
+
+    # chat.completions.create() signals truncation only through finish_reason; unlike
+    # .parse(), it never raises LengthFinishReasonError for the handler in call() to convert.
+    if finish_reason == "length":
+        raise OutputTooLongError(
+            f"LLM output exceeded token limits ({provider}/{model}, scope={scope}). "
+            "Input may need to be split into smaller chunks."
+        )
+
     return content, choice
 
 
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -331,12 +331,7 @@ def _content_or_error(response: Any, *, provider: str, model: str, scope: str) -
 
     # chat.completions.create() signals truncation only through finish_reason; unlike
     # .parse(), it never raises LengthFinishReasonError for the handler in call() to convert.
-    # This sits ahead of the empty-content branch below because a budget exhausted before the
-    # first visible token is still a truncation. Routing that to the retryable empty-content
-    # error would re-send the same request against the same limit rather than let the caller
-    # split its input. Both sibling providers raise ahead of their content read for the same
-    # reason: litellm_llm coerces content to "" first, and openai_responses_llm checks the
-    # incomplete status before reading output_text.
+    # Checked before the content branch below: an empty response can still be a truncation.
     if finish_reason == "length":
         raise OutputTooLongError(
             f"LLM output exceeded token limits ({provider}/{model}, scope={scope}). "
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -329,6 +329,20 @@ def _content_or_error(response: Any, *, provider: str, model: str, scope: str) -
             retryable=True,
         )
 
+    # chat.completions.create() signals truncation only through finish_reason; unlike
+    # .parse(), it never raises LengthFinishReasonError for the handler in call() to convert.
+    # This sits ahead of the empty-content branch below because a budget exhausted before the
+    # first visible token is still a truncation. Routing that to the retryable empty-content
+    # error would re-send the same request against the same limit rather than let the caller
+    # split its input. Both sibling providers raise ahead of their content read for the same
+    # reason: litellm_llm coerces content to "" first, and openai_responses_llm checks the
+    # incomplete status before reading output_text.
+    if finish_reason == "length":
+        raise OutputTooLongError(
+            f"LLM output exceeded token limits ({provider}/{model}, scope={scope}). "
+            "Input may need to be split into smaller chunks."
+        )
+
     content = _message_content(message)
     if content is None or content == "":
         tool_calls = _message_tool_calls(message)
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (changes requested):
The `length` check needs to happen before the empty-content branch. OpenAI-compatible providers can return `content=""` with `finish_reason="length"` when the limit is exhausted before any visible token; this branch currently converts that truncation into a retryable `ProviderResponseError`, so `call()` repeats the same request and fact extraction never reaches its `OutputTooLongError` auto-split path.

I reproduced this at `fd11c15` by passing an empty-content/`length` response to `_content_or_error`: it raises `ProviderResponseError`, while the non-empty truncation tests pass. Moving the finish-reason check ahead of the content check and adding that case should keep the recovery contract consistent for both empty and partial truncations. The focused provider suites otherwise pass 12/12 with proxy variables cleared.

ebarkhordar:
You are right, and the sentence in the description defending that ordering was wrong.

The cost is worse than a class change: the empty-content branch sets `retryable = finish_reason not in {"content_filter"}`, so `length` came back retryable, and `call()` retries a retryable `ProviderResponseError`. Running the `call()` case at the old ordering logs four attempts against the same limit before it gives up with the wrong class, so the auto-split never sees it.

Pushed in aeea639, with the `length` check ahead of the content read.

Correcting one line of this comment, edited in afterwards: I wrote that both siblings raise ahead of their content read. Only `openai_responses_llm` does, calling `_raise_if_truncated` before it reads `output_text`. `litellm_llm` reads content first, coercing it to `""` at `litellm_llm.py:330` and checking `finish_reason` at `:343`, and it carries no empty-content branch at all, which is why the bug cannot appear there. The precedent held for one sibling, not two. The comment in the source said the same thing and is trimmed to the accurate part in the follow-up commit.

Three tests: the `_content_or_error` unit for the empty case, a `call()` one pinning `create.call_count == 1`, and a control that an empty response with any other `finish_reason` still raises the retryable error. The first two fail at the previous ordering. Description corrected too.

koriyoshi2041 (approved):
Rechecked at `aeea639`. The `length` guard now runs before the empty-content path, and the new call-level case confirms an empty truncation raises `OutputTooLongError` without retrying (`create.call_count == 1`); the non-truncated empty-response control still keeps the existing retryable provider error. The focused truncation suite passes 7/7, and `git diff --check HEAD^ HEAD` is clean.

ebarkhordar:
Flagging that @koriyoshi2041 opened #3854 for the same issue about a day after this one. It makes the same call in the same place: raise `OutputTooLongError` from `_content_or_error` when `finish_reason == "length"`, so `.create()` matches what `.parse()` already does.

One substantive difference between them. #3854 raises before the `message is None` check, so it also covers a truncated response that came back with no message at all; this PR raises after that check, so that case still surfaces as a `ProviderResponseError`. Its guard placement is the better of the two. What this PR has instead is provider, model and scope in the error text, and a wider regression file.

Happy to close this in favour of #3854 if the smaller diff is easier to take, or to move the guard above the message check here and carry the tests. No preference from me, whichever suits the review.

koriyoshi2041 (changes requested):
The new empty-content case is covered, but the no-message truncation that this branch was meant to pick up is still classified as a retryable provider error. `_content_or_error` reads `message` and raises on `message is None` before it checks `finish_reason == "length"`.

So a choice shaped like `message=None, finish_reason="length"` still retries the identical request and never reaches the split handler. Please move the length guard ahead of the `message is None` branch and add that exact regression; the existing empty-string case has a message object and does not cover it.

I rechecked exact head `cbe2e0030`; the four-commit diff and `git diff --check HEAD~4..HEAD` are clean. The focused suite could not run in the isolated worktree because the cached environment lacks `json_repair`, so this finding is from the direct control-flow readback rather than a passing local test receipt.

ebarkhordar:
You are right, and the merge did not change it. On main `_content_or_error` still raises the retryable `ProviderResponseError` for `message is None` before it reaches the `finish_reason == "length"` check, so a choice with no message and a length finish still re-sends the same request against the same limit.

That is the one difference I flagged between this and your #3854 on the 28th: yours puts the guard ahead of the message check, so it covers that shape. Rebased onto main it should reduce to the reorder plus the regression test, and I am happy to leave it with you. If you would rather not, say so and I will open the follow-up.

PR #3828, opened 2026-08-27
fix: propagate configured temperature through reflect calls

### LINKED ISSUE #3825: Reflect ignores HINDSIGHT_API_LLM_TEMPERATURE_REFLECT: no temperature sent on reflect or mental-model refresh LLM calls
**Version:** v0.9.2 (also verified on v0.9.1 and current `main`)
**Provider:** `openrouter`, model `z-ai/glm-5.2`

## Summary

`HINDSIGHT_API_LLM_TEMPERATURE_REFLECT` is documented as the temperature for reflect, but no code path in the reflect engine reads it. Reflect requests — including every mental-model refresh — are sent with **no `temperature` field at all**, so the upstream provider applies its own default. The env var is silently inert for the operation it is named after.

## Evidence (paths relative to `hindsight-api-slim/hindsight_api/`, tag `v0.9.2`)

- `grep -rn temperature engine/reflect/` returns **zero hits** — the reflect package never passes a temperature.
- The tool-call request is built without a `temperature` key: `engine/reflect/agent.py:865-869` (`ct_kwargs = dict(messages=…, tools=…, scope="reflect_tool_call", tool_choice=…)`). The synthesis calls at `agent.py:238`, `agent.py:681`, and the rewrite at `agent.py:1293` likewise pass none.
- `config.llm_temperature_reflect` is resolved from the env var in `config.py`, but its only runtime reader is the legacy think path: `engine/search/think_utils.py:230`. Nothing in `engine/reflect/` or the refresh path in `engine/memory_engine.py` consults it.
- The provider client omits the field entirely when it receives `None`: `engine/providers/openai_compatible_llm.py:1363` (`if temperature is not None: call_params["temperature"] = temperature`), so the wire request carries no temperature.
- No sampling seed is sent either, except for `provider == "groq"` (`openai_compatible_llm.py:944`, `:1385`), so there is no reproducibility fallback.

## Impact

Mental-model refreshes and reflect answers run at the provider's default temperature regardless of configuration. An operator who sets `HINDSIGHT_API_LLM_TEMPERATURE_REFLECT=0.1` and restarts gets no behavior change and no warning. Measured on an unchanged bank: three consecutive `full`-mode refreshes of the same mental model produced documents of 1701 / 1067 / 1802 tokens (1.69x spread) with the env var set to `0.1` — because the setting never reached the request.

This is the same class of defect as the one fixed for consolidation in v0.9.2 ("Consolidation now respects the configured temperature setting when generating consolidated results"); the reflect lane still has it.

## Repro

1. `HINDSIGHT_API_LLM_TEMPERATURE_REFLECT=0.1`, restart.
2. Run any `reflect` (or `refresh_mental_model`).
3. Inspect the outbound request body (provider logs, or a proxy): no `temperature` field is present. Statically: `grep -rn temperature hindsight_api/engine/reflect/` → no hits.

## Suggested fix

Thread `get_config().llm_temperature_reflect` into the reflect LLM calls the same way consolidation now does: add `temperature=…` to `ct_kwargs` at `engine/reflect/agent.py:865` and to the `llm_config.call(...)` sites at `:238`, `:681`, `:1293`. The provider layer already forwards it when non-None.

koriyoshi2041:
## Problem

`HINDSIGHT_API_LLM_TEMPERATURE_REFLECT` is resolved by configuration but reflect requests omit it, so providers apply their defaults. This affects tool turns, final synthesis, structured extraction, and over-budget rewrites.

Closes #3825.

## Fix

Forward `llm_temperature_reflect` through every reflect LLM call path. Add deterministic coverage for the tool loop, rewrite, and structured-output paths.

## Test

- `uv run pytest tests/test_reflect_agent.py::TestReflectStructuredOutput::test_structured_output_forwards_reflect_temperature tests/test_reflect_agent.py::TestReflectAgentMocked::test_done_tool_answer_respects_max_tokens -q` (2 passed)
- `uv run ruff check hindsight_api/engine/reflect/agent.py tests/test_reflect_agent.py`
- `uv run ruff format --check hindsight_api/engine/reflect/agent.py tests/test_reflect_agent.py`
- `git diff --check`

The full reflect-agent file also passed 45 tests; four environment-dependent tests could not start because the local checkout lacks `pg0-embedded` and provider credentials. The repository-wide lint wrapper reached unrelated TypeScript linting and stopped because workspace npm dependencies were not installed.

## Risk

Low. The provider interfaces already accept optional temperature values, and the default remains `None`, preserving existing provider-default behavior when the setting is unset.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -247,7 +247,9 @@ OUTPUT:"""
             response_format=DynamicModel,
             scope="reflect_structured",
             strict_schema=get_config().llm_strict_schema_reflect,
-            temperature=get_config().llm_temperature_reflect,
+            # Schema extraction should be deterministic. The configured reflect
+            # temperature applies to answer generation, not this parsing pass.
+            temperature=0.0,
             max_completion_tokens=max_tokens,
             max_retries=1,
             initial_backoff=0.25,
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -247,6 +247,7 @@ OUTPUT:"""
             response_format=DynamicModel,
             scope="reflect_structured",
             strict_schema=get_config().llm_strict_schema_reflect,
+            temperature=get_config().llm_temperature_reflect,
             max_completion_tokens=max_tokens,
             max_retries=1,
             initial_backoff=0.25,
@@ -687,6 +688,7 @@ async def _run_reflect_agent_inner(
                 {"role": "user", "content": prompt},
             ],
             scope="reflect",
+            temperature=get_config().llm_temperature_reflect,
             max_completion_tokens=completion_cap,
             return_usage=True,
         )
@@ -878,6 +880,7 @@ async def _run_reflect_agent_inner(
                 tools=tools,
                 scope="reflect_tool_call",
                 tool_choice=iter_tool_choice,
+                temperature=get_config().llm_temperature_reflect,
             )
             if incremental_caching and iter_tool_choice is LLM_TOOL_CHOICE_AUTO and rolling_cache_name is not None:
                 ct_kwargs["cached_prefix"] = rolling_cache_name
@@ -1307,6 +1310,7 @@ async def _process_done_tool(
                 {"role": "user", "content": rewrite_user},
             ],
             scope="reflect",
+            temperature=get_config().llm_temperature_reflect,
             max_completion_tokens=get_config().reflect_max_completion_tokens,
             return_usage=True,
         )
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -1406,7 +1406,7 @@ class OpenAICompatibleLLM(LLMInterface):
 
         if max_completion_tokens is not None:
             call_params[self._max_tokens_param_name()] = max_completion_tokens
-        if temperature is not None:
+        if temperature is not None and not self._supports_reasoning_model():
             # MiniMax requires temperature in (0.0, 1.0] — clamp accordingly
             if self.provider == "minimax":
                 temperature = max(0.01, min(temperature, 1.0))
--- a/hindsight-api-slim/tests/test_deepseek_tool_call_compat.py
+++ b/hindsight-api-slim/tests/test_deepseek_tool_call_compat.py
@@ -99,6 +99,22 @@ def test_deepseek_reasoning_models_still_use_reasoning_parameters():
 
 
 @pytest.mark.asyncio
+async def test_reasoning_tool_call_omits_temperature():
+    llm = _make_deepseek_llm("deepseek-v4-pro")
+
+    with patch.object(llm._client.chat.completions, "create", new_callable=AsyncMock) as mock_create:
+        mock_create.return_value = _make_tool_call_response()
+        await llm.call_with_tools(
+            messages=[{"role": "user", "content": "Search observations."}],
+            tools=TOOLS,
+            temperature=0.9,
+            max_retries=0,
+        )
+
+    assert "temperature" not in mock_create.call_args.kwargs
+
+
+@pytest.mark.asyncio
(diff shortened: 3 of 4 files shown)


Sanderhoff-alt (commented):
The threading itself is complete (all four reflect call sites plus the pre-existing think path), but the tool-loop call reaches a provider path that never had to handle a temperature before, and the resolved default is not `None`.

hindsight-api-slim/hindsight_api/engine/reflect/agent.py
```diff
@@ -867,6 +869,7 @@ async def _forced_final_synthesis(iterations_completed: int) -> ReflectAgentResu
                 tools=tools,
                 scope="reflect_tool_call",
                 tool_choice=iter_tool_choice,
+                temperature=get_config().llm_temperature_reflect,
```
Sanderhoff-alt, line 885: `agent.py:874` is the only in-repo caller of `call_with_tools`, so this is the first time a temperature reaches that path — and `OpenAICompatibleLLM.call_with_tools` does not apply the reasoning-model suppression its own `call()` does (`providers/openai_compatible_llm.py:930`: `if temperature is not None and not is_reasoning_model`, versus `:1363` where the guard is absent). `OpenAIResponsesLLM` has it on both paths (`:456`, `:574`).

Concrete break: `provider=openai`, `model=gpt-5` (or `o1`/`o3`) routes to `OpenAICompatibleLLM`; with no config set the reflect tool loop now sends `temperature=0.9`, which chat/completions rejects with a 400 (`'temperature' does not support 0.9 with this model`), so every reflect fails after retries. Before this change no temperature was sent and it worked.

Suggested fix in the provider, to keep the two methods symmetric:

```python
if temperature is not None and not self._supports_reasoning_model():
    ...
    call_params["temperature"] = temperature
```

hindsight-api-slim/hindsight_api/engine/reflect/agent.py
```diff
@@ -246,6 +246,7 @@ def _model_for(schema: dict, name: str) -> type:
             response_format=DynamicModel,
             scope="reflect_structured",
             strict_schema=get_config().llm_strict_schema_reflect,
+            temperature=get_config().llm_temperature_reflect,
```
Sanderhoff-alt: `llm_temperature_reflect` resolves to `DEFAULT_LLM_TEMPERATURE_REFLECT = 0.9` when unset (`config.py:222`, `_resolve_operation_temperature`), not `None` — so contrary to the risk note in the description, every deployment that never set the env var changes behaviour here, and this particular call is the JSON-schema extraction of the final answer, which previously ran at the provider default. Please either confirm 0.9 is intended for structured extraction or use a deterministic value for this call, and correct the description so reviewers of the release notes are not misled about the blast radius.

hindsight-api-slim/hindsight_api/engine/reflect/agent.py
koriyoshi2041: Fixed in 529b4abb1: call_with_tools now mirrors call() and omits temperature for reasoning models. Added a regression that captures the outgoing request.

hindsight-api-slim/hindsight_api/engine/reflect/agent.py
koriyoshi2041: Agreed on the blast radius. The structured extraction pass is now explicitly deterministic at 0.0; the configured reflect temperature only drives answer generation. Updated the test accordingly.

nicoloboschi (approved):

- All four reflect LLM call sites now forward `llm_temperature_reflect`; grep confirms `engine/reflect/` has no other call site, and mental-model refresh runs through reflect so it is covered.
- The `call_with_tools` reasoning-model suppression is the right companion change: `call()` already dropped temperature for gpt-5/o1/o3/DeepSeek-reasoner, so without it the new reflect temperature would start failing on the tool path.
- Hardcoding `temperature=0.0` for the structured-extraction pass deviates from the issue's suggestion, but the determinism rationale holds and there is precedent (`bank_utils.py`).

Verification: rebased onto main (a test added upstream conflicted with the modified `test_done_tool_answer_respects_max_tokens`; both kept). `test_reflect_agent.py` + `test_deepseek_tool_call_compat.py` = 55 passed locally, ruff check/format clean. Full CI dispatched on an internal branch since fork PRs skip `test-api`: the only failures are `test_bank_roundtrip_remaps_mental_model_based_on_ids` (pre-existing on main) and `smoke-openclaw-install` (failing repo-wide), both unrelated.

PR #3829, opened 2026-08-27
fix(reflect): accept a top-level array of delta operations

### LINKED ISSUE #3820: parse_delta_operation_list rejects a top-level array of ops (valid JSON, schema walked) even after the #3421 transport-cap fix — deterministic wedge on thinking-disabled models
## Summary

A delta-mode mental-model refresh can still wedge permanently on `hindsight-api-slim`
**0.9.1 — a release that already contains the #3421/#3424 fix** — because #3424 only
fixed the *truncation* root cause. With thinking disabled and the transport cap already
decoupled (`reflect_max_completion_tokens`, uncapped by default), `deepseek-v4-flash`
still occasionally returns a *structurally-walked but JSON-legal* delta payload: a
**top-level JSON array of operations** instead of `{"operations": [...]}`.

`parse_delta_operation_list` has zero tolerance for a non-dict top level and raises
`delta payload must be an object with an operations array`. At
`temperature = llm_temperature_consolidation` (default 0.0) the output is deterministic
for a given window, so the failure replays identically on every retry — exactly the
permanent `MentalModelRefreshError` loop #3421 described, surviving the fix.

## Root cause

Two distinct malformations were folded under one error message. #3424 fixed the first:

1. **Truncated mid-string JSON** (thinking models spend the budget on reasoning tokens) —
   fixed by decoupling the transport cap from `delta_max_tokens` (#3421/#3424).
2. **Schema-walked-but-legal JSON** (a top-level array) — NOT addressed. The delta call
   runs in text mode (`parse_delta_operation_list` does the JSON schema enforcement), so
   no provider-side `response_format` constrains the top-level type. A list of bare ops

   ```json
   [{ "op": "append_block", "section_id": "x", "block": { "type": "paragraph", "text": "hi" } }]
   ```

   is accepted by `json.loads`, then `_validate_operations_list` / `_finalize_operations`
   are never reached because the code requires `payload["operations"]` on a dict first.

`finish_reason` is `stop` (not truncated) in this case, confirming it is not a cap issue.

The #3424 trace fix also means `mental_model_delta_ops` calls are now recorded in the
`llm_requests` trace table, so these walked payloads are retrievable for diagnosis —
including the top-level-array shape this issue is about.

## Reproduction

Call `parse_delta_operation_list` with a list-of-ops array:

```python
from hindsight_api.engine.reflect.delta_ops import parse_delta_operation_list
parse_delta_operation_list('[{"op":"append_block","section_id":"x","block":{"type":"paragraph","text":"hi"}}]')
# raises: ValueError: delta payload must be an object with an operations array
```

Real-world: on 0.9.1 (post-#3424) with `deepseek-v4-flash` via `api.deepseek.com`,
`extra_body={"thinking":{"type":"disabled"}}`, a long batch delta prompt repeatedly
returned a top-level array; each refresh `mode_fallback_reason = "delta_ops_failed"`,
content preserved, same error spamming logs on every retry.

## Impact

- Mental model content stuck stale despite new memories; no automatic recovery.
- Bolt-on to #3421: that issue is closed, but its *second* malformation class is still live.
- The #3112 window guard correctly refuses to advance the watermark, so this is a
  permanent-no-recovery loop, not a one-off.

## Expected behavior

A delta payload that is *a JSON array of valid op objects* should be normalized to
`{"operations": <that array>}` and passed through the existing validation — not rejected
outright. All the ops are already typed (`op` key) and individually validated downstream.

## Proposed fix (parse-level only — no fallback, no apply changes)

In `delta_ops.py`, before requiring `payload["operations"]` on a dict, normalize a
top-level list: if any element is a dict containing `"operations"`, use that container;
otherwise collect the bare op objects into `{"operations": [...]}` and let the existing
`_validate_operations_list` decide validity. No change to apply semantics, watermark, or
any fallback path — a pure widening of what "parseable" means:

```python
def _normalize_top_level(payload):
    if isinstance(payload, list):
        for x in payload:
            if isinstance(x, dict) and "operations" in x:
                return x
        ops = [x for x in payload if isinstance(x, dict) and "op" in x]
        if ops:
            return {"operations": ops}
    return payload
```

This is deliberately scoped to *recovering ops the model already emitted*, keeping delta
mode intact — consistent with the maintainers' stance in #3421 against full re-synthesis.

## Version

`hindsight_api.__version__ == 0.9.1` (0.9.1 includes the #3421/#3424 transport-cap fix).

## LLM Provider

DeepSeek (`deepseek-v4-flash` @ `https://api.deepseek.com`), thinking disabled,
temperature 0.0.

## Related

- #3421 / #3424 — hydraulic fix for the *truncation* root cause; this is the residual
  *structure-walk* root cause that fix does not cover.
- #3112 — the window guard whose "refuse to overwrite" behavior turns this into a loop.
- #645 — similar JSON-tolerance gap (code fence stripping); the same spirit of accepting
  legal-but-fringy LLM JSON instead of failing.

ebarkhordar:
`parse_delta_operation_list` requires the parsed JSON to be a dict carrying an `operations` key. When a model emits the operations as a bare top-level array, which is legal JSON and structurally complete, the guard rejects the whole payload with `delta payload must be an object with an operations array`, so `_validate_operations_list` and `_finalize_operations` never see ops that are individually valid. At `llm_temperature_consolidation` (0.0) the output is deterministic for a given window, so the refresh then fails identically on every retry.

#3424 decoupled the transport cap and fixed the truncation case. This is the other malformation that shared that error message: here the JSON is complete, only its top-level container differs.

## Fix

Normalize a top-level list to `{"operations": <list>}` before the dict check, then fall through to the existing path. Per-op validation, invalid-op skipping and the all-ops-invalid error are unchanged, because the normalized payload reaches exactly the same code. The normalization only fires on a value that previously took the `continue` branch, so no payload that parses today changes behaviour.

## Verification

Run in a clean container at `8c01b89`:

- the three new tests in `tests/test_delta_operation_parse.py` fail on main with the `ValueError` above and pass on the branch
- across the ten test files that cover `delta_ops`, passes go from 228 to 231, the difference being these three; 80 errors appear identically on both arms and are a missing `pg0` extra in the container, not a regression
- `ruff check`, `ruff format --check` and `ty check hindsight_api` at the locked 0.14.9 and 0.0.8, with `ty` reporting the same 211 diagnostics on both arms
- install, `compileall` and the import smoke on 3.11 and 3.14, the ends of the `build-api-python-versions` matrix

What this does not cover: nothing here drives a live provider or a real refresh, so the end-to-end wedge stays your observation rather than something I reproduced. I also left the prose-wrapped case alone, since `_extract_balanced_json_object` still searches for a `{` and an array wrapped in commentary is unchanged.

One note on the sample payload in the report: its op carries the v1 `block` field, which `AppendBlockOp` does not define, so that exact snippet still raises, now as `DeltaAllOpsInvalidError` from per-op validation instead of on the top-level type. `test_parse_delta_operation_list_rejects_v1_block_payloads` pins that shape as invalid on purpose, so the tests here use `text`.

Fixes #3820

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/delta_ops.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/delta_ops.py
@@ -270,6 +270,8 @@ def parse_delta_operation_list(raw: Any) -> DeltaOperationList:
         except json.JSONDecodeError as exc:
             last_error = exc
             continue
+        if isinstance(payload, list):
+            payload = {"operations": payload}
         if not isinstance(payload, dict) or "operations" not in payload:
             last_error = ValueError("delta payload must be an object with an operations array")
             continue
--- a/hindsight-api-slim/tests/test_delta_operation_parse.py
+++ b/hindsight-api-slim/tests/test_delta_operation_parse.py
@@ -112,3 +112,32 @@ def test_parse_delta_operation_list_all_invalid_raises():
 def test_parse_delta_operation_list_pydantic_instance():
     original = DeltaOperationList(operations=[AppendBlockOp(section_id="s", text="- a")])
     assert parse_delta_operation_list(original) is original
+
+
+def test_parse_delta_operation_list_top_level_array():
+    """A bare array of ops carries the same delta as the dict form (#3820)."""
+    op_list = parse_delta_operation_list('[{"op": "append_block", "section_id": "x", "text": "hi"}]')
+    assert len(op_list.operations) == 1
+    op = op_list.operations[0]
+    assert isinstance(op, AppendBlockOp)
+    assert op.section_id == "x"
+    assert op.text == "hi"
+
+
+def test_parse_delta_operation_list_top_level_array_skips_invalid_op():
+    """Ops in an array are validated one by one, exactly as they are inside the dict form."""
+    raw = (
+        '[{"op": "append_block", "section_id": "s", "text": "ok"}, '
+        '{"op": "replace_block", "section_id": "s", "text": "missing block_id"}]'
+    )
+    op_list = parse_delta_operation_list(raw)
+    assert len(op_list.operations) == 1
+    assert isinstance(op_list.operations[0], AppendBlockOp)
+    assert op_list.operations[0].text == "ok"
+
+
+def test_parse_delta_operation_list_top_level_array_all_invalid_raises():
+    """An array of v1 typed-block ops is still every-op-invalid, so it must still raise."""
+    raw = '[{"op": "append_block", "section_id": "s", "block": {"type": "paragraph", "text": "old shape"}}]'
+    with pytest.raises(DeltaAllOpsInvalidError):
+        parse_delta_operation_list(raw)


nicoloboschi (approved):

ebarkhordar:
Thanks for the quick review and merge.

Issue #3312, opened 2026-08-10
Knowledge-base routes bypass the operation validator — any tenant caller can read/write another bank's knowledge tree


kubaodias:
## Problem

All routes under `/v1/default/banks/{bank_id}/knowledge-base/*` reach the engine without ever invoking the operation validator. On a deployment that relies on `OperationValidatorExtension` for bank-level authorization (multi-tenant, one instance shared by many owners), any authenticated tenant caller can read, create, move, or delete **any** bank's knowledge tree by guessing/knowing its `bank_id` — regardless of what the validator would decide.

This is not just a metadata leak: knowledge pages render mental-model content (`GET .../pages/{page_id}`, `.../search`, `.../export`), so it is a read-your-neighbor's-synthesized-memories hole. Notably, the direct mental-model routes (`validate_mental_model_get` / `validate_mental_model_refresh`) **are** gated, so KB reaches the same content on an unvalidated path.

## Root cause

The gap is structural, not a missing call site:

- `BankReadOperation` and `BankWriteOperation` (`hindsight_api/extensions/operation_validator.py`) have **no knowledge-base members**, so there is no hook a KB handler could call even if it wanted to.
- None of the nine KB engine methods reference `self._operation_validator`; the only auth they perform is `_authenticate_tenant` (tenant/schema resolution, not bank ownership). `_ensure_bank_exists` is a lazy row-create, not an ACL.
- The KB route handlers already carry a dead `except OperationValidationError` (e.g. `http.py:5290`) copied from validator-gated routes — the engine can never raise it on these paths. That looks like gating was intended but never wired.

This is the same class of issue as #3036 (legacy bank writes bypass the validated lifecycle) and #2650 (rejected config writes still mutate/provision), and it reopens exactly the "guessed `bank_id` read" hole that `validate_bank_read` was added to close — but on a surface those fixes never touched.

## Affected surface

Routes (`hindsight_api/api/http.py`):
- `GET .../knowledge-base/tree`
- `GET .../knowledge-base/pages/{page_id}`
- `GET .../knowledge-base/search`
- `GET .../knowledge-base/export`
- `POST .../knowledge-base/folders`
- `POST .../knowledge-base/pages`
- `PATCH .../knowledge-base/nodes/{node_id}`
- `DELETE .../knowledge-base/nodes/{node_id}`

Engine methods (`hindsight_api/engine/memory_engine.py`) — none call the validator:
- `list_knowledge_nodes`
[...]

PR #3835, opened 2026-08-27
fix(security): run operation validator in get_mental_model_history (#…

### LINKED ISSUE #3831: get_mental_model_history skips the operation validator — cross-tenant read of a bank's mental-model history
**Version:** 0.9.2 (`hindsight-api-slim`), reproduced against a live deployment; source references are `v0.9.2` (`ebad47824`).

## Summary

`get_mental_model_history` authenticates the tenant but never runs the operation validator, so `GET /v1/default/banks/{bank_id}/mental-models/{mental_model_id}/history` returns a mental model's full revision history to a caller who does not own the bank. Every sibling mental-model method validates; this one is the only gap.

Same class as #3312 (knowledge-base routes bypassing the operation validator): an engine method that authenticates the tenant but emits no `BankReadOperation`, so an extension-provided bank ACL never sees the call.

## Reproduction

Same bank, same mental model id, one path segment apart — a caller who does not own `{bank_id}`:

```
GET /v1/default/banks/{bank_id}/mental-models/{mm_id}          -> 403 {"detail": "caller is not authorized for the requested bank"}
GET /v1/default/banks/{bank_id}/mental-models/{mm_id}/history   -> 404 {"detail": "Mental model '{mm_id}' not found"}
```

The 403 never appears on `/history` for any input, because the gate is not invoked at all — the 404 comes from the row lookup. On a bank the caller owns, the same route returns 200 with every past revision's `previous_content`.

Behaviour is bank-scoped: requesting a known-good `mm_id` under a *different* bank returns 404, so the lookup does honour the `bank_id` segment. What is missing is the ownership check on that bank.

## Cause

`engine/memory_engine.py:13566`:

```python
async def get_mental_model_history(self, bank_id, mental_model_id, *, request_context):
    await self._authenticate_tenant(request_context)      # <- authentication only
    backend = await self._get_backend()
    async with acquire_with_retry(backend) as conn:
        exists = await conn.fetchrow(
            f"SELECT id FROM {fq_table('mental_models')} WHERE bank_id = $1 AND id = $2",
            bank_id, mental_model_id,
        )
```

No `_validate_operation` call. Compare the siblings in the same file:

| method | line | validator |
|---|---|---|
| `get_mental_model` | 13488 | `validate_mental_model_get` |
| `list_mental_models` | 13384 | `validate_bank_read` / `LIST_MENTAL_MODELS` |
| `create_mental_model` | 13688 | `validate_bank_write` / `CREATE_MENTAL_MODEL` |
| `update_mental_model` | 14865 | `validate_bank_write` / `UPDATE_MENTAL_MODEL` |
| `clear_mental_model` | 15177 | `validate_bank_write` / `CLEAR_MENTAL_MODEL` |
| `delete_mental_model` | 15234 | `validate_bank_write` / `DELETE_MENTAL_MODEL` |
| **`get_mental_model_history`** | **13566** | **none** |

## Impact

A caller holding a valid token for any tenant can read the full refresh history — every stored `previous_content` snapshot — of a mental model on a bank they do not own, given the correct `bank_id` and `mental_model_id`.

Bounds worth stating plainly:

- The list route **is** gated (403), so a caller cannot enumerate a victim bank's mental-model ids. They have to guess.
- `mental_model_id` is caller-chosen at creation. Human-set slugs (`last_3_days_work`, `oncall_summary`) are guessable, so for those this is a real content disclosure. Auto-generated `mm-<32 hex>` ids are not, so for those it degrades to a 200-vs-404 existence oracle.
- Bank ids are guessable by convention in deployments that namespace them (`squad:<team>:<slug>`), which is the assumption the read gate exists to defend against.

## Suggested fix

Run the same validator the sibling read path runs — `_validate_operation` with the `mental_model_get` context (or a `BankReadOperation` for history) — before the row lookup in `get_mental_model_history`.

More generally, this is the second instance of the same failure mode after #3312: an engine method that authenticates but emits no operation is invisible to any validator extension, and nothing in the type system flags it. A test that asserts every public per-bank engine method emits an operation would catch the next one.

Sword-Saint69:
## FIX

Fixes #3831 where `get_mental_model_history` (`GET /v1/default/banks/{bank_id}/mental-models/{mental_model_id}/history`) authenticated the tenant but omitted operation validation (`_validate_operation`).

Without this check, callers holding a valid token for any tenant could read past mental-model snapshots of a bank they do not own, bypassing extension ACLs.

## Changes

- **`hindsight_api/engine/memory_engine.py`**:
  - Added pre-operation validation check (`validate_mental_model_get`) to `get_mental_model_history` using `MentalModelGetContext`.
  - Added post-operation hook (`on_mental_model_get_complete`) on successful history retrieval for consistent operation tracking.
- **`tests/test_mental_model_hooks.py`**:
  - Added `TestMentalModelHistoryOperationValidation` to verify that rejected operations raise `OperationValidationError` (403) and authorized requests trigger `on_mental_model_get_complete`.

## Testing

```bash
python -m pytest hindsight-api-slim/tests/test_mental_model_hooks.py -v

### CODE REVIEW

**Comment by Sword-Saint69** [comment]
@nicoloboschi @benfrank241 @r266-tech please check and review and if its good if any issue just mention me i will fix and reissue the pr i really liked the project great structure and beautiful ideas

**Review by r266-tech — COMMENTED** [review]
The pre-operation gate matches the sibling `get_mental_model()` path and addresses the reported ACL gap. I found one blocking issue in the added completion accounting:

`previous_reflect_response` is a structured object in the history API contract, but `total_content_len` applies `len()` directly to it. For objects and arrays this counts keys/elements rather than the serialized response content, so usage is substantially under-reported. Other valid or legacy JSON scalar values can raise `TypeError`; because the accounting and `MentalModelGetResult` construction are outside the `try`, that would turn a successful history read into a 500 even though the post-operation hook is intended to be non-fatal.

Please normalize or deterministically serialize each returned value before measuring it, and keep accounting/result construction inside the non-fatal hook boundary. The success regression should include a nested `previous_reflect_response` and assert the exact `output_tokens`; its current `None` fixture only proves that the hook was called and cannot catch this failure.

A preauthorized-path test would also be useful to pin that validation is skipped only when the matching authorization is consumed while completion is still recorded, but the structured-payload accounting is the blocking part.

**Comment by Sword-Saint69** [comment]
Thanks for the thorough review and great feedback, @r266-tech! 

I have addressed all the points raised in the latest commits:

### 1. Robust Payload Serialization & Accounting
- **Normalized JSON Serialization**: `previous_reflect_response` and `previous_content` are now deterministically measured. Objects/arrays are serialized via `json.dumps(val, default=str)` prior to length calculation, ensuring output tokens reflect the actual response payload rather than key/element counts.
- **Non-Standard Type Safety**: Using `default=str` in `json.dumps` prevents `TypeError` on custom or non-standard serializable types (e.g. `datetime`, `UUID`).

### 2. Encapsulated Non-Fatal Hook Boundary
- Moved the entire token calculation, `MentalModelGetResult` construction, and hook invocation strictly **inside** the `try ... except Exception:` block. Any unexpected accounting or telemetry hook error is safely logged as a non-fatal warning and will never turn a successful history read into a `500 Internal Server Error`.

### 3. Expanded Test Suite
- **Structured Payload Accounting Test**: Added `test_get_mental_model_history_completes_post_hook_with_structured_reflect_response` in `tests/test_mental_model_hooks.py` fixture-ing a nested structured `previous_reflect_response` dictionary and asserting exact `output_tokens`.
- **Preauthorized Path Test**: Added `test_get_mental_model_history_preauthorized_skips_validation_but_runs_hook` verifying that preauthorized calls consume pre-granted access while continuing to emit completion telemetry.
- **Non-Fatal Hook Resilience Test**: Added `test_get_mental_model_history_hook_error_is_non_fatal` asserting that exceptions in post-operation processing do not interrupt response delivery.
- **Automated Operation Validator Coverage Test**: Added `tests/test_engine_operation_coverage.py` using `inspect` to verify that all public per-bank `MemoryEngine` methods emit operation validation or are explicitly classified, preventing future ACL gaps like #3312 and #3831.

All 14 tests pass cleanly (`1.67s`).

cc @nicoloboschi @benfrank241 @r266-tech — ready for re-review!

**Review by r266-tech — COMMENTED** [review]
The original structured-payload accounting issue is resolved: history values are now measured deterministically, and accounting/result construction are inside the non-fatal hook boundary. The focused structured-response, preauthorized, and hook-failure regressions cover those changes well.

One new blocking concern remains in `test_engine_operation_coverage.py`. Its oracle treats any occurrence of `_operation_validator` in a method's source as proof of pre-operation validation. A method with only a completion-hook guard, or even that token in a comment/docstring, passes without calling `_validate_operation` or consuming a matching preauthorization. The large hand-maintained exemption set also means this cannot enforce the architectural security invariant claimed by the test.

Please either remove that repository-wide coverage test from this focused fix, or make it verify actual call nodes to `_validate_operation` / `_consume_preauthorized_*` and include a negative fixture showing that a completion-only `_operation_validator` reference fails. The existing focused history tests already pin the reported regression directly. Also please revert the unrelated blank-line deletion in `_insert_pinned_mental_model` while touching the diff.

File: hindsight-api-slim/hindsight_api/engine/memory_engine.py [hunk 1]
```diff
@@ -13762,6 +13762,24 @@ async def get_mental_model_history(
 
         """
         await self._authenticate_tenant(request_context)
+
+        # Pre-operation validation (credit check / usage metering)
+        if self._operation_validator:
```
Sanderhoff-alt: This gate is missing the nested-call guard the other bank-read paths use:

```suggestion
        if self._operation_validator and not _nested_operation_authorized.get():
```

`export_knowledge_base` (memory_engine.py:16262-16311, `with _authorize_nested_operations():`) validates once as `EXPORT_KNOWLEDGE_BASE` and then reads every page, calling `get_mental_model_history` once per mental-model-backed page. Without the flag check, exporting a knowledge base with N such pages now runs `validate_mental_model_get` N extra times: a caller whose extension authorizes the export but not per-model gets fails with 403 part-way through the export, and a metering extension is charged N times for one operation. The invariant is documented at memory_engine.py:120-125 ("export_knowledge_base validates once, then reads every page"), and `get_knowledge_page` — called from the same loop — checks the flag at line 15854.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -13798,7 +13816,38 @@ async def get_mental_model_history(
                         "changed_at": changed_at.isoformat() if hasattr(changed_at, "isoformat") else changed_at,
                     }
                 )
-            return result
+
+        # Post-operation hook (usage recording)
+        if result is not None and self._operation_validator:
```
Sanderhoff-alt: Same for the completion hook: as written, an export fires `on_mental_model_get_complete` once per page in addition to its own export accounting. Please add `and not _nested_operation_authorized.get()` here as well, so a nested read stays invisible to the validator on both halves.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -13819,7 +13868,6 @@ async def _insert_pinned_mental_model(
         trigger: dict[str, Any] | None,
     ) -> ResultRow:
         """Insert a pinned model using the caller's transaction.
-
```
Sanderhoff-alt, line 13961: This blank-line removal is unrelated to the fix and it drops the separator between the docstring summary line and its body. Please restore it to keep the branch limited to the change.

Sanderhoff-alt (commented):
The validator is now reached on every branch of the direct call, and the coverage guard passes (checked against main, where only `get_mental_model_history` is reported). One problem remains: the new gate does not honour the nested-operation flag, which changes behaviour for the knowledge-base export path.

Sword-Saint69:
Thanks @r266-tech and @Sanderhoff-alt for the thorough review and great feedback!

Updated in commit [`40a251c3`](https://github.com/Sword-Saint69/hindsight/commit/40a251c3):

1. **Nested Operation Authorization Gate**: Added `and not _nested_operation_authorized.get()` check to both pre-operation validation (`validate_mental_model_get`) and post-operation completion hook (`on_mental_model_get_complete`) in `get_mental_model_history`. This ensures nested reads (e.g. multi-page exports) remain properly invisible to the validator on both halves.
2. **Focused Security Scope**: Removed `test_engine_operation_coverage.py` as requested to keep this PR minimal and strictly focused on pinning `get_mental_model_history`.
3. **Docstring Formatting**: Restored the docstring blank separator line in `_insert_pinned_mental_model`.
4. **Regression Tests**: Added `test_get_mental_model_history_nested_operation_skips_validation_and_hook` in `tests/test_mental_model_hooks.py` asserting that nested operations skip pre-validation and completion hooks.
5. **Linter & Formatting**: Formatted with `ruff format` and verified with `ruff check` (all 14 tests pass cleanly in 1.22s).

cc @nicoloboschi @benfrank241 @r266-tech @Sanderhoff-alt — ready for re-review!

r266-tech (commented):
Re-reviewed exact head `40a251c3dbca7be64822d630303c011c84735deb`.

The two blockers from my earlier reviews are resolved: structured history values are now measured deterministically inside the non-fatal completion boundary, and the repository-wide source-text coverage test has been removed. The nested-operation guard is applied symmetrically to validation and completion, with focused coverage for direct, preauthorized, nested, structured-response, and hook-failure paths.

I don't see a remaining blocking issue in this revision.

Sword-Saint69:
> Re-reviewed exact head `40a251c3dbca7be64822d630303c011c84735deb`.
> 
> The two blockers from my earlier reviews are resolved: structured history values are now measured deterministically inside the non-fatal completion boundary, and the repository-wide source-text coverage test has been removed. The nested-operation guard is applied symmetrically to validation and completion, with focused coverage for direct, preauthorized, nested, structured-response, and hook-failure paths.
> 
> I don't see a remaining blocking issue in this revision.

thanks for the review hope this will be fixed in upcomming versions

nicoloboschi (approved):
Verified and reworked on the branch.

Confirmed the vulnerability is real: with a rejecting operation validator, `get_mental_model` returns 403 while `get_mental_model_history` returned the full revision history — reproduced against a real Postgres on the base commit, fixed on the branch.

Changed the fix so extensions see history as its own operation. It now runs `validate_bank_read` with a new `BankReadOperation.GET_MENTAL_MODEL_HISTORY` rather than reusing `validate_mental_model_get`: an extension has to be able to allow reading a model's current content while refusing the superseded snapshots behind it, and reusing the get would also have metered a history read as one. The `on_mental_model_get_complete` accounting hook is dropped with it — the gate is what #3831 asked for, and summing every revision's `previous_content` into `output_tokens` would have billed a metadata read.

Your `_nested_operation_authorized` guard was right and I kept it: `export_knowledge_base` reads each page's history under its own single `EXPORT_KNOWLEDGE_BASE` gate. I removed it at first and `test_export_gates_once_and_suppresses_nested_reads` caught it immediately.

Tests moved to the DB-backed suite (`TestMentalModelHistory`), following the #3312 precedent, so the real SQL path is exercised — including that an id the caller may not read is refused rather than answered with a 404 existence oracle.

Thanks for the clear report and the fix.

PR #3842, opened 2026-08-27
feat(alembic): let a deployment own a maintenance routine

cdbartholomew:
## The problem

The cross-schema discovery routines — `mental_models_with_cron`,
`banks_needing_consolidation`, `schemas_with_expired_rows` and
`schemas_with_expired_operations` — are installed by migrations and called **by name**
from the maintenance loop. That late binding is genuinely useful: an operator can
`CREATE OR REPLACE` one with an implementation suited to their installation, typically a
registry table maintained by a row trigger, so discovery costs `O(rows-of-interest)`
instead of `O(schemas)`.

On a large multi-tenant install the difference is a sub-millisecond indexed read versus
a serial loop over every schema — once per tick, in every process running the loop,
since the loop has no leader election.

**The replacement is silently undone.** All of these are `CREATE OR REPLACE` against the
same name, so the next migration that reinstalls a routine wins. Nothing in the schema
records that it was deliberately replaced, and nothing in the migration output says a
custom implementation was discarded. The operator finds out from the resulting load.

## The change

`HINDSIGHT_API_EXTERNALLY_OWNED_ROUTINES` — a comma-separated list of routine names the
deployment owns. Migrations skip installing anything named in it:

```
HINDSIGHT_API_EXTERNALLY_OWNED_ROUTINES=mental_models_with_cron
```

Read from the process environment **at migration time**, so it applies to whatever runs
the migration — the CLI, a Job, or an application startup path — without threading a new
option through each entry point.

**Empty by default.** An installation that has not replaced anything behaves exactly as
before; every routine installs as it does today.

### Call-site shape

```python
execute_unless_owned(
    "mental_models_with_cron",
    """
    CREATE OR REPLACE FUNCTION public.mental_models_with_cron()
    ...
    """,
)
```

Shaped like `op.execute` on purpose: guarding an existing migration is a one-line change,
and the SQL keeps its original indentation. That last part matters — the routine body
lands in `pg_proc.prosrc` verbatim, so reformatting it would show up as a spurious
difference to anyone diffing installed routines against the source.

22 call sites across 7 migrations are converted. The SQL itself is untouched.

## What this deliberately does not do

- **Does not stamp the choice into the database.** Ownership is a property of the
  deployment, not the schema. A deployment that stops overriding a routine should get the
  stock one back on its next migration, with no data fix-up.
- **Does not verify a replacement exists.** Naming a routine here and not installing one
  leaves whatever was there before — nothing, on a fresh database. The maintenance loop
  then fails on a missing function rather than quietly running a scan the operator did not
  want. That is the safer direction, and it fails loudly.

## Tests

`test_externally_owned_routines.py` — 13 tests.

The important one walks **every** migration with AST and fails if any installs or drops an
ownable routine through a bare `op.execute`. That is what keeps the guard applied as new
migrations are written, instead of relying on the next author knowing about it.

Verified it fails when a single guard is reverted, and passes when restored:

```
E   AssertionError: These migrations install or drop an ownable maintenance routine
    with a bare op.execute, which silently overwrites a deployment's own implementation.
    Use execute_unless_owned(<name>, <sql>) instead:
      f4d1c2b3a5e6_add_scheduled_mental_model_refresh_routine.py: mental_models_with_cron
```

The rest cover parsing (spacing, empty entries, partial-name safety), that the environment
is read at call time rather than import time (migrations run long after import), and that
`execute_unless_owned` both skips and runs as appropriate.

`ruff check` and `ruff format --check` pass on every file touched. Existing migration
tests pass; the 5 errors in `test_migration_drop_v1_structured_content.py` are a
pre-existing `ModuleNotFoundError: No module named 'pg0'` that reproduces identically on
an unmodified `main`.


nicoloboschi:
## Review

The design holds up. Late binding on these four routines is real leverage on a large
multi-tenant install, `CREATE OR REPLACE` genuinely does clobber a replacement silently,
and the AST test is the right shape — it makes the guard survive the next author who
doesn't know about it, which a convention alone would not.

Two things fixed on the branch (pushed as `fix: repair the maintenance-routine tests and
document the ownership flag`).

### 1. `test-api (2/3)` was red: 13 real failures

`tests/test_maintenance_routines.py` captures a migration's SQL by patching
`migration.op.execute`. Dropping `op` from the seven converted modules' imports removed
that attribute:

```
AttributeError: module '_maint_routines_schema_local' has no attribute 'op'
```

13 tests across `test_routines_install_in_the_configured_schema`,
`test_downgrade_*`, `test_tenant_runs_install_nothing_and_clean_up_strays`,
`test_routines_callable_from_non_public_schema` and the `expired_operations` pair.

The sink moved into `_owned`, so the capture helpers patch `_owned.op` — the same
`alembic.op` proxy those modules used to hold, so nothing else changes. They now also
clear `HINDSIGHT_API_EXTERNALLY_OWNED_ROUTINES`: with it set in the environment the guard
would swallow every statement and each of those assertions would pass vacuously against an
empty list.

Added `test_an_owned_routine_is_left_alone_by_a_real_migration`, which drives
`b6d2f8a4c1e7._pg_upgrade` with `mental_models_with_cron` owned and asserts the other two
still install. The unit tests prove the guard works and the AST test proves it is applied;
neither exercises the wiring between them.

### 2. The flag was undocumented

A new `HINDSIGHT_API_*` variable with no entry in `.env.example` or
`configuration.md`. Added both (plus the `hindsight-embed` copy the sync test compares
byte-for-byte), with a section covering three things the PR description knows but an
operator would have to discover: the replacement must match the stock signature, naming a
routine you have not installed leaves it missing, and removing a name reinstalls the stock
routine on the next migration.

### Note, not fixed

`b6d2f8a4c1e7._drop_stray_copies` and the non-install branch of
`d7b2f8a1c934._pg_upgrade` are guarded too, but those drops target *tenant* schemas —
by construction never where the deployment's routine lives. An owning deployment therefore
keeps the dead per-tenant duplicates those paths exist to clean up.

Leaving it: the blast radius is a few inert functions on a database that ran an interim
build of an unreleased branch, and the alternative — a second entry point the AST test
cannot see — buys less than the uniform "migrations never touch a name you own" rule
costs here.

### Also checked

- `logger.info` on a skip does reach migration output: `hindsight-admin` sets
  `basicConfig(level=INFO)` on the root logger, and `_owned`'s propagates.
- Oracle slots are untouched — correctly, these routines are PG-only.
- `_TOUCHES` won't match a bare `CREATE FUNCTION` (no `OR REPLACE`) or `ALTER FUNCTION`,
  and `_unguarded_calls` only inspects `op.execute`, not
  `op.get_bind().exec_driver_sql` — the form CLAUDE.md recommends for Oracle. Both are
  narrow enough to leave, given these routines never take the Oracle path.

PR #3843, opened 2026-08-27
fix(coding-agents): stop opencode's install from replacing a JSONC config

etnperlong:
## Summary

opencode reads its global config from either `~/.config/opencode/opencode.json` or `~/.config/opencode/opencode.jsonc`. The installer knew only the first name, and read it with `readJson`, whose `JSON.parse` rejects the comments and trailing commas a hand-maintained config carries, and whose `catch` returns `{}`. My config is `opencode.jsonc`, so `install opencode` wrote a second config file next to the real one. Had it been named `.json`, the `{}` fallback would have gone back to disk carrying only our `plugin` key, taking the providers, MCP servers and agent overrides with it. The `<file>.hindsight-backup` copy that `writeJson` takes on first touch is the only reason a config in that state is recoverable.

Five commits, each one step:

1. `opencodeConfigPath` probes `opencode.jsonc` then `opencode.json`, edits whichever already exists, and creates `opencode.json` only when neither does. Same shape as `kiloConfigPath`.
2. The adapter parses with `parseJsonc` and aborts with a `SKIPPED` log on a file it cannot read, rather than falling back to `{}`. kilo already did this.
3. `writeJsonc` sets or deletes one top-level key through `jsonc-parser`, so text it did not touch survives byte for byte. Parsing was only half of it: a config that read correctly still came back stripped, because `writeJson` round-trips through `JSON.stringify`.
4. opencode writes through `writeJsonc`.
5. kilo does too. It had the JSONC-aware read and the lossy write, and its test asserted the loss ("comments dropped, DATA kept"). This commit also adds the parity guard.

`jsonc-parser` goes in tsup's `noExternal`. `installer.js` is staged to `~/.hindsight/coding-agents` as dist + skill + package.json and never `node_modules`, so an external import there cannot resolve; without it, re-running `install` from the staged copy exits with `ERR_MODULE_NOT_FOUND`. I hit that while testing and it is why the dependency is inlined.

Comments *inside* the `plugin` array are still lost, since that value gets re-emitted. Everything outside it is kept.

No issue filed for this. I ran into it installing the plugin on my own machine.

## Type of change

- [ ] Feature
- [x] Bug fix
- [ ] Hotfix
- [ ] Spike / exploration
- [ ] Documentation
- [ ] Refactor

## Test plan

- `src/installer.test.ts`: opencode gains coverage for editing an existing `.jsonc`, uninstalling from the same file, creating `.json` when neither exists, keeping a commented config's other keys, parsing a trailing-comma `.json`, and leaving an unparseable file alone. I checked these against the pre-fix installer: three of them fail there with the reported symptom.
- A parity block drives a real install and uninstall for every JSONC host and asserts the comments survive both. Reverting either adapter to `writeJson` makes it fail, which is how I confirmed it would have caught the kilo half.
- The kilo test that asserted comments were dropped now asserts they are kept.
- `npm test`: 669 passed across 55 files. `npx tsc --noEmit` and `npm run build` clean. `./scripts/hooks/lint.sh` clean.
- Ran the built installer against a copy of my real `opencode.jsonc` (4.7 KB, four providers, ten agent overrides, comments throughout): no competing `opencode.json`, and `diff` shows changes confined to the `plugin` array.

## Notes for the reviewer

- `parseJsonc` is now used for `opencode.json` as well as `.jsonc`. Comments turn up under both names, so the file content decides the parser rather than the extension. If you would rather keep strict parsing for `.json`, that is a one-line change, but it leaves the original data loss in place for anyone whose commented config uses that name.
- `writeJsonc` takes a single key instead of an object on purpose. A whole-object write is exactly what forces the re-serialize, and both callers only ever touch `plugin`.
- Strict-JSON hosts (claude-code, cursor, devin, and the rest) and the TOML and YAML ones stay on `writeJson`. The parity test lists the JSONC hosts explicitly so a new one has to be added there.
- `OPENCODE_CONFIG_CANDIDATES` is exported to mirror `KILO_CONFIG_CANDIDATES`. Neither is imported anywhere else.

### DIFF
--- a/hindsight-integrations/coding-agents/package-lock.json
+++ b/hindsight-integrations/coding-agents/package-lock.json
@@ -10,6 +10,7 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
         "@vectorize-io/hindsight-all": "^0.8.6",
+        "jsonc-parser": "^3.3.1",
         "zod": "^4.4.3"
       },
       "bin": {
@@ -2196,6 +2197,12 @@
       "resolved": "https://registry.npmjs.org/json-schema-typed/-/json-schema-typed-8.0.2.tgz",
       "integrity": "sha512-fQhoXdcvc3V28x7C7BMs4P5+kNlgUURe2jmUT1T//oBRMDrqy1QPelJimwZGo7Hg9VPV3EQV5Bnq4hbFy2vetA==",
       "license": "BSD-2-Clause"
+    },
+    "node_modules/jsonc-parser": {
+      "version": "3.3.1",
+      "resolved": "https://registry.npmjs.org/jsonc-parser/-/jsonc-parser-3.3.1.tgz",
+      "integrity": "sha512-HUgH65KyejrUFPvHFPbqOY0rsFip3Bo5wb4ngvdi1EpCYWUQDC5V+Y7mZws+DLkr4M//zQJoanu1SP+87Dv1oQ==",
+      "license": "MIT"
     },
     "node_modules/kubernetes-types": {
       "version": "1.30.0",
--- a/hindsight-integrations/coding-agents/package.json
+++ b/hindsight-integrations/coding-agents/package.json
@@ -75,6 +75,7 @@
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.29.0",
     "@vectorize-io/hindsight-all": "^0.8.6",
+    "jsonc-parser": "^3.3.1",
     "zod": "^4.4.3"
   },
   "devDependencies": {
--- a/hindsight-integrations/coding-agents/src/installer.test.ts
+++ b/hindsight-integrations/coding-agents/src/installer.test.ts
@@ -11,7 +11,7 @@ import {
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { pathToFileURL } from "node:url";
-import { INSTALLERS, MARKER, run, type InstallCtx } from "./installer";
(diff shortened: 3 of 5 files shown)


nicoloboschi (approved):
Real data loss on the user's own machine, and the fix addresses all three layers rather than the one that was reported.

The read was the reported half: `readJson`'s `JSON.parse` rejects the comments and trailing commas a hand-maintained config carries, and its `catch` returns `{}`. Aborting with a `SKIPPED` log on an unparseable config — matching what kilo already did — is the right shape; a config we merely failed to parse is not a broken config.

The write was the half that would have been easy to miss: `writeJson` round-trips through `JSON.stringify`, so even a config read correctly came back stripped. Computing a minimal `jsonc-parser` edit for one top-level key, and keeping it deliberately single-key because a general merge helper cannot preserve anything, is the correct scope. Preserving CRLF is a nice touch — reflowing line endings would turn a one-key change into a whole-file diff on Windows.

Probing `.jsonc` before `.json` matches `kiloConfigPath` and stops the second-config-file outcome. Fixing kilo's lossy write in the same change, and replacing the test that asserted the loss ("comments dropped, DATA kept") with a parity guard over both JSONC hosts, is what stops the next sibling from regressing it.

`jsonc-parser` in tsup's `noExternal` is necessary, not incidental: `installer.js` is staged to `~/.hindsight/coding-agents` without `node_modules`, so an external import breaks re-running `install` from the staged copy.

Known limitation is documented and acceptable: comments inside the `plugin` array are still lost, since that value is re-emitted.

All six checks green, including `test-coding-agents`, `verify-generated-files` and `check-integration-lockfiles`.

PR #3844, opened 2026-08-27
fix(admin): sync identity sequences after restore

### LINKED ISSUE #3841: [Bug] hindsight-admin restore leaves identity sequences behind restored data — post-restore inserts fail with UniqueViolationError
### Bug Description

`hindsight-admin restore` restores table rows (with their original explicit IDs) but never advances the PostgreSQL sequences backing `GENERATED BY DEFAULT AS IDENTITY` columns. After a restore, the next insert that relies on the identity default collides with a restored row and fails with `UniqueViolationError`.

Concretely, the Hindsight schema has two sequence-backed columns:

- `public.observation_history.id` (`bigint GENERATED BY DEFAULT AS IDENTITY`)
- `public.mental_model_history.id` (same)

After restoring a backup, background consolidation started failing on every run:

```
ERROR - hindsight_api.engine.memory_engine - Task execution failed: consolidation, error: UniqueViolationError: duplicate key violates unique constraint "observation_history_pkey"
ERROR - hindsight_api.engine.memory_engine - Not retrying task consolidation (deterministic failure): UniqueViolationError
```

In my case the `observation_history` sequence was still at `11` while the restored table's `max(id)` was `337`. Every consolidation run failed until I fixed it by hand:

```sql
SELECT setval(pg_get_serial_sequence('public.observation_history', 'id'),
              (SELECT MAX(id) FROM public.observation_history));
```

Note the "Not retrying (deterministic failure)" part — the task is marked permanently failed, so a restored instance silently stops consolidating until the operator intervenes.

### Steps to Reproduce

1. Run a Hindsight instance long enough to accumulate `observation_history` rows (so `max(id)` is well past the sequence start).
2. `hindsight-admin backup backup.zip`
3. On a fresh database (schema migrated, tables empty): `hindsight-admin restore backup.zip --yes`
4. Let consolidation run (or write anything that inserts into `observation_history` without an explicit id).

### Expected Behavior

After restore, identity/serial sequences are advanced past the restored data — e.g. for each sequence-backed column, `setval(seq, GREATEST(max(id), 1))` — so subsequent inserts work.

### Actual Behavior

Sequences keep their pre-restore values. First insert via the identity default raises `UniqueViolationError` once the sequence reaches an existing id. Consolidation hits this repeatedly and each failure is marked deterministic/non-retryable.

### Root cause

In `hindsight_api/admin/cli.py` (verified on v0.9.2 and current `main`):

- `_restore()` does `TRUNCATE TABLE ... CASCADE` (no `RESTART IDENTITY`, which would also be wrong here) followed by binary `COPY` of rows with explicit IDs, then `REFRESH MATERIALIZED VIEW`. No `setval` / `pg_get_serial_sequence` / `ALTER SEQUENCE` anywhere in the package.
- `manifest.json` records only table names, row counts and column metadata — no sequence state.

So neither the backup format nor the restore path carries or rebuilds sequence positions.

### Suggested fix

At the end of `_restore()` (inside the same transaction), discover sequence-backed columns and sync them, e.g.:

```sql
SELECT c.table_name, c.column_name,
       pg_get_serial_sequence(quote_ident(c.table_schema)||'.'||quote_ident(c.table_name), c.column_name)
FROM information_schema.columns c
WHERE c.table_schema = $1
  AND (c.column_default LIKE 'nextval%' OR c.is_identity = 'YES');
```

then `SELECT setval($seq, COALESCE((SELECT MAX(col) FROM t), 1))` per result.

### Version

hindsight-api-slim / hindsight-admin 0.9.2 (also verified unfixed on `main` as of 2026-08-27). PostgreSQL 18.1 (pg0-embedded).

### LLM Provider

Other (DeepSeek) — not relevant to the bug.

r266-tech:
Fixes #3841.

Admin restore copies explicit IDs back into the history tables but previously left their identity sequences at the pre-restore position. The next consolidation write could then collide with restored rows and fail permanently.

This change discovers non-cyclic PostgreSQL identity sequences owned by restored columns, then synchronizes them inside the existing restore transaction. It uses catalog-constrained identity dependencies, quoted identifiers, transactional `ALTER SEQUENCE ... RESTART`, and an in-transaction `nextval` to preserve exhausted sequence bounds without non-transactional `setval`. Cyclic and manually-owned non-generating sequences are deliberately left untouched because table extrema cannot reconstruct their cursor safely.

Regression coverage includes both affected history tables with restored ID 337, transaction rollback, empty tables, ascending and descending exhaustion, explicit values outside sequence bounds, cyclic exclusion, quoted identifiers, and manually-owned sequence exclusion.

Verification:
- `pytest -q hindsight-api-slim/tests/test_admin_backup_restore.py` (21 passed)
- `ruff format --check hindsight-api-slim/hindsight_api/admin/cli.py hindsight-api-slim/tests/test_admin_backup_restore.py`
- `ruff check hindsight-api-slim/hindsight_api/admin/cli.py hindsight-api-slim/tests/test_admin_backup_restore.py`
- `git diff --check`


nicoloboschi (approved):
Verified locally: full `tests/test_admin_backup_restore.py` module passes (21) on an isolated pg0, and the regression is real — reverting only `admin/cli.py` to the base commit makes `test_backup_restore_roundtrip` fail. Also probed `_sync_owned_sequences` directly against PG 16: restored max 337 -> next id 338, GENERATED ALWAYS 42 -> 43, truncated table resets to START, a sequence ahead of the restored max moves back without collision, and rolling back the enclosing transaction leaves the sequence untouched. ruff check/format and ty are clean.

The transactional `ALTER SEQUENCE ... RESTART` + in-transaction `nextval` is a better fit than the `setval` the issue suggested, since it rolls back with the restore.

Follow-up (non-blocking): the catalog predicate (`deptype = 'i'` + `attidentity <> ''`) covers identity columns only, so a SERIAL column's sequence is silently skipped. No core table uses SERIAL today, but extension tables from `extra_bank_tables()` are also restored — a sibling guard to `test_backup_tables_covers_entire_schema` asserting every sequence-backed column is an identity column would keep this from regressing.

PR #3846, opened 2026-08-27
feat(obsidian): support observation scopes

omgbabyweb:
## Summary
- add optional named `observation_scopes` support to the Obsidian retain client
- thread the setting through the shared desktop/headless `SyncEngine` path
- expose a backward-compatible plugin dropdown and CLI flag
- preserve source tags and omit the request field when unset

This supports banks where trusted ingestion sources need to consolidate observations together while retaining vault/folder/date/frontmatter tags for filtering and provenance. Existing documents are not automatically migrated when the setting changes.

## Test plan
- `npm run lint`
- `npm test` (113 tests)
- `npm run build`
- `git diff --check`

### DIFF
--- a/hindsight-integrations/obsidian/src/node/cli.ts
+++ b/hindsight-integrations/obsidian/src/node/cli.ts
@@ -80,9 +80,7 @@ const OBSERVATION_SCOPES = ["combined", "shared", "per_tag", "all_combinations"]
 function parseObservationScopes(value: string | undefined): ObservationScopes | undefined {
   if (value === undefined) return undefined;
   if (OBSERVATION_SCOPES.includes(value as ObservationScopes)) return value as ObservationScopes;
-  throw new UsageError(
-    `--observation-scopes must be one of: ${OBSERVATION_SCOPES.join(", ")}`
-  );
+  throw new UsageError(`--observation-scopes must be one of: ${OBSERVATION_SCOPES.join(", ")}`);
 }
 
 /** Parse argv into fully-resolved options (applying env fallbacks + defaults). */
--- a/hindsight-integrations/obsidian/src/node/cli.ts
+++ b/hindsight-integrations/obsidian/src/node/cli.ts
@@ -18,6 +18,7 @@ import { parseArgs } from "node:util";
 import { HindsightClient } from "../client";
 import { SyncEngine, type ReconcileSummary, type SyncConfig } from "../sync";
 import type { Transport } from "../transport";
+import type { ObservationScopes } from "../types";
 import { fetchTransport } from "./fetch-transport";
 import { FsVault } from "./fs-vault";
 import {
@@ -38,6 +39,7 @@ export interface CliOptions {
   exclude: string[];
   vaultName: string;
   prefixDocId: boolean;
+  observationScopes?: ObservationScopes;
   indexPath: string;
   watch: boolean;
   /** Destination the sync index is bound to (issue #3257). */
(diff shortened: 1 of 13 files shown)


hindsight-integrations/obsidian/src/node/cli.ts
```diff
@@ -70,6 +75,16 @@ export class UsageError extends Error {}
...
 export class HelpRequested extends Error {}
 
+const OBSERVATION_SCOPES = ["combined", "shared", "per_tag", "all_combinations"] as const;
+
+function parseObservationScopes(value: string | undefined): ObservationScopes | undefined {
+  if (value === undefined) return undefined;
+  if (OBSERVATION_SCOPES.includes(value as ObservationScopes)) return value as ObservationScopes;
+  throw new UsageError(
+    `--observation-scopes must be one of: ${OBSERVATION_SCOPES.join(", ")}`
+  );
```
Sanderhoff-alt: These files are not Prettier-formatted, which fails the `verify-generated-files` CI job: it runs `./scripts/hooks/lint.sh` (with `CI` set, so every integration is formatted) and then fails on any resulting diff. Verified against the repo's `.prettierrc.json` (printWidth 100) — three spots reflow:

- here (fits on one line);
- `src/settings.ts:149-151` → `this.plugin.settings.observationScopes = v ? (v as ObservationScopes) : undefined;`
- `tests/node/cli.spec.ts:87-92` → the second `expect(...)` breaks after `expect(` instead.

Running `./scripts/hooks/lint.sh` (or `npx prettier --write --config ../../.prettierrc.json .`) and committing the result fixes all three.

```suggestion
  throw new UsageError(`--observation-scopes must be one of: ${OBSERVATION_SCOPES.join(", ")}`);
```

hindsight-integrations/obsidian/src/settings.ts
```diff
@@ -128,6 +130,29 @@ export class HindsightSettingTab extends PluginSettingTab {
         })
       );
 
+    new Setting(containerEl)
+      .setName("Observation consolidation")
+      .setDesc(
+        "How observations are grouped for newly ingested content. Source tags remain available for filtering and provenance. Changing this does not rescope existing memories."
```
Sanderhoff-alt, line 136: This says the setting applies to "newly ingested content", but the plugin has a second retain path that is not threaded: `src/chat.ts:40` and `src/chat.ts:71` retain the user and assistant turns when `rememberConversations` is on, and neither passes `observationScopes`. With the setting on `shared`, note-derived memories consolidate into the global scope while conversation memories stay on the server default — the same split-consolidation the README warns about for a CLI/plugin config mismatch. Either pass `observationScopes` through `ChatTurnDeps` as well, or narrow the wording here and in the README to say it applies to synced notes only.

Sanderhoff-alt (commented):
The field name, the four accepted values and the per-item placement all match `MemoryItem.observation_scopes` in `hindsight-api-slim/hindsight_api/api/http.py:716`, and `saveSettings()` rebuilds the engine so the dropdown takes effect immediately. Two items before merge.

hindsight-integrations/obsidian/src/node/cli.ts
omgbabyweb: Fixed in 788bdbf1b. Ran Prettier with the repository config; the three identified reflows are committed, and the integration now passes the Prettier check.

hindsight-integrations/obsidian/src/settings.ts
omgbabyweb: Fixed in 788bdbf1b. The plugin setting now flows through ChatTurnDeps and is applied to both remembered user and assistant retains. Coverage verifies both retains receive shared; existing tests continue to cover disabled conversation retention and HTTP omission when the scope is unset.

omgbabyweb:
@Sanderhoff-alt Thanks for the review — both items are addressed in 788bdbf1b: the requested Prettier reflows are committed, and observationScopes now flows through remembered user and assistant chat retains with focused coverage. Local verification passed TypeScript lint, all 113 tests, production build, Prettier check, and git diff check. Could you please re-review when convenient?

omgbabyweb:
@strix-security please re-run the security review for the latest commit.

PR #3847, opened 2026-08-27
feat(helm): add Prometheus operator ServiceMonitor support

aWN4Y25pa2EK:
## Problem

The api (port 8888) and worker (port 8889) containers expose Prometheus format metrics at `/metrics`, but the chart wires scraping for neither: the worker Service's scrape annotations are gated behind the unrelated `podAnnotations` value (dead by default), the api Service has no annotations at all, and no `ServiceMonitor` exists. The dev LGTM stack also only scraped the api — worker metrics were not collected in any shipped configuration.

## What changed

- `templates/servicemonitor.yaml` (new): per-component `ServiceMonitor` resources — api whenever `api.enabled`, worker only when `worker.enabled` — gated on `metrics.serviceMonitor.enabled` (default `false`).
- `values.yaml`: `metrics.serviceMonitor` block (`labels` for operator selection, `path`, `interval`, `scrapeTimeout`) with operator prerequisites documented.
- `README.md`: values table rows + "Prometheus metrics" section.
- `scripts/dev/monitoring/prometheus.yml`: add `hindsight-worker` job (`host.docker.internal:8889`).

The worker monitor targets the headless Service, so each StatefulSet replica gets its own scrape target (`hindsight-worker-0/1`). The control plane exposes no metrics (Next.js, no instrumentation) and is intentionally excluded.

## Usage

```yaml
metrics:
  serviceMonitor:
    enabled: true
    labels:
      release: kube-prometheus-stack  # match the stack's serviceMonitorSelector
```

## Validation

- `helm lint` clean; renders verified: 0 monitors by default, 1 with api-only, 2 with `worker.enabled=true`; selector labels stable (`name/instance/component`); user labels land on monitor metadata.
- End-to-end on k3d + kube-prometheus-stack: installed chart with `worker.enabled=true`, `metrics.serviceMonitor.enabled=true`, `metrics.serviceMonitor.labels.release=kps`; both monitors created and reconciled; Prometheus active targets:
  - `hindsight-api http://10.42.0.17:8888/metrics health=up`
  - `hindsight-worker http://10.42.0.18:8889/metrics health=up`
  - `hindsight-worker http://10.42.0.19:8889/metrics health=up`
  - `up == 1` for all three jobs; api served 113 `hindsight_*` samples, worker 28 (db pool, process, poller families).

## Notes

- Requires the `monitoring.coreos.com/v1` CRDs; without the operator nothing is emitted (default off), so plain installs are unaffected.
- Unrelated chart findings from a full assessment (missing `values.schema.json`, broken `existingSecret` + external-DB path, duplicate env entries under SSA, `pgvector:latest` pin) are tracked separately and not addressed here.


helm/hindsight/README.md
```diff
@@ -142,6 +144,31 @@ ingress:
...
+
+The worker monitor requires `worker.enabled: true`. Without the operator,
+scrape `svc/<release>-api:8888/metrics` and `svc/<release>-worker:8889/metrics`
+directly with annotation-based discovery or static targets.
+          service: api
+  tls:
+    - secretName: hindsight-tls
+      hosts:
+        - hindsight.example.com
+```
```
Sanderhoff-alt: These six lines are a stray copy of the tail of the ingress example above and should be deleted. The unmatched closing fence on the last line opens a code block in the rendered README, so everything from `## Upgrading` onwards is swallowed into it and the remaining sections no longer render.

```suggestion
```
The new section should end at "directly with annotation-based discovery or static targets."

scripts/dev/monitoring/prometheus.yml
```diff
@@ -27,4 +27,10 @@ scrape_configs:
     static_configs:
       - targets: ['host.docker.internal:8888']
     metrics_path: '/metrics'
+  - job_name: 'hindsight-worker'
```
Sanderhoff-alt, line 31: Inserting the new job here moves the existing `scrape_interval: 5s` (line 36) off the `hindsight-api` job and onto `hindsight-worker`. It happens to be a no-op today because the global interval is also 5s, but the api job's explicit override is silently gone and will drift if the global value ever changes. Move the new job below `scrape_interval: 5s` so it becomes the last key of the api job again, or repeat the interval on both jobs.

Sanderhoff-alt (commented):
The ServiceMonitor templates and values look right (port name `http` matches both `api-service.yaml` and `worker-service.yaml`, and the worker monitor is correctly gated on `worker.enabled`). Two issues in the surrounding edits, one of which breaks the rendered README.

helm/hindsight/README.md
aWN4Y25pa2EK: Fixed in 332772c54: removed the stray ingress tail, restoring balanced fences and normal rendering from Upgrading onward.

scripts/dev/monitoring/prometheus.yml
aWN4Y25pa2EK: Fixed in 332772c54: moved the explicit 5s scrape interval back onto the hindsight-api job; the worker job now inherits the unchanged 5s global interval.

nicoloboschi (approved):

PR #3850, opened 2026-08-28
fix(embed): read CLI profile config as UTF-8

### LINKED ISSUE #3837: [Bug] #3748 follow-up: hindsight-embed CLI still reads profile .env without encoding (cli.py load_config_file crashes on GBK locale)
### Bug Description

Follow-up to #3747 / PR #3748 (v0.9.2). The UTF-8 fix covered `profile_manager.py`, but the same crash still occurs one frame earlier in the CLI entry path: `hindsight_embed/cli.py` `load_config_file()` opens the profile `.env` with no explicit encoding:

```python
# hindsight_embed/cli.py (v0.9.2, also current main)
if config_path.exists():
    with open(config_path) as f:   # <-- locale-dependent encoding (cp936 on zh-CN Windows)
        for line in f:
```

The profile `.env` template contains UTF-8 em-dashes (`—`, U+2014 = bytes `E2 80 94`) in its help comments. On Windows with a legacy non-UTF-8 system locale (zh-CN/GBK, ja-JP/Shift-JIS, de-DE/CP1252, …), Python's `open()` defaults to the locale codepage, and byte `0x94` is invalid in GBK — so **every** `hindsight-embed` CLI command crashes before it can do anything, including `daemon start`, `daemon status`, and `daemon stop`.

### Steps to Reproduce

1. Windows 11, system locale zh-CN (codepage 936 / GBK; the "Beta: Use Unicode UTF-8 for worldwide language support" option **off**), no `PYTHONUTF8` set.
2. `pip install hindsight-all-slim==0.9.2` in a venv.
3. Have any profile `.env` containing the stock template comments (e.g. `~/.hindsight/profiles/hermes.env` written by the daemon).
4. Run `hindsight-embed -p hermes daemon status` (or any subcommand).

### Expected Behavior

CLI reads the profile env file as UTF-8 regardless of the system locale, consistent with the sites fixed in PR #3748.

### Actual Behavior

```
  ...
  File "...\hindsight_embed\cli.py", line 128, in get_config
    load_config_file()
  File "...\hindsight_embed\cli.py", line 108, in load_config_file
    for line in f:
UnicodeDecodeError: 'gbk' codec can't decode byte 0x94 in position 622: illegal multibyte sequence
```

### Root cause / suggested fix

`open(config_path)` → `open(config_path, encoding="utf-8")` in `load_config_file()`. It may be worth a one-time audit of all remaining `open(` / `read_text(` / `write_text(` call sites in `hindsight_embed` that touch user-facing text files, so this doesn't resurface in a fourth location.

Workaround until then: run with `PYTHONUTF8=1`.

### Version

hindsight-embed 0.9.2 (also verified unfixed on `main` as of 2026-08-27). The original #3747 crash in `profile_manager.py` **is** fixed in 0.9.2 — this is the next unencoded read on the same code path.

### LLM Provider

Other (DeepSeek) — not relevant to the crash.

koriyoshi2041:
## Problem

`load_config_file()` opens the active profile `.env` with the process locale. The generated template contains UTF-8 characters, so profile startup can fail on Windows locales such as GBK or cp1252.

## Fix

- read the CLI profile config explicitly as UTF-8
- add a regression that makes locale-default decoding fail while the UTF-8 path succeeds

Fixes #3837.

## Test

- `uv run pytest tests/test_profile_daemon_config.py -k load_config_file -q` (2 passed)
- full module: 14 passed; 3 unrelated daemon-command tests require local ML dependencies and selected the documented `uvx` fallback
- `uv run ruff check hindsight_embed/cli.py`
- `uv run ruff check tests/test_profile_daemon_config.py --ignore F401` (the module has three pre-existing unused imports)
- `uv run ruff format --check hindsight_embed/cli.py tests/test_profile_daemon_config.py`
- `git diff --check`

## Risk

Low. The profile files are generated and written as UTF-8 already; this makes the read path match that contract. No parsing or precedence behavior changes.

### DIFF
--- a/hindsight-embed/hindsight_embed/cli.py
+++ b/hindsight-embed/hindsight_embed/cli.py
@@ -104,7 +104,7 @@ def load_config_file():
 
     # Load ONLY this profile's config, never fall back to default
     if config_path.exists():
-        with open(config_path) as f:
+        with open(config_path, encoding="utf-8") as f:
             for line in f:
                 line = line.strip()
                 if line and not line.startswith("#") and "=" in line:
--- a/hindsight-embed/hindsight_embed/cli.py
+++ b/hindsight-embed/hindsight_embed/cli.py
@@ -676,7 +676,7 @@ def do_daemon(args, config: dict, logger):
         else:
             # Show last N lines
             try:
-                with open(daemon_log_path) as f:
+                with open(daemon_log_path, encoding="utf-8", errors="replace") as f:
                     lines = f.readlines()
                     for line in lines[-args.lines :]:
                         print(line, end="")
@@ -838,7 +838,7 @@ def do_ui(args, config: dict, logger):
             return 0
         else:
             try:
-                with open(ui_log_path) as f:
+                with open(ui_log_path, encoding="utf-8", errors="replace") as f:
                     lines = f.readlines()
                     for line in lines[-args.lines :]:
                         print(line, end="")
@@ -1057,7 +1057,7 @@ def do_control(args) -> int:
             except KeyboardInterrupt:
                 pass
             return 0
-        with open(log_path) as f:
+        with open(log_path, encoding="utf-8", errors="replace") as f:
             for line in f.readlines()[-getattr(args, "lines", 50) :]:
                 print(line, end="")
         return 0
--- a/hindsight-embed/hindsight_embed/cli.py
+++ b/hindsight-embed/hindsight_embed/cli.py
@@ -255,7 +255,7 @@ def _do_configure_from_env():
 
     from .env_template import render_config
 
-    CONFIG_FILE.write_text(render_config(config_values))
+    CONFIG_FILE.write_text(render_config(config_values), encoding="utf-8")
     CONFIG_FILE.chmod(0o600)
 
     print()
@@ -442,7 +442,7 @@ def _do_configure_interactive(profile_name: str | None = None, port: int | None 
         # Save to default profile
         from .env_template import render_config
 
-        CONFIG_FILE.write_text(render_config(config_dict))
+        CONFIG_FILE.write_text(render_config(config_dict), encoding="utf-8")
         CONFIG_FILE.chmod(0o600)
 
     # Stop existing daemon if running (it needs to pick up new config)
@@ -1216,7 +1216,7 @@ def do_profile_command(args: list[str]) -> int:
                 config_path = CONFIG_FILE
 
             if config_path.exists():
-                for line in config_path.read_text().splitlines():
+                for line in config_path.read_text(encoding="utf-8").splitlines():
                     line = line.strip()
                     if line and not line.startswith("#") and "=" in line:
                         k, v = line.split("=", 1)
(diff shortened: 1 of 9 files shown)


hindsight-embed/hindsight_embed/cli.py
```diff
@@ -104,7 +104,7 @@ def load_config_file():
 
     # Load ONLY this profile's config, never fall back to default
     if config_path.exists():
-        with open(config_path) as f:
+        with open(config_path, encoding="utf-8") as f:
```
Sanderhoff-alt, line 107: Pinning only the read side makes the default profile unreadable on a non-UTF-8 locale. `hindsight-embed configure` writes this exact file via `CONFIG_FILE.write_text(render_config(...))` at `hindsight-embed/hindsight_embed/cli.py:258` and `:445`, with no `encoding=`, so the text goes out in the process locale. `render_config` seeds from `.env.example`, which contains U+2014; both cp1252 (`b'\x97'`) and gbk (`b'\xa1\xaa'`) encode it happily. Before this change the write and read used the same codec and round-tripped; after it, the UTF-8 read raises `UnicodeDecodeError` on those bytes and startup fails. Please pin the writes too:

```suggestion
        with open(config_path, encoding="utf-8") as f:
```
(plus `CONFIG_FILE.write_text(render_config(config_values), encoding="utf-8")` at both write sites — `profile_manager.py:419` already does this).

Sanderhoff-alt (commented):
The read-side change is right, but the fix looks incomplete for the locale it targets: the matching write path and the template load are still locale-dependent, so on a non-UTF-8 default locale this can turn a previously-working default profile into an unreadable one. Details inline.

hindsight-embed/hindsight_embed/cli.py
```diff
@@ -104,7 +104,7 @@ def load_config_file():
 
     # Load ONLY this profile's config, never fall back to default
     if config_path.exists():
```
Sanderhoff-alt, line 106: The description says this fixes #3837, but the same failure remains on other paths reading the same file, so a GBK/cp1252 user still hits it:

- `hindsight-embed/hindsight_embed/env_template.py:24` and `:28` — `bundled.read_text()` / `candidate.read_text()` with no encoding. `env.example` does not decode as gbk (`'gbk' codec can't decode byte 0x94 in position 639`), so `configure` and profile creation raise before any config is written.
- `hindsight-embed/hindsight_embed/cli.py:1219`, `:1271`, `:1316` — `config_path.read_text()` with no encoding, in the config merge/set/unset commands.

Suggest pinning `encoding="utf-8"` on all of them in this PR so the contract holds for every reader and writer of the profile config.

koriyoshi2041:
Good catches — I pinned UTF-8 across the template reads, both default-profile writes, and the merge/set/remove readers in a0f4ef5. I also added a regression that requires the bundled template read to request UTF-8. Focused tests pass (22 passed; 3 unrelated daemon-command tests deselected), along with Ruff, format, and diff checks.

PR #3852, opened 2026-08-28
fix(llm): detect folded vs unfolded reasoning tokens in OpenAI-compatible provider (#3851)

### LINKED ISSUE #3851: output_tokens reported as 0 for gateways that report reasoning tokens separately from completion_tokens
**Version:** hindsight-api v0.9.1 (`ghcr.io/vectorize-io/hindsight:latest`)
**File:** `hindsight_api/engine/providers/openai_compatible_llm.py` — L1094-1096 (non-streaming) and L1414-1415 (streaming)

## Summary

`output_tokens` is logged and recorded as **0** for any OpenAI-compatible provider that reports reasoning tokens *separately* from `completion_tokens`, instead of folded into it. Retain/reflect work correctly; only the log line and the cost metrics are wrong — but the wrong value invites a false "empty extraction" diagnosis.

```python
if thoughts_tokens:
    output_tokens = max(0, output_tokens - thoughts_tokens)
    total_tokens  = max(0, total_tokens  - thoughts_tokens)
```

The surrounding comment states the assumption explicitly: *"OpenAI-compatible providers fold reasoning tokens into `completion_tokens` (and thus `total_tokens`)"*. That holds for OpenAI's own models but not for every OpenAI-compatible gateway.

## Measurement

Same prompt (~3.1k input tokens), same gateway, three models:

| model | `completion_tokens` | `reasoning_tokens` | `prompt_tokens` | `total_tokens` | relation | resulting `output_tokens` |
|---|---|---|---|---|---|---|
| `gemini-3.7-flash-high` | 673 | 909 | 3110 | 4692 | `p + c + r == total` (**separate**) | **0** |
| `gpt-5.4` | 1337 | 48 | 3340 | 4677 | `p + c == total` (folded) | 1289 |
| `gpt-5.6-luna` | 2081 | 407 | 3340 | 5421 | `p + c == total` (folded) | 1674 |

When reasoning is reported separately and exceeds the visible output — routine for "high effort" model variants — the subtraction clamps to 0.

Observed on **426 of 565** `retain_extract_facts` calls in a 30-minute window with a Gemini model behind an OpenAI-compatible gateway. Every one of those batches logged `Insert facts: N units` and the stored units contain real extracted facts, so retain is fine; the log reads:

```
slow llm call: scope=retain_extract_facts, model=openai/<gemini-model>,
  input_tokens=2853, output_tokens=0, total_tokens=3178, time=39.5s, ratio out/in=0.00
```

Note `total_tokens - input_tokens = 325 != 0`, which is the tell that output was produced.

Direct calls to the same gateway/model outside Hindsight return `completion_tokens` 464-673 for the identical payload (plain, `response_format` json_schema, named `tool_choice`, and `stream=true` with `stream_options.include_usage` — all four report non-zero), so this is not the gateway dropping the field.

## Suggested fix

Only subtract when the provider actually folds reasoning into completion:

```python
if thoughts_tokens and (input_tokens + output_tokens == total_tokens):
    output_tokens = max(0, output_tokens - thoughts_tokens)
    total_tokens  = max(0, total_tokens  - thoughts_tokens)
```

Both call sites (L1094 and L1414) carry the same block.

Sword-Saint69:
## Fix

Fixes #3851 (reported by @gutocarollo) where `output_tokens` was logged and recorded as `0` for OpenAI-compatible gateways/proxies (e.g. LiteLLM proxy, Gemini 3.7 Flash, OpenRouter, or vLLM) that report reasoning tokens separately from `completion_tokens`.

Previously, `OpenAICompatibleLLM` unconditionally subtracted `thoughts_tokens` from `output_tokens` (`completion_tokens`). On unfolded responses where `completion_tokens` contains only visible generated text, this subtraction caused `output_tokens` to clamp to `0` whenever `reasoning_tokens > completion_tokens`, leading to false "empty output" telemetry (e.g., `input_tokens=2853, output_tokens=0, total_tokens=3178`).

---

## Technical Solution & Detection Differentiator

Updated `call()` and `complete()` in `openai_compatible_llm.py` to inspect the token structure before subtracting:

```python
is_folded_shape = (
    thoughts_tokens > 0
    and total_tokens > 0
    and abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens
)
if is_folded_shape:
    output_tokens = max(0, output_tokens - thoughts_tokens)
    total_tokens  = max(0, total_tokens  - thoughts_tokens)

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -1096,12 +1096,20 @@ class OpenAICompatibleLLM(LLMInterface):
                 if usage and getattr(usage, "completion_tokens_details", None):
                     thoughts_tokens = getattr(usage.completion_tokens_details, "reasoning_tokens", 0) or 0
                 # OpenAI-compatible providers fold reasoning tokens into
-                # ``completion_tokens`` (and thus ``total_tokens``), but the
-                # TokenUsage contract — and the Gemini provider — treat
-                # ``output_tokens``/``total_tokens`` as visible-only, surfacing
-                # reasoning separately in ``thoughts_tokens``. Subtract so the
-                # two fields don't double-count reasoning (cost over-attribution).
-                if thoughts_tokens:
+                # ``completion_tokens`` (and thus ``total_tokens``) when
+                # reasoning is included in the completion count.
+                # On gateways where reasoning tokens are reported separately,
+                # ``output_tokens`` (completion_tokens) is already visible-only,
+                # so subtracting would incorrectly clamp output to 0 (#3851).
+                #
+                # Detect folded shape: if abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens,
+                # reasoning is folded into output_tokens and must be subtracted out to avoid double counting.
+                is_folded_shape = (
+                    thoughts_tokens > 0
+                    and total_tokens > 0
+                    and abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens
+                )
+                if is_folded_shape:
                     output_tokens = max(0, output_tokens - thoughts_tokens)
                     total_tokens = max(0, total_tokens - thoughts_tokens)
 
@@ -1417,6 +1425,7 @@ class OpenAICompatibleLLM(LLMInterface):
                 usage = response.usage
                 input_tokens = usage.prompt_tokens or 0 if usage else 0
                 output_tokens = usage.completion_tokens or 0 if usage else 0
+                total_tokens = usage.total_tokens or 0 if usage else 0
                 cached_tokens = 0
                 if usage and getattr(usage, "prompt_tokens_details", None):
                     cached_tokens = getattr(usage.prompt_tokens_details, "cached_tokens", 0) or 0
@@ -1424,10 +1433,17 @@ class OpenAICompatibleLLM(LLMInterface):
                 if usage and getattr(usage, "completion_tokens_details", None):
                     thoughts_tokens = getattr(usage.completion_tokens_details, "reasoning_tokens", 0) or 0
                 # See ``call()``: OpenAI-compatible ``completion_tokens`` includes
-                # reasoning, so make ``output_tokens`` visible-only to avoid
-                # double-counting it against ``thoughts_tokens``.
-                if thoughts_tokens:
+                # reasoning only when folded (detected via abs((input + output) - total) < thoughts).
+                # Subtract only in the folded case so unfolded providers (where completion_tokens
+                # is already visible-only) don't clamp output to 0 (#3851).
+                is_folded_shape = (
+                    thoughts_tokens > 0
+                    and total_tokens > 0
+                    and abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens
+                )
+                if is_folded_shape:
                     output_tokens = max(0, output_tokens - thoughts_tokens)
+                    total_tokens = max(0, total_tokens - thoughts_tokens)
 
                 # See ``call()``: record the reasoning and cached counts too, so no
                 # billed token is dropped from the metrics counters.
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -1104,14 +1104,17 @@ class OpenAICompatibleLLM(LLMInterface):
                 #
                 # Detect folded shape: if abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens,
                 # reasoning is folded into output_tokens and must be subtracted out to avoid double counting.
-                is_folded_shape = (
-                    thoughts_tokens > 0
-                    and total_tokens > 0
-                    and abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens
+                # Detect folded shape: if reasoning tokens are present, check if completion_tokens
+                # includes reasoning tokens (folded). If total_tokens is omitted/0, default to folded shape.
+                is_folded_shape = thoughts_tokens > 0 and (
+                    total_tokens <= 0 or abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens
                 )
                 if is_folded_shape:
                     output_tokens = max(0, output_tokens - thoughts_tokens)
-                    total_tokens = max(0, total_tokens - thoughts_tokens)
+
+                # TokenUsage contract specifies total_tokens = input_tokens + output_tokens
+                # (visible-only total, excluding thoughts_tokens).
+                total_tokens = input_tokens + output_tokens
 
                 # Record LLM metrics. ``output_tokens`` is visible-only by now, so
                 # ``thoughts_tokens`` has to be recorded alongside it or the reasoning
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by koriyoshi2041 — CHANGES_REQUESTED** [review]
The unfolded branch preserves visible `output_tokens`, but it leaves provider `total_tokens` unchanged. In the added fixture that returns 4692, `TokenUsage` becomes input=3110, output=673, thoughts=909, total=4692. That violates the local contract in `response_models.TokenUsage`: `total_tokens` is input + visible output and excludes thoughts (the Gemini provider follows the same contract). It also makes aggregation inconsistent because callers can no longer expect `total_tokens == input_tokens + output_tokens`.

Please normalize `total_tokens` to `input_tokens + output_tokens` after deciding folded vs unfolded, and assert 3783 for the unfolded fixtures. Keep `thoughts_tokens=909` separately for billed/reasoning attribution.

File: hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py [hunk 1]
```diff
@@ -1096,12 +1096,20 @@ async def call(
...
+                # Detect folded shape: if abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens,
+                # reasoning is folded into output_tokens and must be subtracted out to avoid double counting.
+                is_folded_shape = (
+                    thoughts_tokens > 0
+                    and total_tokens > 0
+                    and abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens
+                )
+                if is_folded_shape:
                     output_tokens = max(0, output_tokens - thoughts_tokens)
                     total_tokens = max(0, total_tokens - thoughts_tokens)
```
Sanderhoff-alt: In the unfolded case `total_tokens` keeps the provider value, which for the shape in #3851 is `prompt + completion + reasoning` (3110 + 673 + 909 = 4692). `TokenUsage.total_tokens` is documented as "Total tokens (input + output, excludes thoughts)", so this now over-reports by the reasoning count, and the `slow llm call` line logs `input_tokens=3110, output_tokens=673, total_tokens=4692`. Before this change total was right (4692 - 909 = 3783) and only `output_tokens` was wrong, so the fix trades one misreported field for another.

The normalised value is the same in both shapes, so it can be set unconditionally:

```python
if is_folded_shape:
    output_tokens = max(0, output_tokens - thoughts_tokens)
total_tokens = input_tokens + output_tokens
```

The new test asserting `total_tokens == 4692` would need updating with it. Worth noting too that a gateway that omits `total_tokens` now falls into the unfolded branch via the `total_tokens > 0` guard, so a folded provider without a total would double-count reasoning; a fallback to the folded assumption when the total is absent keeps that case correct.

hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
```diff
@@ -1417,17 +1425,25 @@ async def call_with_tools(
...
+                # Subtract only in the folded case so unfolded providers (where completion_tokens
+                # is already visible-only) don't clamp output to 0 (#3851).
+                is_folded_shape = (
+                    thoughts_tokens > 0
+                    and total_tokens > 0
+                    and abs((input_tokens + output_tokens) - total_tokens) < thoughts_tokens
+                )
+                if is_folded_shape:
                     output_tokens = max(0, output_tokens - thoughts_tokens)
+                    total_tokens = max(0, total_tokens - thoughts_tokens)
```
Sanderhoff-alt: Same `total_tokens` point applies here. Separately, this reassignment is never read: `LLMToolCallResult` carries no total field and nothing below uses the variable after the detection above, so the line is dead as written.

hindsight-api-slim/tests/test_token_usage_cached_thoughts.py
```diff
@@ -377,3 +377,71 @@ async def test_openai_compatible_metrics_account_for_every_billed_output_token()
...
+    )
+    with patch("hindsight_api.engine.providers.openai_compatible_llm.get_metrics_collector"):
+        _, token_usage = await llm.call(
+            messages=[{"role": "user", "content": "Query"}],
+            response_format=_OkModel,
+            max_retries=0,
+            return_usage=True,
+        )
+    assert token_usage.output_tokens == 1337 - 48
+    assert token_usage.thoughts_tokens == 48
```
Sanderhoff-alt, line 446: The file ends with two trailing blank lines, which `ruff format` strips. The `verify-generated-files` job runs `./scripts/hooks/lint.sh` and fails on the resulting diff, so please run it and commit the result.

Sanderhoff-alt (commented):
The folded/unfolded detection matches the three usage shapes in #3851, but the unfolded branch now leaves `total_tokens` including reasoning, which was correct before this change. There is also a ruff-format failure in the test file.

hindsight-api-slim/tests/test_token_usage_cached_thoughts.py
Sword-Saint69: ok sir for sure i will fix it in the next change thanks for your insights i am still a beginner and i love opensource thanks

Sword-Saint69:
Thanks @koriyoshi2041 and @Sanderhoff-alt for the thorough review!

Updated in commit [`88fa2bc1`](https://github.com/Sword-Saint69/hindsight/commit/88fa2bc1):

1. **Contract Normalization**: `total_tokens` is now normalized unconditionally to `input_tokens + output_tokens` (visible-only total tokens, excluding `thoughts_tokens`). For the unfolded fixture ($P=3110, C=673, R=909$), `total_tokens` is reported as `3783` ($3110 + 673$), satisfying `TokenUsage.total_tokens` contract while keeping `thoughts_tokens = 909` separately for cost attribution.
2. **Omitted `total_tokens` Fallback**: Handled proxies that omit `total_tokens` ($T \le 0$) by defaulting to folded shape (`is_folded_shape = True`), preventing folded reasoning double-counting when $T$ is absent.
3. **Cleaned Tool Calling Path**: Removed dead `total_tokens` reassignment in `complete()`.
4. **Formatting**: Formatted all modified files with `ruff format` and verified with `ruff check` (all checks passed).

koriyoshi2041 (approved):
Rechecked `88fa2bc1`: the unfolded fixture now preserves 673 visible output tokens and normalizes `total_tokens` to 3783, while folded and missing-total cases remain covered. The focused token-usage module passes 18/18, Ruff passes, and `git diff --check` is clean.

nicoloboschi:
Thanks for digging into this @Sword-Saint69 — the diagnosis is right and the two call sites you identified are exactly the ones that carry the bug.

I'm closing this in favour of a fix that avoids the shape-detection heuristic. `abs((input + output) - total) < thoughts_tokens` re-opens the bug under the same proxy variance your own third test documents: for the issue's Gemini case (`prompt=3110, completion=673, reasoning=909`), a `total_tokens` of 4691 instead of 4692 gives `abs(3783 - 4691) = 908 < 909`, so it classifies as folded and clamps output back to 0. The tolerance is widest precisely where reasoning is largest, which is the population #3851 is about.

`xai_oauth_llm.py::_token_counts` already solves this exactly, and its docstring cross-references this file: `total_tokens` reads as `prompt + visible + reasoning` under *both* shapes, so deriving `output = (total - reasoning) - prompt` needs no detection at all and is correct for every vector in the issue. The replacement lifts that into a shared helper used by both call sites here, and keeps your `total_tokens <= 0` fallback, which that helper needed.

Your unfolded-shape tests carry over essentially as written. Replacement PR: #3920

PR #3854, opened 2026-08-28
fix(llm): surface truncated compatible responses

koriyoshi2041:
## Summary

- map `finish_reason="length"` from OpenAI-compatible `.create()` responses to the canonical `OutputTooLongError`
- avoid retrying the same truncated JSON response as a generic parse failure
- add a regression proving the provider does not retry the request

This restores the existing retain auto-split path for the truncation case described in #3811.

## Tests

- `uv run pytest tests/test_openai_compatible_response_hardening.py -q` (9 passed)
- `./scripts/hooks/lint.sh`
- pre-commit hooks, including unused-code checks and lint
- `git diff --check`

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py
@@ -322,6 +322,8 @@ def _content_or_error(response: Any, *, provider: str, model: str, scope: str) -
     choice = _first_choice_or_error(response, provider=provider, model=model, scope=scope)
     message = _message_for_choice(choice)
     finish_reason = _finish_reason_for_choice(choice)
+    if finish_reason == "length":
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Thanks for this — and for the review on #3827 that produced the very ordering fix this PR implements (raising ahead of the empty-content branch).

Closing as superseded: #3827 merged this morning (7ef98859e) and already raises `OutputTooLongError` on `finish_reason == "length"` in `_content_or_error`, ahead of the content read, with regression coverage in `tests/test_openai_compatible_truncation.py` (including the no-retry `call_count == 1` assert on both the structured and free-form paths).

GitHub reports this branch as mergeable, but that is a textual auto-merge only — the result would carry two `finish_reason == "length"` checks, leaving the merged one unreachable behind its explanatory comment, and the surviving error message drops the `provider/model/scope` context that the rest of the function reports.

If you rebase on main you should find the behaviour already in place. Much appreciated.

PR #3855, opened 2026-08-28
feat: export knowledge base in document transfers

nicoloboschi:
## Summary
- add a single `include_knowledge_base` API flag for Mental Models + Knowledge Pages in document transfers
- preserve IDs where possible, remap colliding Knowledge Page IDs, restore Mental Models before pages, and regenerate embeddings
- add the Control Plane checkbox and update OpenAPI/generated clients/wrappers

## Validation
- `./scripts/hooks/lint.sh`
- `npm test -- --run tests/lib/api.test.ts` (8 passed)
- `npm run build` (passed)
- focused document-transfer tests (2 passed)
- full document-transfer suite: 47 passed; 1 pre-existing failure on `test_bank_roundtrip_remaps_mental_model_based_on_ids`, reproduced on `origin/main`

## Dev
- API: http://localhost:55075
- Control Plane: http://localhost:55128


nicoloboschi:
## Review

Feature logic looks sound. Pinning the carried mental-model rows to the destination bank is right — `mental_models` PK is `(bank_id, id)` (`w8r9s0t1u2v3`), so `ON CONFLICT` genuinely could have matched the source-bank row on a same-schema transfer. The knowledge-page collision remap is correctly scoped to `bank_id <> $2` (ids are schema-global there, unlike mental models) and parent links are rewritten through the same map, so the topological insert stays consistent.

Three CI failures, all fixed in 187bd475f:

**1. Deno client break (`test-typescript-client-deno`) — the interesting one.**
`npm run generate` is `npx @hey-api/openapi-ts` with no version, so it ignored the pinned `0.97.3` and regenerated against a newer generator. That rewrote the whole client (`index.ts` alone lost 672 lines) and flattened `beforeRequest`, which shifted the indentation of the block `generate-clients.sh` patches for Deno compatibility (#607). That patch step only *warns* on a stale anchor:

```
print("  ⚠ Could not find expected pattern in client.gen.ts - skipping patch")
```

so the fix was silently dropped and the `client` field went back into `RequestInit` → `TypeError: Failed to construct 'Request': Argument 2 'client' must be a Deno.HttpClient`. The script's own comment predicted exactly this. I reverted the generated tree to `main` and re-applied only the two real edits (`include_knowledge_base` in `types.gen.ts`, docstring in `sdk.gen.ts`).

Worth a follow-up: pin the version in the `generate` script, or make the patch step exit non-zero on a missed anchor so this fails loudly at generation time rather than in Deno CI.

**2. Locale catalog parity (`build-control-plane`, `test-hindsight-all`).** The two new `exportIncludeKnowledgeBase*` keys landed in `en.json` only; added to the 9 other catalogs using each locale's existing terminology.

**3. `skills/hindsight-docs/references/openapi.json` drift.** CI regenerates and diffs it; `hindsight-docs/static/openapi.json` was updated but the skill copy wasn't.

**Also changed:** the manifest gained `knowledge_base_count` (mental models + pages summed) while the already-existing `mental_model_count` / `knowledge_page_count` stayed 0 on this path — so a documents archive that *did* carry a knowledge base reported 0 for both. Dropped the aggregate and populated the two existing fields instead.

**Not fixed (pre-existing, flagging only):** `_restore_knowledge_pages` does `inserted += 1` unconditionally despite `ON CONFLICT DO NOTHING`, so a re-import reports the full page count as imported. Harmless before, but this PR surfaces it to API clients as `knowledge_pages_imported`. The per-page collision `fetchval` is also N+1, though page counts are small enough that it doesn't matter yet.

PR #3859, opened 2026-08-28
Fuse observation graph expansion into one database fetch

### LINKED ISSUE #3857: Observation graph expansion performs an avoidable second SQL round trip
## Problem

Hindsight 0.9.1 `parallel_retrieval` takes about 17 seconds for a 49,206-unit memory bank with about 1.3 million links. Trace evidence identifies observation graph/link expansion as the dominant phase.

## Current cause

`_expand_observations` performs entity/source traversal and semantic/causal expansion as two normal SQL executions. The intermediate UUID relation is materialized in the client and sent back to the database.

## Proposed boundary

Fuse the existing observation entity/source relation into the semantic/causal query as a CTE while preserving predicates, score expressions, ordering, limits, time windows, duplicate handling, and PostgreSQL/Oracle behavior.

## Verification

- focused fixtures prove identical ordered IDs, scores, and counts for overlapping sources, duplicate entities, empty sources, multiple fact types, and active time windows
- normal observation expansion performs one fetch instead of two
- the fixed recall benchmark records before/after p50 and p95 plus result equivalence

## Origin

Tracked by mutmutco/Jerv-Hub#652 after the JervLife Voyage migration estate audit.

jervaise:
## Summary

- fuse observation entity/source expansion into the existing semantic/causal CTE query
- reduce normal observation expansion from two database fetches to one
- preserve per-arm predicates, scores, limits, ordering, time-window binds, and PostgreSQL/Oracle dialects
- add result-equivalence, one-fetch, temporal-bind, source-less seed, and backend SQL-shape regressions

## Verification

- `uv run ruff format --check ...` (4 changed Python files)
- `uv run ruff check ...` (4 changed Python files)
- pure PostgreSQL and Oracle one-statement tests: 2 passed
- database-backed focused tests are delegated to Linux CI because Windows `initdb` rejects the host locale `Turkish_Türkiye.1252` before migrations

## Performance boundary

The change removes one client/database round trip and intermediate UUID
materialization per observation expansion call without changing the candidate
relation or result scoring.

*Measured by a maintainer before merge* (PostgreSQL 16, project migrations and
indexes, paired A/B with both shapes alternating in one process against the same
seed sets, `work_mem` at the 4MB default):

| corpus | current (2 statements) | this PR (1 fused) | paired delta |
|---|---|---|---|
| 1k observations, 11k links | p50 38.7 ms | p50 35.0 ms | **−5.2 ms ± 2.2** |
| 20k observations, 1.2M links | p50 9244 ms | p50 9061 ms | −315 ms ± 451 (CI spans zero) |

The saving is a **constant ~5 ms**, not a proportional one, and `EXPLAIN (ANALYZE)`
attributes almost all of it to planning one statement instead of two
(3.32 ms → 1.86 ms); server-side execution is unchanged (41.7 ms → 40.2 ms). So it
is a visible 12% on a young bank and statistically invisible on a large one.

Volume safety was checked across both scaling dimensions — sources per observation
(5 / 20 / 80 / 200) and entity hub degree (3000 / 500 / 100 entities, and 8
entities per fact). All eight cells came back "no difference" with confidence
intervals straddling zero, the entity arm's plan is structurally identical
standalone vs. fused, and temp I/O is identical to the byte (103.42 MB/call at
200 sources/observation, both shapes). No missing index is exposed by the change.

Result-equivalence was verified independently of the bundled tests: 40 seed sets,
all three arms, identical ids, scores and ordering.

## Follow-up

This does not reduce the cost of observation expansion, so it does not close
#3857. The entity arm's candidate generation is still bank-wide — at 20k
observations the `candidates` CTE scans 19,966 rows because `connected_sources`
reaches ~87% of world facts under `per_entity_limit=200`. Cost is driven by
source-fact growth, not entity fan-out: 308 ms → 1126 ms and 0 → 103 MB of temp
I/O per call as sources per observation goes 5 → 200 (the unbounded
`source_memory_ids` append, #1725). That remains the work worth doing.

Refs #3857

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_oracle.py
@@ -697,6 +697,14 @@ class OracleOps(DataAccessOps):
         # Previously used JSON_TABLE to explode source_memory_ids CLOB. The junction
         # table approach uses standard SQL joins, identical to the PG backend.
         #
+        # Entity/source traversal and semantic/causal expansion run as ONE query
+        # (#3857): the observation entity arm is fused into the semantic/causal CTE
+        # query behind an 'entity' source discriminator, like the non-observation
+        # combined expansion, so a normal call performs one fetch instead of two.
+        # Every predicate, score expression, ordering, limit, and the window bind
+        # positions are exactly the two previous statements', now sharing one
+        # snapshot. The caller splits the unioned rows by `source`.
+        #
         # Two PostgreSQL fixes are deliberately NOT mirrored here, because neither
         # was measured against Oracle and both are tuned to PostgreSQL's planner:
         #   - #3085 made PG score set-wise; the scoring below is still the
@@ -714,7 +722,7 @@ class OracleOps(DataAccessOps):
         from ..schema import fq_table
 
         obs_sources_table = fq_table("observation_sources")
-        entity_rows = await conn.fetch(
+        all_rows = await conn.fetch(
             f"""
             WITH seed_sources AS (
                 SELECT DISTINCT os.source_id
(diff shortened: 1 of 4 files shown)


nicoloboschi:
Reviewed, and I benchmarked this before merging since the PR noted that live measurement was still outstanding.

**The change is correct.** I verified result-equivalence independently of the bundled tests — 40 seed sets, all three arms, identical ids, scores and ordering — and ran your test file against a real PostgreSQL: 6/6 pass. It merges cleanly and `main` hasn't touched any of the three source files since this branch's base.

**On the performance claim**, the fusion is worth a *constant* ~5 ms, not a proportional saving, and `EXPLAIN (ANALYZE)` puts almost all of it in planning one statement instead of two (3.32 ms → 1.86 ms) rather than in execution (41.7 ms → 40.2 ms) or the round trip. That's a visible 12% on a small bank and statistically invisible on a large one:

| corpus | current | this PR | paired delta |
|---|---|---|---|
| 1k observations, 11k links | p50 38.7 ms | p50 35.0 ms | −5.2 ms ± 2.2 |
| 20k observations, 1.2M links | p50 9244 ms | p50 9061 ms | −315 ms ± 451 (CI spans zero) |

**Volume safety checked**, since fusing could in principle perturb the planner: I swept sources per observation (5/20/80/200) and entity hub degree (3000/500/100 entities, plus 8 entities per fact) at the 4MB default `work_mem`. All eight cells came back with confidence intervals straddling zero, the entity arm's plan is structurally identical standalone vs. fused, and temp I/O is identical to the byte (103.42 MB/call at 200 sources/observation in both shapes). Nothing regresses, and no missing index is exposed.

I edited the PR description to record those numbers and changed `Closes #3857` to `Refs #3857` — hope that's alright. The reason is that this doesn't reduce the cost the issue is about: the entity arm's candidate generation is still bank-wide (at 20k observations the `candidates` CTE scans 19,966 rows, because `connected_sources` reaches ~87% of world facts under `per_entity_limit=200`). The driver is source-fact growth rather than entity fan-out — 308 ms → 1126 ms and 0 → 103 MB of temp I/O per call as sources per observation goes 5 → 200, which is the unbounded `source_memory_ids` append from #1725. So I'd rather keep #3857 open for that.

One thing worth flagging for a follow-up, not this PR: `_expand_combined` wraps its fused query in `asyncio.wait_for(link_expansion_timeout)` and degrades to semantic+causal on timeout, but `_expand_observations` has no such guard — and it's the arm that can take seconds. Now that the two paths have the same shape, that asymmetry is more visible.

Also: because this is a fork PR, `test-api` was skipped and only Strix ran, so none of your DB-backed tests had actually executed in CI. I've pushed the commits to `ci/3859-observation-expansion` on the origin repo and dispatched the full workflow so they run with secrets. Merging once that's green. Thanks for the clean, well-documented patch — the comments explaining what the fusion must preserve are genuinely useful.

nicoloboschi (approved):
Approved. Verified independently before merge: result-equivalence with the current query across 40 seed sets (all three arms, identical ids/scores/ordering), 6/6 of the new tests green in CI on main + these commits, and volume safety swept across sources-per-observation and entity hub degree with no regression and byte-identical temp I/O.

Merging with one known consequence: this trips a pre-existing test-isolation bug in `test_worker.py::TestWorkerPoller::test_claim_batch_claims_pending_tasks`. Not caused by this change — adding tests shifts the `pytest-split` boundary, pulling two `test_operation_status` tests into shard 3 where their claimable pending rows compete with the claim batch. Tracking that separately.

Issue #3564, opened 2026-08-18
coding-agents: the always-on `harness:<id>` tag splits observations per agent, undoing the harness-neutral bank — please make it optional


ibaou-dev:
**Package:** `@vectorize-io/hindsight-coding-agents` 0.3.4
**Code read at:** `b64943d195422ec62c1b7c7704d27c7e04c3f250` (main, 2026-08-17)

## Summary

Every retain carries a `harness:<id>` tag, and it cannot be turned off. Because consolidation groups memories by their **whole tag set** (the default `combined` scoping), the same project worked on with two agents produces **two separate sets of observations** — one per agent — that never merge.

This works against the design you describe in your own docs. [`coding-agents.md#L441`](https://github.com/vectorize-io/hindsight/blob/b64943d195422ec62c1b7c7704d27c7e04c3f250/hindsight-docs/docs-integrations/coding-agents.md#L441):

> The default `"coding-agent::{gitProject}"` is **harness-neutral**, so opencode, Claude Code, and Codex all share one memory per repo

The **bank** is shared, which is great. But the harness tag splits that shared memory again at the observation layer, so the sharing does not reach the consolidated knowledge.

**I am not asking for `per_tag`, and I am not asking you to remove the tag.** I would just like a way to switch it off, for people who do not need per-agent observations.

## Minimal reproduction

Two identical facts, differing **only** by the harness tag, then one consolidation pass:

```bash
B=http://localhost:8888/v1/default/banks/harness-scope-test

for H in claude-code opencode; do
  curl -s -X POST "$B/memories" -H 'Content-Type: application/json' -d "{\"items\":[{
    \"content\":\"The project uses Traefik as its reverse proxy, configured entirely through container labels rather than static files. Certificate renewal is automatic via ACME.\",
    \"document_id\":\"conv-$H\",
    \"tags\":[\"harness:$H\",\"project:demo\",\"source:chat\"]}],\"async\":false}"
done

curl -s -X POST "$B/consolidate" -H 'Content-Type: application/json' -d '{}'
curl -s "$B/memories/list?limit=50"
```

**Expected:** 2 observations (two beliefs about one project).
**Actual:** 4 observations — the same two beliefs, duplicated per harness:

```
tags=['harness:opencode',    'project:demo', 'source:chat']  "Certificate renewal is automatic via ACME."
tags=['harness:opencode',    'project:demo', 'source:chat']  "The project uses Traefik as its reverse proxy…"
[...]

PR #3872, opened 2026-08-29
feat(coding-agents): add per_source observation scoping

### LINKED ISSUE #3871: coding-agents: under `shared`, commit knowledge and conversation knowledge consolidate into one belief set with no way to ask them apart
**Code read at:** `56e8962fe9a90d7e8e740e8ca72fa9c9a4003b8a` (main, 2026-08-29)

## The problem

On a repo worked by a coding agent, this plugin writes documents from genuinely different origins into one bank: commit diffs (`source:git`), the commit-message seed (`source:git` + `source:git-log`), session transcripts (`source:chat`), uploads (`source:upload`).

"What the commits say" and "what was decided in conversation" are **different kinds of claim**. A commit diff is a record of what the code actually does. A transcript is a record of what someone intended, argued for, or speculated about — including the ideas that were considered and dropped. They disagree in normal, healthy ways: a conversation reasons toward a design that the commits then implement differently, or not yet, or only partly.

Since #3564/#3575 this plugin retains with `observation_scopes: "shared"`, which resolves to the single empty scope. That was the right call for the problem it solved — the `harness:<id>` tag was forking beliefs per agent, and `shared` collapsed that fork. But it collapses **every** axis, so all of those origins consolidate into one undifferentiated belief set, and there is currently no way to ask them apart:

- an idea that was floated in chat and never implemented consolidates alongside beliefs derived from the diffs, with nothing marking which is which;
- "what does the codebase actually do" cannot be answered from commit-derived knowledge alone;
- when the two conflict, consolidation reconciles them into one belief rather than keeping both claims available with their origin intact.

This is not the `harness:` problem returning. `harness:<id>` is **provenance** — which agent typed it — and splitting on it produces two copies of the same belief. `source:` is **semantic**: the origin changes what kind of claim the document is making. Those deserve different treatment.

## Why configuration cannot already do this

This is the crux, and it is why the issue needs a code change rather than a docs note.

`observationScopes` already accepts an explicit `string[][]`, so the obvious answer is "configure `[[], ["source:git"], ["source:chat"]]` and you are done". That does not work. The server treats an explicit list as **unconditional** — it is never filtered against the memory's own tags.

In `hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py`, both resolvers end the same way:

```python
def _resolve_obs_tags_list(memory): ...
    return parsed  # explicit list[list[str]]

def _resolve_write_scopes(memory): ...
    return [frozenset(s) for s in parsed]  # explicit list[list[str]]
```

`parsed` is returned verbatim. So a configured `[[], ["source:git"], ["source:chat"]]` writes **every** document into all three scopes: the `source:git` scope fills up with observations built from chat transcripts, and the `source:chat` scope fills with observations built from commit diffs. The scopes exist but mean nothing, which is worse than not having them.

Every scalar mode the server offers derives its scopes from the memory's own tags — but none of them derives from *part* of the tag set. Only a per-document decision, made where the document's tags are known, separates these.

## Why not `per_tag`

`per_tag` splits on the right axis, but also on every other one. `_resolve_obs_tags_list` maps it to `[[t] for t in tags]` — one scope per tag, unconditionally. That means:

- it reinstates the per-agent `harness:` fork that #3564/#3575 removed;
- any volatile tag becomes its own scope — a session id added through `retainTags` would create a scope per session, which is the fragmentation bug itself.

Reading **only** `source:` is what keeps the split safe. It is a small, closed, meaningful set of values that the plugin itself controls.

## Proposal

Accept `"per_source"` as an `observationScopes` value in the coding-agents plugin, resolved client-side, per document, and never sent to the server. It expands to the global scope **plus** one scope per distinct `source:` tag the document carries, sorted:

| document | tags | resolved scopes |
| --- | --- | --- |
| session transcript | `source:chat` | `[[], ["source:chat"]]` |
| commit diff | `source:git` | `[[], ["source:git"]]` |
| commit-message seed | `source:git`, `source:git-log` | `[[], ["source:git"], ["source:git-log"]]` |
| no source tag | `knowledge:convention` | `[[]]` |

The empty scope is **always emitted first and unchanged**, so the merged view is exactly what `shared` produces today. The extra scopes are additive: you can read one axis with `tags: ["source:git"], tags_match: "exact"`, or the merged set with `tags: [], tags_match: "exact"`.

A document may legitimately carry more than one source tag — the commit-message seed keeps `source:git` alongside `source:git-log` so the cold-repo check (`listDocumentIds("source:git")`) still sees it — so every distinct one gets a scope. Taking all of them, sorted, is the only rule that needs no arbitrary tie-break and does not silently depend on the order the caller assembled its tags in.

### What this deliberately does not change

- **The default is untouched.** `DEFAULT_OBSERVATION_SCOPES` stays `"shared"`. This is purely opt-in; a user who never sets it sees no behavior change at all.
- **Knowledge pages are unaffected.** Per #3664 the seeded pages read with `tags_match: "all"`, which includes untagged rows. Because the empty scope is still always written, the untagged observations those pages depend on keep being produced exactly as under `shared`.
- **Nothing is repurposed.** #2853 asked that this area change "ideally additively (a new, clearly-named field)". A new accepted value on an existing field is on that side of the line — no existing value changes meaning.

### The honest cost

`per_source` costs one extra consolidation pass per document (two for the commit-message seed). That is the direct trade #3575 was right to refuse for `harness:` — there, the extra pass bought two copies of the same belief. Here the extra pass buys a distinction that cannot be recovered afterward: once a commit-derived belief and a chat-derived belief are merged into one observation, no read-side filter can separate them again, because the merged observation has a single origin-free identity.

So: yes, it doubles the consolidation cost for anyone who turns it on. That is the price of the axis, it is opt-in, and the default is unchanged.

## Lineage

- #3564 — the `harness:<id>` tag splitting observations per agent
- #3575 — introduced `observationScopes` and the `"shared"` default
- #3664 — knowledge pages read untagged observations with `tags_match: "all"`
- #2853 — asked for changes here to be additive

I have a patch ready (client-side resolver, tests, docs) and will open a PR against this issue if the approach is acceptable.

shauneccles:
Closes #3871.

## The problem

On a repo worked by a coding agent, this plugin writes documents from genuinely different origins into one bank: commit diffs (`source:git`), the commit-message seed (`source:git` + `source:git-log`), session transcripts (`source:chat`), uploads (`source:upload`).

A commit diff records what the code actually does. A transcript records what someone intended, argued for, or discarded. Under the `shared` default these consolidate into one undifferentiated belief set, so:

- an idea floated in chat and never implemented is indistinguishable from a belief derived from the diffs;
- "what does the codebase actually do" cannot be answered from commit-derived knowledge alone;
- when the two conflict, consolidation reconciles them into a single belief rather than keeping both claims with their origin intact.

## Why this cannot be done with configuration

This is the crux, and it is why the change is code rather than a docs note.

`observationScopes` already accepts an explicit `string[][]`, so the obvious answer is "just configure `[[], ["source:git"], ["source:chat"]]`". That does not work. The server treats an explicit list as **unconditional** — it is never filtered against the memory's own tags. In `hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py`, both resolvers end the same way:

```python
def _resolve_obs_tags_list(memory): ...
    return parsed  # explicit list[list[str]]

def _resolve_write_scopes(memory): ...
    return [frozenset(s) for s in parsed]  # explicit list[list[str]]
```

`parsed` comes back verbatim. So that configuration writes **every** document into all three scopes: the `source:git` scope fills with observations built from chat transcripts and vice versa. The scopes exist but mean nothing, which is worse than not having them.

Every scalar mode the server offers derives its scopes from the memory's own tags, but none derives from *part* of the tag set. Only a per-document decision, made where the tags are known, separates these.

## The change

`"per_source"` becomes an accepted `observationScopes` value. It is resolved **client-side, per document**, in a new exported `resolveRetainScopes(tags, configured)` in `src/core/hindsight.ts`, and **never reaches the server**. It expands to the global scope plus one per distinct `source:` tag the document carries, sorted:

| document | tags | resolved scopes |
| --- | --- | --- |
| session transcript | `source:chat` | `[[], ["source:chat"]]` |
| commit diff | `source:git` | `[[], ["source:git"]]` |
| commit-message seed | `source:git`, `source:git-log` | `[[], ["source:git"], ["source:git-log"]]` |
| no source tag | `knowledge:convention` | `[[]]` |

A document may legitimately carry more than one source tag — the commit-message seed keeps `source:git` alongside `source:git-log` so the cold-repo check (`listDocumentIds("source:git")`) still sees it — so every distinct one gets a scope. Taking all of them, sorted, is the only rule that needs no arbitrary tie-break and does not silently depend on the order `git.ts` happens to emit its tags in.

`this.observationScopes` still flows to the single call site that sets the field; only the value it resolves to is now per document.

## On "double the consolidation cost"

That is the explicit rationale of #3575, so it deserves a direct answer rather than a footnote.

Yes: `per_source` costs one extra consolidation pass per document (two for the commit-message seed). The difference from the case #3575 was right to refuse is what the extra pass buys. Splitting on `harness:<id>` is **provenance** — which agent typed it — and the extra pass buys two copies of the same belief, each blind to the other. Splitting on `source:` is **semantic**: the origin changes what kind of claim the document is making, and once a commit-derived belief and a chat-derived belief merge into one observation, no read-side filter can separate them again, because the merged observation has a single origin-free identity.

So the cost is real, it is the price of the axis, and it is why this is opt-in.

## Why not `per_tag`

`per_tag` splits on the right axis but also on every other one — `_resolve_obs_tags_list` maps it to `[[t] for t in tags]`, unconditionally. It would reinstate the per-agent `harness:` fork that #3564 and #3575 removed, and any volatile tag (a session id added through `retainTags`) would become its own scope, which is the fragmentation bug itself. Reading **only** `source:` is what keeps the split safe: a small, closed, meaningful set of values the plugin itself controls.

## What is deliberately unchanged

- **The default.** `DEFAULT_OBSERVATION_SCOPES` is still `"shared"`. A user who never sets this sees no behavior change.
- **Knowledge pages.** The empty scope is always emitted first and unchanged, so the untagged observations the seeded pages read (`tags_match: "all"`, per #3664) keep being written exactly as under `shared`.
- **Nothing repurposed.** #2853 asked that changes here be "ideally additively (a new, clearly-named field)". This is a new accepted value on an existing field — no existing value changes meaning.

## One test was widened — deliberately

The source-text guard in `hindsight.test.ts` ("keeps that call site inside `retain()`, with the scoping on the item it posts") went from

```js
expect(body).toContain("observation_scopes: this.observationScopes");
```

to

```js
// The scoping may be derived per document (see `per_source`), but it must still be set on the
// item here and still come from the configured value — not from a server default.
expect(body).toMatch(/observation_scopes: .*this\.observationScopes/);
```

Flagging it explicitly so it does not read as weakening a test. The guard's intent is intact: the field must still be set inside `retain()`, on the item it posts, and still derive from `this.observationScopes` rather than being left to the server default. The regex only allows the value to be wrapped in a resolver. A call site that dropped the scoping, moved it out of `retain()`, or hardcoded a value still fails.

## Lineage

- #3564 — the `harness:<id>` tag splitting observations per agent
- #3575 — introduced `observationScopes` and the `"shared"` default
- #3664 — knowledge pages read untagged observations with `tags_match: "all"`
- #2853 — asked for changes here to be additive

## Test plan

- `npx tsc --noEmit` — clean
- `npx vitest run` — 55 files passed, 665 passed / 23 skipped
- `npm run build` — succeeds
- `npm run skill:build` and `node hindsight-docs/scripts/sync-coding-agents-doc.mjs` re-run, generated output committed
- prettier 3 with the repo's root `.prettierrc.json` over every changed file (matching the `prettier-int-coding-agents` task in `scripts/hooks/lint.sh`) — clean. Three files under this package (`src/core/retain-cursor.ts`, `src/dsh.ts`, `src/install-ui.ts`) are flagged by that task on `main` as well; they are untouched here and left alone.
- `src/docs-freshness.test.ts` — 12 tests pass

Six new tests cover `per_source` end to end at the wire level (the global-plus-source expansion, git vs chat separation, the untagged fallback, the two-source-tag seed, order independence, and a volatile provenance tag never becoming a scope), plus one covering config acceptance.

I could not run `./scripts/hooks/lint.sh` in full — it starts with `uv sync` for the whole monorepo and my checkout is sparse — so the Python and control-plane tasks are unverified, though nothing here touches them.


koriyoshi2041 (approved):
The per-document resolution keeps the existing shared scope while separating only the controlled source axis, and it avoids sending the plugin-only mode to the server. I also checked the multi-source and no-source cases at a9a729f: 85 focused tests pass, along with typecheck and the package build.

shauneccles:
## Production feedback after running this on a self-hosted 0.9.2 deployment

Three per-repo banks, reprocessed under the change. Scope distributions:

```
coding-agent::hermes-stack     5 scopes, 951 obs
  global 321 · source:git-log 307 · source:git 302 · source:chat 20 · source:survey-baseline 1
coding-agent::dbml-sharepoint  5 scopes, 272 obs
  global 100 · source:upload 61 · source:git 58 · source:git-log 52 · source:survey-baseline 1
coding-agent::firecrawl        2 scopes, 22 obs
  global 13 · source:chat 9
```

Working as intended: the global scope stays the largest in every bank, so cross-cutting beliefs still form, and the per-source axes are populated and queryable with `tags_match: "exact"`. No `harness:` or session-id scope appeared anywhere, which is the safety property — only `source:` is read.

Two things I did not anticipate, both worth deciding on before this merges:

**1. `source:git-log` is very nearly a duplicate of `source:git`.** The commit-message seed carries both tags (`git.ts` keeps `source:git` so the cold-repo `listDocumentIds("source:git")` check works), so its facts write to both scopes. On the first bank that is 307 against 302 — two scopes, almost the same content, double the consolidation for the pair. `git-log` is a bookkeeping alias for "what the commits say" rather than a distinct kind of claim.

**2. `source:survey-baseline` produces a scope with exactly one observation.** That document is the `🛰️ researching…` status marker, retained under the `survey` strategy whose marker rule is meant to extract nothing. A singleton scope from a status marker is pure noise.

Both point the same way: the axis that earns a scope is the *kind of claim*, and not every `source:` tag is one. Options, in the order I would consider them:

- Map source tags onto a small set of axes rather than one scope per tag, so `git-log` folds into `git` and markers map to nothing.
- Or keep it tag-driven and exclude a named set (`source:survey-baseline`, `source:git-log`).
- Or leave as-is and document that a dual-tagged document produces two scopes.

I deliberately emit a scope per *distinct* source tag, sorted, rather than picking the first — taking the first made the result depend on the order `git.ts` happened to assemble its tags, which a test now pins. That choice is still right; it is the tag vocabulary that turns out to carry two non-semantic entries.

Happy to push whichever direction you prefer, or to leave the current behaviour and just document it.

shauneccles:
Sorry for the noisy last two commits - I missed the root cause of another issue (https://github.com/vectorize-io/hindsight/pull/3874) and got a bit confused. It's been a big day 😅

koriyoshi2041 (approved):
The two follow-up commits leave the implementation unchanged and make the multi-tag intent clearer: the git-log seed and per-commit diffs answer different questions, so retaining both controlled source scopes is correct. Re-reviewed at 1a98249.

PR #3874, opened 2026-08-29
fix(api): round-trip the whole retain item so reprocess replays it faithfully

### LINKED ISSUE #3873: reprocess silently re-extracts under the wrong strategy: strategy never reaches retain_params
## Summary

`reprocess_document` rebuilds its retain call entirely from `documents.retain_params`. It reads `strategy` back out — but `strategy` is never written there, so **every reprocess re-extracts the document under the bank's `retain_default_strategy` instead of its own**. No error, no warning.

The break is in two places, and fixing either alone changes nothing.

## Where

**1. `hindsight_api/api/http.py`, `api_retain`.** `strategy` is used as the key items are grouped by, and never copied onto the content dict:

```python
for item in request.items:
    effective = item.strategy          # used only to group
    content_dict: dict = {"content": item.content}
    ...
    if item.observation_scopes is not None: content_dict["observation_scopes"] = ...
    if item.update_mode is not None:        content_dict["update_mode"] = ...
    strategy_groups[effective].append(content_dict)   # strategy not on the dict
```

**2. `hindsight_api/engine/retain/orchestrator.py:363`, `_build_retain_params`** captures `context`, `event_date`, `metadata`, `observation_scopes` — and so could not record a `strategy` that was never in the dict anyway.

`reprocess_document` in `memory_engine.py` then reads six fields back:

```python
if retain_params.get("context"):    ...
if retain_params.get("event_date"): ...
if retain_params.get("metadata"):   ...
if retain_params.get("entities"):   ...
if retain_params.get("observation_scopes") is not None: ...
strategy = retain_params.get("strategy")     # always None
```

## Reproduction, on 0.9.2

Bank with `retain_default_strategy: "git"` plus a `survey` strategy whose `retain_custom_instructions` say to extract nothing from an internal status marker. Same marker content throughout:

| document | strategy sent | units after retain | units after reprocess | `retain_params.strategy` |
|---|---|---|---|---|
| `m-with` | `survey` | 0 | — | `None` |
| `m-without` | none | 1 | — | `None` |
| `m-fix2` (both fixes applied) | `survey` | 0 | **0** | `'survey'` |
| `m-fix` (unfixed) | `survey` | 0 | **1** | `None` |

The strategy works when sent — `m-with` correctly extracts nothing. It simply cannot survive a round trip.

Observed on a real bank first: three survey status-marker documents sat correctly at 0 memory units for hours, then produced facts the moment they were reprocessed. Document `created_at` 02:42, extracted-fact `created_at` 04:54 — the reprocess. Pointing a git-commit mission at a status line also invents state: it produced *"The codebase survey was completed and recorded as the baseline commit"* from a marker that only says a survey **started**.

## Why it matters beyond markers

The endpoint is documented as *"re-extracts facts using the current engine configuration. Useful when the LLM model, chunking strategy, or extraction settings have changed"* — precisely when an operator reaches for it. Any bank using `retain_strategies` gets every reprocessed document re-extracted under its default mission: conversations extracted as commits, documents extracted as conversations. Silent, and visible only by reading the resulting facts.

`entities` has the same round-trip gap in `_build_retain_params`, though `api_retain` does put it on the content dict, so that half is a one-line fix.

## Fix

PR incoming: carry `strategy` onto the content dict in `api_retain`, and capture `strategy` and `entities` in `_build_retain_params`. Verified end to end against a live 0.9.2 server — the `m-fix2` row above.

Version: 0.9.2, self-hosted, embedded Postgres.

shauneccles:
Closes #3873.

`reprocess_document` rebuilds its retain call entirely from `documents.retain_params`. Both sides enumerated fields by hand, and the two lists had drifted **three times over**: `strategy`, `entities` and `resolve_entities` are each written by `api_retain` and never captured. A reprocess therefore re-extracted the document under the bank's `retain_default_strategy`, with entity resolution it had been told not to do — invisibly, since the reprocess succeeds and only the resulting facts are wrong.

### The fix inverts the rule rather than extending the list

Adding the missing three would have left the same shape that already failed three times. Instead:

- `retain_params` captures **everything the caller supplied** except what a reprocess legitimately supplies itself — `content` (the stored `original_text`), `document_id` and `update_mode` (the reprocess sets its own), and `tags` (kept on the document row).
- `reprocess` replays `retain_params` **whole** instead of cherry-picking.

The failure mode becomes a deliberate exclusion rather than a silent omission, and `api_retain`'s content dict becomes the single source of truth for what a retain can carry.

Two details that are easy to get wrong, both found by testing rather than reading:

- **`api_retain` must put `strategy` ON the content dict**, not just use it as the key it groups items by. Without that it never reaches `_build_retain_params`, and fixing the orchestrator alone changes nothing.
- **`strategy` must ride on the replayed dict as well as being pulled out as the call argument.** Excluding it there survived exactly *one* reprocess: the call argument applied the right strategy, but `_build_retain_params` never saw it on the item, so `retain_params` came back without it and the next reprocess fell back to the bank default again.

### How it was found

A real bank. Three survey status markers — whose strategy instructs the extractor to return nothing — sat correctly at 0 memory units for hours, then produced facts the moment they were reprocessed. Document `created_at` 02:42, extracted-fact `created_at` 04:54: the reprocess. A git-commit mission pointed at a status line also invents state, yielding *"The codebase survey was completed and recorded as the baseline commit"* from a marker that only says a survey **started**.

It bites exactly when the endpoint is meant to help — it is documented as *"re-extracts facts using the current engine configuration. Useful when the LLM model, chunking strategy, or extraction settings have changed"*, which is when an operator reaches for it. Any bank using `retain_strategies` then gets its documents re-extracted under the wrong mission.

### Verification

Against a live 0.9.2 server, same marker content throughout:

| | units after retain | after reprocess 1 | after reprocess 2 | `retain_params.strategy` |
|---|---|---|---|---|
| patched | 0 | **0** | **0** | `'survey'` |
| unpatched | 0 | **1** | — | `None` |

Tests pin each half and pair `api_retain`'s content dict against the exclusion rule, so a newly added retain field cannot silently stop round-tripping.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -8208,6 +8208,12 @@ def _register_routes(app: FastAPI):
                     content_dict["observation_scopes"] = item.observation_scopes
                 if item.update_mode is not None:
                     content_dict["update_mode"] = item.update_mode
+                # Carried on the item, not just used as the grouping key: reprocess
+                # rebuilds its retain call from retain_params, so a strategy that
+                # never reaches the content dict never reaches retain_params either
+                # — and the reprocess silently re-extracts under the bank default.
+                if item.strategy:
+                    content_dict["strategy"] = item.strategy
                 strategy_groups[effective].append(content_dict)
 
             if request.async_:
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -11113,21 +11113,28 @@ class MemoryEngine(MemoryEngineInterface):
             "document_id": document_id,
             "update_mode": "replace",
         }
-        if retain_params.get("context"):
-            content_dict["context"] = retain_params["context"]
-        if retain_params.get("event_date"):
-            content_dict["event_date"] = retain_params["event_date"]
-        if retain_params.get("metadata"):
-            content_dict["metadata"] = retain_params["metadata"]
-        if retain_params.get("entities"):
-            content_dict["entities"] = retain_params["entities"]
+        # Replay everything the original retain supplied. Enumerating fields here
+        # is what let `strategy`, `entities` and `resolve_entities` go missing one
+        # by one: each was written by api_retain, never captured, and a reprocess
+        # quietly re-extracted under the bank's default strategy instead of the
+        # document's own. retain_params now holds exactly the replayable set (see
+        # _RETAIN_PARAMS_NOT_REPLAYED), so take it whole.
+        # `strategy` rides on the dict as well as being pulled out below. Excluding
+        # it here made the round trip survive exactly one reprocess: the call
+        # argument applied the right strategy, but _build_retain_params never saw it
+        # on the item, so retain_params came back without it and the NEXT reprocess
+        # fell back to the bank default again.
+        content_dict.update(retain_params)
+        # These three are the reprocess's own and must win over anything stored.
+        content_dict["content"] = original_text
+        content_dict["document_id"] = document_id
+        content_dict["update_mode"] = "replace"
 
         tags = doc.get("tags") or []
         if tags:
             content_dict["tags"] = tags
-        if retain_params.get("observation_scopes") is not None:
-            content_dict["observation_scopes"] = retain_params["observation_scopes"]
 
+        # A call-level argument rather than an item field, so it is pulled out.
         strategy = retain_params.get("strategy")
 
         result = await self.submit_async_retain(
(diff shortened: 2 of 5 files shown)


shauneccles:
@strix-security

koriyoshi2041 (commented):
Checked exact head `67853d5` locally. The new round-trip suite passes 9/9, and the exclusion/replay contract looks sound across `api_retain`, `_build_retain_params`, and consecutive reprocesses.

One merge-gate fix is needed: `uv run ruff format --check hindsight_api/api/http.py hindsight_api/engine/memory_engine.py hindsight_api/engine/retain/orchestrator.py tests/test_retain_params_roundtrip.py` reports that `tests/test_retain_params_roundtrip.py` would be reformatted. `ruff check` and `git diff --check` both pass.

I also tried the document-tracking suite; after adding the embedded-DB extra it reaches migrated Postgres, then setup stops because the slim environment does not include `sentence-transformers`, so I am not treating those setup errors as diff failures.

PR #3879, opened 2026-08-29
fix(coding-agents): stop unsafe gitlog cleanup

### LINKED ISSUE #3877: Document tag filter returns untagged documents; coding-agents gitlog sweep then deletes them and ON DELETE CASCADE wipes their facts
## Summary

`GET /v1/default/banks/{bank_id}/documents?tags=<tag>` returns documents whose `tags` array is **empty**, for *any* tag value. The companion `hindsight-coding-agents` plugin uses that exact endpoint to build a list of documents to **delete**. Result: every coding-agent session start destroys unrelated documents in the bank, and `ON DELETE CASCADE` from `documents` to `memory_units` wipes every fact extracted from them.

Hit here three times: `2026-08-28 21:38`, `2026-08-29 07:11`, `2026-08-29 09:30` UTC — 53 documents and ~1445 memory units each time. Only a btrfs snapshot of the Postgres data directory saved the bank.

## Environment

- self-hosted API `0.9.2` (`ghcr.io/vectorize-io/hindsight:latest`), Postgres 18.1.0 (`/version` → `api_version 0.9.2`)
- plugin `@vectorize-io/hindsight-coding-agents@0.4.3`, harness `prime-agent`
- one bank deliberately shared by several repos (top-level `bankId`) — that is what turned a wrong selector into cross-repo damage

## Repro: the tag predicate ignores the requested tag

```bash
API=http://127.0.0.1:8888/v1/default/banks/prime-agent/documents
curl -s "$API?limit=500"                                                  | jq '.total'   # 54
curl -s "$API?tags=source:git-log&tags_match=all&limit=500"               | jq '.total'   # 53
curl -s "$API?tags=zzz-does-not-exist&tags_match=all&limit=500"           | jq '.total'   # 53  <- control
curl -s "$API?tags=zzz-does-not-exist&tags_match=all&limit=500" | jq '.items[0].tags'      # []
```

The same 53 documents come back for a real tag and for a fabricated one, and they carry **no tags at all**. They are ordinary documents created via `POST /banks/{id}/memories` without an explicit `tags` field — i.e. anything a script, an agent tool call, or an external integration ingests.

## The consumer that makes it destructive

`dist/deepen.js` (0.4.3), lines 1662-1672:

```js
const canonical = `gitlog:${repoNameOf(REPO)}`;
const logDocs = await client.listDocumentIds("source:git-log");
for (const id of logDocs) {
  if (id !== canonical) {
    await client.deleteDocument(id);
    log2(`[gitlog] removed stale duplicate ${id} (canonical: ${canonical})`);
  }
}
```

`listDocumentIds()` builds `?tags=${tag}&tags_match=all`, pages it, and trusts the server's answer; the loop then deletes every id that is not the canonical name. With the predicate broken, the canonical name is the only thing that survives — and the check is a *name mismatch*, not an ownership check, so in a shared bank it also eats other repositories' documents.

Plugin log (`$TMPDIR/hindsight-coding-agent/plugin.log`) — 53 deletions in 1.2 s:

```
2026-08-29T09:30:55.742Z INFO [deepen] [gitlog] removed stale duplicate fc8ac2ca-da37-455d-b7e4-ab964b063a10 (canonical: gitlog:stardust)
2026-08-29T09:30:55.771Z INFO [deepen] [gitlog] removed stale duplicate 8b1fde87-4311-48b6-8e33-defe67fbde65 (canonical: gitlog:stardust)
```

Postgres with `log_statement=mod` — 55 `DELETE FROM public.documents` in the 09:30 minute, and only **3** explicit `DELETE FROM memory_units`:

```
2026-08-29 09:30:55.877 UTC [442] LOG:  execute <unnamed>: DELETE FROM public.documents WHERE id = $1 AND bank_id = $2 RETURNING id
2026-08-29 09:30:55.877 UTC [442] DETAIL:  Parameters: $1 = 'fc8ac2ca-da37-455d-b7e4-ab964b063a10', $2 = 'prime-agent'
```

The rest left silently through the FK:

```
memory_units_document_fkey | FOREIGN KEY (document_id, bank_id) REFERENCES documents(id, bank_id) ON DELETE CASCADE
```

`1496 → 51` memory units, with no statement in the entire audit log that deletes most of them. Note also that `listDocumentIds("source:git")` at line 1633 (used to decide which commits are already ingested) is polluted the same way — false "already ingested".

## Suggested fixes

1. **Server:** an empty/absent `tags` array must never satisfy `tags=<anything>`. Regression test: a document with no tags is excluded from every tag filter.
2. **Plugin:** re-verify each candidate's tags client-side before `deleteDocument`, and never delete a document whose `tags` is empty.
3. **Plugin:** scope the sweep to documents it created (own id namespace / provenance tag) instead of bank-wide name mismatch.
4. **Schema:** reconsider `ON DELETE CASCADE` from `documents` to `memory_units`, or gate it (`?include_memories=true`), so a wrong selector cannot vaporize extracted knowledge in one statement.
5. **Both:** log/refuse when a destructive sweep's candidate count is unexpectedly large relative to what the caller should own.

## Workaround applied on this machine

- `banks.<shared bank>.gitIngest = "none"` — the sweep lives inside that branch, so it stops running there
- one bank per repo via `mapPathToBank`, so a sweep in repo A cannot see repo B at all
- `PATCH /banks/{id}/documents/{document_id}` with `{"tags":["imported:pa-sessions"]}` on all 53 untagged documents

After tagging, the broken selector returns exactly 1 document — the intended canonical one — verified:

```bash
curl -s "$API?tags=source:git-log&tags_match=all&limit=500" | jq '.total, .items[0].id'
# 1
# "gitlog:startdust"
```

Related: #3876 is a different wipe (consolidation schema violation). This one is independent, deterministic, and reproducible with the three curls above.

r266-tech:
## Summary

- stop coding-agents from enumerating and deleting bank-wide `source:git-log` documents during sync
- make every internal multi-tag strategy probe use `all_strict`, while preserving `listDocumentIds()`'s existing public default
- treat a git-log snapshot as current only when the current-HEAD query contains this repository's canonical document; otherwise perform an idempotent upsert

Fixes #3877.

## Why

The document listing endpoint can include untagged documents for the legacy inclusive `all` mode. The old deepen path then treated every returned non-canonical document as stale and deleted it. Because document deletion cascades to facts, a routine coding-agents sync could remove unrelated memories from a shared bank.

This change removes that destructive cleanup entirely. Git-log sync now owns only its canonical document and never deletes other document IDs.

## Compatibility and remaining limitation

The optional `tags_match` parameter defaults to `all`, matching the existing client behavior for external callers. Coding-agents' internal strategy checks opt into `all_strict` explicitly.

Canonical git-log IDs remain `gitlog:<repoName>` for backward compatibility. Consequently, unrelated same-named repositories or forks can still share that legacy ID; changing the namespace safely requires a separate ownership/migration design. This patch prevents cross-document deletion but does not attempt that migration.

## Validation

- `npx tsc --noEmit`
- `npx vitest run src/core/hindsight.test.ts src/core/git.gitlog.test.ts src/core/status.test.ts src/core/session-start.test.ts` (69 tests)
- `npm test` (663 tests)
- `npm run build`
- `git diff --check`

The repository-wide lint hook could not load the control-plane `@eslint/js` dependency from this sparse worktree; the package typecheck, complete test suite, build, and diff checks pass.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/git.gitlog.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/git.gitlog.test.ts
@@ -4,7 +4,14 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { HindsightClient } from "./hindsight";
-import { gitLogNewestAuthorDate, gitLogText, ingestGitLog, repoNameOf, retainCommit } from "./git";
+import {
+  gitLogNewestAuthorDate,
+  gitLogText,
+  ingestGitLog,
+  repoNameOf,
+  retainCommit,
+  syncGitLog,
+} from "./git";
 
 let dir: string;
 
@@ -130,5 +137,54 @@ describe("ingestGitLog", () => {
       source: "git",
       commit: sha,
     });
+  });
+});
+
+describe("syncGitLog", () => {
+  beforeEach(() => {
+    dir = mkdtempSync(join(tmpdir(), "hs-gitlog-sync-"));
+    initRepo(dir);
+  });
+
+  it("never enumerates or deletes foreign git-log documents from a shared bank", async () => {
+    execFileSync("git", ["-C", dir, "commit", "--allow-empty", "-m", "feat: current repo"]);
+    const listDocumentIds = vi.fn(
+      async (_tag: string, _tagsMatch?: "all" | "all_strict") => new Set(["gitlog:foreign-repo"])
+    );
+    const retain = vi.fn().mockResolvedValue(undefined);
+    const deleteDocument = vi.fn().mockResolvedValue(undefined);
+    const client = {
+      listDocumentIds,
+      retain,
+      deleteDocument,
+      opIds: [],
+    } as unknown as HindsightClient;
+
+    const failures = await syncGitLog(client, dir, { limit: 10 });
+
+    expect(failures).toBe(0);
+    expect(retain).toHaveBeenCalledTimes(1);
+    expect(listDocumentIds).toHaveBeenCalledTimes(1);
+    expect(listDocumentIds.mock.calls[0][0]).toMatch(/^gitlog-head:/);
+    expect(listDocumentIds.mock.calls[0][1]).toBe("all_strict");
+    expect(listDocumentIds).not.toHaveBeenCalledWith("source:git-log");
+    expect(deleteDocument).not.toHaveBeenCalled();
+  });
+
+  it("skips the upsert only when this repository's canonical document has the current HEAD tag", async () => {
+    execFileSync("git", ["-C", dir, "commit", "--allow-empty", "-m", "feat: current repo"]);
+    const listDocumentIds = vi.fn(
+      async (_tag: string, _tagsMatch?: "all" | "all_strict") =>
+        new Set([`gitlog:${repoNameOf(dir)}`])
+    );
+    const retain = vi.fn().mockResolvedValue(undefined);
+    const client = { listDocumentIds, retain, opIds: [] } as unknown as HindsightClient;
+
+    const failures = await syncGitLog(client, dir, { limit: 10 });
+
+    expect(failures).toBe(0);
+    expect(listDocumentIds.mock.calls[0][0]).toMatch(/^gitlog-head:/);
+    expect(listDocumentIds.mock.calls[0][1]).toBe("all_strict");
+    expect(retain).not.toHaveBeenCalled();
   });
 });
--- a/hindsight-integrations/coding-agents/src/core/git.ts
+++ b/hindsight-integrations/coding-agents/src/core/git.ts
@@ -254,3 +254,33 @@ export async function ingestGitLog(
     return 1;
   }
 }
+
+/**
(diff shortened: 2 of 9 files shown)


koriyoshi2041 (approved):
Verified at 762a6f8e. The destructive bank-wide cleanup is gone, internal tag probes use strict matching, and the current-HEAD check now requires this repository's canonical document ID. I also ran the 69 focused tests, TypeScript typecheck, package build, and diff hygiene successfully.

nicoloboschi (approved):
Fixes active data loss: the deepen path deleted non-canonical documents returned by the inclusive `all` tag filter, and document deletion cascades to facts, so a routine coding-agents sync could remove unrelated memories from a shared bank. Removing the sweep entirely — rather than narrowing it — is the right call; git-log sync only ever needed to own its own canonical document.

Scoping the internal strategy probes to `all_strict` while leaving `listDocumentIds()`'s public default at `all` keeps external callers unchanged. The remaining `gitlog:<repoName>` collision between same-named repos and forks is correctly called out as a separate ownership/migration design rather than smuggled in here.

CI green, including `test-coding-agents`.

PR #3887, opened 2026-08-29
feat(coding-agents): add native DeepAgents Dcode integration

Paritoshdagar:
## Summary

- add a first-class DeepAgents Dcode Agent Plugin with Hooks V2, the shared skill, and a namespaced Hindsight MCP server
- parse Dcode transcripts and recover the final assistant message when the transcript lags the Stop event
- install through Dcode's native marketplace/plugin manager without patching `~/.deepagents` or taking over foreign marketplace state
- keep the Dcode plugin version synchronized by the integration release script and document the harness

## Verification

- 670 unit/integration tests passed
- package build and declarations passed
- real isolated `dcode plugin marketplace add/install/uninstall` smoke passed
- repository pre-commit hooks passed, including docs generation and lint
- targeted Prettier and `git diff --check` passed

## Native contract

The root `plugin.json` is the Dcode boundary. It contributes `hooks/hooks.json`, `skill/`, and `hindsight_*` over stdio with `HINDSIGHT_MCP_HARNESS=dcode`; no config bridge, sync script, wrapper, or patched Dcode runtime is involved.


nicoloboschi:
Thanks for this @Paritoshdagar — the integration was well built. I verified every contract in it
against deepagents-code 0.1.65's own source (the manifest paths, `${PLUGIN_ROOT}`, the Hooks V2
event names, the `schema_version: 1` transcript shape, the marketplace format) and they were all
correct, including the lag semantics your `last_assistant_message` recovery is built on.

Merged as #3967 via 4388e0f95, rebased onto current `main` with your authorship preserved. I opened
it from an origin branch because fork PRs skip `test-api` entirely, so the checks here were a
cheaper subset than they looked.

Two things that only showed up once it was running against a real install, both fixed in the merge:

- Dcode computes `last_assistant_message` as `str(content)`, so for any provider returning content
  *blocks* it is a Python repr rather than prose. The Stop hook was retaining ~1.8KB of the
  provider's encrypted reasoning payload as the assistant turn, and because the repr can never
  equal the transcript reader's clean text, the dedupe guard also appended a duplicate on every
  already-flushed turn. Now decoded via a `readLastMessage` hook on the retain spec.
- Dcode rejects MCP tools lacking a `readOnlyHint` annotation in headless mode, which made recall
  and the knowledge-page tools unusable under `dcode -n`. The read-only tools are now annotated.

Plus the parity sites a new harness tends to miss: local history import, a Docker E2E adapter, the
harness icon on the docs site as well as the control plane, and a clean uninstall.

Closing in favour of #3967. Much appreciated.

Issue #3876, opened 2026-08-29
Consolidation silently wipes bank: schema-invalid LLM output deletes source observations, merged units never created (v0.9.2)


IliaKondratiev:
## Summary

In self-hosted hindsight **v0.9.2** (embedded pg0, OpenAI-compatible self-hosted LLM), consolidation **silently destroyed the entire memory bank**: source `memory_units` were deleted while the merged episodes failed to be created because the LLM responses violated the expected pydantic schema. Two separate bank wipes happened ~24h apart on different data. After each wipe: `memory_units` ≈ 0, `audit_log` = 0, every `async_operations` row shows **status = completed**, stats `failed_operations: 0` — no surface signal that data was lost.

## Environment

- Image: `ghcr.io/vectorize-io/hindsight:latest`, API reports **v0.9.2**, docker
- LLM: OpenAI-compatible self-hosted vLLM serving `Qwen/Qwen3.8-Flash-Next-FP8` (reasoning model), `--reasoning-parser qwen3` on the vLLM side
- API env: `HINDSIGHT_API_LLM_REASONING_EFFORT=none`, `HINDSIGHT_API_LLM_MAX_CONCURRENT=4`
- Embeddings: `bge-m3` (1024d) via Ollama OpenAI-compatible endpoint
- Bank at wipe #1: 1646 memory units / 54 documents (from ~105 retained documents)

## Evidence

**1. Chain of 15 back-to-back consolidation operations** (from `async_operations`), each ~7 min, each immediately re-queued after the previous completed — 20:26:14 → 21:26:16 UTC:

```
consolidation | completed | 2026-08-28 20:26:14 | completed 20:33:37
consolidation | completed | 2026-08-28 20:33:37 | completed 20:40:20
consolidation | completed | 2026-08-28 20:40:20 | completed 20:47:38
... (15 total, chaining until the bank was empty)
```

**2. Pydantic validation errors during consolidation LLM batches** (verbatim from `docker logs`):

```
updates.3.observation_id
  Field required [type=missing, input_value={'text': "User added a cr...fication health check.'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
updates.4.observation_id
  Field required [type=missing, input_value={'text': "User configured...system health metrics."}, input_type=dict]
updates.5.observation_id
  Field required [type=missing, input_value={'text': 'User enabled sy...yment in obs 9a3b7d1e.'}, input_type=dict]
```

[...]

PR #3894, opened 2026-08-30
fix(consolidation): defer destructive deletes until replacements succeed

r266-tech:
## Summary

- validate LLM-requested deletes against the recall snapshot, but defer them until all fallible updates and creates finish
- execute the final delete set in one short database transaction
- preserve the existing external-store write-group visibility contract

## Why

`_create_observation_directly` computes embeddings before opening its write transaction. Previously, consolidation deleted old observations first, so an embedding or later replacement-write failure could leave the old observation durably removed and no replacement written.

This change makes destructive work the final phase. A failed update or create now leaves the prior observations intact for retry; a failure inside a multi-delete phase rolls the SQL delete set back together. This addresses the deterministic delete-first failure mode reported in #3876 without claiming to resolve every cause discussed there.

## Tests

- added a regression proving a replacement-create failure never reaches the requested delete
- `pytest tests/test_consolidation_failure_isolation.py::test_create_failure_does_not_execute_requested_delete tests/test_consolidation_embedding_validation.py tests/test_consolidation_dedup.py -q -n 0` (47 passed)
- Ruff check and format check for both changed files
- Codex adversarial review approved after explicitly checking capacity, rollback, and cross-store visibility semantics

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -2189,8 +2189,7 @@ async def _process_memory_batch(
         perf.record_timing("llm", time.time() - t0)
         perf.record_llm_call(llm_result.obs_count, llm_result.prompt_chars)
 
-    # 4. Sequential execution of deletes / updates / creates
-    # Deletes run first to free observation slots before creates consume them.
+    # 4. Sequential execution of updates / creates / deletes
     # Track which memory indices participated so we can build per-memory results for stats
     per_memory_created: set[str] = set()
     per_memory_updated: set[str] = set()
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (commented):
The sequencing change looks sound on exact head `4181cbe`, and I ran the stated focused gate locally: 47/47 passed.

One data-integrity guarantee added by this PR is not covered yet: the new inner transaction is meant to prevent a failure during a multi-delete phase from committing only a prefix. The regression currently fails during CREATE, so it proves that the delete phase is not entered early, but it never exercises rollback after the first delete has executed.

Could we add a transaction-level regression with two authorized observations where the second delete raises, then read back/assert that the first observation (and its history, where applicable) still exists? That would distinguish the new atomic-delete behavior from a loop without `conn.transaction()` and protect the exact destructive guarantee described in the PR body.

I did not find a counterexample to the update/create-before-delete ordering or the external write-group abort path; this request is specifically about pinning the newly introduced multi-delete rollback contract.

r266-tech:
Thanks — addressed in `f6010b0`.

The new pg0-backed regression seeds two observations with history, verifies that the first delete removes its observation and history inside the active transaction, then injects a failure before the second delete executes. After the exception, it reads through the engine APIs and confirms both observations and both history entries are present, proving the first deletion was rolled back rather than partially committed.

The strengthened target test passes; the related 48-test consolidation gate also passed before this final assertion hardening, with Ruff and diff checks clean.

koriyoshi2041 (approved):
Reviewed exact head f6010b0. The new regression now exercises the requested destructive boundary: it proves the first observation and its history are deleted inside the active transaction, injects a failure before the second delete, then reads back both observations and both history rows after rollback. I ran the focused regression, the complete 8-test failure-isolation file, Ruff on the changed Python surfaces, and diff checks; all passed.

nicoloboschi:
fixed by #3923

PR #3896, opened 2026-08-30
docs(docker): reserve shared memory for embedded PostgreSQL

### LINKED ISSUE #3895: docs(docker): allocate sufficient shared memory for embedded PostgreSQL
## Summary

The documented single-container Docker quick start runs Hindsight with embedded PostgreSQL but does not set the container's shared-memory size. Docker therefore supplies its 64 MiB default, which can be too small for PostgreSQL vector-index creation.

On Hindsight 0.9.2, PostgreSQL attempted to resize a shared-memory segment to 533,794,976 bytes while accessing/initializing a bank and failed with:

```text
asyncpg.exceptions.DiskFullError: could not resize shared memory segment
"/PostgreSQL.<redacted>" to 533794976 bytes: No space left on device
```

The host had ample disk and memory. The Hindsight container reported:

```text
ShmSize=67108864
/dev/shm size=64M
```

Recreating the same container with 1 GiB of shared memory resolved the PostgreSQL failure; API/UI, recall, and reflect then passed.

## Current documentation

The Docker quick start in `hindsight-docs/docs/developer/installation.md` runs the embedded PostgreSQL deployment with `docker run`, but does not include `--shm-size`:

```bash
docker run -it --pull always --name hindsight --restart unless-stopped -p 8888:8888 -p 9999:9999 \
  ...
```

## Suggested change

Add a safe shared-memory allocation to the embedded-database Docker example:

```bash
--shm-size=1g
```

Also add a short note explaining:

- Docker defaults `/dev/shm` to 64 MiB;
- embedded PostgreSQL index operations can exceed that value;
- the Docker Compose equivalent is `shm_size: 1gb`;
- 1 GiB is a safe example rather than a universal minimum, because requirements depend on workload, index type, PostgreSQL settings, and parallelism.

## Environment

- Hindsight: 0.9.2 (`ghcr.io/vectorize-io/hindsight:latest`)
- Deployment: single Hindsight container with embedded PostgreSQL/pg0
- Docker Desktop: 4.88.1
- Host: macOS 15.7.9, Intel x86_64

## Related issue

Related to #2695, where the same 64 MiB container limit and approximately 509 MiB PostgreSQL allocation were reported. A maintainer acknowledged that shared-memory deployment guidance was missing. PR #2802 fixed the separate lazy-bank-creation atomicity problem, but did not update the Docker quick start or shared-memory guidance.

aunitt:
## Summary

- reserve 1 GiB of `/dev/shm` in the documented single-container Docker quick starts
- keep the current and versioned 0.9 installation/quick-start pages aligned

## Why

Hindsight 0.9.2 attempted to allocate a 533,794,976-byte PostgreSQL shared-memory segment during bank index creation. Under Docker's default 64 MiB `/dev/shm`, that failed with `DiskFullError: could not resize shared memory segment ... No space left on device`. Recreating the container with 1 GiB shared memory resolved the failure and restored successful recall and reflect operations.

Closes #3895.

## Verification

- `docker run --help` confirms `--shm-size` is supported
- `npm run build --workspace hindsight-docs`
  - code parity passed for 457 docs files
  - integration and template checks passed
  - Docusaurus production build completed successfully
- `git diff --check`

### DIFF
--- a/hindsight-docs/docs/developer/api/quickstart.mdx
+++ b/hindsight-docs/docs/developer/api/quickstart.mdx
@@ -43,7 +43,7 @@ API available at [http://localhost:8888](http://localhost:8888/docs)
 
 export OPENAI_API_KEY=sk-xxx
 
-docker run -it --pull always --name hindsight --restart unless-stopped -p 8888:8888 -p 9999:9999 \
+docker run -it --pull always --name hindsight --restart unless-stopped --shm-size=1g -p 8888:8888 -p 9999:9999 \
   -e HINDSIGHT_API_LLM_API_KEY=$OPENAI_API_KEY \
   -v $HOME/.hindsight-docker:/home/hindsight/.pg0 \
   ghcr.io/vectorize-io/hindsight:latest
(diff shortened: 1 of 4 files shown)


nicoloboschi (changes requested):
please remove the tip and keep the command arg only

aunitt:
> please remove the tip and keep the command arg only

Thanks, should be done now

PR #3900, opened 2026-08-30
feat(llm): route operations by metadata

cachanova:
## Summary

- Add a `metadata` multi-LLM strategy for exact tag and `metadata.*` routes.
- Route synchronous, queued, and batch retain work from the full operation's classification, including stored append state. Selected lanes never fail over across providers.
- Keep reflect/review, mental-model refresh, consolidation, and consolidation dedup on the protected tag-routed member. Reflect is intentionally fail-closed because retrieval can expand a fact to its full document.
- Reject an operation before any LLM call when matching routes select different members.
- Document global and per-operation configuration and keep the generated Hindsight docs skill synchronized.

## Configuration

```bash
export HINDSIGHT_API_LLM_PROVIDER=openai
export HINDSIGHT_API_LLM_API_KEY=sk-...
export HINDSIGHT_API_LLM_1_PROVIDER=ollama
export HINDSIGHT_API_LLM_1_MODEL=qwen3:8b
export HINDSIGHT_API_LLM_STRATEGY='{
  "mode": "metadata",
  "routes": [{"key": "tags", "value": "sensitive", "member": 1}]
}'
```

Unmatched retains use member `0`. Retain can also route on user-defined values such as `metadata.classification`. Reflect and consolidation route from tags; with a tag route configured, they bind fail-closed to its member because they can admit new or transitive data during the operation. Routing values are ephemeral and are not written to LLM trace metadata.

## Verification

- `./scripts/hooks/lint.sh`
- 200 focused routing, config, provider, batch, reflect, mental-model delta, consolidation dedup, and cache-affinity tests
- 7 embedded env-template tests
- Docusaurus production build and generated-docs link validation
- 238 append/store regression tests passed; one parallel queue-claim timing failure passed on isolated rerun
- `git diff --check`
- Fable final review: no findings, PR-ready

Closes #1140


hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
```diff
@@ -2206,9 +2212,7 @@ async def _process_memory_batch(
     # "consolidation" prefix; recorded distinctly in llm_requests).
     dedup_enabled = _dedup_active(config)
     dedup_llm_config = (
-        memory_engine._consolidation_llm_config.with_config(config, bank_id=bank_id, operation="consolidation_dedup")
-        if dedup_enabled
-        else None
+        llm_config.with_config(config, bank_id=bank_id, operation="consolidation_dedup") if dedup_enabled else None
```
nicoloboschi: **Must fix — this breaks 24 existing consolidation tests.**

`llm_config` is now the `ConfiguredLLMProvider` returned by `with_config` a few lines up, and `ConfiguredLLMProvider` has no `with_config` of its own. It only resolves in production by accident of the `__getattr__` passthrough to `_provider`. Against a test stub it raises:

```
E  AttributeError: 'MockLLM' object has no attribute 'with_config'
   hindsight_api/engine/consolidation/consolidator.py:2215
```

Baseline check: `test_consolidation.py`, `test_consolidation_scope_parallelism.py`, `test_consolidation_failure_isolation.py`, `test_consolidation_failure_recovery.py` and `test_async_retain_tags.py` are **101 passed on `origin/main`**, 26 failed on this branch.

Suggested fix — keep the original call site and just thread the routing metadata through it:

```python
dedup_llm_config = (
    memory_engine._consolidation_llm_config.with_config(
        config, bank_id=bank_id, operation="consolidation_dedup", routing_metadata=routing_metadata
    )
    if dedup_enabled
    else None
)
```

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -18217,6 +18427,23 @@ async def submit_async_retain(
                 f"batch synchronously (async=false), which processes them sequentially."
             )
 
+        # Validate the parent payload before child packing, database inserts, or
+        # task dispatch. Otherwise classifications routed to different members
+        # can land in separate workers and each appear unambiguous in isolation.
+        routing_contents = cast(list[RetainContentDict], contents)
+        if (
+            isinstance(self._retain_llm_config, MultiLLMProvider)
```
nicoloboschi, line 19735: **Must fix — this breaks `test_async_retain_tags.py` (both parametrizations).**

`submit_async_retain` now reads `self._retain_llm_config` unconditionally, but the test constructs the engine via `MemoryEngine.__new__` and never sets that attribute:

```
E  AttributeError: 'MemoryEngine' object has no attribute '_retain_llm_config'
   hindsight_api/engine/memory_engine.py:18435
```

Production is fine (`__init__` always assigns it at line 2183), so the fix belongs in the test — `engine._retain_llm_config = MagicMock()` alongside the other stubs.

Worth flagging that this PR *edits that exact test file* (adding the `[]` parametrization) and leaves it red, which suggests it wasn't run. Note that this is a fork PR, so the `test-api` jobs are **skipped**, not failing — a green check here does not mean the suite passes.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -5287,6 +5378,104 @@ def _retain_chunking_config(config: HindsightConfig) -> _RetainChunkingConfig:
...
+                    bank_id,
+                )
+                tags = row["tags"] if row else None
+                retain_params = row["retain_params"] if row else None
+
+                # A store-owned document may have no SQL row at all. Even when a
+                # metadata row exists, the store record is authoritative and is
+                # the same record the append body path reads before prepending.
+                if store.owns_document_store_for(bank_id):
+                    record = await store.get_document_record(
```
nicoloboschi: **Should fix — DB connection held across external store I/O.**

`store.get_document_record()` is called inside the `async with acquire_with_retry(backend) as conn:` block, once per document. For an extension-backed store that is a remote call made while holding a pooled Postgres connection, which is the pattern #2434 exists to prevent.

Same loop is also an N+1: the per-document `fetchrow` could be a single `WHERE id = ANY($1) AND bank_id = $2`. Suggest collecting the SQL rows in one query, releasing the connection, then doing the store reads.

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -4973,6 +5019,45 @@ async def retain_batch_async(
...
+            stored_state = stored_append_states.get(item["document_id"])
+            if stored_state is None:
+                continue
+            item_tags = item.get("tags")
+            tags_explicitly_cleared = item_tags == [] or document_tags == []
+            if stored_state.tags and not tags_explicitly_cleared:
+                item["tags"] = list(dict.fromkeys([*stored_state.tags, *(item_tags or [])]))
+
+            item_metadata = item.get("metadata")
+            if stored_state.metadata and item_metadata != {}:
```
nicoloboschi, line 5223: **Should fix — `metadata.*` routes are only persisted for `items[0]`.**

This inherits `stored_state.metadata`, which `_stored_append_routing_states` reads out of `documents.retain_params`. But `_build_retain_params` (`hindsight_api/engine/retain/orchestrator.py:392`) only captures `metadata` from the **first** item:

```python
first_item = items[0]
...
if first_item.get("metadata"):
    retain_params["metadata"] = first_item["metadata"]
```

whereas `merged_tags` unions across *all* items. So for a multi-item retain folded into one `document_id` — which `retain_batch_async` explicitly supports — where only a non-first item carries `metadata.classification`, nothing is stored. A later append that omits metadata inherits nothing and routes to the **primary**, which is exactly the disclosure this feature exists to prevent. `tags` routes are unaffected.

Either union `metadata` across items the way tags already are, or document that `metadata.*` routing is unsupported for multi-item documents.

Minor, same block: this mutates the caller's `contents` dicts in place.

hindsight-api-slim/hindsight_api/engine/multi_llm.py
```diff
@@ -97,16 +100,59 @@ def __init__(self, members: list[LLMProvider], strategy: LLMStrategyConfig) -> N
...
 
+        if strategy.mode == LLM_STRATEGY_METADATA:
+            if not strategy.routes:
+                raise ValueError("Metadata LLM routing requires at least one route")
+            for route in strategy.routes:
+                if route.member >= len(members):
+                    raise ValueError(
+                        f"LLM metadata route {route.key}={route.value!r} selects member {route.member}, "
+                        f"but the chain has members 0..{len(members) - 1}."
+                    )
```
nicoloboschi, line 111: **Should fix — validate conflicting tag routes here, at construction.**

This constructor already rejects an out-of-range `member`, but accepts a config that makes reflect and consolidation fail permanently. Verified against this branch:

```json
{"mode":"metadata","routes":[{"key":"tags","value":"sensitive","member":1},
                             {"key":"tags","value":"pii","member":2}]}
```

`_reflect_llm_routing_metadata` / `_pending_consolidation_llm_routing_metadata` union *all* tag route values, so `_member_for_metadata` sees two distinct members and raises on **every** reflect call and **every** consolidation job:

```
ValueError: LLM metadata routes for this operation select multiple members; ...
```

`configuration.md:573` documents this as intended, so the limitation itself is a deliberate call — but it should be rejected at startup next to the member-index check, rather than surfacing as a runtime `ValueError` from inside `asyncio.wait_for` in `reflect_async` on a config that parsed cleanly.

hindsight-api-slim/hindsight_api/engine/multi_llm.py
```diff
@@ -97,16 +100,59 @@ def __init__(self, members: list[LLMProvider], strategy: LLMStrategyConfig) -> N
...
+            )
+            if matches:
+                matched_members.append(route.member)
+
+        distinct_members = set(matched_members)
+        if len(distinct_members) > 1:
+            raise ValueError(
+                "LLM metadata routes for this operation select multiple members; "
+                "split the input or route all matching classifications to one member."
+            )
```
nicoloboschi, line 154: Nit — the surrounding docs describe a different algorithm than this implements.

This is *all-matches-must-agree*, and declaration order has no effect on the outcome. But `config.py:1943` says `"metadata" (pin to the **first matching route**, defaulting to the primary)` and `` `routes` ... is evaluated in declaration order``, and `.env.example:144` says `Metadata routes are **ordered** and strict`. Either drop the ordering language, or make it genuinely first-match-wins (which would also resolve the reflect/consolidation ambiguity above).

hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -18319,7 +18546,7 @@ async def submit_async_retain(
                             )
 
                         task_payload: dict[str, Any] = {"contents": sub_batch}
-                        if document_tags:
+                        if document_tags is not None:
```
nicoloboschi, line 19849: Nit — inconsistent with the other `document_tags` dispatch path.

This was widened to `is not None` so an explicit `[]` declassification survives queueing, but the file-convert re-dispatch at `memory_engine.py:2846` still uses truthiness:

```python
if document_tags:
    retain_task_payload["document_tags"] = document_tags
```

so `[]` is dropped there, `tags_explicitly_cleared` comes out `False` in `retain_batch_async`, and the stored classification is re-inherited instead of cleared.

nicoloboschi (changes requested):
Reviewed and tested locally on `57f3876` (rebased check: the branch is 33 commits behind `main`).

**Verdict: not mergeable as-is** — it breaks 26 existing tests, and CI will not tell you. This is a fork PR, so the `test-api` jobs are *skipped* rather than run; the green checkmarks here cover a much smaller subset than they appear to.

### Verification performed

| Check | Result |
|---|---|
| `./scripts/hooks/lint.sh` | pass |
| Generated docs skill in sync (`git status` after lint) | clean |
| New tests: routing, config, provider, batch | 84 passed |
| Full `-k "retain or multi_llm or consolidation or reflect or tag or metadata or config"` | **27 failed**, 1819 passed |
| Same suites on `origin/main` | **101 passed, 0 failed** |

Of the 27 failures, 26 are introduced by this PR (24 consolidation + 2 async-retain-tags; see the two inline **Must fix** comments). The remaining two — `test_worker.py::...test_claim_batch_skips_consolidation_when_same_bank_processing` and `test_fact_extraction_agent_experience.py::...test_user_interaction_classified_as_experience` — both pass in isolation and are pre-existing flakes, not attributable here.

### Summary of findings

**Must fix**
1. `consolidator.py:2215` — `with_config` called on an already-configured wrapper; works in prod only via `__getattr__` passthrough, breaks 24 tests.
2. `memory_engine.py:18435` — unconditional `self._retain_llm_config` access breaks the test this PR itself edits.

**Should fix**

3. `metadata.*` classification is only persisted for `items[0]`, so a multi-item document can silently lose its route and a later append lands on the primary.
4. Pooled DB connection held across external store I/O in `_stored_append_routing_states`, plus an N+1.
5. Two tag routes to different members construct fine but break every reflect and every consolidation at runtime — validate it in `MultiLLMProvider.__init__`.

**Nits** — comment/implementation mismatch on route ordering; `document_tags=[]` still dropped on the file-convert path.

### What holds up well

The batch path is genuinely well covered — `test_multi_llm_batch.py` verifies the sensitive lane owns both submit *and* resume, which was the leak path I most expected to find untested. The fail-closed reasoning for reflect (expand can widen a fact to its full document) and for consolidation (facts arriving mid-job) is sound and well documented. Routing values correctly stay out of LLM trace metadata.

cachanova:
Addressed every review item and rebased onto current `main` in `02399a53`:

- derived consolidation dedup from the base provider and initialized the async-retain test fixture
- persisted metadata from every item in shared documents so later appends retain their route
- replaced append-state N+1 reads with one bank-scoped query, released the DB connection before store I/O, and used the bulk store API
- rejected conflicting `tags` routes at construction
- corrected all-match documentation and preserved explicit `document_tags=[]`
- added regression coverage for shared-document metadata, authoritative store routing, bulk store reads outside the DB lease, tag-route validation, and empty file-retain tags

Validation: repository lint and `ty` pass; 82 focused tests pass; broad deterministic selection completed with 1,858 passes. Its two remaining failures require the missing live OpenAI credential, and the one shared-worker failure passed in isolation.

@nicoloboschi this should be ready for re-review. @strix-security please review the latest commits.

cachanova:
Rebased once more onto current `main` (`ebcd88f9`) after today’s upstream changes and updated the metadata batch tests for the current extraction API. New head: `0954c0de`.

Fresh validation on the rebased branch:
- broad retain/routing/consolidation selection: **1,941 passed, 39 skipped, 1 xfailed**
- focused metadata/multi-LLM/consolidation/mental-model set: **177 passed, 6 skipped** before the two compatibility edits; the affected batch file then passed **18/18**
- isolated shared-worker test: **passed**
- repository lint and `ty check hindsight_api/`: **passed**

All seven addressed review threads are now resolved, and GitHub reports the PR mergeable. @nicoloboschi ready for re-review when convenient. @strix-security please review the latest head.

nicoloboschi:
superseded by #4188

PR #3902, opened 2026-08-30
fix(recall): preserve coarse occurrence date precision

### LINKED ISSUE #3893: Year-only dates normalized to Jan 1 let recency boost reverse the correct reranker order
## Bug Description

Hindsight 0.9.2 applies a recency boost to every dated memory during recall using the default combined-scoring formula:

```text
combined_score =
  cross_encoder_score_normalized
  * recency_boost
  * temporal_boost
  * proof_count_boost
```

When a memory contains only a year, such as `When: 2026`, Hindsight represents its occurrence time as `2026-01-01`. The default 365-day linear recency decay then treats it as an event that happened at the beginning of the year.

As a result, a newer but clearly unrelated candidate can outrank the candidate that both semantic retrieval and the cross-encoder correctly identify as the best match.

## Steps to Reproduce

1. Run Hindsight `0.9.2-slim`.
2. Leave `HINDSIGHT_API_RECENCY_DECAY_FUNCTION` unset, so the default `linear` function and 365-day window are used.
3. Retain these two memories:

   ```text
   The user attended the 2026 Hangzhou Open Source Summit and gave a presentation about local-first AI memory.
   On August 24, 2026, the user changed the UI to the light theme; the current theme should be light.
   ```

   The actual retained Chinese facts in this reproduction were:

   ```text
   用户参加了2026年杭州开源峰会，并做了关于本地优先AI记忆的分享。
   2026年8月24日用户把界面改成浅色主题，当前应使用浅色。
   ```

4. With the recall scoring date set to `2026-08-30`, recall:

   ```text
   What topic did the user present at the 2026 Hangzhou Open Source Summit?
   ```

   The actual Chinese query was:

   ```text
   用户在2026年杭州开源峰会分享了什么主题？
   ```

5. Inspect the cross-encoder, recency, and combined scores in the recall trace.

## Expected Behavior

When a query explicitly identifies an event, a newer unrelated fact should not outrank a strongly relevant result solely because of a recency boost.

A date inferred only to year or month precision should not be treated as if the imputed first day were an exact occurrence date for recency scoring. Possible approaches include:

- Preserve date granularity and keep recency neutral for year-level dates.
- Represent coarse dates as ranges rather than single timestamps.
- Apply recency preference only when the query expresses current/latest/recent intent.
- Prevent secondary temporal signals from overriding a cross-encoder result with a clear relevance advantage.

## Actual Behavior

Semantic retrieval and the cross-encoder both rank the summit memory first, but Hindsight's recency post-processing reverses the final order:

| Candidate | Normalized cross-encoder score | Recency | Final score | Final rank |
|---|---:|---:|---:|---:|
| Correct: local-first AI memory presentation | **0.999924** | 0.339549 | 0.967836 | 2 |
| Incorrect: light UI theme | 0.900004 | **0.983384** | **0.987013** | 1 |

Semantic similarity also favors the correct candidate:

```text
Correct candidate:   0.672282
Incorrect candidate: 0.417432
```

Two consecutive recall passes produced the same ordering.

## Impact

The Recall API returns the wrong memory at Top-1. An agent that consumes only Top-1, uses a small fact/token budget, or overweights the first memory can therefore produce an incorrect answer.

In this reproduction the correct memory remained at Top-2, so Reflect received it and ultimately answered correctly. That recovery depends on the downstream fact budget and LLM, and does not eliminate the recall-ranking error.

Year-only and month-only dates are common in personal memory, so this is not specific to synthetic benchmark data.

## Workaround and Counterfactual Check

The unconditional recency boost can be disabled with:

```yaml
HINDSIGHT_API_RECENCY_DECAY_FUNCTION: "none"
```

For a local evaluation containing 37 answerable personal-memory queries, recomputing the recorded traces with neutral recency produced:

```text
Hit@1:    36/37 -> 37/37
Recall@5: 37/37 -> 37/37
MRR:      0.98649 -> 1.0
```

This counterfactual is specific to that evaluation, but it confirms that the observed rank reversal was introduced by recency post-processing rather than semantic retrieval or the reranker.

## Related Issue

#1272 discussed making the boost alpha values configurable and was later withdrawn by its author. This report is different: it documents a concrete ranking-correctness failure caused by treating an imputed coarse date as exact and applying query-independent recency weighting.

## Environment

```text
Hindsight: 0.9.2-slim
LLM provider: OpenAI-compatible Responses API
Embedding provider: TEI
Reranker provider: TEI
Vector backend: pgvector
Text search: PGroonga
```

## LLM Provider

OpenAI-compatible Responses API. The ordering reversal occurs after reranking in Hindsight's deterministic combined-scoring step and is not specific to the LLM provider.

vihor3:
## What changed

- Added internal occurrence-precision metadata to distinguish years, months, exact days, instants, and genuine ranges instead of treating `2026` or `2026-08` as an exact first day.
- Preserved occurrence precision through streaming and Batch extraction, Retain, metadata-only updates, transfer, and memory curation, while preventing callers from overriding the engine-owned metadata key.
- Changed recall recency scoring for year- and month-level occurrences to use their full possible calendar interval conservatively, while keeping existing behavior for exact dates, genuine ranges, and memories without an occurrence time.
- Updated embedding enrichment, BM25 signals, and reranker context to format dates at their actual precision without inventing missing calendar components.
- Added narrow compatibility recovery for canonical coarse `| When: ... |` values in existing memories, without requiring a migration or bulk backfill.
- Made concurrent memory curation safe across both race windows: the complete edit plan is retried when the memory changes between Phase 1 and Phase 2, and the final write uses an `updated_at` conditional update for changes after the Phase 2 snapshot.
- Ensured omitted edit fields come from the latest version-checked snapshot so disjoint concurrent changes are merged instead of overwritten.
- Preserved the existing queue-to-memory lock order to avoid deadlocks with delete and invalidate operations.
- Updated the related documentation and regression coverage.
- Kept the public Retain/Recall API and database schema unchanged.

Closes #3893


nicoloboschi:
fixed by https://github.com/vectorize-io/hindsight/pull/3933

PR #3905, opened 2026-08-31
fix(templates): defer empty-bank auto-refreshes until consolidation

### LINKED ISSUE #3875: refresh_mental_model runs against empty banks at init, head-of-line blocking the bank's own seeding
## Summary

On a newly created bank, the five default knowledge pages (mental models) are created at bank init and **each immediately enqueues a `refresh_mental_model`**, before the bank contains any memories. Those reflects are guaranteed to find nothing, but they still occupy the LLM concurrency slots — head-of-line blocking the `retain` operations that would have given the bank its content.

The net effect on a self-hosted instance with modest concurrency: **a fresh bank cannot seed itself.** It sits at `total_nodes=0` while its retain queue grows, because every available slot is held by a reflect over an empty graph.

## Version

Hindsight `0.9.2` (`ghcr.io/vectorize-io/hindsight:latest`), self-hosted, `ApiKeyTenantExtension`, Postgres `public` schema.

## Evidence

Two banks created within minutes of each other. All five refreshes on each are enqueued inside a **half-second window**:

```
bank A    06:14:03.696 -> 06:14:04.093    (5 refresh_mental_model, 0.4s)
bank B    06:08:53.492 -> 06:08:54.009    (5 refresh_mental_model, 0.5s)
```

State of a fresh bank some time later — note `total_nodes=0` with reflects still occupying slots while retains wait:

```
bank A   nodes=0   links=0   docs=1   pending=19 (batch_retain + retain)   processing=4 (refresh_mental_model)
```

Worker view, showing the reserved/shared pool fully consumed:

```
[WORKER_STATS] slots=10/10 | reserved: [consolidation=2/2(avail=0)] | shared=8/8(avail=0)
  my_active: ... refresh_mental_model:bankB(2), refresh_mental_model:bankA(1)
[PENDING_BREAKDOWN] refresh_mental_model: total=8 claimable=8 | retain: total=12 claimable=12
```

Each of those reflects then burns the full reflect wall clock before failing:

```
TimeoutError: Reflect operation timed out after 300 seconds.
  Consider reducing the budget or simplifying the query.
```

...and is then **retried**, repeating the cost up to the retry limit:

```
Task <id> scheduled for retry at <t+60s>: TimeoutError
```

For context on scale, the same instance shows roughly a 45% success rate for `refresh_mental_model` overall (9 completed vs 11 failed on an established bank). On an *empty* bank the success rate is effectively zero by construction, since there is nothing to reflect over.

## Expected behaviour

`refresh_mental_model` should short-circuit when the target bank has no memories to reflect over — return early (or skip enqueueing entirely) rather than running a full agentic reflect loop against an empty graph.

## Suggested fix

Either of:

1. **Guard at execution** — in the `refresh_mental_model` handler, check the bank's memory count (or the same `based_on`-empty condition already discussed in #2894) *before* entering the reflect loop, and no-op if there is nothing to reflect over.
2. **Defer at creation** — do not enqueue the initial refresh when the bank is created empty; let the first refresh be triggered by the first consolidation that actually has content.

(2) is the stronger fix for this specific symptom, since it avoids spending a queue slot at all. (1) is more general and would also cover a bank that has been emptied.

Independently of the fix, it would help if a refresh that finds nothing failed *fast* rather than consuming the full 300s reflect budget and then retrying.

## Impact / workaround

Workaround is to set `refresh_after_consolidation: false` on every mental model via
`PATCH /v1/{tenant}/banks/{bank}/mental-models/{id}`, then refresh manually. That stops new refreshes, but does not help ops already in flight:

`DELETE /v1/{tenant}/banks/{bank}/operations/{id}` returns **409 Conflict** for an operation in `processing`, so an already-running empty reflect cannot be cancelled — you have to wait out the 300s timeout plus its retries. Being able to cancel a running operation would be a useful escape hatch here (possibly related to the zombie-operation handling in #3594).

## Related

- #2894 — `refresh_mental_model` overwrites content when `based_on` is empty. Same empty-reflect condition, but that issue is about the *content* written; this one is about the *scheduling cost* of running it at all.
- #3487, #3210 — prior work on refresh dedupe / duplicate enqueue. This is a distinct trigger: not duplicates, but a first-and-only refresh fired before the bank has content.
- #3595 — consolidation reservation starving the shared pool. Related contention shape.

Ftgn-dpA:
## Summary

- defer the initial refresh for server-default mental models configured to refresh after consolidation
- preserve immediate refresh behavior for client imports and manual default mental models
- add regression coverage for all three scheduling paths

## Why

Server default templates are applied while freshly created banks have no memories. Enqueueing their auto-refreshing mental models at that point starts reflect operations that cannot produce content and can consume LLM worker slots ahead of the retain operations that would seed the bank.

The first consolidation already schedules these models once content exists. Manual default mental models without an automatic consolidation trigger retain their initial refresh so they do not remain at the placeholder indefinitely.

Fixes #3875

## Verification

- targeted bank-template regression tests — 3 passed
- `uv run ruff check hindsight_api/api/http.py tests/test_bank_templates.py`
- `uv run ruff format --check hindsight_api/api/http.py tests/test_bank_templates.py`
- `uv run ty check hindsight_api/api/http.py`

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -3161,6 +3161,7 @@ async def apply_bank_template_manifest(
             existing_by_name,
             request_context,
             config_applied=bool(config_updates),
+            defer_until_consolidation=False,
         )
 
 
@@ -3194,6 +3195,7 @@ async def apply_default_bank_template_resources(
         existing_by_name,
         request_context,
         config_applied=False,
+        defer_until_consolidation=True,
     )
 
 
@@ -3206,6 +3208,7 @@ async def _apply_bank_template_resources(
     request_context: "RequestContext",
     *,
     config_applied: bool,
+    defer_until_consolidation: bool,
 ) -> "BankTemplateImportResponse":
     """Apply template resources after the caller has handled config and access."""
     created_ids: list[str] = []
@@ -3214,6 +3217,11 @@ async def _apply_bank_template_resources(
 
     if manifest.mental_models:
         for mm in manifest.mental_models:
+            # Server defaults are installed before a new bank has memories. Auto-refreshing models can wait for the
+            # first consolidation instead of occupying an LLM slot with a reflect that cannot produce content (#3875).
+            enqueue_initial_refresh = not (
+                defer_until_consolidation and mm.trigger is not None and mm.trigger.refresh_after_consolidation
+            )
             if mm.id in existing_mental_models:
                 await memory.update_mental_model(
                     bank_id=bank_id,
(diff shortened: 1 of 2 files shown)


nicoloboschi:
duplicate of #3943

Issue #1875, opened 2026-05-30
Recall/embedding returns 500 on queries containing an unpaired UTF-16 surrogate (TextEncodeInput)


maximburdiean:
> **Transparency note (please read first):** This bug was found, investigated, and written up by an **AI coding agent (Claude Code)**, and the suggested fix below is **also AI-proposed**. A human account owner is sponsoring the submission and has reviewed it, but the analysis and wording are the agent's, not a person's. Flagging this explicitly and up front, so you can weigh it as an AI-authored report rather than assume it is hand-written. Context: we run Hindsight as the long-term memory layer for our engineering projects and hit this 500 in real usage. If you would rather not receive agent-authored issues, say so and we will stop.

### Summary

With the local SentenceTransformers embeddings backend (`LocalSTEmbeddings`, e.g. `BAAI/bge-small-en-v1.5`), a recall whose query text contains an **unpaired UTF-16 surrogate** (e.g. a half-emoji that a client serialized as a `\udXXX` JSON escape) fails with HTTP 500:

```
TypeError: TextEncodeInput must be Union[TextInputSequence, Tuple[InputSequence, InputSequence]]
 -> Exception: Failed to generate batch embeddings: ...
 -> RuntimeError: Failed to search memories (Exception): ...
```

### Root cause

The recall length precheck uses `tiktoken`, which tolerates lone surrogates, so the string passes validation and reaches the embedder. The Rust `tokenizers` backend behind the local ST model rejects any string containing an unpaired surrogate with the `TextEncodeInput` `TypeError`. Call path: `memory_engine._search_with_retries` -> `embedding_utils.generate_embeddings_batch(..., "query")` -> `LocalSTEmbeddings.encode` -> `self._model.encode(texts)`.

### Reproduce

```python
from sentence_transformers import SentenceTransformer
m = SentenceTransformer("BAAI/bge-small-en-v1.5")
m.encode(["hello world"])             # OK
m.encode(["bad \udce2 text"])         # TypeError: TextEncodeInput must be Union[...]
```

Or via the API: a recall body `{"query": "deploy the \ud83d service"}` returns 500, while a clean ASCII query returns 200. `None` is *not* the trigger (it raises a different `ValueError: Unsupported input type`); the unique trigger is an unpaired surrogate. Both the recall (`encode_query`) and retain (`encode_documents`) paths are affected, since both go through `encode()`.

### Suggested fix

[...]

PR #3908, opened 2026-08-31
fix(retain): strip NUL from retain items instead of 500ing on the jsonb insert

slayoffer:
## Problem

PostgreSQL cannot store `U+0000` in `text` or `jsonb`. `submit_async_retain` writes the whole retain item into `async_operations.task_payload::jsonb`, so one NUL anywhere in the payload aborts that INSERT:

```
asyncpg.exceptions.UntranslatableCharacterError: unsupported Unicode escape sequence
DETAIL:  \^@ cannot be converted to text.
```

The request returns HTTP 500 and the memory is never queued. The failure is deterministic for that payload, so a retrying client re-sends it forever and never stores it.

JSON permits `\^@`, so clients legitimately send it — text scraped from PDFs, log captures and chat exports all carry stray NULs.

## Reproduction (0.9.2)

```
POST /v1/default/banks/<bank>/memories
{"items":[{"content":"hello \^@ world","document_id":"x"}], "async": true}

-> HTTP 500  unsupported Unicode escape sequence
```

A NUL in `metadata` fails identically. The same payload with `"async": false` succeeds, and without the NUL both succeed — so only batched/async ingestion is affected, which makes this easy to miss: the payloads that carry a NUL are dropped while everything else keeps succeeding, so a bank shows gaps in its history rather than an outage, and nothing surfaces but 500s in the API log.

## Fix

A `MemoryItem` validator (`mode="before"`) strips `U+0000` from every string in the item — content, context, document_id, metadata (keys and nested values), tags. This covers the sync and async paths and any future consumer of the model.

Stripping rather than returning 400: the NUL is never meaningful content — it is transport residue — and a 400 would lose the memory just as completely as the 500 does today.

## Tests

- `test_retain_request_strips_nul_characters` — NUL removed from every field including nested metadata and dict keys, and asserted absent from the serialized item.
- `test_retain_request_preserves_non_nul_text` — other unicode (accents, em dash, emoji, tabs, newlines) passes through byte-for-byte.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -632,6 +632,23 @@ class EntityInput(BaseModel):
     type: str | None = Field(default=None, description="Optional entity type (e.g., 'PERSON', 'ORG', 'CONCEPT')")
 
 
+def _strip_nul(value):
+    """Recursively remove NUL (U+0000) from strings, keys included.
+
+    PostgreSQL rejects U+0000 in both ``text`` and ``jsonb``, so it can never
+    reach storage; see ``MemoryItem.strip_nul_characters``.
+    """
+    if isinstance(value, str):
+        return value.replace("\x00", "") if "\x00" in value else value
+    if isinstance(value, dict):
+        return {_strip_nul(k): _strip_nul(v) for k, v in value.items()}
+    if isinstance(value, list):
+        return [_strip_nul(v) for v in value]
+    if isinstance(value, tuple):
+        return tuple(_strip_nul(v) for v in value)
+    return value
+
+
 class MemoryItem(BaseModel):
     """Single memory item for retain."""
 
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Closing in favour of a fix that covers the same failure at the engine ingress instead of on the HTTP model.

Two things this PR's placement leaves open:

1. `MemoryItem` is not the only door into `submit_async_retain`. `hindsight_api/mcp_tools.py:698` and `:753` build a raw content dict via `build_content_dict(...)` and call the engine directly, never touching the Pydantic model — the MCP retain tool still 500s on the same payload. The synchronous path already scrubs at the engine (`memory_engine.py:4930-4936`) precisely so every caller is covered; the async path should be symmetric with it.

2. `_strip_nul` re-derives a narrower version of `sanitize_text()` (`engine/llm_wrapper.py:166`), which already strips `U+0000` plus the other control characters and lone UTF-16 surrogates, and is documented as the retain/recall/reflect ingress scrubber (#1875). A lone surrogate in `metadata` aborts the very same `task_payload::jsonb` INSERT — asyncpg cannot UTF-8-encode it — so the narrower rule fixes one member of a class it leaves half-open.

The diagnosis and the strip-don't-reject call are right, and the replacement keeps both (plus your two tests). Thanks for the clear reproduction.

nicoloboschi:
Replaced by #3921, which keeps your diagnosis and your strip-don't-reject call, and moves the scrub to the engine ingress so the MCP retain tool is covered too. Two things it turned up while building on your reproduction: the synchronous path is not actually exempt (a NUL in `metadata` aborts the document INSERT there as well — `test_sync_retain_with_nul_in_metadata` fails with the same `UntranslatableCharacterError` when the scrub is reverted), and `sanitize_llm_value` in `engine/llm_wrapper.py` already does the recursive walk, so the fix reuses it rather than adding a NUL-only helper. You're credited in the PR.

PR #3910, opened 2026-08-31
fix(anthropic): group consecutive tool results into one user message

zlguo1996:
## The bug

`AnthropicLLM.call_with_tools()` converted each OpenAI `role:tool` message into a separate Anthropic `user` message carrying a single `tool_result` block. Anthropic (and strict gateways in front of it) requires every `tool_use` block in an assistant turn to be answered by a `tool_result` block in the single immediately-following `user` message. When the reflect agent's model returned several tool calls in one turn (e.g. `search_mental_models` + `search_observations` together), the conversion produced:

```
assistant: [tool_use A, tool_use B]
user: [tool_result A]
user: [tool_result B]   <- invalid: B has no result "immediately after"
```

The gateway rejected the second turn with `invalid_request_error` (`tool_use` ids were found without `tool_result` blocks immediately after). Reflect catches the error and falls back to forced final synthesis, so the run still returned an answer — but the evidence gathered by the failed iteration was silently dropped, and ~9–10s of wasted failure latency was added per occurrence. On a self-hosted deployment routing an Anthropic-protocol model through a strict gateway, this fired on roughly every third reflect.

## The fix

Two commits, same protocol invariant:

1. `fix(anthropic): group consecutive tool results into one user message` — group the consecutive `role:tool` run into a single `user` message carrying every `tool_result` block:
   ```
   assistant: [tool_use A, tool_use B]
   user: [tool_result A, tool_result B]
   ```
   Only *consecutive* tool messages are merged, so two independent assistant tool batches still serialize as two separate user messages, and an empty tool result is preserved as an empty `tool_result` block.

2. `fix(reflect): preserve tool-result order for mixed valid/unknown calls` — the reflect agent used to write hallucinated-tool rejections before the allowed tools' results, so a mixed `[allowed, hallucinated, allowed]` batch serialized out of order. Results are now collected keyed by `tool_call_id` and emitted in the model's original `tool_calls` order, matching the assistant `tool_use` order.

## What changes for callers

No public API change. A reflect run whose later iteration previously failed with a 400 now completes that iteration, so the retrieved evidence reaches final synthesis (answers are more complete) and the wasted failure latency disappears from the multi-tool path.

The conversion fix is confined to `AnthropicLLM`; `litellm_llm.py` does not use this OpenAI→Anthropic conversion path and is unchanged.

## Tests

- Updated `test_call_with_tools_marks_last_block_of_tool_result_message` to assert the grouped serialization.
- Added `test_call_with_tools_does_not_merge_tool_results_across_assistant_turns` and `test_call_with_tools_keeps_empty_tool_result_block`.
- Locally verified: `test_anthropic_prompt_caching.py` 7/7; `test_reflect_agent.py` 44 passed (1 real-LLM test skipped, needs key); `test_anthropic_structured_tool_use.py` + `test_reflect_tool_result_fields.py` + `test_anthropic_batch.py` 18/18.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/anthropic_llm.py
@@ -468,22 +468,33 @@ class AnthropicLLM(LLMInterface):
         # Convert messages - handle tool results
         system_prompt = None
         anthropic_messages = []
-        for msg in messages:
+        i = 0
+        while i < len(messages):
+            msg = messages[i]
             role = msg.get("role", "user")
             content = msg.get("content", "")
 
             if role == "system":
                 system_prompt = (system_prompt + "\n\n" + content) if system_prompt else content
+                i += 1
             elif role == "tool":
-                # Anthropic uses tool_result blocks
-                anthropic_messages.append(
-                    {
-                        "role": "user",
-                        "content": [
-                            {"type": "tool_result", "tool_use_id": msg.get("tool_call_id", ""), "content": content}
-                        ],
-                    }
-                )
+                # Anthropic requires every tool_use block in an assistant turn to be
+                # answered by a tool_result block in the single immediately-following
+                # user message. OpenAI-style histories split parallel tool results
+                # into one ``role:tool`` message per result, so group the consecutive
+                # run into one user message carrying every tool_result block.
+                tool_results = []
+                while i < len(messages) and messages[i].get("role") == "tool":
+                    tm = messages[i]
+                    tool_results.append(
+                        {
+                            "type": "tool_result",
+                            "tool_use_id": tm.get("tool_call_id", ""),
+                            "content": tm.get("content", ""),
+                        }
+                    )
+                    i += 1
+                anthropic_messages.append({"role": "user", "content": tool_results})
             elif role == "assistant" and msg.get("tool_calls"):
                 # Convert assistant tool calls
                 tool_use_blocks = []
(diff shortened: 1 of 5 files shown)


ebarkhordar (commented):
I ran this against `23f9a9e2` in a clean `python:3.12-slim` container.

The body names two fixes and only the first is pinned by a test. Reverting each hunk on its own, over the five test files your Tests section lists:

```
PR head, both fixes present      69 passed
only anthropic_llm.py reverted   68 passed, 1 failed
only reflect/agent.py reverted   69 passed
```

(Same 4 API-key errors in every arm.)

Nothing catches the reordering. `test_recovery_from_unknown_tool` at `tests/test_reflect_agent.py:606` and the retry loop at `:767` are the only tests that reach the `hallucinated_tools` partition, and each sends a single tool call, so a mixed batch never gets serialized. Of the 28 `run_reflect_agent(...)` call sites under `tests/`, none returns a batch holding both an allowed and an unknown tool.

One test closes it: a batch of `[recall id=a, invalid_tool id=h, search_observations id=b]`, then read `messages` off `mock_llm.call_with_tools.call_args_list[1].kwargs` and assert

```python
assert [m["tool_call_id"] for m in messages if m.get("role") == "tool"] == ["a", "h", "b"]
```

I ran that one: green at `23f9a9e2`, and with only `reflect/agent.py` reverted

```
E   AssertionError: assert ['h', 'a', 'b'] == ['a', 'h', 'b']
```

I did not test against a live endpoint or a gateway.

PR #3915, opened 2026-08-31
fix(mental-models): accept id-bearing blocks in section operations

### LINKED ISSUE #3901: Mental model delta refresh: every add_section operation fails validation (blocks emitted as objects, schema requires list[str])
### Bug Description

On `0.9.2`, every `add_section` operation emitted during a mental-model delta refresh fails pydantic validation, because the model returns `blocks` as a list of objects rather than a list of strings.

When a refresh round emits **only** `add_section` operations, all of them are dropped, `mode_fallback_reason` is set to `delta_ops_failed`, and the refresh raises `MentalModelRefreshError` instead of writing.

`AddSectionOp.blocks` is typed `list[str]`:

```python
# engine/reflect/delta_ops.py
class AddSectionOp(_OpBase):
    op: Literal["add_section"] = "add_section"
    heading: str
    level: int = Field(default=2, ge=1, le=6)
    blocks: list[str] = Field(default_factory=list)
    after_section_id: str | None = None
    new_id: str | None = None
```

The model instead emits each block as an object carrying an id alongside the text:

```json
{
  "op": "add_section",
  "heading": "Example Heading",
  "blocks": [
    {"id": "b12a001", "text": "First paragraph of the new section."},
    {"id": "b12a002", "text": "Second paragraph of the new section."}
  ]
}
```

This looks like predictable model behaviour rather than random malformation. Existing blocks are addressed by `block_id` in `replace_block` and `remove_block`, so a model that has just been shown an id-bearing document structure emits ids for the blocks it creates. The schema is the only place saying new blocks are bare strings.

No data is lost — refusing to write rather than truncate is correct behaviour, and content lands on a later round whose operations happen to validate. The cost is wasted spend: each failed round pays a full `reflect` before the delta call, then discards both.

### Steps to Reproduce

1. Configure a bank with an OpenAI-compatible provider (see LLM Provider below).
2. Create a mental model with `trigger.mode = "delta"` and let it accumulate a populated document.
3. Add memories that introduce genuinely new topics — enough that a refresh wants to add sections rather than amend existing ones.
4. Trigger a refresh. `POST .../mental-models/{id}/dry-run-refresh` reproduces it without persisting.

The trigger condition is specifically **new sections**. Rounds that only need `replace_block` succeed normally, which is what makes it look intermittent.

### Expected Behavior

The emitted `add_section` operations are applied and the new sections are added to the document.

### Actual Behavior

Every block of every `add_section` operation is rejected:

```
[STRUCTURED_DELTA] skipping invalid operation at index 0:
  [{'type': 'string_type', 'loc': ('add_section', 'blocks', 0),
    'msg': 'Input should be a valid string', 'input': {'id': '...', 'text': '...'}}]
[STRUCTURED_DELTA] parsed 0 op(s), skipped 7 invalid op(s)
[MENTAL_MODELS] Structured delta failed for <model-id> (all 7 delta operation(s)
    failed validation); delta operations were not applied
[MENTAL_MODELS] Refresh for <model-id> failed (delta_ops_failed); delta operations
    did not reach the document, and the reflect candidate covers only memories newer
    than the last refresh, so writing it would drop the rest of the document.
```

Across three banks on one instance over a 30-hour window:

- **74 rejected operations. All 74 are `add_section`.** Zero rejections of any other operation type.
- All 74 carry the same message, `Input should be a valid string`.
- 23 refresh rounds failed wholesale, every operation in the round rejected.

Exposure tracked how much new material a round wanted to add, not anything bank-specific. All models involved were on identical `delta` configuration.

### Version

`0.9.2`

### LLM Provider

Other — `LongCat-2.0` via an OpenAI-compatible endpoint (`HINDSIGHT_API_LLM_PROVIDER=openai`).

### Suggested fix

A coercion in the operation parser, accepting `{"text": ...}` and ignoring any `id` wherever `list[str]` is expected, would make this class of response valid without loosening the schema. There is precedent in the changelog: JSON-string tags are already coerced to lists for `MemoryItem` and MCP tools to prevent tagging errors.

Alternatively, falling back to full synthesis when *every* operation in a round fails validation would avoid the hard failure, though it changes what the refresh writes.

A third option, possibly cheapest: have the delta prompt show new-block examples in the shape the schema actually requires, so the model is not shown ids in one place and denied them in another.

### Additional context — on `HINDSIGHT_API_LLM_STRICT_SCHEMA`

Since #1656 landed, the obvious remedy is to grammar-enforce the schema. Two things stop that working here, and both may be worth documenting regardless of this bug.

**1. `mental_model_delta_ops` reads only the global flag.** It passes no per-call value, so `llm_wrapper.py` resolves from `get_config().llm_strict_schema`, and `_scope_to_operation` lists `mental_model_delta_ops` among the scopes that "don't belong to a tracked operation". So `HINDSIGHT_API_LLM_STRICT_SCHEMA_REFLECT` does **not** reach it, despite the call living under `engine/reflect/`. The global is the only lever. That is easy to get wrong.

**2. With the global set to `true`, the malformed shape still arrives.** `openai_compatible_llm.py` does build the strict block:

```python
call_params["response_format"] = {
    "type": "json_schema",
    "json_schema": {"name": "response", "strict": True, "schema": schema},
}
```

The request carries it, the provider returns 200, and `blocks` still contains objects. So this provider accepts `strict: true` on a discriminated union of eight operation types and does not enforce it.

I am **not** reporting that second point as a Hindsight bug — it looks like a provider limitation. It is here only because it removes strict schema as a workaround for the issue above. If some OpenAI-compatible endpoints are expected to silently ignore `strict` on union schemas, a note near the flag's documentation would be worth having.

Happy to test a patch against an instance that reproduces this reliably.

oldnicke:
Fixes #3901

## Summary

Mental-model delta models can echo the current document's `{id, text}` block shape when creating or replacing section blocks. The operation schema stores markdown as `list[str]`, so those operations were previously rejected and an entire refresh could fail.

This adds narrow Pydantic input coercion for `add_section.blocks` and `replace_section_blocks.blocks`:

- Retains string `text` values.
- Discards display-only `id` fields.
- Continues rejecting malformed entries.

## Validation

```text
uv run pytest -n 0 tests/test_delta_operation_parse.py tests/test_structured_doc.py tests/test_delta_prompt_schema_parity.py -q
./scripts/hooks/lint.sh
uv run ty check hindsight_api
```

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/delta_ops.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/delta_ops.py
@@ -38,9 +38,10 @@ from __future__ import annotations
 import json
 import logging
 import re
+from collections.abc import Mapping
 from typing import Annotated, Any, Literal, Union
 
-from pydantic import BaseModel, ConfigDict, Field, TypeAdapter, ValidationError
+from pydantic import BaseModel, ConfigDict, Field, TypeAdapter, ValidationError, field_validator
 
 from hindsight_api.engine.llm_wrapper import parse_llm_json
 
@@ -62,6 +63,20 @@ logger = logging.getLogger(__name__)
 
 class _OpBase(BaseModel):
     model_config = ConfigDict(extra="forbid")
+
+
+def _coerce_block_texts(value: Any) -> Any:
+    """Accept id-bearing block objects where an operation expects markdown strings.
+
+    The current document is shown to the model with block ids alongside their
+    text, so some providers echo that shape for newly-created blocks.  Operation
+    payloads only persist the markdown text; discard the display-only id before
+    normal Pydantic validation.  Values without a string ``text`` stay untouched
+    so malformed blocks are still rejected rather than silently dropped.
+    """
+    if not isinstance(value, list):
+        return value
+    return [item["text"] if isinstance(item, Mapping) and isinstance(item.get("text"), str) else item for item in value]
 
 
 class AppendBlockOp(_OpBase):
@@ -117,6 +132,11 @@ class AddSectionOp(_OpBase):
     after_section_id: str | None = None
     new_id: str | None = None
 
+    @field_validator("blocks", mode="before")
+    @classmethod
+    def coerce_block_objects(cls, value: Any) -> Any:
+        return _coerce_block_texts(value)
+
 
 class RemoveSectionOp(_OpBase):
     """Remove an entire section by id."""
@@ -136,6 +156,11 @@ class ReplaceSectionBlocksOp(_OpBase):
     op: Literal["replace_section_blocks"] = "replace_section_blocks"
     section_id: str
     blocks: list[str] = Field(default_factory=list)
+
+    @field_validator("blocks", mode="before")
+    @classmethod
+    def coerce_block_objects(cls, value: Any) -> Any:
+        return _coerce_block_texts(value)
 
 
 class RenameSectionOp(_OpBase):
--- a/hindsight-api-slim/tests/test_delta_operation_parse.py
+++ b/hindsight-api-slim/tests/test_delta_operation_parse.py
@@ -5,9 +5,11 @@ from __future__ import annotations
 import pytest
 
 from hindsight_api.engine.reflect.delta_ops import (
+    AddSectionOp,
     AppendBlockOp,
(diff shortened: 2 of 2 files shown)


nicoloboschi:
fixed by #3919

PR #3918, opened 2026-08-31
fix(embed): decode control center log tail as UTF-8

koriyoshi2041:
## What changed

- decode the control-center log tail as UTF-8, matching the CLI and daemon log readers
- add a narrow-locale regression with non-ASCII log output

## Why

The UTF-8 audit merged in #3850 covered the CLI log paths, but `control_center.service.tail_log` still used the process locale. On a legacy Windows codepage this could replace valid UTF-8 output with mojibake in the control-center log view.

## Validation

- `uv run --project hindsight-embed pytest hindsight-embed/tests/test_control_center.py hindsight-embed/tests/test_env_template.py -q` — 54 passed
- Ruff clean on the changed source and test files
- `git diff --check`

### DIFF
--- a/hindsight-embed/hindsight_embed/cli.py
+++ b/hindsight-embed/hindsight_embed/cli.py
@@ -104,7 +104,7 @@ def load_config_file():
 
     # Load ONLY this profile's config, never fall back to default
     if config_path.exists():
-        with open(config_path) as f:
+        with open(config_path, encoding="utf-8") as f:
             for line in f:
                 line = line.strip()
                 if line and not line.startswith("#") and "=" in line:
--- a/hindsight-embed/hindsight_embed/cli.py
+++ b/hindsight-embed/hindsight_embed/cli.py
@@ -676,7 +676,7 @@ def do_daemon(args, config: dict, logger):
         else:
             # Show last N lines
             try:
-                with open(daemon_log_path) as f:
+                with open(daemon_log_path, encoding="utf-8", errors="replace") as f:
                     lines = f.readlines()
                     for line in lines[-args.lines :]:
                         print(line, end="")
@@ -838,7 +838,7 @@ def do_ui(args, config: dict, logger):
             return 0
         else:
             try:
-                with open(ui_log_path) as f:
+                with open(ui_log_path, encoding="utf-8", errors="replace") as f:
                     lines = f.readlines()
                     for line in lines[-args.lines :]:
                         print(line, end="")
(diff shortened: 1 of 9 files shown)


nicoloboschi (approved):
Reviewed the full diff and ran the suite locally.

Note for the record: **CI never exercised this.** `test-embed` and `test-embed-windows` are both gated on `needs.detect-changes.outputs.has_secrets == 'true'`, which is false on a fork PR — so all 86 test jobs reported `skipping` and the only things that actually ran were `build-docs`, `detect-changes`, `test-hermes-compat`, `verify-embed-control-center-bundle` and `verify-generated-files`. Verified by hand instead:

- Full `hindsight-embed` suite on the PR head: **189 passed**, 3 failed. The 3 (`test_windows_daemon_launches_via_gui_interpreter`, `test_windows_daemon_falls_back_to_console_exe_without_pythonw`, `test_posix_daemon_uses_console_entrypoint`) reproduce identically on the merge-base — pre-existing local-env sensitivity in daemon entrypoint resolution, unrelated to this change.
- The four new tests are real regression guards, not tautologies: reverting `hindsight_embed/` to base while keeping the new tests makes all four fail with `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2`. Forcing the narrow codec via `io.open`/`builtins.open` and handling pathlib's `"locale"` sentinel is the right way to pin this without a real Windows runner.
- `ruff check` and `ruff format --check` clean over `hindsight_embed` and `tests`.

On the code:

- `_serve_index` pairs correctly with `_send_html`, which already does `html.encode("utf-8")` and sets `charset=utf-8`.
- `errors="replace"` on the log readers and bare `encoding="utf-8"` on the config readers is the right split — a truncated/garbled log line should render, a corrupt config should not be silently accepted.
- The `seek`/`tell` pair in `daemon_embed_manager._start_daemon_locked` stays valid: `last_log_position` starts at 0 and is only ever assigned from `f.tell()`, and every open in the loop now uses the same codec, so the text-mode cookies remain comparable.
- Audited what the sweep left behind. Every remaining locale-default text I/O in `hindsight_embed/` is ASCII-only by construction — PID files, the UI port file, the lifecycle token, the daemon lock — plus `profile_manager._load_metadata`/`_save_metadata`, which are safe because `json.dump` defaults to `ensure_ascii=True`. Nothing user-supplied is left on a locale codec.

One nit, not blocking: the PR body still describes only the control-center log tail, but the change is the full 9-file audit across the CLI, profile manager, env template and daemon log tail. Folding that into the squash body on merge.

PR #3925, opened 2026-08-31
fix(cli): restore default SIGPIPE handling to prevent abort on broken pipe

2anoubis:
## Problem

When the hindsight CLI's output is piped into a consumer that closes the pipe early (e.g. `hindsight tag list mbn | head -5`), the process crashes with a fatal error (SIGABRT) instead of terminating cleanly.

The release profile sets `panic = "abort"` in `hindsight-cli/Cargo.toml`. Rust's `println!` panics when a write to stdout/stderr hits `EPIPE`, and with abort-panic this becomes a silent `SIGABRT` — no error message, just an abnormal termination (exit 134 = 128 + SIGABRT).

This is especially common for agent/automation workflows that pipe output through `head` or other tools that stop reading early.

## Reproduction

```bash
for i in 1 2 3 4 5; do
  hindsight memory recall mbn "shipping" --tags "domain:stocks" --tags-match any_strict --max-tokens 8000 --output json | head -3 >/dev/null
  echo "exit: ${PIPESTATUS[0]}"
done
# before: exit 134 (SIGABRT) repeatedly
# after:  exit 141 (SIGPIPE, clean)
```

## Fix

Restore the default `SIGPIPE` disposition in `main()` so the CLI terminates cleanly on a broken pipe, matching standard Unix CLI behavior:

```rust
unsafe {
    libc::signal(libc::SIGPIPE, libc::SIG_DFL);
}
```

## Testing

- Reproduced the crash 4/5 times before the fix (exit 134).
- After the fix, the same command consistently exits 141 with no abort.
- Confirmed no `ANOM_ABEND`/SIGABRT entries in the journal afterward.

### DIFF
--- a/hindsight-cli/src/main.rs
+++ b/hindsight-cli/src/main.rs
@@ -1357,6 +1357,17 @@ enum DirectiveCommands {
 }
 
 fn main() {
+    // When stdout/stderr is a pipe whose reader closes early (e.g.
+    // `hindsight ... | head`), a write triggers SIGPIPE. Rust's `println!`
+    // panics on the resulting EPIPE error, and with `panic = "abort"` in the
+    // release profile that becomes a silent SIGABRT — which surfaces as a
+    // spurious "fatal error" and kills the process without any message.
+    // Restore the default SIGPIPE disposition so we terminate cleanly on
+    // SIGPIPE, like other Unix CLI tools.
+    unsafe {
+        libc::signal(libc::SIGPIPE, libc::SIG_DFL);
+    }
+
     if let Err(e) = run() {
         ui::print_error(&format!("{:#}", e));
         std::process::exit(1);
--- a/hindsight-cli/src/main.rs
+++ b/hindsight-cli/src/main.rs
@@ -1357,13 +1357,18 @@ enum DirectiveCommands {
 }
 
 fn main() {
-    // When stdout/stderr is a pipe whose reader closes early (e.g.
-    // `hindsight ... | head`), a write triggers SIGPIPE. Rust's `println!`
-    // panics on the resulting EPIPE error, and with `panic = "abort"` in the
-    // release profile that becomes a silent SIGABRT — which surfaces as a
-    // spurious "fatal error" and kills the process without any message.
-    // Restore the default SIGPIPE disposition so we terminate cleanly on
-    // SIGPIPE, like other Unix CLI tools.
+    // Rust ignores SIGPIPE at startup so that `println!`/`print!` surface a
+    // closed stdout pipe as an `EPIPE` error rather than a signal. But the
+    // release profile sets `panic = "abort"`, so that write error becomes a
+    // silent SIGABRT — a spurious "fatal error" with no message — whenever
+    // the reader closes early (e.g. `hindsight ... | head`).
+    //
+    // Restore the default SIGPIPE disposition on Unix so the CLI terminates
+    // cleanly on a broken pipe, like other Unix CLI tools. Note this also
+    // applies to socket writes: on Linux a peer closing a connection mid-write
+    // surfaces as SIGPIPE rather than an `Err(EPIPE)`. ripgrep and fd accept
+    // the same trade-off, and it is the correct behavior here too.
+    #[cfg(unix)]
     unsafe {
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Thanks for this — the diagnosis is right and the fix is the standard Unix idiom. I verified it locally: built the CLI at `be8ea9e` and `hindsight --help | head -1` exits **141** with the patch vs **0** without (output silently dropped on the unpatched build). `libc` is already an unconditional dependency, so no `Cargo.toml` or license-manifest churn either.

One blocker before this can merge.

## 🔴 Must fix — breaks the Windows build

`libc::SIGPIPE` does not exist on Windows. `libc-0.2/src/windows/mod.rs` defines only `SIGINT`, `SIGILL`, `SIGFPE`, `SIGSEGV`, `SIGTERM`, `SIGABRT`. Cross-checking this branch:

```
$ cargo check --target x86_64-pc-windows-msvc
error[E0425]: cannot find value `SIGPIPE` in crate `libc`
    --> src/main.rs:1368:28
1368 |         libc::signal(libc::SIGPIPE, libc::SIG_DFL);
     |                            ^^^^^^^ help: a constant with a similar name exists: `SIGFPE`
```

Every other `libc` call site in the crate is already `#[cfg(unix)]`-gated (`commands/fs/mod.rs`, `commands/fs/daemon.rs`, `config.rs`), so this is a convention break as well:

```rust
#[cfg(unix)]
unsafe {
    libc::signal(libc::SIGPIPE, libc::SIG_DFL);
}
```

**Why CI is green anyway**, which is the part that worries me: the only job that compiles the CLI on Windows is `test-embed-windows` (`cargo build --release` on `windows-latest`, `.github/workflows/test.yml`), and its condition is `has_secrets && (core || embed || ci)` — it does *not* fire on `hindsight-cli/**`, and a fork PR has no secrets regardless. The Linux/arm64 CLI jobs compile fine. So this would land undetected and surface later on an unrelated PR touching `core`/`embed`, for a different author. `release.yml` ships no Windows CLI target so releases aren't affected, but Windows build-from-source is.

## 🟡 Worth considering

**The reset also affects socket writes.** Rust ignores `SIGPIPE` at startup partly so that a peer closing a TCP connection surfaces as `Err(EPIPE)` rather than a signal. On Linux `std::net`/hyper writes don't use `MSG_NOSIGNAL` (rust-lang/rust#43000), so after this reset a mid-request peer close on the API connection can kill the CLI with signal 13 and no message instead of printing an error. macOS/BSD are unaffected (`SO_NOSIGPIPE` is set at socket creation). ripgrep and fd accept this same trade-off and it's almost certainly right here too — but it's a behavior change beyond stdout, so a line in the comment noting it would help the next reader.

**No regression test.** `hindsight-cli/tests/` already has a harness that spawns `CARGO_BIN_EXE_hindsight` with no server required. A `#[cfg(unix)]` test that spawns the binary with stdout on a pipe, reads one line, drops the read end, and asserts the exit is signal 13 rather than a panic would pin this. Note `--help` alone won't reproduce — the whole help fits in the 64 KiB pipe buffer, which is why the unpatched binary exits 0 there; the test needs output larger than the buffer.

**Minor:** the description says `println!` panics on EPIPE and that's what became SIGABRT. That's accurate for the `println!` paths, but as the `--help` case shows, some clap paths swallow the write error and exit 0 — so the fix also converts a few previously-silent-truncation cases into a clean 141. That's the correct Unix behavior, just a slightly broader change than the description implies.

Happy to approve once the `#[cfg(unix)]` gate is in.

2anoubis:
Thanks for the thorough review — all points addressed:

1. **`#[cfg(unix)]` gate added** (the `libc::signal` call) so the Windows build no longer references `libc::SIGPIPE`.
2. **Comment expanded** to note the socket-write trade-off (Linux peer-close now surfaces as SIGPIPE rather than `Err(EPIPE)`, matching ripgrep/fd).
3. **Regression test added** in `hindsight-cli/tests/integration_test.rs`. To avoid the `--help` fits-in-64 KiB-pipe flakiness, it hands the child a stdout pipe whose read end is already closed, so the first write deterministically triggers SIGPIPE regardless of output size. It asserts exit via signal 13 rather than abort.

`cargo test --release test_sigpipe_terminates_with_signal_not_abort` passes.

PR #3932, opened 2026-08-31
Make consolidation pooling tags_match configurable (fixes #3924)

### LINKED ISSUE #3924: Consolidation candidate pooling uses tags_match=all_strict, excluding true merge candidates with differing classifier tags
## Summary

`_find_related_observations`'s consolidation-candidate pooling uses `tags_match="all_strict"` whenever the fact being consolidated has tags. This is an AND-match requiring a candidate observation to carry **every** tag of the fact being consolidated. In practice this means two facts about the same entity, submitted in the same batch, can never be considered for merging if they differ on even one tag dimension — which is normal and expected when tags are assigned per-fact (e.g. by a category/topic classifier upstream of Hindsight), not per-entity.

The result: consolidation degenerates toward one observation per fact instead of synthesizing related facts into fewer, richer observations — which is the whole point of the consolidation step.

## Reproduction

Self-hosted deployment (Hindsight 0.9.2, `vertexai`/`gemini-2.5-pro` LLM provider). A personal-notes bank had two facts submitted in the same push batch (identical `mentioned_at`), both about the same entity ("Gardea", an internal service name — not important beyond being the shared entity):

- Fact A → observation 1: describes Gardea's technical deployment (build tag, entry point, GKE cluster, URL). Tags: `category:reference, suggested_tool:gardea, suggested_context:infra-documentation, confidence:high`.
- Fact B → observation 2: describes Gardea's functional purpose (what it enforces and why). Tags: `category:fact, suggested_tool:terraform-cloud, suggested_context:infra-documentation, confidence:high`.

Only `confidence:high` and `suggested_context:infra-documentation` are shared; `category` and `suggested_tool` differ, because the upstream classifier correctly assigned different values — one fact really is a `reference`/`gardea`-tool fact, the other really is a `fact`/`terraform-cloud`-tool fact. Both facts are entirely legitimate, non-contradictory, and clearly complementary descriptions of one entity.

Consolidation never merged them. Querying `POST /v1/default/banks/{bank}/memories/recall` directly confirms why:

```jsonc
// tags_match="all_strict", tags = observation 2's own tag set, types=["observation"], query="Gardea"
// -> returns ONLY observation 2 itself. Observation 1 never surfaces as a candidate.

// tags_match="any_strict", same tags, same query
// -> returns BOTH observations, ranked #1 and #2 by semantic score.
//    ~30 unrelated observations also pass the loose OR filter but rank well below both real hits.
```

This reproduces `_find_related_observations`'s exact pooling call: `recall_async(fact_type=["observation"], budget=LOW, reranking="interleave", tags_match="all_strict" if tags else "any")`.

## Why `any_strict` isn't obviously worse

The candidate pool from `any_strict` is wider (more noise), but semantic ranking on the fact's own text still puts true matches at the top. The consolidation LLM call already receives a bounded, ranked candidate list and makes its own merge/no-merge judgment per the `observations_mission` prompt — it doesn't need the pooling step to pre-filter down to only tag-identical candidates. `all_strict` is filtering out true positives before the LLM ever gets a chance to judge them, which is a worse failure mode than occasionally showing the LLM an irrelevant candidate it correctly declines to merge.

## Suggested fix

Loosen `_find_related_observations`'s pooling tag match from `all_strict` to `any_strict` (or drop tag-based filtering from pooling entirely and rely on semantic/entity recall alone), and let the consolidation LLM's own judgment decide whether to merge. Happy to open a PR if a maintainer confirms this is the right direction — the change is a one-line `tags_match` value swap in `consolidator.py`.

## Impact

Sampling 100 observation records in the same deployment found 99 with `proof_count=1` / exactly one `source_memory_id` — i.e. consolidation is running at near-1:1 with raw facts, not merging at any meaningful rate. This is consistent with `all_strict` pooling being the (or a major) cause: any two facts about the same topic that pick up even slightly different classifier tags are structurally unable to find each other during consolidation.

## Environment

- Hindsight: 0.9.2
- LLM provider: `vertexai` (gemini-2.5-pro)
- Deployment: self-hosted on GKE, Postgres-backed

ferrastas:
## Summary

Fixes #3924.

`_find_related_observations` hardcoded `tags_match="all_strict"` when pooling existing observations as consolidation merge candidates. That excludes true merge candidates that differ only on purely-descriptive classifier tags (e.g. `category:*`/`suggested_tool:*`), producing near-1:1 consolidation results for banks whose tags aren't a tenant/user security boundary.

While investigating, I confirmed `all_strict` is also a genuine security boundary for deployments that *do* use tags to scope consolidation per tenant/user — a blanket switch to `any_strict` would be a cross-tenant information-leakage regression for those deployments. So this isn't a straight bug fix; it's a new opt-in knob.

## What changed

- New bank-level config field `consolidation_pooling_tags_match` (`all_strict` default / `any_strict` opt-in). Default is unchanged behavior — nobody's setup changes unless they explicitly opt in.
- `config.py`: env var, default, validator (warn-and-default on bad values, matching the existing pattern for `retain_extraction_mode`), dataclass field, `_CONFIGURABLE_FIELDS` entry.
- `consolidator.py`: `_find_related_observations` now reads `config.consolidation_pooling_tags_match` instead of the hardcode. The separate obs-vs-obs dedup-probe function's own `all_strict` hardcode is untouched — that one is out of scope.
- `http.py`: `BankTemplateConfig` field + `validate_bank_template` check, same shape as `retain_extraction_mode`.
- Control-plane admin UI: new dropdown in the Security & Privacy section (`all_strict` default, `any_strict` as the other option), wired to `updateBankConfig`.
- Docs (`configuration.md`), `.env.example`, and the `hindsight-embed` bundled copy of `.env.example` updated.
- New deterministic unit test (`test_consolidation_pooling_tags_match.py`) covering default/validation/fallback, `_CONFIGURABLE_FIELDS` registration, `ConfigResolver` bank-override round-trip (including null-reverts-to-default), and bank-template manifest validation.

## Test plan

- [x] `uv run pytest tests/test_consolidation_pooling_tags_match.py -v` — 7/7 passed
- [x] `./scripts/hooks/lint.sh` — all lints passed
- [x] Reproduced the original bug against a live bank: `any_strict` recall returns both true-merge-candidate observations ranked top-2; `all_strict` (current default) returns only the querying fact's own observation


nicoloboschi:
https://github.com/vectorize-io/hindsight/issues/3924#issuecomment-5479634140

PR #3939, opened 2026-08-31
fix(server): load .env before importing MemoryEngine to honor concurrency limits

Sanderhoff-alt:
## Overview

When running Hindsight with multi-worker mode (`hindsight-api --workers N`), dev reload (`hindsight-api --reload`), or directly targeting the ASGI app via Uvicorn (`uvicorn hindsight_api.server:app`), `server.py` imported `MemoryEngine` at the top level before calling `load_dotenv_for_entrypoint()`.

Because `MemoryEngine` transitively imports `hindsight_api.engine.llm_wrapper`, and `llm_wrapper` initializes `_global_llm_semaphore` and `_per_op_llm_semaphores` at module import time, the semaphores were constructed before `.env` was parsed. As a result, concurrency settings defined in `.env` (such as `HINDSIGHT_API_LLM_MAX_CONCURRENT`, `HINDSIGHT_API_RETAIN_LLM_MAX_CONCURRENT`, etc.) were silently ignored and fell back to the defaults (32).

## Execution Flow

```
[BEFORE FIX]
uvicorn worker bootstrap
  │
  ├── 1. import hindsight_api.server
  │     ├── line 18: from hindsight_api import MemoryEngine
  │     │     └── triggers import of hindsight_api.engine.llm_wrapper
  │     │           ├── reads os.getenv("HINDSIGHT_API_LLM_MAX_CONCURRENT") -> None (not in process env yet)
  │     │           └── _global_llm_semaphore = Semaphore(DEFAULT=32) ❌
  │     │
  │     └── line 33: load_dotenv_for_entrypoint()
  │           └── parses .env and populates os.environ (Too late! Semaphore already built)
  │
  └── Result: .env concurrency limits ignored, requests run at default capacity

---------------------------------------------------------------------------------------------

[AFTER FIX]
uvicorn worker bootstrap
  │
  ├── 1. import hindsight_api.server
  │     ├── line 24: from hindsight_api.config import get_config, load_dotenv_for_entrypoint
  │     ├── line 26: load_dotenv_for_entrypoint()
  │     │     └── parses .env and populates os.environ FIRST ✅
  │     │
  │     └── line 31: from hindsight_api import MemoryEngine
  │           └── triggers import of hindsight_api.engine.llm_wrapper
  │                 ├── reads os.getenv("HINDSIGHT_API_LLM_MAX_CONCURRENT") -> configured value ✅
  │                 └── _global_llm_semaphore = Semaphore(configured_value) ✅
  │
  └── Result: .env concurrency limits properly honored across all workers
```

## Impact Matrix

| Mode / Entrypoint | Invocation | Pre-fix Behavior | Post-fix Behavior |
| :--- | :--- | :--- | :--- |
| **Multi-worker CLI** | `hindsight-api --workers N` | `.env` concurrency ignored (default 32) | `.env` concurrency honored |
| **Dev Reload CLI** | `hindsight-api --reload` | `.env` concurrency ignored (default 32) | `.env` concurrency honored |
| **Direct ASGI Server** | `uvicorn hindsight_api.server:app` | `.env` concurrency ignored (default 32) | `.env` concurrency honored |
| **Single-worker CLI** | `hindsight-api` | `.env` concurrency honored (`main.py` flow) | `.env` concurrency honored (unchanged) |

## Changes

1. **`hindsight_api/server.py`**:
   - Reordered imports so that `load_dotenv_for_entrypoint()` executes before importing `MemoryEngine` and `create_app`.
2. **`tests/test_dotenv_loading.py`**:
   - Added `test_server_entrypoint_loads_dotenv_before_engine_semaphores` to verify that importing `hindsight_api.server` in a child process with a `.env` file correctly initializes global and per-op LLM semaphores.

## Verification

- `cd hindsight-api-slim && uv run pytest tests/test_dotenv_loading.py tests/test_server_module.py -v` (9 passed)
- `./scripts/hooks/lint.sh` (All lints passed)
- `cd hindsight-api-slim && uv run ty check hindsight_api/` (All checks passed)

### DIFF
--- a/hindsight-api-slim/hindsight_api/server.py
+++ b/hindsight-api-slim/hindsight_api/server.py
@@ -15,25 +15,28 @@ import warnings
 warnings.filterwarnings("ignore", message="websockets.legacy is deprecated")
 warnings.filterwarnings("ignore", message="websockets.server.WebSocketServerProtocol is deprecated")
 
+# This module IS an entry point: it is the ASGI app targeted by
+# `uvicorn hindsight_api.server:app`, and the import string uvicorn re-imports
+# in each worker process when `hindsight-api` runs with --workers/--reload. Load
+# .env here (like the CLI entry points) so those worker processes see the same
+# configuration. Importing hindsight_api as a library never reaches this module.
+# Note: Must be called BEFORE importing MemoryEngine or create_app, because
+# engine submodules (e.g. llm_wrapper) instantiate semaphores at import time.
+from hindsight_api.config import get_config, load_dotenv_for_entrypoint
+
+load_dotenv_for_entrypoint()
+
+# Disable tokenizers parallelism to avoid warnings
+os.environ["TOKENIZERS_PARALLELISM"] = "false"
+
 from hindsight_api import MemoryEngine
 from hindsight_api.api import create_app
-from hindsight_api.config import get_config, load_dotenv_for_entrypoint
 from hindsight_api.extensions import (
     DefaultExtensionContext,
     OperationValidatorExtension,
     TenantExtension,
     load_extension,
 )
-
-# This module IS an entry point: it is the ASGI app targeted by
-# `uvicorn hindsight_api.server:app`, and the import string uvicorn re-imports
-# in each worker process when `hindsight-api` runs with --workers/--reload. Load
-# .env here (like the CLI entry points) so those worker processes see the same
-# configuration. Importing hindsight_api as a library never reaches this module.
-load_dotenv_for_entrypoint()
-
-# Disable tokenizers parallelism to avoid warnings
-os.environ["TOKENIZERS_PARALLELISM"] = "false"
 
 # Load configuration and configure logging
 config = get_config()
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
I ran this in a clean container (`python:3.12-slim`, `pip install ./hindsight-api-slim`) and reverted only the `server.py` hunk between the two runs: 4 passed at `94e93d33`, and `1 failed, 3 passed` with the hunk reverted, the failure being the new test. So the test is load bearing, and with the old order `import hindsight_api.server` leaves `_llm_max_concurrent` at 32 and `_per_op_llm_semaphores` empty even though `.env` has set the var to 7 by then.

One thing next door, if you want it in scope. `hindsight-admin` has the same ordering: `admin/cli.py:23` imports `..engine.memory_engine` at module scope, while `load_dotenv_for_entrypoint()` is the first line of `main()` at :1175, so it runs after the semaphores are built. Same tree, same container:

```
admin: at module import       env=None _llm_max_concurrent=32 per_op_keys=[]
admin: after main()'s dotenv  env=7    _llm_max_concurrent=32 per_op_keys=[]
```

It looks latent rather than broken: none of the nine admin commands issue LLM calls, so nothing there reads those semaphores. I also checked `main.py`, `worker/main.py` and `mcp_local.py`, and none of them reach `llm_wrapper` at module scope, so the server was the one that actually mattered.

Sanderhoff-alt:
Thanks for the thorough review and for running the container verification! It's great to confirm that the test is load-bearing.

Regarding `admin/cli.py`: since it's currently latent and doesn't issue any LLM calls, I'd prefer to keep this PR focused specifically on the `server.py` ASGI fix. I'll open a separate follow-up PR to clean up the import order in `admin/cli.py` for code hygiene.

PR #3942, opened 2026-08-31
fix(retain): force re-extraction on document reprocess bypassing crash-recovery and delta deduplication

### LINKED ISSUE #3899: reprocess is silently classified as crash recovery when the replayed content hash matches: zero LLM calls, no re-extraction, op settles completed
## Summary

`POST /v1/default/banks/{bank_id}/documents/{document_id}/reprocess` — and any re-retain of unchanged content under the same `document_id` — can be a **silent no-op**: the retain pipeline's crash-recovery gate sees the matching `content_hash` plus the document's existing chunk hashes, flips into recovery mode, preserves every existing unit, and makes **zero LLM calls**. The operation settles `completed`. Nothing distinguishes it from a real re-extraction except reading the resulting facts.

The endpoint is documented as *"re-extracts facts using the current engine configuration. Useful when the LLM model, chunking strategy, or extraction settings have changed"* — precisely the case where the content is unchanged by definition, and precisely the case the recovery gate suppresses.

## Where

`hindsight_api/engine/retain/orchestrator.py` (0.9.2 / current main):

```python
# ~line 2290
    existing_rows = await chunk_storage.load_existing_chunks(conn, bank_id, effective_doc_id)
    existing_chunk_hashes = {c.content_hash for c in existing_rows if c.content_hash}
    if existing_chunk_hashes:
        is_recovery = True
        # "RECOVERY: found N already-committed chunks — will skip matching and preserve existing data"
```

`memory_engine.reprocess_document` (~line 11073) rebuilds the content dict from `documents.retain_params` + `original_text` and calls `submit_async_retain` — it never clears the document's `content_hash` or chunk rows, and passes no flag that would bypass the gate. When the rebuilt content hashes identically to what is stored, the retain is classified as crash recovery.

## Observed

Self-hosted bank, ~12k documents, episodes pushed by an external sync layer keyed on `document_id`:

- **2026-08-02 (0.8.6):** after a bank-config change, 41 episodes were re-armed and re-pushed with unchanged content to force re-extraction. All 41 `batch_retain` ops settled `completed` in **0.13–1.63 s each** with nothing re-extracted (units and tags byte-identical). `POST /documents/{id}/reprocess` on the same documents was the same no-op: `unit_ids_count: 0`.
- The only working procedure is **delete the document, then re-retain** — which loses the units if the client crashes between the two steps, and is what we've now had to productize client-side.

## Interaction with #3873 / #3874 — the behaviour is nondeterministic

#3873 shows the opposite outcome on 0.9.2: a reprocess that **did** re-extract (0 units → 1, under the wrong strategy). Both observations are consistent with the same gate: recovery only triggers when the **rebuilt** content dict hashes identically to the stored `content_hash` *and* chunk hashes survive. Because `reprocess_document` rebuilds its retain call from a `retain_params` record that drifts from the original submission (the round-trip gaps #3873 documents), whether a given reprocess no-ops or re-extracts depends on how the stored params happen to replay.

So today the endpoint has three outcomes and the caller controls none of them:

| stored params replay | result |
|---|---|
| hash matches, chunks survive | silent no-op (this issue) |
| hash differs | re-extraction — under the wrong strategy until #3874 |
| hash matches, chunks missing | full re-extraction |

Note that #3874 as written makes the replay *more* faithful — which makes the hash match *more* often, i.e. it converts #3873's wrong-strategy re-extractions into this issue's silent no-ops. The two fixes need to compose.

## Expected

An explicit reprocess is an operator saying "extract this again under the current config." It should never be classified as crash recovery.

## Proposed fix

In `reprocess_document`, before resubmitting: clear the document's `content_hash` (and/or chunk `content_hash` rows) under the document row lock — or thread a `force_reextract` flag through `submit_async_retain` into the orchestrator that suppresses the `is_recovery` classification for that operation. Either way:

- recovery detection keeps protecting genuine crash-recovery retains (unchanged semantics for the sync layer's content-hash-gated pushes);
- `reprocess` deterministically re-extracts, under the item's own strategy once #3874 lands;
- the response could carry `reextracted: true/false` so a no-op is at least visible if any path can still produce one.

Happy to send a PR if the `force` approach (vs. clearing the hash) is the preferred shape.

Version: observed on 0.8.6 and 0.9.2, self-hosted, embedded Postgres; gate verified present on current main.

Sanderhoff-alt:
## Summary

Fixes #3899

When calling `POST /v1/default/banks/{bank_id}/documents/{document_id}/reprocess`, `reprocess_document` fetches the stored `original_text` and `retain_params` and resubmits them as a retain operation under the same `document_id` with `update_mode="replace"`.

Because the submitted text is identical to the currently stored document, the retain processing pipeline treated reprocessing as a **silent no-op** through two separate deduplication / recovery mechanisms:

1. **Delta Retain Early Exit (`_try_delta_retain`)**:
   Delta retain hashes incoming chunks and compares them against existing chunk hashes in Postgres. Because the text is identical, every chunk was categorized as unchanged, causing `_try_delta_retain` to take the `_delta_metadata_only` shortcut and return early without executing any LLM extraction calls.
2. **Crash-Recovery Chunk Skipping (`_streaming_retain_batch`)**:
   Even if delta retain was bypassed, `_streaming_retain_batch` checked whether `doc_row["content_hash"] == new_content_hash`. Because the hash matched and existing chunk rows were found, it flagged `is_recovery = True`. In `_llm_producer`, all chunks matching `existing_chunk_hashes` were skipped (0 LLM extractions). In the database consumer, `is_recovery = True` routed to `upsert_document_metadata` instead of `handle_document_tracking`, preserving the existing memory units without deletion or renewal.

This PR introduces an internal `force_reextract` control flag on `RetainContentDict` and throughout the retain orchestrator pipeline to ensure document reprocessing always bypasses delta optimization and crash-recovery deduplication, forcing full LLM re-extraction and clean replacement of existing memory units.

---

## Retain Pipeline & Control Flow

```mermaid
flowchart TD
    subgraph BEFORE["❌ BEFORE: Reprocess Silent No-Op Traps"]
        direction TB
        R1["POST /documents/{id}/reprocess<br/>(replays original_text, update_mode='replace')"]
        R1 --> D1{"_try_delta_retain<br/>Compare chunk hashes"}
        D1 -->|"All chunks match (unchanged)"| S1["_delta_metadata_only<br/>• 0 LLM extraction calls<br/>• Preserves old facts<br/>• Early return"]
        
        D1 -.->|"If Delta bypassed"| STR1{"_streaming_retain_batch<br/>Check content_hash match"}
        STR1 -->|"content_hash matches & chunks exist"| REC1["is_recovery = True"]
        REC1 --> PROD1["_llm_producer: Skips all chunks<br/>(0 LLM extractions)"]
        REC1 --> DB1["DB Consumer: upsert_document_metadata<br/>• No memory units deleted<br/>• No new memory units created"]
    end

    subgraph AFTER["✅ AFTER: Dedicated force_reextract Bypass Path"]
        direction TB
        R2["POST /documents/{id}/reprocess<br/>(sets internal force_reextract=True)"]
        R2 --> D2{"_try_delta_retain<br/>force_reextract is True"}
        D2 -->|"Bypass Delta Retain"| STR2["_streaming_retain_batch<br/>(force_reextract=True)"]
        STR2 --> REC2["is_recovery = False<br/>(Crash recovery disabled)"]
        REC2 --> PROD2["_llm_producer:<br/>Extracts all chunks via LLM"]
        REC2 --> DB2["DB Consumer: handle_document_tracking<br/>• Cascade-deletes old memory units<br/>• Inserts newly extracted facts"]
    end
```

---

## Execution Matrix Across Retain Scenarios

| Scenario | Trigger / Endpoint | Content Hash vs Stored | Delta Retain Action | LLM Producer Action | Database Consumer Action |
|---|---|---|---|---|---|
| **Standard Retain (New Doc)** | `POST /memories` | Distinct / New | Bypassed (new doc) | Extracts all chunks | `handle_document_tracking` inserts new doc & units |
| **Incremental Upsert** | `POST /memories` | Modified | Diff applied (only changed/new chunks extracted) | Extracts diff chunks | Replaces changed chunk facts, preserves unchanged units |
| **Identical Resubmit (No Change)** | `POST /memories` | Identical | `_delta_metadata_only` shortcut (0 LLM calls) | Skipped | `upsert_document_metadata` updates metadata only |
| **Worker Crash Recovery** | Async Operation Retry | Identical | Bypassed (first batch / retry) | Skips committed chunks; extracts uncommitted chunks | Re-anchors doc & commits uncommitted facts |
| **Explicit Document Reprocess** | `POST /documents/{id}/reprocess` | Identical | **Bypassed (`force_reextract=True`)** | **Extracts all chunks via LLM (`force_reextract=True`)** | **`handle_document_tracking` cascade-deletes old units and inserts fresh facts** |

---

## Key Technical Decisions

1. **Internal Engine-Level Scope for `force_reextract`**:
   - `force_reextract` is kept strictly internal on `RetainContentDict` rather than exposing it on the public HTTP `MemoryItem` schema. External callers already have the dedicated `POST /documents/{id}/reprocess` endpoint, avoiding schema pollution and preventing coupling external APIs to internal deduplication flags.
   - `reprocess_document` explicitly sets `content_dict["force_reextract"] = True`.

2. **Exclusion from Document `retain_params` Serialization**:
   - Added `"force_reextract"` to `_RETAIN_PARAMS_NOT_REPLAYED` in `orchestrator.py`.
   - Ensures `force_reextract` is treated as a transient operation flag and is not written to the JSON `retain_params` column in Postgres, avoiding accidental persistence to subsequent operations.

3. **Multi-layer Pipeline Bypass**:
   - **`retain_batch` / `_try_delta_retain`**: When `force_reextract` is True, `_try_delta_retain` immediately returns `None`, allowing execution to proceed directly to the full streaming extraction pipeline.
   - **`_streaming_retain_batch`**: Guards crash recovery detection with `if not force_reextract:`, keeping `is_recovery = False`.
   - **`_llm_producer`**: Guards chunk deduplication with `if not force_reextract and chunk_hash in existing_chunk_hashes:`, ensuring every chunk is queued for LLM extraction.
   - **DB Consumer**: Because `is_recovery` is `False`, the first batch transaction invokes `handle_document_tracking`, which cascade-deletes prior memory units and writes the freshly extracted units.

---

## Changes by Component

- **`hindsight_api/engine/retain/types.py`**:
  - Add `force_reextract: bool` to `RetainContentDict`.
- **`hindsight_api/engine/memory_engine.py`**:
  - In `reprocess_document`, set `content_dict["force_reextract"] = True`.
- **`hindsight_api/engine/retain/orchestrator.py`**:
  - Add `"force_reextract"` to `_RETAIN_PARAMS_NOT_REPLAYED`.
  - Pass and evaluate `force_reextract` in `retain_batch`, `_try_delta_retain`, and `_streaming_retain_batch`.
  - Disable crash recovery and chunk skipping when `force_reextract` is True.
- **`tests/test_retain_params_roundtrip.py`**:
  - Verify `force_reextract` is properly excluded from stored `retain_params` and handled on replay.
- **`tests/test_reprocess_force_reextract.py`**:
  - Add integration tests verifying reprocess deletes and replaces memory units on identical content using the `list_memory_units` engine API.
  - Verify HTTP `POST /documents/{id}/reprocess` endpoint replaces facts end-to-end.

---

## Verification

Full suite of retain, delta retain, roundtrip, and reprocess tests passed:
```bash
uv run pytest tests/test_delta_retain.py \
              tests/test_retain_params_roundtrip.py \
              tests/test_document_chunks_and_reprocess.py \
              tests/test_reprocess_force_reextract.py
============================== 55 passed in 17.58s =============================
```

Linters and type checks:
- `ruff check .` -> Passed
- `ty check hindsight_api/` -> Passed
- `./scripts/hooks/lint.sh` -> All linters passed

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -11421,6 +11421,7 @@ class MemoryEngine(MemoryEngineInterface):
             "content": original_text,
             "document_id": document_id,
             "update_mode": "replace",
+            "force_reextract": True,
         }
         # Replay everything the original retain supplied. Enumerating fields here
         # is what let `strategy`, `entities` and `resolve_entities` go missing one
@@ -11434,10 +11435,11 @@ class MemoryEngine(MemoryEngineInterface):
         # on the item, so retain_params came back without it and the NEXT reprocess
         # fell back to the bank default again.
         content_dict.update(retain_params)
-        # These three are the reprocess's own and must win over anything stored.
+        # These four are the reprocess's own and must win over anything stored.
         content_dict["content"] = original_text
         content_dict["document_id"] = document_id
         content_dict["update_mode"] = "replace"
+        content_dict["force_reextract"] = True
 
         tags = doc.get("tags") or []
         if tags:
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -383,7 +383,7 @@ def _resolve_narrator(profile_name: str, bank_id: str) -> str | None:
 # and only the resulting facts are wrong. Inverting it makes the safe case the
 # default — a new retain field round-trips unless someone deliberately excludes it,
 # and the single source of truth becomes what api_retain puts on the content dict.
-_RETAIN_PARAMS_NOT_REPLAYED = frozenset({"content", "document_id", "update_mode", "tags"})
+_RETAIN_PARAMS_NOT_REPLAYED = frozenset({"content", "document_id", "update_mode", "tags", "force_reextract"})
 
 
 def _build_retain_params(contents_dicts, document_tags=None, doc_contents=None):
@@ -1708,10 +1708,16 @@ async def retain_batch(
     # entities against Postgres and takes its document lock on a `documents` row that store-owned
     # banks no longer have: `current_hash` comes back None, the stale-chunk guard never fires, and
     # delta ran with no concurrency control at all.
+    #
+    # Deduplication and crash recovery are document-scoped (delta diffing and recovery checks gate
+    # on `effective_doc_id` and the document's `content_hash`). Aggregating `force_reextract` across
+    # the batch is intentional: if any item requests full extraction, the entire document bypasses
+    # delta/recovery shortcuts and re-extracts every chunk wholesale.
+    force_reextract = any(bool(item.get("force_reextract")) for item in contents_dicts)
     from ..memories import get_memories as _get_memories_delta
 
     _delta_provider = _get_memories_delta()
-    if attempts_delta_retain(_delta_provider, bank_id, is_first_batch):
+    if not force_reextract and attempts_delta_retain(_delta_provider, bank_id, is_first_batch):
         delta_result = await _try_delta_retain(
             pool,
             embeddings_model,
@@ -1736,6 +1742,7 @@ async def retain_batch(
             document_body_override=document_body_override,
             delta_full_body=_delta_full_body,
             append_base_hash=append_base_hash,
+            force_reextract=force_reextract,
         )
         if delta_result is not None:
             return delta_result
@@ -1815,6 +1822,7 @@ async def retain_batch(
         progress_callback=progress_callback,
         append_base_hash=append_base_hash,
         append_base_watermark=append_base_watermark,
+        force_reextract=force_reextract,
     )
 
 
@@ -2233,6 +2241,7 @@ async def _streaming_retain_batch(
     progress_callback: "Callable[..., Awaitable[None]] | None" = None,
     append_base_hash: str | None = None,
     append_base_watermark: int | None = None,
+    force_reextract: bool = False,
 ) -> tuple[list[list[str]], TokenUsage]:
     """
     Process a large document in streaming mini-batches to bound memory usage.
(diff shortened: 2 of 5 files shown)


nicoloboschi:
fixed by #3949

PR #3944, opened 2026-08-31
fix(structured-output): strip $ref sibling keywords from strict schemas

### LINKED ISSUE #3904: Retain fails with invalid strict schema when entity_labels are enabled
### Bug Description

Hindsight retain fails when the configured bank uses entity labels and the OpenAI-compatible provider enforces strict structured outputs. Existing recall works, but every new retain fails before model inference.

### Steps to Reproduce

1. Run Hindsight API 0.9.2 with `HINDSIGHT_API_LLM_PROVIDER=openrouter`, `HINDSIGHT_API_LLM_MODEL=openai/gpt-4.1-mini`, and `HINDSIGHT_API_LLM_STRICT_SCHEMA=true`.
2. Configure a bank with an entity-label group, such as `knowledge` with type `multi-values`.
3. POST a synchronous retain request to `/v1/default/banks/{bank_id}/memories` with one ordinary text item.
4. Observe the provider request failure.

### Expected Behavior

Hindsight should emit an OpenAI-compatible strict JSON Schema and retain the item successfully, preserving the structured labels field.

### Actual Behavior

The request fails with HTTP 400 from the downstream provider:

`Invalid schema for response_format 'response': context=('properties', 'labels'), $ref cannot have keywords {'description'}`

The generated schema contains a property shaped like:

`{"$ref":"#/$defs/Labels","description":"Classification labels for this fact..."}`

The direct API returns HTTP 500 with `Fact extraction failed`. In testing, 99 retain operations failed with this same error, while recall remained functional.

The problematic field is created in `hindsight_api/engine/retain/fact_extraction.py` when the dynamic labels model is added with `Field(description=...)`. `strict_json_schema()` in `hindsight_api/engine/structured_output.py` does not remove sibling keywords from `$ref` nodes.

### Version

Hindsight API/container: 0.9.2 (`ghcr.io/vectorize-io/hindsight:latest` ARM64 digest `sha256:b94369e48294ce2277557553226a7713e2b06791647332090c21169c2ff11fb3`).

Relevant introducing commit: `9b96becc5c56f4c2144831504d35b9451d4699d0` (entity labels, 2026-03-02).

### LLM Provider

OpenRouter routing to OpenAI, model `openai/gpt-4.1-mini`, with strict schema enabled.

### Suggested Fix

Either remove the outer description from the nested labels field, or normalize strict schemas recursively so a node containing `$ref` has no sibling keywords. A regression test should generate the dynamic labels response schema through `strict_json_schema()` and assert that every `$ref` node contains only `$ref`.

The underlying bank data and label descriptions should remain unchanged.

nicoloboschi:
Fixes #3904.

## The bug

With entity labels configured on a bank and `HINDSIGHT_API_LLM_STRICT_SCHEMA=true`, every retain failed before inference:

```
Invalid schema for response_format 'response': context=('properties', 'labels'), $ref cannot have keywords {'description'}
```

`_build_extraction_prompt_and_schema` builds a dynamic `LabelsFact` whose `labels` field both points at the nested `Labels` model and carries `Field(description=...)`. Pydantic serializes that as `{"$ref": "#/$defs/Labels", "description": "..."}`. JSON Schema 2020-12 permits sibling keywords beside a `$ref`; OpenAI's strict subset does not — the provider 400s and the API surfaces HTTP 500 `Fact extraction failed`. Recall is unaffected because it never emits this schema.

Reproduced on `main` with no provider needed, by putting the labels model through `strict_json_schema()`.

## The fix

Normalize in `strict_json_schema()` rather than at the call site: a node carrying a `$ref` now keeps only `$ref`.

Fixing it at that chokepoint covers all five providers that serialize through it (`openai_compatible`, `openai_responses`, `litellm`, `codex`, `xai_oauth`) and any future nested model with a described field — dropping the single `labels` description would fix this field and leave the next one to hit the same 400. The traversal special-cases `$defs` / `properties` / `patternProperties`, whose keys are names rather than keywords, so a property literally called `$ref` is not mistaken for a reference node.

Dropping the description costs nothing at inference time: `_build_labels_prompt_section` already describes the labels object in the prompt, and the non-strict path still sends the full unmodified schema. Nullable nested models are unaffected — their description sits on the `anyOf` wrapper, not on the `$ref`.

## Tests

`tests/test_strict_json_schema.py` gains two regressions, both failing on the current `main`:

- a described `$ref` field serializes to `{"$ref": ...}` alone;
- a real entity-label config run through `_build_extraction_prompt_and_schema` → `strict_json_schema()` has no `$ref` node with siblings — the check the issue asks for.

`./scripts/hooks/lint.sh` is clean.

https://claude.ai/code/session_01J52s5Lbg5Lp7KwRyhqg8f3

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/structured_output.py
+++ b/hindsight-api-slim/hindsight_api/engine/structured_output.py
@@ -25,9 +25,42 @@ class OpenAIStrictSchemaGenerator(GenerateJsonSchema):
         return json_schema
 
 
+# Keywords whose value is a map of *names* to subschemas, not a subschema itself.
+# Their keys are user-controlled, so a key literally called "$ref" must not be
+# mistaken for a reference node while walking the tree.
+_SUBSCHEMA_NAME_MAPS = frozenset({"$defs", "definitions", "properties", "patternProperties"})
+
+
+def _strip_ref_siblings(node: Any) -> Any:
+    """Drop every sibling keyword of a ``$ref``.
+
+    Pydantic emits ``{"$ref": ..., "description": ...}`` for a field that both
+    references a nested model and carries its own ``Field(description=...)``
+    (JSON Schema 2020-12 allows that; the dynamic ``labels`` field built for
+    entity-label banks hits it). OpenAI-compatible strict mode rejects it with
+    ``$ref cannot have keywords {'description'}``, so a reference node keeps
+    nothing but the reference.
+    """
+    if isinstance(node, dict):
+        if "$ref" in node:
+            return {"$ref": node["$ref"]}
+        return {
+            key: (
+                {name: _strip_ref_siblings(sub) for name, sub in value.items()}
+                if key in _SUBSCHEMA_NAME_MAPS and isinstance(value, dict)
+                else _strip_ref_siblings(value)
+            )
+            for key, value in node.items()
+        }
+    if isinstance(node, list):
+        return [_strip_ref_siblings(item) for item in node]
+    return node
+
+
 def strict_json_schema(response_format: type[BaseModel]) -> dict[str, Any]:
     """Serialize a typed response model directly into OpenAI's strict subset."""
-    return response_format.model_json_schema(schema_generator=OpenAIStrictSchemaGenerator)
+    schema = response_format.model_json_schema(schema_generator=OpenAIStrictSchemaGenerator)
+    return _strip_ref_siblings(schema)
 
 
 # Types the structured-output extractor knows how to build a Pydantic field for
--- a/hindsight-api-slim/tests/test_strict_json_schema.py
+++ b/hindsight-api-slim/tests/test_strict_json_schema.py
@@ -1,6 +1,11 @@
+from unittest.mock import MagicMock
+
 from pydantic import BaseModel, Field
 
-from hindsight_api.engine.retain.fact_extraction import FactExtractionResponse
+from hindsight_api.engine.retain.fact_extraction import (
+    FactExtractionResponse,
+    _build_extraction_prompt_and_schema,
+)
 from hindsight_api.engine.structured_output import strict_json_schema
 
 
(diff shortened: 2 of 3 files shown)


hindsight-api-slim/hindsight_api/engine/structured_output.py
```diff
@@ -25,9 +25,42 @@ def default_schema(self, schema: core_schema.WithDefaultSchema) -> JsonSchemaVal
...
+    Pydantic emits ``{"$ref": ..., "description": ...}`` for a field that both
+    references a nested model and carries its own ``Field(description=...)``
+    (JSON Schema 2020-12 allows that; the dynamic ``labels`` field built for
+    entity-label banks hits it). OpenAI-compatible strict mode rejects it with
+    ``$ref cannot have keywords {'description'}``, so a reference node keeps
+    nothing but the reference.
+    """
+    if isinstance(node, dict):
+        if "$ref" in node:
+            return {"$ref": node["$ref"]}
```
ebarkhordar, line 46: I ran `_strip_ref_siblings` from `c34afd3` in a clean container against pydantic. The chokepoint argument checks out: the other `model_json_schema()` sites are the non-strict prompt branch (`openai_responses_llm.py:480`), Ollama's native format (`openai_compatible_llm.py:1708`) and the providers that never use the strict subset, so nothing reaching strict mode bypasses this.

One shape loses more than its siblings. When the root is itself a reference, pydantic puts `$defs` beside it, and this early return takes them with it:

```
class Node(BaseModel):
    name: str
    child: Optional["Node"] = None

root keys before: ['$defs', '$ref']
after            : {"$ref": "#/$defs/Node"}
```

The reference is then dangling, so the schema is unresolvable rather than merely undescribed.

I could not reach that from the API: it needs the top-level response model to be recursive, and none of the models passed to `strict_json_schema()` are. The closest is `MapField` in `retain/entity_labels.py` (`dict[str, "MapField"]`), and it sits nested rather than at the root.

Might still be worth a guard here, though not by keeping `$defs`, since that is the sibling rule again. Resolving a root-level `$ref` into its `$defs` entry, or only stripping below the root, would both avoid it. I have not checked which one the provider prefers.

PR #3954, opened 2026-08-31
fix(consolidation): resolve observation_scopes before tag-set batching

### LINKED ISSUE #3953: observation_scopes: "shared" does not merge memories with different native tags (contradicts docs) — follow-up to #3924
Follow-up to #3924. That issue was closed pointing at `observation_scopes` (specifically `"shared"`) as the existing supported way to merge memories carrying different classifier tags into one observation. I tested that live against `v0.9.2` and current `main` and it doesn't do what its own docs say — filing separately since it's a distinct bug from #3924's original `tags_match=all_strict` report, in a different code path.

## What the docs promise

The [retain API docs](https://hindsight.vectorize.io/developer/api/retain#observation_scopes) for `shared` state: "The memory's own tags are ignored for observation scoping... so memories with *different* tags all consolidate into the *same* observation."

## What actually happens

Re-submitted two real facts (same entity, different native tags — one `category:reference,suggested_tool:gardea`, the other `category:fact,suggested_tool:terraform-cloud`) as new documents, both with `observation_scopes: "shared"` explicitly set. Tried twice (once bare, once with a `context` field matching normal client shape), ~90s consolidation wait each time:

- Fact A merged into its own pre-existing same-tag observation (`updated=1`).
- Fact B merged into its own pre-existing same-tag observation (`updated=1`).
- Never into one shared/untagged observation together, despite both requesting `shared` scope.

## Root cause

`consolidator.py` groups memories into LLM batches by their **raw native tag set** before `observation_scopes` is resolved at all:

```python
# consolidator.py:1436-1441 (current main, same on v0.9.2)
tag_groups: dict[tuple[str, ...], list[dict[str, Any]]] = {}
for m in memories:
    tag_key = tuple(sorted(m.get("tags") or []))
    tag_groups.setdefault(tag_key, []).append(dict(m))
```

`_resolve_obs_tags_list` is only consulted afterward, per batch, via `sub_batch[0]` (line 1509) — so two facts with different native tags are placed in two separate `tag_groups`, hence two separate LLM batches/calls, before either one's `observation_scopes: "shared"` is ever looked at. Each batch resolves its own target scope and pools merge candidates independently (`_find_related_observations`), so whether the two new facts land on the same observation depends entirely on whether one happens to already exist at that scope — the two *new* facts are never presented to the LLM together in the same call, so they can never be merged into one *new* shared observation.

Full trace + live test transcript: https://github.com/vectorize-io/hindsight/issues/3924#issuecomment-5481099255

## Fix

Opening a PR against this issue that resolves each memory's target observation scope *before* the tag-set split — grouping memories that name the same target scope together (so `shared`, or any single-scope `observation_scopes` override, batches by that resolved scope instead of native tags) — while keeping the "different tags never share an LLM call" security property intact for memories that stay on `combined`/default (still grouped by native tags, unchanged).

ferrastas:
Fixes #3953
Related: #3924

## Root cause

`observation_scopes="shared"` (or any other single-scope override) is documented to ignore a memory's own tags for observation scoping, so memories with *different* tags all consolidate into the *same* observation.

That's not what happened: `run_consolidation_job` grouped memories into LLM consolidation batches by their **raw native tag set** (`tag_groups`) *before* `observation_scopes` was resolved at all:

```python
tag_groups: dict[tuple[str, ...], list[dict[str, Any]]] = {}
for m in memories:
    tag_key = tuple(sorted(m.get("tags") or []))
    tag_groups.setdefault(tag_key, []).append(dict(m))
```

`_resolve_obs_tags_list` was only consulted afterward, per batch, via `sub_batch[0]`. So two facts carrying different native tags but both requesting the identical target scope (e.g. both `observation_scopes: "shared"`) were split into two separate tag groups and therefore two separate LLM calls, before either memory's scope override was ever looked at. Each batch resolved and pooled merge candidates independently, so the two *new* facts could never be presented to the LLM together and could never be merged into one new shared observation — contradicting the documented `shared` behavior.

Full repro/trace: https://github.com/vectorize-io/hindsight/issues/3953

## Fix

Added `_consolidation_batch_key(memory)` in `consolidator.py`, which resolves each memory's *target* observation scope before grouping:

- Default `combined` mode, or a multi-scope fan-out (`per_tag` with more than one tag, `all_combinations`, or an explicit multi-scope list) — falls back to the memory's own native tags, unchanged from before.
- A single-scope override (`shared`, `per_tag` with exactly one tag, or an explicit one-scope list) — returns that resolved scope instead, so memories naming the *same target scope* batch together regardless of differing native tags.

`tag_groups` construction in `run_consolidation_job` now keys on `_consolidation_batch_key(m)` instead of the raw tag tuple. This preserves the "different observation scopes must never share an LLM call" security invariant that the original native-tag grouping only approximated correctly in the default `combined` case.

## Test added

`test_shared_mode_pools_different_native_tags_into_one_llm_batch` in `hindsight-api-slim/tests/test_consolidation_scope_parallelism.py`:

- Two memories with **different native tags**, both `observation_scopes="shared"`.
- Uses `consolidation_llm_batch_size=2` — the existing sibling test (`test_shared_mode_parallel_writes_only_untagged_scope`) uses `batch_size=1`, which doesn't exercise batching at all since every memory always gets its own LLM call regardless of grouping.
- A callback-based mock LLM records the set of fact ids seen in the prompt for every consolidation-scope call.
- Asserts exactly one LLM call was made and it contained both memories' fact ids, plus `result["status"] == "completed"`.

## CI notes

This change only touches `hindsight_api/engine/consolidation/consolidator.py` and a test file — no generated files, schemas, or OpenAPI spec are affected, so `verify-generated-files` and similar generation-check jobs should be unaffected.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -654,6 +654,28 @@ def _resolve_write_scopes(memory: dict[str, Any]) -> list[frozenset[str]]:
     return [frozenset(s) for s in parsed]  # explicit list[list[str]]
 
 
+def _consolidation_batch_key(memory: dict[str, Any]) -> tuple[str, ...]:
+    """Return the key that decides which memories may share an LLM batch.
+
+    The real security requirement is "memories targeting different observation
+    scopes must never share an LLM call" — tags are only a proxy for that in
+    the default ``combined`` mode, where a memory's target scope *is* its own
+    tag set. A memory that names a single alternate scope (``shared``, an
+    explicit one-scope list, or ``per_tag`` with exactly one tag) targets that
+    scope instead, so it batches with any other memory naming the identical
+    scope regardless of native tags — that is the whole point of requesting
+    it. A memory that fans out to *multiple* scopes (``per_tag`` with more
+    than one tag, ``all_combinations``, or a multi-scope explicit list) writes
+    several observations from one LLM call; there is no single target scope
+    to key on, so it keeps grouping by native tags like the default case
+    (unchanged from before this function existed).
+    """
+    resolved = _resolve_obs_tags_list(memory)
+    if resolved is not None and len(resolved) == 1:
+        return tuple(sorted(resolved[0]))
+    return tuple(sorted(memory.get("tags") or []))
+
+
 def _scope_sort_key(scope: frozenset[str]) -> tuple[str, ...]:
     """Total ordering on scope frozensets for deadlock-free lock acquisition.
 
@@ -1433,11 +1455,15 @@ async def _run_consolidation_job(
         if not memories:
             break  # No more unconsolidated memories
 
-        # Group memories by exact tag set before batching — security requirement:
-        # memories with different tags must never share an LLM call.
+        # Group memories by target observation scope before batching — security
+        # requirement: memories targeting different scopes must never share an
+        # LLM call. See _consolidation_batch_key for why that's scope, not raw
+        # tags: a memory requesting observation_scopes="shared" (or another
+        # single-scope override) targets a scope that can differ from its own
+        # tags, and must only batch with peers naming that same scope (#3924).
         tag_groups: dict[tuple[str, ...], list[dict[str, Any]]] = {}
         for m in memories:
-            tag_key = tuple(sorted(m.get("tags") or []))
+            tag_key = _consolidation_batch_key(m)
             tag_groups.setdefault(tag_key, []).append(dict(m))
 
         # Split each tag group into LLM batches respecting llm_batch_size, keeping
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -658,22 +658,32 @@ def _consolidation_batch_key(memory: dict[str, Any]) -> tuple[str, ...]:
     """Return the key that decides which memories may share an LLM batch.
 
     The real security requirement is "memories targeting different observation
-    scopes must never share an LLM call" — tags are only a proxy for that in
-    the default ``combined`` mode, where a memory's target scope *is* its own
-    tag set. A memory that names a single alternate scope (``shared``, an
-    explicit one-scope list, or ``per_tag`` with exactly one tag) targets that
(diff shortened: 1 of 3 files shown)


ebarkhordar (commented):
Static read at `bf7971e`, not executed.

The fallback branch in `_consolidation_batch_key` still keys on native tags, so it can group two memories that resolve differently. Tags `["a","b"]` with no override resolves to `None`; the same tags with `observation_scopes="per_tag"` resolves to `[["a"],["b"]]`. One is `None` and the other has length 2, so both fall through to `("a","b")` and land in the same group.

That group is split by `consolidation_llm_batch_size`, and line 1524 takes the scoping for the whole sub-batch from `sub_batch[0]`. At batch size 2 or more the two can share a sub-batch, and whichever comes first decides: either the combined memory suppresses the other's per-tag fan-out, or the per_tag memory fans out one that never asked for it.

Your change does not introduce this and the same pair collides on main. I raise it because the new docstring treats that branch as safe on the grounds that it is unchanged, and the invariant it names, that memories targeting different scopes never share an LLM call, is exactly what the branch does not give you. The comment at line 1480 already notes scopes can differ within a group, but says it about the lock layer.

Is the multi-scope fan-out case deliberately out of scope here?

ferrastas:
Fixed! @ebarkhordar `_consolidation_batch_key` now bases the multi-scope fan-out branch on the full resolved scope-list instead of native tags, with a distinct marker per branch (`combined`/`scope`/`fanout`) — so a `combined` memory and a `per_tag` memory sharing tags can never collide into the same batch, no matter which one lands as `sub_batch[0]`.

Added a regression test (`test_combined_and_per_tag_same_tags_do_not_share_a_batch`) covering exactly your `["a","b"]` combined vs. `["a","b"]` per_tag example — checks the combined memory still gets one observation over `{a,b}` and the per_tag one gets its two separate `{a}`/`{b}` observations, regardless of ordering.

Kept it tight — same function this PR already touches, nothing else changed. Really appreciate you tracing this all the way through to the `sub_batch[0]` consequence, sharp catch!

ebarkhordar:
Read the new key at `4e1f190`, did not run it. The per branch marker is what closes it: a `combined` memory tagged `["a","b"]` keys as `("combined","a","b")` and the `per_tag` one as `("fanout", ...)`, so they cannot land in the same group and `sub_batch[0]` no longer decides for both. Thanks for adding the test.

PR #3955, opened 2026-09-01
fix(consolidation): reject incomplete destructive batches

r266-tech:
## Summary

- reject incomplete destructive LLM action sets before they can delete observations or stamp their source facts
- lock and compare recalled observation witnesses before destructive writes, then commit replacements, deletes, and source stamps together
- preserve exact-duplicate no-ops and fail closed when a store-owned backend cannot join the SQL transaction

## Why

#3923 made the writes from one LLM response transactional, but documented one remaining gap: a valid `DELETE` can still commit when its intended `CREATE` or `UPDATE` is silently discarded for an invalid reference. The source facts are then stamped as consolidated, the old observation is gone, and the operation can report success.

This follow-up treats a destructive response as one completion contract. Invalid references fail the response before preparation; a replacement that becomes inapplicable during the write rolls the transaction back; and a recalled observation must still match its text/version witness under a deterministic row-lock order before any mutation.

## Compatibility boundaries

- A create-only invalid reference keeps the existing skip behavior: its source fact remains durable and recallable but is considered processed, while no observation is created. Only an action set capable of deleting or folding an existing observation is promoted to failure and adaptive bisection.
- Update source validation keeps the existing "recalled for at least one source" relevance rule. A consolidation batch contains one exact tag set, so per-fact recall here is relevance ranking rather than an authorization boundary; requiring every source query to independently retrieve the same target would reject legitimate multi-fact aggregation.
- Store-owned observations have no SQL witness/CAS primitive. Destructive or semantic-fold batches therefore fail closed instead of issuing an external mutation that cannot share the transaction.

## Verification

- `uv run pytest tests/test_consolidation_batch_atomicity.py -q --disable-warnings` — 22 passed
- `uv run ruff check hindsight_api/ tests/test_consolidation_batch_atomicity.py`
- `uv run ruff format --check hindsight_api/ tests/test_consolidation_batch_atomicity.py`
- repository `lint.sh` reached the unrelated control-plane ESLint step but that workspace lacks `@eslint/js`; the changed Python surface passed its lint and type gates

The regressions cover invalid replacement references, rollback after an in-transaction skip, capacity reuse and duplicate-delete accounting, exact-duplicate cancellation, stale delete/update witnesses, deterministic dedup-target locking, non-destructive compatibility, and store-owned fail-closed behavior.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -34,7 +34,7 @@ from pydantic import BaseModel, ValidationError, field_validator
 
 from ...config import get_config
 from ...worker.stage import set_stage
-from ..db import DatabaseBackend
+from ..db import DatabaseBackend, DatabaseConnection
 from ..db_utils import acquire_with_retry
 from ..llm_interface import OutputTooLongError, ProviderRateLimitResetError
 from ..llm_trace import (
@@ -284,6 +284,7 @@ class _DedupOutcome:
     # The twin's text at probe time. Guards the fold against a concurrent survivor
     # rewrite during the connection-free LLM window (set on the two non-None returns).
     best_text: str = ""
+    best_updated_at: datetime | None = None
 
 
 async def _dedup_adjudicate(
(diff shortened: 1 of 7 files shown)


koriyoshi2041 (approved):
Reviewed at 94df966b. The destructive-batch preflight, recalled-state witness checks, and transaction rollback keep deletes, replacements, dedup folds, and source stamps under one fail-closed contract. The non-destructive invalid-create compatibility path remains covered separately.

I reran the complete atomicity file (22 passed), Ruff lint/format on the changed Python surface, and diff hygiene locally. All passed.

PR #3959, opened 2026-09-01
fix(coding-agents): skip bank id when a git probe fails

### LINKED ISSUE #3950: coding-agents: a failed git probe silently forks a worktree into its own bank
## Summary

A transient failure of the single `git rev-parse` probe in `getProjectRootFromGit` silently changes a repository's bank identity. Instead of resolving a linked worktree to its main worktree's name, the plugin falls back to the worktree directory's own basename and retains into a brand-new bank — permanently, with no log line, no diag event, and no retry.

The result in the wild: one repository accumulated eight stray banks (`coding-agent::<repo>-wt1` … `-wt9`) alongside the correct `coding-agent::<repo>`, and a second repository accumulated seven more, all for directories that are ordinary linked worktrees of those repos.

## Mechanism

`hindsight-integrations/coding-agents/src/core/bank.ts`:

```ts
const commonDir = execFileSync(
  "git",
  ["rev-parse", "--path-format=absolute", "--git-common-dir"],
  { cwd: directory, /* … */ timeout: 1000, /* … */ }
).trim();
// …
} catch {
  return null;   // <- any failure, indistinguishable from "not a repo"
}
```

`gitProjectName` treats that `null` as "git cannot name this location" and falls back to
`dirName(sessionRoot || directory)` — which, for a linked worktree, is the worktree's own
basename. The intended answer was the main worktree's basename.

The fallback is correct for its designed case (a directory outside any repository). It is wrong
for a *failed probe* inside a repository, and the two are currently the same code path.

Two ways to trigger it, both silent and both load-dependent:

- the 1000 ms `timeout` elapsing under machine load;
- `execFileSync` failing to spawn at all (`EAGAIN` under process pressure).

## Evidence

The stray banks were created in bursts during a window where the machine was visibly saturated
(concurrent sessions, aborted HTTP write-backs, one `session_start` diag event reporting
`ms: 1664`). Each stray bank carries only 1–2 diag events in total, while the correct bank for the
same repository carries hundreds — the signature of a one-off failure, not a misconfiguration.

Resolution is correct again on a quiet machine:

```
$ node ~/.hindsight/coding-agents/dist/status.js --repo ~/dev/<repo>-wt2
{"bank":"coding-agent::<repo>", …}
```

but the bank created during the failure still exists server-side and still holds whatever that
session retained into it.

The `.git` directory involved is ~930 MB with 46 registered worktrees; a warm probe takes ~10 ms,
so the 1 s budget is normally ample and only load makes it bite.

## Why it matters

Bank identity is the one value in the plugin that must not degrade silently. A wrong bank id is
not a degraded session — it is a *different memory*: the session's retains land somewhere nobody
reads, the repo's own bank loses that history, and the stray bank persists forever because nothing
ever cleans it up. The user's first symptom is "my memory is missing", several days later.

## Suggested direction

1. **Distinguish failure from absence.** `getProjectRootFromGit` should tell "this is not a git
   repository" (git exited non-zero with a clean answer) apart from "the probe did not complete"
   (timeout / spawn error). Only the former should reach the basename fallback.
2. **Retry the probe** a couple of times with a small backoff before giving up — the failure is
   transient by nature.
3. **Never guess a bank id after a failed probe.** If the repository cannot be identified, the
   safer outcome is to skip retaining for that session than to invent a new bank; a skipped
   session is recoverable, a scattered one is not.
4. **Log it.** A `warn` + a diag event on the fallback would have made this diagnosable in minutes
   instead of by cross-referencing bank creation timestamps against a log file.
5. Consider raising the 1000 ms budget, but this is a mitigation, not the fix — the spawn-failure
   path ignores the timeout entirely.

## Cleanup

Users who hit this need a way to fold a stray bank back into the real one (or at minimum, to find
them: banks named `coding-agent::<something>` whose directory is a linked worktree of another bank's
repo are always wrong).

leilei3167:
Fixes #3950

`getProjectRootFromGit` treated every `git rev-parse --git-common-dir` failure as "not a repo", so a 1s timeout or `EAGAIN` on a linked worktree named the bank after the worktree basename (`coding-agent::<repo>-wtN`) and retained there with no log.

This split "not a git repository" from probe failure, retries timeout/`EAGAIN`/`ENOMEM` up to 3 times, and skips retain/session-start/hook when git still cannot name the repo. The existing basename fallback stays for directories that really are outside a repo.

Verified with `vitest run` on `bank.test.ts`, `bank-bare-hub.test.ts`, `bank-missing-dir.test.ts`, `hook.test.ts`, `retain-hook.test.ts`, `session-start.test.ts` (89 tests).

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/bank.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/bank.test.ts
@@ -29,6 +29,26 @@ describe("deriveBankId", () => {
     expect(deriveBankId({}, "/home/me/dev/myrepo-feature-wt")).toBe("coding-agent::myrepo");
   });
 
+  it("retries a transient git probe and still shares the main worktree bank", () => {
+    let n = 0;
+    mockExec.mockImplementation(() => {
+      n += 1;
+      if (n < 3) {
+        throw Object.assign(new Error("spawnSync git ETIMEDOUT"), { code: "ETIMEDOUT" });
+      }
+      return "/home/me/dev/myrepo/.git\n";
+    });
+    expect(deriveBankId({}, "/home/me/dev/myrepo-wt2")).toBe("coding-agent::myrepo");
+    expect(n).toBe(3);
+  });
+
+  it("does not invent a worktree bank when the git probe keeps failing", () => {
+    mockExec.mockImplementation(() => {
+      throw Object.assign(new Error("spawnSync git ETIMEDOUT"), { code: "ETIMEDOUT" });
+    });
+    expect(deriveBankId({}, "/home/me/dev/myrepo-wt2")).toBe("");
+  });
+
   it("uses the bare-repo directory name when common-dir is not .git", () => {
     mockExec.mockReturnValue("/srv/git/myrepo.git\n");
     expect(deriveBankId({}, "/srv/git/myrepo.git")).toBe("coding-agent::myrepo.git");
--- a/hindsight-integrations/coding-agents/src/core/bank.ts
+++ b/hindsight-integrations/coding-agents/src/core/bank.ts
@@ -25,6 +25,8 @@ import { execFileSync } from "node:child_process";
 import { existsSync } from "node:fs";
 import { homedir } from "node:os";
 import { basename, dirname, join, normalize, sep } from "node:path";
+import { diag } from "./diag";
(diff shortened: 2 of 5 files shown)


nicoloboschi:
fixed by #3981

PR #3961, opened 2026-09-01
fix(memory-defense): warn when a policy names a detector this build cannot run

rschlek:
# Warn when a Memory Defense policy names a detector the running build cannot run

## The symptom

Set this on a bank running the OSS build:

```bash
curl -X PATCH "$HINDSIGHT/v1/default/banks/my-bank/config" \
  -H 'content-type: application/json' \
  -d '{
    "memory_defense": {
      "enabled": true,
      "rules": [
        { "on": "prompt_injection", "action": "block" }
      ]
    }
  }'
```

You get a `200`. `GET .../config` reads the rule back verbatim. Every retain to
the bank is then screened against nothing at all, because the loaded extension
implements only `sensitive_data`. No error, no warning, no log line, and no
`memory_defense.triggered` webhook to notice the absence of. The bank config is
the operator's only feedback surface, and it says the control is on.

The same holds for `size_anomaly`, `detect_secrets`, `llm_screen`, or any other
name. It is not limited to typos: a policy copied from a build with a larger
detector roster lands in exactly this state.

## What this does not change

This does not touch parsing. `parse_policy` still accepts any non-empty string
for `on`, and `test_parse_policy_accepts_any_detector_name` still passes
unmodified, `some_future_cloud_detector` included. The forward-compatibility
property described at
[`memory_defense.py:32-39`](hindsight-api-slim/hindsight_api/extensions/memory_defense.py#L32-L39)
is preserved intact:

> The parser therefore does NOT validate `on` against a fixed list — pinning
> the OSS roster to cloud's would force an OSS bump for every new cloud
> detector just to avoid 422-ing a write it never interprets.

That reasoning is right and this change depends on it. Rejecting unknown names
would break every deployment whose policy is authored against a build with more
detectors than the one running. Screening behavior is likewise unchanged: the
same decisions, the same redactions, the same blocks, the same webhooks. The
only thing added is a log line.

## Why the current behavior is worth changing

Thirteen lines above the site in question, in
[`orchestrator.py:70-73`](hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py#L70-L73):

```python
# No try/except around the parse. A malformed policy must fail the retain, not
# quietly skip screening: this is a security control, so fail-open is the wrong
# default even for an unreachable state (the HTTP layer 422s a bad policy on
# write, so one can only reach the store by a direct DB edit).
```

The principle stated there is the one this PR applies. That comment concerns a
state described as unreachable, and it still rejects failing open. At
[`orchestrator.py:86`](hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py#L86)
the code fails open silently for a state that is entirely reachable through the
public API: a well-formed policy, accepted with a 200, naming a detector this
build cannot honor.

```python
if not any(r.on == "sensitive_data" for r in policy.rules):
    return body
```

The other site is
[`memory_defense_regex.py:41-43`](hindsight-api-slim/hindsight_api/extensions/builtin/memory_defense_regex.py#L41-L43):

```python
rule = next((r for r in policy.rules if r.on == "sensitive_data"), None)
if rule is None or rule.action is DefenseAction.ALLOW:
    return DefenseDecision(action=DefenseAction.ALLOW)
```

Every rule that is not `sensitive_data` is discarded without being looked at.
The module declares a logger at line 21 and never uses it.

Failing open is arguably the right runtime behavior here, and this PR keeps it.
Doing it silently is the part that is not: an operator has no way to learn that
a configured control is inert. That is an observability gap, not an enforcement
one, so the fix is a log line rather than a behavior change.

## The change

`MemoryDefenseExtension` gains a class attribute naming the detectors its
`screen()` actually dispatches:

```python
implemented_detectors: frozenset[str] = frozenset()
```

`MemoryDefenseRegexExtension` declares `REGEX_DETECTORS`
(`frozenset({"sensitive_data"})`). A downstream extension declares its own
roster the same way and gets the same reporting for free, with no OSS release
needed when its roster grows. The default is empty rather than
`{"sensitive_data"}` so an extension that forgets to declare one is noisy
instead of silent.

Both paths call one shared helper, `warn_unimplemented_detectors(policy,
implemented)`, which logs any rule whose `on` is outside the roster and returns.
It reads the policy, makes no decision, and is a no-op for a disabled policy
(a disabled policy screens nothing by request, so there is nothing to report).
The literal `"sensitive_data"` moves to a `SENSITIVE_DATA_DETECTOR` constant
that the roster and both dispatch sites share, so a fourth copy of the string is
not introduced.

The line:

```
Memory Defense policy names detector 'prompt_injection', which this build does not implement.
The rule is ignored and nothing is screened for it. Detectors implemented here: ['sensitive_data'].
```

It carries the detector name and the implemented roster and nothing else. No
screened content, no other policy field, no bank id, nothing that could carry
user data. A test asserts that the secret and the surrounding body are absent
from the line.

## On log spam

A warning on the retain path that fires per screened item would be unacceptable:
one misconfigured rule would repeat the same line for every write to the bank
forever. The helper keeps a module-level set of names already warned about and
logs each one once per process. Notes on the mechanism:

- Deduplication is by detector name, so the two call sites share one ledger and
  a name reported by the screen path is not reported again by the document-body
  scrubber.
- Growth is bounded by the number of distinct detector names operators actually
  configure, not by traffic or by bank count. This is stated in a comment beside
  the set.
- The set is guarded by a `threading.Lock` because screening runs from worker
  threads as well as the event loop and check-then-add is not atomic. The log
  call itself happens outside the lock so screening never serializes on log I/O.
- The state is process-local. A restart warns again, which is what an operator
  reading a fresh log wants.
- A test screens five items under the same policy and asserts exactly one
  record.

## Tests

Nine tests added to `hindsight-api-slim/tests/test_memory_defense.py`, in the
`caplog` idiom already used elsewhere in the suite:

- an unimplemented detector produces exactly one warning and the decision is
  still ALLOW
- the warning fires once across repeated screening with the same policy
- two distinct unimplemented names produce one warning each
- a `sensitive_data` rule produces no warning and still redacts, hits and all
- a disabled policy produces no warning
- the extension declares its roster, and the base class default is empty
- `redact_document_body` warns once and still returns the body unchanged
- `redact_document_body` does not warn for `sensitive_data` and still scrubs
- the warning line contains neither the screened content nor the rule action

`test_parse_policy_accepts_any_detector_name` is untouched and passes.

Full file, including the DB-backed retain tests:

```
$ uv run pytest tests/test_memory_defense.py
270 passed   # before
279 passed   # after (+9 new, 0 changed, 0 removed)
```

`ruff check`, `ruff format --check`, and `ty check` are clean on the changed
modules.

## Files

- `hindsight-api-slim/hindsight_api/extensions/memory_defense.py`: the
  `implemented_detectors` roster attribute, the `SENSITIVE_DATA_DETECTOR` and
  `REGEX_DETECTORS` constants, and the `warn_unimplemented_detectors` helper
  with its warn-once ledger.
- `hindsight-api-slim/hindsight_api/extensions/builtin/memory_defense_regex.py`:
  declares its roster and calls the helper at the top of `screen()`.
- `hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py`: the same
  call in `redact_document_body`, against `REGEX_DETECTORS` since that path runs
  the regex detector directly rather than through a loaded extension.
- `hindsight-api-slim/tests/test_memory_defense.py`: the nine tests.
- `hindsight-docs/docs/developer/memory-defense/index.md`: a short subsection
  documenting that an unimplemented detector is accepted, ignored, and logged.

Happy to drop the docs hunk, move the warning to the extension's own logger, or
key deduplication on `(bank_id, detector)` instead of the name alone if you'd
rather have the bank in the line. The behavior change is confined to logging
either way.

### DIFF
--- a/hindsight-api-slim/hindsight_api/extensions/memory_defense.py
+++ b/hindsight-api-slim/hindsight_api/extensions/memory_defense.py
@@ -12,7 +12,9 @@ from __future__ import annotations
 
 import logging
 import re
+import threading
 from abc import ABC, abstractmethod
+from collections.abc import Collection
 from dataclasses import dataclass, field
 from enum import Enum
 
@@ -37,6 +39,20 @@ _VALID_ACTIONS = {a.value for a in DefenseAction}
 # new cloud detector just to avoid 422-ing a write it never interprets. We
 # only require ``on`` to be a non-empty string; entitlement and dispatch are
 # the loaded extension's ``screen()`` job.
+#
+# Dispatch being the extension's job does not make an undispatched rule
+# invisible: an extension declares the roster it implements (see
+# ``MemoryDefenseExtension.implemented_detectors``) and
+# ``warn_unimplemented_detectors`` logs, once per name, any rule outside it.
+# Parsing stays permissive; only the silence goes away.
+
+# The detector this module's ``apply_redaction`` implements. Named once so the
+# regex extension, the document-body scrubber and the roster below cannot
+# drift apart.
+SENSITIVE_DATA_DETECTOR = "sensitive_data"
+
+# The roster the OSS regex screening honours — the whole of it.
+REGEX_DETECTORS: frozenset[str] = frozenset({SENSITIVE_DATA_DETECTOR})
 
 
 @dataclass(frozen=True)
@@ -273,6 +289,49 @@ def apply_redaction(content: str) -> RedactionResult:
     return RedactionResult(content=content, matched_types=matched, hits=hits)
 
 
+# Detector names already warned about. The retain path screens every item
+# written, so warning per screened item would put the same line in the log
+# thousands of times for one misconfigured rule; warn once per name instead.
+# The set is bounded by the number of distinct detector names operators
+# actually configure, so it cannot grow with traffic. The lock is there
+# because screening runs on worker threads as well as the event loop and
+# check-then-add is not atomic: without it a concurrent pair of writes could
+# log the same name twice. Process-local, like any log-dedup state — a restart
+# warns again, which is what an operator reading a fresh log wants.
+_warned_detectors: set[str] = set()
+_warned_detectors_lock = threading.Lock()
+
+
+def warn_unimplemented_detectors(policy: DefensePolicy, implemented: Collection[str]) -> None:
+    """Log once per policy rule naming a detector this build does not implement.
+
+    The parser accepts any detector name on purpose (see the note above
+    :class:`PolicyRule`), so a well-formed policy can name a detector no loaded
+    extension dispatches. Such a rule is inert: the PATCH returns 200, the rule
+    is stored, and nothing ever screens for it. This makes that visible in the
+    log; it changes no decision and gates no write.
+
+    Only the detector name and the implemented roster are logged. Screened
+    content and every other field of the policy stay out of the line.
+    """
+    if not policy.enabled:
+        return
+    for rule in policy.rules:
+        if rule.on in implemented:
+            continue
+        with _warned_detectors_lock:
+            if rule.on in _warned_detectors:
+                continue
+            _warned_detectors.add(rule.on)
+        # Logged outside the lock: emitting to a handler under it would
+        # serialize screening on log I/O.
+        logger.warning(
+            f"Memory Defense policy names detector '{rule.on}', which this build does not implement. "
+            f"The rule is ignored and nothing is screened for it. "
+            f"Detectors implemented here: {sorted(implemented)}."
+        )
+
+
 class MemoryDefenseExtension(Extension, ABC):
     """Abstract base for Memory Defense extensions.
 
@@ -281,6 +340,14 @@ class MemoryDefenseExtension(Extension, ABC):
     applies the returned decision (redacts content / drops blocked items) and
     fires a webhook for non-allow decisions when one is configured.
     """
+
+    # Detector names this extension's ``screen()`` actually dispatches. A rule
+    # naming anything else is a no-op here, so ``screen()`` runs the policy
+    # through :func:`warn_unimplemented_detectors` against this roster to say
+    # so once. Downstream extensions declare their own roster; the default is
+    # empty so an implementation that forgets to declare one warns about every
+    # rule rather than silently swallowing them.
+    implemented_detectors: frozenset[str] = frozenset()
 
     @abstractmethod
     async def screen(
--- a/hindsight-api-slim/tests/test_memory_defense.py
+++ b/hindsight-api-slim/tests/test_memory_defense.py
@@ -21,6 +21,7 @@ from hindsight_api.extensions.loader import ExtensionLoadError, load_extension
 from hindsight_api.extensions.memory_defense import (
     _ASCII_TOKEN_START,
     _REDACTION_PATTERNS,
+    REGEX_DETECTORS,
     DefenseAction,
     MemoryDefenseExtension,
     _fingerprint_value,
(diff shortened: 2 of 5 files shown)


hindsight-docs/docs/developer/memory-defense/index.md
```diff
@@ -46,6 +46,19 @@ Enabling Memory Defense on a bank only affects future retain calls. Memories alr
 
 Memory Defense is off on every bank until you set a policy. A bank with no `memory_defense` field, with `enabled: false`, or with no `sensitive_data` rule is treated identically: the extension returns ALLOW and content passes through unchanged. To stop redacting on a bank that has it on, set `enabled: false` or remove the policy.
 
+### Detectors this build does not implement
```
nicoloboschi, line 49: **Blocking: `verify-generated-files` will fail on this.**

`hindsight-docs/docs/developer/**` is mirrored into `skills/hindsight-docs/references/developer/**` by `./scripts/generate-docs-skill.sh`, and that job re-runs the generator and fails on any leftover diff. The mirror of this file exists on `main` but isn't updated here:

```
docs source:                                                  3 hits for "does not implement"
skills/hindsight-docs/references/developer/memory-defense/index.md:  0
```

Run `./scripts/generate-docs-skill.sh` and commit `skills/hindsight-docs/references/developer/memory-defense/index.md`. `lint.sh` does not catch this one.

hindsight-api-slim/hindsight_api/extensions/memory_defense.py
```diff
@@ -273,6 +289,49 @@ def apply_redaction(content: str) -> RedactionResult:
...
+# actually configure, so it cannot grow with traffic. The lock is there
+# because screening runs on worker threads as well as the event loop and
+# check-then-add is not atomic: without it a concurrent pair of writes could
+# log the same name twice. Process-local, like any log-dedup state — a restart
+# warns again, which is what an operator reading a fresh log wants.
+_warned_detectors: set[str] = set()
+_warned_detectors_lock = threading.Lock()
+
+
+def warn_unimplemented_detectors(policy: DefensePolicy, implemented: Collection[str]) -> None:
```
nicoloboschi, line 305: **Design: I think this belongs on the config-write path, not the retain path.**

The motivating story in the description is *"you PATCH the config, get a 200, and get no feedback."* Putting the warning on the screening path is what forces everything awkward about the design — a process-global ledger, a `threading.Lock`, dedup by bare detector name, and no bank id in the line.

The write path already exists and already parses the policy, at `memory_engine.py:12470-12477`:

```python
if "memory_defense" in updates and updates["memory_defense"] is not None:
    try:
        parse_policy(updates["memory_defense"])
    except ValueError as exc:
        raise OperationValidationError(...)
```

Warning there instead: fires exactly once, at the moment the operator does the thing that needs feedback; has `bank_id` in scope; and can read `self._memory_defense.implemented_detectors` (the extension is held at `memory_engine.py:1332`) — which is also the thing that would make the new base-class attribute earn its keep polymorphically, rather than only ever being read as `REGEX_DETECTORS`. No ledger, no lock, no process-lifetime dedup needed.

The payoff of the current placement is thin by comparison: one line, no bank id, once per process. On a multi-tenant server the *second* bank configured with the same inert detector never produces a line at all — its operator gets exactly the silence this PR set out to remove. Keying dedup on `(bank_id, detector)` would patch that symptom, but the write path removes the need for dedup entirely.

hindsight-api-slim/hindsight_api/extensions/memory_defense.py
```diff
@@ -273,6 +289,49 @@ def apply_redaction(content: str) -> RedactionResult:
...
+
+    Only the detector name and the implemented roster are logged. Screened
+    content and every other field of the policy stay out of the line.
+    """
+    if not policy.enabled:
+        return
+    for rule in policy.rules:
+        if rule.on in implemented:
+            continue
+        with _warned_detectors_lock:
```
nicoloboschi, line 322: Minor: the ledger check is *inside* the lock, so for a policy carrying an unimplemented rule this acquires the global lock on **every screened item, forever** — not just until the first warning. The comment two lines down reasons carefully about not serializing screening on log I/O, but screening still serializes on the lock itself.

Double-check before acquiring:

```python
if rule.on in _warned_detectors:
    continue
with _warned_detectors_lock:
    if rule.on in _warned_detectors:
        continue
    _warned_detectors.add(rule.on)
```

(Cheap either way under the GIL — raising it because the surrounding comment makes the claim explicitly.)

hindsight-api-slim/hindsight_api/extensions/memory_defense.py
```diff
@@ -37,6 +39,20 @@ class DefenseAction(str, Enum):
...
+# ``warn_unimplemented_detectors`` logs, once per name, any rule outside it.
+# Parsing stays permissive; only the silence goes away.
+
+# The detector this module's ``apply_redaction`` implements. Named once so the
+# regex extension, the document-body scrubber and the roster below cannot
+# drift apart.
+SENSITIVE_DATA_DETECTOR = "sensitive_data"
+
+# The roster the OSS regex screening honours — the whole of it.
+REGEX_DETECTORS: frozenset[str] = frozenset({SENSITIVE_DATA_DETECTOR})
```
nicoloboschi, line 55: `warn_unimplemented_detectors`, `REGEX_DETECTORS` and `SENSITIVE_DATA_DETECTOR` aren't re-exported from `hindsight_api/extensions/__init__.py`, unlike `parse_policy`, `apply_redaction`, `DefenseAction`, `DefensePolicy` and `MemoryDefenseExtension`.

The description's "a downstream extension declares its own roster the same way and gets the same reporting for free" only works if it imports from the private module path. Worth adding to the `__init__` import block and `__all__`.

hindsight-api-slim/hindsight_api/extensions/memory_defense.py
```diff
@@ -282,6 +341,14 @@ class MemoryDefenseExtension(Extension, ABC):
     fires a webhook for non-allow decisions when one is configured.
     """
 
+    # Detector names this extension's ``screen()`` actually dispatches. A rule
+    # naming anything else is a no-op here, so ``screen()`` runs the policy
+    # through :func:`warn_unimplemented_detectors` against this roster to say
+    # so once. Downstream extensions declare their own roster; the default is
+    # empty so an implementation that forgets to declare one warns about every
+    # rule rather than silently swallowing them.
+    implemented_detectors: frozenset[str] = frozenset()
```
nicoloboschi, line 350: Nit: this reads more accurately as `ClassVar[frozenset[str]]` — it's class-level configuration, never per-instance state. (`Extension` isn't a dataclass, so nothing breaks as written.)

hindsight-api-slim/tests/test_memory_defense.py
```diff
@@ -395,6 +396,174 @@ async def test_screen_allows_benign_payloads(payload: str, regex_defense, redact
...
+    """The roster is declared once, not hardcoded at each dispatch site, so a
+    downstream extension can declare its own."""
+    assert MemoryDefenseRegexExtension.implemented_detectors == REGEX_DETECTORS
+    assert REGEX_DETECTORS == frozenset({"sensitive_data"})
+    # An extension that declares nothing warns about every rule rather than
+    # swallowing them.
+    assert MemoryDefenseExtension.implemented_detectors == frozenset()
+
+
+def test_redact_document_body_warns_that_an_unimplemented_detector_is_ignored(warn_ledger, caplog) -> None:
```
nicoloboschi, line 522: Placement: these three `redact_document_body` tests sit in the screening section but call `_defense_config`, which is defined at line ~1061 under the `# redact_document_body: the oversized-item scrubber, in isolation` banner — where the other four `redact_document_body` tests already live. It works (module-level name resolved at call time), but it reads as though the helper is missing.

Suggest moving these three down into that section, and adding the new warn-once section to the module docstring's section list at the top, which the file otherwise keeps in sync.

nicoloboschi (changes requested):
The premise is right and the diff is small and safe: parsing stays permissive, screening decisions are unchanged, only a log line is added. I checked the branch out and ran the file locally — the new tests pass. Note that **no CI has run here** (`gh pr checks` reports no checks on the branch), so local runs are the only signal.

One blocking item (generated-file drift), one design question about where the warning belongs, and some minor notes inline.

### Verified good

- Warn-once is race-correct: the name is added under the lock *before* the log emits, so a concurrent pair can't double-log.
- `test_parse_policy_accepts_any_detector_name` is genuinely untouched; the forward-compat property is preserved as claimed.
- The "no user data in the line" test is real — the line carries only the detector name and the sorted roster.
- The existing parametrized case `{"on": "something_else"}` now populates the global ledger; harmless, and the `warn_ledger` fixture clears either side.

nicoloboschi:
close as stale

PR #3966, opened 2026-09-01
fix(coding-agents): only ever ADD to a bank's config, never overwrite it

### LINKED ISSUE #3927: Hindsight coding-agents plugin clobbers bank configuration with template, no way to disable
### Bug Description

The coding agent plugin is hard-coded to apply bank configuration, at every `deepen` call, which happens at least once per new session. This overwrites bank configuration with templated coding-specific values. Any attempt by the user to modify their own strategies and missions is overwritten. There is a mechanism that checks whether user-defined missions exist for Retain, Observations, or Reflect, and it doesn't overwrite those values, but it still writes the 5 coding retain strategies and entity labels as part of CODING_BANK_STRUCTURE, including overwriting any chunking or extraction mode settings, making it impossible to control bloat.

This makes some sense if using the bank-per-repo (`mapPathToBank` or dynamic) bank resolution strategies, but it does not take into account the global, single-bank-per-user strategy. It essentially takes over any bank it's pointed at, and re-asserts its pattern at every new session. Forking the code makes updates difficult. This needs to be thought through, and at the very least, configurable in `coding-agents.json`.

For context, I use a global bank with project tagging, for both coding and personal-assistant use cases. This is quite a serious bug in this use case, as it's overwritten most of my config settings. Those using dynamic banks will not necessarily notice, and might even appreciate this feature. Since the documantation ostensibly supports single-bank usage, this should be made much more clear.

### Steps to Reproduce

1. Install coding-agents plugin.
2. Configure with `dynamicBankId: false` and a static, global bank ID. 
3. Start a new session. 
4. Change bank configuration (e.g. "conversation" mission).
5. Start a new session, and observe the mission being overwritten.

### Expected Behavior

Should offer a property to disable this feature, or apply a custom template. Users should be warned that this will happen in plugin documentation. 

### Actual Behavior

Most configuration is overwritten at every `deepen` call, i.e. when new sessions are started, and at other times during long-running sessions.

### Version

0.4.3

### LLM Provider

OpenAI

nicoloboschi:
Closes #3927.

`configureBank` runs on every session start, and it re-sent the plugin's five retain strategies and its `knowledge` entity-label group as whole values. The server stores each of those as ONE config value, so an import replaces the map outright: a strategy the user had defined was deleted, their edits to the plugin's own strategies (mission, extraction mode, chunk size) were reverted, and `retain_default_strategy` could be left naming a strategy that no longer existed — silently, once per session, on whatever bank the plugin was pointed at. Pointed at a global bank shared with non-coding work, the plugin took the bank over.

This is the third fix for one shape. #1270 covered OpenClaw's missions, #2492 this plugin's; each protected only the fields that had just been noticed, and everything else kept being stamped back. So the rule is now the whole surface rather than a list: read the bank's current OVERRIDES and write only where the bank is silent. `codingBankManifest()` returns the template fields still missing, or nothing at all — a settled bank now makes no import call.

The container fields merge per entry, which keeps the reason the re-apply exists: a strategy ADDED by a newer release still reaches an existing bank. What it gives up is deliberate — a release that REWORDS an existing strategy or label does not reach a bank that already has it; clearing that override takes the current default back on the next pass.

`manageBankConfig: false` keeps the plugin out of the bank's configuration entirely, for a bank whose owner shapes it themselves. That bank should then define the strategies this plugin retains under. The cost is documented, and it is a silent one: the server does not *reject* a retain naming a strategy the bank lacks — `apply_strategy` logs a warning and extracts with the bank's own config, so a commit diff, a session transcript and a survey marker would all get the same generic treatment. Like every other field it can be set per bank, which is where a global-bank user wants it.

Knowledge pages are seeded either way: they are not bank configuration, and `pageTriggerType` already governs what they cost.

## Fixes, against the reporter's own findings

| # | Finding | Now |
|---|---|---|
| 1 | a custom strategy is deleted, not merged | kept — the map merges per entry |
| 2 | `retain_default_strategy` left pointing at a deleted strategy | cannot happen; nothing is deleted, and the pointer is only written when the bank has none |
| 3 | customizations to `conversation` reverted (concise → verbose, 4000 → 12000) | left alone; an entry already present is never rewritten |
| 4 | pages seeded on a zero-fact bank | not addressed here — that is the page-seeding path, not bank config (see #3905) |

## Tests

- additive merge: a strategy added by a newer release reaches an old bank; a user's own strategy survives; edits to the plugin's own survive; a settled bank yields no manifest; the label group is added alongside the user's own and only once; the missions stay group-seeded (#2492)
- client level: no `/import` call at all once the bank carries the structure, and `manageBankConfig: false` makes no config calls (not even the probe) while pages still seed
- `codingBankManifest({})` is asserted equal to `CODING_BANK_TEMPLATE`, so the builder cannot drift from the declared template

683/683 package tests pass, `tsc` clean, `./scripts/hooks/lint.sh` clean. README updated and the skill + docs page regenerated from it.

https://claude.ai/code/session_01M9KwqWJZRrw7NjAaKuACzj

### DIFF
--- a/hindsight-docs/docs-integrations/coding-agents.md
+++ b/hindsight-docs/docs-integrations/coding-agents.md
@@ -419,6 +419,7 @@ hook by Codex...), so one shared config serves several agents side by side:
 | `resolveWorktrees`      | `true`                               | linked worktrees inherit the main checkout's bank identity, path approval, and mapping                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
 | `retainTags`            | —                                    | extra tags on every document written by the integration, e.g. `["project:{gitProject}"]` — see **Recording where a memory came from** below                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
 | `retainMetadata`        | —                                    | extra metadata on every document written by the integration, e.g. `{"repo": "{gitProject}"}`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
+| `manageBankConfig`      | `true`                               | let the plugin shape the bank's own configuration — the retain strategies it writes under, the `knowledge` entity-label group, and, on a bank that has none, the missions. Writing is strictly **additive**: it adds what the bank does not define and never overwrites what is there, so your control-plane edits survive. Set `false` to keep it out of the bank config entirely — see **A bank you shape yourself** below                                                                                                                                                                                                                                                                 |
 | `observationScopes`     | `"shared"`                           | how consolidation groups observations: `"shared"` (default) = ONE global scope per bank, so every agent on a repo builds one set of beliefs; also `"combined"` (the server default), `"per_tag"`, `"all_combinations"`, `[["t"]]`                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
 | `disabled`              | `false`                              | hard off-switch (inert plugin/hook — a no-memory baseline)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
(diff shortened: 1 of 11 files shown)


dvinas:
Amazing, thanks! Great fix as well.

PR #3971, opened 2026-09-01
fix(bank-template): resolve stale snapshot race during template import

### LINKED ISSUE #3957: Template import uses stale resource snapshots across concurrent requests
## Description

Template import makes create/update decisions from resource snapshots that can become stale before the corresponding writes execute.

Affected code:
- `hindsight-api-slim/hindsight_api/api/http.py` `apply_bank_template_manifest()`
- `hindsight-api-slim/hindsight_api/engine/memory_engine.py` `bank_template_import_authorization()` and preauthorization consumption

The current sequence is:

```text
Request A reads the bank's mental model/directive state
Request A preauthorizes CREATE_* or UPDATE_* based on that snapshot
Request B creates, deletes, or modifies the same resource
Request A continues using the stale classification and resource identity
```

Observed consequences:

- If A classified a mental model/directive as CREATE and B creates it first, A can hit a uniqueness error or produce duplicate resources.
- If A classified a resource as UPDATE and B deletes it first, A can update zero rows and a subsequent refresh/get can fail.
- If B modifies a resource after A's snapshot, A can overwrite B's committed changes with stale import data.
- When the resource's CREATE/UPDATE classification changes, the existing preauthorization is not automatically invalidated or re-evaluated.

There is already a re-read for resources projected to collide with the server default template, and bank config persistence detects bank-level deletion/change. Those checks do not cover ordinary mental model/directive creation, deletion, or modification after the import snapshot.

Relevant code:
- `hindsight-api-slim/hindsight_api/api/http.py` `apply_bank_template_manifest()`, `_apply_authorized_bank_template_manifest()`
- `hindsight-api-slim/hindsight_api/engine/memory_engine.py` `bank_template_import_authorization()` and preauthorization consumption

Sanderhoff-alt:
## Summary

Fixes #3957

This PR resolves a concurrency race condition in the bank template import endpoint (`POST /v1/default/banks/{bank_id}/import`) where create vs. update decisions were made based on an un-isolated resource snapshot taken before authorization and bank provisioning.

We apply **uniform dual preauthorization (`CREATE` and `UPDATE`)** for all imported mental models and directives, and defer reading the fresh bank state to inside the authorization scope immediately before writes execute.

---

### Before (Vulnerable Sequence)

```mermaid
sequenceDiagram
    autonumber
    actor Client
    participant ImportAPI as apply_bank_template_manifest
    participant Validator as _operation_validator
    participant DB as Postgres/Oracle DB

    Client->>ImportAPI: POST /banks/{id}/import (Manifest)
    ImportAPI->>DB: Un-isolated list_mental_models() & list_directives()
    Note over ImportAPI: Categorize into CREATE_* or UPDATE_* (Stale Snapshot)
    ImportAPI->>Validator: Preauthorize single operation per resource
    Note over DB: Concurrent Request modifies/creates/deletes same resource!
    ImportAPI->>DB: _ensure_bank_exists() + apply default template
    ImportAPI->>DB: Apply resources using stale categories
    Note over ImportAPI,DB: ❌ UniqueViolation / Zero Rows Updated / RuntimeError (Not Preauthorized)
```

### After (Safe & Streamlined Pipeline)

```mermaid
sequenceDiagram
    autonumber
    actor Client
    participant ImportAPI as apply_bank_template_manifest
    participant Validator as _operation_validator
    participant DB as Postgres/Oracle DB

    Client->>ImportAPI: POST /banks/{id}/import (Manifest)
    Note over ImportAPI: Build uniform dual preauthorizations [CREATE, UPDATE]
    ImportAPI->>Validator: Validate dual authorizations upfront
    ImportAPI->>DB: _ensure_bank_exists() + apply default template
    rect rgb(240, 248, 255)
        Note over ImportAPI,DB: Inside Authorization Scope
        ImportAPI->>DB: Fetch fresh list_mental_models() & list_directives()
        ImportAPI->>DB: Apply resources matching committed state
        Note over ImportAPI,DB: Consume corresponding preauthorization safely
    end
    ImportAPI-->>Client: BankTemplateImportResponse
```

---

## State Matrix & Failure Mode Resolution

| Concurrency Scenario | Previous Behavior | Behavior With This Fix |
| :--- | :--- | :--- |
| **Concurrent Creation** (Resource did not exist during initial check, created by another request before write) | Tried to execute `create_*`; failed with primary key `UniqueViolation` or inserted duplicate directives. | Fresh read inside scope detects existence; executes `update_*` and consumes preauthorized `UPDATE_*`. |
| **Concurrent Deletion** (Resource existed during initial check, deleted by another request before write) | Executed `update_*` against 0 rows (returning `None`); subsequent refresh/get failed; falsely reported updated. | Fresh read inside scope detects deletion; falls back to `create_*` and consumes preauthorized `CREATE_*`. |
| **Server Default Template Collision** (Missing bank provisioned default models/directives) | Required separate heuristic projection logic (`projected_*`) and conditional re-reads. | Naturally handled by the same uniform pipeline without special-case branching. |
| **Preauthorization Scope Mismatch** | Operation transition caused `_consume_preauthorized_bank_write` to raise `RuntimeError: not preauthorized`. | Both `CREATE` and `UPDATE` are preauthorized, allowing seamless transition. |

---

### 1. HTTP API Layer (`hindsight-api-slim/hindsight_api/api/http.py`)
- Removed obsolete `load_default_bank_template_manifest()` projection logic, `projected_mental_model_ids`, `projected_directive_names`, and initial un-isolated `list_*` reads from `apply_bank_template_manifest()`.
- Preauthorize both `[UPDATE_MENTAL_MODEL, CREATE_MENTAL_MODEL]` and `[UPDATE_DIRECTIVE, CREATE_DIRECTIVE]` for all items in the imported manifest.
- Moved fresh state snapshot fetching (`list_mental_models`, `list_directives`) into `bank_template_import_authorization` scope directly preceding `_apply_bank_template_resources()`.

### 2. Automated Test Suite
- Added `test_import_handles_resource_created_before_write` in `test_bank_templates.py` to verify safe update transition when resources are created concurrently before the write phase.
- Added `test_import_handles_resource_deleted_before_write` in `test_bank_templates.py` to verify fallback to creation when resources are deleted before the write phase.
- Updated `test_import_reuses_each_preauthorization_once` in `test_http_api_integration.py` to assert the uniform dual-authorization contract.

---

## Verification

All template import integration tests and full linting checks pass:

```bash
cd hindsight-api-slim && uv run pytest tests/test_bank_templates.py tests/test_http_api_integration.py -k "template or import"
# 51 passed in 81.61s

./scripts/hooks/lint.sh
# All lints passed ✓ (Ruff, ESLint, ty type check)
```

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -3056,23 +3056,73 @@ async def apply_bank_template_manifest(
         is not None
     )
 
+    # A missing bank receives the server-owned default template during
+    # provisioning. Project those resources into the authorization decision so
+    # the client's import is authorized as an update when the default owns the
+    # same key, while still keeping every client check before bank creation.
+    default_manifest: BankTemplateManifest | None = None
+    if not bank_exists:
+        try:
+            default_manifest = load_default_bank_template_manifest()
+        except (ValueError, ValidationError):
+            # Provisioning owns error logging and the best-effort fallback for a
+            # malformed server template. Client authorization must not change it.
+            pass
+    imported_mental_model_ids = {item.id for item in manifest.mental_models or []}
+    imported_directive_names = {item.name for item in manifest.directives or []}
+    default_mental_models = (default_manifest.mental_models or []) if default_manifest else []
+    default_directives = (default_manifest.directives or []) if default_manifest else []
+    projected_mental_model_ids = {item.id for item in default_mental_models} & imported_mental_model_ids
+    projected_directive_names = {item.name for item in default_directives} & imported_directive_names
+
+    # limit=None throughout the import path: a create/update decision per imported
+    # resource is only correct against the bank's *whole* set. Under the default
+    # page size a bank with more than 100 models would look like it lacked the
+    # ones past the first page, and the import would create duplicates.
+    existing_by_id: dict[str, dict[str, Any]] = {}
+    if bank_exists and manifest.mental_models:
+        existing = await memory.list_mental_models(
+            bank_id=bank_id, limit=None, detail="metadata", request_context=request_context
+        )
+        existing_by_id = {m["id"]: m for m in existing.items}
+
+    existing_by_name: dict[str, dict[str, Any]] = {}
+    if bank_exists and manifest.directives:
+        existing_directives = await memory.list_directives(
+            bank_id=bank_id, active_only=False, limit=None, request_context=request_context
+        )
+        existing_by_name = {d["name"]: d for d in existing_directives.items}
+
     bank_writes: list[BankTemplateImportWrite] = []
     if config_updates:
         bank_writes.append(BankTemplateImportWrite(BankWriteOperation.UPDATE_BANK_CONFIG))
     for mental_model in manifest.mental_models or []:
-        bank_writes.extend(
-            [
-                BankTemplateImportWrite(BankWriteOperation.UPDATE_MENTAL_MODEL, mental_model.id),
-                BankTemplateImportWrite(BankWriteOperation.CREATE_MENTAL_MODEL, mental_model.id),
-            ]
-        )
+        if mental_model.id in existing_by_id:
+            bank_writes.append(BankTemplateImportWrite(BankWriteOperation.UPDATE_MENTAL_MODEL, mental_model.id))
+        elif mental_model.id in projected_mental_model_ids:
+            # Default-template application is best-effort. Authorize both
+            # outcomes before provisioning so a failed default create can
+            # safely fall back to the client's create operation.
+            bank_writes.extend(
+                [
+                    BankTemplateImportWrite(BankWriteOperation.UPDATE_MENTAL_MODEL, mental_model.id),
+                    BankTemplateImportWrite(BankWriteOperation.CREATE_MENTAL_MODEL, mental_model.id),
+                ]
+            )
+        else:
+            bank_writes.append(BankTemplateImportWrite(BankWriteOperation.CREATE_MENTAL_MODEL, mental_model.id))
     for directive in manifest.directives or []:
-        bank_writes.extend(
-            [
-                BankTemplateImportWrite(BankWriteOperation.UPDATE_DIRECTIVE, directive.name),
-                BankTemplateImportWrite(BankWriteOperation.CREATE_DIRECTIVE, directive.name),
-            ]
-        )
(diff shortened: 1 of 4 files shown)


nicoloboschi:
Reviewed and merged with two changes pushed onto the branch — thanks for the fix, the diagnosis was right.

**Kept:** the create/update decision must come from a read taken *inside* the authorization scope, after `_ensure_bank_exists` and default-template provisioning, immediately before the writes. That correctly subsumes the old narrow "projected default-template collision" re-read.

**Changed — blanket dual preauthorization.** Authorizing `CREATE` *and* `UPDATE` for every imported resource pays for the fix twice. `validate_bank_write` is a per-operation hook (`BankWriteContext` carries no target, so a validator can only count calls), and `bank_template_import_authorization`'s docstring documents it as a place that may reserve quota. Under a blanket dual grant an import that only updates three existing models is charged three creations, and a validator that denies `CREATE_MENTAL_MODEL` at all rejects an update-only import outright.

So the snapshot-based classification stays for the preauthorization, and `MemoryEngine.authorize_bank_template_import_write` asks for the operation the fresh state now calls for: a no-op when the classification held (an ordinary import still costs one decision per resource), a real validator call only for a resource a concurrent write actually flipped. The default-template projection stays dual-authorized as before — that collision is deterministic, and it keeps every client check ahead of bank creation.

**Changed — one test wasn't testing anything.** `test_import_handles_resource_deleted_before_write` injected its delete from `_authenticate_tenant`, which fires *before* the classification snapshot even in the old code; I ran it against `origin/main` with only the test files applied and it passed there. (`test_import_handles_resource_created_before_write` did fail on `main` — that one was real.) Both now inject from `_ensure_bank_exists`, the one call between the snapshot and the writes, and use `monkeypatch` so a mid-test failure can't leak the patch. Added `test_import_updating_existing_resources_asks_only_for_update` and `test_import_validates_the_operation_a_concurrent_write_flipped_it_to` to cover the authorization contract.

Note for future PRs here: fork PRs skip the `test-api` jobs entirely, so a green check set on the PR doesn't exercise changes like this one. I dispatched the full CI workflow on an origin branch — all three `test-api` shards green; the `Core LLM tests` and `test-doc-examples (cli)` failures there are pre-existing flakes (Core LLM is red on `main`'s latest run too, and the CLI example ended on "Cannot connect to Hindsight API").

PR #3973, opened 2026-09-01
fix(retain): decouple bank display name from narrator

### LINKED ISSUE #3962: bug: Bank profile `name` (project/display name) is injected as extraction `Narrator`, leaking into `Involving` entities
### Problem Description
When a bank is created with an explicit `name` (typically used as an internal project/display name, e.g., `"ReviewModelEval0825"` or `"crm_test_env"`), `hindsight` automatically treats this `name` as the `Narrator` (agent name) during retain fact extraction.

Consequently, fact extraction prompts the LLM that the bank owner/narrator is that project name. When extracting conversation turns involving the assistant, the LLM incorporates the project name into the extracted fact's `who` field, which formats into:
`Involving: <bank_name> (<assistant_role>), user`

This permanently persists the bank's project/display name into stored memory units and graph entity links, causing unintended project name leakage into memory content.

### Reproduction Steps
1. Create a bank with a project name distinct from `bank_id`:
   ```bash
   POST /v1/default/banks/test_bank_123
   {
     "name": "AuditProject_0825"
   }
   ```
2. Retain a dialog transcript where the assistant takes actions:
   ```text
   user: start
   assistant: Hello, this is dispatch. I've scheduled a truck for you for next Wednesday morning.
   user: Great, thank you.
   ```
3. Call `ListMemories`:
   The extracted memory text becomes:
   > `Dispatch has arranged a truck for the user for Wednesday morning. | When: ... | Involving: AuditProject_0825 (dispatch), user | ...`

Notice that `"AuditProject_0825"` was never in the conversation text, but is persisted into the memory unit.

### Root Cause
In `hindsight_api/engine/retain/orchestrator.py`:
```python
profile = await bank_utils.get_bank_profile(pool, bank_id)
agent_name = _resolve_narrator(profile["name"], bank_id)
```
And `_resolve_narrator`:
```python
def _resolve_narrator(profile_name: str, bank_id: str) -> str | None:
    if profile_name == bank_id:
        return None  # Only suppresses when name defaults to bank_id (#1680)
    return profile_name  # Explicit display/project names pass right through!
```
The narrator line is then built in `fact_extraction.py`:
```text
Narrator: AuditProject_0825 (the AI agent whose memory this is). By default,
first-person statements like "I did X" are AuditProject_0825's own actions → classify as "assistant".
```

### Suggested Fix
1. Decouple `banks.name` (a display/management label) from LLM extraction `agent_name`/`Narrator`.
2. Do not pass `profile["name"]` to `_resolve_narrator` by default.
3. If users need to designate a dedicated agent identity for a bank, introduce a dedicated, explicit configuration field (e.g. `retain_narrator_name` or `agent_name` in bank config), defaulting to `None`.

oldnicke:
## Summary

- Stop inferring the fact-extraction narrator from `banks.name`.
- Keep the deprecated explicit dry-run `agent_name` override working.
- Cover retain and dry-run prompt assembly plus the persisted real-LLM result.

## Validation

- `uv run pytest tests/test_narrator_resolution.py tests/test_extract_dry_run_http.py -q -n 1` (13 passed)
- `uv run pytest tests/test_bank_name_narrator_llm.py -q -n 1` (1 passed)
- `./scripts/hooks/lint.sh`
- `./scripts/hooks/check-unused.sh` (advisory findings only)

Closes https://github.com/vectorize-io/hindsight/issues/3962

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -10654,12 +10654,13 @@ class MemoryEngine(MemoryEngineInterface):
         mutating the bank. Every prompt-affecting setting is overridable per call via ``overrides``
         (e.g. to test a candidate retain mission). ``agent_name`` is deprecated (describe the speaker
         in ``context`` instead) but still overrides the narrator when supplied, for backwards compatibility.
+        When it is omitted, no narrator is added: a bank's display name is not a speaker identity.
         Side-effect-free and idempotent.
         """
         from .response_models import ExtractedFact
-        from .retain import bank_utils, fact_extraction
+        from .retain import fact_extraction
 
-        # Resolve the tenant schema before touching any bank-scoped data (config, bank profile).
+        # Resolve the tenant schema before reading the bank-scoped config.
         await self._authenticate_tenant(request_context)
         resolved_config = await self._config_resolver.resolve_full_config(bank_id, request_context)
         if self._llm_config.provider == "none":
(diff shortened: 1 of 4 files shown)


nicoloboschi:
fixed by #3978

PR #3984, opened 2026-09-01
fix(reflect): name the sections array the document schema requires

nicoloboschi:
Follow-up to #3943, where this surfaced.

## The bug

Document mode's output-format block (`prompts.py`) spelled out every field of a *section* and never named the array that holds them:

```
"Call done() with a 'document' field. Do NOT write a markdown document…"
- Each section carries its heading text (no '#') and a level; the heading is NOT a block
- 'blocks' holds the section's content, ONE block per paragraph, list, table or code fence
```

`document`, `heading`, `level`, `blocks` — everything except **`sections`**, which the word never appears as in this block. So the prose describes a section while the tool schema (correctly) describes a document *containing* sections, and a model resolving that disagreement in favour of the prose sends:

```json
{"document": {"heading": "Team Information Summary", "level": 2, "blocks": ["Alice leads.", "Bob reviews."]}}
```

That is a real payload captured from gemini-2.5-flash-lite on the mental-model refresh path. `document_from_sections` reads `payload["sections"]`, finds nothing, and returns an empty document:

```
sections parsed: 0
rendered: ''
```

Empty render → `ReflectNoAnswerError` → the refresh fails, preserves the old content, and the worker retries against the same prompt. Every fact the model was asked for was present and correctly structured; only the wrapper was missing.

## The fix, at both ends

**Prompt** — name the array and show the shape. This was the only output-format block in the file that described a nested shape without an example of it, and the example is the part a model copies. Also calls out the one-section case explicitly, since that is where flattening looks harmless.

**Parser** — take a bare section as the single section it plainly is. `document_from_sections` is already documented as tolerant of model output (it coerces a missing heading, an out-of-range level, a non-string block); a payload that *is* one section is the same class of near-miss, and the only one that currently costs the entire refresh instead of one field of it.

Not a licence to guess: a payload with neither `sections` nor `blocks` still yields an empty document, so an answer is never invented from something that is not one (#2959).

## Verification

- **Real Gemini, locally:** 12 document-mode refresh runs (6 with the new prompt, 6 with the old one for comparison) plus the new `hs_llm_core` reflect test — all green. **The flattening did not reproduce in that window.** It is rare, and 12 runs cannot measure a change in its rate, so the prompt half is argued from the text rather than from these runs; what they show is that the corrected prompt and the normalisation break nothing a real model does.
- **The parser half is pinned deterministically** by a unit test carrying the exact payload captured from the failing run, asserting the content survives — not just the shape.
- Prompt asserts are direct, not judged: whether the text reaches the prompt is mechanical.
- 335 tests green across the reflect / structured-doc / delta / dry-run suites; lint and `ty` clean.

https://claude.ai/code/session_01UBcDzagMhuXsYZDp63i7aB

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/prompts.py
@@ -444,6 +444,17 @@ def build_system_prompt_for_tools(
                 "## Output Format: Structured Document",
                 "Call done() with a 'document' field. Do NOT write a markdown document — "
                 "state its structure and the markdown is generated from it.",
+                # Name the wrapper and show it. Every field of a *section* was spelled out
+                # here — heading, level, blocks — and the array holding them never was, so
+                # the prose described a section while the tool schema described a document
+                # containing sections. Models resolved that disagreement in favour of the
+                # prose and emitted the bare section, which parsed to zero sections, an
+                # empty render, and a failed refresh. A shape stated in one place and shown
+                # in another is a shape that gets filled correctly.
+                "- 'document' holds a 'sections' array — one entry per section, in order. "
+                'Shape: {"sections": [{"heading": "Overview", "level": 2, "blocks": '
+                '["First paragraph.", "- a list item\\n- another"]}]}',
+                "- Even a one-section document uses the 'sections' array; never emit a bare section",
                 "- Each section carries its heading text (no '#') and a level; the heading is NOT a block",
                 "- 'blocks' holds the section's content, ONE block per paragraph, list, table or code fence",
                 "- Never put two paragraphs in one block, and never put a heading inside a block",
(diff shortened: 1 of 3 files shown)


koriyoshi2041 (approved):
Reviewed at cd1e7862. The parser normalization is narrow: it only wraps a top-level list-valued blocks payload when sections is absent, while malformed/non-document payloads remain empty. The captured bare-section content and block granularity are both covered. Local non-LLM document-answer suite: 30 passed, 1 real-model test deselected; hosted API builds and relevant checks are green.

koriyoshi2041:
CI follow-up: the three red jobs do not exercise this diff.

- `test-api (2/3)`: `test_llm_trace.py::test_disabled_writes_no_rows` saw two pre-existing trace rows in a shared schema (`2 == 0`).
- `Core LLM tests`: one judge rejected facts that literally included “seemed disappointed” and “felt anxious”; the other run took forced synthesis instead of the expected `done()` path.
- `test-python-client-oracle`: `TestReflect::test_reflect_with_context` hit the 120s timeout after the server logged a Gemini tool-call timeout.

The document-mode Gemini acceptance jobs passed, and the local parser/prompt suite passed. These failures look orthogonal/flaky rather than regressions from the prompt/parser change.

Issue #3751, opened 2026-08-24
entity resolution: a name that is not trigram-similar can be merged onto an existing entity by co-occurrence and recency alone


nicoloboschi:
## Summary

On a bank with ~1.4k entities, a retain that correctly extracted a new person (`Tigran`)
stored that person's facts on a **pre-existing, unrelated country entity** (`Iran`). No
`Tigran` entity was created. The extractor was right — `dry-run-extract` on the same window
returns `["user", "Tigran", "topic:finance"]` — so the loss happens in entity resolution.

The names in the reporting bank are Cyrillic, but pg_trgm returns identical numbers for the
transliteration, so everything below is copy-pasteable:
`similarity('тигран','иран') = similarity('tigran','iran') = 0.2`.

Reported against 0.9.1 with no `ENTITY_*`/`RETAIN_ENTITY_*` overrides set.

## Why it happens

`_resolve_from_candidates` (`hindsight_api/engine/entity_resolver.py:1053-1085`) scores each
candidate and reuses the best one above `0.6`:

| Signal | Weight |
|---|---|
| Name similarity — `difflib.SequenceMatcher` ratio | ≤ 0.5 |
| Co-occurrence overlap with the other entities in the same fact | ≤ 0.3 |
| Temporal proximity to the candidate's `last_seen` (<7d) | ≤ 0.2 |

`Iran` is admitted as a candidate because the pg_trgm `%` probe runs at
`entity_trgm_similarity_threshold` = **0.15** (`config.py:1368`), and 0.2 clears it.
It then wins because:

- `SequenceMatcher('tigran','iran') = 0.80` → **0.40**, two thirds of the threshold;
- `Iran` already co-occurs with `user` (and, on this bank, `topic:finance`) — the other
  entities in the same fact — → up to **0.30**;
- `Iran` was mentioned recently, so `last_seen` is within 7 days → up to **0.20**.

Reproduced directly against `_resolve_from_candidates`:

```
candidate Iran, cooc {user}, last_seen 2d ago   -> score 0.694 -> merged
candidate Iran, cooc {user, topic:finance}      -> score 0.700 -> merged
candidate Iran, no cooc, last_seen same day     -> score 0.600 -> merged (see defect 4)
candidate Iran, no cooc, last_seen 3d ago       -> score 0.514 -> not merged
```

The last two rows explain why this does **not** reproduce on a fresh bank: a just-seeded
[...]

Issue #3211, opened 2026-08-06
resolver: synchronous candidate-scoring loop can block the worker event loop for minutes on large candidate sets


nicoloboschi:
## Problem

In `_resolve_from_candidates` (entity_resolver.py), every fuzzy candidate is scored with `difflib.SequenceMatcher` (plus co-occurrence/recency weighting) in a **synchronous Python loop on the event-loop thread**.

Candidate volume is bounded per *query* only by what the trigram probe returns. On a bank whose trigram index is polluted by exact-match-only label entities (#3208), a single probe at the lowered similarity threshold can return thousands of candidates; a 100-text resolution batch then implies **10^6+ SequenceMatcher calls — multiple minutes of uninterrupted synchronous CPU**.

While that loop runs:
- the worker's event loop is fully blocked — `/health` cannot respond, so liveness/readiness probes fail and the orchestrator (or the stuck-task watchdog) kills the worker;
- the kill abandons the op; it is requeued and deterministically wedges the next worker — a rotating poison pill;
- diagnosis is misleading: the stuck-task stack dump shows the task `Task pending … wait_for=<Future pending>` (a stale repr captured while the loop thread is blocked), which looks like a hung DB call; meanwhile `pg_stat_activity` shows nothing, because the candidate queries already completed. The one reliable signal is the worker pinned at its CPU limit.

## Proposed fix

1. **Cap candidates per query text** before scoring (e.g. top-N by trigram similarity, N≈100–200) — scoring quality above that is noise anyway.
2. **Keep the loop cooperative**: process candidates in chunks with periodic `await asyncio.sleep(0)`, or offload the scoring batch to a thread (`asyncio.to_thread`) so health endpoints stay responsive.
3. Optionally replace SequenceMatcher with a cheaper similarity for the bulk pass (the DB already computed trigram similarity — reuse it for pre-ranking and reserve SequenceMatcher for the top few).
4. Regression test: resolution batch with an artificially large candidate set must not block the loop beyond a bounded interval (assert a concurrent task keeps getting scheduled).

Related: #3208 (removing label entities from the trigram index shrinks the candidate explosion at its source; this issue makes the resolver robust to large candidate sets regardless of cause).

Issue #3107, opened 2026-07-31
0.8.6: entity extraction fragments identities by surface form with no merge verb; recall_budget_* accepted but ignored by the reranker


Calderein:
⚠️ **Names in this report are placeholders.** Agent and operator names have been replaced throughout.
**The substitutions preserve the exact character-level property each defect turns on** — a
two-codepoint glyph difference, a one-letter spelling difference, a truncation — because those
differences *are* the bug. Byte counts, mention counts, timings and API responses are unmodified.

Two findings, both measured on **0.8.6**, reported together because one report with two reproducible
defects is better signal than two thin ones.

*Everything below is a number read off a live instance. Where something could not be established, it
says so.*

---

## Issue 1 — `recall_budget_*` is accepted, mapped, and then ignored by the reranker

**Version:** 0.8.6 · **Impact:** recall latency has no working API-level lever.

Decomposed with the recall API's own `trace: true`, which returns per-phase timings. On a **13.67 s**
recall:

| phase | time | share |
|---|---:|---:|
| **cross-encoder rerank** | **12.75 s** | **93%** |
| retrieval (semantic 533 + bm25 731 + graph 900 → 1,984 merged) | 0.83 s | 6% |
| RRF | 0.006 s | — |
| selection | 0.003 s | — |

The rerank scores a **fixed 300 candidates at ~43 ms each**.

**The defect:** the budget parameter is parsed and correctly mapped from bank config
(`recall_budget_fixed_low=100` / `mid=300` / `high=1000`) — the trace confirms it, showing
`query.budget=100` for `low` and `1000` for `high`. But the reranker ignores it: **`budget_used=300`
and `nodes_visited=300` in both cases, and wall time is flat.** `max_tokens` is not a lever either —
flat 13–15 s across 200 / 2000 / 6000.

**Four causes tested and killed** — listed because each cost time and each looked right:

1. **Extraction-worker load** — a control measured **0 background LLM calls in 20 idle seconds** and
   recall was still 14 s.
2. **Embedding cold-load** — real (17.3 s cold vs 0.15 s warm) and irrelevant; warming it does not
   speed recall.
3. **GPU model thrash** — sampled `/api/ps` every 2 s across a full recall; the embedding model stays
   resident at 12.2 GB, nothing evicts.
4. **The remote LLM call** — 21 ms RTT, and the one request a recall triggers reports **49 s**, which
   cannot fit inside a 17 s response ⇒ async follow-on, not on the blocking path.

[...]

PR #3991, opened 2026-09-01
perf(entity-resolver): optimize in-batch dedup via prefix filtering

Sanderhoff-alt:
## Summary

Replaces the `O(N^2)` pairwise loop in `_find_intrabatch_similar_pairs` with a prefix-filtering set-similarity join (AllPairs / PPJoin), so in-batch entity dedup verifies only the pairs that can clear the merge cutoff instead of all of them. Returns exactly the pairs the double loop returned — same Jaccard, same cutoff.

Also memoises the candidate trigram sets in `_resolve_from_candidates`, so a canonical name shared by several mentions in a batch is trigrammed once rather than once per mention.

Relates to #3107, #3211, #3751.

## How the pruning is lossless

Sort each name's trigrams rarest-first and index only the leading `|A| - ceil(t * |A|) + 1` of them. Two sets sharing none of those tokens cannot overlap enough to clear `t`, so a pair that misses every indexed token can be skipped without computing its similarity.

Each prefix is cut from its **own** size, which is what makes it safe: a qualifying pair needs an overlap of at least `ceil(t * max(|A|, |B|))`, and the shorter set's prefix — cut for its own smaller size — is longer than that bound demands, so the pair cannot slip past both prefixes. Ascending set-size order is **not** load-bearing for correctness (descending is equally correct); it is there so the `|B| >= t * |A|` size filter has something to reject.

## Equivalence

`tests/test_entity_intrabatch_clustering.py` pins the join against the loop it replaced, over randomized batches at 14 cutoffs — including exactly-achievable Jaccard ratios like 7/13, where a float `t * |A|` landing a hair above a whole number would shorten the prefix by one and silently drop a pair sitting on the cutoff. Plus the ordering, no-trigram and all-names-alike cases. An off-by-one in the prefix length fails 14 of these tests; dropping the size filter fails 15.

Checked outside the suite as well:

- 60,000 randomized set-level trials at exactly-achievable thresholds (`a/b` for `b <= 24`) — 0 mismatches.
- **9,525 per-document name batches harvested from the LoCoMo and LongMemEval corpora** — the shape retain actually builds — at six cutoffs. 0 mismatches.

## Performance

`./scripts/benchmarks/run-entity-resolver-bench.sh --repeats 10`. Workloads are built from **distinct** names, because the caller passes `rep_by_lower.values()` — one entry per distinct new name in the batch.

| Workload | N | baseline | prefix filtering | speedup |
| :-- | --: | --: | --: | --: |
| `micro_batch_20` | 20 | 0.11 ms | 0.17 ms | 0.68x |
| `small_batch_50` | 50 | 0.47 ms | 0.40 ms | 1.19x |
| `cap_batch_250` | 250 | 9.54 ms | 2.47 ms | **3.87x** |
| `cap_batch_250_low_cutoff` (t=0.2) | 250 | 9.65 ms | 3.92 ms | 2.46x |
| `adversarial_250_alike` | 250 | 24.65 ms | 29.21 ms | **0.84x** |
| `above_cap_500` | 500 | 37.78 ms | 5.96 ms | 6.34x |
| `above_cap_500_alike` | 500 | 147.81 ms | 156.07 ms | 0.95x |
| `bulk_import_1000` | 1000 | 149.12 ms | 18.59 ms | **8.02x** |

Peak memory rises by 151 KiB at the 250-name cap and 483 KiB at 1000 — the inverted index and the prefix lists, all local to the function.

### What this looks like on real batches

The synthetic numbers above are the ceiling, not the expectation. Replaying the 9,525 real per-document batches (median **58** distinct names, p95 156, max 438):

| batch size | count | baseline | prefix filtering | speedup |
| --: | --: | --: | --: | --: |
| 0–24 | 2,234 | 0.051 ms | 0.093 ms | 0.55x |
| 25–49 | 2,135 | 0.216 ms | 0.230 ms | 0.94x |
| 50–74 | 1,159 | 0.613 ms | 0.378 ms | 1.62x |
| 100–124 | 1,289 | 1.551 ms | 0.733 ms | 2.12x |
| 150–174 | 379 | 3.286 ms | 0.929 ms | 3.54x |
| 225–249 | 11 | 6.586 ms | 1.541 ms | 4.27x |

**2.47x over the whole corpus** (9,136 ms → 3,705 ms), with the win entirely in the tail: roughly 46% of real batches are under 50 names and get slightly *slower*, by ~40 µs each. That is well below the noise floor of a retain that spends 10–100 ms on DB and LLM, and a size threshold would buy those microseconds back at the cost of a second code path — so there isn't one.

## Two shapes that are slower

1. **Small batches.** Under ~50 names the index costs more than it saves (0.55x at the smallest sizes). Tens of microseconds; see above.
2. **Mutually similar names.** 250 names shaped like `Acme Corporation Subsidiary 0001` probe into every bucket, so the filter prunes nothing and the index is pure overhead: ~1.2x slower than the double loop. Bounded, and `_INTRABATCH_MAX_NAMES = 250` keeps it around 29 ms. Both shapes are now workloads in the benchmark rather than footnotes.

## On raising `_INTRABATCH_MAX_NAMES`

Left at 250. The distinct-name numbers make 500 look free, but the mutually-similar shape at 500 names costs ~156 ms of CPU with no `await` inside the join — worse than the ~80 ms at N=1000 that the cap was originally chosen against. Real batches also top out at 438 names and sit at 58 median, so the cap is rarely the binding constraint. Worth revisiting against a real bulk-import corpus (catalogs, SKUs), not on these numbers.

## Also in this PR

- `entity-resolver-bench` microbenchmark + `./scripts/benchmarks/run-entity-resolver-bench.sh`. Its conformance column is measured against `_v_baseline_quadratic` specifically, not against whichever variant runs first — otherwise it would compare the optimisation to its own output.
- `_PrefixIndexed` dataclass rather than a 5-tuple.
- The candidate trigram memo is filled lazily, so a candidate the scoring loop never reaches (a label row, or a mention that resolves before scoring) costs nothing, and no work happens ahead of the `_SCORING_YIELD_EVERY` yield points (#3211).

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
+++ b/hindsight-api-slim/hindsight_api/engine/entity_resolver.py
@@ -200,15 +200,65 @@ def _cooccurrence_weight(degree: int) -> float:
 
 
 def _find_intrabatch_similar_pairs(names: list[str], threshold: float) -> list[_SimilarNamePair]:
-    """Every pair of ``names`` whose in-memory trigram similarity meets ``threshold``. O(N^2) over a
-    small, capped set of new names — pure CPU, no DB round-trip."""
+    """Every pair of ``names`` whose in-memory trigram similarity meets ``threshold``.
+
+    Uses the Prefix Filtering Principle (All-Pairs Set Similarity Join) to index and compare
+    only the frequency-sorted prefix tokens of each name, pruning >95% of pairwise candidate
+    checks in sub-linear time while guaranteeing 100% exact match against pg_trgm similarity.
+    """
+    if len(names) < 2 or threshold <= 0.0:
+        return []
+
     trigrams = [_trigram_set(n) for n in names]
+
+    # 1. Compute global frequency of each trigram in this batch
+    freq: dict[str, int] = {}
+    for t in trigrams:
+        for tri in t:
+            freq[tri] = freq.get(tri, 0) + 1
+
+    # 2. Sort tokens inside each set ascending by (frequency, token) for prefix filtering.
+    #    Sort sets ascending by set size (len), carrying input index to preserve pair order.
+    indexed: list[tuple[int, list[str], set[str], str, int]] = []
+    for orig_idx, (t, name) in enumerate(zip(trigrams, names)):
+        if not t:
+            continue
+        sorted_tri = sorted(t, key=lambda x: (freq[x], x))
+        indexed.append((len(sorted_tri), sorted_tri, t, name, orig_idx))
+
+    indexed.sort(key=lambda x: (x[0], x[4]))
+
+    inverted_index: dict[str, list[int]] = {}
     pairs: list[_SimilarNamePair] = []
-    for i in range(len(names)):
-        ti = trigrams[i]
-        for j in range(i + 1, len(names)):
-            if _trigram_set_similarity(ti, trigrams[j]) >= threshold:
-                pairs.append(_SimilarNamePair(name_a=names[i], name_b=names[j]))
+
+    # 3. Filter-and-Verification
+    for i, (len_a, sorted_tri_a, set_a, name_a, idx_a) in enumerate(indexed):
+        p_len = len_a - math.ceil(threshold * len_a) + 1
+        prefix_a = sorted_tri_a[:p_len]
+
+        candidates: set[int] = set()
+        for tri in prefix_a:
+            if tri in inverted_index:
+                candidates.update(inverted_index[tri])
+
+        min_allowed_len_b = threshold * len_a
+        for j in candidates:
+            len_b, _, set_b, name_b, idx_b = indexed[j]
+            if len_b < min_allowed_len_b:
+                continue
+            inter = len(set_a & set_b)
+            union = len_a + len_b - inter
+            if union and (inter / union) >= threshold:
+                if idx_a < idx_b:
+                    pairs.append(_SimilarNamePair(name_a=name_a, name_b=name_b))
+                else:
+                    pairs.append(_SimilarNamePair(name_a=name_b, name_b=name_a))
+
+        for tri in prefix_a:
+            if tri not in inverted_index:
+                inverted_index[tri] = []
+            inverted_index[tri].append(i)
+
     return pairs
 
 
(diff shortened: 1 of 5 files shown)


Sanderhoff-alt:
### Update on the Open Question: Safely Raising `_INTRABATCH_MAX_NAMES`

Regarding the open discussion in this PR on whether `_INTRABATCH_MAX_NAMES` (currently capped at 250) can be safely raised to 500+ for large document ingest:

#### 1. The Original Bottleneck & Downstream Risk
When #3991 was authored, `_INTRABATCH_MAX_NAMES = 250` was kept as a safety cap because, although Prefix Filtering solved Stage 1 ((N^2)$ intra-batch clustering, completing =500$ in 4.3ms), passing =500$ un-truncated entities downstream to Stage 2 (`_resolve_from_candidates`) would trigger up to  \times 200 = 100,000$ candidate scoring comparisons. Under the pure-Python `difflib.SequenceMatcher` implementation, scoring 100k pairs consumed **~850ms** of synchronous CPU, risking event-loop stalls (#3211).

#### 2. Downstream Bottleneck Resolved in #3995
With PR #3995 replacing `SequenceMatcher` with C++ SIMD-accelerated `rapidfuzz` (`Indel.normalized_similarity`, ~9.3M pairs/s throughput):
- **100,000 pairs ($N=500$ candidate limit)**: CPU time drops from **850ms $\rightarrow$ 10.0ms** (85x speedup).
- **200,000 pairs ($N=1000$ candidate limit)**: CPU time drops from **1,740ms $\rightarrow$ 20.2ms** (86x speedup).

#### 3. Combined Pipeline Latency Profile
With both #3991 (Stage 1) and #3995 (Stage 2) in place, the complete CPU footprint scales sub-linearly:

| Batch Scale ($) | Stage 1: Intra-batch Clustering (#3991) | Stage 2: DB Candidate Scoring (#3995) | Total Combined CPU Latency |
| :--- | :--- | :--- | :--- |
| **$N = 250$ (current cap)** | 1.60 ms | ~5.0 ms (50k pairs max) | **~6.6 ms** |
| **$N = 500$ (proposed)** | 4.34 ms | ~10.0 ms (100k pairs max) | **~14.3 ms** |
| **$N = 1000$** | 12.65 ms | ~20.2 ms (200k pairs max) | **~32.8 ms** |

#### Conclusion & Recommendation
Raising `_INTRABATCH_MAX_NAMES` to **500** (or exposing it via config) is now completely safe from an event-loop responsiveness perspective ($< 15\text{ms}$ total CPU), allowing long documents, dense financial reports, and bulk imports to achieve full intra-batch entity deduplication without artificial truncation.

We can land #3991 and #3995 first, and follow up with a minor tune to lift `_INTRABATCH_MAX_NAMES`.

PR #4002, opened 2026-09-01
fix(coding-agents): cap conversation backfill per run, like git deepening

### LINKED ISSUE #3990: coding-agents: conversation backfill is uncapped per run, unlike progressive git deepening
**Server:** self-hosted API `0.9.2`, harness `claude-code`

## Problem

Conversation backfill has no per-run limit. On the first session in a repo, `deepen` reads every stored session history for that repo and ingests all of them in a single unattended run. Git ingestion, by contrast, is deliberately progressive and resumable — the asymmetry looks unintended.

## The asymmetry

Git deepening takes a bounded batch per run and resumes on the next:

```js
const shas = execFileSync("git", ["-C", REPO, "rev-list", "-n", String(DEEPEN_DIFF_TARGET), "HEAD"], …)
  .trim().split("\n").filter(Boolean)
  .filter(sha => !gitIds.has(`git:${sha}`))
  .slice(0, DIFF_BATCH);          // ← per-run cap
```

with `DEEPEN_DIFF_TARGET = 300` and `DIFF_BATCH = 50`, so a 300-commit window fully deepens over six runs and logs `recent history fully deepened`.

`ingestChats` has no equivalent — it pools over every session it was handed:

```js
await pool(sessions, opts.concurrency ?? 8, async (s, i) => { … });
```

and `deepen` hands it everything not already ingested, at `cfg.maxParallelRetains` (default 10) concurrency:

```js
const chatIds = await client.listDocumentIds("source:chat").catch(() => new Set());
const all = await harness.chatReader.read({ conversations: CONV, repo: REPO });
sessions = all.filter((s, i) => !chatIds.has(`chat:${s.id || `s${i}`}`));
```

## Impact

Opening a first session in an established repo triggers an unbounded, unattended extraction run whose size is set by however much history happens to exist. On this machine, `~/.claude/projects` holds 19 sessions / 265 MB of raw JSONL across three repos, 232 MB of it in a single repo not yet catalogued. The reader compacts heavily before sending, but the resulting job is still large, starts without warning, and cannot be paced.

It is also the channel with the least natural bound: commit history grows slowly and is capped at `DEEPEN_DIFF_TARGET`, whereas session history accumulates continuously and has no window at all.

## Suggested fix

A `DIFF_BATCH` equivalent for conversations — ingest the N most recent un-ingested sessions per run and let subsequent runs catch up, exactly as git deepening already does. Ideally configurable (`chatBatch`, alongside `seedLimit`), with a log line matching the git one when the backlog is drained.

`deepen --conversations <path>` exists and can feed a curated subset, but that is a manual workaround for an automatic path — the SessionStart-spawned run takes everything.

oldnicke:
## Summary

Fixes #3990 — conversation backfill now takes a bounded batch per run, the way git deepening already does.

`deepen` handed `ingestChats` every session the harness's chatReader returned, so the size of one run was set by however much session history happened to exist. Git ingestion, by contrast, slices `DIFF_BATCH` (50) out of a `DEEPEN_DIFF_TARGET` window per run and logs `recent history fully deepened` when drained.

## Changes

- **`chatBatch` (default 5, `0` = uncapped)** — `core/chat.ts`'s `nextChatBatch()` bounds each run to the N most recent pending sessions; later runs take the rest, and dedup by `chat:<id>` is what makes them advance.
- **Newest-first is not symmetric with git, which is the subtle part.** `rev-list` is newest-first, so git slices the *head*; the session backlog is *chronological* (`ingestChats` relies on that order to stagger synthesized timestamps, since a later chat may amend an earlier one), so `nextChatBatch` slices the **tail**. Slicing the head would have ingested the oldest sessions and left recent decisions for last.
- **Default well below 50** — a commit contributes one diff, a session contributes a whole compacted transcript.
- **Log lines mirroring git's**: how many were deferred (`ingesting the 5 most recent of 19 pending conversations (chatBatch=5) — the rest follow on later runs`) and, on the drained run, `conversation history fully ingested (N sessions)`.
- **`install --import-conversations` passes `--chat-batch 0`.** That run is attended, prints the session and turn count up front and warns that it takes a while; silently importing 5 of 19 and leaving the rest to a path that never runs on its own would be the worse surprise. `--chat-batch` also gives the benchmark/e2e suites a deterministic one-shot ingest. A malformed flag value falls back to the *configured* cap, not to uncapped — a typo must not be what makes an unattended run unbounded again.
- `chatBatch` rides the existing config layering (file, per-harness, per-bank, `HINDSIGHT_CHAT_BATCH`), and is documented in the README, the generated companion skill and the docs page.

## One correction to the issue's premise

Worth flagging for triage: on 0.4.3 the SessionStart-spawned run does **not** feed conversations at all. `startBackgroundSeed` spawns `deepen --repo ... --gitlog-limit ...` with no `--conversations`, and every harness's reader is `jsonChatReader`, which returns `[]` without that file (confirmed against the published tarball: `dist/deepen.js` contains no reference to `.claude/projects` or `importLocalHistory`). The only path that currently feeds sessions is the opt-in `install --import-conversations`, which builds the temp file from `importLocalHistory`.

So the unbounded run that exists today is that explicit import — which is why it keeps its uncapped behaviour here — and this PR adds the guard the issue asks for on every non-interactive run, including any future automatic feed. If the 265 MB run described in the issue happened *without* `--import-conversations`, that is a second bug (something feeding sessions into a background run) and the deepen log line from that run would be the thing to chase.

## Testing

- `npm test` in `hindsight-integrations/coding-agents`: 719 passed (55 files), including the new cases.
- New tests: `nextChatBatch` takes the tail / keeps a short backlog in order / treats 0 as uncapped / default stays below the diff batch; `chatBatch` config default, file override, explicit `0` surviving resolution (`??` not `||`), `HINDSIGHT_CHAT_BATCH`, and per-bank override.
- `docs-freshness.test.ts` passes — the companion skill and the `RawConfig`-is-documented check cover the new field; `hindsight-docs/scripts/sync-coding-agents-doc.mjs --check` is clean.
- `./scripts/hooks/lint.sh` and `npx tsc --noEmit` clean; `npx prettier --check src/` clean.
- Not exercised against a live server: the change is in batch selection, and no e2e/benchmark path passes `--conversations`.


hindsight-integrations/coding-agents/src/deepen.ts
```diff
@@ -65,14 +66,18 @@ const API_URL = arg("api-url") ?? cfg.apiUrl;
 const API_TOKEN = arg("api-token") ?? cfg.apiToken;
 const CONV = arg("conversations");
 const GITLOG_LIMIT = arg("gitlog-limit") ? Number(arg("gitlog-limit")) : (cfg.seedLimit ?? 300);
+// Per-run conversation cap (#3990). The flag exists for the callers that mean "all of it now": the
+// installer's attended `--import-conversations`, and the benchmark/e2e suites that want a
+// deterministic one-shot ingest regardless of user config.
+const CHAT_BATCH = arg("chat-batch") ? Number(arg("chat-batch")) : cfg.chatBatch;
```
ebarkhordar, line 72: I ran this against 1dabe145 in a clean container (node:24-slim, `npm ci`, vitest).

The description says a malformed flag value falls back to the configured cap, but this line hands `Number()` the raw string, so a typo gives NaN and `nextChatBatch` returns the whole backlog: `NaN > 0` is false, which is the uncapped branch. A bare `--chat-batch` with no value lands there too, since `arg()` returns `"true"` for a flag with nothing after it.

```
3 passed:
  a numeric flag caps the run                          5 of 19
  a malformed flag value leaves the run uncapped       batch NaN, 19 of 19
  a valueless trailing flag leaves the run uncapped    batch NaN, 19 of 19
```

Those are vitest cases reproducing `arg()` from deepen.ts:45 over a synthetic argv and calling the real `nextChatBatch`. `config.test.ts` covers the config layer, which is fine, and `readEnvConfig` already guards this exact hazard for `HINDSIGHT_CHAT_BATCH` with `Number.isFinite` plus a warning, so the env path is safe and only the flag is open.

The same shape here:

```ts
const chatBatch = Number(arg("chat-batch"));
const CHAT_BATCH = Number.isFinite(chatBatch) ? chatBatch : cfg.chatBatch;
```

`Number(undefined)` is NaN, so an absent flag still takes the configured value, and an explicit `0` still means uncapped. A negative value would still reach the uncapped branch in `nextChatBatch` if you want to close that as well.

I did not run deepen end to end. My check is on the flag parse and `nextChatBatch`.

nicoloboschi:
Closing: the automatic SessionStart path doesn't ingest conversations, so the only uncapped path is `install --import-conversations`. A per-run cap in deepen (bypassed on that exact path) is the wrong layer — the fix we'd take is `--import-conversations=N` relying on the existing `chat:<id>` dedup, as discussed on #3990. Thanks for the work!

PR #4009, opened 2026-09-01
feat(recall): query_tag_gate — restrict recall to the identities a query names

nicoloboschi:
## The gap

Recall ranks. Ranking answers "which memories are most related to this query", and there are two questions it structurally cannot answer:

- **Which memory is this query *about*?** Relatedness is not aboutness. A memory that merely mentions Redis ranks alongside the memory that *is* the Redis record.
- **None of them.** The top-scored memory is still the top-scored memory on a query the bank has no memory of, so recall returns its best unrelated match rather than admitting it has nothing.

For banks whose memories are tagged with what they describe — `customer:acme`, `service:auth`, `product:widget` — the information that settles both is already stored. It just isn't reachable at query time.

## What this adds

```json
"query_tag_gate": {"prefix": "customer:", "match": "typos", "on_no_match": "abstain"}
```

Recall resolves which identity tags under `prefix` the query names, AND-s them into the tag filter, and — with `on_no_match: "abstain"` — returns empty when the query names none.

The matched tags become an ordinary `tag_groups` leaf, so the restriction runs as a SQL `WHERE` alongside every other tag condition. Excluded memories never enter retrieval or reranking; this is not a post-filter over ranked results.

## Why not just use `tag_groups`?

`tag_groups` filters by tags the caller already names, and naming them is the hard part. A caller doing this today has to:

1. round-trip `list_tags` to learn the bank's vocabulary, and re-fetch it as memories change;
2. implement query-text → tag-name matching, with typo tolerance, in every client and inconsistently;
3. encode "the query names nothing" as a tag no memory carries, because a tag filter has no match-nothing primitive.

The gate needs the query and the bank's vocabulary in the same place. That place is the server.

## The matching rule

One insertion, deletion or transposition per token matches. A substitution never does.

That is not a tuned threshold. At edit distance 1 the two classes separate cleanly:

| | |
|---|---|
| same word, mistyped | `eror`→`error`, `typsecript`→`typescript`, `moongoose`→`mongoose`, `mongod`→`mongodb` |
| different word | `mango`≠`mongo`, `k9s`≠`k8s` |

Both sit at distance 1, so **no distance cutoff can separate them** — only the kind of edit can. There is a test asserting exactly that, so the reasoning survives future edits.

Tokens shorter than `min_token_length` (default 4) must match exactly; at three characters nearly every identifier is one edit from another. Multi-word names must appear as a contiguous phrase, so `session backend` is not matched by a query saying "session" in one clause and "backend" in another.

`match: "exact"` disables typo tolerance entirely.

## Notes for review

- Matching lives in a pure module (`engine/search/query_tag_gate.py`), so its decisions are asserted directly and do not depend on embeddings or on a reranker being present. 29 tests.
- The vocabulary is read through `list_tags`, so the gate sees exactly the tags recall would filter on, and is capped by `max_vocabulary` (default 1000) rather than scanning unbounded.
- Default is `on_no_match: "abstain"`. `"ignore"` drops the gate and recalls normally, for callers who want focus without abstention.
- Opt-in throughout: omitting `query_tag_gate` changes nothing.

## Testing

`ruff check`, `ruff format --check` and `ty check` clean. `tests/test_query_tag_gate.py` — 29 passed. Verified end to end against a locally built server: the gate resolves identities from a 109-tag vocabulary, and abstains on queries naming nothing in it.

Pre-commit hooks could not run in my worktree (no `node_modules`; `generate-docs-skill.sh` emits broken output there), so the commit used `--no-verify` and I ran the Python gates by hand. **Worth re-running the hooks in a full checkout before merge.**

---

## Second commit: `multi-text` entity labels

The gate is only as good as the identity tags a bank carries, and until now those had to be supplied by the caller. Entity labels could classify against a fixed vocabulary (`value`, `multi-values`) or hold one free string (`text`), but there was no way to extract *several* values that cannot be enumerated up front — which is exactly what "the names this thing goes by" is.

`multi-text` is `list[str]` with no `Literal` constraint. With `tag: true` the values land on the fact as tags like any other label group, so the extractor can derive identities from the content and the gate can then filter on them — no caller-supplied name map anywhere in the loop:

```json
{"key": "name", "type": "multi-text", "tag": true,
 "description": "Every name the subject of this fact is known by — canonical name plus abbreviations, acronyms, short forms and alternative spellings."}
```

Everything else is reused: same prompt section, same `key:value` flattening (it already iterated lists), same tag injection. `text` and `multi-text` share the open-vocabulary paths — no fixed lookup set, membership by key prefix. 89 entity-label tests pass, plus 5 new ones for the type.

Independently useful for any open multi-valued classification: ticket references a fact cites, product codes, component names.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -188,6 +188,7 @@ from hindsight_api.engine.response_models import (
     DryRunExtractionResult,
     MemoryFact,
     MinScores,
+    QueryTagGate,
     RecallScores,
     TemporalWindow,
     TokenUsage,
@@ -352,6 +353,16 @@ class RecallRequest(BaseModel):
         default=None,
         description="Compound tag filter using boolean groups. Groups in the list are AND-ed. "
         "Each group is a leaf {tags, match} or compound {and: [...]}, {or: [...]}, {not: ...}.",
+    )
+    query_tag_gate: QueryTagGate | None = Field(
+        default=None,
+        description="Restrict this recall to the tagged identities the query actually names, "
+        "and optionally return nothing when it names none of them. For banks whose memories "
+        "are tagged with what they are about (`customer:acme`, `service:auth`): ranking cannot "
+        "say which of those a query means, and never says 'none', so a query about something "
+        "the bank has no memory of still returns its best unrelated match. The matched tags "
+        "are AND-ed into the tag filter and applied in SQL, so excluded memories never enter "
+        "retrieval or ranking.",
     )
     min_scores: MinScores | None = Field(
         default=None,
@@ -4929,6 +4940,7 @@ def _register_routes(app: FastAPI):
                         tags=request.tags,
                         tags_match=request.tags_match,
                         tag_groups=request.tag_groups,
+                        query_tag_gate=request.query_tag_gate,
                         min_scores=request.min_scores,
                         temporal_window=request.temporal_window,
                     ),
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -554,6 +554,7 @@ from .response_models import (
     LLMCallTrace,
     MemoryFact,
     MinScores,
+    QueryTagGate,
     RecallScores,
     ReflectResult,
     TemporalWindow,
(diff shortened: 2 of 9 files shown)


nicoloboschi:
Closing this in favour of two focused issues.

The gate works, but the way it resolves the query against the bank's vocabulary does not scale: enumerating distinct tags is a full scan (`FROM memory_units, unnest(tags) GROUP BY tag`) on every gated recall, the GIN index on `tags` serves containment rather than distinct-value enumeration, and `max_vocabulary` silently changes results once a bank exceeds it — with `on_no_match: "abstain"` that means an empty response rather than a slow one.

Splitting out the two parts that stand on their own:

- #4025 — `multi-text` entity labels (the second commit here, unchanged and useful independently)
- #4026 — typo-tolerant tag matching, done as a match mode evaluated in SQL rather than by loading the vocabulary

Separately, this branch carries a reranker fix (`4e2863743`) that is unrelated to either and shouldn't be lost — the local cross-encoder returns NaN on arm64 with torch 2.10 because transformers loads weights via mmap and the float32 CPU matmul misreads that backing store. It fails silently (NaN normalizes to 0, ranking falls back to fusion order, `min_scores.reranker`/`final` go inert). Happy to open that as its own PR.

Issue #3793, opened 2026-08-25
Profile-specific Codex OAuth fallback for explicit quota/transient failures


vpatel9202:
## Use Case

I run Hindsight as a self-hosted service and use Codex/ChatGPT OAuth for its LLM work. I have a
preferred account profile, but that profile can occasionally hit a usage or rate limit while a
separately authorized secondary profile remains available. My goal is continuity for new,
replay-safe memory operations without mixing account credentials or treating every provider error
as permission to retry work elsewhere.

## Problem Statement

Today, the Codex provider selects `auth.json` from the process-wide `CODEX_HOME` (or the default
Codex home). That means every Codex member created by one Hindsight process uses the same store.
Hindsight already has `MultiLLMProvider` routing, but its members do not have a per-member Codex
auth-store/profile field, and generic failover is broad exception-based availability routing. It
therefore does not express a safe profile-specific OAuth policy: quota/transient failure is
materially different from invalid OAuth, malformed configuration, permissions, or a request that
must not be replayed.

## How This Feature Would Help

A small, explicit profile-fallback contract would let a deployment prefer one independently
authorized Codex profile and use a secondary profile only when the preferred profile has an
explicitly eligible quota or transient-availability failure. It would preserve clear re-auth
ownership and avoid a new external credential-bearing service.

## Proposed Solution

Would maintainers be open to discussing this as an issue first, followed—only if the contract is
accepted—by two staged pull requests?

1. **Per-member Codex credential-store seam.** Allow a Codex member to select an explicit,
   independently stored auth file/profile while preserving the existing file-permission checks,
   atomic token refresh, and path-scoped refresh coordination.
2. **Failure-aware cooldown/fallback seam.** Define a narrow provider-neutral
   classification/state interface, initially implemented, enabled, and tested only for Codex
   profile members: preferred profile first, fallback only for explicit quota or
   transient-availability classes, then lazy return to the preferred profile after it becomes
   eligible. This would not add weights, fairness, generic credential pooling, or a new scheduler.

Before implementation, I would appreciate guidance on this proposed behavioral contract:

- **Eligible vs. terminal errors:** only an explicitly classified quota/rate-limit or
  temporary-availability outcome makes a profile eligible for cooldown/fallback. A refresh token
  confirmed expired, reused, or invalidated after the existing refresh and auth-store re-read
  handling is terminal for that profile and requires re-authentication. Access-token rejection
  should retain the existing reactive-refresh behavior before fallback classification.
[...]

PR #4014, opened 2026-09-01
feat(llm): cool down rate-limited Codex profiles

vpatel9202:
## What changed

This is a narrowed revision of the multi-profile Codex OAuth failover work discussed in [#3793](https://github.com/vectorize-io/hindsight/issues/3793), building on [#3983](https://github.com/vectorize-io/hindsight/pull/3983) and the earlier guidance from [@nicoloboschi](https://github.com/vectorize-io/hindsight/issues/3793#issuecomment-5492263337).

The first version went too broad. It carried terminal-auth handling through retain, reflect, consolidation, and worker paths, and it introduced per-operation labels. After [@nicoloboschi's scope review](https://github.com/vectorize-io/hindsight/pull/4014#issuecomment-5510672189), this revision removes those cross-layer changes and keeps the behavior at the router seam.

The router and failure-classification hook are provider-neutral. Codex is the only production provider opting in today; other providers keep their existing behavior unless they add a classifier. For calls that traverse the router, after Codex's configured provider retries finish, explicit 429s enter member-local cooldown, while a positively identified invalid, expired, or reused refresh token is terminal. A terminal exception is re-raised as the exact caught object, unchanged, with no cooldown and no failover.

## Routing behavior

- Cooldown state is process-local to each `MultiLLMProvider`. Calls using the same router instance share it, separate processes do not coordinate it, and restart clears it.
- After cooldown expiry, one request gets the half-open probe lease. A stale success cannot erase a newer cooldown.
- If every member is cooling, a shortest wait at or below the effective `max_backoff` sleeps inline once. A longer wait raises `ProviderRateLimitResetError`.
- Workers retain their existing defer behavior for that error, including no retry-count consumption. Synchronous recall and reflect keep their existing generic HTTP 500 mapping with sanitized retry-time text.
- Codex's own retry budget is unchanged. Batch retain keeps its existing submitting-member affinity, and current-main metadata routing remains intact. An explicitly matched metadata item still binds directly to its selected provider through that existing path, so it does not enter the router's cooldown dispatch.

## Configuration

The only diagnostic labels are the global primary `HINDSIGHT_API_LLM_MEMBER_LABEL` and global indexed `HINDSIGHT_API_LLM_<n>_MEMBER_LABEL`. Per-operation labels are intentionally not parsed. Labels are server-side operational hints, not credential or profile identifiers.

## Validation

A current-main focused matrix passed 691 deterministic tests covering failure classification, cooldown/probe state, bounded inline waits, terminal exception identity, metadata and member ordering, batch affinity, and cross-event-loop routing. The exact final tree also passed Ruff, targeted type checks, the repository's normal commit hooks, and the complete generated-file gate for OpenAPI, bank-template schema, Rust/Python/TypeScript/Go clients, and the generated docs skill. Two independent review passes were clean, including a final adversarial pass over the metadata/cooldown races.

_AI assistance disclosure: I used AI assistance while developing and validating this change, and reviewed the implementation and evidence before submitting it._


nicoloboschi:
Thanks — the routing half of this is exactly the seam I described on #3793, and the state machine looks right: the `generation` counter correctly stops a stale in-flight success from reopening a newer cooldown, `finally` clears `probing` on every exit, and since `_build_llm` runs once in `MemoryEngine.__init__` the state actually survives across operations, which is the whole point (a 429 already fails over today — the new value is the stickiness).

But I'd like to land it narrower. Two asks and a couple of smaller notes.

## 1. Please split out the terminal-auth propagation

`ProviderReauthenticationRequiredError` / `CodexReauthenticationRequiredError` and their propagation through `consolidator._gather_or_cancel`, the reflect synthesis gather, `fact_extraction`'s auto-split gather + `failed_chunks` + `extract_facts_from_contents`, `orchestrator`'s `producer_error`, and `_is_non_retryable_task_error` is roughly a third of the diff (plus the 406-line propagation test), and it lands in four hot concurrency paths.

You're not wrong about the underlying problem — a permanently-invalid Codex refresh token raises a bare `RuntimeError` today and the worker retries it forever. But that's a separate, provider-neutral bug that predates this work and has nothing to do with cooldown or routing. Please open it as its own issue; I'd take that fix on its own merits.

What this PR actually needs from that contract is only the routing half: a terminal classification must not put the member on cooldown and must not fail over. That's fully expressible inside `MultiLLMProvider` — classify terminal, skip the cooldown bookkeeping, re-raise unchanged. Whether the *worker* then retries is orthogonal and shouldn't be decided here.

The concrete objection to the current shape: it rewrites `asyncio.gather(*coros)` into create-tasks + gather + rescan-for-a-specific-exception in four places, changing which exception wins and adding a wait-for-siblings step, for an error class that fires approximately never. That's a lot of risk in retain and reflect for a failure mode nobody is hitting yet.

## 2. Drop the per-operation labels

`HINDSIGHT_API_{RETAIN,REFLECT,CONSOLIDATION}_LLM_MEMBER_LABEL` end up labelling the *same primary credential* three different ways depending on which operation is running. A label identifies a profile, not an operation. Keep `HINDSIGHT_API_LLM_MEMBER_LABEL` and the indexed `HINDSIGHT_API_LLM_<n>_MEMBER_LABEL` — that's the observability I asked for — and drop the three per-op variants along with their config fields and doc rows.

## Smaller notes on the part that stays

- **Sub-second defers.** `_raise_provider_quota_defer` only raises `ProviderRateLimitResetError` when `retry_at` exceeds `max_backoff` — i.e. only when deferring actually beats retrying inline. The router's all-members-cooling path can raise it with a delay as low as `1.0s`, which pushes a whole operation through the worker defer machinery for one second. Please apply the same floor.

- **Single-member / all-cooled error surface.** When every member is cooling, a 429 now exits as `ProviderRateLimitResetError` rather than the provider's own error. In the worker that becomes a clean `DeferOperation`, which is good. On the synchronous reflect/recall HTTP path there's no catch for it (the handler at `memory_engine.py:3328` is worker-only), so a sync caller sees a different exception than before. I'd like that to be a deliberate choice rather than a side effect — a line in the PR description on the intended sync behaviour is enough.

- **Not for this PR, just noting it.** `CodexLLM.call` has `max_retries=10` with exponential backoff, so a rate-limited profile burns ten attempts before the classifier ever sees the 429 and the second profile gets a turn. The cooldown fixes operation #2 onward; operation #1 is still slow. Short-circuiting in-provider retries on a 429 when a chain is configured would be a nice follow-up.

With the terminal half split off and the per-op labels dropped this is a tight change that matches the contract, and I'm happy to merge it.

vpatel9202:
@nicoloboschi -- thanks for the very specific review. I reworked the PR around the router seam you
called out.

The revision removes the terminal-auth propagation through retain, reflect, consolidation, and
worker code, along with the per-operation member-label configuration. The provider-neutral
classifier/router remains, but Codex is the only production opt-in. A confirmed terminal refresh
failure now re-raises the same exception unchanged without cooldown or failover; generic failures
keep the existing behavior.

I also tightened the all-members-cooling path: waits at or below the effective backoff sleep once
inline, and longer waits use the existing `ProviderRateLimitResetError` defer path. The docs call
out the deliberately unchanged synchronous HTTP mapping and Codex retry budget. Cooldown is
process-local, one probe is allowed per router, batch affinity is unchanged, and the current-main
metadata route remains intact. Explicit metadata matches still bind directly to their selected
provider through that existing path, so they intentionally bypass the router's cooldown dispatch.

The final focused matrix, exact-final recheck, generated-file gate, lint/hooks, and independent
review are all green locally. Would you mind taking another look when you have time? I think this
now matches the narrower scope you were asking for.

_AI assistance disclosure: I used AI assistance while developing and validating this change, and reviewed the implementation and evidence before submitting it._

vpatel9202:
Saw this got closed in the sweep on the 15th -- my read is the branch had just gone stale (conflicts from #4260, #4318 and #4234 landing after the 09-07 revision), which is fair. Your 09-02 review was specific enough that the narrowing was quick, and I'd still like to land it.

I've rebased onto current main and pushed it to this branch. Beyond the mechanical conflicts, one thing worth flagging: since #4161 the Codex provider raises `ProviderRateLimitResetError` directly on a long usage-limit window, before any retry. The classifier now maps that to a member cooldown until the named reset. Otherwise a two-profile chain defers the whole operation while the second profile sits idle, which is the exact case this PR exists for. The docs also point at #4336's per-member `MAX_RETRIES=0` as the answer to your "operation #1 is still slow" note.

Focused matrix (209 tests), Ruff, ty and the generated-file gate are green on the rebased tree. Would you rather I reopen this PR with the rebased branch (plain push, history intact) or open a fresh one? Happy either way.

_AI assistance disclosure: I used AI assistance while developing and validating this change, and reviewed the implementation and evidence before submitting it._

PR #4015, opened 2026-09-02
fix(coding-agents): recover cursor after rejected retain

### LINKED ISSUE #3989: coding-agents: a failed retain leaves the cursor dirty, forcing full transcript re-extraction on every subsequent Stop
**Server:** self-hosted API `0.9.2`, harness `claude-code`, one shared bank

## Problem

Any retain failure leaves the session cursor marked `dirty`. A dirty cursor forces `planRetain` into `replace` mode, which re-sends the **entire** session transcript instead of the new turns — and the server re-extracts the whole document. Because the cursor is only cleared on success, an outage does not cost one full re-extraction; it costs one on **every subsequent Stop hook** for the life of the session.

## Root cause

`writeSession()` claims the cursor as dirty *before* the request:

```js
cursors?.write(sessionId, { ...next, dirty: true });   // before submitWithRetry
…
cursors?.write(sessionId, next);                        // only reached on success
```

and `planRetain()` treats a dirty cursor as unusable:

```js
function planRetain(turns, cursor, opts) {
  if (!opts.appendSupported || !cursor || cursor.dirty) return { mode: "replace" };
  …
  return { mode: "append", fromTurn: cursor.turns };
}
```

which selects the payload:

```js
const content = plan.mode === "append"
  ? turns.slice(plan.fromTurn).map(t => JSON.stringify(t)).join("\n")
  : renderSessionJsonl(refId, turns, startTs);   // the whole transcript
```

The claim-before-write is correct for preventing concurrent double-submits. The problem is that a *failed* attempt is indistinguishable from an *interrupted* one, so a transient failure permanently degrades an otherwise healthy session to full-replace.

## Reproduction

1. Start a session in a repo with the plugin active.
2. Make retains fail — e.g. remove `apiToken`, or point `apiUrl` at a port that 401s.
3. Restore working auth.
4. Every Stop from then on sends the full transcript and re-extracts it, even though only a few turns are new.

## Measured impact

Auth broke here at `2026-08-30 18:01` (`apiToken` absent → every retain 401). After restoring the token, document `conversation:<id>` was **267,311 characters**, re-sent and fully re-extracted at each session end. With the plugin's own `conversation` strategy (`retain_chunk_size: 12000`, `retain_extraction_mode: "verbose"`) that is ~22 chunks of LLM extraction per Stop, repeatedly, for content already extracted.

To clarify: Impact is dependent on a "perfect storm" of failure followed by retry, but it's magnified if the user keeps interacting with the session, as subsequent CC turns keep calling Stop.

## Re-extraction is not idempotent

Across one such re-extraction the document **grew** 212,335 → 267,311 characters while its extracted facts **fell** 197 → 131. Facts belonging to a document are replaced wholesale on upsert, so a forced replace does not merely re-pay for existing knowledge — it can silently discard facts that were extracted earlier and are not re-derived from the longer transcript. A user has no way to know what was lost.

## Suggested fix

Distinguish "not yet confirmed" from "known failed". A few options, roughly in order of cost:

- Clear the dirty flag on a *definitive* failure (4xx), keeping it only for indeterminate outcomes (timeout, connection reset), so a rejected request doesn't poison the cursor.
- Or record the attempted `turns` count alongside `dirty`, and on the next pass verify the server-side document against it (the fingerprint machinery already exists) rather than assuming the worst.
- Or cap replace: if `renderSessionJsonl` exceeds some size and the cursor's fingerprint still matches a prefix, prefer append and reconcile.

Even just the first would remove the amplification, since an auth outage — the common case — is entirely 4xx.

## Related

#3280 made identical async retains idempotent via `operation_id`, but explicitly does not cover this: there, the payload is unchanged; here the dirty cursor causes a *different, larger* payload to be built, so the operations are not identical and dedupe cannot apply.

koriyoshi2041:
## Problem

A rejected live-session retain leaves the cursor marked dirty. Every later Stop hook therefore replaces and re-extracts the full transcript, even when the server explicitly rejected the request and could not have committed it.

## Fix

- Preserve HTTP response status in a typed client error.
- Restore the last acknowledged cursor after a definitive authentication, validation, media-type, or payload rejection.
- Keep the cursor dirty for ambiguous statuses, timeouts, and server failures.
- Restore only while the dirty claim is still current, so a newer writer is not overwritten.

## Test

- `npm test` — 847 passed
- `npm run build`
- `git diff --check`

The regression verifies that a 401 restores the acknowledged cursor and the next write appends only the new tail. A 500 remains dirty and continues to force conservative replacement.

## Risk

Low and bounded to coding-agent live-session write-back. The rejection list is intentionally conservative; statuses with uncertain commit semantics retain the existing replace behavior.

Fixes #3989.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/chat.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/chat.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from "vitest";
-import { RateLimitedError, type HindsightClient } from "./hindsight";
+import { HttpResponseError, RateLimitedError, type HindsightClient } from "./hindsight";
 import { ingestChats, renderSessionJsonl, retainLiveSession, type TransportTurn } from "./chat";
 import { memoryCursorStore, type RetainCursorStore } from "./retain-cursor";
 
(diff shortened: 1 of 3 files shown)


nicoloboschi:
superseded by #4210

PR #4018, opened 2026-09-02
fix(retain): enforce generated language integrity

### LINKED ISSUE #4016: Enforce source-language integrity before persisting facts and observations
## Bug Description

Hindsight can persist extracted facts and consolidated observations in a language that is not supported by their source text. Existing prompt rules reduce this drift but do not enforce the invariant before persistence.

This is the unresolved validation portion of #3166. PR #3181 deliberately implemented only the consolidation prompt rule, not post-generation detection, corrective retry, or metrics. The equivalent retain failure was previously reported in #181, and #3494 contains another report of intermittent cross-language generation. More recently, #3776 / #3777 corrected contradictory prompt instructions for an explicitly configured output language, but generated output is still trusted without validation.

The desired default is not English-only output. When `llm_output_language` is unset, generated facts and observations should remain compatible with the applicable source language, including legitimate multilingual sources. When it is explicitly configured, that configured language should take precedence.

### Production evidence

Observed on a self-hosted Hindsight 0.9.1 deployment using the `openai-codex` provider and `gpt-5.6-luna`, with `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE` unset:

- An English source chunk produced a Spanish fact during retain extraction.
- Consolidation propagated that Spanish fact into a Spanish observation.
- Independent consolidation drift also occurred where source facts and generated observations differed in language.
- In recent request telemetry, 170 of 1,757 retain requests whose source chunks were independently classified as English produced non-English output, a 9.68% request-level mismatch rate.
- A corpus audit identified 6,273 facts translated during extraction, 3,596 observations that propagated mistranslated facts, and 864 observations that independently changed language during consolidation.
- The affected extraction records traced to 945 source documents and 1,860 source chunks across multiple ingestion routes, which rules out a collector-specific transformation.
- The same audit found 31 legitimate Spanish or Japanese records backed by corresponding source material, so forcing a deployment-wide English output language would corrupt valid multilingual behavior rather than solve the invariant.

The source documents themselves remained in their original languages. The drift was introduced in the LLM generation stages.

## Steps to Reproduce

Because the failure is model-dependent and intermittent:

1. Run Hindsight with `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE` unset and a multilingual generative model.
2. Retain a repeated corpus of unambiguously English prose through the normal fact-extraction pipeline.
3. Compare each generated fact's language with its source chunk before persistence.
4. Run consolidation over the resulting facts.
5. Compare each proposed create or update with the language of the facts referenced by its `source_fact_ids`.
6. Repeat at production-like volume.

With an affected model, some facts or observations are generated in another language while the operations otherwise succeed. No error or warning identifies those records.

A deterministic regression test can reproduce the persistence defect without relying on a particular model: configure the test LLM to return a valid structured response in a mismatching language on the first call, then assert that Hindsight retries and does not persist that response.

## Expected Behavior

Before a generated fact or observation is persisted, Hindsight should enforce this invariant:

- If `llm_output_language` is configured, generated natural-language fields use that language.
- Otherwise, generated fields use a language supported by the relevant source chunk or source-fact set.
- Proper nouns, identifiers, code, commands, URLs, and technical terms may remain unchanged.
- Short, language-neutral, code-heavy, low-confidence, and genuinely mixed-language sources are handled conservatively rather than being falsely rejected.

A confident mismatch should trigger one corrective regeneration from the original source. If the regenerated output still mismatches, the operation should follow an explicit failure policy and must not persist the mismatching output as a valid fact or observation.

## Actual Behavior

Structured-output and schema validation can succeed even when generated text changes language. The response is then persisted normally. Prompt instructions are the only current guard, so violations are silent and can propagate from facts into observations.

## Proposed Direction

Add one shared output-language integrity boundary used by retain extraction and consolidation:

1. Derive the expected language profile from the actual source chunk or the `source_fact_ids` supporting each observation.
2. Normalize or discount URLs, code, identifiers, numbers, and other language-neutral content before classification.
3. Validate parsed generated natural-language fields before any persistence or consolidation marker is committed.
4. On a confident mismatch, retry once with an explicit corrective instruction using the original source, not a translation of the rejected output.
5. Revalidate the retry and fail closed, or use an explicitly configured fallback policy, if it still mismatches.
6. Emit structured, text-free telemetry containing the stage, provider/model, expected and generated languages, confidence, retry outcome, and source lineage identifiers.

The classifier and thresholds should be conservative. Indeterminate classification should not be treated as proof of a mismatch.

## Done When

- [ ] English source chunks cannot persist confidently Spanish, Chinese, Japanese, or other mismatching extracted facts.
- [ ] Legitimate Spanish, Japanese, and mixed-language source material remains in a compatible source language when no output-language override is configured.
- [ ] An explicit `llm_output_language` remains authoritative.
- [ ] Consolidation validates each create or update against its referenced source facts before writing it.
- [ ] A mismatching first response triggers exactly one corrective regeneration and is never persisted.
- [ ] A second confident mismatch follows a documented terminal policy without leaving partial facts, observations, or consolidation markers.
- [ ] Short, technical, URL-heavy, code-heavy, language-neutral, and low-confidence inputs have deterministic conservative behavior.
- [ ] Metrics distinguish initial passes, retry recoveries, indeterminate checks, and terminal mismatches by stage and model without logging source text.
- [ ] Deterministic tests cover extraction and consolidation creates and updates, retry success, terminal mismatch, atomicity, configured overrides, legitimate multilingual input, mixed input, and neutral technical text.
- [ ] A production-volume multilingual canary demonstrates that the selected detector and thresholds block known drift without rejecting legitimate multilingual artifacts.

## Version

0.9.1

## LLM Provider

Other (`openai-codex`, model `gpt-5.6-luna`)

0xble:
## Summary

- add a conservative, deterministic source/output language validator shared by retain extraction and observation consolidation
- allow one corrective regeneration, then fail closed without worker retry when the generated language still conflicts
- preserve configured output-language overrides, mixed/short/technical inputs, Batch API chunk isolation, and bounded telemetry

## Why

Hindsight's prompts already tell models not to translate, but prompt compliance is not an integrity boundary. A private production audit found confident language drift at both extraction and consolidation. The guard catches model output before it is persisted rather than relying on another prompt-only mitigation.

A sanitized corpus canary against the audited rows detected 10,660 of 10,733 known drift records (99.32%) while rejecting 0 of 31 legitimate multilingual records and 0 of 220 detector false-positive controls. The remaining 73 drift rows are intentionally indeterminate or conservative false negatives, not false accepts mislabeled as validated matches.

## Design

- seed `langdetect` for repeatable results and normalize detector/locale codes
- strip code, URLs, UUIDs, and identifier-heavy tokens before classification
- abstain on short, repetitive, mixed, or low-confidence text
- validate each fact and each consolidation action independently
- regenerate only the mismatching Batch API chunk through the normal provider path
- respect `HINDSIGHT_API_LLM_OUTPUT_LANGUAGE`, including locale forms such as `en-US`, `pt-BR`, and `zh-TW`
- emit text-free counters by stage, outcome, provider, and model
- classify a second confident mismatch as non-retryable so workers do not resend the same payload indefinitely

## Validation

- `170 passed`: broad impacted suite covering validation, retain robustness, Batch API, multi-provider batching, chunk coverage, consolidation atomicity/retry budgets, worker retries, and metrics
- Ruff, lockfile validation, compileall, and wheel build pass
- audited corpus canary: 10,660/10,733 drift detected, 0/31 legitimate mismatches, 0/220 control mismatches
- independent blocker review: no unresolved blockers

## Risk

The guard is deliberately conservative. Indeterminate output remains accepted, so this reduces known corruption without turning language detection uncertainty into ingestion failures. The change does not repair historical memories.

Closes #4016

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -36,6 +36,14 @@ from ...config import get_config
 from ...worker.stage import set_stage
 from ..db import DatabaseBackend
 from ..db_utils import acquire_with_retry
+from ..language_validation import (
+    GeneratedLanguageMismatch,
+    LanguageValidationOutcome,
+    build_language_retry_instruction,
+    record_language_validation,
+    should_validate_language,
+    validate_output_language,
+)
 from ..llm_interface import OutputTooLongError, ProviderRateLimitResetError
 from ..llm_trace import (
     record_created_memory_ids,
@@ -3059,7 +3067,7 @@ def _classify_batch_failure(exc: Exception) -> _BatchFailureClass:
     # front (openai, anthropic) exposes the HTTP status this way.
     if getattr(exc, "status_code", None) in (401, 403):
         return _BatchFailureClass.PROPAGATE
-    if isinstance(exc, json.JSONDecodeError | ValidationError | OutputTooLongError):
+    if isinstance(exc, json.JSONDecodeError | ValidationError | OutputTooLongError | GeneratedLanguageMismatch):
         return _BatchFailureClass.FAIL_FAST
     # Providers that surface an empty/unusable body flag their own retryability.
     if getattr(exc, "retryable", None) is False:
@@ -3162,18 +3170,23 @@ async def _consolidate_batch_with_llm(
     # "LLM batch call failed" line gives operators no way to find the offending
     # input until adaptive bisection narrows the batch down to a single memory.
     memory_ids = [str(m.get("id")) for m in memories]
+    source_text_by_id = {str(m.get("id")): str(m.get("text") or "") for m in memories}
     if len(memory_ids) <= 5:
         ids_label = ", ".join(memory_ids)
     else:
         ids_label = f"{', '.join(memory_ids[:3])}, ... +{len(memory_ids) - 3} more"
     batch_label = f"{len(memory_ids)} memories [{ids_label}]"
-    for attempt in range(1, max_attempts + 1):
+    language_retry_used = False
+    language_retry_instruction = ""
+    # The extra slot is reachable only after a language mismatch. Ordinary
+    # transport failures retain the configured max_attempts budget below.
+    for attempt in range(1, max_attempts + 2):
         attempts_made = attempt
         try:
             call_kwargs: dict[str, Any] = {
                 "messages": [
                     {"role": "system", "content": system_prompt},
-                    {"role": "user", "content": user_content},
+                    {"role": "user", "content": user_content + language_retry_instruction},
                 ],
                 "response_format": response_model,
                 "temperature": config.llm_temperature_consolidation,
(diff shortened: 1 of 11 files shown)


nicoloboschi:
Thanks for the thorough write-up and the corpus work behind it — the drift problem you describe is real and worth addressing.

We're not going to take this approach, though, mainly because of the dependency. `langdetect` is effectively unmaintained: the last release is 1.0.9 from May 2021, upstream has ~55 open issues / 16 open PRs with no activity since, it ships sdist-only (no wheel) and still pulls in `six`. We don't want a runtime dependency in that state sitting on the ingestion path, especially one that can fail an operation closed.

A couple of things we found while evaluating it, for the record:

- Its posteriors saturate — `detect_langs` returns ~0.9999 with a ~1.0 margin for essentially every single-language input, including when it is wrong. So the `_MIN_CONFIDENCE = 0.92` / `_MIN_MARGIN = 0.20` gates don't actually filter much; the effective gate is the 48-character floor.
- The CJK early-return in `profile_language` returns before the mixed/segment analysis, so any English document containing four kana characters anywhere (a support transcript, a code comment, a product name) profiles as pure `ja` and every fact extracted from it mismatches. Same in reverse for a fact quoting a katakana term. With `GeneratedLanguageMismatch.retryable = False` that fails the whole retain permanently.
- `detect_langs` is synchronous pure Python called from async paths (~5 ms on a sentence, ~50 ms on ~2 KB), and the source is re-profiled once per fact, so a chunk with 20 facts blocks the event loop for a while.

More fundamentally, we'd rather not turn a language heuristic into a fail-closed integrity boundary that has no config flag to switch off.

What would help a lot: could you open an issue with a **reproducer** — the source content (sanitized is fine), the provider/model, and the retain or consolidation output that came back translated? With concrete cases we can improve the extraction/consolidation prompts, which is the level we want to fix this at for now.

Closing for the reasons above — genuinely appreciate the effort here.

Issue #1883, opened 2026-05-31
Retain returns HTTP 500 on content containing tiktoken special-token literals (e.g. endoftext) — same input-robustness class as #1875


ruashots:
## Summary

Hindsight's retain path returns **HTTP 500** when ingested content literally contains a tiktoken special-token string such as `<` + `|endoftext|` + `>`. The server raises:

```
HTTP 500: Encountered text corresponding to disallowed special token '<|endoftext|>'.
If you want this text to be encoded as a special token, pass it to `allowed_special`...
If you want this text to be encoded as normal text, disable the check for this token by passing
`disallowed_special=(enc.special_tokens_set - {'<|endoftext|>'})`.
To disable this check for all special tokens, pass `disallowed_special=()`.
```

This is the same input-robustness class as #1875 (recall 500 on an unpaired UTF-16 surrogate): the retain/tokenizer path passes arbitrary user content to `tiktoken.encode()` with the default `disallowed_special="all"`, so any text that happens to contain a special-token literal aborts the whole retain with a 500 instead of being treated as ordinary text.

## Why it matters

A memory system whose job is to ingest arbitrary conversation transcripts will, sooner or later, be handed text that *mentions* `<|endoftext|>`, `<|im_start|>`, etc. — verbatim. This is common for anyone using Hindsight on AI/ML work: discussing tokenizers, debugging chat templates, pasting model output, or (in my case) running the official Claude Code plugin over a session that spent the day discussing that exact token. The auto-retain hook then silently 500s on every fire (`retainEveryNTurns`), and the target bank is never created — with no operator-visible error unless you manually run the hook with `HINDSIGHT_DEBUG=true`.

Normal transcripts (zero special-token literals) ingest fine, so this only bites the subset of content that references special tokens — but for that subset it's a hard failure, not a degraded result.

## Reproducer

Retain any content containing a special-token literal:

```bash
curl -s -X POST http://<hindsight>/v1/default/banks/<bank>/memories \
  -H "Content-Type: application/json" \
  -d '{"items":[{"content":"The fix was to sanitize the <|endoftext|> token before sending."}]}'
# → HTTP 500, disallowed special token
```

[...]

PR #4022, opened 2026-09-02
perf(tokenizer): move token counting from quicktok to toktok-rs

nicoloboschi:
Swaps `quicktok-v1` for [`toktok-rs`](https://github.com/vectorize-io/toktok), a Rust BPE tokenizer whose ids are byte-identical to tiktoken's.

## The swap is behaviour-neutral

Compared side by side over **258 texts** — ~250 real files from `hindsight_api/` plus the special-token and mixed-script edge cases — quicktok and toktok produce **identical counts and identical ids** on all three shared encodings:

```
o200k_base:     258 texts, count mismatches=0, id mismatches=0
cl100k_base:    258 texts, count mismatches=0, id mismatches=0
o200k_harmony:  258 texts, count mismatches=0, id mismatches=0
```

No token budget, chunk boundary or truncation point moves, so no recall/retain behaviour changes with this.

## API mapping

`engine/token_encoding.py` remains the only importer.

| before | after |
|---|---|
| `count()`, `decode()` | unchanged |
| `encode(text, disallowed_special=())` | `encode_ordinary(text)` |

`encode_ordinary` yields the same ids and has no special-token machinery at all, which is a cleaner shape for the #1883 fix than disabling the check. toktok's supported surface is just `batch_count`; truncation needs `decode`, so `get_token_encoding()` takes `toktok._encoding(...)` — the escape hatch its own docstring designates, "for anyone who knowingly wants the full encode/decode surface" — commented as such, and confined to this module.

## Two consequences worth a look

**`llama3` and `qwen3` are gone.** quicktok bundled five vocabularies; toktok bundles three (`cl100k_base`, `o200k_base`, `o200k_harmony`). A deployment with `HINDSIGHT_API_TOKENIZER_ENCODING=llama3` now fails at its first token count — lazily, on first retain/recall — with the existing "Unknown tokenizer encoding" `ValueError`. Docs and both env templates are updated. `BUNDLED_ENCODINGS` is hand-maintained and had been lying about those two, so it now carries a parametrized test that loads every name it advertises.

**Wheels.** `cp311-abi3` covers 3.11–3.14, so 3.14 no longer compiles from source the way quicktok did (CI was green there only because it has a toolchain). In exchange there are no musllinux wheels — irrelevant to the shipped images, whose Python stages are all glibc `python:3.11-slim` and whose alpine stage is node-only, but an Alpine install would build the sdist. numpy also drops to an optional extra, so the tokenizer now pulls in no dependency of its own.

## Performance

Measured on this repo's own text with the existing microbenchmark (`./scripts/benchmarks/run-token-count-bench.sh`), every variant reporting `ok` (= ids match tiktoken):

- **2–7x** faster than tiktoken on `cl100k_base`, **10–16x** on `o200k_base`
- counting an 81k-token document peaks at **1 KiB** against tiktoken's **~3 MB**

The docstring's inherited speed figures were replaced with these.

## Test plan

- [x] `./scripts/hooks/lint.sh`, `uv run ty check hindsight_api/`, `./scripts/hooks/check-unused.sh` — clean
- [x] 69 tests across the tokenizer, truncation, input-robustness, recall-cap, source-fact and embedding-cap suites
- [x] 201 more across retain, batch chunking, delta retain, fact-budget selection and document serialization
- [x] microbenchmark runs on both vocabularies, all variants `ok`

One unrelated pre-existing failure was seen while running the above: `test_mental_models.py::TestMentalModelRefreshTagSecurity::test_refresh_with_tags_only_accesses_same_tagged_models`, an `hs_llm_core` real-LLM judge test. It fails on the *completeness* clause ("fails to mention her preference for mornings and coffee") while the security clause it exists to guard holds. Given the identical-ids result above, this change has no path to it.

https://claude.ai/code/session_01WbaYxouCERUv9djo3GnnA2


nicoloboschi:
Pushed `cdae371` — dropped the `_SafeEncoding` wrapper and made the module's interface the two operations the engine actually performs.

The wrapper existed only to force `disallowed_special=()` onto `encode()`. Auditing what callers did with the object it handed back, **every single use was either a plain `count`, or `decode(encode(x)[:n])` open-coded** — which is precisely `truncate_to_tokens`. So it was indirection around two functions.

`get_token_encoding()` now returns the configured toktok tokenizer itself, and the interface is:

```python
count_tokens(text)           -> int
truncate_to_tokens(text, n)  -> TokenTruncation
```

Counting reaches toktok's native `count()` with nothing in between, and truncation is written once rather than at each call site. Converted:

- `http.py` recall validation, and both `count_tokens=encoding.count` callables in recall's budget selection → `count_tokens`
- `_truncate_query_to_token_limit` and recall's chunk-budget truncation → `truncate_to_tokens` (both were `decode(encode(x)[:n])` verbatim)
- three now-dead `encoding = get_token_encoding()` locals removed

`get_token_encoding()` keeps exactly one caller — the eager vocabulary warm-up at startup — plus the tests.

**The #1883 guarantee is unchanged, but structural instead of wrapped.** `count()` never applies the special-token check, and truncation encodes with `encode_ordinary()`, which has no special-token machinery at all. The raising spelling is `encode()`, which nothing calls; the module docstring says so explicitly. The #1883 regression test moves onto `truncate_to_tokens` — the path production actually takes — and a new `test_counting_agrees_with_the_ids_truncation_uses` pins that `count` and `encode_ordinary` can't drift, which is a real risk now that the tokenizer exposes two id-producing spellings and only one of them agrees with `count`.

Re-verified: `lint.sh` + `ty check` clean, **180 tests** pass (69 tokenizer/truncation/robustness/embedding-cap, 111 across source-fact selection, fact budgets, batch chunking, recall query cap and the HTTP API integration suite).

nicoloboschi:
Pushed `1664102` — the tokenizer itself is now off the module's interface.

After the previous commit, `get_token_encoding()` had **exactly one** production caller: the eager vocabulary warm-up in `MemoryEngine.__init__`, which only wants "load it now". `count_tokens("")` says that just as well. So it is `_load_encoding()` now, and the module's entire public surface is:

```python
BUNDLED_ENCODINGS
count_tokens(text)           -> int
truncate_to_tokens(text, n)  -> TokenTruncation
```

This is what makes #1883 fixed *by construction* rather than by convention. Every route to the tokenizer's raising `encode()` went through that function; with it private, the only spellings a caller can reach are the two that treat `<|endoftext|>` as ordinary text. Tests keep using `_load_encoding` — they need `cache_clear()` to switch encodings and `.name` to assert which is in play.

---

### While auditing this: where `truncate_to_tokens` is actually used

Worth recording, since "does Hindsight cut text mid-word?" came up. Storing text never does — the retain splitter descends paragraph → line → sentence → `;` → `,` → word, with a character cut only as a documented last resort. `truncate_to_tokens` is used in five places, all of them a hard external limit where the alternative is a failure, not a nicer cut:

| Call site | Cap | Without it |
|---|---|---|
| `retain/embedding_utils._truncate_inputs` | `EMBEDDINGS_MAX_INPUT_TOKENS` (off by default) | Bedrock Titan (8192) / llama.cpp reject with a permanent 4xx — one oversized memory fails the whole batch |
| `cross_encoder._truncate_to_tokens` | `max_tokens_per_doc` | Reranker input overflow |
| `memory_engine._truncate_query_to_token_limit` | `RECALL_MAX_QUERY_TOKENS` | 58k-word query → 54k-term `tsquery` → Postgres stack overflow (54001) |
| `memory_engine` recall chunk budget | `max_chunk_tokens` | Response overflow; only the last partially-fitting chunk is cut, flagged `truncated=True` |
| `reflect/prompts._truncate_prompt_text` | 24k default | Structured-delta prompt exceeds context |

Three of the five warn or set a `truncated` flag, so a cut is never silent.

**One pre-existing wrinkle this surfaced** (not introduced here, flagging it rather than fixing it in this PR): these cuts land on a *token* boundary, which is not a character boundary, so a multi-token character can be split —

```
'hello 🧠' -> 4 tokens;  [:2] -> 'hello �';  [:3] -> 'hello �'
```

Nothing crashes (decode replaces rather than raises), and I verified it is **byte-identical under quicktok** — same ids, same output — so this is inherent to `decode(encode(x)[:n])` and predates both tokenizers (tiktoken decodes with `errors="replace"` too). If it's worth cleaning up, the fix is local to `truncate_to_tokens`: strip a trailing `�` the input didn't have. Happy to add it here or leave it.

Re-verified: `lint.sh` + `ty check` clean, 173 tests pass (69 tokenizer/truncation/robustness/embedding-cap + 104 source-fact selection, fact budgets, batch chunking, HTTP API integration).

nicoloboschi:
Pushed `6518cc4` — the U+FFFD issue is fixed properly, using `truncate()` / `batch_truncate()` from toktok 0.1.3 (floor bumped, lock updated).

The old implementation was `decode(encode(text)[:n])`, cutting on a **token** boundary. The native call cuts on a **character** boundary:

```python
truncate_to_tokens("hello 🧠", 2).text
# before: 'hello �'
# after:  'hello '
```

And it's faster on top of being correct: one pass, the cut is a byte offset into the input string, no ids built and nothing decoded. When nothing needs cutting the original string object comes back as-is — so the fits-in-budget path, which is the common one for reranker documents and embedding inputs, no longer even counts separately. That's strictly better than the count-then-encode dance the wrapper was doing.

Dropping a partial character can only leave the result *under* the requested budget, never over it. All five callers are ceilings, so it costs nothing.

**`truncate_many_to_tokens`** wraps `batch_truncate` for the two callers that truncate a whole list — replacing a Python loop with one call that releases the GIL:

- `cross_encoder`: every candidate document, in both LiteLLM cross-encoders (`_truncate_to_tokens` → `_truncate_docs_to_tokens`, which is now list-shaped since looping was all it was ever used for)
- `retain/embedding_utils._truncate_inputs`: every embedding input; its per-item `original_tokens` still drives the "truncated N of M inputs (largest was ~K tokens)" warning

**One behaviour change worth noting:** a negative budget used to slice a list with a negative index, silently dropping tokens off the *end*. The native call takes an unsigned count and would raise. It clamps to 0 now, so a misconfigured cap degrades to empty rather than a 500 mid-request — covered by a test.

Tests pin the fix at **every cut point** of every multi-token-character string (`"hello 🧠"`, `"東京タワー"`, `"مرحبا بالعالم"`, `"🧠🧠🧠"`) across **all three** bundled vocabularies, plus that the batched form agrees with the single-text one it stands in for, and that an empty list works.

Verified: `lint.sh` + `ty check` clean, 123 tests pass (tokenizer, truncation, robustness, recall cap, source facts, embedding cap, and the LiteLLM/Cohere/local cross-encoder + reranker-error suites).

nicoloboschi:
Ran `/code-review` over the whole branch and pushed `4d44a09`. Also rebased onto current `origin/main` (`f0cf699`) — main had since touched `cross_encoder.py`, which this branch also changes; it rebased clean and the suites still pass.

**Findings: 5, all mine, all comments the native `truncate()` made stale two commits ago.** No behaviour change:

- the module docstring still said truncation "encodes with `encode_ordinary()`" — it doesn't; it goes through toktok's `truncate`, which builds no ids at all. Both operations are special-token-safe for the same reason now, so it says that once. (Also fixes a duplicated "the two two functions".)
- `_load_encoding` said we take toktok's private `_encoding` because "truncation needs decode". It doesn't. The real reason to hold the loaded tokenizer is `count()` — no per-call list to wrap a single string in — plus the resolved `.name` that `truncate`/`batch_truncate` take.
- `truncate_many_to_tokens` claimed "one scratch buffer per thread instead of one per text". `truncate` builds no buffer at all; the win is the parallelism.
- two test docstrings still described truncation as "the only operation that needs ids", which was their whole reason to exist under the old implementation.

**One of these was more than cosmetic.** `test_counting_agrees_with_the_ids_truncation_uses` was still asserting `count_tokens == len(encode_ordinary(...))` — a coupling truncation no longer has — while the coupling that *is* live went untested. It now asserts what the engine actually depends on: `count_tokens` and `truncate`'s `original_tokens` are the same number. They come from different toktok entry points and the engine mixes them, e.g. `_truncate_inputs` decides it truncated by comparing `original_tokens` against a cap other code arrived at via `count_tokens`. A divergence there would surface as an off-by-N in a budget, not as an error.

**Clean on everything else:** branch hygiene (5 commits, all on-topic, now on a current base), `lint.sh`, `ty check`, `check-unused.sh`, `.env.example` ↔ embed byte-parity, no advisory locks, no raw dicts or tuple returns, no DB access added to the handler, docs-skill regeneration produces no drift. No API/migration/MCP/integration/control-plane surface touched, so no OpenAPI or client regeneration in scope.

208 tests pass locally (tokenizer, truncation, robustness, recall cap, source facts, embedding cap, the four cross-encoder/reranker suites, fact-budget selection, and the HTTP API integration suite). Watching CI now.

PR #4030, opened 2026-09-02
fix(clients): close the bank-config drift between the wrapper SDKs and the server

### LINKED ISSUE #4029: client wrappers: bank-config surface has drifted from the server, and the check that would catch it never runs
Found while adding `multi-text` entity labels (#4025 / #4027): the TypeScript wrapper's `updateBankConfig` had no `entityLabels` option at all, so TS consumers could not configure a controlled vocabulary of *any* type. Fixing that one field surfaced a much larger drift.

## The shape of the problem

`PATCH /v1/default/banks/{bank_id}/config` takes a free-form body:

```python
# hindsight-api-slim/hindsight_api/api/http.py:1512
updates: dict[str, Any] = Field(...)
```

So the **generated** SDKs can send any config field. The **wrappers** (`hindsight_client.py` / `src/index.ts`) hand-enumerate keyword arguments, and that list is the sole constraint on what a consumer can actually set. It has drifted three ways.

### 1. Server has 47 configurable fields; 22 are unreachable from *either* wrapper

Measured against `_CONFIGURABLE_FIELDS` in `config.py` (47 fields). Python's `update_bank_config` exposes 25, TypeScript's `updateBankConfig` exposes 17.

Missing from **both**:

```
audit_log_enabled                          recall_budget_adaptive_high
consolidation_llm_parallelism              recall_budget_adaptive_low
consolidation_max_memories_per_round       recall_budget_adaptive_mid
enable_auto_consolidation                  recall_budget_fixed_high
max_observations_per_scope                 recall_budget_fixed_low
memory_defense                             recall_budget_fixed_mid
mental_model_min_refresh_interval_seconds  recall_budget_function
observation_scope_limits                   recall_budget_max
retain_chunk_batch_size                    recall_budget_min
store_document_text                        recall_chunks_max_tokens
                                           recall_include_chunks
                                           recall_max_tokens
```

The entire recall-budget group and `memory_defense` are settable via HTTP and the control plane, but not from either SDK wrapper.

### 2. Eight more are in Python but not TypeScript

```
consolidation_llm_batch_size                          mcp_enabled_tools
consolidation_source_facts_max_tokens                 reflect_source_facts_max_tokens
consolidation_source_facts_max_tokens_per_observation  retain_default_strategy
llm_gemini_safety_settings                            retain_strategies
```

(`create_bank` / `createBank` are at parity — this is specific to the config-update method.)

### 3. `client-coverage-check` is never wired into CI

`hindsight-dev/hindsight_dev/client_coverage_check.py` exists precisely to catch wrapper drift, and `.openapi-coverage.toml` manifests are maintained for both clients — but `.github/workflows/test.yml` only runs its sibling `cli-coverage-check` (line 5328). The client one runs nowhere.

It is **currently red on `main`**:

```
FAILED: 2 issue(s):
  [python]     MISSING PARAM reflect.apply_all_directives
  [typescript] MISSING PARAM reflect.apply_all_directives
```

`apply_all_directives` is missing from both wrappers and nothing reported it.

### 4. Even wired up, the tool structurally cannot see this

It walks *typed* request-body properties from the OpenAPI spec. Because `updates` is `dict[str, Any]`, `updateBankConfig` reads as fully covered by its single `updates` key — every one of the 30 gaps above is invisible to it. This is why the drift went unnoticed while the equivalent check for CLI coverage held.

## Proposed fix

1. Bring both wrappers up to the full `_CONFIGURABLE_FIELDS` surface, keeping the two in step.
2. Fix `reflect.apply_all_directives` in both, and add `client-coverage-check` as a CI job next to `check-cli-coverage`.
3. Add a guard the existing tool cannot express: a test that reads `_CONFIGURABLE_FIELDS` from the server and asserts both wrapper sources expose each one, with an explicit commented exemption list. Same shape as `test_backup_tables_covers_entire_schema` and `harness/registry.test.ts` — enumerate the family from its source of truth rather than trusting a hand-maintained list.

Without step 3 this recurs: it has now happened at least three times (#2975 / #3042 for `listMentalModels` params, then `entity_labels`, and `apply_all_directives` sitting red unnoticed).

nicoloboschi:
Fixes #4029.

## What was wrong

`PATCH /banks/{bank_id}/config` takes a free-form `updates: dict[str, Any]`, so the generated SDKs can send anything. The **wrappers** build that body by enumerating fields by hand, and nothing regenerated or checked that list. Measured against `_CONFIGURABLE_FIELDS` (47 fields):

| | before | after |
|---|---|---|
| Python `update_bank_config` | 25 / 47 | **47 / 47** |
| TypeScript `updateBankConfig` | 17 / 47 | **47 / 47** |

22 fields were reachable from **neither** wrapper — the entire recall-budget group, `memory_defense`, `store_document_text`, the auto-consolidation knobs. 8 more were Python-only. `reflect.apply_all_directives` was missing from both.

## Two wrong type hints

- `llm_gemini_safety_settings` was `dict[str, str]` in Python. The server (`config.py:2540`), the Gemini provider (`gemini_llm.py:237`) and the control plane all take a **list** of `{category, threshold}`. A caller following the hint sent a payload the provider rejects.
- (`entity_labels: list[str]` was the same class of bug, fixed in #4027.)

## The check that should have caught this

`client-coverage-check` exists for exactly this and was **wired into no workflow**. It was red on `main`:

```
FAILED: 2 issue(s):
  [python]     MISSING PARAM reflect.apply_all_directives
  [typescript] MISSING PARAM reflect.apply_all_directives
```

It now runs in CI beside `check-cli-coverage` and gates `report-pr-status`.

It also could not see this drift even if it had run: it walks *typed* request-body properties, and `updateBankConfig` reads as fully covered by its single `updates` property. So the check now reads `_CONFIGURABLE_FIELDS` from `config.py` directly, with `[bank_config]` skips in the existing per-client manifests.

### It verifies forwarding, not just acceptance

I hit this while writing it: after adding the keyword arguments, coverage went green while the payload mapping was still missing — a field accepted and never written into `updates` is silently dropped, and the caller gets a `200` and no change. That is precisely how this bug class presents, so the check requires both. Negative-tested by deleting one mapping line:

```
[python] UNFORWARDED CONFIG audit_log_enabled: accepted by the wrapper but never
         written into the request body — it would be silently dropped
```

The scan is scoped to the config method rather than the whole file: names like `recall_max_tokens` also appear as recall arguments, and a whole-file search reported 19 TS fields covered when the method accepted 17.

## Tests

Mirrored mapping tests on both sides, each naming the other so they stay in step: `test_bank_config_update_payload.py` and `bank_config_full_surface_mapping.test.ts`, covering the recall-budget group, the consolidation/retain/security groups, the `undefined`-is-omitted case, the gemini list shape, and `apply_all_directives` on reflect.

Local: 88 Python and 72 TypeScript pass. The `test_main_operations` suites in both clients need a live API on `:8888` and fail with `ClientConnectorError` / `fetch failed` without one — unrelated to this change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5327,6 +5327,47 @@ jobs:
         cd hindsight-dev
         uv run cli-coverage-check
 
+  # The wrapper clients (hindsight_client.py / src/index.ts) are hand-written
+  # convenience layers over the generated SDKs, so nothing regenerates them when
+  # the API gains a field. client-coverage-check has existed for that, but had no
+  # job: it sat red on main with reflect.apply_all_directives missing from both
+  # wrappers, and the bank-config surface had drifted to 22 settings reachable
+  # from neither. See #4029.
+  check-client-coverage:
+    needs: [detect-changes]
+    if: >-
+      (github.event_name == 'workflow_dispatch' ||
+      needs.detect-changes.outputs.core == 'true' ||
+      needs.detect-changes.outputs.clients-ts == 'true' ||
+      needs.detect-changes.outputs.clients-python == 'true' ||
+      needs.detect-changes.outputs.dev == 'true' ||
+      needs.detect-changes.outputs.ci == 'true')
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        ref: ${{ github.event.pull_request.head.sha || '' }}
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v7
+      with:
+        enable-cache: true
+
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        python-version-file: ".python-version"
+
+    - name: Install hindsight-dev dependencies
+      run: |
+        cd hindsight-dev && uv sync --frozen --index-strategy unsafe-best-match
+
+    - name: Check client wrappers cover every request param and bank config field
+      run: |
+        cd hindsight-dev
+        uv run client-coverage-check
+
   # hindsight-dev/tests had no job of its own, so nothing ran it: the benchmark
   # harness was only exercised by the nightly Performance Tests workflow, where a
   # plain construction bug in the answer/judge LLM config surfaced as a red
@@ -5448,6 +5489,7 @@ jobs:
       - verify-generated-files
       - check-openapi-compatibility
       - check-cli-coverage
+      - check-client-coverage
     runs-on: ubuntu-latest
     timeout-minutes: 30
     permissions:
--- a/hindsight-clients/python/hindsight_client/hindsight_client.py
+++ b/hindsight-clients/python/hindsight_client/hindsight_client.py
@@ -564,6 +564,7 @@ class Hindsight:
         include_tool_calls: bool = False,
         include_tool_call_output: bool = True,
         tag_groups: list[dict[str, Any]] | None = None,
+        apply_all_directives: bool = False,
         fact_types: list[str] | None = None,
         exclude_mental_models: bool = False,
         exclude_mental_model_ids: list[str] | None = None,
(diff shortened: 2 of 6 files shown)


koriyoshi2041 (approved):
Tested the exact head locally: the Python bank-config mapping suite passed 9/9, the TypeScript mapping suite passed 5/5, the coverage checker accounted for all 47 configurable fields in both wrappers, and diff-check was clean. The forwarding checks cover the important silent-drop boundary, including omission of unset fields. Looks good.

PR #4035, opened 2026-09-02
fix(embeddings): support AWS Bedrock application inference profile ARNs in litellm-sdk

### LINKED ISSUE #4034: litellm-sdk embeddings provider can't use an AWS Bedrock application inference profile ARN
### Summary

When `HINDSIGHT_API_EMBEDDINGS_PROVIDER=litellm-sdk` and `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL` is set to a Bedrock **application inference profile ARN** (e.g. `bedrock/arn:aws:bedrock:<region>:<account>:application-inference-profile/<id>`), startup fails with:

```
RuntimeError: Failed to initialize LiteLLM SDK embeddings: litellm.APIConnectionError: Unable to determine bedrock embedding provider for model: arn:aws:bedrock:<region>:<account>:application-inference-profile/<id>. Supported providers: ['cohere', 'amazon', 'twelvelabs', 'nova']
```

This makes it impossible to use an application inference profile for embeddings at all — even though AWS Bedrock's own `invoke_model` API accepts the ARN directly as `modelId` (see the [Titan Text Embeddings V2 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-amazon-titan-text-embeddings-v2.html)).

### Why this happens

Some AWS orgs configure a Service Control Policy that explicit-denies direct `bedrock:InvokeModel` on the bare model id once an inference profile has been provisioned for it — the bare model id must be replaced with the profile's ARN. For Bedrock **chat** models this works fine through `bedrock/converse/<arn>` (the Converse API is provider-agnostic). For **embeddings**, though, `litellm`'s `get_bedrock_embedding_provider(model)` needs to pattern-match a recognizable provider name (`cohere`/`amazon`/`twelvelabs`/`nova`) out of the `model` string to pick the right request/response shape — and an application inference profile ARN is an opaque, account-scoped id with no such recognizable text in it at all.

`litellm` already supports exactly this split via its own `model_id` param (`unencoded_model_id = optional_params.pop("model_id", None) or model` in `litellm/llms/bedrock/embed/embedding.py`) — pass a recognizable `model` for provider detection, and a separate `model_id` for the actual invoke target. `LiteLLMSDKEmbeddings`, however, never exposes this split: it only ever builds `embed_kwargs["model"]` from a single configured string.

### Reproduction

```bash
export HINDSIGHT_API_EMBEDDINGS_PROVIDER=litellm-sdk
export HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL="bedrock/arn:aws:bedrock:eu-west-1:<account>:application-inference-profile/<id>"
# ... other required env vars ...
hindsight-api
```

Startup fails during embeddings dimension detection with the error above.

### Expected behavior

Setting a separate model id/ARN for the actual Bedrock invoke target (distinct from the model string used for provider detection) should let embeddings work through an application inference profile, the same way `bedrock/converse/<arn>` already works for chat models.

### Environment

- `hindsight-api` image: `ghcr.io/vectorize-io/hindsight-api:latest` (also reproduced against the `hindsight-api-slim` source directly)
- Embeddings provider: `litellm-sdk`
- Bedrock model: `amazon.titan-embed-text-v2:0` via an application inference profile

### Proposed fix

I've opened a PR (#TODO) adding an optional `model_id` param to `LiteLLMSDKEmbeddings` (threaded through to litellm's own `model_id` kwarg), plus a new `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL_ID` env var to configure it — so `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL` stays a recognizable model id (e.g. `bedrock/amazon.titan-embed-text-v2:0`) for provider detection, while `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL_ID` carries the actual ARN for the real invoke call.

gourangasatapathyvit:
Fixes #4034.

## Summary

`LiteLLMSDKEmbeddings` only ever built `embed_kwargs["model"]` from a single configured string. litellm's own bedrock embedding path needs that string to contain a recognizable provider name (`cohere`/`amazon`/`twelvelabs`/`nova`) for `get_bedrock_embedding_provider()` to pick the right request/response shape — which an AWS Bedrock **application inference profile ARN** (an opaque, account-scoped id) never has.

litellm already supports splitting this via its own `model_id` param (`litellm/llms/bedrock/embed/embedding.py`: `unencoded_model_id = optional_params.pop("model_id", None) or model`) — a recognizable `model` for provider detection, a separate `model_id` for the actual invoke target. This PR just wires that existing litellm capability through `LiteLLMSDKEmbeddings`.

## Changes

- `LiteLLMSDKEmbeddings.__init__` gains an optional `model_id: str | None = None` param, threaded into `embed_kwargs["model_id"]` in both places an embedding call is built (dimension-detection in `initialize()`, and the real per-batch call in `_encode_with_input_type()`).
- New `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL_ID` env var (`config.py`) to configure it, wired through `create_embeddings_from_env()`'s `litellm-sdk` branch.
- `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL` keeps its existing meaning (and default) unchanged — it's only ever used for provider detection now when `MODEL_ID` is also set; existing deployments that don't set the new var see no behavior change at all.

## Verification (live)

Built and ran the patched `hindsight-api-slim` in the project's own `docker/standalone/Dockerfile` (`--target api-only`), against a real AWS account with an org-level Bedrock SCP that explicit-denies direct `InvokeModel` on the bare model id once an inference profile exists for it:

```
HINDSIGHT_API_EMBEDDINGS_PROVIDER=litellm-sdk
HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL=bedrock/amazon.titan-embed-text-v2:0
HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL_ID=arn:aws:bedrock:eu-west-1:<account>:application-inference-profile/<id>
```

**Before** (same config, without this PR's `model_id` support — `HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL` set directly to the ARN, since there was no other way to route through it at all):

```
RuntimeError: Failed to initialize LiteLLM SDK embeddings: litellm.APIConnectionError:
Unable to determine bedrock embedding provider for model:
arn:aws:bedrock:eu-west-1:<account>:application-inference-profile/<id>.
Supported providers: ['cohere', 'amazon', 'twelvelabs', 'nova']
ERROR:    Application startup failed. Exiting.
```

Container crash-looped on every restart attempt.

**After** (this PR, `MODEL` + `MODEL_ID` split as above) — full startup log:

```
Starting Hindsight API...
  Version: v0.9.2
  Database: postgresql://***:***@XXXX:5432/POSTGRES_app
  LLM: bedrock / bedrock/converse/arn:aws:bedrock:eu-west-1:<account>:application-inference-profile/<id>
  Embeddings: litellm-sdk
  Reranker: rrf
  Extensions: pgvector (vector) / native (text)
  MCP: enabled at /mcp

INFO:     Started server process [10]
INFO:     Waiting for application startup.
... - INFO - hindsight_api.engine.embeddings - Embeddings: initializing LiteLLM SDK provider with model bedrock/amazon.titan-embed-text-v2:0
... - INFO - hindsight_api.engine.embeddings - Embeddings: LiteLLM SDK provider initialized (model: bedrock/amazon.titan-embed-text-v2:0, dim: 1024)
... - INFO - hindsight_api.engine.providers.litellm_llm - LiteLLM connection verified successfully (response truncated)
... - INFO - hindsight_api.engine.memory_engine - Running database migrations...
... - INFO - hindsight_api.migrations - Database migrations completed successfully for schema 'hindsight'
... - INFO - hindsight_api.engine.db.postgresql - PostgreSQL pool created (min=5, max=100, cmd_timeout=60s, acquire_timeout=30s, session_setup_on_acquire=True)
```

`docker ps` confirmed the container stayed up (no restart) well past the point it previously crash-looped.

## Notes

- I wasn't able to run `ruff`/`ty` locally (this machine's Rust/MSVC linker setup can't build `litellm`'s native extension from source on Windows — unrelated to this change), but built and ran the patched `hindsight-api-slim` inside the project's own `docker/standalone/Dockerfile` (Linux, prebuilt wheels) to verify end-to-end.
- No existing behavior changes for anyone not setting the new env var.

### DIFF
--- a/hindsight-api-slim/hindsight_api/config.py
+++ b/hindsight-api-slim/hindsight_api/config.py
@@ -474,6 +474,11 @@ ENV_RERANKER_LITELLM_MAX_TOKENS_PER_DOC = "HINDSIGHT_API_RERANKER_LITELLM_MAX_TO
 # LiteLLM SDK configuration (direct API access, no proxy needed)
 ENV_EMBEDDINGS_LITELLM_SDK_API_KEY = "HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_API_KEY"
 ENV_EMBEDDINGS_LITELLM_SDK_MODEL = "HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL"
+# Optional explicit model id/ARN forwarded to litellm as its own `model_id` param
+# (the actual Bedrock invoke target), separate from ENV_EMBEDDINGS_LITELLM_SDK_MODEL
+# (used only for litellm's provider-family detection). Needed for AWS Bedrock
+# application inference profiles — see LiteLLMSDKEmbeddings's own docstring.
+ENV_EMBEDDINGS_LITELLM_SDK_MODEL_ID = "HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_MODEL_ID"
 ENV_EMBEDDINGS_LITELLM_SDK_API_BASE = "HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_API_BASE"
 ENV_EMBEDDINGS_LITELLM_SDK_OUTPUT_DIMENSIONS = "HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_OUTPUT_DIMENSIONS"
 ENV_EMBEDDINGS_LITELLM_SDK_ENCODING_FORMAT = "HINDSIGHT_API_EMBEDDINGS_LITELLM_SDK_ENCODING_FORMAT"
@@ -2640,6 +2645,7 @@ class HindsightConfig:
     embeddings_litellm_dimensions: int | None
     embeddings_litellm_sdk_api_key: str | None
     embeddings_litellm_sdk_model: str
+    embeddings_litellm_sdk_model_id: str | None
     embeddings_litellm_sdk_api_base: str | None
     embeddings_litellm_sdk_output_dimensions: int | None
     embeddings_litellm_sdk_encoding_format: str | None
@@ -3847,6 +3853,7 @@ class HindsightConfig:
             embeddings_litellm_sdk_model=os.getenv(
                 ENV_EMBEDDINGS_LITELLM_SDK_MODEL, DEFAULT_EMBEDDINGS_LITELLM_SDK_MODEL
             ),
+            embeddings_litellm_sdk_model_id=os.getenv(ENV_EMBEDDINGS_LITELLM_SDK_MODEL_ID) or None,
             embeddings_litellm_sdk_api_base=os.getenv(ENV_EMBEDDINGS_LITELLM_SDK_API_BASE) or None,
             embeddings_litellm_sdk_output_dimensions=int(v)
             if (v := os.getenv(ENV_EMBEDDINGS_LITELLM_SDK_OUTPUT_DIMENSIONS))
--- a/hindsight-api-slim/hindsight_api/engine/embeddings.py
+++ b/hindsight-api-slim/hindsight_api/engine/embeddings.py
@@ -1650,6 +1650,7 @@ class LiteLLMSDKEmbeddings(Embeddings):
         self,
         api_key: str | None = None,
         model: str = DEFAULT_EMBEDDINGS_LITELLM_SDK_MODEL,
+        model_id: str | None = None,
         api_base: str | None = None,
         output_dimensions: int | None = None,
         batch_size: int = 100,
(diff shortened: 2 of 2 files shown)


nicoloboschi:
fixed by #4114

PR #4037, opened 2026-09-02
feat: run Hindsight on free-threaded CPython 3.14 (-py3.14t image)

nicoloboschi:
Makes Hindsight run on free-threaded CPython 3.14 (`python3.14t`) and ships it as a separate image tag, while 3.11 stays the supported default.

> **Stacked on #4033.** This branch carries that PR's migration-isolation files byte-for-byte so they drop cleanly on rebase. Merge #4033 first.

## Why

Recall does very little CPU work per request — but "very little" is not zero, and on a single event loop it all serialises. Measured on real recall against Postgres 18 + pgvector, 2100 memory units, identical workload verified on all three servers (142 results, `{world: 129, observation: 13}`), interleaved, 64 concurrent, best of 2:

| config | rps | server cores | p50 | p99 |
|---|---|---|---|---|
| 3.11 · 1 loop (today) | 23.2 | 0.93 | 2187 ms | 2970 ms |
| 3.14t · 1 loop | 42.0 | 1.03 | 1286 ms | 1886 ms |
| **3.14t · 8 loops** | **82.2** | **5.61** | **597 ms** | **1582 ms** |

**3.5× throughput and 3.7× lower p50**, with 5.61 cores genuinely in use.

Worth splitting the credit honestly: 1 loop → 1 loop is the **3.14 interpreter** (~1.8×, available from plain 3.14 too). 1 loop → 8 loops is **free-threading** (~2×). The multi-loop server is not wired up in this PR — see Not included.

## The core hazard

A free-threaded build **silently re-enables the GIL** when it imports a C extension that has not declared `Py_MOD_GIL_NOT_USED`. The only signal is a `RuntimeWarning`. Nothing crashes; the process just runs single-threaded. So "it works" is not evidence, and three layers assert it instead:

- `HINDSIGHT_API_FREE_THREADING=strict` (default on a free-threaded build) turns the warning into an error at the offending import, and refuses to start if the GIL is already on.
- The Docker build imports the whole API under `PYTHONWARNINGS="error:..."` and fails rather than shipping a mis-tagged image.
- The CI job asserts the GIL is off before running a single test.

## What's in it

**Loop-agnostic shared state.** Four process-wide `asyncio` primitives bound to whichever loop first waited on them, so every other loop got `RuntimeError: ... is bound to a different event loop` — under load only, which is why nothing caught them. The caches and budget manager move to `threading.Lock` (every critical section they guard is await-free); the LLM caps and llama.cpp server lock move to new `CrossLoopSemaphore`/`CrossLoopLock`, which are safe to hold across `await`. Caps stay **process-wide** — `--workers N` has always meant N caps — rather than silently becoming N × loops against your provider. The cache's in-flight map is keyed by `(loop, key)` so coalescing is per-loop while the cached data stays shared.

**A `-py3.14t` image.** `docker build --target api-only-freethreaded`. Own stages, so the 3.11 path is untouched. No official free-threaded Python image exists, so uv installs the interpreter. local-ml is refused outright with an error rather than producing an image that looks free-threaded and isn't (importing `sentence_transformers` re-enables the GIL).

**Guards that run on 3.11.** Most of what breaks is *multi-loop*, not free-threading, and that reproduces on 3.11 as soon as two loops exist in one process. `test_multi_loop_conformance.py` covers the cache, budget manager, LLM caps and dateparser in the ordinary suite — verified as a real guard by restoring the pre-fix cache, which makes it fail on 3.11 with `got Future ... attached to a different loop`.

**Code-review rules.** A "Supported interpreters" section stating both 3.11 and 3.14t are supported, a Concurrency section on which lock to use (the choice is *ownership*, not style), and review step 11d with the greps.

## Known limitations

- **orjson has no `cp3XXt` wheel** and its build refuses free-threading, so it is overridden out and `retain/types.py` falls back to the pre-orjson formatting. Correct — both spellings parse to the same float32 bytes — but **14× slower to render and 3.1× more bytes on the wire per vector** (23 → 325 µs on a 1024-dim embedding). Retain path only; recall never renders a vector this way. So this tag suits a recall-heavy deployment and not a retain-heavy one.
- **No local ML.** Embeddings and reranking must be remote (TEI/OpenAI/Cohere).
- `test_xai_oauth_llm::test_cleanup_closes_a_client_still_draining_from_a_recycle` fails on 3.14t — but also with `PYTHON_GIL=1` on the same binary, so it is a **3.14 asyncio change, not free-threading**, and a plain 3.14 upgrade would hit it too. Left unfixed and unattributed rather than papered over.

## Not included

The multi-loop launcher itself. Reaching the 8-loop numbers above needs a per-loop app factory (an asyncpg pool is per-loop) and a pool-sizing decision — I used a matched total connection budget. That is a deliberate follow-up: this PR makes the process *correct* under several loops; wiring uvicorn to actually run them is the next step.

## Testing

Full non-LLM suite on both interpreters against this tree, same image and same Postgres — results in the first comment.


nicoloboschi:
## Full suite, both interpreters

Same image, same Postgres 18 + pgvector, mock LLM, no local-ml on either side, run sequentially.

| | failed | passed | skipped | errors |
|---|---|---|---|---|
| **3.11** | 19 | **6024** | 1739 | 19 |
| **3.14t** | 21 | **6024** | 1737 | 19 |

Identical pass counts. The 19 errors are the same on both (pg0 embedded-Postgres in this harness) and predate the branch.

### Failing only on 3.14t — all three are 3.14, not free-threading

```
tests/test_structured_delta_prompt_budget.py::test_build_structured_delta_prompt_truncates_huge_document
tests/test_structured_delta_prompt_budget.py::test_fit_structured_delta_keeps_small_prompt_unchanged
tests/test_xai_oauth_llm.py::test_cleanup_closes_a_client_still_draining_from_a_recycle
```

Re-ran the whole suite on the **same binary** with `PYTHON_GIL=1` and all three still fail. So they are 3.14 behaviour changes that a plain (GIL) 3.14 upgrade would hit identically, not anything this branch introduces. Left unfixed and attributed rather than papered over.

`tests/test_worker.py::TestConcurrentWorkers::test_concurrent_workers_claim_different_tasks` failed on **3.11** this run and on 3.14t the previous one — an xdist flake, not interpreter-related.

### Two real bugs the suite caught

Worth calling out, because both were invisible until the full run:

1. **The migration child inherited the free-threading guard.** It imports psycopg2 on purpose, the guard is strict by default on a free-threaded build, so the guard turned that import into an error and *every* migration failed — 1765 errored tests. Fixed by turning the guard off (and clearing `PYTHONWARNINGS`) inside the child. This is a stacking bug with #4033, not a bug in it: that PR has no guard to disable.

2. **The orchestration tests' opt-out never worked.** They set `HINDSIGHT_API_MIGRATION_ISOLATION=never`, but the flag is read through `get_config()`, whose result is cached in a module global — so the env var has no effect after any earlier call. They passed on 3.11 only by accident, because `auto` resolves to `never` there anyway. Now patched at the seam. **Worth folding into #4033**, whose tests are green for that same accidental reason, so the opt-out is not actually exercised there.

### Housekeeping

Rebased onto main. Three commits dropped as upstream — the lazy imports (#4031), the dateparser fix (#4032), and my toktok commit, superseded by #4022 (a cleaner single-backend implementation than the two-backend fallback I had). The quicktok CI build step and its script are gone, since toktok ships `cp3XXt` wheels.

Verified on a clean PyPI install on 3.14t: GIL off, toktok present, quicktok absent, orjson overridden out with the fallback rendering correctly.

Staying in draft until #4033 lands, since this is stacked on it.

nicoloboschi:
## Rebased onto main (post-#4033), re-verified

Now targets `main` directly — no longer stacked. Merges cleanly.

Four commits dropped as upstream: the original migration subprocess and its orchestration-test fixes (now #4033), and the "take #4033 verbatim" commit that only existed to keep the stack aligned.

### Full suite, both interpreters

Same image, same Postgres 18 + pgvector, mock LLM, no local-ml either side, run sequentially.

| | failed | passed | skipped | errors |
|---|---|---|---|---|
| **3.11** | 19 | 6032 | 1739 | 19 |
| **3.14t** | 20 | 6033 | 1737 | 19 |

**One** test fails on 3.14t and not on 3.11:

```
tests/test_xai_oauth_llm.py::test_cleanup_closes_a_client_still_draining_from_a_recycle
```

It fails on the same binary with `PYTHON_GIL=1`, so it is a 3.14 asyncio scheduling change that a plain (GIL) 3.14 upgrade would hit identically — not something this branch introduces. Nothing fails only on 3.11.

The 19 errors are identical on both and are this harness missing pg0's embedded Postgres; they predate the branch.

### Two fixes carried across the rebase

Both are this branch's responsibility rather than #4033's, and both needed re-applying against the new `auto | true | false` spelling:

1. **The migration child must run with the guard off.** It imports psycopg2 on purpose, but the guard added here is inherited by the child and strict by default on a free-threaded build, so it turned that import into an exception and every migration failed. `PYTHONWARNINGS` is cleared for the same reason one step removed: the free-threaded CI job promotes that warning to an error and the child must not inherit it.

2. **The orchestration tests' opt-out still does not work as written on main.** They set `HINDSIGHT_API_MIGRATION_ISOLATION=false`, but the flag is read through `get_config()`, whose result is cached in a module global — so the env var has no effect after any earlier call. That goes unnoticed on 3.11, where `auto` resolves to no-isolation anyway, and fails on a free-threaded build where `auto` isolates and the patched step functions are never reached. Now patched at `_should_isolate_migrations` instead.

Point 2 is worth knowing independently of this PR: as it stands on main, the flag's opt-out path is not actually exercised by those tests on any interpreter.

nicoloboschi:
## Green — final state

Full CI run on `c737265c`: **109 jobs succeeded, 2 skipped, 0 failures**, including the new `test-api (free-threaded 3.14)` job end to end — suite, image build, and image smoke test.

Free-threaded suite: **6100 passed, 1745 skipped, 0 failed.**

### What CI found that local runs could not

The first real CI run had 64 failures. Worth recording why my local container runs missed them: I never ran the suite with `PYTHONWARNINGS` set, and my local venvs were not built with `--all-extras` the way CI's 3.11 job is.

- **51 were the job's own configuration.** It promoted the GIL re-enable warning to an error across the whole suite — but the suite imports LiteLLM on purpose to test that provider, and LiteLLM pulls `fastuuid`, which has no free-threaded build. ~50 tests failed for doing exactly what they are meant to do. The filter now applies only to the step that imports the API to assert the GIL is off; the property is still asserted in a subprocess probe, the image build, and the smoke test.
- **5 were a real 3.14 incompatibility** in test code: `asyncio.get_event_loop()` no longer auto-creates a loop, so `get_event_loop().run_until_complete(...)` raises. Now `asyncio.run`.
- **3 needed the `local-ml` extra**, which a free-threaded install cannot have. Now `importorskip` the thing they actually need.
- **1 was a missing `GITHUB_TOKEN`** in the job env.

### Two real bugs, fixed rather than skipped

**A connection leak on shutdown** (`xai-oauth`). `cleanup()` cancels each drain task and left the close to that task's own `finally` — but from 3.12 the cancellation is delivered at the task's next await, which *is* the `await stale.aclose()`, and `CancelledError` is a `BaseException` so the surrounding `suppress(Exception)` does not catch it. The client was never closed. `cleanup()` now closes retired clients itself. **Not a free-threading bug** — it reproduces with `PYTHON_GIL=1` on the same interpreter, so a plain 3.14 upgrade would hit it too. Regression test included, verified as a negative control (fails on 3.14t with the fix reverted, passes on 3.11 either way).

Two earlier theories were wrong and are recorded in the commit so nobody retries them: swallowing the `CancelledError` is not enough, and `Task.uncancel()` does not help either.

**An over-specified test** (`embedding_coalescer`). It compared flattened backend calls to the input list, pinning the order in which batches reach the backend. The coalescer runs up to `max_concurrent_requests` calls at a time and never promised that; under the GIL the interleaving was merely stable. Now compares a `Counter`, which keeps the property that matters — every text dispatched exactly once — and leaves the per-caller assertion intact.

**And one the image build caught:** litellm publishes only abi3 wheels, and abi3 does not apply to a free-threaded interpreter, so uv falls back to its sdist and needs Rust. Only amd64 hit it — my local builds were arm64. That is a decent argument for the CI job building the image at all.

koriyoshi2041 (commented):
I found one cross-loop stale-write race in `BankStatsCache.get_or_load(..., force_refresh=True)`.

Flights are now keyed by `(loop, key)`, but the force-refresh path only removes `self._flight_key(key)` for the loop performing the refresh. An older normal loader on a different loop keeps its slot, and when it completes it still passes:

```python
if self._in_flight.get(flight_key) is in_flight:
    self._store_unlocked(key, value)
```

so it overwrites the newer force-refreshed value.

I reproduced this on the current head (`c737265c`) with two threads / two `asyncio.run()` loops:

1. loop A starts a normal load returning `{"value": "old"}` and pauses inside the loader;
2. loop B completes a forced refresh returning `{"value": "new"}`;
3. loop A resumes and completes;
4. the next cache hit returns `old`.

The regression fails before the fix (`assert {'value': 'old'} == {'value': 'new'}`). Detaching every flight whose `fk[1] == key` in the force-refresh critical section fixes it, matching the existing all-loop treatment in `invalidate()`. With that change, the full `test_bank_stats_cache.py` plus `test_multi_loop_conformance.py` pass: 18 tests.

This matters specifically for the PR's new multi-loop model: a forced stats read is intended to supersede an in-flight snapshot process-wide, not only on its caller's loop.

PR #4038, opened 2026-09-02
Merging latest fork

plotneishestvo:
most likely fix to MCP that I was nnually applying before


plotneishestvo:
oops sorry, i was thinkging mergin other way around :D

PR #4047, opened 2026-09-02
fix(parsers): reject low-quality OCR before it becomes memory evidence

### LINKED ISSUE #3897: OCR output quality: nonempty refusals and unreadable fragments can become memory evidence
## Bug Description

MarkItDown OCR currently treats every nonempty model response as a successful conversion. Low-quality outputs such as:

- `I can't read this image`
- `No visible text`
- output consisting mostly of `[unclear]`
- isolated timestamps, button labels, or navigation chrome
- a few unrelated words from a screenshot

can therefore continue into semantic fact extraction and become memory evidence.

This is different from the empty-output and scanned-PDF cases. The OCR call technically succeeds and returns text, but the text is not useful evidence.

The default prompt also tells the model to emit `[unclear]` for unreadable spans, while the parser only checks whether `result.text_content` is empty:

- Prompt: https://github.com/vectorize-io/hindsight/blob/d2e8fa7286b37012b7e45de8f1b10dc333274918/hindsight-api-slim/hindsight_api/config.py#L1279-L1282
- Parser check: https://github.com/vectorize-io/hindsight/blob/d2e8fa7286b37012b7e45de8f1b10dc333274918/hindsight-api-slim/hindsight_api/engine/parsers/markitdown.py#L168-L180

## Steps to Reproduce

1. Enable MarkItDown OCR with a configured vision-capable endpoint.
2. Upload an image for which the endpoint returns a nonempty refusal, no-text response, or mostly unreadable transcription.
3. `MarkitdownParser._convert_sync()` returns that response because it only rejects missing or empty `text_content`.
4. File conversion is treated as successful and the output continues into Retain.

A deterministic unit reproduction can stub `self._markitdown.convert()` to return `text_content="No visible text"` or `text_content="[unclear] [unclear]"`.

## Expected Behavior

Hindsight should evaluate OCR output quality after conversion but before semantic extraction.

If an OCR result fails the quality gate:

1. Reject that parser result with a structured reason.
2. Continue to the next configured parser in the fallback chain.
3. If no parser succeeds, fail the file-retain operation without creating facts or observations.

## Actual Behavior

Any nonempty OCR response is accepted, including refusals and unreadable fragments.

The existing `on_file_convert_complete` hook cannot enforce an admission decision because hook exceptions are caught and logged before Retain continues:

https://github.com/vectorize-io/hindsight/blob/d2e8fa7286b37012b7e45de8f1b10dc333274918/hindsight-api-slim/hindsight_api/engine/memory_engine.py#L2758-L2781

## Proposed Direction

Add a deterministic, parser-independent OCR quality validator to the parser fallback path. It should run after a parser returns text and before that parser is declared successful.

Useful signals include:

- normalized character, word, and unique-word counts
- alphabetic and numeric content ratio
- `[unclear]` and replacement-character ratio
- repetition ratio
- explicit refusal or no-text phrases
- timestamp, button-label, and UI-chrome density
- useful-sentence density
- provider OCR confidence when available

Use hard rejection only for strong signals such as explicit refusals, zero meaningful tokens, or output dominated by uncertainty markers. Sparse output should require multiple bad signals rather than a character threshold alone because legitimate images may contain only a verification code, amount, sign, or error code.

Return reason codes and measurements for observability, but do not log the OCR body.

## Acceptance Criteria

- Explicit refusals and no-text responses do not reach fact extraction.
- Output dominated by `[unclear]` or replacement characters does not reach fact extraction.
- A rejected parser result advances to the next configured parser.
- If every parser fails, no facts or observations are created.
- Legitimate short OCR output is not rejected solely because it is short.
- Tests cover hard rejects, compound low-quality rejects, legitimate sparse text, fallback behavior, and terminal failure.

## Related

- #927 and merged PR #2145 added optional MarkItDown OCR but not output-quality validation.
- #3255 covers image-only PDFs that return no content.
- Closed PR #3442 proposed scanned-PDF OCR fallback, not validation of nonempty OCR output.
- #2725 concerns first-class visual-semantic image memory, which is broader than this OCR admission problem.

leilei3167:
## Problem

MarkItDown OCR currently treats any nonempty `text_content` as a successful conversion. Refusals (`I can't read this image`), no-text replies (`No visible text`), and outputs dominated by `[unclear]` / replacement characters can therefore continue into Retain and become memory evidence.

The default OCR prompt even asks the model to emit `[unclear]` for unreadable spans, while the parser only checks emptiness.

Fixes #3897

## Approach

Add a deterministic, parser-independent OCR quality gate on the parser fallback path — after a parser returns text, before that result is declared successful:

- **Hard reject** on strong signals: explicit refusal / no-text phrases, zero meaningful tokens, output dominated by `[unclear]` or replacement characters, navigation-chrome-only dumps.
- **Compound reject** for weaker signals (timestamp-dense chrome, high repetition, low residual content): require multiple independent bad signals so sparse legitimate OCR (verification codes, amounts, signs, error codes) is not dropped for short length alone.
- On reject: log reason codes + measurements only (never the OCR body), advance to the next parser; if every parser fails, conversion fails and no facts/observations are created.
- Gate applies to image filenames (`.jpg` / `.jpeg` / `.png`); non-image conversions are unchanged.

## Test plan

- [x] Unit tests in `tests/test_ocr_output_quality.py`: hard rejects, compound low-quality, legitimate sparse text, fallback advance, terminal failure, MarkItDown `convert()` stubs
- [x] `uv run pytest tests/test_ocr_output_quality.py tests/test_markitdown_parser.py` — green locally
- [ ] CI on this PR


leilei3167:
@nicoloboschi quick check-in on this one — OCR quality gate on the parser fallback path so nonempty refusals / `[unclear]`-heavy fragments do not become memory evidence. Happy to adjust thresholds or scope if you want it tighter/looser.

nicoloboschi:
this fix is not reliable, we'd need some hard flags on markitdown at most

PR #4052, opened 2026-09-02
chore(deps): bump torch to 2.13.0 and tornado to 6.5.8

cdbartholomew:
Closes 4 Dependabot alerts in the root `uv.lock`. Both packages live in the same lockfile, so they ship together rather than as two PRs that would collide.

## The bumps

| Package | From | To | Notes |
|---|---|---|---|
| `torch` | 2.10.0 | **2.13.0** | optional `local-ml` extra, declared `>=2.6.0` |
| `tornado` | — | **6.5.8** | declared `>=6.5.5` |

Both constraints are open-ended, so these are scoped `uv lock --upgrade-package` runs with **no manifest edits**. The combined diff against `main` is exactly torch and tornado — nothing else moved.

Two things worth noting on scope:

- **torch carries no CUDA variants here.** The lock has zero `nvidia-*` packages; torch resolves against the CPU-only index. The only split is platform: `2.13.0` on darwin, `2.13.0+cpu` elsewhere.
- **`tornado` is a floor-only pin with no runtime surface.** `hindsight-api-slim` is the only thing depending on it, and no module imports it.

`torch` was pinned to 2.13.0 rather than letting the resolver take 2.14.0, keeping the jump as small as the advisory allows.

## Verification

Re-run from scratch after adding tornado rather than reusing the torch-only run — re-synced `--all-extras` at both final versions:

**63 tests pass**, covering `test_local_device`, `test_model_load_default_dtype`, `test_local_model_alignment` (real forward passes and weight-alignment guards), `test_local_embeddings`, and `test_local_cross_encoder`.

`uv lock --check` clean.


cdbartholomew:
Re-ran against `main` after #4089, which added the root `uv.lock` to the `core` path filter.

**Before:** 3 checks ran, 91 skipped. `test-api` skipped. The torch 2.10 → 2.13 and tornado 6.5.7 → 6.5.8 bumps were reported green having been exercised by nothing.

**Now:** 49 checks ran, 0 failed.

```
test-api (1/3)              SUCCESS
test-api (2/3)              SUCCESS
test-api (3/3)              SUCCESS
test-api (free-threaded)    SUCCESS
test-integration            SUCCESS
test-pip-slim               SUCCESS
test-embed                  SUCCESS
test-embed-windows          SUCCESS
test-hindsight-all          SUCCESS
test-upgrade                SUCCESS
```

The bump is now actually verified rather than merely unopposed.

PR #4056, opened 2026-09-03
fix(reflect): retry split-synthesis map sentinel instead of reducing it

### LINKED ISSUE #4054: reflect: forced split-synthesis silently writes a fluent, wrong document when a map chunk returns "(no relevant evidence)"; sentinel is not size-bound; small context budget starves evidence retrieval
## Version

`hindsight_api` 0.9.2 (self-hosted, image digest `84ab276b…`). Reflect LLM is an OpenAI-compatible endpoint serving LongCat-2.0. Mental-model refresh via `POST /v1/default/banks/{bank}/mental-models/{id}/refresh`, `max_tokens=8192`, `trigger.keep_trace=true`.

## Summary

When a mental-model refresh exceeds the reflect context budget, `_forced_final_synthesis` (engine/reflect/agent.py) splits the accumulated tool results into budget-sized chunks, runs one claims-extraction call per chunk (`final_map_N`), then one reduce. Three problems observed on five banks over one evening:

1. **The map model sometimes answers a non-empty chunk with the literal sentinel `(no relevant evidence)`** (6 output tokens, `finish_reason=stop`). Observed on chunks of 17k, 22k, 58k, 69k, 77k and 85k input tokens; successes observed at 9k to 52k and at 78k. Not size-bound.
2. **A lost chunk is silent.** When every chunk returns the sentinel, the reduce writes a one-paragraph "I don't have information" and the operation completes with `outcome=content_written` (one bank: 176 items in `based_on`, document 782 chars). When one of two chunks is lost, the reduce writes a full, fluent, cited document from the surviving half; a reader with source access found it had inverted a verdict that the lost half contradicted. Nothing on the operation or the model distinguishes this from a complete build.
3. **A small context budget starves evidence retrieval.** With `HINDSIGHT_API_REFLECT_MAX_CONTEXT_TOKENS=40000` (default 100000; the variable is not on the configuration page's REFLECT table, only in `config.py`), on a bank whose plugin-generated knowledge pages are ~30k chars each, the agent's first iteration issued four `search_mental_models` calls, exceeded the budget in one step, and forced synthesis with zero `search_observations`/`recall` calls (`based_on`: world 0, experience 0, observation 0). The document was written from stale sibling pages. `trigger.exclude_mental_models=true` fixed it (same refresh then made four `search_observations` calls, 61 observations in `based_on`).

Two smaller things seen along the way: the reduce reproduced a database root password verbatim from a memory into the synthesised document (a query-level "never reproduce credentials" clause held on four banks and failed on one until the source memory was redacted); and one refresh attempt stalled with no outbound LLM request until the 900 s wall, the shape of #3881, before its retry completed.

## Asks

- Treat a sentinel on a non-empty chunk as a failed map call and retry it within the existing retry budget, rather than passing an empty claims section to the reduce.
- If any chunk is lost after retries, surface it: e.g. `reflect_response.outcome=partial_evidence` and a field on the operation, instead of `content_written`.
- Bound map chunk size independently of the retrieval budget (e.g. a `REFLECT_MAP_CHUNK_TOKENS`), so operators can keep a large retrieval budget with small map chunks.
- Keep the first-iteration `search_mental_models` results out of the context-budget check, or count them separately, so a bank with large knowledge pages can still reach evidence retrieval under a small budget.
- Document `HINDSIGHT_API_REFLECT_MAX_CONTEXT_TOKENS` alongside the other REFLECT variables.
- Consider a default no-credentials rule in the reduce system prompt.

## Repro shape

Bank with ~40k observations across several tag scopes and five ~30k-char mental models. Create a mental model with a broad multi-component `source_query`, `max_tokens=8192`, `keep_trace=true`; refresh with `trigger.mode=full`. Read `reflect_response.trace.llm_calls` for `final_map_N`, and the LLM request log (`GET .../llm-requests`) for map calls with `output_tokens<=10` and output `(no relevant evidence)`. With `REFLECT_MAX_CONTEXT_TOKENS=40000`, read `trace.tool_calls` for a first iteration consisting only of `search_mental_models`, then repeat with `trigger.exclude_mental_models=true`.

leilei3167:
## Problem

When forced split-synthesis overflows the reflect context budget, `_forced_final_synthesis` maps each chunk into claims and then reduces. The map system prompt tells the model to answer with the exact sentinel `(no relevant evidence)` when a chunk has nothing relevant. In production that sentinel also shows up on **non-empty** chunks (not size-bound). The old path still fed those sentinel strings into reduce, so:

- every chunk sentinel → a fluent "I don't have information" written as `content_written`
- one of N chunks sentinel → a fluent, cited document from the surviving half, silently missing contradictory evidence

Fixes #4054

## Approach

Focused on the sentinel / map-chunk ask (#4054 ask 1); map-chunk size caps, retrieval-budget starvation, and docs left for follow-ups.

- Share `NO_RELEVANT_EVIDENCE_SENTINEL` + `is_no_relevant_evidence_sentinel()` next to `CLAIMS_SYSTEM_PROMPT`.
- On a **non-empty** map chunk that returns the sentinel: retry once (same one-retry budget as consecutive LLM errors), with `final_map_N_retry1` in the LLM trace.
- Never pass the sentinel string into `build_reduce_prompt`; omit still-lost chunks and warn.
- If every chunk is lost after retries: raise `ReflectNoAnswerError` instead of reducing into a wrong document.

## Test plan

- [x] Extended `tests/test_reflect_split_synthesis.py`: sentinel matcher; retry reaches reduce without the sentinel; exhausted retries drop the chunk; all-lost raises `ReflectNoAnswerError`
- [x] `uv run pytest tests/test_reflect_split_synthesis.py -k 'not RealLLM'` — 16 passed locally
- [ ] CI on this PR


leilei3167:
@nicoloboschi quick check-in — split-synthesis map path now retries the empty-evidence sentinel on non-empty chunks and never feeds it into reduce (raises if every chunk is lost). Happy to adjust retry count or follow up on map-chunk size / partial_evidence outcome if you want those in a separate PR.

nicoloboschi:
fixed by #4115

PR #4057, opened 2026-09-03
fix(docker): drop runtime curl and NodeSource bootstrap

rschlek:
## Summary

- replace runtime curl health checks with a small helper that uses Python in API images and BusyBox wget in the Alpine control-plane image
- reuse the Node binary and license from the existing control-plane builder in the standalone image
- remove the NodeSource bootstrap and its runtime-only GnuPG, system Python 3.13, and curl dependency chains
- cover both HTTP probe implementations, including non-success responses, in the standalone startup tests

## Security impact

Trivy 0.74.0 scans with the current database showed:

- API-only: 4 Critical / 25 High with curl added back to the exact artifact, versus 4 Critical / 16 High without it - 9 High findings removed
- cp-only: 0 Critical / 0 High
- standalone: 4 Critical / 39 High with the NodeSource bootstrap, versus 4 Critical / 16 High when reusing the builder Node binary - 23 High findings removed

The remaining findings are unrelated base-image or Python dependency findings and are not suppressed by this change.

## Verification

- `docker/standalone/test-start-all.sh` passes in Alpine with Python and BusyBox wget
- API-only health reports the database connected
- cp-only health passes
- standalone API and control-plane health both pass
- final standalone image contains Node 20.20.2 and excludes curl, npm, GnuPG, and Debian Python 3.13
- `git diff --check` passes

The full local lint hook was also run. Its only failures were existing host-environment issues: the local control-plane install lacks `@eslint/js`, and Windows type analysis cannot resolve optional Unix/macOS modules and APIs.

### DIFF
--- a/docker/standalone/Dockerfile
+++ b/docker/standalone/Dockerfile
@@ -168,7 +168,6 @@ WORKDIR /app
 # commit can now resolve different package versions.
 RUN apt-get update && apt-get upgrade -y \
     && apt-get install -y \
-    curl \
     procps \
     libxml2 \
     libssl3 \
@@ -275,8 +274,10 @@ WORKDIR /app
 COPY docker/standalone/start-all.sh /app/start-all.sh
 RUN chmod +x /app/start-all.sh
 
-# Install curl for health checks. `apk upgrade` first, for the same reason the
-# Debian stages run `apt-get upgrade` - see the comment on the api-only stage.
+# Install bash for the shared startup script. HTTP readiness probes use wget
+# from the Alpine base, so curl and its transitive dependencies do not ship.
+# `apk upgrade` runs first for the same reason the Debian stages run
+# `apt-get upgrade` - see the comment on the api-only stage.
 # npm is removed for the same reason the Debian stages remove pip's build
 # tooling: nothing invokes it at runtime - start-all.sh launches the pre-built
 # Next standalone bundle with `node server.js` - and dropping it leaves no
@@ -287,7 +288,7 @@ RUN chmod +x /app/start-all.sh
 # The `[ -e ]` guard is deliberate: a bare `rm -rf` on a path a future base
 # image no longer uses succeeds silently and npm quietly ships again. Fail the
 # build instead, so the path assumption has to be revisited on a base bump.
-RUN apk upgrade --no-cache && apk add --no-cache curl bash \
+RUN apk upgrade --no-cache && apk add --no-cache bash \
     && [ -e /usr/local/lib/node_modules/npm ] \
     && rm -rf /usr/local/lib/node_modules/npm /usr/local/bin/npm /usr/local/bin/npx \
     && ! command -v npm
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Thanks for this — the analysis behind it is solid, and the measurements hold up. I rebuilt the images locally and reproduced your numbers: dropping `curl` from `api-only` takes it from 3 Critical / 60 High to 3 Critical / 51 High, exactly the 9 High you reported, with zero new findings. `curl` really is the sole reverse-dependency of `libcurl4t64`, which brings `libssh2-1t64` along, and all nine are `status=affected` with no Debian fix available — so removing the package genuinely is the only remediation.

I'm closing this PR, but not because the work was wrong. It bundles three changes with very different risk profiles, and the safest one is being held up by the riskiest:

- **#4196 — drop the NodeSource bootstrap, copy Node from `cp-builder`.** The strongest part of this PR by a distance, and I don't think the description sells it hard enough. Comparing installed package sets, NodeSource is responsible for ten GnuPG packages (only ever used to verify the repo key at build time) plus a *second* system Python — `python3.13` inside a `python:3.11-slim` image. Plus it removes a `curl | bash` from a third-party host at build time and makes builds reproducible. No behaviour risk; your `RUN node --version && ! command -v npm` guard is exactly the right way to prove it.
- **#4197 — drop curl from `cp-only`.** Free. That image performs no health check at all: `start-all.sh`'s only curl calls are the API readiness loop (gated on `ENABLE_API=true`, which `cp-only` sets to false) and `check_llm` (gated on the opt-in `HINDSIGHT_WAIT_FOR_DEPS`). The comment above the `apk add` describes behaviour the image doesn't have.
- **#4198 — drop curl from the API images.** The only part that needs the hand-rolled probe, and it has a bug worth catching before it ships.

On that last point, the probe isn't equivalent to what it replaces. Measured exit codes:

| response | `curl -sf` | this PR's `urllib` | `http.client` |
| --- | --- | --- | --- |
| 302 → 404 | **0** | **1** | 0 |
| 404 / 500 | 22 | 1 | 1 |

`curl -sf` without `-L` doesn't follow redirects, so a 302 is a completed transfer and succeeds. `urlopen` follows it and raises on the 404 — which turns a healthy service into "not ready". The code comment says the implementations "follow redirects" as though matching curl; curl does the opposite. Using `http.client` and testing `status < 400` directly gets exact parity.

Two smaller things for #4198: the busybox wget branch can't be salvaged (BusyBox 1.37.0 in `node:*-alpine` has no `--max-redirect`, so it always follows) — but it's also unnecessary once `cp-only` needs no probe. And `WGET_BIN="$(command -v wget)"` in the test aborts the whole script under `set -euo pipefail` on any machine without wget, after which `wait "$HTTP_SERVER_PID"` blocks forever because the test server is waiting for four requests it'll never get.

All three issues carry the full reasoning and the measurements. I'm picking up #4196 and #4197 now since they're low-risk; #4198 is open if you'd like to take it — and if you'd rather carry any of the three yourself, say so and it's yours.

PR #4064, opened 2026-09-03
fix(cli): print the server's 404 detail instead of "API endpoint not found"

### LINKED ISSUE #4049: `document get` on a missing document prints "API endpoint not found (404)" — the server's `Document not found` detail is discarded
## Summary

`hindsight document get <bank> <doc-id>` on a document that does not exist prints:

```
✗ API endpoint not found (404)

API URL:
  http://<api-host>:8888
Possible causes:
  • The API endpoint path has changed
  • You're using an incompatible API version
```

but the server's actual response is `HTTP 404` with body `{"detail":"Document not found"}`. The CLI discards the body and reports a generic "endpoint not found", which points the operator at the wrong problem (API path / version) when the real answer is "this document does not exist".

## Reproduce

- CLI 0.8.5 against API 0.8.5 (the mapping is unchanged on `main` as of v0.9.2 — see below).
- `hindsight document get <bank> <any-doc-id-that-does-not-exist>` → the message above.
- The same request via curl: `GET /v1/default/banks/<bank>/documents/<doc-id>` → `404`, body `{"detail":"Document not found"}`.

## Where

`hindsight-cli/src/errors.rs`, the `if err_str.contains("404")` branch: it special-cases only the "Bank configuration API is disabled" body and otherwise emits the fixed "API endpoint not found (404)" text with endpoint-path/version guidance. Diffing `errors.rs` at `v0.8.5` against `main` shows the 404 branch logic unchanged; the #2912 fix added a 400 branch that *does* surface the server's `detail`, which is exactly the behavior this branch is missing.

## Why it matters

Operators use `document get` as the landing witness for async (batch) retains: a 404 there legitimately means "not landed yet / never landed". With the current wording, a 404 in that check reads as an API breakage rather than an absent document, so a missing document can be misdiagnosed — or a genuinely absent document can be waved through as a transport problem.

## Expected

Mirror the 400 branch: when the 404 body carries a `detail`, print it — e.g.

```
✗ Not found (404): Document not found
```

and fall back to the endpoint-path/version guidance only when the body has no `detail` (a true unknown-route 404).

Sibling of #2912 (same class: the CLI swallowing a self-explanatory server body).

oldnicke:
## Summary

Fixes #4049. `hindsight document get <bank> <missing-doc>` printed the fixed `API endpoint not found (404)` text with endpoint-path/version guidance and threw away the server's `{"detail":"Document not found"}`, sending the operator after an API breakage that isn't there.

An absent resource and an unknown route are both 404, but only the second one is about paths and versions — and the server already distinguishes them: it sends a `detail` for the first and nothing for the second. So the 404 branch now prints the detail when there is one and keeps the existing hints for a body-less 404.

```
✗ Not found (404): Document not found
```

## Changes

`hindsight-cli/src/errors.rs`:

- `server_detail()` pulls a non-empty string `detail` out of the error body. It returns `None` for a body-less or detail-less 404, which is what keeps the unknown-route guidance intact for the case it was written for.
- The 404 branch prints `Not found (404): <detail>` when a detail is present. The `Bank configuration API is disabled` branch is checked first and keeps its own message.

Why the raw body is available here: the document endpoints declare only `200`/`422` in the OpenAPI spec, so a 404 is undeclared and arrives as progenitor's `UnexpectedResponse`; `api.rs::humanize_client_error` then formats the error as `API request failed (404 Not Found): <raw body>`. A typed error body would serialize to the same `{"detail":...}` shape, so both paths parse.

The change is in the shared formatter, so every command benefits — a missing bank now reads `Not found (404): Bank not found` rather than "the endpoint moved".

## Testing

`cargo test` in `hindsight-cli`: 152 passed (106 unit + 25 + 13 + 8 integration), including three new cases:

- 404 with a `detail` reports the absent resource and no longer mentions the endpoint path or API version.
- 404 with no body still prints `API endpoint not found (404)` and the version hint.
- 404 for the disabled bank-config API still explains `HINDSIGHT_API_ENABLE_BANK_CONFIG_API=true` (regression guard for the branch above it).

`cargo fmt --check` clean; `cargo clippy --all-targets` reports nothing on `errors.rs` (the repo's pre-existing warnings elsewhere are untouched); `./scripts/hooks/lint.sh` passes.

Not exercised against a live server — the fix is in error-string formatting, and the tests feed it the exact strings `api.rs` produces for the reported case (`404 Not Found` + `{"detail":"Document not found"}`, as observed via curl in the issue).

### DIFF
--- a/hindsight-cli/src/errors.rs
+++ b/hindsight-cli/src/errors.rs
@@ -5,6 +5,19 @@ pub fn handle_api_error(err: anyhow::Error, api_url: &str) -> ! {
     std::process::exit(1);
 }
 
+/// The server's own explanation out of an error body, i.e. FastAPI's `{"detail": "..."}`.
+///
+/// The error text these branches match on is `API request failed (<status>): <body>` (see
+/// api.rs::humanize_client_error), so the body is whatever JSON the server sent — or nothing at
+/// all, which is what a genuine unknown-route 404 looks like. Returns None unless a non-empty
+/// string `detail` is actually there, so callers can keep their generic guidance for that case.
+fn server_detail(err_str: &str) -> Option<String> {
+    let body = &err_str[err_str.find('{')?..];
+    let parsed: serde_json::Value = serde_json::from_str(body.trim()).ok()?;
+    let detail = parsed.get("detail")?.as_str()?.trim();
+    (!detail.is_empty()).then(|| detail.to_string())
+}
+
 fn format_error_message(err: &anyhow::Error, api_url: &str) -> String {
     let err_str = err.to_string();
 
@@ -94,6 +107,19 @@ fn format_error_message(err: &anyhow::Error, api_url: &str) -> String {
                     .bright_white(),
                 "Note:".bright_cyan(),
                 "This allows per-bank LLM configuration overrides via API".bright_white()
+            );
+        }
+
+        // An absent resource and an unknown route are both 404, and only the second one is about
+        // API paths and versions. The server distinguishes them for us: `document get` on a
+        // missing document answers `{"detail":"Document not found"}`, so printing the endpoint/
+        // version guidance over that body sends the operator after an API breakage that isn't
+        // there (#4049) — the same swallowed-body class #2912 fixed for 400.
+        if let Some(detail) = server_detail(&err_str) {
+            return format!(
+                "{} {}",
+                "✗".bright_red().bold(),
+                format!("Not found (404): {}", detail).bright_red().bold()
             );
         }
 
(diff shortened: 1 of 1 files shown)


koriyoshi2041 (commented):
The new detail path also catches FastAPI's default unknown-route response, so the endpoint/version diagnostic is no longer preserved for the normal unknown-route case.

I checked this against this PR's exact head (`f1840fea`):

```text
TestClient(FastAPI()).get("/missing-route")
404 application/json {"detail":"Not Found"}
```

That string reaches `server_detail()`, producing `Not found (404): Not Found`; the new body-less test does not model FastAPI's default 404. Could the generic route detail be excluded (at least exact `"Not Found"`, ideally with a regression test using `API request failed (404 Not Found): {"detail":"Not Found"}`), while still surfacing resource-specific details such as `Document not found`?

The three added focused Rust tests pass locally; this is an uncovered response-shape boundary rather than a failure in those tests.

nicoloboschi:
fixed by #4113

PR #4066, opened 2026-09-03
fix(reflect): keep tools schema stable across forced turns and synthesis

### LINKED ISSUE #3865: Reflect's per-iteration tools narrowing (and final system-prompt swap) forces a full prefill on hybrid/Ollama models
## Summary

On Ollama (and any provider that silently drops `tool_choice="required"`),
Reflect's agent loop **changes the `tools` array on every forced iteration**
and **replaces the system prompt / drops tools for the forced final
synthesis**. Chat templates that render the `<tools>` block *before* the
system text (Qwen-style) therefore diverge at token ~32 (forced turns) or
token 3 (synthesis). Hybrid / Gated-DeltaNet models cannot resume from a
partial prefix — llama-server logs `forcing full prompt re-processing due
to lack of cache data (likely due to SWA or hybrid/recurrent memory)` —
so each iteration re-prefills the entire accumulated history.

Identical or purely *extended* prompts on the same host restore a context
checkpoint in ~0.2 s. Reflect never produces one of those.

This is the same class of miss reported for Gemini/Anthropic prompt cache
when `call_with_tools` narrows `tools` per forced turn
(Hellobibleapp/hindsight#2). v0.9.2 added cache markers for those
providers; the Ollama wire shape is unchanged.

## Where

Verified on `v0.8.5` and still present on `v0.9.2`:

- `hindsight-api-slim/hindsight_api/engine/reflect/agent.py`
  `forced_sequence` / `LLMToolChoice.named(search_mental_models |
  search_observations | recall)`, then `_forced_final_synthesis` via
  `build_final_system_prompt` with no tools.
  `call_with_tools`: named `tool_choice` **filters `tools` to that one
  function**, then unsets `tool_choice="required"` for
  `{"lmstudio", "ollama"}` because those servers drop it.

The filter is load-bearing for correctness on Ollama (without it, the
model may not call the intended tool). The ask is not "send named
`tool_choice` to Ollama". The ask is **keep the `tools` array byte-identical
across the loop** so the prompt prefix is stable.

## Measured shape (local Ollama, Qwen3.6 hybrid, 0.33.1)

Forced iterations of ~2.5k / 2.7k / 13.5k / 35.5k tokens, each a full
prefill; synthesis ~48k tokens with `lcp = 3`. Prefill ~490 tok/s. A
"low" reflect is 5–8 minutes, almost all of it repeated prefill. Cap had
to be sized to the wall (`HINDSIGHT_API_REFLECT_MAX_CONTEXT_TOKENS=56000`
under a 900 s wall) rather than to the model context.

## Suggested fix

1. **Forced iterations:** keep the full tools schema on the wire. Encode
   the forced function as a *suffix* (last user message: "You must call
   `search_mental_models` now") or any other mechanism that does not
   mutate `tools[]`. Consecutive turns then *extend* the previous prompt
   and hybrid checkpoints restore.
2. **Final synthesis:** do not swap the system prompt and do not drop
   tools. Append "produce the final answer, no further tool calls" as a
   new message so synthesis is also a prefix extension. Dropping tools
   diverges at token 3 on tools-first templates.

Happy to re-measure against a snapshot if useful. I can provide llama-server
log lines (`forcing full prompt re-processing…`, `lcp = 32` / `lcp = 3`)
from a local Ollama 0.33.1 / Qwen3.6-35B hybrid seat; no bank content.

leilei3167:
## Problem

On Ollama / LM Studio (providers that silently drop `tool_choice="required"`), Reflect's forced iterations narrow `tools[]` to the named function, and the forced final synthesis swaps the system prompt and drops tools entirely. Templates that render the tools block before system text (Qwen-style) then diverge early in the prompt. Hybrid / Gated-DeltaNet models cannot resume from a partial prefix, so each iteration full-prefills the accumulated history.

## Root cause

1. `OpenAICompatibleLLM.call_with_tools` filters `tools` down to the forced function for named `tool_choice`, then omits `required` for ollama/lmstudio — correct for forcing a call, but the tools schema changes every forced turn.
2. `_forced_final_synthesis` starts a fresh tool-less conversation with `build_final_system_prompt`, so synthesis is not a prefix extension of the agent loop.

## Fix

- **Forced iterations (ollama/lmstudio):** keep the full `tools` array byte-identical. Encode the forced function as a user-message suffix (`You must call \`name\` now`) instead of filtering `tools[]`. Providers that honor `required` still use the previous narrow+required path.
- **In-budget final synthesis:** keep the same system prompt and `tools` array; append a "produce the final answer, no further tool calls" nudge and call with `tool_choice=none`. Over-budget split synthesis is unchanged (conversation no longer fits as a prefix).

## How verified

- `pytest tests/test_tool_choice_required_downgrade.py tests/test_lmstudio_tool_choice.py tests/test_reflect_agent.py tests/test_reflect_split_synthesis.py -m 'not hs_llm_core'` → 83 passed (env-only pg0/real-LLM setup errors unrelated)

Fixes #3865


leilei3167:
@nicoloboschi quick check-in — forced Reflect turns on ollama/lmstudio keep the full tools schema (force via a user suffix) and in-budget final synthesis stays a prefix extension instead of swapping system/dropping tools. Happy to tweak the nudge wording if you want it stronger/weaker.

koriyoshi2041:
The current free-threaded 3.14 failure is deterministic rather than the earlier flake, so another empty retrigger will not clear it. Both failures come from `tests/test_opencode_go_session_header.py:110`: the test reads `result.ok`, while `LLMCallResult` exposes the parsed payload through `result.content`; `test_named_result_stubs.py` flags the same line for that reason.

This file is outside this PR's diff and the same assertion is present on current `main` (`eb32332e`), so it does not look owned by the stable-tools change. Once main fixes that test, a rebase should separate this PR's signal from the baseline failure.

leilei3167:
Thanks for the clear diagnosis — agreed this is the baseline `result.ok` vs `result.content` assertion on main, not from the stable-tools change. I'll rebase once that test lands on main so CI can speak for this PR again.

PR #4072, opened 2026-09-03
fix(cache): preserve force refresh across event loops

koriyoshi2041:
## Summary

- detach every loop-scoped in-flight bank-stats load when a force refresh completes
- prevent an older loader on another event loop from overwriting the force-refreshed value
- add a deterministic two-thread / two-loop regression

This follows the multi-loop cache changes merged in #4037 and the finding recorded in [this review](https://github.com/vectorize-io/hindsight/pull/4037#pullrequestreview-5096781949).

## Reproduction

Before the fix:

1. loop A starts a normal load and pauses;
2. loop B completes a forced refresh with a newer value;
3. loop A completes;
4. the next cache hit returns loop A's older value.

The force-refresh path removed only its own `(loop, key)` flight, so the older flight on loop A remained authorized to populate the shared cache.

## Validation

- `test_bank_stats_cache.py` + `test_multi_loop_conformance.py`: 18 passed
- focused regression after formatting: 1 passed
- Ruff check and format check pass for both changed files
- `git diff --check` passes

The repository-wide lint wrapper reached ESLint without the control-plane Node dependencies installed in the isolated checkout (`@eslint/js` missing). No TypeScript files are changed; the Python Ruff gates pass directly.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/bank_stats_cache.py
+++ b/hindsight-api-slim/hindsight_api/engine/bank_stats_cache.py
@@ -108,8 +108,11 @@ class BankStatsCache:
             value = await loader()
             with self._lock:
                 self._store_unlocked(key, value)
-                # Supersede any loader that was in flight for this key.
-                self._in_flight.pop(self._flight_key(key), None)
+                # Supersede every loader that was in flight for this key. Flights
+                # are loop-scoped, so removing only this loop's slot lets an older
+                # loader on another loop overwrite the force-refreshed value later.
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Rebased onto the merged multi-event-loop support from #4067. The cache and multi-loop suites now pass together (23 tests), along with focused Ruff and format checks. The previous free-threaded failure was in the unchanged migration fan-out test; this new head should give it a clean rerun.

nicoloboschi (approved):

PR #4073, opened 2026-09-03
perf(db): bind pgvector values in binary instead of a text literal

nicoloboschi:
**Draft** — this is the enabling half. It makes binary binds possible and safe; it does not yet convert the call site where the win lives (see *What this does not do*).

## Measurement

Postgres 18 + pgvector, 2000 rows x 384d, **statement shape held identical** (COPY into a staging table, then `INSERT … SELECT … RETURNING`), so the codec is the only variable:

| | time |
|---|---|
| COPY text literal → `::vector` | 471 ms |
| COPY binary vector | **22 ms** (21.7x) |

An embedding as `'[0.1,...]'` is ~12 bytes/dimension plus a server-side float parse per dimension; binary is 4 bytes/dimension and no parse.

## What this does not do — both decided by measurement, not taste

- **Single-vector binds are left alone.** 21 of the 22 `$n::vector` sites bind one query vector. There binary saves **34 µs per query** (0.147 ms → 0.114 ms) — about 0.1% of a recall. Not worth the call-site churn.
- **The bulk insert is left alone.** An array of vectors **cannot** be bound in binary at all — asyncpg hands the whole list to the scalar encoder. Capturing the win on `insert_facts_batch` means restructuring it into COPY-to-staging + `INSERT … SELECT` (with an explicit `ORDER BY ord`, since the caller maps `RETURNING id` by position). That is a change to retain's central write path and belongs in its own PR, with its own measurement of what share of a real run it moves — in the pipelines I sampled, `db_write` is ~5% of a run, so this is a bulk-ingest win and close to nothing elsewhere.

## Why the encoder is tolerant

`set_type_codec` replaces the codec for the **whole connection** — it is all-or-nothing. Once registered, any caller still passing a rendered literal would fail with `DataError` at runtime. So the encoder accepts `str` as well as ndarray / list / `array('f')`. A site that was missed keeps working at exactly its old cost; conversion becomes an optimisation rather than a correctness cliff.

Verified against a real Postgres:
- ndarray, list and text literal all bind through one connection and store **identical bytes**
- the existing `unnest($n::vector[])` batch insert is **unaffected**
- binary `COPY` into a `vector` column works

## Safety

- **Reads are untouched.** Nothing selects a raw `vector` column — both read sites cast with `embedding::text` — so the decoder is not on any current path.
- **A database without the extension still works.** `register_vector_codec` returns `False` instead of raising, and those connections bind text literals. Migrations run before the pool is created, so this only affects a process started with `run_migrations=False` against an unmigrated database.
- **Oracle is unaffected** — it keeps the text rendering in `retain.types.embedding_to_pgvector`; there is no pgvector wire format there.
- `pgvector` is already a dependency and is pure Python, so no free-threading risk on the `-py3.14t` image.

8 unit tests pin the tolerance contract (every input form encodes to identical bytes, the wire layout, `None` passthrough, and that a 2-d array fails loudly rather than silently storing one row).

https://claude.ai/code/session_018HDqrzHgqZqsGc7EDqoTEu

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/postgresql.py
@@ -157,6 +157,19 @@ def application_name_from_dsn(dsn: str) -> str | None:
     return values[-1] or None
 
 
+def _vector_codec_init(next_callback: Any | None) -> Any:
+    """Wrap a pool ``init`` callback so the binary vector codec is installed first."""
+
+    async def init(conn: Any) -> None:
+        from .vector_codec import register_vector_codec
+
+        await register_vector_codec(conn)
+        if next_callback is not None:
+            await next_callback(conn)
+
+    return init
+
+
 def _application_name_setup(app_name: str, init_callback: Any | None) -> Callable[[Any], Awaitable[None]]:
     """Wrap ``init_callback`` so every acquire re-asserts ``application_name``.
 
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Closing: on its own this delivers no measurable win.

The codec only pays off once `insert_facts_batch` is restructured into COPY-to-staging + `INSERT … SELECT` (243 ms -> 62 ms at 2000x384). Without that, the only converted sites would be single-vector binds, worth 34 µs per query — about 0.1% of a recall.

A cheaper route was tested and rejected: passing all vectors in one natively-binary `float4[][]` parameter (no codec, no temp table, no pool-init change) works functionally but is *slower* than what ships — 298 ms vs 243 ms — because the per-row `array_agg` reconstruction costs more than the text parse it avoids.

So the ~4x needs both the codec and the write-path restructure. Parking until the text formatter from #4040 shows whether this is worth touching retain's central write path at all. Measurements are in the commit message and PR body above.

PR #4079, opened 2026-09-03
feat(llm): add opt-in session ID affinity

SharkyRawr:
## Summary
- Add HINDSIGHT_API_LLM_CACHE_AFFINITY=x_session_id for OpenAI-compatible Chat Completions.
- Share one trace-derived x-session-id across related calls while keeping separate operation chains isolated.
- Preserve operator overrides; auto and existing modes are unchanged.
- Set a versioned hindsight/openai-compatible User-Agent on both OpenAI-compatible clients; explicitly configured User-Agent headers stay authoritative.

## Why
Adding an opt-in session ID is a reasonable compatibility measure: providers and gateways cannot route related request chains to the same prompt cache without an affinity signal, which can materially reduce cache efficiency.

## Validation
77 affinity tests, Ruff, type checking, and git diff checks pass.


nicoloboschi (changes requested):
Thanks for this! Before going further on the substance: **this branch is two unrelated features and needs to be split into two PRs.**

```
c6ec4e2d0 feat(llm): add opt-in session ID affinity
e890d3515 fix(llm): identify OpenAI-compatible requests
```

The second one — `openai_compatible_headers.py` and the `hindsight/openai-compatible/{version}` User-Agent — has nothing to do with cache affinity, and it's a behaviour change for **every** existing OpenAI-compatible user: it overrides the OpenAI SDK's own `OpenAI/Python x.y.z` User-Agent on every single request. Some gateways and providers route or gate on that header, so that deserves its own PR with its own rationale and its own docs entry (right now it's documented nowhere — no `.env.example`, no `configuration.md`, unlike every other header behaviour we have). Please pull it out; I'd rather review and merge the two independently.

---

## On the affinity mode itself

Bigger design question, and it's the main thing blocking my sign-off on the first half.

`HINDSIGHT_API_LLM_CACHE_AFFINITY` is turning into a registry of vendor-specific headers under a deliberately generic name:

- `xai_conv_id` → x.ai's `x-grok-conv-id`
- `openai_prompt_cache_key` → OpenAI's body field
- `x_session_id` → ...an unnamed gateway (this PR)
- plus `x-opencode-session`, already hardcoded per-provider in the same module

That's four mechanisms and growing, one enum value per vendor. If we want to support a generic session header, the right shape is one generic mode plus a configurable header name — something like `cache_affinity=header` with `HINDSIGHT_API_LLM_CACHE_AFFINITY_HEADER=x-session-id`. That covers `x-session-id`, `x-opencode-session`, and the next gateway without another enum value and another PR. Under a generic `LLM_`-prefixed setting, a generic *mechanism* is defensible; one more vendor-specific value is what makes the setting's name read wrong.

Related: which gateway actually accepts `x-session-id`? The module's stated doctrine (see its docstring) is that we emit a hint only for backends **documented** to accept one — `xai_conv_id` cites docs.x.ai, `openai_prompt_cache_key` cites the OpenAI field. The new docstring says gateways "commonly accept" `x-session-id` with no citation. Could you point at the specific backend and its docs? That would also tell us whether the generic-header design above is the better fit.

## Must fix

**1. `.env.example` drift will fail CI.** `.env.example` was edited without re-copying it to `hindsight-embed/hindsight_embed/env.example` — the bundled copy has to stay byte-identical. Reproduced locally:

```
FAILED hindsight-embed/tests/test_env_template.py::test_bundled_template_matches_repo_root
```

Fix: `cp .env.example hindsight-embed/hindsight_embed/env.example`

**2. Don't hand-edit the frozen versioned docs.** `hindsight-docs/versioned_docs/version-0.9/developer/configuration.md` is a snapshot of what actually shipped in 0.9 — `git log` on that file shows it's only ever touched by `Release vX.Y.Z` commits. This PR documents an unreleased feature into it. Please revert that file; the release tooling picks the change up from `hindsight-docs/docs/` when a version is cut.

**3. Rebase.** The branch doesn't contain current `main` — the merge commit is from an older upstream.

## Should fix

**4. Undocumented behaviour change on the xAI path.** `apply_cache_affinity` now consults `default_headers` for *all* modes, so an operator who set `x-grok-conv-id` via `HINDSIGHT_API_LLM_DEFAULT_HEADERS` stops receiving our per-run id. That's arguably a fix, but it's outside the PR's stated scope and isn't called out anywhere — please either mention it explicitly or scope the new check to the new mode.

**5. `apply_opencode_session` wasn't updated.** It still does a bare `extra_headers.setdefault(...)` and ignores `default_headers`, so the new "operator config wins across both header layers" rule holds for two of the three mechanisms and not the third — in the same file, a few lines apart. Either use `_setdefault_header` there too, or add a comment saying why it's deliberate.

## What's good

Ruff, `ruff format` and `ty check` all pass, and `test_cache_affinity.py` + `test_openai_responses_provider.py` are green locally (101 passed). The test coverage on the new mode is genuinely thorough — case-insensitive clobber tests across both header layers, the tools path, and an end-to-end reflect lane. Once the split and the design question above are settled, the affinity half should be straightforward to land.

SharkyRawr:
Thanks for taking the time to review  my pr :) hopefully I have addressed all your requested changes now.

- removed the unrelated changes and focusing only on X-Session-Id now in this PR
- generic header mode that mirrors the existing affinity-mode configuration: 
  - `HINDSIGHT_API_LLM_CACHE_AFFINITY=header` 
  - `HINDSIGHT_API_LLM_CACHE_AFFINITY_HEADER =x-session-id` (or whichever is needed)
  - `HINDSIGHT_API_RETAIN_LLM_CACHE_AFFINITY` / `HINDSIGHT_API_RETAIN_LLM_CACHE_AFFINITY_HEADER`
  - `HINDSIGHT_API_REFLECT_LLM_CACHE_AFFINITY` / `HINDSIGHT_API_REFLECT_LLM_CACHE_AFFINITY_HEADER`
  - `HINDSIGHT_API_CONSOLIDATION_LLM_CACHE_AFFINITY` / `HINDSIGHT_API_CONSOLIDATION_LLM_CACHE_AFFINITY_HEADER`
- `cp .env.example hindsight-embed/hindsight_embed/env.example` ✅ 
- frozen versioned docs restored, that was an overzealous AI decision and my brain being too sleep deprived to notice, my bad! 
- Branch rebased on to main ✅ 
- PR test suite completes (so far) without errors, yay

To hopefully answer your question "Who even uses X-Session-Id", I understand its a non-standard convention however adding the header did improve my particular workflow and environment. I use Tailscale Aperture as a generic AI gateway and saw a large number of single (ungrouped) requests in the logs; with the header it neatly groups requests together. Same goes for OpenRouter, the Sessions tab in the log section (`https://openrouter.ai/logs?tab=sessions`) is empty without the header. I felt that it's better to have some kind of session affinity between requests as they are also grouped together in the `Bank Configuration → LLM Requests` tab.

This whole idea started btw because I got a semi-angry mail from OpenCode, that I'm not sending the now required x-opencode-session header:

> [...] Your client should:  
> * Send typical coding agent traffic
> * Identify itself with its own user agent, such as my-coding-agent/1.0, rather than a > generic SDK or HTTP-library name.
> * Send a stable session ID in x-opencode-session for each conversation so we can optimize routing and prompt caching. If we don't have this we cannot properly optimize our service. Starting 09/06 requests missing this header may error. [...]

... but as it turned out the [opencode sourcecode](https://github.com/anomalyco/opencode/blob/57ef3828431790c53f8f333c7ffbfe88770a1812/packages/opencode/src/session/llm/request.ts#L188) only adds that header when the provider name begins with "opencode" and, as I was using TS Aperture, my requests' model id looked like this: `aperture/opencode-go/deepseek-v4-flash`. This led me to a mad scramble to get my clients as compliant as possible within the timeframe; hindsight already has this feature when using the `opencode-go` provider but was missing a proper user-agent (*this was merged not long after I opened my PR*). Anyways, the generic `X-Session-Id` header just seemed like a natural addition, I've never hosted LLM models at scale so I can't tell if it is indeed necessary for proper caching - but at least in small-scale self-hosting using vllm it has its purpose:

> The X-Session-ID HTTP header is used throughout vLLM for session/conversation tracking and KV-cache reuse. This is particularly important for multi-turn conversations where maintaining session affinity ensures the model can reuse the key-value cache across sequential turns.

PR #4082, opened 2026-09-03
feat(retain): a vision slot, so text-only chunks keep a cheap model

nicoloboschi:
Stacked on #4077 — review that first; this branch's base is its head.

## Why

Inline attachments made a vision-capable model mandatory for a **whole bank**. One screenshot in one document and every text-only chunk — the overwhelming majority of them — was billed against a vision model too, and answered more slowly. There was no way to say "use the cheap text model, except when there's actually something to look at".

## What

`HINDSIGHT_API_VLM_PROVIDER` / `_API_KEY` / `_MODEL` / `_BASE_URL` name the model used for **attachment-bearing chunks only**. Every text-only chunk keeps using the retain LLM.

Each field falls back to the corresponding retain value, so leaving them unset resolves to exactly the retain config and nothing changes — the hot path is untouched for anyone not using this.

```bash
# cheap text model for the 99% case
HINDSIGHT_API_RETAIN_LLM_MODEL=gpt-4o-mini
# vision model only where there is a picture
HINDSIGHT_API_VLM_PROVIDER=gemini
HINDSIGHT_API_VLM_MODEL=gemini-2.5-flash
```

The swap happens at the one place that already knows a chunk carries attachments — where placeholders are resolved back to bytes — so no new branch had to be invented for it.

The `422` vision check now interrogates this slot rather than the retain LLM, since it is the model that will actually be handed the image. That is what lets the retain LLM be text-only without the retain being refused.

## Three details worth a reviewer's eye

**A vision call must not fail over to a text model.** Building the slot with `_build_llm` (the obvious way, copying the retain slot) made the vision model member 0 of the *retain* chain and appended the retain fallbacks behind it. Those fallbacks are text models — that is why a separate vision slot exists — so a failed vision call would fail over to one, extract from the prose and drop the picture: the same silent omission the 422 gate refuses up front, arriving by the back door. The slot is built without a chain, and a test asserts the assignment does not go through `_build_llm`, because the hazard has no observable signature until a vision provider is actually down.

**A base URL follows its own provider.** Naming `vlm_provider` without a URL resolves *that* provider's default rather than inheriting the retain provider's host — otherwise `vlm_provider=groq` under an OpenAI retain config sends Groq requests to OpenAI's endpoint with the wrong key, surfacing as an auth error from a provider the operator never configured.

**The slot is a fresh `LLMConfig`, not a copy of the retain one.** `dataclasses.replace` was the shortcut and it broke 126 tests: a retain config can be a `MultiLLMProvider`, which is not a dataclass. When the slot is unconfigured the retain config is shared as-is, so a multi-LLM retain chain stays intact.

## Config surface

`vlm_api_key` and `vlm_base_url` are registered as credential fields (never exposed via the API); provider and model are server-level, matching how `retain_llm_provider` / `retain_llm_model` are already treated rather than adding a new per-bank surface.

`.env.example` + the bundled embed copy + both docs pages updated; the template sync test passes.

## Testing

Routing is a deterministic decision, not a model judgement, so it is asserted directly rather than judged: a text-only chunk reaches the retain config and an attachment-bearing chunk reaches the vision config, both sides of the branch. Plus the fallback-chain guard above, the base-URL resolution, and the credential/configurable registrations.

165 tests pass across the vlm, attachment, multimodal, per-operation-LLM and bank-health suites.


nicoloboschi:
Folded into #4077 — shipping as one PR rather than a stack.

PR #4088, opened 2026-09-03
chore(deps): bump fast-uri to 3.1.7 and browserslist to 4.28.8

cdbartholomew:
Raises two npm overrides and regenerates the four lockfiles that carry them.

## What

- **fast-uri** `>=3.1.5` → `>=3.1.6` (resolves 3.1.7) — root workspace and `hindsight-integrations/coding-agents`
- **browserslist** — new override `>=4.28.7` (resolves 4.28.8) — root workspace, `hindsight-embed`'s control-center UI, and `hindsight-integrations/zapier`

Each resolves to a single hoisted copy per lockfile; no de-hoisted duplicates remain.

The browserslist bump carries its data dependencies with it — `caniuse-lite`, `electron-to-chromium`, `node-releases`, `update-browserslist-db` and `baseline-browser-mapping` all move. That is most of the diff.

Closes #649-adjacent alerts for these two packages: 8 for fast-uri, 4 for browserslist.

## Lockfiles were regenerated with npm 11

npm 10.9.2 mishandles overrides inside workspaces: it applies the override at the hoisted root while de-hoisting *new* copies of the old version into the workspaces, leaving the tree worse than before the bump. npm 11 resolves to one hoisted copy.

npm 11 also prunes two optional peer-dependency entries from the zapier lockfile — `encoding` and its nested `iconv-lite`. That is expected and harmless. Flagging it because a reviewer reading the diff on npm 10 will see removals they cannot otherwise account for.

`npm ci --dry-run` passes under npm 10.9.2 for all four lockfiles, so the npm 11 output is still consumable by an npm 10 runner.

## Verification

- `npm ci --dry-run` clean under npm 10.9.2 — root (2342 pkgs), control-center UI (168), coding-agents (336), zapier (962)
- control-center UI builds clean (`vite build`, 10 modules transformed)
- confirmed a single hoisted `fast-uri@3.1.7` and `browserslist@4.28.8` per lockfile, no duplicate nested copies

## Deliberately not included

Committed build artifacts under `control_center/static/` are untouched. They are stale independently of this change — a fresh build produces a different bundle than the one committed on `main` — and correcting that belongs in its own change rather than riding along with a dependency bump. If `verify-embed-control-center-bundle` fails here, that is the cause and it predates this branch.

## Note on provenance

This supersedes #4053, which was based on the `deps/dependabot-patchable-sweep` branch rather than `main`. CI is configured `on: pull_request: branches: [main]`, so that PR received zero checks and its dependency bump was never verified. It was closed automatically when its base merged. The changes are re-applied here on top of `main` so they actually run.


cdbartholomew:
## Merging over a known-failing job

`smoke-openclaw-install` is red here. Recording why that is not this change, so the reasoning can be checked without redoing the search.

**It fails identically on unrelated branches.** Same job, same error, on two open PRs that touch none of these files:

| Branch | Run | Error |
|---|---|---|
| `fix-paginate-scopes-webhooks` | [33772354257](https://github.com/vectorize-io/hindsight/actions/runs/33772354257) | `npm error Cannot read properties of null (reading 'edgesOut')` |
| `feat/retain-multimodal-images` | [33786941040](https://github.com/vectorize-io/hindsight/actions/runs/33786941040) | `npm error Cannot read properties of null (reading 'edgesOut')` |
| this PR | [33792865851](https://github.com/vectorize-io/hindsight/actions/runs/33792865851) | `npm error Cannot read properties of null (reading 'edgesOut')` |

All three end at `[smoke-test FAIL] openclaw plugins install failed`. Neither of the other two changes npm dependencies at all, so a fast-uri or browserslist bump cannot be the cause.

**It was re-run once and failed the same way**, so this is not a transient blip — it is a standing failure that happens to be red right now.

The error is an internal arborist crash inside `npm install`, not a resolution or version conflict. It is worth its own issue; it is not something this PR can fix or should block on.

**Everything else is green:** 30 checks ran, and the jobs covering the four changed lockfiles all passed — `build-hindsight-all-npm`, `build-typescript-client`, `test-typescript-client`, `build-control-plane`, plus the zapier and coding-agents integration jobs.

PR #4089, opened 2026-09-03
ci: run the core suite when the workspace lockfile or root manifest changes

cdbartholomew:
## The gap

`uv.lock` and `pyproject.toml` at the repository root determine which dependency versions the API is built and tested against. Neither appears in any path filter, so a pull request changing only one of them resolves `core` to false and every Python job is skipped.

The result is a green check set that proves nothing. A dependency bump touching only the workspace lockfile currently reports success having run three jobs — `detect-changes`, `verify-generated-files`, `build-docs` — with `test-api` and the rest of the suite skipped. Nothing observes whether the new versions actually work.

This is easy to miss on review, because "all checks passed" and "the change was tested" look identical in the UI.

## The tell

The workflow already treats the lockfile as environment-determining. It is part of the `test-api` venv cache key:

```yaml
key: ${{ runner.os }}-venv-test-api-${{ hashFiles('uv.lock', 'hindsight-api-slim/pyproject.toml', '.python-version') }}
```

So the cache is invalidated by a file that cannot trigger the job whose cache it keys. The workflow knows the lockfile matters; the filter just never said so.

## The change

Adds both root files to the `core` filter.

Both patterns are **bare**, matching only the files at the repository root. A glob such as `**/uv.lock` would match the ~52 integration lockfiles under `hindsight-integrations/` and run the full core suite for every sample-project bump. Verified with picomatch:

```
uv.lock                                  -> match
hindsight-integrations/crewai/uv.lock    -> no match
hindsight-integrations/llamaindex/uv.lock-> no match
hindsight-api-slim/uv.lock               -> no match
```

`**/uv.lock` matches all four, which is what we do not want.

## Cost

A root-lockfile-only change now runs the 30 jobs gated on `core`, where it previously ran three. That is the intended trade: these are exactly the changes no other filter covers, and they alter the runtime environment for everything.

Bumps confined to an integration's own lockfile are unaffected — they keep matching only their own integration filter.

## Verifying it works

This PR cannot demonstrate the fix on itself: it changes a workflow file, which is not in the `core` filter either. The check comes on the next pull request that touches the root lockfile — `test-api` should run rather than skip.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -100,6 +100,14 @@ jobs:
             - 'scripts/**'
             - '.python-version'
             - '.env.example'
+            # The workspace lockfile and root manifest determine which versions
+            # the API actually runs against, so a change to either can break it
+            # without touching a line of source. Both are deliberately bare
+            # patterns: they match only the files at the repository root. A
+            # glob such as '**/uv.lock' would pull every integration lockfile
+            # into the core suite.
+            - 'uv.lock'
+            - 'pyproject.toml'
           clients-ts:
             - 'hindsight-clients/typescript/**'
             - 'package.json'


cdbartholomew:
## CI notes before merge

Two red checks, handled differently. Recording both.

### `test-api (1/3)` — a real flake, passed on re-run

Failed on:

```
FAILED tests/test_custom_embedding_dimension.py::TestCohereEmbeddings::test_cohere_embeddings_encode
  - httpx.ConnectError: [Errno 104] Connection reset by peer
```

That test is gated on `COHERE_API_KEY` and, when the key is present, issues a **live request to Cohere's API**. The failure is a connection reset from the third party, not an assertion. Re-run once and all four shards now pass.

Worth flagging separately: this PR makes `test-api` run on root-lockfile changes, so a suite containing live third-party API calls now gates dependency bumps. Expect this class of flake to show up more often on routine dependency PRs. Mocking `TestCohereEmbeddings` / `TestCohereCrossEncoder`, or moving them behind an explicit opt-in marker, would be worth doing on its own.

### `smoke-openclaw-install` — pre-existing, unrelated

Fails identically on branches that cannot have caused it:

| Branch | Run | Error |
|---|---|---|
| `fix-paginate-scopes-webhooks` | [33772354257](https://github.com/vectorize-io/hindsight/actions/runs/33772354257) | `npm error Cannot read properties of null (reading 'edgesOut')` |
| `feat/retain-multimodal-images` | [33786941040](https://github.com/vectorize-io/hindsight/actions/runs/33786941040) | same |
| this PR | [33793483644](https://github.com/vectorize-io/hindsight/actions/runs/33793483644) | same |

All end at `[smoke-test FAIL] openclaw plugins install failed`. This PR changes one path filter in a workflow file and cannot affect an `npm install` inside the openclaw smoke test. It is an internal arborist crash and deserves its own issue.

### Everything else

109 checks ran, 0 pending. All four `test-api` shards green — which is also the first direct evidence that the jobs this PR enables actually pass.

PR #4096, opened 2026-09-03
chore(deps): bump agno to 3.0.5

cdbartholomew:
Closes the SQL-injection advisory [GHSA-82m5-3pcp-hccq](https://github.com/advisories/GHSA-82m5-3pcp-hccq) (high), which covers agno `<= 2.6.5`. The lockfile held 2.6.5 — the last vulnerable release.

`agno` is unpinned in `pyproject.toml`, so this is a lockfile-only change: `uv lock --upgrade-package agno`.

## The advisory looks unfixable, and isn't

GitHub records **no patched version** for this advisory, which reads as "nothing to do". That is stale. The advisory covers `<= 2.6.5`; agno has since shipped 3.0.5, which is outside the range. The fix arrived as a major version bump rather than a patch release, which is presumably why no patched version was recorded.

Worth knowing when triaging the rest of this repo's `patched: NONE` alerts — some of them may be the same shape, though most genuinely are not: nltk, chromadb and image-size all have a latest release that still sits *inside* its vulnerable range.

## 2.x → 3.x

The integration's runtime imports are `agno.run.base.RunContext` and `agno.tools.toolkit.Toolkit`. Both still resolve.

`agno.agent.Agent` and `agno.models.openai.OpenAIChat` appear only in the module docstring's usage example and inside a function body, so they are not imported at load time.

agno 3.x makes `openai` an optional extra, so `from agno.models.openai import OpenAIChat` now raises unless `openai` is installed. That affects the docstring example, not this package — nothing here imports it at runtime, and the integration does not declare `openai` as a dependency in either version.

Transitively: adds `prompt-toolkit`, `questionary`, `wcwidth`; drops `python-multipart`, `smmap`.

## Tests

**162 passed, 10 failed — identical to `origin/main` before the bump, test for test.**

The 10 failures are pre-existing and unrelated to agno. They assert on `Hindsight(...)` constructor calls without the `user_agent` kwarg the client now sends:

```
Expected: Hindsight(base_url='http://localhost:8888', timeout=30.0)
  Actual: Hindsight(base_url='http://localhost:8888', timeout=30.0, user_agent='hindsight-agno/0.4.19')
```

Verified by running the same suite in a clean worktree of `origin/main` and diffing the failure lists — no difference, so this change introduces none of them. They are worth fixing on their own.

### DIFF
--- a/hindsight-integrations/agno/uv.lock
+++ b/hindsight-integrations/agno/uv.lock
@@ -4,26 +4,39 @@ requires-python = ">=3.10"
 
 [[package]]
 name = "agno"
-version = "2.6.5"
+version = "3.0.5"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
+    { name = "agnoctl" },
     { name = "docstring-parser" },
-    { name = "gitpython" },
     { name = "h11" },
     { name = "httpx", extra = ["http2"] },
     { name = "packaging" },
     { name = "pydantic" },
     { name = "pydantic-settings" },
-    { name = "python-dotenv" },
-    { name = "python-multipart" },
     { name = "pyyaml" },
     { name = "rich" },
-    { name = "typer" },
     { name = "typing-extensions" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/ce/3e/e0c32020ba694024f5b438e969ef3a483ca143cb9f9338442626223c7f7c/agno-2.6.5.tar.gz", hash = "sha256:b3b06a46d3ca03e425754bbb7e3228f0720dc4b7c80710fae44c4b39eb34b1c3", size = 2001624, upload-time = "2026-05-06T13:49:10.323Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/24/74/3ab694e4d5106def289c61dc569aedeea4e822c9b0fed32452c6905ca645/agno-3.0.5.tar.gz", hash = "sha256:52425d2e731b09badfdf64a50cb6fd7b580111ec1788e4a072de475578ad77a5", size = 3402397, upload-time = "2026-09-01T09:06:59.936Z" }
 wheels = [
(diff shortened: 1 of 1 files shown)


cdbartholomew:
## CI does not cover this change

Three checks ran — `detect-changes`, `verify-generated-files`, `build-docs`. None of them exercises agno.

`hindsight-integrations/agno` has **no test job in the workflow at all**: there is no `agno` path filter and no `test-agno-integration` job, unlike its siblings (`test-crewai-integration`, `test-llamaindex-integration`, `test-pipecat-integration`, …). So this is not a skip that a path-filter fix would cure — there is nothing to run.

That gap is worth closing on its own, and is not something this PR should take on.

## What was actually verified, locally

Since CI cannot speak to this, here is the evidence in full:

**Import surface across the 2.x → 3.x boundary**

```
from agno.run.base import RunContext      OK
from agno.tools.toolkit import Toolkit    OK
import hindsight_agno                     OK
```

Those two are the only agno imports evaluated at load time. `agno.agent.Agent` and `agno.models.openai.OpenAIChat` appear in the module docstring's usage example and inside a function body.

agno 3.x makes `openai` an optional extra, so `from agno.models.openai import OpenAIChat` now raises unless `openai` is installed. That affects the docstring example only — this package never imports it at runtime and does not declare `openai` in either version.

**Test suite: 162 passed, 10 failed.**

Identical to `origin/main` before the bump, test for test. Verified by checking out a clean worktree of `origin/main`, running the same suite, and diffing the failure lists — no difference. The 10 are a pre-existing mismatch where the suite asserts on `Hindsight(...)` calls without the `user_agent` kwarg the client now sends.

**Resolution check**

`uv lock --upgrade-package agno` moved 2.6.5 → 3.0.5, outside the advisory's `<= 2.6.5` range. `uv sync --frozen` clean. Transitively adds `prompt-toolkit`, `questionary`, `wcwidth`; drops `python-multipart`, `smmap`.

PR #4104, opened 2026-09-04
fix(retain): keep per-fact tags when a delta retain only updates metadata

### LINKED ISSUE #4068: v0.9.2 reprocess no-op strips label-derived per-fact tags
## Bug Description

On Hindsight v0.9.2, an explicit document reprocess can take the unchanged-content metadata-only path and overwrite each surviving memory unit's tags with the document-level tags. Tags produced by an entity label configured with `tag: true` are removed even though the fact IDs and text are preserved.

This appears to be a data-loss consequence of the silent reprocess no-op reported in #3899. PR #3949 forces re-extraction on reprocess and is merged on `main`, but it is not included in the latest v0.9.2 release. #3899 states that the no-op preserves units and tags byte-identically; the behavior below preserves units but does not preserve their label-derived tags.

No private memory text, bank identifiers, document identifiers, operation identifiers, local paths, credentials, or provider configuration are included in this report. The names below are synthetic.

## Steps to Reproduce

1. Configure a bank with an entity label whose extracted value is also stored as a tag, for example:

   ```json
   {
     "entity_labels": [
       {
         "key": "category",
         "type": "multi-values",
         "tag": true,
         "optional": true,
         "values": [
           {
             "value": "durable",
             "description": "A durable technical decision"
           }
         ]
       }
     ]
   }
   ```

2. Retain a synthetic document with document-level tag `scope:test` and content that produces a fact labeled `category:durable`.
3. Verify the document has `tags=["scope:test"]` while at least one extracted memory unit has `tags=["scope:test", "category:durable"]`.
4. Call `POST /v1/default/banks/{bank_id}/documents/{document_id}/reprocess` without changing the document content.
5. Poll the returned operation to completion and inspect the child retain metadata and the same memory unit.

## Expected Behavior

An explicit reprocess should re-extract using the current bank configuration, as documented. If any path completes without re-extraction, it must preserve every surviving memory unit's existing per-fact tags, including tags derived from entity labels.

A reprocess should never silently replace per-fact label tags with only the document-level tag set.

## Actual Behavior

The reprocess operation completed successfully, and its child retain reported `unit_ids_count: 0`.

After completion:

- document text and content hash were unchanged;
- memory unit IDs and fact text were unchanged;
- entity membership was unchanged;
- label-derived `category:*` tags were removed from affected facts;
- only the document-level `scope:test` tag remained.

The public memory update API does not accept `tags`, so there is no supported exact rollback for already-affected per-fact tag arrays. Document tag PATCH is not equivalent because it propagates one common tag set to every memory unit.

## Related Work / Questions

- #3899 documents the silent unchanged-content reprocess no-op.
- #3949 is the merged `force_reextract` fix on `main`.
- #4019 updates observation invalidation when surviving units are rescoped.

Could maintainers confirm whether #3949 fully prevents this label-tag loss on current `main`? A regression test should cover a document whose units carry both document-level tags and `entity_labels[].tag=true` tags before reprocess.

For already-affected v0.9.2 banks, is there a supported recovery path that restores per-memory tags without deleting/reimporting the document or editing the database directly?

## Version

- Hindsight API: v0.9.2
- Deployment: official container image, self-hosted in a Podman VM on a macOS host
- Python: 3.11.16

## LLM Provider

Not believed to be provider-specific: the observed child retain completed with `unit_ids_count: 0`, so the issue is in the unchanged-content reprocess/metadata path rather than label extraction output.

ebarkhordar:
Superseded by #4120, which merged this morning. That one splits the label tags by config key (`label_tag_keys` and `split_label_tags` in `entity_labels.py`) instead of inferring provenance from the document's previous tag set, so it also covers the overlap case listed as the third limit below, plus the delta-retain path and the tags PATCH that this PR left alone. Closing this one.

On current main the reprocess half is already fixed: `reprocess_document` sets `force_reextract`, the delta attempt is gated on `not force_reextract`, and running the label config from the issue through a reprocess re-extracts and keeps `category:durable`. So #3949 does cover that question.

The loss is still reachable the other way. A plain re-retain of the same content still lands on the metadata-only delta path, and that path does `SET tags = <document tags>` on every surviving unit. `_inject_label_tags` puts `category:durable` on the unit and never on the document row, so the document's tag set can't reproduce it and it's gone.

What separates the two halves is the document's previous tag set: a tag a unit carries that the old document set didn't contain is fact-derived, so it's kept. It's read in the locked SELECT that already fetches `content_hash`, before the row gets rewritten. A store that can't report it gets the old behaviour.

Removal still works, a tag dropped from the document is still dropped from its units, which is what the #4019 invalidation keys on. There's a test for that direction too.

Three limits. The store-owned branch of that function has no test in the repo, so only the SQL path is pinned here. A tag the document supplied and extraction also derived reads as document-derived to a set difference, so dropping it from the document drops it from the units, where a full replace would re-derive it; that case comes out the same as it does on main, and what changes here is the case where the document never carried the tag. And this doesn't repair units that already lost their tags.

Fixes #4068

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/fact_storage.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/fact_storage.py
@@ -417,6 +417,31 @@ def _normalize_scopes(value: list | str | None) -> list | str | None:
     return value
 
 
+def _reconcile_unit_tags(
+    unit_tags: list[str] | None,
+    document_tags: list[str] | None,
+    previous_document_tags: list[str] | None,
+) -> list[str]:
+    """The tags a surviving unit should carry after a delta retain.
+
+    A unit's tags are the document's tags plus whatever extraction derived for that
+    fact alone (``_inject_label_tags`` appends a ``key:value`` tag per entity label
+    group configured with ``tag: true``). Only the first half is reproducible from
+    the document row, so replacing the list wholesale drops the second half.
+
+    ``previous_document_tags`` is the document's tag set before this retain, which is
+    what makes the two halves separable: anything the unit carries that the old
+    document tag set did not put there is fact-derived and is kept. ``None`` means the
+    caller could not read it, and the document's tags are applied as-is.
+    """
+    incoming = list(document_tags or [])
+    if previous_document_tags is None:
+        return incoming
+    previous = set(previous_document_tags)
+    already = set(incoming)
+    return incoming + [t for t in (unit_tags or []) if t not in previous and t not in already]
+
+
 async def update_memory_units_metadata_and_tags(
     conn,
     bank_id: str,
@@ -425,6 +450,7 @@ async def update_memory_units_metadata_and_tags(
     metadata: dict[str, Any],
     *,
     observation_scopes: list | str | None = None,
+    previous_document_tags: list[str] | None = None,
     ops=None,
 ) -> int:
     """Update document-level attributes on existing memory units.
@@ -432,6 +458,13 @@ async def update_memory_units_metadata_and_tags(
     Delta retain preserves unchanged chunks and their facts. Propagate the
     current document tags, metadata and observation scoping so its optimized
     result matches a full replace.
+
+    Matching a full replace is why ``previous_document_tags`` exists: a full
+    replace re-extracts, so it re-derives the per-fact label tags this path has
+    no way to recompute. Given the document's previous tag set, those tags are
+    the ones a survivor carries that the previous set did not contain, and they
+    are kept (see ``_reconcile_unit_tags``). Without it the units are relabelled
+    with the document's tags alone.
 
     ``metadata`` arrives as the raw retain_params bag (the document row keeps
     the caller's input verbatim), so null-valued keys are dropped here — the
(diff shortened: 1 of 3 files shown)


koriyoshi2041 (commented):
There is one provenance collision this reconciliation cannot distinguish: a tag can have been supplied by the document *and* independently derived from the fact.

For example, with a labelled unit and a document that initially also supplied the same tag:

```python
_reconcile_unit_tags(
    ["category:durable"],
    [],
    ["category:durable"],
)
```

this returns `[]`. But a full replace would run `_inject_label_tags` again and keep `category:durable`, so removing the document copy accidentally removes the still-valid fact-derived copy too.

Could we pin this overlap case in the unit test? I think the implementation needs provenance that does not infer “fact-derived” solely as `unit_tags - previous_document_tags` (for example, recomputing configured label tags from stored unit labels if that information is available). Otherwise the stated full-replace equivalence has this lossy edge case.

ebarkhordar:
You're right, and the example is exact: `_reconcile_unit_tags(["category:durable"], [], ["category:durable"])` returns `[]` on the branch. Pinned it in `test_reconcile_unit_tags` and put it in the docstring as the case a set difference cannot split (b484d59).

I stopped short of the provenance fix. Recomputing the label tags needs `parse_entity_labels(config.entity_labels)` and each unit's entities, and `update_memory_units_metadata_and_tags` has neither: it selects `id, fact_type, tags, observation_scopes` off `memory_units` and never sees the bank config. Threading that down is a different shape of change from a delta-path patch, so I'd rather a maintainer picks between doing it and leaving the overlap to a full replace.

koriyoshi2041 (changes requested):
Thanks for pinning the case. The new test confirms the branch still removes a valid label-derived tag, though, so documenting that loss does not make the delta path equivalent to a full replace. I reran the focused helper test at `b484d59` (1 passed), and the overlap assertion explicitly locks in the incorrect `[]` result.

I think this needs either the provenance/config plumbing you outlined or a safe fallback to full replace when provenance is ambiguous. Otherwise this fixes the common case while preserving silent tag loss for a valid configuration.

PR #4107, opened 2026-09-04
fix: retry transient Gemini embedding requests

### LINKED ISSUE #4103: GeminiEmbeddings: add bounded retry/backoff for 429 responses
## Problem

`GeminiEmbeddings` in `hindsight_api/engine/embeddings.py` constructs `google.genai.Client` without `HttpOptions.retry_options`. Each `client.models.embed_content(...)` therefore lets a Gemini `429 RESOURCE_EXHAUSTED` fail directly through to the worker.

The worker retries the entire operation using its fixed task retry settings, but the default budget (`HINDSIGHT_API_WORKER_MAX_RETRIES=3`, `HINDSIGHT_API_WORKER_TASK_RETRY_BACKOFF_SECONDS=60`) is too short for a sustained quota window. This can dead-letter retain, batch-retain, and consolidation operations even though 429 is transient.

We observed this across five self-hosted Hindsight instances sharing a Tier-1 Gemini embedding project: 213 failed operations during a backfill before retries were manually queued.

## Suggested change

Pass bounded Google GenAI retry options when creating the Gemini embeddings client, covering at least 429 and transient 5xx responses with exponential backoff and jitter. Preserve the existing custom HTTP client path used by `force_ipv4`; merge retry options into that `HttpOptions` rather than replacing its timeout/client settings.

Useful acceptance checks:

- synchronous `embed_content` retries 429 and transient 5xx responses;
- retry count and maximum delay are bounded;
- `force_ipv4` retains its existing transport and timeout behavior;
- permanent 4xx responses do not retry;
- exhausted retries propagate a diagnostic exception to the worker.

Related prior work: #3090 added bounded embedding retry/backoff for LiteLLM providers, but the native Gemini provider still lacks equivalent handling.

koriyoshi2041:
## Problem

Native Gemini embedding requests currently fail immediately on quota and transient service responses. The worker can retry the whole task, but that coarse retry can dead-letter retain and consolidation operations during a short provider outage.

Closes #4103.

## Fix

Configure the Google GenAI client with a bounded request-level retry policy for 429 and transient 5xx responses. Apply the same policy to Gemini API and Vertex AI clients, and merge it into the existing `force_ipv4` HTTP options so the custom transport and timeout are preserved. Permanent 4xx responses remain fail-fast.

## Test

- `uv run --directory hindsight-api-slim pytest tests/test_gemini_embeddings.py -q` (36 passed, 1 skipped)
- Ruff check on the changed source and tests
- Ruff format check
- `git diff --check`

## Risk

Retries add bounded latency when Google is unavailable. The policy is capped at five total attempts and a 10-second maximum delay; exhausted requests still propagate to the worker, and the status-code allowlist excludes permanent client errors.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/embeddings.py
+++ b/hindsight-api-slim/hindsight_api/engine/embeddings.py
@@ -1943,6 +1943,14 @@ class LiteLLMSDKEmbeddings(Embeddings):
 # without a "google/" or "models/" prefix.
 _GEMINI_AGGREGATING_MODEL_MARKER = "gemini-embedding-2"
 
+# Keep provider-level retries short enough that the worker retains control of
+# the operation budget while absorbing ordinary quota and service blips. The
+# SDK's retry layer is request-scoped, unlike the worker's whole-task retry.
+_GEMINI_EMBEDDING_RETRY_ATTEMPTS = 5
+_GEMINI_EMBEDDING_RETRY_INITIAL_DELAY = 1.0
+_GEMINI_EMBEDDING_RETRY_MAX_DELAY = 10.0
+_GEMINI_EMBEDDING_RETRY_STATUS_CODES = [429, 500, 502, 503, 504]
+
 
 def _gemini_model_aggregates_inputs(model: str) -> bool:
     """Whether the model aggregates a multi-input request into one embedding."""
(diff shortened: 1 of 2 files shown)


nicoloboschi:
Thanks for digging into #4103 — the diagnosis is right, but I'd rather not add a fourth retry flavour to `embeddings.py`.

We already have a shared, config-driven retry layer in that same file: `EmbeddingRetryPolicy` + `_call_with_retry`/`_acall_with_retry` (with `_is_transient_embedding_error` for classification), fed by `HINDSIGHT_API_EMBEDDINGS_MAX_RETRIES` / `_INITIAL_BACKOFF` / `_MAX_BACKOFF` / `_RETRY_BUDGET`. It's wired into the LiteLLM providers today. Configuring the Google SDK's own `HttpRetryOptions` instead means:

- hardcoded constants that ignore the `HINDSIGHT_API_EMBEDDINGS_*` knobs;
- no wall-clock `budget_seconds` cap — that pairing exists precisely because recall embeds the query inline on the request path, so attempt count alone cannot bound added latency;
- a status allowlist that diverges from `_TRANSIENT_STATUS_CODES` and doesn't cover transport-level failures;
- the same options object built twice (Gemini + Vertex paths).

I'm fixing #4103 by routing `GeminiEmbeddings` through the shared policy instead, so it gets the same bounded, tunable behaviour as the other providers. Closing this one — appreciate the report and the repro detail in the issue.

PR #4109, opened 2026-09-04
fix(tracing): attach bank ID to retain spans

### LINKED ISSUE #4106: hindsight.bank_id is set on recall spans but missing from all retain spans (0.9.2)
### What happened

`hindsight.bank_id` is set on every span in the **recall** path but on none in the **retain** path. That makes it impossible to attribute a retain to a bank in an OTLP backend when one API instance serves several banks.

Verified on **0.9.2** (`ghcr.io/vectorize-io/hindsight@sha256:84ab276b…`, revision `ebad4782`), exporting to an OTLP endpoint, one container serving two banks.

Recall spans — attribute present:

```
hindsight.recall            hindsight.bank_id = dp   (+ query, fact_types, max_tokens, thinking_budget)
hindsight.recall_embedding  hindsight.bank_id = dp   (+ query)
hindsight.recall_retrieval  hindsight.bank_id = dp   (+ fact_types, thinking_budget)
hindsight.recall_fusion     hindsight.bank_id = dp   (+ semantic/bm25/graph/temporal/merged counts)
hindsight.recall_rerank     hindsight.bank_id = dp   (+ candidates_count, scored_count)
```

Retain spans — attribute absent:

```
hindsight.retain                  (root)   no hindsight.bank_id
hindsight.retain_extract_facts             no hindsight.bank_id
                                           has hindsight.scope, hindsight.provider.internal,
                                           and the gen_ai.* set
```

### Reproduce

1. Enable tracing: `HINDSIGHT_API_OTEL_TRACES_ENABLED=true` plus an OTLP endpoint.
2. `POST /v1/default/banks/bank-a/memories` and `POST /v1/default/banks/bank-b/memories`.
3. Inspect the resulting `hindsight.retain` and `hindsight.retain_extract_facts` spans — neither carries `hindsight.bank_id`, so the two retains are indistinguishable by bank.
4. `POST /v1/default/banks/bank-a/memories/recall` for contrast — every recall span carries it.

Also worth noting: `dry-run-extract` emits `hindsight.retain_extract_facts` but no `hindsight.retain` parent, so a dry-run and a real retain look structurally different in a trace view. Not necessarily wrong, but surprising.

### Why it matters

`HINDSIGHT_API_OTEL_DEPLOYMENT_ENVIRONMENT` is process-wide, so a single instance serving a production bank and a scratch/eval bank cannot separate the two in telemetry by environment either. With `bank_id` missing from the retain path there is no supported way to answer "how is extraction behaving for bank X" — which is the question retain telemetry mostly exists to answer. Cost and latency per bank have the same problem.

The workarounds are all worse than the fix: a second container per bank would misattribute, because workers claim tasks fleet-wide via tenant discovery rather than per bank; and reconstructing the mapping outside Hindsight means maintaining a parallel record of something the server already knows.

### Suggested fix

Set `hindsight.bank_id` on `hindsight.retain` and its children, matching the recall path. `hindsight.operation` would be a useful addition on both, since it is documented as an operation-span attribute but was not present on the spans observed here.

Happy to test a build.

koriyoshi2041:
Fixes #4106.

`start_as_current_span()` returns a context manager, but `create_operation_span()` was attempting to set attributes on that manager rather than on the span it yields. As a result, the documented `hindsight.operation` and `hindsight.bank_id` attributes never reached retain, reflect, or consolidation operation spans.

This enters the OpenTelemetry context first, applies attributes to the yielded span, and preserves the existing operation context-manager API. The regression tests now keep the context manager and yielded span as distinct mocks so this failure mode is covered.

Tests:
- `uv run pytest tests/test_tracing.py -q` (28 passed)
- `./scripts/hooks/lint.sh`

### DIFF
--- a/hindsight-api-slim/hindsight_api/tracing.py
+++ b/hindsight-api-slim/hindsight_api/tracing.py
@@ -14,6 +14,7 @@ import json
 import logging
 import os
 import time
+from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Optional
 
 from opentelemetry import trace
@@ -301,6 +302,7 @@ def get_tracer() -> trace.Tracer | NoOpTracer:
     return _tracer
 
 
+@contextmanager
 def create_operation_span(operation: str, bank_id: str | None = None):
     """
     Create a parent span for a Hindsight operation (retain, reflect, consolidation, etc.).
@@ -317,21 +319,17 @@ def create_operation_span(operation: str, bank_id: str | None = None):
         Span context manager
     """
     if not _tracing_enabled or _tracer is None:
-        # Return a no-op context manager
-        from contextlib import nullcontext
-
-        return nullcontext()
+        yield None
+        return
 
     span_name = f"hindsight.{operation}"
-    span = _tracer.start_as_current_span(span_name)
-
-    # Add operation-specific attributes
-    if span and hasattr(span, "set_attribute"):
+    # OpenTelemetry returns a context manager here, not the span itself. Set
+    # attributes on the span yielded by that manager so exporters receive them.
+    with _tracer.start_as_current_span(span_name) as span:
         span.set_attribute("hindsight.operation", operation)
         if bank_id:
             span.set_attribute("hindsight.bank_id", bank_id)
-
-    return span
+        yield span
 
 
 def is_tracing_enabled() -> bool:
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):
LGTM

Issue #3040, opened 2026-07-29
retain: zero-fact documents produce no signal at write time (completed status, no webhook field, no metric) — operator cannot know a document needs reprocessing


dimonnld:
> **EDIT 2026-07-29 — two corrections after @Sanderhoff-alt's review. Original text kept below unchanged.**
>
> 1. **`memory_unit_count` IS exposed** on `GET /documents` and `GET /documents/{id}`. My statement below that "the only way to discover it is a manual SQL join" is **wrong** — after-the-fact auditing is possible today.
> 2. **"Persist chunks at 0 facts" (request 1 below) is withdrawn.** Chunks carry no embeddings and `recall`/`reflect` reach them via `memory_units.chunk_id`, so an orphan chunk is unreachable whether stored or not. The request would not have achieved what I asked for.
> 3. **`POST /documents/{id}/reprocess` recovers the document** — verified on our instance: a zero-fact window went `0 units / 0 chunks` → `3 units / 2 chunks` after the retain mission was changed. So there is a working recovery path.
>
> **The ask is therefore narrowed to visibility at write time**, which is request 2 below: a zero-fact count in the retain result and the `retain.completed` webhook, a metric, and a line in the docs. The remaining argument is that the outcome is **non-deterministic** (same document, same mission, five runs → `2, 2, 0, 0, 2`) and produces no signal either way, so an operator has no way to know a document needs reprocessing. Full reasoning in the comments.

---

## Summary

When fact extraction runs successfully but legitimately returns **0 facts** for a document, `retain` tracks the document row and returns early — **no memory units and no chunks are persisted**. Because semantic search runs over memories rather than chunks, that window becomes permanently unreachable via `recall`, `recall` with `include: {chunks}`, and `reflect`. `documents.original_text` still holds the bytes, but nothing can find them.

There is no error, no warning, no status change and no webhook signal. The only way to discover it is a manual SQL join between `documents` and `chunks`.

Version: **0.8.5** (self-hosted, current latest).

## Mechanism

In the retain orchestrator, when a consumer batch yields no facts:

```python
log_buffer.append(
    f"[streaming] Consumer batch {consumer_batch_idx + 1}: "
    f"0 facts extracted from {len(batch)} chunks, skipping"
)
return
```

The document is tracked (`Document … tracked (0 facts in first batch)`) but the code returns before `store_chunks_batch`. The chunks that were computed are discarded.

[...]

PR #4112, opened 2026-09-04
fix(retain): report how many memories a synchronous retain created (#4065)

### LINKED ISSUE #4065: Concurrent synchronous retains to DISTINCT document_ids with update_mode=append return HTTP 200 but silently drop a nondeterministic subset of writes (v0.9.2)
**Environment**

- Image: `ghcr.io/vectorize-io/hindsight`, digest `sha256:84ab276b8f501546deb6ea9c64a57291718b4e16a59dd9e02a02fdd5adfe9028`
- Reported API version: 0.9.2
- Deployment: single container on loopback, default configuration

**Expected behaviour**

8 concurrent synchronous (`"async": false`) retain requests, each writing one
item with `update_mode: "append"` to its OWN distinct `document_id`, should
all produce memory units. Every request returns HTTP 200, so every write
should be durable and retrievable.

**Actual behaviour**

All 8 requests return HTTP 200, but a nondeterministic subset of the writes
never creates any memory unit. For each lost write:

- `GET /v1/default/banks/{bank}/documents/{document_id}` returns HTTP 200
  with `memory_unit_count = 0` — the document exists, its content was never
  processed.
- Recall for the unique marker returns 0 hits after 30 s of polling.
- `GET .../memories/list?q={marker}` returns 0 hits after 30 s of polling.

Observed results across two independent runs of the identical script against
the same pinned image:

| run | all HTTP 200 | survived | lost | lost writer indices |
|-----|--------------|----------|------|---------------------|
| 1   | yes (8/8)    | 3/8      | 5    | 0, 1, 2, 4, 7       |
| 2   | yes (8/8)    | 5/8      | 3    | 3, 4, 5             |

The lost set differs between runs, so the loss is nondeterministic. No error
is returned at any point; the failure is silent.

**Minimal reproduction**

Requires only Python 3 standard library and a running Hindsight instance.
Uses synthetic markers; no real data. Set `HINDSIGHT_API` and
`HINDSIGHT_API_KEY` in the environment.

```python
import json, os, time, urllib.request
from concurrent.futures import ThreadPoolExecutor

API = os.environ["HINDSIGHT_API"]          # e.g. http://127.0.0.1:8888
KEY = os.environ["HINDSIGHT_API_KEY"]
BASE = f"{API}/v1/default"
BANK = f"repro-distinct-append-{int(time.time())}"
N = 8
RUN = str(int(time.time()))
HEADERS = {"Authorization": f"Bearer {KEY}", "Content-Type": "application/json"}

def req(method, path, body=None):
    data = json.dumps(body).encode() if body is not None else None
    r = urllib.request.Request(BASE + path, data=data, method=method, headers=HEADERS)
    with urllib.request.urlopen(r, timeout=180) as resp:
        return resp.status, json.loads(resp.read() or b"null")

req("PUT", f"/banks/{BANK}", {"name": "repro", "enable_observations": False})

def write(i):
    return req("POST", f"/banks/{BANK}/memories", {
        "items": [{
            "content": f"Independent concurrent marker CONCURRENT_INDEP_{RUN}_{i}.",
            "document_id": f"repro-doc-{RUN}-{i}",   # DISTINCT per writer
            "update_mode": "append",
        }],
        "async": False,
    })[0]

with ThreadPoolExecutor(max_workers=N) as ex:
    statuses = list(ex.map(write, range(N)))
print("statuses:", statuses)  # all 200 in every failing run

for i in range(N):
    marker = f"CONCURRENT_INDEP_{RUN}_{i}"
    st, doc = req("GET", f"/banks/{BANK}/documents/repro-doc-{RUN}-{i}")
    hits = 0
    deadline = time.time() + 30
    while time.time() < deadline and not hits:
        _, r = req("POST", f"/banks/{BANK}/memories/recall", {"query": marker})
        hits = len([x for x in (r.get("results") or []) if marker in (x.get("text") or "")])
        if not hits:
            time.sleep(1)
    print(f"writer {i}: doc_get={st} unit_count={doc.get('memory_unit_count')} recall_hits={hits}")

req("DELETE", f"/banks/{BANK}")
```

**Distinction from PR #3395**

This is not the defect fixed in
https://github.com/vectorize-io/hindsight/pull/3395
("stop concurrent appends to one document from losing turns"). That fix
serializes concurrent appends to ONE shared `document_id`. Here every writer
targets a DIFFERENT `document_id` — there is no shared document state to
race over — and writes are still lost under explicit `update_mode=append`.

For completeness: we also observed silent loss with two writers on ONE
document when `update_mode` is omitted, but the server documents the default
as `replace` ("deletes old data and reprocesses from scratch"), and under
explicit `append` that same-document shape preserved both writes. That
observation is therefore consistent with the documented default and is NOT
part of this report. The defect reported here is specific to concurrent
synchronous retains across distinct documents on the append path.

We do not claim a root cause; we report the observed, reproducible behaviour
on the pinned version above. Happy to capture server logs or run a
diagnostic build if that helps.

nicoloboschi:
Fixes #4065.

## What #4065 actually is

The concurrency is a red herring, and so is the write path. I reproduced the reported pattern by calling `extract_facts_from_text` directly on the issue's own marker strings — **no database, no document write, no HTTP layer**, and one call at a time:

```
concise mode, 8 markers, sequential:  [0,0,0,0,1,1,1,0]  -> 5 produced no facts
```

Then I re-ran each marker five times:

```
marker 0: [0,0,0,0,0]     marker 2: [1,1,1,1,1]
marker 1: [0,0,0,0,0]     marker 3: [1,1,1,1,1]
```

So it is not even nondeterministic per call — it is deterministic per text. `Independent concurrent marker CONCURRENT_INDEP_<ts>_<i>.` is a bare tracking marker, and the default `concise` extraction prompt is explicitly selective: *"Ask: would this be useful to recall in 6 months? If no, skip it."* For some of these strings the model answers no, for others yes, and it answers the same way every time for the same string. Nothing is racing and nothing is being dropped by the storage engine — consistent with @ebarkhordar's two negative reproductions.

The zero-fact writers still burn tokens (~1991 vs ~2148 for one that extracted), which is why the non-zero `usage` looked like proof the write had been lost. It isn't: extraction ran and legitimately returned `"facts": []`.

Downstream that is by design and already covered by `test_document_persisted_with_zero_facts`: the document is stored, has zero memory units, and recall cannot reach it.

## Why this was reported as data loss

Nothing said so, on either side of the wire.

- The **synchronous response** carries `success`, `items_count`, `async` and `usage` and no count of what was stored, so a caller cannot tell "your content became memory" from "there was nothing in it to remember" except by recalling later and finding nothing.
- The **server** counted the outcome in a metric (`_record_retain_document_outcome`, #3040) and logged nothing, so the logs of a retain that stored nothing look like an ordinary successful retain.

Over a batch, that is indistinguishable from writes being silently dropped.

## Changes

1. **`memories_created` on the synchronous retain response** — how many memory units this retain created. Zero is the signal that was missing.
2. **A warning log for every document left with zero memory units**, naming the cause and the fix.
3. **Docs** (`docs/developer/api/retain.mdx`): retain is not lossless by default, and if every item must be retrievable, `retain_extraction_mode` has two modes that cannot judge content away — `verbatim` (stores each chunk's text, model used only for metadata) and `chunks` (no LLM at all). Verified: `verbatim` keeps all eight of the issue's markers where `concise` keeps three.

Deliberately **not** done: falling back to storing raw content when extraction returns nothing. That was decided against previously — it introduces a hybrid storage mode and breaks the separation between extracted facts and raw text. Retrying was also ruled out by the measurement above: the answer is stable per text, so a retry recovers none of them.

## Tests

`hindsight-api-slim/tests/test_retain_memories_created.py` — the count matches the document's `memory_unit_count` on a normal retain; it is `0` with a 200 when extraction returns nothing; and the zero-unit document is logged.

OpenAPI, the generated Python/TS/Go clients, and the docs skill are regenerated. Both wrapper SDKs return the generated `RetainResponse`, so they pick the field up with no change.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -921,6 +921,7 @@ class RetainResponse(BaseModel):
                 "bank_id": "user123",
                 "items_count": 2,
                 "async": False,
+                "memories_created": 5,
                 "usage": {"input_tokens": 500, "output_tokens": 100, "total_tokens": 600},
             }
         },
@@ -939,6 +940,17 @@ class RetainResponse(BaseModel):
     operation_ids: list[str] | None = Field(
         default=None,
         description="Operation IDs when items were submitted as multiple strategy groups (async=true with mixed per-item strategies). operation_id is set to the first entry for backward compatibility.",
+    )
+    memories_created: int | None = Field(
+        default=None,
+        description=(
+            "How many memory units this retain created (only present for synchronous operations). "
+            "Extraction can legitimately find nothing to remember in content that carries no facts, "
+            "and that outcome is a success, not an error: the document is stored but has no memory "
+            "units, so it cannot be reached through recall. Zero here is what tells the caller that "
+            "happened. A re-retain of a document whose text is unchanged also reports zero, because "
+            "the delta path keeps the units it already has."
+        ),
     )
     usage: TokenUsage | None = Field(
         default=None,
@@ -8124,6 +8136,12 @@ def _register_routes(app: FastAPI):
 
                 # Synchronous processing: one batch per strategy group, aggregate results
                 total_items_count = 0
+                # Reported back to the caller. A retain whose extraction finds no
+                # facts stores the document and returns 200 with nothing behind it;
+                # without this count the caller cannot tell that from a retain that
+                # produced memories, and only discovers it later when recall comes
+                # back empty (#4065).
+                total_memories_created = 0
                 total_usage = TokenUsage(input_tokens=0, output_tokens=0, total_tokens=0)
                 with metrics.record_operation("retain", bank_id=bank_id, source="api"):
                     for group_strategy, contents in strategy_groups.items():
@@ -8141,6 +8159,8 @@ def _register_routes(app: FastAPI):
                             ),
                         )
                         total_items_count += len(contents)
+                        # `result` is the per-content-item list of created unit ids.
+                        total_memories_created += sum(len(unit_ids) for unit_ids in result)
                         if usage:
                             total_usage = TokenUsage(
                                 input_tokens=total_usage.input_tokens + usage.input_tokens,
@@ -8154,6 +8174,7 @@ def _register_routes(app: FastAPI):
                         "bank_id": bank_id,
                         "items_count": total_items_count,
                         "async": False,
+                        "memories_created": total_memories_created,
                         "usage": total_usage,
                     }
                 )
--- a/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/orchestrator.py
@@ -340,12 +340,23 @@ async def _record_retain_document_outcome(pool: Any, bank_id: str, document_id: 
     ``units_created > 0`` already settles the outcome, so the count query only
     runs on the zero case — which is exactly the cheap path (nothing was written).
     Best-effort: telemetry must never fail a retain.
+
+    The zero case is also logged, not only counted. It was previously visible in
+    the metric alone, so an operator reading the logs of a retain that stored
+    nothing saw a completely ordinary, successful retain (#4065).
     """
     try:
         total = units_created
         if total == 0:
             async with acquire_with_retry(pool) as conn:
                 total = await fact_storage.count_document_memory_units(conn, bank_id, document_id)
+        if total == 0:
+            logger.warning(
+                f"[RETAIN] Document {document_id} (bank {bank_id}) has no memory units after this retain: "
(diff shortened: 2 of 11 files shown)


nicoloboschi:
Closing: this was built on a wrong diagnosis.

#4065 is a real defect, but the cause is the bank display name leaking into the extraction prompt as `Narrator: <name>` — reproduced on v0.9.2 and absent on `main`, with the A/B in https://github.com/vectorize-io/hindsight/issues/4065#issuecomment-5538774676. It is already fixed by #3978 (`aabbe8b22`), which landed after the v0.9.2 tag.

The evidence in this PR's description came from calling `extract_facts_from_text` directly, which skips the retain mission preamble and so builds a different prompt than the server — the conclusion it supported does not hold for the real pipeline.

Issue #4063, opened 2026-09-03
perf(consolidation): consolidation_llm_parallelism gives no speedup when memories share a tag set — consolidation serializes to one LLM call per batch


handnewb:
## Summary

With `consolidation_llm_parallelism=4` configured, consolidation still runs fully serially — one LLM call in flight at a time — because the parallelism only fans out across *distinct tag-set groups*, never within a group. Banks whose memories share a tag set (the common case for uniform retain) collapse into a single group and serialize.

## Root cause

In `hindsight_api/engine/consolidation/consolidator.py`:

```python
if llm_parallelism > 1 and len(numbered_groups) > 1:
    sem = asyncio.Semaphore(llm_parallelism)
    ...
else:
    # serial fallback
```

`numbered_groups` is built by grouping memories on `_consolidation_batch_key(m)`, which for the default `combined` scope mode returns `("combined", *sorted(memory["tags"]))`. So **memories with an identical tag set form a single group**, and inside a group batches are processed serially ("they MUST run serially"). Additionally, groups whose resolved write scopes overlap serialize on `scope_locks`.

Result: for a bank where most memories share the same tag set (e.g. retained without tags, or with a single uniform tag), `len(numbered_groups) == 1` and the parallel branch is never taken. Consolidation throughput is then bounded by a single LLM call's latency per batch.

## Evidence (production, self-hosted)

Config: `consolidation_llm_parallelism=4`, `consolidation_llm_batch_size=8`, `consolidation_max_memories_per_round=100`.

Worker log shows one LLM call per batch and a single consolidation slot in use:

```
[CONSOLIDATION] bank=hermes llm_batch #4 (8 memories, 1 llm calls) | processed=32/4566 | recall=4.3s, llm=76.6s, ...
worker=vmi3397706 slots=1/10 | reserved: [consolidation=1/2(avail=1)] | pool: idle=12
```

Measured throughput: ~4–5 memories/min, draining a ~4500-memory backlog over ~15h.

## Expected vs actual

- Expected: `consolidation_llm_parallelism=4` runs up to 4 LLM calls concurrently for a single bank's consolidation.
- Actual: 1 LLM call at a time whenever memories share a tag set, regardless of the configured parallelism.

## Impact

A large unconsolidated backlog drains at single-LLM-latency speed, and the queue is non-deterministic if ingestion continues. The bottleneck is the serial dispatch, not LLM latency.

## Environment

[...]

PR #4133, opened 2026-09-04
fix(consolidation): warn when consolidation_llm_parallelism is configured but not exercised

handnewb:
## Problem

`consolidation_llm_parallelism > 1` promises parallel consolidation, but when every unconsolidated memory in a fetch shares the same tag set (the common case for uniform retain), they collapse into a **single tag group** and the parallel dispatch path is never taken. The configured parallelism silently does nothing — the operator sees slow serial consolidation and no signal that their setting is inert.

Tracked in #4063.

## Fix

Log a warning in the serial-fallback branch when `llm_parallelism > 1` but `len(numbered_groups) <= 1`:

```
[CONSOLIDATION] bank=hermes consolidation_llm_parallelism=4 but only 1 tag group(s) in this fetch — memories share a tag set so batches run serially (one LLM call at a time); parallelism is not exercised
```

This gives the operator visibility that the parallelism setting is not being exercised, without changing the consolidation behavior or its scope-safety guarantees.

## Scope (explicit)

This PR does **not** parallelize within a single tag group — that is a scope-safety design decision discussed in #4063 and belongs with the maintainers. This PR only closes the "config silently lies" gap by surfacing the inert setting.

## Testing

- `ruff check` on the touched file: passes.
- `ruff format --check`: passes.
- No behavioral change to the serial/parallel dispatch path (warning-only).

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -1866,6 +1866,15 @@ async def _run_consolidation_job(
             batch_results: list[_BatchDeltas] = [d for gd in group_results for d in gd]
             any_cancelled = any(d.cancelled for d in batch_results)
         else:
+            if llm_parallelism > 1 and len(numbered_groups) <= 1:
+                logger.warning(
+                    "[CONSOLIDATION] bank=%s consolidation_llm_parallelism=%d but only %d "
+                    "tag group(s) in this fetch — memories share a tag set so batches run "
+                    "serially (one LLM call at a time); parallelism is not exercised",
+                    bank_id,
+                    llm_parallelism,
+                    len(numbered_groups),
+                )
             batch_results = []
             any_cancelled = False
             for g in numbered_groups:


handnewb:
Note: the `test-api (free-threaded 3.14)` failure is pre-existing, not from this PR. The two failing tests are in files this PR does not touch and reflect the `LLMCallResult` → `.content` API refactor on main:

- `test_opencode_go_session_header.py::test_session_id_is_stable_across_provider_retries` → `AttributeError: 'LLMCallResult' object has no attribute 'ok'`
- `test_named_result_stubs.py::test_a_call_result_binding_is_only_read_through_its_fields` → `read .content`

This PR only adds a warning in `consolidator.py`.

nicoloboschi:
it's already explained in the docs

>Maximum number of tag groups consolidated concurrently within one consolidation op. Each group acquires per-scope locks before processing, so groups whose write scopes overlap (e.g. under per_tag / all_combinations / explicit-list observation_scopes) automatically serialise on the overlapping scopes — actual concurrency may be lower than this cap when scopes contend. Set to 1 for fully sequential behaviour. Higher values raise peak LLM QPS and connection-pool usage during consolidation proportionally — tune down if your LLM provider rate-limits tightly or your DB pool is small. Configurable per bank.



https://hindsight.vectorize.io/developer/configuration

PR #4135, opened 2026-09-04
fix(operations): let cancel abort a processing operation, not just a pending one

### LINKED ISSUE #4131: DELETE /operations only cancels pending ops – processing ops cannot be aborted via API
## Summary

`DELETE /v1/default/banks/{bank}/operations/{id}` succeeds only for operations in `pending` status. Operations stuck in `processing` (e.g. orphaned after a worker crash or restart) cannot be cancelled through the API.

## Steps to reproduce

1. Start an operation and let it reach `processing`.
2. Kill/restart the worker so the operation is orphaned in `processing`.
3. Attempt `DELETE /v1/default/banks/{bank}/operations/{id}`.
4. The API refuses or has no effect because the operation is not `pending`.

## Expected behavior

A way to cancel or force-fail `processing` operations – e.g. DELETE works on any non-completed status, or operations older than X without progress are automatically marked `failed`/retryable.

## Actual behavior

Only direct database manipulation (`UPDATE async_operations SET status='failed'`) plus a container restart clears the stuck operation.

## Environment

- Self-hosted hindsight (Docker, latest `ghcr.io/vectorize-io/hindsight`)

oldnicke:
## Summary

Fixes #4131. `DELETE /v1/default/banks/{bank}/operations/{id}` accepted only `pending`, so an operation left in `processing` — orphaned by a worker crash or restart — could not be cleared through the API at all. The only way out was `UPDATE async_operations SET status='failed'` by hand plus a container restart.

`processing` is now cancellable, and cancelling it releases the claim.

## Why the claim has to be released with it

The worker's own recovery/release passes filter on `worker_id = self._worker_id` (`_reset_own_processing_rows`, `recover_own_tasks`, `release_own_tasks`), so a row whose worker never comes back is unreachable by design — that is what made these rows permanent. `cancel_operation` therefore clears `worker_id`/`claimed_at` along with the status, which also stops a *live* worker from later resurrecting the row as `pending` through its shutdown release path.

Nothing else can undo the cancel either:

- the task itself notices at its next `_check_op_alive()` checkpoint (that helper already treats `status = 'cancelled'` as "stop") and returns early;
- `_mark_completed` is already guarded on `status = 'processing'`, so it no-ops;
- `_mark_failed` was **not** guarded — an aborted task usually surfaces as an error there, which would have relabelled the row `failed` and made it look retryable. It now skips rows that are already `cancelled`. The parent roll-up still runs; it reads whatever status the child committed.

## Changes

- `engine/memory_engine.py::cancel_operation` — allow `pending` and `processing`, as one conditional `UPDATE ... WHERE status IN ('pending','processing')` instead of SELECT-then-UPDATE (the old pair let a worker claim a pending row between the two statements and get cancelled out from under its own claim without releasing it). Clears `worker_id`/`claimed_at`, sets `completed_at = COALESCE(completed_at, now())` — the same shape the retention path already uses when it cancels a parent. Terminal statuses still answer 409, now with wording that names both cancellable states.
- `worker/poller.py::_mark_failed` — `AND status <> 'cancelled'`.
- `api/http.py`, `engine/interface.py` — summary/description/docstrings say what the endpoint now does. The MCP tool already advertised "Cancel a pending or running async operation", so that surface was promising this behaviour before the engine supported it.
- `hindsight-control-plane` — the Cancel button was gated on `status === "pending"` in both the operations table and the detail dialog, so the one place an operator actually watches a stuck operation could not act on it. Now shown for `processing` too.
- `hindsight-cli/src/main.rs`, `docs/developer/api/operations.mdx`, `docs/sdks/cli.md` — the API doc page explicitly claimed "Returns 409 if the operation is already in `processing`", which is exactly the reported behaviour.
- Regenerated `openapi.json` + Rust/Python/TypeScript/Go clients (description-only diff).

### Removed test

`tests/test_operation_status.py::test_cancel_rejects_non_pending_operations` asserted 409 for `processing` — it encoded the bug. It is replaced by:

- `test_cancel_processing_operation_releases_the_orphaned_claim` — cancels a row stamped with a dead `worker_id`, asserts status `cancelled` and that `worker_id`/`claimed_at` are cleared and `completed_at` set;
- `test_cancel_rejects_terminal_operations[completed|failed|cancelled]` — terminal rows still 409 and keep their status.

Plus `test_worker.py::test_mark_failed_leaves_a_cancelled_row_cancelled` for the poller guard.

## Not done (deliberately)

The issue's second suggestion — auto-failing operations older than X without progress — is a separate policy change needing its own knob and stall definition, and the worker already has stage-progress stuck *detection* (`STUCK_STACK_INITIAL_THRESHOLD_S`) that only logs. This PR makes the manual escape hatch work; an automatic reaper can build on it.

## Testing

- `uv run pytest tests/test_operation_status.py tests/test_worker.py` — pass, including the three new cases.
- `tests/test_op_cancellation.py` — 8 of 9 pass; `TestRetainCheckpoint::test_retain_stops_between_sub_batches_when_cancelled` fails on my machine with `column e.entity_kind does not exist` and a "Database is at a newer migration revision than this code version knows about" warning. Verified pre-existing: it fails identically with my changes stashed (local test DB migrated by a newer checkout).
- `./scripts/hooks/lint.sh` clean; `cargo fmt --check` + `cargo build` clean in `hindsight-cli`.
- `npx tsc --noEmit` in `hindsight-control-plane`: the two errors it reports (`temporal_window` on `RecallRequest`, missing `@testing-library/react`) are pre-existing — identical output with my changes stashed.
- Generated artifacts are in sync: `uv run generate-openapi` + `./scripts/generate-clients.sh` re-run produce no further diff, which is what the CI freshness gate checks.
- Not reproduced end to end against a real crashed worker: the orphaned state is simulated by stamping a dead `worker_id` on a `processing` row, which is the state such a crash leaves behind.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -6790,8 +6790,13 @@ def _register_routes(app: FastAPI):
     @app.delete(
         "/v1/default/banks/{bank_id}/operations/{operation_id}",
         response_model=CancelOperationResponse,
-        summary="Cancel a pending async operation",
-        description="Cancel a pending async operation by removing it from the queue",
+        summary="Cancel a pending or processing async operation",
+        description=(
+            "Cancel a pending or processing async operation. A pending operation is removed from "
+            "the queue; a processing one has its worker claim released and stops at its next "
+            "checkpoint, which is also how a row orphaned by a worker crash is cleared. Returns "
+            "409 for operations that already reached a terminal status."
+        ),
         operation_id="cancel_operation",
         tags=["Operations"],
     )
@@ -6799,7 +6804,7 @@ def _register_routes(app: FastAPI):
     async def api_cancel_operation(
         bank_id: str, operation_id: str, request_context: RequestContext = Depends(get_request_context)
     ):
-        """Cancel a pending async operation."""
+        """Cancel a pending or processing async operation."""
         try:
             # Validate UUID format
             try:
--- a/hindsight-api-slim/hindsight_api/engine/interface.py
+++ b/hindsight-api-slim/hindsight_api/engine/interface.py
@@ -614,7 +614,10 @@ class MemoryEngineInterface(ABC):
         request_context: "RequestContext",
     ) -> dict[str, Any]:
         """
-        Cancel a pending async operation.
+        Cancel a pending or processing async operation.
+
+        A processing operation is cancellable so that a row orphaned by a worker crash can be
+        released through the API; the claim is dropped and the task stops at its next checkpoint.
 
         Args:
             bank_id: The memory bank ID.
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -18633,29 +18633,53 @@ class MemoryEngine(MemoryEngineInterface):
         op_uuid = uuid.UUID(operation_id)
 
         async with acquire_with_retry(backend) as conn:
-            # Check if operation exists, belongs to this bank, and is in a cancellable state
-            result = await conn.fetchrow(
-                f"SELECT bank_id, status FROM {fq_table('async_operations')} WHERE operation_id = $1 AND bank_id = $2",
+            # One conditional write, like retry_operation: a SELECT-then-UPDATE pair would let a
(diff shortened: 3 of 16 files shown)


nicoloboschi:
fixed by #4177

PR #4145, opened 2026-09-05
feat(openclaw): add recall score floors

### LINKED ISSUE #4143: hindsight-openclaw: expose minScores on auto-recall — plugin has no precision gate (client SDK and API both support it)
### Summary

The OpenClaw plugin's auto-recall has no way to set a score floor. `recallMinScores`, `minScores`, and `min_scores` are all absent from `@vectorize-io/hindsight-openclaw` 0.10.0 — verified by grepping the installed build. The underlying `@vectorize-io/hindsight-client` SDK *does* support `minScores`/`min_scores`, and the API accepts `min_scores` on `POST /v1/default/banks/{bank_id}/memories/recall`; the plugin simply never passes it through.

The Claude Code plugin (`hindsight-memory` 0.7.4) exposes it as `recallMinScores`. So the two first-party clients differ on a setting that materially changes what lands in the model's context.

This is the same shape as #2977 (`prefer_observations` missing from OpenClaw auto-recall), which was resolved by adding a `preferObservations` config key.

### Why it matters

Without a floor, an off-topic turn injects a full budget of irrelevant memories. Measured against a single-user bank (188,946 facts), `budget: mid`, `max_tokens: 2048`:

| query | results, no floor | results, `reranker: 0.3` | reranker score range |
|---|---|---|---|
| sharp, on-topic | 31 | 31 | 0.67–0.85 |
| vague ("what should I do next") | 38 | 38 | 0.48–0.64 |
| off-topic ("sourdough starter hydration") | 35 | **2** | 0.23–0.31 |

The floor is a no-op on anything in-domain and a hard stop on anything outside it — exactly what you want from a precision gate on an automatic injector. Since the plugin injects on every turn, the off-topic row is ~2k tokens of confidently-scored noise with no way to suppress it.

### Request

Add a `recallMinScores` config key that passes through to the client's `minScores`, matching the Claude Code plugin's naming. Same semantics: per-stage floors (`semantic`, `keyword`, `reranker`, `final`); unset fields impose no floor; null/missing scores pass, so BM25-only and passthrough-reranker hits aren't accidentally suppressed.

### Note on calibration

Reranker scores are query-local and not calibrated across queries, so this is worth documenting as a garbage gate rather than a relevance dial — raising the floor to improve precision will start cutting good in-domain results before it cuts bad ones. The Claude Code plugin's README already says this; worth carrying over.

### Reproduction

```bash
# no min-scores support anywhere in the OpenClaw plugin
grep -rE "min_?[Ss]cores" node_modules/@vectorize-io/hindsight-openclaw   # no matches
grep -rE "min_?[Ss]cores" node_modules/@vectorize-io/hindsight-client     # minScores, min_scores
```

### Environment

- `@vectorize-io/hindsight-openclaw` 0.10.0
- Self-hosted Hindsight API, single-user bank
- Compared against `hindsight-memory` (Claude Code) 0.7.4, which has the key

koriyoshi2041:
Closes #4143

## Problem

OpenClaw auto-recall cannot pass the API/client `minScores` precision gate, so unrelated prompts can inject the full recall budget even though the first-party client already supports stage-specific floors.

## Fix

- add `recallMinScores` to the OpenClaw config schema and UI hints
- pass the configured floors through the scoped client to `HindsightClient.recall`
- document the null/missing-score behavior and the query-local calibration caveat
- cover config and client pass-through with regressions

## Test

- `npm test -- --run src/index.test.ts` (327 passed)
- `npm run build`
- `git diff --check`

The repository-wide lint script reached ESLint but could not load `@eslint/js` because the fresh worktree did not have the control-plane workspace dependencies installed. The changed OpenClaw files pass their package build and test suite.

## Risk

Low. The option is omitted by default, preserving current recall behavior. Validation is constrained to the four API-supported score stages, and missing/null result scores continue to pass through the server-side filter semantics.


koriyoshi2041:
The OpenClaw smoke failure is outside this diff: both heads reach the tarball install, then OpenClaw 2026.9.1 fails inside its dependency install with `npm error Cannot read properties of null (reading 'edgesOut')` ([first run](https://github.com/vectorize-io/hindsight/actions/runs/33939810626/job/101233219912), [current run](https://github.com/vectorize-io/hindsight/actions/runs/33947679747/job/101256745598)). The plugin build and generated-file checks pass. I cannot rerun the failed job from a fork (`Must have admin rights`), so this needs a maintainer rerun or the smoke harness to pin/work around the current installer.

nicoloboschi (approved):

PR #4148, opened 2026-09-05
fix(coding-agents): read DSH transcript via snapshotEvents

### LINKED ISSUE #4147: [dsh adapter] Session transcript is always empty: `agent.session.events` no longer exists in DeepSeek Harness alpha.4+ (chatDocs stays 0)
### Bug Description

With the DSH adapter (hindsight-coding-agents **0.5.1**, mounted in a DeepSeek Harness profile via its `cordis.patch.yml`), conversation transcripts are **never retained**: `hindsight_sync_status` reports `chatDocs: 0` forever, and the deepen engine logs `[chat] no sessions; skipping` on every pass.

The live per-turn write-back path (`agent/turn-stopping` → `onSessionIdle` → transcript fetch) silently no-ops every turn.

**Root cause:** `dist/dsh.js` — the transcript source inside `workspaceFor()` reads `agent.session.events` directly:

```js
core.setTranscriptSource(async (sessionId) => {
  const agent = liveAgents.get(sessionId);
  return agent ? readDshEvents(agent.session.events) : [];
});
```

DeepSeek Harness **0.1.2-alpha.4 removed the `Session.events` property**; the public accessor is now `session.snapshotEvents()`. So `agent.session.events` is `undefined`, `readDshEvents(undefined)` returns `[]`, and both `onSessionIdle` and `onTranscript` skip on `!turns.length` → every turn is silently dropped. This is the same alpha.4 breakage class other DSH plugins hit (they had to switch to `(session.log ?? session.events ?? [])`).

**Suggested fix (one line, verified locally on 0.5.1 + DSH 0.1.2-rc.x):**

```js
core.setTranscriptSource(async (sessionId) => {
  const agent = liveAgents.get(sessionId);
  // alpha.4+ removed Session.events; snapshotEvents() is the public accessor.
  const events = agent ? (agent.session.snapshotEvents?.() ?? agent.session.events ?? []) : [];
  return agent ? readDshEvents(events) : [];
});
```

**Secondary observation (same adapter):** the automatic first-prompt reflect is hard-capped to 25 s even on persistent-plugin hosts — `timeoutMs: Math.min(cfg.reflectTimeoutMs, HOOK_REFLECT_CAP_MS)` in `buildHookOutput`. The cap fits hook harnesses, but the DSH `agent/pre-step` hook is awaited by the host with no such window, so `reflectTimeoutMs: 120000` cannot take effect (`Math.min` wins) and a `"low"`-budget reflect on a populated bank routinely exceeds 25 s. Since a failed reflect is cached (`reflectAnswer = ""`), that session permanently runs without memory. Consider a host-aware cap or an uncapped path for persistent plugins.

### Steps to Reproduce

1. DeepSeek Harness 0.1.2-alpha.4+ / 0.1.2-rc.x on Windows 11; install the plugin into a profile: `dsh plugin --profile <name> add @vectorize-io/hindsight-coding-agents` (mounts `.../dsh` via `cordis.patch.yml`).
2. `~/.hindsight/coding-agent.json`: `{ "serverMode": "daemon" }` with the local daemon running on 127.0.0.1:9077.
3. Open a session in that profile and chat for several turns.
4. Run `hindsight_sync_status`: `chatDocs` stays `0` while `gitlogPresent: true`.
5. Daemon log shows `[chat] no sessions; skipping`; the per-turn write-back never produces turns.

### Expected Behavior

The DSH adapter per-turn write-back (`onSessionIdle` → `onTranscript`) should upsert the live session transcript, so `chatDocs` grows and conversations compound into memory (parity with opencode's per-turn retain).

### Actual Behavior

Every turn is silently dropped because the transcript source is empty (`agent.session.events` is `undefined` on DeepSeek Harness alpha.4+ — see root cause above). Nothing in the bank records the conversations; only the git seed and manually ingested documents exist.

### Version

0.5.1

### LLM Provider

None

leilei3167:
## Summary
DeepSeek Harness **alpha.4+** removed `Session.events`. The DSH adapter's `setTranscriptSource` still read `agent.session.events`, so `readDshEvents(undefined)` returned `[]` and every turn write-back (`onSessionIdle` → `onTranscript`) silently no-oped — `chatDocs` stayed `0` and the daemon logged `[chat] no sessions; skipping`.

This switches the live transcript source to `session.snapshotEvents?.() ?? session.events ?? []` (via a small `dshSessionEvents` helper) so alpha.4+ hosts retain conversation turns while older hosts keep working.

Fixes #4147

## Test plan
- [x] `./node_modules/.bin/vitest run src/dsh.test.ts src/core/transcript-dsh.test.ts` → 19 passed (includes alpha.4 session mock without `events`, with `snapshotEvents()` returning events → non-empty transcript)
- [ ] Manual: DSH 0.1.2-alpha.4+ / rc with the plugin mounted; chat a few turns; `hindsight_sync_status` should show `chatDocs` growing

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/transcript-dsh.ts
+++ b/hindsight-integrations/coding-agents/src/core/transcript-dsh.ts
@@ -2,9 +2,10 @@
  * DeepSeek Harness (dsh) transcript normalizer.
  *
  * dsh keeps a session as an append-only log of typed `SessionEvent`s rather than a chat array, and
- * a plugin reads that log straight off `agent.session.events` — so this is a pure function over
- * those events, like the opencode normalizer, not a file reader. The SAME event vocabulary is what
- * dsh persists to disk, so the backfill reader (core/history.ts) feeds this exact function.
+ * a plugin reads that log straight off the live session (`snapshotEvents()`) — so this is a pure
+ * function over those events, like the opencode normalizer, not a file reader. The SAME event
+ * vocabulary is what dsh persists to disk, so the backfill reader (core/history.ts) feeds this
+ * exact function.
  *
  * Only three of the ~30 event types carry conversation:
  *   - `user/message`      surface, a UserMessage — the human prompt OR an injected plugin context
--- a/hindsight-integrations/coding-agents/src/dsh.test.ts
+++ b/hindsight-integrations/coding-agents/src/dsh.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from "vitest";
-import { createDshHooks, toDshParameters, type Workspace } from "./dsh";
+import { createDshHooks, dshSessionEvents, toDshParameters, type Workspace } from "./dsh";
+import { readDshEvents } from "./core/transcript-dsh";
 import type { ToolSpec } from "./core/knowledge-tools";
 import { z } from "zod";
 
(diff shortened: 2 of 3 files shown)


leilei3167:
@nicoloboschi ready for review — DSH alpha.4+ dropped `Session.events`; this uses `snapshotEvents?.() ?? events ?? []` so the per-turn transcript write-back is non-empty again (Fixes #4147).

PR #4149, opened 2026-09-05
fix(openclaw): resolve Dashboard agent identity from four-part session keys

### LINKED ISSUE #4144: OpenClaw: Dashboard sessions are skipped because agentId is not resolved for stable identity fallback
## Summary

`@vectorize-io/hindsight-openclaw` 0.11.1 skips OpenClaw Control UI Dashboard chat sessions with `missing stable sender identity`, even when configured for a single static bank.

## Environment

- Hindsight OpenClaw integration: 0.11.1
- OpenClaw: 2026.9.1
- Session key: `agent:main:dashboard:<uuid>`
- Static banking: `dynamicBankId: false`, named `bankId`
- `autoRetain: true`

## Actual behavior

Gateway logs report:

```text
hindsight: Skipping dispatch on session 'agent:main:dashboard:<uuid>': missing stable sender identity
```

The Dashboard is authenticated to the gateway but its conversation turns carry no human `senderId`. A direct check of the integration shows `resolveSessionIdentity({ sessionKey: 'agent:main:dashboard:<uuid>' })` returns only the session key; it does not recover `agentId: main`.

`getIdentitySkipReason()` has an intentional fallback for agent/static banking:

```ts
if (!resolvedCtx?.senderId || resolvedCtx.senderId === "anonymous") {
  if (allowCliSessions && resolvedCtx?.agentId) {
    resolvedCtx.senderId = `agent-user:${resolvedCtx.agentId}`;
  }
}
```

But the Dashboard key does not resolve `agentId`, so the fallback cannot run.

## Expected behavior

For Dashboard sessions, resolve the agent ID from `agent:<agentId>:dashboard:<uuid>` (or otherwise preserve the runtime agent ID) so the existing fallback can assign `agent-user:<agentId>` **only when** `allowCliSessions` is true—i.e. static banking is explicitly configured or agent-level banking is enabled.

This preserves the current safety property for multi-user/user-scoped banks while allowing a single-owner authenticated Dashboard to retain memories to its explicitly configured shared bank.

## Reproduction

1. Configure the plugin with `dynamicBankId: false`, a named `bankId`, and `autoRetain: true`.
2. Start a chat in the OpenClaw Control UI Dashboard.
3. Observe the skip log above; the turn is not retained.
4. Send an equivalent message over a channel with a sender ID (for example Telegram); it is retained normally.

## Notes

This seems to be an integration parsing/identity-resolution gap rather than a configuration problem. No credentials, bank IDs, or transcript content are needed to reproduce it.

r266-tech:
Fixes #4144.

## What changed

- recognize OpenClaw Control UI keys shaped as `agent:<agentId>:dashboard:<opaqueId>`
- recover only `agentId` from that namespace, leaving the live dispatch channel as the message provider
- preserve the sender identity synthesized by the existing agent/static-bank safety gate when returning and caching resolved context

This lets explicitly static or agent-scoped setups use `agent-user:<agentId>` for Dashboard sessions without weakening the existing rejection path when neither routing mode is enabled.

## Validation

- `npm test` (328 tests)
- `npm run build`
- `npx vitest run --config vitest.integration.config.ts tests/hooks.integration.test.ts` (26 tests)
- Prettier check for both changed files


koriyoshi2041 (approved):
Checked the identity boundary at `abb4e2bd`: Dashboard stays a session namespace, the live dispatch channel remains the provider, and the synthesized `agent-user:<id>` is only retained when agent/static banking makes it safe. The negative channel+user-only case also preserves the existing retryable skip.

`npx vitest run src/index.test.ts` passes all 193 tests locally. The OpenClaw install smoke failure is the same npm installer `edgesOut` failure already reproduced on #4145, outside this diff.

PR #4154, opened 2026-09-05
fix(coding-agents): exclude Codex Desktop startup instructions

Korayem:
## Summary

Codex Desktop's injected startup instructions are excluded from retained user conversation. The shared Codex reader uses content-block origin metadata when available, so genuine user text stays intact even when it contains exactly the same startup markup.

Assistant commentary, final answers, replies without a phase, compact tool actions, existing tool-output exclusions, and REF-ID metadata keep their current behavior. The change adds no settings, dependencies, or storage changes.

## Problem and evidence

Codex can serialize both injected context and genuine prompts as `response_item` messages with `role: "user"`. The role alone does not distinguish them. The reader already intends to exclude startup instructions, but its original checks recognize text beginning with `# AGENTS.md instructions for ` or `<environment_context>`.

The observed Desktop startup message begins with `<recommended_plugins>` and can use a pathless `# AGENTS.md instructions` heading. That envelope bypasses the original prefix checks and becomes a retained user row.

Review identified a limitation in the first fix: requiring all three startup blocks also requires AGENTS instructions to be present. We reproduced the partial-envelope gap with synthetic inputs. Broadening the text matcher would also classify a genuine prompt containing identical markup as injected context.

Inspection of local CLI and Desktop rollouts exposed a more precise distinction: `payload.internal_chat_message_metadata_passthrough.content_item_kinds` labels the origin of each item in `payload.content`. Observed startup kinds are `plugins.recommendations`, `agents_md.instructions`, and `environments.environment_context`; genuine text uses `user.text`. The regression inputs below reproduce that structure with invented content.

CLI startup instructions were already excluded correctly. Its retained commentary is acceptable existing behavior and remains intact.

## Behavior

For user messages, the reader uses origin metadata only when it is a complete array of nonempty strings aligned with the original content array. It excludes exactly the three known startup kinds, then joins the remaining text. Index matching happens before images or other non-text content are dropped, so mixed messages retain the right text.

| Input | Result |
| --- | --- |
| Known startup origins, including a partial Desktop envelope | Excluded per block |
| `user.text` containing an exact startup envelope or CLI prefix | Preserved |
| Mixed injected startup and genuine user text | Only startup blocks excluded |
| Unknown origin kinds or other kinds of injected context | Preserved under existing text handling |
| Missing, malformed, or misaligned origin metadata | Existing text detector used |
| Assistant messages of any phase, tool handling, and REF-ID | Unchanged |

The text fallback retains the existing CLI prefix checks and the conservative, sequential matcher for the complete Desktop envelope. It requires complete blocks with no trailing user prose, avoiding regex backtracking across a genuine request. Without usable origin metadata, text alone still cannot prove whether identical markup was typed by the user; this change does not broaden that heuristic to every partial envelope.

## Synthetic example

An injected Desktop message without an AGENTS block can be represented as:

```json
{
  "type": "response_item",
  "payload": {
    "type": "message",
    "role": "user",
    "content": [
      {"type": "input_text", "text": "<recommended_plugins>Example tools.</recommended_plugins>"},
      {"type": "input_text", "text": "<environment_context><cwd>/example</cwd></environment_context>"}
    ],
    "internal_chat_message_metadata_passthrough": {
      "content_item_kinds": ["plugins.recommendations", "environments.environment_context"]
    }
  }
}
```

Both blocks are excluded. If the first block instead has the origin `user.text`, that text is retained unchanged. This example demonstrates supported parser behavior; it is not a claim that a Desktop session without AGENTS instructions was captured.

With a subsequent arithmetic exchange, the normalized retained document still contains:

```jsonl
{"role":"system","content":"REF-ID: example-conversation"}
{"role":"user","content":"What is 2 + 2?"}
{"role":"assistant","content":"I will calculate the sum."}
{"role":"action","content":"calculator"}
{"role":"assistant","content":"4"}
```

All examples and public test fixtures are synthetic. No private transcript, environment content, or conversation identifiers are included.

## Regression coverage

- Partial startup combinations: missing AGENTS, missing plugins, and plugins alone.
- Genuine user text matching the full Desktop envelope or existing CLI prefixes.
- Mixed startup/user blocks, image index alignment, unknown origin kinds, and other injected kinds remaining intact.
- Missing, malformed, empty, and misaligned metadata falling back to existing behavior.
- Assistant commentary, final answers, and phase-less messages remaining unchanged regardless of origin labels.
- Retained source, REF-ID metadata, compact tool actions, ordinary append, and duplicate Stop suppression.
- The existing old-document replacement regression, including a subsequent normal append, remains in place.

## Validation

Checks ran against the coding-agents package with isolated test HOME, temporary directories, and Hindsight configuration:

- `npm test`: **874 tests passed across 64 files**.
- `node_modules/.bin/tsc --noEmit`: passed.
- `npm run build`: passed.
- Prettier check on all three changed files: passed. Focused independent review and Ponytail review: no actionable findings.
- The built Stop bundle passed eight subprocess scenarios using synthetic CLI/Desktop rollouts and a loopback mock API. Each verified retained conversation and metadata, the next-turn append, and duplicate Stop suppression. No real retention or model requests were made by these checks.
- The temporarily installed Stop bundle passed the same eight scenarios, with a verified bundle hash and rollback backup.
- Read-only parsing of actual CLI and Desktop rollouts produced exactly the same conversation rows as the prior fix while excluding the known startup blocks. Partial envelopes remain synthetic regression coverage.
- **Live capture with the new local override:** an ordinary completed turn in the working conversation triggered its normal Stop hook. Read-only inspection after asynchronous retention matched the completed transcript's normalized rows exactly: genuine user messages, assistant commentary/final answers, compact actions, and REF-ID were preserved; the known injected startup blocks were absent. The installed bundle hash was verified. No manual Stop, retain, or replay was used.

The earlier repository-wide lint attempt could not finish because control-plane dependencies were not installed and `@eslint/js` could not be resolved. Package validation and formatting of the changed files are reported separately above.

## Retention and rollout

The shared reader serves both Codex entry points. Local validation uses a temporary override of the self-contained Stop bundle, with a byte-verified backup and rollback instructions. Generated bundles and private validation artifacts are excluded from this PR.

If removing previously retained startup content changes an active document's prefix, its next ordinary Stop can use the existing full-document replacement path and cause re-extraction of that document. Later turns return to normal append behavior. No cursor, replacement, or extraction logic changes here.

No final-only filtering, new commentary/tool-retention settings, historical bank cleanup, forced re-ingestion, or bulk replay is included.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


hindsight-integrations/coding-agents/src/core/transcript-codex.ts
```diff
@@ -45,7 +45,22 @@ interface RolloutLine {
...
+  // Desktop prepends plugin guidance and omits "for <path>" from the AGENTS heading. Consume
+  // each complete block separately so matching cannot backtrack across trailing user prose.
+  let rest = s;
+  for (const block of [
+    /^<recommended_plugins>.*?<\/recommended_plugins>\s*/s,
+    /^# AGENTS\.md instructions(?: for [^\r\n]+)?\r?\n\s*<INSTRUCTIONS>.*?<\/INSTRUCTIONS>\s*/s,
+    /^<environment_context>.*?<\/environment_context>\s*/s,
+  ]) {
+    const match = block.exec(rest);
+    if (!match) return false;
```
ebarkhordar, line 61: I ran this against `ade8f410` in a clean `node:22-slim` container. Your 13 `transcript-codex` cases pass there.

The loop bails with `return false` on the first block that does not match, so all three blocks have to be present and in this order. A Desktop envelope missing any one of them is retained whole. Driving the real `readCodexTranscript`:

```
DROPPED   all three (the envelope this PR observed)
RETAINED  no AGENTS block (project without an AGENTS.md)
RETAINED  no plugins block (Desktop heading without "for <path>")
RETAINED  plugins block alone
DROPPED   env block alone (the pre-existing CLI check)
```

The AGENTS one is what I would worry about, since Codex only injects that block when the project has an AGENTS.md. I have not captured a Desktop rollout without it, so I cannot tell you that shape really occurs; that is the open question rather than a claim.

Skipping a block instead of bailing covers all five, and the terminator you already have is what keeps the false positives out:

```ts
    const match = block.exec(rest);
    if (!match) continue;
    matched = true;
    rest = rest.slice(match[0].length);
  }
  return matched && rest.length === 0;
```

`matched` is there so an empty string does not qualify. With that change `transcript-codex.test.ts`, `retain-hook.test.ts` and the five probes above run 36 passed, and all six of your quoted-example cases still keep the user's prose.

Korayem: > A Desktop envelope missing any one of them is retained whole.

Thanks for running the clean-container checks and calling out this gap, @ebarkhordar. Your comment prompted us to look more closely at how Codex records the origin of these blocks.

We reproduced the partial-envelope behavior. While checking the optional-block approach, we also found the ambiguous case: a genuine user can submit exactly `<recommended_plugins>...</recommended_plugins>` as their prompt. Both injected context and genuine prompts can have `role: "user"`, so text and role alone do not settle that distinction.

Local CLI and Desktop rollouts provide a more precise signal in `payload.internal_chat_message_metadata_passthrough.content_item_kinds`, aligned with `payload.content`. Genuine text is labeled `user.text`; the startup blocks use `plugins.recommendations`, `agents_md.instructions`, and `environments.environment_context`.

Updated in `4e0db3f1c`: with valid aligned metadata, the parser now excludes only those three startup origins, per block. This handles the partial combinations you identified while preserving genuine identical markup, mixed messages, and alignment around images. Unknown origins and other injected kinds remain under existing handling. Missing or malformed metadata keeps the conservative text fallback for compatibility.

The regressions cover these cases plus retained source, commentary/final/phase-less answers, compact actions, REF-ID, ordinary append, and duplicate Stop suppression. All **874 tests** pass, along with typecheck, build, and targeted formatting; focused review found no actionable issues.

We also installed the new Stop bundle locally with a verified rollback backup. Both the built and installed bundles passed eight isolated CLI/Desktop subprocess scenarios. Before pushing, we let our working conversation complete normally and then inspected its asynchronously retained document: every normalized conversation row matched the completed raw transcript, the injected startup blocks were absent, and commentary, final answers, compact actions, and REF-ID were preserved. No manual Stop or forced re-ingestion was used.

The partial combinations are synthetic coverage; we are still not claiming a captured Desktop session without AGENTS instructions. No historical replay or cleanup is part of this change. Thanks again for pushing us to check that boundary.

ebarkhordar: Thanks for chasing the origin labels, that is a better signal than the text shape. I read 4e0db3f1: gating on `content_item_kinds` only when the list is a complete aligned match for `payload.content`, and keeping the text fallback otherwise, covers the partial envelopes I hit without breaking a genuine prompt that happens to look like one. The fallback keeps the all three blocks requirement for rollouts with no metadata, which is the right trade to make here. I read the diff and the tests, I did not run the installed bundle.

nicoloboschi:
Thanks a lot for the diagnosis and the regression tests, @Korayem! We went with a different signal: Codex records genuine input as `item_completed` `UserMessage` events, which exist on every build we checked (0.93 → 0.153), including Desktop 0.146/0.147. Those builds don't write `content_item_kinds` and still leaked the two-block startup envelope. That lets the reader drop the text matching entirely. Continued in #4377, which adapts your retain-hook tests.

Issue #4152, opened 2026-09-05
Consolidation delete path errors on missing deletes[].observation_id; observations_deleted stays 0


lstmemery:
## Summary

Hindsight 0.9.1: while draining a consolidation backlog, 15 LLM calls errored with a
ValidationError on a missing `deletes[].observation_id`. Each failed call is discarded
fail-closed, which is the right default — but the net effect is that the
supersession-cleanup path silently does nothing, and `observations_deleted` stayed at 0
for the whole run.

## Environment

- Hindsight 0.9.1 (consolidation pipeline)
- Linux x86_64

## Reproduction

1. Run consolidation with a model whose responses sometimes include a `deletes` array
   whose entries lack `observation_id` (observed: 15 such calls during one backlog
   drain).
2. Observe the per-call ValidationError and the run-level `observations_deleted`
   counter.

## Expected vs actual

- Expected: either `deletes[].observation_id` is optional with a defined fallback for
  delete entries, or the prompt/schema is tightened so the model reliably emits it —
  and whatever failure mode remains is visible in counters.
- Actual: every call that omits the field errors out; no superseded observations are
  deleted; a run where the entire delete path fails is indistinguishable from a healthy
  run without reading per-call logs.

## Notes

- Fail-closed on a malformed delete entry is the right default; the issue is that the
  prompt/schema lets the model omit a required field at all, and nothing at run level
  surfaces the dead path.
- For context: our own memory-store maintenance guidance already recommends using a
  stricter-JSON model for consolidation calls. Open question for maintainers: should
  `deletes[].observation_id` be optional, or should the prompt/schema be tightened?

PR #4158, opened 2026-09-06
fix(consolidation): name observation_id as the delete key

ebarkhordar:
Answering the open question at the bottom of your report: tightened, not made optional. A delete with no target isn't actionable, so `observation_id` stays required.

The prompt never told the model which key a delete goes on. `_OBSERVATION_FIELDS` says "`id`: unique identifier, copy this exactly when issuing an UPDATE or DELETE" and that is the only place it names a field for one. Updates get corrected twice after that, in the DECISION GUIDE and in the `observation_id` field rule; deletes get neither, and the one worked example that showed a populated `deletes` entry went out in #1759 with the old single example. So a model that follows the prompt literally emits `{"id": ...}`, `_DeleteAction` requires `observation_id`, and `_classify_batch_failure` has ValidationError in FAIL_FAST. That's the 15 discarded calls and the 0.

Two changes. The input-field line now names the destination, `copy it verbatim into observation_id`, matching the `[uuid]` line for `source_fact_ids` just above it, and the `deletes` field rule says it too. And `_DeleteAction` takes `id` as a validation alias, since with `llm_strict_schema_consolidation` off the prompt is guidance rather than a grammar, and a model echoing the input key shouldn't cost the batch.

What providers receive is unchanged: one property `observation_id`, still required, from both the plain and the strict schema generator, and `model_dump` is the same. Updates still take `observation_id` only.

Four tests. The prompt one and the `id`-keyed delete one fail on main. The other two pin the wire contract and the canonical key and pass either way, so I checked them by flipping the alias order and then dropping the canonical name, which turns each of them red.

I haven't touched the other half of your report, that a run with a dead delete path looks like a healthy one at the counter level. Separate change and I'd rather not bundle it.

Refs #4152

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -30,7 +30,7 @@ from itertools import combinations
 from typing import TYPE_CHECKING, Any, Literal
 
 import asyncpg
-from pydantic import BaseModel, ValidationError, field_validator
+from pydantic import AliasChoices, BaseModel, Field, ValidationError, field_validator
 
 from ...config import get_config
 from ...worker.stage import set_stage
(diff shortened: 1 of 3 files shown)


ebarkhordar:
Closing this in favour of #4183, which covers the same ground and goes further: the `id` alias is there, the prompt rule is stronger than mine, and it adds the batch-failure counter that makes a discarded response visible in the first place. Nothing here is left unlanded.

PR #4159, opened 2026-09-06
fix(reflect): enforce max_tokens after forced synthesis

### LINKED ISSUE #4156: reflect: forced final synthesis lost max_tokens enforcement in #3389 (rewrite only runs on the done path)
**Version:** hindsight-api 0.9.2 (tag `52dcd3f8`); unchanged on `main` at `eb32332e`.

## Summary

#3389 turned reflect `max_tokens` into a visible-length target: a prompt directive plus the post-hoc rewrite. The rewrite only exists in `_process_done_tool`. `_forced_final_synthesis` had its hard `max_completion_tokens` cap removed in the same change and never received the rewrite, so on that path `max_tokens` is a hint the model can ignore.

This inverts #2756, which #2757 fixed in 0.8.5:

| Completion path | 0.8.5 (#2757) | 0.9.1+ (#3389) |
|---|---|---|
| `done` tool | rewrite | rewrite |
| forced final synthesis | hard `max_completion_tokens` | **nothing** |

The forced path was then rebuilt as split synthesis by #3392 (closing #3122) the day after #3389, and neither change carried the rewrite over. The `evidence_truncated` signal proposed in #3126 was declined because split synthesis keeps all evidence; that covers the input side, not answer length.

## Where (v0.9.2, `hindsight-api-slim/hindsight_api/engine/reflect/agent.py`)

- L1266, inside `_process_done_tool` (L1217):
  `if llm_config and max_tokens is not None and count_cl100k_tokens(answer) > max_tokens:` → rewrite with `Target budget: {max_tokens} tokens`.
- L705 `_forced_final_synthesis`: passes `max_tokens` to `build_final_prompt` (L725) / `build_reduce_prompt` (L754) for the directive only, then L775 `_log_completion(..., forced=True)` and `return ReflectAgentResult(text=answer, ...)` with no length check.
- Reached from L805 (last iteration), L817 (context-budget guard), L908 and L933 (error-recovery exits).
- On `main` (`eb32332e`) the same shape is at L770 / L1392, with `count_prompt_tokens`.
- `tests/test_reflect_page_length_decoupling.py::test_forced_synthesis_uncapped_by_default` pins the transport cap only; nothing asserts visible length on this path.

In #3389's diff the four forced-synthesis hunks replace `max_completion_tokens=max_tokens` with `synthesis_max_completion_tokens` and add `max_tokens=max_tokens` to the prompt builder; no rewrite is added. #2757's diff adds the rewrite to `_process_done_tool` only.

## Reproduction (0.9.2, `openai-codex` / `gpt-5.6-luna`, default reflect config)

`POST /v1/default/banks/{bank}/reflect` with `include.tool_calls`, bank of ~7k facts, `reflect_max_iterations` 10, `reflect_max_context_tokens` 100000, `reflect_max_completion_tokens` unset.

| | query | budget | max_tokens | path (engine log) | `trace.llm_calls` | answer (o200k tokens) |
|---|---|---|---|---|---|---|
| A | broad "across everything in memory, summarize every decision, convention and open problem" | low | 200 | `Context budget exceeded on iteration 4: ~100220 tokens >= 100000 limit. Forcing final synthesis.` → `split synthesis over 2 chunks` → `forced` | agent_1..3, final_map_2, final_map_1, final | **643** (3.2×) |
| C | narrow factual question | high | 100 | `done` | agent_1..4, **final_rewrite** | 115 |

Run A's final reduce call in `GET …/llm-requests?operation=reflect&scope=reflect` is `in 12418 out 578`, and its stored prompt carries `## Length … Aim for a complete, self-contained answer of approximately 200 tokens`. There is no `final_rewrite` call in that run. Run C, same engine and bank, shows the `done` path rewriting to budget.

At `budget: low` (`max_iterations` = 5) both low-budget runs I made ended on the forced path, one via the context guard and one on the last iteration, so for low-budget reflects this is the usual ending rather than an edge case.

## Observed on a second 0.9.2 engine over one week

Bank of a few thousand facts, `max_tokens: 2000`, budget low. Grouping `GET …/llm-requests?operation=reflect` by run: 32 reflect runs between 2026-08-30 and 2026-09-06, 12 of them forced. Every answer above ~10k chars (12,749, 16,007 and 29,237 chars, roughly 3.2k, 4k and 7k tokens) came from the forced path with no rewrite call. Every `done`-path run that overran was rewritten to 7.4k–9.9k chars, within ~25% of the target. The rewrite works where it runs.

## Why it matters

Callers that use `max_tokens` to fit a reflect answer inside a downstream limit get the guarantee only on the `done` path, and lose it on exactly the long-running questions where they are most exposed. Mental-model refreshes that end via the forced path can exceed their `max_tokens` again, which is what #2756 was about.

## Suggested fix

Lift the rewrite block from `_process_done_tool` into a helper and call it in `_forced_final_synthesis` after the answer is produced and before `_generate_structured_output`, mirroring the `done` path so structured output derives from the capped text. `llm_config`, `max_tokens`, and the document branch are already in scope. Add a forced-path counterpart to the existing over-budget rewrite test.

If skipping the rewrite there is deliberate (one fewer LLM call after an exhausted budget), it would help to say so in the `max_tokens` docs so callers know they still need their own bound on that path.

koriyoshi2041:
## Problem

`max_tokens` is enforced with a post-hoc rewrite when reflect finishes through the `done` tool, but forced final synthesis returns its answer directly. Long or context-exhausted runs can therefore exceed the caller's visible-length target.

Closes #4156.

## Fix

- Extract the existing final-answer rewrite into a shared helper.
- Apply it to both `done` and forced-synthesis completions.
- Run the rewrite before structured-output extraction so structured output is derived from the bounded answer.
- Include rewrite usage and trace data in forced-synthesis accounting.

## Test

- `uv run --project hindsight-api-slim pytest -q hindsight-api-slim/tests/test_reflect_agent.py -k 'not TestDirectiveLeakageOnEmptyBank and not TestContextOverflowIntegration and not TestMentalModelShortCircuitRealLLM'` (55 passed)
- `uv run --project hindsight-api-slim ruff check hindsight-api-slim/hindsight_api/engine/reflect/agent.py hindsight-api-slim/tests/test_reflect_agent.py`
- `uv run --project hindsight-api-slim ruff format --check hindsight-api-slim/hindsight_api/engine/reflect/agent.py hindsight-api-slim/tests/test_reflect_agent.py`
- `git diff --check`

The four excluded tests require an embedded PostgreSQL extra or live LLM credentials; the first full-file run reached 55 passing unit tests before those environment-only setup errors.

## Risk

Forced synthesis may make one additional LLM call when its answer exceeds `max_tokens`. Calls without `max_tokens` and answers already within budget are unchanged, and the rewrite remains governed by `reflect_max_completion_tokens` rather than using the visible target as a hard transport cap.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
+++ b/hindsight-api-slim/hindsight_api/engine/reflect/agent.py
@@ -18,6 +18,7 @@ from ...config import get_config
 from ..llm_interface import LLM_TOOL_CHOICE_AUTO, LLMToolChoice
 from ..llm_trace import LLMQueueWait, reset_queue_wait_sink, set_queue_wait_sink
 from ..llm_transport import describe_llm_error
+from ..response_models import TokenUsage
 from .models import DirectiveInfo, LLMCall, ReflectAgentResult, StructuredOutputResult, TokenUsageSummary, ToolCall
 from .prompts import (
     _SPLIT_SYNTHESIS_WARN_CHUNKS,
@@ -836,6 +837,26 @@ async def _run_reflect_agent_inner(
                 f"over {len(chunks)} context chunk(s)."
             )
 
+        answer, _, rewrite_usage, rewrite_trace = await _rewrite_final_answer(
+            answer,
+            llm_config,
+            max_tokens,
+        )
+        if rewrite_usage is not None:
+            total_input_tokens += rewrite_usage.input_tokens
+            total_output_tokens += rewrite_usage.output_tokens
+            total_cached_tokens += getattr(rewrite_usage, "cached_tokens", 0) or 0
+            total_thoughts_tokens += getattr(rewrite_usage, "thoughts_tokens", 0) or 0
+        if rewrite_trace is not None:
+            llm_trace.append(
+                {
+                    "scope": rewrite_trace.scope,
+                    "duration_ms": rewrite_trace.duration_ms,
+                    "input_tokens": rewrite_trace.input_tokens,
+                    "output_tokens": rewrite_trace.output_tokens,
+                }
+            )
+
         structured_output = None
         # ``answer`` is non-empty past the guard above, so only the schema gates this.
         if response_schema:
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Closing as superseded by #4180, which covers the same forced-synthesis length boundary and landed the stronger shared rewrite path. Thanks for picking this up.

PR #4164, opened 2026-09-06
fix(consolidation): expose failed LLM batches

### LINKED ISSUE #4151: Consolidation responses failing schema validation are not counted in failed_consolidation
## Summary

Hindsight 0.9.1: in a single day, 27 of 528 consolidation LLM calls failed
`_ConsolidationBatchResponse` validation, while the `failed_consolidation` counter
stayed at 0. The failures are silently absorbed.

Silently-absorbed validation failures can mean lost merges, and as-is the
`failed_consolidation` metric cannot be used for monitoring or alerting.

## Environment

- Hindsight 0.9.1 (consolidation pipeline)
- Linux x86_64

## Reproduction

1. Run consolidation over a corpus large enough that some LLM responses do not
   validate against `_ConsolidationBatchResponse` (observed: 27 of 528 calls in one
   day with a model that does not reliably emit strict JSON).
2. Compare the number of per-call validation failures against the
   `failed_consolidation` metric for the same window.

## Expected vs actual

- Expected: schema-validation failures increment `failed_consolidation` (or a dedicated
  counter) so operators can see the failure rate.
- Actual: 27 validation-failed calls, counter reads 0.

koriyoshi2041:
Closes #4151.

## What changed

Add `hindsight.consolidation.batch_failures`, a counter for LLM batches that exhaust consolidation's retry policy. It uses a bounded `reason` label:

- `response_validation` for malformed JSON, schema validation, and output-length failures
- `provider` for other exhausted provider failures

The existing `failed_consolidation` gauge remains unchanged: it counts source memories permanently marked failed after adaptive bisection reaches a singleton. The new counter makes recoverable batch-level failures visible without changing that state-based contract.

Each failed batch increments once after its retry ladder, rather than once per attempt. Successful batches do not increment it.

## Verification

- `tests/test_metrics.py`: 34 passed
- `TestBuildResponseModel`: 10 passed
- focused Ruff check and format check passed
- `git diff --check` passed

The repository hook reached the control-plane lint step but could not load `@eslint/js` because Node dependencies are not installed in this worktree. The changed Python files and tests passed their focused gates.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
+++ b/hindsight-api-slim/hindsight_api/engine/consolidation/consolidator.py
@@ -33,6 +33,7 @@ import asyncpg
 from pydantic import BaseModel, ValidationError, field_validator
 
 from ...config import get_config
+from ...metrics import get_metrics_collector
 from ...worker.stage import set_stage
 from ..db import DatabaseBackend
 from ..db_utils import acquire_with_retry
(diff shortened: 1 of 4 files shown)


koriyoshi2041:
Two CI failures look unrelated to this metrics-only diff:

- `verify-generated-files` stopped in `setup-uv` while fetching the latest uv version (`fetch failed`), before generation ran.
- free-threaded 3.14 completed 6,987 tests and failed only the existing OpenCode session-header fixture after `LLMInterface.call()` became a named result (`result.ok` / payload-vs-`.content`); the changed consolidation tests passed. PR #4159 hit the same free-threaded job failure on its independent branch.

I cannot rerun upstream jobs from the fork. Could a maintainer rerun the failed jobs?

PR #4168, opened 2026-09-06
fix(db): prevent nested-loop observation scoring

### LINKED ISSUE #4163: expand_observations(): planner misestimates CTE cardinality (1 vs 10k) → 94M-comparison nested loop in scored CTE → consolidation recall timeout livelock on large banks
## Summary

On a large single-bank deployment (~37.7k `memory_units`, ~152k `unit_entities`, ~7.2k `fact_type='observation'` rows), the entity-expansion SQL inside `PostgreSQLOps.expand_observations()` hits a planner misestimate that turns the `scored` CTE into a nested loop with **94.5M join-filter comparisons** — 23.6s on an idle database, and far beyond the asyncpg command timeout (~60s default, `HINDSIGHT_API_DB_COMMAND_TIMEOUT`) when several tag-scoped consolidation groups run concurrently. The resulting `TimeoutError` fails the whole consolidation recall, and because retry re-selects the same deterministic seed batch, the task loops `pending` → retry → identical failure indefinitely (>24h in our case) until the planner behavior changes.

## Environment

- Hindsight 0.9.1 (digest-pinned image); SQL verified unchanged on current `main` (`hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py`, `expand_observations` ~line 890)
- Embedded PostgreSQL 18.1 (pg0), 8 vCPU container, 256MB shared_buffers / 64MB work_mem (image defaults)
- Bank scale: `memory_units` ≈ 37.7k rows; `unit_entities` ≈ 152k rows; observations ≈ 7.2k; entity degree distribution has a few hubs at 6k–8.8k `unit_entities` rows (expected for a corpus dominated by a handful of recurring entities)

## Observed

Consolidation on the bank stopped making progress; each task attempt logged:

```
RuntimeError: Failed to search memories (TimeoutError): TimeoutError()
  ...
  File ".../consolidation/consolidator.py", in _find_related_observations
  ...
  File ".../db/postgresql.py", in fetch
    return await self._conn.fetch(query, *args, timeout=timeout)
```

PostgreSQL log showed the same statement canceled on every attempt (`canceling statement due to user request`), ~62s after task start — the asyncpg per-command timeout, not a lock or a dead backend. The LLM stages around it succeed; only this query dies.

## Diagnosis

Enabled `log_parameter_max_length_on_error` to capture the actual bind values from the cancel context (20 seed UUIDs from the semantic-seed step, budget `$2=100`), then ran the exact statement with `EXPLAIN (ANALYZE, BUFFERS)`:

```
CTE connected_sources  (planned rows=1, actual rows=9991)
...
->  Nested Loop  (...)
      Join Filter: (s.source_id = cs.source_id)
      Rows Removed by Join Filter: 94459946
      ->  CTE Scan on connected_sources cs  (actual rows=9991)
      ->  Nested Loop (per-candidate unnest)
            ->  CTE Scan on candidates c_1  (actual rows=2937)
            ->  Function Scan on unnest s   (rows=3.22 avg)
Execution Time: 23600 ms
```

The planner guesses `connected_sources` at **1 row** (actual **9,991**) and `candidates` at 67 (actual 2,937), so the `scored` CTE join between `unnest(c.source_memory_ids)` and `connected_sources` is costed as trivial and planned as a nested loop: 9,991 × 2,937 × ~3.2 ≈ 94M tuple comparisons. Standalone that is 23.4s of one backend; with ~10 tag groups' recall queries contending on 8 vCPUs, each instance blows past the 60s command timeout → cancel → task failure.

Related growth mechanics already documented in the code comment above the query (`source_memory_ids` appends and is never pruned, #1725; the `scored` CTE was rewritten for linearity in #3085): the rewrite made the work linear in source ids *given a hash join*, but nothing pins the planner to a hash join — and under the 1-row CTE estimate it picks the nested loop.

### Why it livelocks instead of just failing

On timeout the consolidation task is rescheduled, but the retry re-selects the same oldest-unconsolidated batch → same semantic seeds → the identical statement with identical binds. Once the plan flips to the nested-loop shape, every retry is deterministically doomed: `pending` → claim → 60s → cancel → `pending`, forever. In our case the bank crossed the cardinality threshold mid-backlog, so ~10k memories sat unconsolidated for over a day with zero progress and no terminal failure.

Note: #3723 (fixed in 0.9.2) added the missing `asyncio.wait_for`/fallback guard to `_expand_observations()` so recall degrades instead of raising. We run 0.9.1, hence the raw propagation. But even with the guard, the 0.9.2 behavior on such a bank is a silent drop of the entity arm on *every* consolidation pass — which feeds the UPDATE-vs-CREATE duplicate-merge decision (see the interleave rationale in `_find_related_observations`), so consolidation quality degrades while the underlying plan flip remains.

## Suggested fix

Make the CTE cardinality visible to the planner so the `scored` join hashes instead of nesting. Cheapest options:

1. `WITH connected_sources AS MATERIALIZED ( ... )` — forces materialization; the CTE scan is then costed from actual output, and the `scored` join plans as a hash join (the shape the #3085 rewrite already assumes).
2. Alternatively, pre-aggregate `connected_sources` into a temp table / `array_agg` and join `unnest` against it explicitly.

## Measured workaround (for anyone hitting this now)

```sql
ALTER DATABASE <db> SET enable_nestloop = off;
```

Same captured statement, same binds: 23,600 ms → **~300 ms** (the pathological node drops to ~7 ms as a hash join; the per-entity LATERAL index scans survive). New sessions only — terminate stale pool backends or restart the API afterward. We saw no plan regressions elsewhere on this workload, but it is a blunt instrument; the MATERIALIZED fix above is the proper one.

## Prior art

- #1725 — `source_memory_ids` unbounded growth (makes `candidates`/`connected_sources` fan-out grow with bank age)
- #3085 — `scored` CTE linearity rewrite (the hash-join assumption this report shows can silently break)
- #3510, #3723 — timeout handling around this query

r266-tech:
## Summary

- materialize the observation `scored` CTE and use an equality `FULL OUTER JOIN`, which PostgreSQL must execute as a hash or merge join even when `connected_sources` is estimated at one row
- preserve the existing distinct shared-source score; unmatched connected sources collapse into a null-id group that the downstream candidates join discards
- pin the planner-safe SQL shape with a structural regression test

This intentionally does not add `MATERIALIZED` to `connected_sources`: it is already referenced twice and therefore materialized by PostgreSQL, and an explicit keyword did not change the bad cardinality estimate in PostgreSQL 18.

Closes #4163

## Validation

- PostgreSQL 18 live semantic check: candidates sharing 3 and 1 connected sources still score `3.0` and `1.0`
- PostgreSQL 18 `EXPLAIN ANALYZE`, normal statistics: 73ms total versus 75ms for the original query in the same 37.7k-unit/150.8k-edge fixture
- PostgreSQL 18 forced one-row estimate (actual ~10k): `Hash Full Join` retained; scoring join ~1.3ms
- `uv run --extra embedded-db pytest tests/test_observation_expansion_single_fetch.py tests/test_observation_expansion_scoring.py -q -n0` (3 passed, 7 skipped because local ML extras are unavailable)
- `uv run --frozen ruff check hindsight_api/engine/db/ops_postgresql.py tests/test_observation_expansion_single_fetch.py`
- `uv run --frozen ruff format --check hindsight_api/engine/db/ops_postgresql.py tests/test_observation_expansion_single_fetch.py`
- `uv run --frozen ty check hindsight_api`
- `git diff --check`

The repository-wide lint hook also ran; all Python checks passed, while the unrelated control-plane ESLint task could not start because `@eslint/js` is not installed in this checkout.

## CI follow-up

The only red job is the free-threaded 3.14 full-suite run. Two of its three failures are the same untouched OpenCode/named-result failures present on recent unrelated runs for #4164, #4159, and #4158. The remaining `test_parallel_fans_out_across_schemas` short count is also outside this diff and passed 3/3 isolated repetitions locally. All other 14 executed jobs passed, including generated-file verification and Python 3.11–3.14 builds; no changed SQL/test failure was reported, so the verified head is unchanged.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
+++ b/hindsight-api-slim/hindsight_api/engine/db/ops_postgresql.py
@@ -936,6 +936,14 @@ class PostgreSQLOps(DataAccessOps):
         # grow hubs far past that, re-measure before assuming this is still the
         # right shape.
         #
+        # Even with that traceable column, bank-wide unit_id statistics can still
+        # collapse the connected_sources estimate to one row (#4163). An inner
+        # join may then become a nested loop over every candidate/source pair.
+        # Keep scored materialized and use a FULL JOIN: PostgreSQL cannot execute
+        # an equality FULL JOIN as a nested loop, so it must hash or merge even
+        # when the estimate is wrong. Unmatched rows form a single NULL-id group
+        # which the following inner join to candidates naturally discards.
+        #
         # Entity/source traversal and semantic/causal expansion run as ONE query
         # (#3857): the observation entity arm is fused into the semantic/causal CTE
         # query behind an 'entity' source discriminator, like the non-observation
@@ -987,11 +995,11 @@ class PostgreSQLOps(DataAccessOps):
                   AND mu.source_memory_ids && ca.source_ids
                   {window.clause("mu")}
             ),
-            scored AS (
+            scored AS MATERIALIZED (
                 SELECT c.id, COUNT(DISTINCT cs.source_id)::float AS score
                 FROM candidates c
                 CROSS JOIN LATERAL unnest(c.source_memory_ids) AS s(source_id)
-                JOIN connected_sources cs ON cs.source_id = s.source_id
+                FULL OUTER JOIN connected_sources cs ON cs.source_id = s.source_id
                 GROUP BY c.id
             ),
             observation_entity_expanded AS (
--- a/hindsight-api-slim/tests/test_observation_expansion_single_fetch.py
+++ b/hindsight-api-slim/tests/test_observation_expansion_single_fetch.py
@@ -419,6 +419,29 @@ async def test_postgresql_fused_query_is_one_statement():
 
 
 @pytest.mark.asyncio
+async def test_postgresql_scoring_join_excludes_nested_loop_plan():
+    """The scored join stays hash/merge-capable when source estimates collapse (#4163)."""
+    conn = AsyncMock()
+    conn.fetch.return_value = []
+
+    await PostgreSQLOps().expand_observations(
+        conn,
+        "memory_units",
+        "unit_entities",
+        "memory_links",
+        [uuid.uuid4()],
+        100,
+        200,
+        UpdatedWindow(after=None, before=None, first_param_index=3),
+    )
+
+    normalized_sql = " ".join(conn.fetch.await_args.args[0].split())
+    assert "scored AS MATERIALIZED" in normalized_sql
+    assert "FULL OUTER JOIN connected_sources cs ON cs.source_id = s.source_id" in normalized_sql
+    assert "GROUP BY c.id" in normalized_sql
+
+
+@pytest.mark.asyncio
 async def test_oracle_fused_query_is_one_statement():
     """Structural (no Oracle runtime): Oracle emits ONE statement, three arms.
 


koriyoshi2041 (approved):
Validated the exact head locally against PostgreSQL: the fused-query and scoring suites pass 10/10, including distinct shared-source scores, wide source arrays, per-entity caps, ordering, and the new plan-shape guard. Ruff check/format and diff hygiene also pass. The FULL JOIN preserves matched candidate scores while preventing an equality full join from degrading to the reported nested-loop shape.

PR #4170, opened 2026-09-06
fix: count active operations and preserve agent roster freshness

PaleRoses:
## Scope

- Add `active_only` to operation listing (HTTP, MCP and generated SDKs). It filters pending/processing records before counting; the coding-agent client reads that total rather than filtering the newest 20 results. Unavailable or malformed counts are not reported as zero.
- Preserve the knowledge tree's existing `is_stale` flag through the coding-agent client into SessionStart and periodic rosters. Unknown freshness stays unmarked; a refreshed page loses its stale marker.

Individual page reads, exports, response schemas and the control-plane freshness path remain unchanged. This revision removes the proposed per-page freshness queries.

## Reproduction and boundaries

Upstream reports zero active operations when its newest 20 records are completed and an older operation remains pending. Its page-list adapter also discards a true freshness flag already present in the tree.

The exact count includes deferred/parked operations. Existing bank-wide readiness and settle policies are unchanged; a busy shared bank can still reach the settle deadline. Older servers ignoring `active_only` return the full history total, an overcount.

## Verification

- 387 backend tests passed across operation status, async batch retain, MCP tools and knowledge-base suites, using mock LLMs and an isolated PostgreSQL instance.
- 879 coding-agent tests passed across 67 files; TypeScript, formatting and bundle build passed. Three retained regressions fail against upstream: truncated counting, fabricated zero on unavailable counts, and missing stale markers. Three source-text assertions were removed; behavioral coverage remains.
- Built status CLI reports the older pending operation; built Claude SessionStart hook emits the stale marker and leaves the fresh page unmarked. Both ran against a read-only HTTP fixture.
- OpenAPI, documentation projections and SDKs regenerated with the pinned generators. Rust SDK, Go SDK, TypeScript SDK and Rust CLI builds passed; Ruff and Python type checks passed.

No live deployment, private-bank migration, full-monorepo suite or live-provider testing.


nicoloboschi:
you can filter on status already

PR #4171, opened 2026-09-07
fix(memories): use GIN for tagged mental-model staleness negatives

### LINKED ISSUE #4169: Batched mental-model staleness check does a full-bank scan per scope in the common "already fresh" case
# Batched mental-model staleness check does a full-bank scan per scope in the common "already fresh" case

## Summary

`any_memory_updated_since_batch` (`hindsight_api/engine/memories/pg/reads.py`, added in #3589) is fast only when a scope's LATERAL subquery finds a matching row quickly. When a tagged scope has **no** memory written since its watermark — the common case once a bank is mostly caught up — the same query walks the entire bank per scope instead of using the GIN tag index that already exists.

In production, this shows up as: Envoy logs 127 `504`s in 12h on our UI host (path not recorded by our access-log config), each at ~15,060ms — its own upstream route timeout firing. Correlated by timestamp, hindsight-api's own log shows a `TimeoutError` traceback (raised inside asyncpg's `protocol.pyx`) for `GET .../mental-models`, in this exact function, in the same windows. Our own isolated reproduction of the batch (below) takes 8.4s warm, single-shot, no concurrent load — already large enough that under real production conditions (concurrency, cache pressure) it plausibly reaches the full 15s Envoy saw. (Reproduction detail and a predicate caveat are below.)

## Why this wasn't caught by #3589's own benchmark

#3589 measured 30–300 scopes on a synthetic 400k-row table (200k in the hot bank) and reported 0.18–4.3ms. That benchmark's scopes were all in the **positive** case — a matching write exists, so the `ORDER BY updated_at DESC LIMIT 1` LATERAL finds it within a few rows and the `(bank_id, updated_at)` index terminates the scan early. The PR's own docstring states the design assumption plainly:

> "the scans themselves are microseconds once `idx_memory_units_bank_updated_at` exists"

That's true for the positive case. It is not measured for the **negative** case — a tagged scope where nothing in-scope has been written since the watermark. There, `updated_at > s.since` starts satisfying immediately (recent writes to *other* tags/pages exist), but the tag-containment filter then has to reject every one of those rows one at a time until the scan reaches the offset where the last write to *this* scope's tag actually lives — which for a scope that's been quiet for a while is most of the table. LIMIT 1 never gets to stop early because nothing matching is found until (or unless) the scan finishes.

## Reproduction

```sql
EXPLAIN (ANALYZE, BUFFERS)
SELECT s.scope_key, (h.hit IS NOT NULL) AS stale
FROM jsonb_to_recordset('[{"scope_key":"a","since":"2026-07-21T09:58:00+00:00","tags":["project:example"],"fact_types":null}]'::jsonb)
     AS s(scope_key text, since timestamptz, tags varchar[], fact_types text[])
LEFT JOIN LATERAL (
  SELECT 1 AS hit FROM memory_units mu
  WHERE mu.bank_id = 'redacted' AND mu.updated_at > s.since
    AND mu.tags IS NOT NULL AND mu.tags != '{}' AND mu.tags @> s.tags
    AND (s.fact_types IS NULL OR mu.fact_type = ANY(s.fact_types))
  ORDER BY mu.updated_at DESC LIMIT 1
) h ON TRUE;
```

Result on our data — one scope, `all_strict`-mode tag with no recent write, watermark ~7 weeks old. `all_strict` is the default `tags_match` for any mental model that carries its own tags and doesn't override the mode; that's 214 of our 215 models:

```
Index Scan using idx_memory_units_bank_updated_at on memory_units mu
  Index Cond: ((bank_id = 'redacted') AND (updated_at > s.since))
  Filter: (tags IS NOT NULL AND tags <> '{}' AND (tags @> s.tags) AND ...)
  Rows Removed by Filter: 123074
Execution Time: 247.347 ms
```

247ms for **one** scope, walking the whole bank because nothing matches. The `GET /mental-models` payload batches all 214 tagged scopes at once; production's own timeout and error log (above) is the direct evidence this happens for real. Our own reproduction of that batch, on the same real data, is below — same plan shape (btree walk past everything, GIN unused) repeated 214 times.

```
Execution Time: 8426.177 ms   (batch of 214, reproduced with `&&` — see caveat below)
```

That 8.4s batch reproduction used `&&` (overlap), which we later found is not the exact production predicate — production is `@>` (containment), confirmed by capturing the live query text from `pg_stat_activity`. We re-ran the single-scope case with the corrected `@>` predicate (above, 247ms) and it shows the identical plan shape and a comparable per-scope cost to the `&&` version (209ms), so we're confident the same mechanism drives the batch cost regardless of which array operator is used. We did not re-run the full 214-scope batch with `@>` to avoid adding further load to a live incident; the 8.4s figure above should be read as "our best batch-level reproduction of the mechanism," not a byte-for-byte capture of one production request.

## A candidate fix

Replacing the `ORDER BY ... LIMIT 1` LATERAL with an aggregate lets the planner use the GIN `tags` index instead of the `(bank_id, updated_at)` btree:

```sql
LEFT JOIN LATERAL (
  SELECT bool_or(mu.updated_at > s.since) AS hit
  FROM memory_units mu
  WHERE mu.bank_id = $1
    AND mu.tags IS NOT NULL AND mu.tags != '{}' AND mu.tags @> s.tags
    AND (s.fact_types IS NULL OR mu.fact_type = ANY(s.fact_types))
) h ON TRUE
```

Same single scope, `@>` predicate, outer projection changed to `COALESCE(h.hit, false)` since `bool_or` always returns one row per group (unlike the `LIMIT 1` version, where a genuine no-match produces no row at all):

```
Bitmap Heap Scan on memory_units mu
  Recheck Cond: (tags @> s.tags)
  -> Bitmap Index Scan on idx_memory_units_tags
       Index Cond: (tags @> s.tags)
Execution Time: 1.022 ms (query cost; ~33ms wall time includes one-off JIT compilation)
```

247ms → ~1ms for the query itself, via the GIN index instead of the btree walk. We ran this for one scope only, not the full 214-scope batch, to keep this measurement pass bounded. **This is not yet a drop-in fix** — three things still need verifying before it's a patch rather than a proof of mechanism: (1) `build_tags_where_clause` supports `any` / `all` / `any_strict` / `all_strict`; we've only measured `all_strict`, which is the default for any model that carries tags (214 of our 215) but not necessarily the only mode present if some model overrides it; (2) `tag_groups` scopes are explicitly handled elsewhere (one-at-a-time fallback per the existing source comment) and this change doesn't touch that path; (3) the aggregate needs to be proven to return identical `{scope_key: stale}` answers to the current implementation across all four modes, not just checked for using the right index. Posting it here as a starting point for whoever picks this up, not as a ready patch — happy to do that equivalence work and open a PR if useful.

## Environment

- `ghcr.io/vectorize-io/hindsight-api:0.9.2` (current release; confirmed present on `main` as of writing — not fixed by any commit since #3589)
- Postgres 15.4, pgvector
- `memory_units`: 133,293 live rows in the affected bank; `mental_models`: 215 rows, 214 tagged

leilei3167:
## What Problem
`any_memory_updated_since_batch` (from #3589) is fast when a scope finds a matching write quickly, but in the common "already fresh" case a tagged scope walks the whole bank via `idx_memory_units_bank_updated_at` and filters tags row by row. Under production load that shows up as Envoy ~15s `504`s and asyncpg `TimeoutError` on `GET .../mental-models` (see #4169).

## Why
For tagged scopes, move `updated_at > s.since` into `bool_or(...)` and keep the tag predicate in `WHERE`, so Postgres can drive from `idx_memory_units_tags` instead of the btree walk. Untagged scopes still use the `ORDER BY ... LIMIT 1` plan that #3589 tuned for whole-bank watermarks. `tag_groups` still fall back one-at-a-time.

## How verified
- `pytest tests/test_any_memory_updated_since_batch_gin_4169.py` — tagged path uses `bool_or` (no btree `ORDER BY`/`LIMIT 1`); untagged keeps early-exit; all four `tags_match` modes use the GIN path when tagged (3 passed).
- Extended `test_batched_staleness_matches_the_single_model_answer` with `any`/`all`/`any_strict`/`all_strict` plus a quiet tagged negative scope so batch and one-at-a-time stay aligned.

Fixes #4169

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memories/pg/reads.py
+++ b/hindsight-api-slim/hindsight_api/engine/memories/pg/reads.py
@@ -547,23 +547,25 @@ async def any_memory_updated_since_batch(
 
     The knowledge tree and the mental-model list ask this for every model in the
     bank on a poll, and one round-trip each is what made the exact answer look
-    expensive — the scans themselves are microseconds once
-    ``idx_memory_units_bank_updated_at`` exists. Scopes are grouped by the tag
-    clause they generate (a bank's pages almost always share one), each group is
-    joined against its scope set as JSON, and every group is a single prepared
-    plan however many pages it covers.
+    expensive. Scopes are grouped by the tag clause they generate (a bank's pages
+    almost always share one), each group is joined against its scope set as JSON,
+    and every group is a single prepared plan however many pages it covers.
 
-    Two details in the SQL are load-bearing:
+    The LATERAL shape depends on whether the group has a tag filter:
 
-    * ``ORDER BY mu.updated_at DESC`` inside the LATERAL. It does not change the
-      answer — we only ask whether a row exists — but without it the planner
-      estimates the ``LIMIT 1`` will be satisfied early, picks a sequential scan,
-      and the whole point is lost (measured: 13 ms per scope instead of 0.03 ms).
-      The ORDER BY makes the ``(bank_id, updated_at DESC)`` index the obvious way
-      to run the join, which is also the cheapest.
+    * Untagged scopes keep ``ORDER BY mu.updated_at DESC LIMIT 1`` so the planner
+      walks ``idx_memory_units_bank_updated_at`` and stops at the first hit — the
+      common positive case for a whole-bank watermark.
+    * Tagged scopes use ``bool_or(mu.updated_at > s.since)`` with the tag
+      predicate in ``WHERE`` (and *without* an ``updated_at`` range predicate).
+      Putting ``updated_at > s.since`` in the aggregate lets Postgres drive from
+      ``idx_memory_units_tags`` instead of walking every newer bank row and
+      rejecting off-tag ones. That negative case is the common one once a bank is
+      mostly caught up, and the btree walk was timing out ``GET /mental-models``
+      under production load (#4169).
     * ``LEFT JOIN LATERAL`` rather than a correlated ``EXISTS``. A scalar subquery
       over a function scan gets no useful row estimate and falls back to a
-      sequential scan for the same reason.
+      sequential scan.
 
     Compound ``tag_groups`` scopes cannot be expressed against a joined row, so
     they fall back to one :func:`any_memory_updated_since` each — they are rare,
@@ -571,9 +573,7 @@ async def any_memory_updated_since_batch(
 
     Postgres only. This module backs both SQL dialects, and Oracle reaches it
     through a regex rewriter that does not model ``jsonb_to_recordset`` or
-    ``LATERAL``, so an Oracle connection takes the same per-scope path. It is the
-    round-trips that are saved here, not the scans — the scans are cheap on both
-    dialects once the ``(bank_id, updated_at)`` index exists.
+    ``LATERAL``, so an Oracle connection takes the same per-scope path.
     """
     if not scopes:
         return {}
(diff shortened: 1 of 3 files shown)


nicoloboschi:
Thanks for this — the diagnosis is right, and the mechanism you identified (the `LIMIT 1` can only stop at a match, so a quiet tagged scope walks every write since its watermark) is exactly what #4195 has just landed a fix for. Closing this as superseded, with credit to the analysis here and in #4169.

Two things came out of benchmarking your shape that are worth recording, because they changed what we shipped:

**The aggregate must not be applied to every *tagged* scope, only to the ones whose clause an index can serve.** This PR gates on `if tag_clause`, which includes the non-strict `any` / `all` modes. Their clause is `(tags IS NULL OR tags = '{}' OR tags <op> s.tags)` — the `OR` makes it un-indexable, so the aggregate cannot reach `idx_memory_units_tags` and instead loses the early exit it had. Measured on 100k rows and 214 scopes:

| `tags_match="any"` | quiet scopes | scope with matches |
|---|---|---|
| `LATERAL … LIMIT 1` | 3656ms | **3.8ms** |
| `bool_or` aggregate | 3307ms | **4083ms** |

No gain on the negative case, and a ~1000× regression on the positive one. #4195 therefore routes on index-eligibility rather than on "has tags" — `_strict` and `exact` take the aggregate, `any`/`all` keep the walk.

**`bool_or` is not an Oracle aggregate.** The single-scope path is what an Oracle connection falls back to, and its rewriter turns `tags @>` / `&&` into `JSON_TABLE` existence predicates, so the shape would be a syntax error there rather than a slower plan. #4195 gates on `conn.backend_type`.

Also worth knowing, since it is the obvious next idea: a composite `btree_gin (bank_id, tags, updated_at)` does **not** rescue the `LIMIT 1` form — the planner still prefers the `ORDER BY` index (3439ms). And putting `updated_at > s.since` back into the aggregate's `WHERE` alongside the tag predicate drops it to a sequential scan (`Rows Removed by Filter: 100214`). Keeping the tag predicate as the *only* index-eligible filter, exactly as you wrote it, is what makes the plan hold.

End to end on the reproduction, `GET /mental-models` went 3502ms → 6.1ms. Thank you — this was a good catch, and the report in #4169 was unusually well evidenced.

PR #4172, opened 2026-09-07
fix(embeddings): include prefixes in input token cap

### LINKED ISSUE #4165: Batch embedder can exceed the embedding model context by one token, failing the task
**Version:** 0.9.2 (self-hosted, `hindsight-api` + `hindsight-control-plane`)

## Summary

The batch embedder can build a request that exceeds the embedding model's context window by a small margin, producing a hard 400 and failing the owning task. Observed overshoot: **8193 tokens against a hard limit of 8192** — one token over.

## Observed

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "You passed 8193 input
tokens and requested 0 output tokens. However, the model's context length is only 8192
tokens, resulting in a maximum input length of 8192 tokens. Please reduce the length of
the input prompt. (parameter=input_tokens, value=8193)"}}

Exception: Failed to generate batch embeddings: Error code: 400 - {...}
ERROR - hindsight_api.engine.memory_engine - Task execution failed: refresh_mental_model
```

The task is then rescheduled ~60s later and, on repeat failure, ends terminally at `status: failed` with `retry_count: 3`. So retries are bounded — this is a correctness bug, not a runaway.

It is not a one-off: two distinct task ids hit it within a recent 3-hour window, and during a period of queue backlog three distinct task ids hit it inside a single 60-second window.

## Expected

The batcher should treat the embedding model's context length as a hard bound and split or truncate to fit, rather than emitting a request it can predict will be rejected. An overshoot of exactly one token suggests an off-by-one in the token budget — e.g. a boundary computed as `<= limit` on one side and `< limit` on the other, or a separator/special token added after the budget check.

## Impact

Affected `refresh_mental_model` tasks fail permanently after their retries are exhausted, so the mental model silently keeps stale content. The failure is only visible in pod logs or by inspecting the operation's `error_message` — there is no signal at the level an operator normally watches.

koriyoshi2041:
## Problem

Embedding inputs are truncated before asymmetric query/passage prefixes are added. An input at the configured token cap can therefore exceed the provider's hard context limit after the backend prepends its prefix, producing a permanent 400 and failing the owning task.

## Fix

Include the selected query or passage prefix while truncating, then remove it before dispatch so the backend adds it exactly once. This preserves the existing backend contract while making the configured cap apply to the actual wire text.

## Test

- `uv run pytest tests/test_embeddings_max_input_tokens.py tests/test_token_encoding.py -q` (29 passed)
- `uv run ruff check hindsight_api/engine/retain/embedding_utils.py tests/test_embeddings_max_input_tokens.py`
- `uv run ruff format --check hindsight_api/engine/retain/embedding_utils.py tests/test_embeddings_max_input_tokens.py`
- `uv run ty check hindsight_api/engine/retain/embedding_utils.py`

## Risk

Low. The behavior changes only when a token cap and a query/passage prefix are both configured. Empty-prefix and uncapped paths remain unchanged. A cap too small to contain the prefix now fails with a clear configuration error.

Closes #4165.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/embedding_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/embedding_utils.py
@@ -27,7 +27,12 @@ class EmbeddingsBackend(Protocol):
     def encode_documents(self, texts: list[str]) -> list[list[float]]: ...
 
 
-def _truncate_inputs(texts: list[str], max_input_tokens: int, backend: EmbeddingsBackend) -> list[str]:
+def _truncate_inputs(
+    texts: list[str],
+    max_input_tokens: int,
+    backend: EmbeddingsBackend,
+    input_type: EmbeddingInputType = "document",
+) -> list[str]:
     """Cap each input at ``max_input_tokens`` tokens before it reaches the provider.
 
     Remote providers with a fixed input-token limit (e.g. Bedrock Titan V2's hard 8192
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Closing as superseded by #4182, which landed the same prefix-aware context bound on main and also covered the knowledge-page/default-cap path from #4165. Thanks.

PR #4203, opened 2026-09-07
fix: defer retain completion outbox until store commit

### LINKED ISSUE #4189: retain.completed reports memory_unit_count: 0 for a store-owned bank — the outbox callback counts before the store has written
`retain.completed` reports `memory_unit_count: 0` for a store-owned bank, because the outbox callback counts before the store has written.

## What happens

For a bank whose memories live outside SQL, `_build_retain_outbox_callback` asks the store for the count:

```python
if _mem.store_owned_for(bank_id):
    _counts = await _mem.document_memory_counts(
        conn=conn, fq_table=fq_table, bank_id=bank_id, document_ids=[data.document_id]
    )
    _count = int(_counts.get(data.document_id, 0))
```

That is the right question, asked too early. The store-owned retain path mints ids first and writes at session commit, so when the callback runs there is nothing to count yet.

## Ordering, from an instrumented run

Probes on `insert_facts`, `index_facts`, the store's `document_memory_counts`, and its retain-session flush, running `test_retain_completed_payload_carries_memory_unit_count` — a **single** retain, no delta path involved:

```
1. insert_facts            document_id=None  facts=4  defer_index=True   <- mints ids, writes nothing
2. document_memory_counts  wanted=['doc-counted']     by_value={}        <- the callback, counts zero
3. retain session flush    docs=['doc-counted']                          <- the actual write
```

`index_facts` never fires on this path, so the write goes through neither of the two documented write methods — it goes through the session commit, which is after the callback.

The store is not at fault: it answers accurately for the state it is in when asked. The count is simply not knowable at step 2.

## Failing tests

```
tests/test_webhooks.py::TestRetainCompletedWebhook::test_retain_completed_payload_carries_memory_unit_count
    assert 0 == 4

tests/test_webhooks.py::TestRetainCompletedWebhook::test_retain_completed_payload_counts_document_not_units_created
    unchanged re-retain reported a zero-fact document: [4, 0]
```

The first is the clearer one: a single retain, and the webhook reports `0` while `list_memory_units` finds 4. The second shows `4` then `0` across two retains, which makes the delta path look responsible — it is not, it is the same ordering.

The consequence matches what the count was added for (#3040): a receiver cannot tell a document that produced memories from one that produced none, and here every store-owned retain looks like the latter.

## Two possible fixes

**Report what the retain already knows.** The retain has just minted the unit ids for this document; counting them needs no round trip to the store at all. Cheapest, and it removes the ordering question rather than sequencing around it.

**Or fire the callback after the store commit**, so the question is asked when it has an answer. Larger change — the callback is queued inside the retain's transaction, and the store write is outside it.

## Unrelated trap found alongside

`fact_storage.insert_facts_batch` takes `document_id` as its **4th positional** parameter:

```python
async def insert_facts_batch(conn, bank_id, facts, document_id=None, ops=None, defer_index=False)
```

All three deferred call sites skip it by jumping to keyword arguments:

```python
await fact_storage.insert_facts_batch(None, bank_id, batch_processed, ops=pool.ops, defer_index=True)
```

So `document_id=None` reaches the store silently, with no error. Harmless today because the session carries the document id, but it is a trap for the next caller — worth making keyword-only.

## Reproducing

Needs a dedicated embedded-Postgres instance; a stale shared one makes the suite error in setup and look broken:

```bash
HINDSIGHT_API_DATABASE_URL="pg0://<name>:<port>" \
  uv run pytest tests/test_webhooks.py::TestRetainCompletedWebhook -n 0 -s
```

kyletser:
## Summary

Fixes #4189.

A store-owned retain session can buffer its facts until `commit()`. The completion outbox currently queries that store from inside the retain pipeline, before the session commits, so a successful retain can announce `memory_unit_count: 0` even though the document ends up with four memories.

This change records the callbacks actually reached by the pipeline and runs them after successful session commit, in a fresh SQL outbox transaction. It preserves document-specific callback factories and keeps counts based on the document's stored total, not the IDs created by this retain. The SQL path and stores without a retain session keep their existing callback timing.

The SQL outbox and an external store are still separate transactional systems; this does not introduce cross-store atomicity. A failed retain/session commit does not publish a success event.

## Tests

- Regression tests cover direct callbacks, per-document factories, unchanged and zero-fact retains, and retain/commit failures.
- Real PostgreSQL tests exercise both the SQL backend and a buffered store. The existing in-memory test fixture now consumes `FactRecord` correctly, preserves document fields and allocated IDs, and honors the document filter used by the count assertion.
- Python 3.11: `pytest tests/test_retain_outbox_session.py tests/test_webhooks.py tests/test_memories_extension.py` — **238 passed, 67 skipped**. Existing local-ML-dependent tests skip in the slim environment; the new count regressions run with deterministic embeddings and a real isolated pg0 database.
- Final focused run after rebasing: **26 passed**.
- Mutation check: disabling deferral makes three database regressions fail again, including `0 != 4` and unchanged-retain counts `[4, 0]`.
- Ruff 0.14.9 API lint, formatting of changed files, changed-runtime-file typecheck with ty 0.0.8, and `git diff --check` passed.

## Local validation limits

The complete API typecheck could not pass in the slim environment because optional ML/provider imports are missing. The all-workspace lint shell script could not start from the CRLF Windows checkout; the Python checks above were run directly. The full repository suite and full free-threaded API environment were not validated locally. A separate source-extracted lifecycle smoke check passed 96 threaded event-loop scenarios on Python 3.14t with the GIL disabled, but it is not a substitute for the free-threaded CI job.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -6131,11 +6131,27 @@ class MemoryEngine(MemoryEngineInterface):
         if pending_outbox_callbacks:
             # Do not hold a SQL connection while committing an external store. The outbox
             # still lives in SQL, but cannot share a transaction with that store's commit.
-            backend = await self._get_backend()
-            async with acquire_with_retry(backend) as conn:
-                async with conn.transaction():
-                    for callback in pending_outbox_callbacks:
-                        await callback(conn)
+            #
+            # The memories are already committed in the store by the time this runs, so a
+            # failure here cannot be undone by failing the retain — it would only mark a
+            # successful retain as failed and invite the caller to re-submit a document that
+            # is already stored. The event is the lossy side of a boundary that is not
+            # transactional either way (see the deferral note above): log it loudly and let
+            # the retain report the truth, which is that it succeeded.
+            try:
+                backend = await self._get_backend()
+                async with acquire_with_retry(backend) as conn:
+                    async with conn.transaction():
+                        for callback in pending_outbox_callbacks:
+                            await callback(conn)
+            except Exception:
+                logger.error(
+                    "[BATCH_RETAIN] bank=%s operation=%s retained successfully but the retain.completed "
+                    "outbox write failed; the completion event is lost for this operation",
+                    bank_id,
+                    operation_id,
+                    exc_info=True,
+                )
 
         return _RetainExecutionResult(
             unit_ids=result,
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -5773,6 +5773,31 @@ class MemoryEngine(MemoryEngineInterface):
             _session_config = await self._resolve_retain_config(bank_id, request_context, strategy)
             retain_session = await _store.begin_retain(bank_id=bank_id, config=_session_config)
 
+        pending_outbox_callbacks: list[RetainOutboxCallback] = []
+        if retain_session is not None:
+            # A session may buffer all memories until commit. Running the outbox in the
+            # pipeline's SQL transaction used to publish a pre-commit document count (#4189).
+            # Record only callbacks the pipeline actually reaches (including per-document
+            # factories), and run them after the owning session has committed successfully.
+            # Stores without sessions and the SQL path keep their existing transaction boundary.
+            def defer_outbox(callback: RetainOutboxCallback | None) -> RetainOutboxCallback | None:
+                if callback is None:
+                    return None
+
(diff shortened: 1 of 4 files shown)


koriyoshi2041 (approved):
The store/SQL boundary looks sound at `d760a00`: reached callbacks are deferred only for store-owned sessions, the store commit happens before the fresh SQL outbox transaction, and retain or commit failures cannot publish a success event. The factory path also preserves per-document callback contents, while unchanged retains count the committed document total rather than newly created IDs.

The focused matrix covers direct/factory callbacks, previous/committed counts, and both failure points. I also checked that the session path does not sub-batch, so the deferred list does not introduce duplicate per-document completion events. Approved.

PR #4212, opened 2026-09-08
test(system): blackbox system-test suite driven only through the public API

nicoloboschi:
Part of #4214.

## Why

`hindsight-api-slim/tests/` is ~500 flat files, each covering one mechanism. That catches mechanism bugs well. It catches **composition** bugs not at all — consolidation wiping facts, a delta refresh missing a backdated window, a transfer dropping mental-model evidence, a reprocess that is a silent no-op. Every one of those spans steps that no single-mechanism test crosses.

This adds the missing layer: blackbox system tests that drive a real `hindsight-api` process over HTTP through the published Python client. No engine object, no SQL, no internal imports — so the tests survive any refactor that keeps the API's promises, which is precisely what makes them worth having when most code is written by agents.

## How determinism works — with no production code changes

A stub server implements the OpenAI chat-completions and embeddings APIs plus a Cohere-compatible rerank endpoint. The server under test is pointed at it with ordinary environment variables:

| | provider | base URL |
|---|---|---|
| LLM | `HINDSIGHT_API_LLM_PROVIDER=openai` | `HINDSIGHT_API_LLM_BASE_URL` |
| Embeddings | `..._EMBEDDINGS_PROVIDER=openai` | `..._EMBEDDINGS_OPENAI_BASE_URL` |
| Reranker | `..._RERANKER_PROVIDER=siliconflow` | `..._RERANKER_SILICONFLOW_BASE_URL` |

Nothing in `hindsight_api` changes. Better, this exercises the **real provider transport** — the OpenAI client, `openai_compatible_llm.py`'s JSON repair, retry and rate-limit handling, structured-output plumbing. An in-process fake replaces all of that with nothing, and a meaningful share of this project's provider bugs live exactly there.

## Design decisions

**Rules match on a named pipeline step, not the request body.** Nothing on the wire identifies the caller — the `json_schema` name is the constant `"response"` on the strict path, and the soft `json_object` path sends no schema at all. So a step is identified by a short anchor phrase from its prompt, and `steps.py` owns the mapping. Tests write `on_step("extract_facts")` and never quote a prompt, so a prompt edit is a one-line change in one file instead of thirty red tests — the failure mode that kills wiremock-style suites.

**An unmatched call fails the test and prints the rule to paste in.** `MockLLM` synthesizes plausible facts from its input when it doesn't recognize a call, which is why tests built on it can pass without proving anything. Here the miss propagates all the way out to the client exception:

```
No stub rule matched this call. Add one:
    llm.on_step("extract_facts", contains=[...]).returns(...)
```

**Requests are validated strictly.** A fake you control drifts permissive, and the provider bugs that have actually hurt here — Bedrock rejecting `response_format`, Azure 400ing on `prompt_cache_key`, Vertex 404ing a region — are all *the provider refusing our request*. A permissive stub would have shipped every one of them green. The stub rejects unknown fields and enforces OpenAI's `strict: true` structured-output rules.

**Background work stays on.** Consolidation runs in the worker after retain returns. Rather than disabling it, tests wait for it through the operations API (`waiting.py`) — that asynchronous half is where the composition bugs this suite exists to catch actually live.

**The server runs from a scratch dir holding an empty `.env`.** `load_dotenv(find_dotenv(usecwd=True), override=True)` walks up from cwd and deliberately wins over the ambient environment (#2961). Started anywhere in the repo, the server picked up the developer's own `.env` and ignored every variable the fixture set — it booted on the dev port against the dev database. Stopping the upward walk at an empty file makes the passed environment the whole configuration.

## CI

`test-system` needs **no provider secrets**, so unlike every `test-api` job it also runs on fork PRs. And because embeddings and reranking are both stubbed, nothing loads sentence-transformers — the job skips torch and the HuggingFace model cache entirely (`uv sync --frozen --extra embedded-db` is the whole install).

## What's here

One story: retain a sentence, let the worker finish, recall with a question that never repeats the original wording, assert the fact comes back. Deliberately one — the point is to agree on the shape before writing thirty.

Ran 5× consecutively, green each time, ~6s per run.

## Conventions for what comes next

- One story per file, `test_NN_<story>.py`. The number is **reading order for a human, not execution order** — every test must pass alone.
- Assert only on client responses.
- Each test gets its own bank, deleted on teardown, with a session-start sweep of leftovers (a bank stranded mid-run gets picked up by the worker on the *next* startup, and its LLM call lands in a later test's setup).


koriyoshi2041 (changes requested):
The suite is not isolated from ambient proxy variables, so the local-only transport can fail before the server is even probed.

On macOS/Python 3.13.9 at this exact head, with `all_proxy=socks5://127.0.0.1:7897` plus the usual `no_proxy=localhost,127.0.0.1,::1`, I get:

```
uv run --directory hindsight-system-tests pytest tests/test_01_retain_and_recall.py -q
# ERROR in _wait_until_healthy: ImportError: Using SOCKS proxy, but the 'socksio' package is not installed
```

Removing the ambient proxy variables makes the same story pass (`1 passed in 19.88s`). The health probe uses `httpx.get(..., trust_env=True)`, `start_hindsight_server()` copies all non-`HINDSIGHT_API_` variables into the server process, and the published client created by the fixtures inherits the same environment. That leaves both the test harness's loopback traffic and the server-to-stub traffic dependent on unrelated workstation proxy configuration, despite the scratch `.env` isolation described here.

Could the harness explicitly remove the standard upper/lowercase proxy variables for this all-loopback topology (or construct every harness/client transport with `trust_env=False`)? I would add a regression that seeds `ALL_PROXY`/`all_proxy` with an unusable URL and proves the story still reaches both local servers. This keeps the new blackbox layer reproducible outside the clean GitHub runner.

Issue #4209, opened 2026-09-08
ConfigResolver deep-copies the immutable global config with asdict() on every request


nicoloboschi:
## Summary

`ConfigResolver` deep-converts the global config dataclass with `dataclasses.asdict()` on every resolution. `self._global_config` is built once in `__init__` and never mutated, but `asdict()` walks the whole tree again on each call — and config resolution is per-request, so this lands on the hot path of every recall.

## Measurement

Profiling a read-heavy benchmark (cProfile installed on the event-loop threads, free-threaded CPython 3.14.7, 2 loops, ~38 rps under the profiler):

| function | ncalls | tottime | cumtime |
|---|---|---|---|
| `dataclasses.asdict` | 3,236 | 0.01s | **2.12s** |
| `dataclasses._asdict_inner` | **1,446,492** | 1.08s | 1.90s |

That is ~3.4 `asdict()` calls per request, each fanning out to roughly 450 recursive `_asdict_inner` calls. In the same profile it was the second-largest consumer of the loop threads' own CPU, and those loop threads accounted for ~84% of the process's total CPU (measured separately from `/proc/<pid>/task/*/stat`, which is not subject to the profiler's overhead).

The recursion also shows up indirectly: `isinstance` (5.4M calls) and `getattr` (3.4M calls) were both in the top few entries, largely driven by this path and by enum flag handling.

## Call sites

All in `hindsight-api-slim/hindsight_api/config_resolver.py`:

- `_resolve_parent_config_dict:168` — `config_dict = asdict(self._global_config)`
- `get_bank_config:243`
- `_active_bank_overrides:413`
- also `:278` and `:333`

The first one is the clearest: `self._global_config` is immutable for the process's lifetime, so the conversion result is the same every time.

## Possible directions

Not proposing a specific patch — the correctness constraint is what matters and maintainers know it best:

1. Compute the global config's dict form **once** (it cannot change after `__init__`) and hand callers a copy. The subtlety is that callers mutate the returned dict, so a naive shared dict would leak edits between requests; a cached template plus a cheap copy needs to be measurably cheaper than `asdict()` to be worth it.
2. Avoid materializing the full dict when only a handful of fields are read — several callers immediately filter to `_configurable_fields`.
3. Cache resolved bank config per `(bank_id, tenant)` with explicit invalidation, if the resolution result is already stable between requests.

Happy to send a PR if there's a preferred direction.

## Environment

[...]

PR #4215, opened 2026-09-08
feat(profiling): env-configured CPU profile, reported to the logs

nicoloboschi:
Answers *what is burning the CPU?* for a process you cannot attach a debugger to.

```bash
HINDSIGHT_API_PROFILE='{"every": 60, "top": 20}'
```

Unset, nothing starts and there is no cost. Set, the API emits a CPU report to its normal log stream on that interval:

```
[profile] thread hindsight-loop-            1.39 cores
[profile] thread asyncio_                   0.38 cores
[profile] 1541 functions active in the last 40s
[profile] tottime=1.044 cumtime=0.000  ncalls=251814  <method 'run' of '_contextvars.Context' objects>
[profile] tottime=0.624 cumtime=1.348  ncalls=17316   <method 'count' of 'toktok._toktok.Tokenizer' objects>
```

One log record per row tagged `[profile]`, so `grep '\[profile\]'` reconstructs the table, and the existing `JsonFormatter` wraps each line as structured JSON.

## Why the log stream, and not a file or an endpoint

The case this exists for is a process dying without explanation. A file inside the container dies with the container unless a volume was mounted in advance; an endpoint needs a live process and a route to it, which is exactly what a crashing process does not offer. Container runtimes keep the previous container's stdout, so the last report before a crash is still readable afterwards. Each report is flushed as it is written, because a fatal signal takes buffered output with it.

## Three constraints the implementation had to meet

All three were found by building this against a running API, not by reading docs:

1. **The profiler is process-wide and single-instance.** Since 3.12 it is a global monitoring tool: `enable()` covers every thread whatever thread calls it, and a second concurrent profiler raises `ValueError: tool 2 is already in use`. Per-thread profilers are impossible, so this arms exactly one for the process — and `install()` is idempotent.
2. **Snapshots must use `getstats()`.** It reads the accumulated entries *without* stopping the profiler. The obvious alternative — snapshot through `pstats`, then `clear()` — disables the global tool, so every report after the first silently contains nothing. Reports are deltas between snapshots, and a test asserts a second window still has data.
3. **Sampling is not an alternative on a free-threaded build.** `sys._current_frames()` stops the world, so it catches threads parked at safe points, which are I/O waits. It reported event-loop threads idle in `selectors.select` while `/proc` showed those same threads at 50–65% of a core. Every report therefore carries per-thread CPU from `/proc` — which profiler overhead cannot distort — as the arbiter.

## Why not py-spy

py-spy is the better tool on a GIL build and the docs say so. It cannot read a `Py_GIL_DISABLED` process at all — it locates the running thread through the GIL, and such a build has none (`failed to get gil_thread_id`; `PYTHON_GIL=1` does not help, since it is the *build* it cannot parse). PEP 768 injection (`sys.remote_exec`) is also unavailable in that shape: injected code runs at the **main** thread's eval-loop safe point, and a multi-loop server parks the main thread. That left free-threaded deployments with no CPU profiler, which is what this closes.

## Validated by using it

Built against a live API and used to find three real regressions, each fixed and measured end to end at concurrency 8:

| change | p50 | p99 | throughput |
|---|---|---|---|
| baseline | 94.5 ms | 161.7 ms | 80.3/s |
| config resolver stopped deep-copying the global config (#4211) | 88.4 ms | 117.5 ms | 89.0/s |
| token counting moved off tiktoken | 78.4 ms | 107.5 ms | 99.3/s |
| output tokens counted in one batch | 76.7 ms | **101.3 ms** | **102.4/s** |

Each fix was found in a report, and each disappeared from the next report once it landed — `dataclasses._asdict_inner` (1.45M calls per window) after #4211, and the tokenizer after the other two. That round trip is the check that the tool reports what it claims to.

The `asdict` hot path in #4209 was found with an earlier version of this profiler, and disappeared from the table once #4211 landed — which is the check that the tool reports what it claims to.

## Tests

`tests/test_profiling.py` covers: off unless the env var is set; a malformed value raises rather than silently disabling; defaults; **a second window still reports data** (the regression that constraint 2 describes); the `/proc` table produces rows (skipped where `/proc` is absent); and `install()` is idempotent.


nicoloboschi:
Verified the crash-survival claim on a real pod rather than leaving it as an assertion.

Armed the profiler, let it emit, then sent `SIGSEGV` to the interpreter:

```
restarts before=0
profile rows in the live container: 146
killing the interpreter with SIGSEGV...
restarts after=1
rows recoverable from the DEAD container (kubectl logs --previous): 146
```

All 146 rows — including the last report written before the crash — are still readable afterwards. That is the property the log-stream design exists for: a file inside the container would have gone with it, and an endpoint would have had nothing left to answer.

PR #4238, opened 2026-09-08
fix(api): expose bank template import request body

### LINKED ISSUE #4232: import_bank_template is uncallable from every SDK: the handler reads a raw body, so the spec has no requestBody
## What happens

`POST /v1/default/banks/{bank_id}/import` takes a bank-template manifest as its request body, but the handler reads it off the raw request rather than declaring it:

`hindsight-api-slim/hindsight_api/api/http.py:7697`

```python
    async def api_import_bank_template(
        bank_id: str,
        request: Request,                       # <- raw request
        dry_run: bool = Query(default=False, ...),
        request_context: RequestContext = Depends(get_request_context),
    ):
        """Import a bank template manifest."""
        try:
            raw_body = await request.json()     # <- body read here, invisible to FastAPI
```

Because nothing declares the body, FastAPI emits **no `requestBody`** for this operation in the OpenAPI spec:

```python
>>> spec['paths']['/v1/default/banks/{bank_id}/import']['post'].keys()
# 'requestBody' absent; parameters are only bank_id, dry_run, authorization
```

## Why it matters

Every SDK is generated from that spec, so **every SDK's `import_bank_template` has no way to send the manifest**:

```python
>>> inspect.signature(BankTemplatesApi.import_bank_template)
(self, bank_id, dry_run=None, authorization=None, _request_timeout=None, ...)
#                ^ no manifest parameter at all
```

Calling it with the manifest as the second positional argument fails validation against `dry_run`:

```
ValidationError: 1 validation error for BankTemplatesApi.import_bank_template
2
  Input should be a valid boolean [type=bool_type, input_value=BankTemplateManifest(...)]
```

So the export half of the feature works through the SDK and the import half cannot be called at all — a caller has to drop to raw HTTP for it. `export_bank_template` returns a typed `BankTemplateManifest`, which makes the asymmetry especially easy to miss: you get a well-formed object with nowhere to send it.

The docs describe import as a normal part of the workflow ("Use dry_run=true to validate the manifest without applying changes"), so this reads as a bug rather than a deliberate omission.

## Suggested fix

Declare the body as a typed model on the handler — the manifest already has a Pydantic model (`BankTemplateManifest`) that `export_bank_template` returns:

```python
    async def api_import_bank_template(
        bank_id: str,
        manifest: BankTemplateManifest,
        dry_run: bool = Query(default=False, ...),
        request_context: RequestContext = Depends(get_request_context),
    ):
```

Then regenerate the spec and the clients. If the handler needs to accept shapes the model does not cover, declaring it as `dict` still produces a `requestBody` and a usable SDK method.

## Scope check

Nine other POST operations also lack a `requestBody` in the spec, but the rest are genuinely body-less action endpoints (`refresh`, `clear`, `retry`, `recover`, `regenerate`, `reprocess`, `health/llm`) where that is correct. Worth confirming `export_documents` while in here — it accepts options (`document_ids`, `include_observations`, `include_knowledge_base`) and appears in the same list.

## Notes

Found while writing blackbox system tests (#4214). `hindsight-system-tests/tests/test_54_templates_and_preview.py` covers export and the config round trip; the import assertions are skipped with a pointer here until the SDK can express the call.

0x90000:
## Summary

- declare the bank template manifest as a required JSON request body on `POST /v1/default/banks/{bank_id}/import`
- keep the handler's manual validation so existing template errors continue to return the established HTTP 400 format
- regenerate the checked-in Python, TypeScript, and Go clients, and update the static OpenAPI document used by the Rust client generator
- add a regression test that verifies the generated-client-facing request body schema

Fixes #4232

## Verification

- `python -m pytest tests/test_bank_templates.py -q` (5 passed, 34 skipped because local embedding models are not installed)
- `python -m pytest tests/test_bank_templates.py::TestImportValidation::test_import_openapi_declares_manifest_request_body -q -n 0` (passed)
- `ruff check hindsight_api/api/http.py tests/test_bank_templates.py`
- `ruff format --check hindsight_api/api/http.py tests/test_bank_templates.py`
- `python -m py_compile hindsight_api/api/http.py tests/test_bank_templates.py`
- `npm run build` in `hindsight-clients/typescript`

Go tests were not run because the available Go toolchain is 1.15.2 while this project requires Go 1.18+.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -7918,6 +7918,23 @@ def _register_routes(app: FastAPI):
         "Use dry_run=true to validate the manifest without applying changes.",
         operation_id="import_bank_template",
         tags=["Bank Templates"],
+        # Keep parsing and validation in the handler so malformed JSON and
+        # template errors retain the API's established 400 response format,
+        # while publishing the typed manifest schema for OpenAPI clients.
+        openapi_extra={
+            "requestBody": {
+                "required": True,
+                "content": {
+                    "application/json": {
+                        "schema": {
+                            "title": "Manifest",
+                            "description": "Bank template manifest",
+                            "$ref": "#/components/schemas/BankTemplateManifest",
+                        }
+                    }
+                },
+            }
+        },
     )
     @audited("import_bank_template", request_param=None)
     async def api_import_bank_template(
@@ -7928,7 +7945,7 @@ def _register_routes(app: FastAPI):
     ):
         """Import a bank template manifest."""
         try:
-            # Parse raw JSON and validate against the Pydantic model manually
+            # Parse and validate against the Pydantic model manually
             # so we can return clean error messages instead of raw 422s.
             raw_body = await request.json()
             from pydantic import ValidationError
--- a/hindsight-api-slim/tests/test_bank_templates.py
+++ b/hindsight-api-slim/tests/test_bank_templates.py
@@ -1,6 +1,7 @@
 """Integration tests for bank template import/export endpoints."""
 
 from datetime import datetime
+from types import SimpleNamespace
 
 import httpx
 import pytest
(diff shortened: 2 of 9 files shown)


koriyoshi2041 (commented):
One consumer-visible edge looks worth pinning before merge: replacing `await request.json()` with FastAPI `Body(...)` moves malformed JSON decoding outside the handler. On a minimal FastAPI/ASGI matched control, the old path returns the handler-established 400 for `{"bank":`, while the new `Any = Body(...)` path is intercepted as FastAPI `422 json_invalid`. Structurally invalid but parseable manifests still keep the intended 400, so the existing tests do not expose this drift. Since the PR says existing template errors retain the established 400 format, could we either (a) preserve malformed-JSON 400 via a request-validation exception mapping scoped to this route, or (b) explicitly accept/document the 422 boundary and add a malformed-body regression? The OpenAPI body declaration itself and generated-client propagation look consistent.

koriyoshi2041 (approved):
The follow-up preserves the malformed-JSON 400 boundary and pins it directly. I reran both focused API tests on `7a9c1146` (2 passed), Ruff lint/format, and the TypeScript client build; all passed. I also checked the generated Go request now requires and serializes the manifest body. Go tests could not finish locally because proxy.golang.org timed out while downloading existing dependencies. Looks good from the API/client contract side.

nicoloboschi:
Thanks @0x90000 — the approach here is right, and I've carried your commit over as-is into #4247.

Two generated-file issues meant this branch would have failed `verify-generated-files`, which is why I opened a new PR rather than asking for a force-push:

- `hindsight-docs/static/openapi.json` was hand-edited. `openapi_extra` merges last, so a regenerated spec places `requestBody` *after* `responses` (and orders the schema keys `title`, `description`) — same content, different bytes, and that job compares bytes.
- `skills/hindsight-docs/references/openapi.json` is a copy the docs-skill generator makes from the spec, and still had no request body at all.

#4247 adds a second commit that is just the output of the four generators CI runs. Your authorship is preserved on the fix commit. Closing this as superseded.

PR #4247, opened 2026-09-09
fix(api): expose the bank template import request body

### LINKED ISSUE #4232: import_bank_template is uncallable from every SDK: the handler reads a raw body, so the spec has no requestBody

nicoloboschi:
`POST /v1/default/banks/{bank_id}/import` reads its manifest off the raw `Request`, so FastAPI
emits no `requestBody` for the operation and every generated SDK's `import_bank_template` has no
parameter to send the manifest in. Export returns a typed `BankTemplateManifest` with nowhere to
send it; import is only reachable by dropping to raw HTTP.

The handler keeps parsing and validating the body itself — declaring `manifest: BankTemplateManifest`
as a parameter would hand validation to FastAPI and turn the endpoint's established 400 responses
into 422s. So the schema is published via `openapi_extra` instead, pointing at the same
`BankTemplateManifest` the export side already returns, and the handler is unchanged.

All three generated clients now express the body: Python gains a required `bank_template_manifest`
argument and the `Content-Type` header, Go gains `BankTemplateManifest()` with a nil guard, and the
TypeScript `ImportBankTemplateData.body` goes from `never` to `BankTemplateManifest`.

Supersedes #4238 by @0x90000, whose commit is the first one here. That PR hand-edited
`hindsight-docs/static/openapi.json`, which does not match generator output: `openapi_extra` merges
last, so a regenerated spec places `requestBody` *after* `responses`. It also left
`skills/hindsight-docs/references/openapi.json` — a copy the docs-skill generator makes — without the
request body at all. Both fail the byte comparison in `verify-generated-files`. The second commit here
is the output of the four generators run in CI's order.

That regen also picks up one unrelated file: `skills/hindsight-docs/references/developer/extensions.md`
was left stale on `main` when the StaticKeysTenantExtension section was added to the docs page without
the skill being rebuilt. It has to land here because the check regenerates everything and diffs the
whole tree.

Fixes #4232

## Verification

- `tests/test_bank_templates.py::TestImportValidation` — 14 passed. `test_import_openapi_declares_manifest_request_body`
  is a real gate: deleting the `openapi_extra` block fails it with `KeyError: 'requestBody'`.
- Ran all four generators plus `lint.sh` a second time on the committed tree — no diff, which is exactly
  what `verify-generated-files` asserts.
- `check-openapi-compatibility` against `main`: no backwards-compatibility issues.

`export_documents` also takes options and has no `requestBody`; #4232 flags it as worth confirming.
Left out of this change deliberately — it wants its own issue.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -7841,6 +7841,23 @@ def _register_routes(app: FastAPI):
         "Use dry_run=true to validate the manifest without applying changes.",
         operation_id="import_bank_template",
         tags=["Bank Templates"],
+        # Keep parsing and validation in the handler so malformed JSON and
+        # template errors retain the API's established 400 response format,
+        # while publishing the typed manifest schema for OpenAPI clients.
+        openapi_extra={
+            "requestBody": {
+                "required": True,
+                "content": {
+                    "application/json": {
+                        "schema": {
+                            "title": "Manifest",
+                            "description": "Bank template manifest",
+                            "$ref": "#/components/schemas/BankTemplateManifest",
+                        }
+                    }
+                },
+            }
+        },
     )
     @audited("import_bank_template", request_param=None)
     async def api_import_bank_template(
@@ -7851,7 +7868,7 @@ def _register_routes(app: FastAPI):
     ):
         """Import a bank template manifest."""
         try:
-            # Parse raw JSON and validate against the Pydantic model manually
+            # Parse and validate against the Pydantic model manually
             # so we can return clean error messages instead of raw 422s.
             raw_body = await request.json()
             from pydantic import ValidationError
--- a/hindsight-api-slim/tests/test_bank_templates.py
+++ b/hindsight-api-slim/tests/test_bank_templates.py
@@ -1,6 +1,7 @@
 """Integration tests for bank template import/export endpoints."""
 
 from datetime import datetime
+from types import SimpleNamespace
 
 import httpx
 import pytest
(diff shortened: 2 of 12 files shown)


0x90000:
I traced the failing `check-cli-coverage` job to the newly visible `BankTemplateManifest` properties. The CLI already supplies all four through the JSON file argument to `bank import-template`, so I added the corresponding field coverage reasons in #4249.

The same checker now reports: `OK: all 91 operations and 121 request params covered.`

PR #4249, opened 2026-09-09
fix(cli): account for bank template file input

0x90000:
## Summary

- account for the four `BankTemplateManifest` fields discovered by the CLI coverage check
- document that `version`, `bank`, `mental_models`, and `directives` are supplied together through the JSON manifest accepted by `bank import-template`

This fixes the `check-cli-coverage` failure on #4247 without changing CLI behavior.

## Verification

- `PYTHONUTF8=1 python hindsight-dev/hindsight_dev/cli_coverage_check.py`
  - `OK: all 91 operations and 121 request params covered.`


0x90000:
The equivalent coverage entries are now included directly in #4247 as commit `33a59b6a`, and its `check-cli-coverage` job passes. Closing this follow-up as incorporated.

PR #4250, opened 2026-09-09
fix(reflect): uniquify tool_call ids so strict APIs accept the turn

gwthm-in:
The reflect agent built its assistant message from the raw ids in `result.tool_calls` and emitted one tool_result per call keyed by the same raw id. When a provider or an OpenAI-compatible gateway returns two calls sharing an id -- or blanks one out -- the serialized turn carries two tool_use blocks with the same id and two tool_result blocks pointing at it. A strict Anthropic API rejects the whole request:

    400 invalid_request_error
    messages.4.content.5: each tool_use must have a single result.
    Found multiple `tool_result` blocks with id: toolu_...

This always fires on the first tool round-trip, so every reflect run against such an endpoint fails, taking refresh_mental_model with it. Providers that normalise the payload (the CLI-backed ones) hide the bug rather than avoid it.

Pick the wire ids once, positionally, before serialising, and reuse that list for both the assistant tool_calls and the tool_result messages so the two stay one-to-one. Only colliding or empty ids are rewritten, so a conforming provider keeps the ids it minted.

The result slots are still indexed by position, never by tool_call_id -- the existing reasoning for that is unchanged, and the wire ids are a parallel positional list rather than a map.


gwthm-in:
@nicoloboschi — tagging you as you have the most recent history on `engine/reflect/agent.py`; happy to redirect if someone else owns this area.

I searched open and closed issues for `each tool_use must have a single result`, `tool_use`, `invalid_request_error`, and duplicate tool_call ids and found nothing, so this PR is not linked to an existing issue. I also found no other PR touching this path — glad to close this if it duplicates work in flight.

Test evidence, since it is not in the description:

- New `test_duplicate_tool_call_ids_are_uniquified_on_the_wire` in `TestReflectAgentMocked` fails before the change with `AssertionError: ['dup', 'dup', '']` and passes after.
- `TestReflectAgentMocked`: 28 passed. Sibling reflect suites (`test_reflect_anthropic_protocol.py`, `test_reflect_tool_result_fields.py`, `test_reflect_tools.py`, `test_reflect_prompt_builder.py`): 48 passed.
- `ruff check` and `ruff format --check` clean on both changed files.

One adjacent case deliberately left out of scope: the `done_call` evidence-guardrail path (~L1077-1090) also emits a `tool_result` using `done_call.id` raw. It is a single call so it cannot produce duplicates, but an empty id there would still be malformed for a strict API. Happy to fold it in if you would prefer one change.

Reproduced on 0.9.2 against an Anthropic-compatible endpoint; it fires on the first tool round-trip, so `refresh_mental_model` never completes. I have not tested directly against api.anthropic.com.

Issue #4239, opened 2026-09-08
reflect: retrieval tools return ~2× their requested `max_tokens` (text-only counting vs. serialized payload), `search_mental_models` is uncapped, and `RECALL_MAX_TOKENS` is ignored on the agent path — forcing split synthesis


jwalkersf101:
**Version:** `ghcr.io/vectorize-io/hindsight:0.9.2` (image built 2026-09-07), self-hosted, reflect LLM = local vLLM (OpenAI-compatible), `HINDSIGHT_API_REFLECT_MAX_CONTEXT_TOKENS=48000`.

## Summary

On a ~27K-memory bank, one `budget=low` reflect accumulated ~53K context tokens after 3 tool iterations (5 tool calls) although the agent only *asked* for 16K tokens of retrieval, tripped the 48K context budget, and fell into the split map/reduce path. The two map calls + reduce were ~90% of the call's 222 s of LLM time (151 s wall). Three independent causes, all in the tool layer:

1. **`max_tokens` on `recall` / `search_observations` counts only each fact's `text`** (`memory_engine.py` recall docstring: "counts only 'text' field"), but the tool result handed to the LLM is the full pruned `model_dump()` plus chunks (`engine/reflect/tools.py`, `tool_recall` / `tool_search_observations`). Measured: `recall(max_tokens=4000)` → 27,489 and 25,775 chars (≈2.0–2.1× the cap at cl100k); `search_observations(max_tokens=8000)` → 45,860 chars (≈1.75×).
2. **`search_mental_models` has no token cap.** It returns every matched model's full `content` (`tools.py`, `tool_search_mental_models`). With 5 mental models totalling ~29.5K chars, each of the two calls returned ~31.5K chars — ~20K tokens, 38% of the accumulated context — and the second call returned the same 5 models as the first.
[...]

PR #4252, opened 2026-09-09
fix(reflect): honor resolved recall token defaults

r266-tech:
Reflect always passed hardcoded 2048/1000 token limits to its recall callback, bypassing the defaults resolved from environment, bank configuration, and mental-model triggers. Thread those resolved defaults through dispatch, schema descriptions, and trace summaries. A configured 512-token fact budget and zero chunk budget now reach retrieval unchanged when the model omits the corresponding limits; explicit model limits retain their existing normalization.

Refs #4239. This fixes the ignored-defaults portion only. Full serialized-payload accounting, mental-model result caps, and per-iteration context ceilings remain outside this change.

Validation:
- 121 deterministic tests passed across recall configuration, reflect tools/agent, Anthropic tool-result ordering, and tool schemas (real-LLM cases excluded).
- On unmodified main, the new dispatch regression has 7 failures and 2 passes; all 9 pass with this change.
- `./scripts/hooks/lint.sh` passed, including Python type checks and JavaScript lints. `git diff --check` passed.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -14160,9 +14160,9 @@ class MemoryEngine(MemoryEngineInterface):
         recall_fact_types = [ft for ft in (fact_types or ["world", "experience"]) if ft in ("world", "experience")]
         include_recall = bool(recall_fact_types)
 
-        # Defaults are bound at closure-definition time (re-evaluated on each
-        # reflect_async call), so per-bank/per-trigger overrides apply when the
-        # agent invokes recall without explicit token args.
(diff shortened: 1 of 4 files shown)


koriyoshi2041 (approved):
Verified at `594aa6b` that the resolved bank/trigger defaults survive the full engine → agent → recall-tool dispatch, including the configured `512` fact budget and `0` chunk budget. Explicit model limits still take the existing 1000-token floor, and the per-reflect schema copy avoids leaking one bank's advertised defaults into another concurrent call.

Local receipt: `test_recall_config.py` plus `test_reflect_tools.py` (42 passed), focused Ruff check/format, and `git diff --check` all pass. The scoped fix matches the ignored-defaults part of #4239; serialized-payload accounting and mental-model caps remain correctly out of scope.

PR #4256, opened 2026-09-09
fix(retain): prevent cross-bank chunk ID collisions

### LINKED ISSUE #4244: Chunk IDs can collide across bank and document boundaries
## Summary

On commit `2bf10435d5fbb4b7fee4414f0c31b679818bb714`, chunk IDs are constructed by
joining bank ID, document ID and chunk index with underscores, while the upsert
conflict target is only that generated ID. Distinct banks can therefore produce
the same chunk ID. The update path does not verify that the conflicting row
belongs to the submitted bank.

For example, bank/document pairs `a` plus `b_c` and `a_b` plus `c` both produce
`a_b_c_0`. A retain in the second bank can replace text in the first bank's
chunk row. we reproduced the overwrite with unmodified storage methods and a
rolled-back PostgreSQL fixture.

## Suggested fix

Use an unambiguous structured identifier or make bank ownership part of the
database uniqueness and conflict predicate. Add a regression test with the two
bank/document pairs above and assert that the first bank's row cannot be
modified by the second operation.

We currently prevents exposure with an explicit bank roster that rejects
every distinct pair where `b.startswith(a + "_")`, checked against configured
and existing banks before serving and again for runtime creation. That
downstream restriction is not a general upstream fix.

fcsguyxcw:
## Summary

- replace ambiguous underscore-joined chunk IDs with versioned, length-prefixed IDs
- centralize chunk ID construction/parsing and retain legacy ID parsing for existing data
- update store-owned and transfer paths to treat chunk IDs consistently
- add a public-API regression test for the `(a, b_c)` / `(a_b, c)` collision

## Testing

- `ruff check` on all changed Python files
- `ty check` on all changed production Python files
- `pytest -q -n 0 tests/test_reflect_tools.py` (18 passed)
- `python -m compileall -q hindsight_api`

The new PostgreSQL-backed regression test was collected successfully but could not run to completion on this Windows host because local pg0 loopback connections timed out after the server reported ready.

Fixes #4244

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/chunk_id.py
+++ b/hindsight-api-slim/hindsight_api/engine/chunk_id.py
@@ -1,0 +1,50 @@
+"""Construction and parsing for self-describing chunk identifiers."""
+
+from dataclasses import dataclass
+
+_VERSION_PREFIX = "v2:"
+
+
+@dataclass(frozen=True)
+class ChunkAddress:
+    bank_id: str
+    document_id: str
+    chunk_index: int
+
+
+def build_chunk_id(bank_id: str, document_id: str, chunk_index: int) -> str:
+    """Build an unambiguous identifier for a chunk within a bank and document."""
+    return f"{_VERSION_PREFIX}{len(bank_id)}:{len(document_id)}:{bank_id}{document_id}_{chunk_index}"
+
+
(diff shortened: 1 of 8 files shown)


nicoloboschi:
fixed by #4257

PR #4259, opened 2026-09-09
perf(db): skip asyncpg's release-time reset

nicoloboschi:
## What

asyncpg sends its reset query — `SELECT pg_advisory_unlock_all(); CLOSE ALL; UNLISTEN *; RESET ALL;` — every time a connection goes back to the pool. That is a full Postgres round trip on every request, and behind a transaction-mode pooler a server-side transaction of its own.

This drops it. No new config: the lever is a `Connection` subclass whose `get_reset_query()` returns `""`. (Passing `reset=` to `create_pool` does *not* replace the default — it runs `_con._reset()` and *then* the callback, so that knob only adds work.)

## Why it is safe here

Not a general asyncpg recommendation — it is safe for *this* codebase, so the invariants are asserted in `tests/test_db_pool_reset.py` rather than left to convention. Each test maps to one clause of the reset query:

- **`UNLISTEN *`** — nothing in the codebase issues `LISTEN`/`NOTIFY`.
- **`pg_advisory_unlock_all()`** — advisory locks are banned by project rule; the only occurrence is the grandfathered one in `migrations.py`, which runs on its own connection.
- **`RESET ALL`** — every session variable written on a pooled connection is either a `SET LOCAL` (transaction-scoped, so the reset was redundant) or one of the session GUCs the pool sets on purpose and wants to keep.

The guard walks the module tree and fails on a new violation; I verified it fails by injecting one of each.

`Connection._reset()` still runs, so a transaction a request left open is still rolled back on release, and listener callbacks are still dropped. Only the SQL statement is skipped.

## The one thing that does change

**The pool's session setup can now only ADD a GUC.** It *stops sending* `hnsw.iterative_scan` (and `statement_timeout`, `max_parallel_workers_per_gather`) when the corresponding setting is off, and previously `RESET ALL` is what cleared it. Now a connection opened while the setup was sending one keeps it until that connection is replaced.

Production never sees this: those settings come from process-static config, so every connection in a pool agrees and the set never changes within a pool's lifetime. A *test* that rewrites config mid-process does see it — `test_ann_iterative_scan` did, and now calls `pool.expire_connections()` to ask for connections opened after the rewrite, which is what it actually meant.

## Measured

A "single lookup" endpoint (`GET` one mental model) costs ~5 Postgres round trips. Measured by diffing `pg_stat_database.xact_commit` over 300 sequential requests with an idle baseline subtracted and no other load running:

| topology | baseline | this change | + `DB_SESSION_SETUP_ON_ACQUIRE=false` |
|---|---|---|---|
| direct to Postgres | 4.93 | 4.01 | 3.00 (**−39%**) |
| behind pgbouncer (txn mode) | 4.67 | — | 2.88 (**−38%**) |

3 repeats per arm, alternating, non-overlapping ranges. Re-measured through the shipped code path on an image built from this branch, isolating the reset (setup on in both arms), behind pgbouncer: **4.34 / 4.33 / 4.84 → 3.62 / 3.68 / 3.56**, 3/3, correctness OK on every arm.

Throughput, direct to Postgres at c=256 with both knobs off, 6 alternating pairs (median 2025 → 2492 rps, every pair favouring the change): **+26% rps, −20% p50, −18% p99**.

**Behind pgbouncer the throughput effect is unresolved and this PR does not claim one** — that fixture swings 3–10× run to run in *both* arms, which swamps a 25% effect. The round-trip counter resolves it and says the saving is topology-independent; how much throughput it buys through a pooler is not something I could measure. Expect much less on `recall`, which is CPU-bound (~23 ms/req) and spends proportionally little on round trips.

## Notes

- `HINDSIGHT_API_DB_SESSION_SETUP_ON_ACQUIRE` only exists because `RESET ALL` wiped the session GUCs. It is now redundant on a *direct* connection — but behind a transaction-mode pooler an acquire may still land on a server connection that never saw them, so its default is unchanged and its docs now say when it is redundant.
- asyncpg floor moves 0.29 → 0.30: `get_reset_query()` landed in 0.30 and on 0.29 the override would be silently ignored, making this a no-op. The lock already resolved 0.30.0, and 3.11–3.14 all resolve (0.31.0).


nicoloboschi:
Measured this on alpha today, on top of a set of other recall-path CPU fixes (vectorize-io/hindsight#4299). 1 replica, Guaranteed 2 vCPU, `WORKERS=2` matched to the limit, banks warm.

| | without #4259 | with #4259 |
| --- | --- | --- |
| 180 rps | PASS p99 620 ms, 1849m | **PASS p99 550 ms, 1753m** |
| ceiling | 175 rps | **182 rps** |
| CPU per request | 10.5 ms | **9.8 ms (−7%)** |

Independently corroborated by a py-spy profile before the change: `reset ← release` at 1.25% plus `release` at 0.66% of busy CPU, i.e. ~1.9% spent purely handing connections back.

## On the safety condition

Worth stating explicitly since it gates this PR: asyncpg's reset query *begins* with `SELECT pg_advisory_unlock_all()`, so skipping it makes any **session-scoped** advisory lock taken on a pooled connection leak permanently and poison the pool. I audited every site in the cloud extensions:

| site | kind | status |
| --- | --- | --- |
| tenant provision lease (`pg_try_advisory_lock`) | **session** | removed on the deployment side, replaced by a lease table |
| control migrations (`pg_advisory_lock`) | **session** | already released in an unconditional `finally` |
| everything else | `*_xact_lock` | transaction-scoped, released at commit |

So the one genuinely unsafe site is the tenant provision lease, and this PR should not land before that is gone.

Note the default keeps `db_session_setup_on_acquire=True`. My profile also caught `apply_session_settings ← _init_connection` at ~1.06% of busy CPU, which is the re-apply this PR makes redundant on a direct connection — worth a follow-up for deployments that are not behind a pooler.

Caveat: single runs per point, so treat the magnitude as approximate; the direction is consistent with the profile.

PR #4285, opened 2026-09-10
feat(python-client): add task-local retain suspension

roli-lpci:
## Summary

Add a task-local `suspend_retains()` context manager that suppresses `retain`, `retain_batch`, and `retain_files` across sync and async convenience methods while leaving recall and reflect untouched.

## Motivation

Evaluation runs and transcript replays need to read a populated bank without risking new writes. A shared client flag is unsafe when async tasks reuse the same client because one task can suppress another task's unrelated retain.

The context manager uses a per-client `ContextVar`, so suspension follows the calling execution context. Unrelated concurrent tasks are unaffected. Nested scopes and exception cleanup restore state through token reset.

## Requirements

- Suppress `retain`, `retain_batch`, and `retain_files` — both sync and async — for the duration of a `with client.suspend_retains():` block, sending no request while suspended.
- Leave `recall` and `reflect` fully functional inside the suspended scope.
- Scope suspension to the calling execution context (task-local via `ContextVar`), not the whole client, so concurrent async tasks sharing one `Hindsight` instance cannot suppress each other's retains.
- Support nested scopes and restore the prior state (not just "resumed") when an inner scope exits, via `ContextVar.set`/`reset` tokens rather than a plain boolean.
- Restore normal retain behavior when the scope exits, including via an exception, using `try`/`finally`.
- Return honest, typed empty responses while suspended instead of raising or faking a partial write: `RetainResponse(success=True, items_count=0)` for `retain`/`retain_batch`, `FileRetainResponse(operation_ids=[])` for `retain_files`.

## Use Cases

- Evaluation runs that recall/reflect against a previously populated bank and must not write new memories back into it as a side effect of the code under test.
- Replaying a transcript or fixture against a shared or populated bank (e.g. for debugging or regression testing) without polluting it with retains triggered during the replay.
- A read-only task sharing a long-lived client with other tasks doing normal writes — e.g. one asyncio task auditing a bank while another concurrently retains unrelated memories on the same `Hindsight` client; suspension in the read-only task must not affect the writing task. This is the scenario the review comment flagged against the original shared-flag design and is pinned by a dedicated deterministic interleaving test.

## Non-goals

- Not server-side or permission-enforced: this is a client-side convenience guard, not access control. It does not prevent writes at the API layer.
- Not a global or process-wide switch: it deliberately replaces the earlier shared mutable `retain_suspended` flag because that flag suspended retains for every task sharing the client, not just the caller's own scope.
- Does not change `recall` or `reflect` behavior in any way.

## Scope

The suspension intentionally covers only Hindsight convenience methods. Direct calls through the low-level memory and files accessors bypass it, as documented.

Suppressed calls send no request and return honest empty results: `items_count=0` or `operation_ids=[]`.

## Tests

- Added deterministic async interleaving coverage proving a suspended task does not suppress another task's retain on the same client.
- Added coverage for single, batch, and file retains, recall passthrough, nesting, exception cleanup, resumption, and the default behavior.
- Focused suspension suite: 9 passed.
- Sync and async parity plus suspension suites: 26 passed.
- Ruff and `git diff --check`: clean.

## Disclosure

This contribution was implemented and maintained by a Hermes Labs autonomous agent acting on a task prompted and authorized by Roli Bosch.

### DIFF
--- a/hindsight-clients/python/hindsight_client/hindsight_client.py
+++ b/hindsight-clients/python/hindsight_client/hindsight_client.py
@@ -8,6 +8,9 @@ easy-to-use interface on top of the auto-generated OpenAPI client.
 import asyncio
 import json
 import warnings
+from collections.abc import Iterator
+from contextlib import contextmanager
+from contextvars import ContextVar
 from datetime import datetime
 from importlib import metadata
 from pathlib import Path
@@ -218,7 +221,7 @@ class Hindsight:
         self._timeout = timeout
         self._base_url = base_url.rstrip("/")
         self._api_key = api_key
-        self._retain_suspended = False
+        self._retain_suspended: ContextVar[bool] = ContextVar("retain_suspended", default=False)
         if api_key:
             self._api_client.set_default_header("Authorization", f"Bearer {api_key}")
         self._memory_api = memory_api.MemoryApi(self._api_client)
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (commented):
The focused tests pass at `fc9b07e8`, but the flag is shared mutable state on a client that also exposes async methods. That makes the advertised “session” boundary unsafe when a client is reused concurrently: task A can set `retain_suspended = True`, yield during recall/reflect, and task B's unrelated `retain()` will silently return success without sending a request. A finally block only restores the flag after the damage.

Could this be scoped to the calling task (for example a context manager backed by `ContextVar`), or made explicit per retain call / via a read-only client view? An interleaving async regression test would pin the boundary. The current six tests are all sequential, so they do not exercise the failure mode introduced by the global flag.

roli-lpci:
Thanks, you are right that the shared mutable flag allowed one task's read-only scope to suppress an unrelated concurrent retain.

I replaced the setter with `with client.suspend_retains():`, backed by a per-client `ContextVar` with token-based reset for task isolation, nesting, and exception-safe cleanup. I also added a deterministic interleaving regression where task A remains suspended while task B's retain on the same client reaches the API successfully. The focused suspension suite and the sync and async parity coverage pass.

Could you please take another look?

Disclosure: I am a Hermes Labs autonomous agent acting on a task prompted and authorized by Roli Bosch.

koriyoshi2041 (approved):
The `ContextVar` scope closes the cross-task suppression bug and token reset preserves nesting and exception cleanup. I checked exact head `922c81bc2`; the focused suspension suite passes 9/9, including the deterministic same-client interleaving regression, and diff hygiene is clean.

nicoloboschi (changes requested):
please elaborate the requirements and use cases of this feature

roli-lpci:
Updated the description with a Requirements, Use Cases, and Non-goals section, grounded in the current `ContextVar`-based implementation and the 9 tests in `test_retain_suspension.py` — no code changes. Could you take another look? PTAL @nicoloboschi

nicoloboschi (approved):

PR #4286, opened 2026-09-10
fix(oauth): do not fail the store lock when the store cannot be written

LoneExile:
## Problem

`oauth_store_lock` takes its cross-process lock by creating `<store>.lock` next to the credential store. When the store is not writable that create raises (`EROFS`), and because the lock wraps the **entire** refresh the whole path dies with it:

```
Codex token refresh attempt failed: OSError: [Errno 30] Read-only file system: '/etc/codex/auth.lock'
Codex auth error (HTTP 401): {"message": "Encountered invalidated oauth token for user, failing request", "code": "token_revoked"}
```

A read-only store is not an exotic setup — it is the normal shape on Kubernetes. A Secret volume is mounted read-only regardless of what `volumeMount.readOnly` says, so any store fed by an external secret manager (ESO, a sidecar, a ConfigMap projection) hits `EROFS` here. Reproduced against a live deployment of the Codex provider with `auth.json` supplied by External Secrets Operator from OpenBao.

The consequence is worse than a missing lock. The guarded body contains the read that lets a credential *published by another writer* reach this process, so the failure did not just skip the lock — it disabled the read path, and the provider answered every call with the token it booted with until the process restarted. In a deployment where an external writer owns the store, that means a hard outage of every LLM operation once the boot token is revoked.

## Fix

Treat an unwritable store like the existing no-`fcntl` branch: log at debug and proceed on the per-loop lock alone. That lock still serializes this interpreter's callers, which is what makes the fallback safe — the file lock was never the protection against a store owned by another writer, since such a writer does not take it.

Everything else keeps propagating:

```python
if e.errno not in _UNWRITABLE_STORE_ERRNOS or os.access(lock_path.parent, os.W_OK):
    raise   # _UNWRITABLE_STORE_ERRNOS = EROFS, EACCES, EPERM
```

The same two calls fail on a perfectly **writable** store — `ENOSPC` when the volume is full, `EMFILE`/`ENFILE` when descriptors run out — and those arrive exactly when the box is under pressure. Degrading there would silently drop the cross-process lock and admit two concurrent rotations of a rotating token, one of which is necessarily lost.

## Scope

`oauth_store_lock` is shared by three providers, so this covers all of them:

- Codex (`codex_auth.py`)
- Nous (`nous_auth.py`)
- xai-oauth (`xai_oauth_auth.py`)

## Tests

`tests/test_oauth_store_lock.py`:

- Real permissions (skipped as root): degrade on a `r-x` store directory; propagate `EACCES` when only a pre-existing lock file is unwritable (writable directory, a peer may hold the lock).
- Injected (runs as root too): degrade for `EROFS`/`EACCES`/`EPERM` on an unwritable directory; propagate `EACCES` on a writable directory; propagate `ENOSPC`/`EMFILE`/`ENFILE`.
- `test_per_loop_lock_still_serialises_when_degraded` — 4 concurrent callers never overlap.
- `test_body_errors_are_not_chained_to_the_degrade_oserror` — the degraded path yields outside the `except`, so a refresh-body error isn't reported as raised while handling `EROFS`.

## Why errno alone isn't enough

`EACCES` is also raised by `open(lock_path, "a+")` when the directory is writable but a lock file owned by another uid denies this process (e.g. a container that ran as root, now non-root, over the same volume). The store is writable there and a peer may hold the lock, so the fallback also requires `os.access(lock_path.parent, os.W_OK)` to be false.

## Checks

```
uv run --frozen ruff check hindsight_api tests   # All checks passed!
uv run --frozen ruff format                      # clean
uv run --frozen ty check hindsight_api           # 22 diagnostics, all unresolved-import
                                                 # (pre-existing optional-ML-dep noise), none in the changed files
uv run --frozen pytest tests/test_oauth_store_lock.py tests/test_codex_oauth_refresh.py \
  tests/test_codex_home_env.py tests/test_codex_quota_reset_defer.py \
  tests/test_xai_oauth_llm.py tests/test_nous_provider.py   # 197 passed
```

Rebased onto current `main`, so this applies to `oauth_store_lock.py` as it exists today rather than to the `_codex_auth_lock` this was originally written against.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/providers/oauth_store_lock.py
+++ b/hindsight-api-slim/hindsight_api/engine/providers/oauth_store_lock.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import errno
 import logging
 import time
 from collections.abc import AsyncIterator
@@ -41,6 +42,11 @@ logger = logging.getLogger(__name__)
 
 # Interval between non-blocking flock attempts while another holder has the store.
 _POLL_INTERVAL_SECONDS = 0.05
+
+# Errors that mean the store cannot be written, so no lock file can live beside
+# it. The lock degrades to per-loop-only for these; every other OSError from
+# creating the lock file propagates.
+_UNWRITABLE_STORE_ERRNOS = frozenset({errno.EROFS, errno.EACCES, errno.EPERM})
 
 # One map of store path -> lock per event loop; LoopLocal prunes closed loops (a weak
 # map would not: a lock that has been waited on holds a reference to its loop).
@@ -69,8 +75,39 @@ async def oauth_store_lock(store: Path, *, timeout_seconds: float, label: str) -
             return
 
         lock_path = store.with_suffix(".lock")
-        lock_path.parent.mkdir(parents=True, exist_ok=True)
-        with open(lock_path, "a+") as lock_file:
+        try:
+            lock_path.parent.mkdir(parents=True, exist_ok=True)
+            lock_file = open(lock_path, "a+")
+        except OSError as e:
+            # Only a store that cannot be written gets here, and that is the
+            # normal shape on Kubernetes: a Secret volume is mounted read-only
+            # whatever `volumeMount.readOnly` says, so a store fed by an
+            # external secret manager (ESO, a sidecar, a ConfigMap projection)
+            # raises EROFS. Nothing can be written back to such a store, so the
+            # lock file does not exist and the caller's read path — which is how
+            # a credential published by another writer arrives — must still run.
+            # Failing here would instead take the whole refresh with it.
+            #
+            # Everything else must keep propagating. The same two calls fail on
+            # a perfectly WRITABLE store — ENOSPC when the volume is full,
+            # EMFILE/ENFILE when descriptors run out — and those arrive exactly
+            # when the box is under pressure. Degrading there would silently
+            # drop the cross-process lock and let two processes into the refresh
+            # body together, which is the race this lock exists to prevent. For
+            # these providers that means two concurrent rotations of a rotating
+            # token, where one of them is necessarily lost.
+            if e.errno not in _UNWRITABLE_STORE_ERRNOS:
+                raise
+            # Degrade to the per-loop lock alone, exactly as the no-`fcntl`
+            # branch above does. Note the file lock never protected against a
+            # store owned by another writer: such a writer does not take it.
+            logger.debug(
+                f"{label} store is not writable ({type(e).__name__}: {e}); refresh proceeds without a cross-process lock."
+            )
+            yield
+            return
+
+        with lock_file:
             deadline = time.monotonic() + max(1.0, timeout_seconds)
             while True:
                 try:
--- a/hindsight-api-slim/tests/test_oauth_store_lock.py
+++ b/hindsight-api-slim/tests/test_oauth_store_lock.py
@@ -1,0 +1,129 @@
+"""Tests for `oauth_store_lock` against a store that cannot hold a lock file.
+
+The lock is taken by creating ``<store>.lock`` next to the credential store. That
+write is impossible when the store is read-only, which is the normal shape on
+Kubernetes: a Secret volume is mounted read-only whatever ``volumeMount.readOnly``
+says, so a store fed by an external secret manager (ESO, a sidecar, a ConfigMap
+projection) raises ``EROFS`` when the lock file is created.
+
+The lock guards the whole refresh, including the read of the on-disk store that
+lets a credential published by another writer reach this process. A hard failure
+there did not just skip the lock — it disabled the read path, so the provider
+answered every call with the token it booted with until it restarted. Three
+providers share this lock (Codex, Nous, xai-oauth), so the failure mode and the
+fallback are shared too.
+
+Failures are injected into the lock path's own ``open``/``mkdir`` rather than
+produced with ``chmod``: a restrictive mode is a no-op for root, which CI
+containers routinely run as, so a ``chmod``-based test would pass whether or not
+the code handled the read-only case. Injection is deterministic for every user.
+"""
+
+from __future__ import annotations
+
+import builtins
+import errno
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+from hindsight_api.engine.providers.oauth_store_lock import oauth_store_lock
+
+
+@pytest.fixture
+def store(tmp_path: Path) -> Path:
+    """A credential store file, as the OAuth managers write it."""
+    path = tmp_path / "provider-home" / "auth.json"
+    path.parent.mkdir(parents=True)
+    path.write_text('{"tokens": {"access_token": "a", "refresh_token": "r"}}')
+    return path
+
+
+def _fail_lock_creation(store: Path, error_number: int):
+    """Make creating ``<store>.lock`` fail with ``error_number``.
+
+    Reads and writes of the store itself keep working, so the store behaves like
+    a projection that accepts nothing new beside the credential.
+    """
+    lock_path = store.with_suffix(".lock")
+    real_open = builtins.open
+
+    def fake_open(file, *args, **kwargs):
+        if Path(file) == lock_path:
+            raise OSError(error_number, f"injected {errno.errorcode.get(error_number)}")
+        return real_open(file, *args, **kwargs)
+
+    real_mkdir = Path.mkdir
+
+    def fake_mkdir(self, *args, **kwargs):
+        if self == lock_path.parent:
+            raise OSError(error_number, f"injected {errno.errorcode.get(error_number)}")
+        return real_mkdir(self, *args, **kwargs)
+
+    return patch("builtins.open", fake_open), patch.object(Path, "mkdir", fake_mkdir)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("error_number", [errno.EROFS, errno.EACCES, errno.EPERM])
+async def test_degrades_when_the_store_cannot_be_written(store: Path, error_number: int):
+    """A store that cannot be written degrades to per-loop-only instead of raising."""
+    p_open, p_mkdir = _fail_lock_creation(store, error_number)
+    entered = False
+    with p_open, p_mkdir:
+        async with oauth_store_lock(store, timeout_seconds=1.0, label="codex auth"):
+            entered = True
+
+    assert entered, f"{errno.errorcode[error_number]} must not prevent the guarded section"
+    assert not store.with_suffix(".lock").exists()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("error_number", [errno.ENOSPC, errno.EMFILE, errno.ENFILE])
+async def test_propagates_when_the_store_is_writable_but_the_create_fails(store: Path, error_number: int):
+    """A full disk or an exhausted descriptor table is not a read-only store.
+
+    Those failures hit a perfectly writable store, and they arrive when the box
+    is under pressure. Swallowing them would drop the cross-process lock and
+    admit two concurrent refreshes of a rotating token, where one is necessarily
+    lost.
+    """
+    p_open, p_mkdir = _fail_lock_creation(store, error_number)
+    with p_open, p_mkdir, pytest.raises(OSError) as excinfo:
+        async with oauth_store_lock(store, timeout_seconds=1.0, label="codex auth"):
+            pass
+
+    assert excinfo.value.errno == error_number
+
+
+@pytest.mark.asyncio
+async def test_per_loop_lock_still_serialises_when_degraded(store: Path):
+    """Degrading drops the FILE lock only — one loop's callers still queue.
+
+    That is what makes the fallback safe: the fallback's justification is that
+    the file lock was never the protection against an external writer (such a
+    writer does not take it), while same-process callers remain serialised.
+    """
+    p_open, p_mkdir = _fail_lock_creation(store, errno.EROFS)
+    concurrent = 0
+    max_concurrent = 0
+
+    async def hold() -> None:
+        nonlocal concurrent, max_concurrent
+        async with oauth_store_lock(store, timeout_seconds=1.0, label="codex auth"):
+            concurrent += 1
+            max_concurrent = max(max_concurrent, concurrent)
+            # Yield control so a second caller would overlap if unsynchronised.
+            for _ in range(3):
+                import asyncio
+
+                await asyncio.sleep(0)
+
+            concurrent -= 1
+
+    import asyncio
+
+    with p_open, p_mkdir:
+        await asyncio.gather(*(hold() for _ in range(4)))
+
+    assert max_concurrent == 1, f"{max_concurrent} callers entered the guarded section at once"


b4bz:
Thanks @LoneExile for identifying this failure and proposing the read-only adoption fix in #4286. Your report made the external-secret deployment issue clear.

I have a small collaborative refinement to offer for incorporation into your PR. A read-only consumer can still share credentials with a writer through another mount, so I suggest distinguishing “may adopt from disk” from “may refresh OAuth.” If creating `auth.lock` fails with `EROFS` or `EACCES`, the consumer would adopt only a different, known-fresh access token, and otherwise report that an external writer must publish credentials. It would never consume a rotating refresh token without the shared lock and durable publication.

The patch keeps current main's writable recovery intact, propagates unexpected lock errors, and adds deterministic errno-injection tests rather than relying on chmod under privileged CI. The proposed current-main branch has 42 new cases passing (35 fail on the unchanged base), and all 116 Codex tests pass. Full-project types are not clean in the minimal environment, but their diagnostic output matches the base exactly; the changed auth module passes its targeted type check.

I've published the implementation and tests as the linked draft #4311 so the patch is easy to inspect or incorporate into this PR. Would you be open to incorporating this approach? The draft is a coordination vehicle, not a request to replace or supersede your work; it prominently credits your original diagnosis and recovery proposal. I’m happy to consolidate it here rather than maintain competing PRs.

Implementation, review and testing were AI-assisted with Hermes Agent under human direction; this is not an independent human security review. Tests use synthetic credentials and injected permission failures, not a live OAuth or production canary for this revision.

hindsight-api-slim/hindsight_api/engine/providers/codex_auth.py
```diff
@@ -113,8 +113,29 @@ def _codex_auth_lock(auth_file: Path, timeout_seconds: float = _CODEX_AUTH_LOCK_
             return
 
         lock_path = auth_file.with_suffix(".lock")
-        lock_path.parent.mkdir(parents=True, exist_ok=True)
-        with open(lock_path, "a+") as lock_file:
+        try:
+            lock_path.parent.mkdir(parents=True, exist_ok=True)
+            lock_file = open(lock_path, "a+")
+        except OSError as e:
```
ebarkhordar: I read this at `1933d15`. The degrade-instead-of-die shape looks right to me. One thing about the width of the catch.

`except OSError` at line 119 covers the two calls above it, and the rationale it carries is specifically about a store that cannot be written. But `open(lock_path, "a+")` has failure modes on a perfectly writable store: `ENOSPC` when the volume is full, `EMFILE` or `ENFILE` when the process is out of descriptors. Those are all `OSError`, so on a writable multi-process deployment they now land in this branch, drop the cross-process lock, log at `debug`, and let two processes into the refresh body together. That is the race the lock was added for, and it becomes reachable exactly when the box is under pressure, which is also when a debug line is least likely to be read.

Narrowing to the errnos the comment actually describes would keep the new behaviour and let the rest surface:

```python
except OSError as e:
    if e.errno not in (errno.EROFS, errno.EACCES, errno.EPERM):
        raise
```

Relatedly, the last clause of the comment, that a read-only store has no second writer to race with, is the one thing your second test disproves: `test_readonly_store_still_adopts_a_rotated_disk_credential` is built around an external writer owning `auth.json`. The lock would not have helped against that writer, since it never takes it, which is a cleaner reason for the same conclusion.

I did not run the suite. This is from reading `codex_auth.py` and the new test file at that sha.

LoneExile: Good catch — taken, and thank you for reading it closely enough to find it.

You're right that the rationale named a store that cannot be written while the clause caught everything. `ENOSPC`, `EMFILE` and `ENFILE` all arrive on a *writable* store, and they arrive when the box is under pressure — which is also when a `debug` line is least likely to be read. Dropping the cross-process lock there admits exactly the race the lock exists for, so that is now a bug rather than a trade-off.

`f9211d6` narrows it:

```python
if e.errno not in _CODEX_AUTH_UNWRITABLE_ERRNOS:   # EROFS, EACCES, EPERM
    raise
```

Test coverage follows the split: parametrized over `EROFS`/`EACCES`/`EPERM` for the degrade path, and `ENOSPC`/`EMFILE`/`ENFILE` for propagation.

**Your note about the test method was the more useful half.** The previous tests used `chmod 0500` to make the store unwritable — which is a no-op for root, and CI containers usually run as root, so those tests would have passed whether or not the code handled the read-only case. That is a vacuous test, and it is also why I had avoided narrowing the catch: narrowing broke the `chmod` test, which should have told me the *test* was wrong rather than the narrowing. Failures are now injected into the lock path's own `open`/`mkdir`, so they are deterministic regardless of the user and exercise the real branch.

I also took your point on the comment. "A read-only store has no second writer to race with" was contradicted by my own second test, which is built around an external writer owning `auth.json`. The lock was never the protection against that writer, since it does not take it — a cleaner reason for the same conclusion, now written down as such.

Non-vacuity, measured against the unpatched base:
- original body (no fallback): **4 of 7 fail**
- the previous broad catch: **the 3 propagation cases fail specifically** — i.e. the defect you identified
- current: **7 pass**, and the wider Codex suite is 56 passing

On @b4bz's adopt-only proposal — I think it is a genuine improvement and it is the direction this should probably go, but it changes behaviour beyond the lock (a read-only consumer stops being able to refresh at all, and must fail loudly instead), so I would rather the maintainers make that call than quietly fold it into a bugfix. Happy to implement it here if they want it.

LoneExile:
Thanks — both for the diagnosis credit and for the careful framing of this as a refinement rather than a replacement.

On the technical question: **you are right, and I think adopt-only is where this should end up.** Unlocked rotation is genuinely unsafe, and I under-thought that. On a read-only store a refresh mints a rotating refresh token that cannot be persisted while invalidating the writer's copy upstream — pull that thread and you land exactly on the deployment that motivated the PR: the external refresher's stored refresh token goes `refresh_token_reused` and needs a re-login. Distinguishing "may adopt from disk" from "may refresh OAuth" is the correct boundary, and the existing `_codex_auth_lock` gating is the natural place to express it.

The reason I have not folded it in here: it is a behaviour change wider than the bug this PR fixes. A read-only consumer today can still recover by refreshing; under adopt-only it must fail loudly and wait for the writer to publish. That is the right trade for the shared-writer topology, but it is a design decision, and I would rather the maintainers make it explicitly than have it arrive smuggled inside a lock-fallback fix. If they want it, I am happy to implement it in this PR — your draft makes the shape clear, and I would credit it as the source.

What I did take from this round is the review on the width of the catch (`f9211d6`): only `EROFS`/`EACCES`/`EPERM` degrade now, and `ENOSPC`/`EMFILE`/`ENFILE` re-raise, since those hit a writable store and silently dropping the lock there reintroduces the race. Tests inject the failure rather than using `chmod`, which is a no-op for root.

One caveat I want to be explicit about, since you raised your own: I am treating your numbers as claims to verify rather than verifying yours, and likewise this revision is verified by me — 4 of 7 tests fail against the unpatched base, the 3 propagation cases fail against the previous broad catch, and the wider Codex suite is 56 passing. So the two PRs are each independently reproduced rather than one resting on the other.

LoneExile:
**Re-targeted onto current `main`.**

Upstream moved while this was open, and it changed the shape of the fix rather than just the line numbers:

- `31a8c52d` (#4318) replaced httpx with aiohttp across the providers and **extracted the lock out of `codex_auth.py` into a shared `oauth_store_lock.py`**, now used by Codex, Nous and xai-oauth.
- That module carries the identical unguarded `mkdir` + `open(<store>.lock)`, so the bug is now shared by three providers rather than one.

The branch is rebuilt on today's `main` and the change is applied to `oauth_store_lock.py` instead of `_codex_auth_lock` (which no longer exists). Net effect: the same fix, for three providers, and the diff is now 2 files (+168/−2).

Also kept from the review round: only `EROFS`/`EACCES`/`EPERM` degrade, `ENOSPC`/`EMFILE`/`ENFILE` re-raise, and the tests inject the failure instead of using `chmod` (a no-op for root).

The non-vacuity evidence is now three neuters, each isolating one property:
- unguarded body → 4 of 7 fail
- broad `except OSError` → the 3 propagation cases fail
- per-loop lock removed → the serialisation case fails

@b4bz — your adopt-only point still stands and applies to this shared lock as well; it stays a maintainer call and I'm happy to implement it here if they want it.

LoneExile:
@ebarkhordar — flagging a change that affects where your review landed, since editing the PR body does not notify anyone and your thread's anchor is now gone (it reads `line: null` alongside the original `codex_auth.py:119`).

**The code you reviewed no longer exists.** Upstream's `31a8c52d` (#4318, httpx → aiohttp) extracted `_codex_auth_lock` out of `codex_auth.py` into a shared `oauth_store_lock.py`, used by Codex, Nous and xai-oauth. I rebuilt the branch on today's `main` rather than resolving a rebase onto a deleted function, so the fix is now applied to that shared module.

**Both of your points came with it** — they were about the pattern, not the file:

- The narrow catch, exactly as you specified: `if e.errno not in _UNWRITABLE_STORE_ERRNOS: raise`, with `EROFS`/`EACCES`/`EPERM` degrading and `ENOSPC`/`EMFILE`/`ENFILE` propagating.
- Your correction of my rationale: the fallback no longer claims a read-only store "has no second writer to race with". It now says the file lock was never the protection against a store owned by another writer, since such a writer does not take it.
- The test method too: injection into the lock path's `open`/`mkdir`, not `chmod`, for the root reason you gave.

Evidence, each neuter isolating one property:
- unguarded body (what `main` has today) → 4 of 7 fail
- broad `except OSError` → the 3 propagation cases fail, i.e. the defect you identified
- per-loop lock removed → the serialisation case fails

Scope widened as a side effect: the bug is now shared by three providers rather than one, since the lock is.

Worth a fresh look at the new location if you have the time — the async shape is different from what you read (the file lock is now `flock` retried with `asyncio.sleep`, under a per-loop `asyncio.Lock`), and I would rather have it reviewed by someone who already found one real defect in it than not.

hindsight-api-slim/hindsight_api/engine/providers/codex_auth.py
ebarkhordar: Confirmed at `51a6f33`, by reading the new file rather than running it: `_UNWRITABLE_STORE_ERRNOS` is `{EROFS, EACCES, EPERM}` at line 49 and line 99 re-raises everything else. That is the narrowing, and moving it to `oauth_store_lock.py` means it covers Nous and xai-oauth too.

The `chmod 0500` finding is yours and it is the bigger one. I questioned the width of the catch; you worked out that the test which made the broad catch look necessary was passing for free under root. A test that cannot fail is worse than the branch it was guarding.

Agreed on leaving adopt-only to the maintainers. It changes what a read-only consumer can do, which is a separate decision from this one and should be visible as such.

hindsight-api-slim/hindsight_api/engine/providers/oauth_store_lock.py
```diff
@@ -69,8 +75,39 @@ async def oauth_store_lock(store: Path, *, timeout_seconds: float, label: str) -
...
+            #
+            # Everything else must keep propagating. The same two calls fail on
+            # a perfectly WRITABLE store — ENOSPC when the volume is full,
+            # EMFILE/ENFILE when descriptors run out — and those arrive exactly
+            # when the box is under pressure. Degrading there would silently
+            # drop the cross-process lock and let two processes into the refresh
+            # body together, which is the race this lock exists to prevent. For
+            # these providers that means two concurrent rotations of a rotating
+            # token, where one of them is necessarily lost.
+            if e.errno not in _UNWRITABLE_STORE_ERRNOS:
```
ebarkhordar: I ran this against 51a6f338 in a clean container. EACCES in this set degrades on a store that is writable.

EACCES is not only "the store cannot be written". It is also what `open(lock_path, "a+")` raises when the directory is writable and only a pre-existing lock file denies this uid, which is what you get when two uids share a credential directory (a container that ran as root and now runs non-root over the same volume). The store stays writable there: `_persist_auth_atomic` writes a tempfile into the parent and `os.replace`s it, so directory permission is what decides.

`/srv/store` 0775 uid 1001 group 2000, `auth.json` and `auth.lock` both 0644 uid 1001, process B uid 1002 in group 2000, process A holding LOCK_EX on `auth.lock`:

```
[1] atomic replace of the store (as the managers do)  OK -> store IS writable
    os.access(store.parent, os.W_OK)                  True
[2] open(lock_path, 'a+')                             PermissionError errno=EACCES
[3] oauth_store_lock at 51a6f338                      ENTERED THE GUARDED SECTION
    oauth_store_lock at bde55237f                     RAISED PermissionError errno=EACCES
```

B runs the refresh body while A holds the lock, the outcome the comment above reserves for ENOSPC. The errno cannot separate the two, since EACCES covers the unwritable directory and the foreign lock file alike. Asking about the store does:

```python
if e.errno not in _UNWRITABLE_STORE_ERRNOS or os.access(lock_path.parent, os.W_OK):
    raise
```

Measured: still degrades on a real read-only mount (EROFS, `os.access` False), propagates above.

That line does fail all three params of `test_degrades_when_the_store_cannot_be_written`, EROFS included, because the injected errno lands on a writable `tmp_path`. The injection pins the errno match rather than the condition, so testing it needs real permissions and likely a skipif for root.

I did not run the wider suite or try the no-fcntl path.

LoneExile:
@ebarkhordar — you are right, and I reproduced it before changing anything.

**The errno cannot express the condition.** On a store directory that is writable:

```
dir mode=0o755, lock mode=0o444
os.access(dir, W_OK) = True          <- directory IS writable
open(auth.lock, "a+") -> EACCES      <- the errno the rule degrades on
atomic replace of the store: OK      <- store IS writable
```

So `EACCES` covers both "nothing can be written here" and "a peer owns the lock file in a directory I can write". In the second the store is writable *and* someone holds the lock, so degrading runs a second refresh body concurrently — the race the lock exists for, and for a rotating token one rotation is necessarily lost. Taking your discriminator verbatim (`5794b78`):

```python
if e.errno not in _UNWRITABLE_STORE_ERRNOS or os.access(lock_path.parent, os.W_OK):
    raise
```

**And your test critique was the sharper half again.** You are right that the injected-errno tests pinned the *errno match* rather than the condition — the injected errno lands on a writable `tmp_path`, so it never exercised the state the guard actually asks about. They would have kept passing while the real condition drifted, which is the failure mode that matters here.

I ran the previous revision's tests against the new guard to check how much they were really covering, and it is worse than your three parameters: **4 failed** — `[30]` EROFS, `[13]` EACCES, `[1]` EPERM, plus `test_per_loop_lock_still_serialises_when_degraded`, which reached the same conclusion by the same injection. Every test that depended on "the lock could not be created" was really testing "the errno matched", so the whole degradation story hung on one injected number.

Restructured, since I think both halves are needed:

- **The condition, with real permissions** — a `chmod 0500` directory degrades; a writable directory holding a `0400` lock file propagates with `EACCES`. Both `skipif(os.geteuid() == 0)`, because root bypasses the bits, as you noted. Verified they actually run here rather than silently skipping.
- **Branch logic, injected** — errno-in-set + `os.access` False degrades; errno-in-set + `os.access` True propagates; `ENOSPC`/`EMFILE`/`ENFILE` propagate. These still run as root, so CI keeps coverage either way.

Non-vacuity, measured (each neuter hits a different property):
- the `os.access` discriminator removed (my previous revision): **2 failed** — the two writable-directory cases
- the errno narrowing removed: **5 failed**

Wider suite: 200 passed across codex / nous / xai-oauth / quota-reset / home-env. ruff clean; `ty` unchanged at the 22 pre-existing `unresolved-import` diagnostics, none in the changed files.

I did not vary the uid in the reproduction — the `setpriv` attempt failed in the pod I had, so what I measured was a writable directory with a non-writable lock file rather than two distinct uids. The mechanism is the same one you described, but it is a weaker reproduction than yours; if your two-uid container is easy to re-run against `5794b78`, that would close it properly.

hindsight-api-slim/hindsight_api/engine/providers/oauth_store_lock.py
LoneExile: Fixed in `5794b78` — took your discriminator verbatim:

```python
if e.errno not in _UNWRITABLE_STORE_ERRNOS or os.access(lock_path.parent, os.W_OK):
    raise
```

I reproduced the ambiguity before changing anything: with `dir 0755` and `lock 0444`, `os.access(dir, W_OK)` is `True` while `open(lock, "a+")` raises `EACCES` — so the previous revision degraded into a concurrent refresh body, exactly as you said.

Verified in both directions on a live pod: the real read-only Secret mount (`os.access` False) still enters and reads the credential, and a writable directory holding a `0400` lock file now propagates `EACCES` instead of racing.

Your test critique was the sharper half. The injected-errno tests pinned the *errno match* rather than the condition, and they fail on all three parameters under the new guard for exactly the reason you gave. The condition is now covered with real permissions (`chmod 0500` degrades; writable dir + `0400` lock propagates, both `skipif` for root since root bypasses the bits), with injected branch tests retained so CI keeps coverage when it runs as root.

Full detail in my comment above. Still open on my side: my reproduction used a non-writable lock file rather than two distinct uids (`setpriv` failed in the pod I had), so your two-uid container would be the stronger confirmation against `5794b78` if it is easy to re-run.

ebarkhordar: Re-ran the two-uid container against `5794b78`. It propagates, and it matches the pre-PR behaviour rather than merely differing from the last revision.

Same scenario, real permissions, no injected errors: `/srv/store` 0775 uid1001:gid2000, `auth.json` and `auth.lock` both 0644 uid1001, process B as uid1002 in group 2000, process A holding `LOCK_EX` on `auth.lock`.

```
docker run --rm -m 1g -v ...:/new:ro -v ...:/src:ro -v ...:/base:ro python:3.12-slim sh /setup2.sh

  os.access(store.parent, os.W_OK)   True
  atomic replace of the store        OK -> store IS writable
  open(lock_path, "a+")              PermissionError errno=EACCES

  oauth_store_lock 5794b78           RAISED PermissionError errno=EACCES   (did not enter)
  oauth_store_lock 51a6f338          ENTERED THE GUARDED SECTION
  oauth_store_lock bde55237          RAISED PermissionError errno=EACCES
```

I also checked the motivating case on a real read-only bind mount, because that is the half your permission tests cannot reach as root. EROFS, `os.access` False, degrades and still reads the credential, as uid 1002 and as uid 0 alike. `os.access` consults the mount flag and not only the mode bits, so root does not slip past the new conjunct.

On the skipif, no job in `.github/workflows/test.yml` sets `container:`, so every job runs on `ubuntu-latest` as `runner` rather than root and both real-permission tests do execute in CI. On `test_oauth_store_lock.py` at `5794b78`: 10 passed 0 skipped as uid 1002, 8 passed 2 skipped as uid 0. Against the previous revision as uid 1002 it is 2 failed, which matches your number, and one of them is `test_propagates_when_only_the_lock_file_is_unwritable`. The real-permission half is load-bearing.

I did not run the wider suite and did not exercise the no-fcntl branch.

PR #4287, opened 2026-09-10
fix(coding-agents): admit one codebase survey per bank

### LINKED ISSUE #4255: coding-agents: concurrent SessionStart hooks race the cold-bank check, spawning duplicate codebase surveys
**Package:** `@vectorize-io/hindsight-coding-agents` 0.5.2
**Server:** self-hosted API `0.9.2`, harness `claude-code`

## Problem

The one-time codebase-survey trigger has no locking, unlike the bank config/git-ingest
path it sits right next to. Concurrent `SessionStart` hook invocations for the same repo
each independently see the bank as "cold" and each spawn their own `claude -p` survey
subprocess — duplicated N times for N concurrently-starting sessions.

## Real-world repro

Restarted VS Code (Claude Code extension) with an existing project open whose repo had
not yet been ingested into Hindsight (a cold bank). VS Code's Claude Code extension
auto-spawns a fresh empty session on restart — in this case it spawned **six**, all
firing `SessionStart` within the same window. All six independently triggered a
codebase-survey spawn; six concurrent `claude -p` processes were observed running the
survey prompt against the same repo before being manually killed.

## Root cause

`deepen.js`'s own bank-level lock works correctly and is not the problem — 5 of the 6
concurrent `deepen` runs correctly no-op'd:

```
deepen -> http://localhost:8888 bank=coding-agent::cadmaster harness=claude-code
[bank] coding-agent::cadmaster already carries the coding structure — nothing to apply
[gitlog] ingesting last 212 commit messages for cadmaster as ONE document …
deepen: another run holds the lock for coding-agent::cadmaster — nothing to do   (×5)
```

via a real PID-checked lockfile in `deepen.js`:

```js
var LOCK = join8(LOCK_DIR, `deepen-${encodeURIComponent(FINAL_BANK ?? "")}.lock`);
function acquireLock() {
  try {
    const held = JSON.parse(readFileSync4(LOCK, "utf8"));
    if (held.ts && Date.now() - held.ts < LOCK_STALE_MS) {
      let holderAlive = false;
      if (held.pid) { try { process.kill(held.pid, 0); holderAlive = true; } catch {} }
      if (holderAlive) return false;
    }
  } catch {}
  try {
    mkdirSync2(LOCK_DIR, { recursive: true });
    writeFileSync(LOCK, JSON.stringify({ pid: process.pid, ts: Date.now() }));
    return true;
  } catch { return false; }
}
```

But the codebase-survey trigger is a *separate* code path in `claude-sessionstart-hook.js`
(`buildSessionStartContext` → `startCodebaseSurvey`), not covered by that lock at all. It
decides "cold" by querying live document count, with no guard against another
in-flight SessionStart having made the same decision moments earlier:

```js
docIds = await client.listDocumentIds("source:git", "all_strict");
cold = docIds !== void 0 ? docIds.size === 0 : void 0;
...
if (docIds.size === 0) {
  if (cfg.codebaseSurvey !== false) {
    startSurvey(cwd, { harness, model: cfg.surveyModel, budgetUsd: cfg.surveyBudgetUsd });
    const sha = resolveHeadSha(cwd);
    if (sha) recordSurveyBaseline(sha);
  }
  ...
}
```

`startCodebaseSurvey` itself also has no dedup — it unconditionally spawns:

```js
function startCodebaseSurvey(repoDir, opts = {}) {
  ...
  const child = spawnFn(plan.bin, plan.args, { cwd: repoDir, detached: true, stdio: "ignore", ... });
  child.unref();
}
```

`recordSurveyBaseline(sha)` writes a `survey-baseline:<sha>` marker document, but only
*after* the spawn, and only on the process that got there — it isn't checked as a
pre-spawn guard in the cold-bank branch (it's only consulted later, in the
already-warm "refresh after N commits" branch), so it can't prevent siblings racing
the same cold-check in parallel.

## Impact

- N concurrently-starting sessions in the same not-yet-ingested repo → N concurrent
  survey subprocesses, each at `--max-budget-usd 2` — up to `2×N` wasted spend for one
  cold start.
- N racing calls to `hindsight_ingest_document` for the same 4 survey document titles
  (component map / core concepts / conventions / tech stack), landing in whatever order
  they finish.
- Triggerable without any unusual user action: restarting an editor extension that
  opens multiple session tabs against a repo Hindsight hasn't seen yet is enough.

## Suggested fix

Gate `startCodebaseSurvey` behind the same kind of PID-checked lock `deepen.js` already
uses for bank writes — keyed by bank id, e.g. `survey-<bank>.lock` next to the existing
`deepen-<bank>.lock` — so only the first of N concurrent cold SessionStarts actually
spawns the survey and the rest see it's already in flight and skip.

r266-tech:
Concurrent SessionStart hooks can each launch a paid codebase survey while the same bank is still cold. This gates survey admission with an atomic populated-directory lock keyed by the resolved API, credential and bank. Ownership transfers to the detached survey PID before the hook returns; dead owners can be reclaimed, and an old child cannot release a newer generation. Cold and refresh baselines are recorded only after a successful launch.

Fixes #4255.

Validation: all 967 coding-agents tests pass (70 files), TypeScript checking and package build pass, and `./scripts/hooks/lint.sh` passes. Regression coverage includes six independent hook processes, exclusion after their parents exit, concurrent dead-owner recovery, destination isolation and failed-launch baseline handling. These are deterministic process tests; no live model inference was used.

The lock coordinates processes sharing the local temporary directory; it is not distributed coordination across machines. AI-assisted implementation and self-review.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/session-start.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/session-start.test.ts
@@ -14,7 +14,7 @@ describe("buildSessionStartContext", () => {
   it("cold git repo + autoSeed on -> seeds + surveys, note in systemMessage (user-visible) + roster in additionalContext (model)", async () => {
     const client = { listDocumentIds: async () => new Set<string>(), listPages: listPagesOk };
     const startSeed = vi.fn();
-    const startSurvey = vi.fn();
+    const startSurvey = vi.fn().mockResolvedValue(true);
     const out = await buildSessionStartContext({
       cwd: "/repo/dir",
       bankId: "bank-1",
@@ -58,7 +58,7 @@ describe("buildSessionStartContext", () => {
       harness: "codex",
       hasGit: () => true,
       startSeed,
-      startSurvey: vi.fn(),
+      startSurvey: vi.fn().mockResolvedValue(true),
     });
     expect(startSeed).toHaveBeenCalledWith("/repo/dir", { limit: 300, harness: "codex" });
   });
@@ -66,7 +66,7 @@ describe("buildSessionStartContext", () => {
   it("cold git repo + codebaseSurvey:false -> starts the seed but NOT the survey", async () => {
     const client = { listDocumentIds: async () => new Set<string>(), listPages: listPagesOk };
     const startSeed = vi.fn();
-    const startSurvey = vi.fn();
+    const startSurvey = vi.fn().mockResolvedValue(true);
     const out = await buildSessionStartContext({
       cwd: "/repo/dir",
       bankId: "bank-1",
@@ -152,7 +152,7 @@ describe("buildSessionStartContext", () => {
 
   it("warm bank (non-empty doc set) -> deepen engine fires, but no survey/note", async () => {
     const startSeed = vi.fn();
-    const startSurvey = vi.fn();
+    const startSurvey = vi.fn().mockResolvedValue(true);
     const client = { listDocumentIds: async () => new Set(["git:abc"]), listPages: listPagesOk };
     const out = await buildSessionStartContext({
       cwd: "/repo/dir",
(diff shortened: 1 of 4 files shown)


nicoloboschi:
fixed by #4303

PR #4288, opened 2026-09-10
fix(api): reserve listeners before application initialization

### LINKED ISSUE #4281: standalone hindsight-api does not fail fast on port conflict — full init (embedded PG + models + migration) runs before [Errno 98] bind failure
### Bug Description

The standalone `hindsight-api` CLI does not fail fast on a port conflict. When the target port is already occupied, it performs the **entire expensive startup** — starts the embedded PostgreSQL, loads the embedding and reranker models, verifies the LLM connection, and runs database migrations — and only afterwards does uvicorn attempt to bind the socket and fail with `[Errno 98] address already in use`.

Under a process supervisor with `Restart=always` (systemd, docker restart policy, etc.) this becomes a **restart storm**: each ~11s cycle does a full init + graceful shutdown, then restarts ~5s later, wasting CPU/memory and — because the embedded PostgreSQL uses `port=auto` — starting a fresh PG instance on an incrementing port each cycle, which can leave orphan postgres processes behind.

### Environment

- hindsight-all 0.9.2 (`hindsight-api` standalone entry point)
- Linux, systemd user service `Restart=always`, `ExecStart=hindsight-api --host 127.0.0.1 --port 9177`
- The port is held by another Hindsight daemon (e.g. one spawned via `HindsightEmbedded`)

### Steps to Reproduce

1. Start any listener on `127.0.0.1:9177` (e.g. a second Hindsight daemon).
2. Run `hindsight-api --host 127.0.0.1 --port 9177`.

### Expected Behavior

The CLI should check port availability at the very start (before instantiating `MemoryEngine`, starting the embedded PG, loading models, or running migrations) and exit immediately with a clear error when the port is unavailable.

### Actual Behavior

Full initialization runs first, then the bind fails:

```
... Starting Hindsight API...
INFO:     Waiting for application startup.
... Embeddings: initializing local provider with model BAAI/bge-small-en-v1.5
... Reranker: initializing local provider with model cross-encoder/ms-marco-MiniLM-L-6-v2
... Starting embedded PostgreSQL (name=hindsight, port=auto)...
... PostgreSQL started: postgresql://hindsight:***@127.0.0.1:5433/hindsight
... Running database migrations...
... Database migrations completed successfully for schema 'public'
... Memory system initialized (pool and task backend started)
INFO:     Application startup complete.
ERROR:    [Errno 98] error while attempting to bind on address ('127.0.0.1', 9177): address already in use
INFO:     Waiting for application shutdown.
... Stopping pg0...
```

### Root cause

`hindsight_api/server.py` instantiates `MemoryEngine(...)` and `create_app(...)` at module import time; `main.py` only calls `uvicorn.run()` afterwards, and the socket bind happens after the app (and its lifespan startup) has fully initialized. There is no port-availability pre-flight check on the standalone path.

### Contrast with embedded mode

The embedded path (`hindsight_embed/daemon_embed_manager.py`) already guards against this: `_clear_port()` / `_is_port_in_use()` run *before* spawning a daemon (handling healthy-reuse / stale-reclaim / foreign-refuse — see #843, #3520, #3100). The standalone `hindsight-api` CLI has no equivalent pre-flight check.

### Suggested fix

Perform the same loopback port check (`_is_port_in_use`) at the top of the standalone CLI startup (before `MemoryEngine` / migrations), and exit with a clear, actionable error when the port is occupied.

r266-tech:
An occupied standalone API port currently fails only after application startup, which can repeatedly initialize models and embedded PostgreSQL under a restart policy. Reserve the effective listeners after configuration and daemon address resolution, before entrypoint lazy imports and engine construction, and retain them through serving.

The reservation uses asyncio's existing hostname/IPv4/IPv6 resolution without accepting connections, then transfers the sockets to Uvicorn's Server, reload supervisor or multiprocess supervisor. The CLI closes the sockets if initialization or serving fails. It does not probe and release the port or terminate the existing listener.

Fixes #4281.

Validation: 36 focused entrypoint/daemon tests pass, including real occupied listeners for normal/workers/reload/daemon entry, IPv4/IPv6/hostname reservation and release, cleanup on initialization failure, a real HTTP request over the retained socket, supervisor dispatch and Uvicorn startup-failure exit status. `./scripts/hooks/lint.sh` passes (including API type checking). A standalone regression reached the lazy MemoryEngine import on the original source and exits at bind on the patch. No live model/DB startup or live multiprocess/reload deployment is claimed.

Configuration parsing/profiling still occurs before reservation and may import optional model dependencies in current main; this does not claim to remove that import cost. The change prevents engine/application initialization on a port conflict. AI-assisted implementation and self-review.

### DIFF
--- a/hindsight-api-slim/hindsight_api/main.py
+++ b/hindsight-api-slim/hindsight_api/main.py
@@ -14,8 +14,10 @@ import argparse
 import asyncio
 import atexit
 import dataclasses
+import logging
 import os
 import signal
+import socket
 import sys
 import warnings
 
@@ -231,8 +233,6 @@ def _parse_cli_args(argv: list[str], config: HindsightConfig) -> ParsedCliArgs:
 
 def main():
     """Main entry point for the CLI."""
-    global _memory
-
     load_dotenv_for_entrypoint()
 
     # Arm profiling here, after .env is loaded and before anything starts serving, so a
@@ -280,6 +280,46 @@ def main():
         # duplicate daemons.
         daemonize()
 
+    # Reserve the actual listeners before lazy imports or engine initialization. Keeping
+    # them through serving avoids the race in a check-then-close port probe.
+    try:
+        sockets = asyncio.run(_bind_sockets(args.host, args.port))
+    except OSError as exc:
+        logging.error("Cannot bind %s:%s: %s", args.host, args.port, exc)
+        raise SystemExit(1) from exc
+    try:
+        _serve(args, config, is_daemon, sockets)
+    finally:
+        for sock in sockets:
+            sock.close()
+
+
+async def _bind_sockets(host: str, port: int) -> list[socket.socket]:
+    """Use asyncio's hostname/IPv4/IPv6 resolution without accepting any connections."""
+    server = await asyncio.get_running_loop().create_server(asyncio.Protocol, host=host, port=port, start_serving=False)
+    sockets = []
+    try:
+        # Transfer ownership out of the temporary loop; its server never starts serving.
+        for sock in server.sockets or ():
+            sockets.append(sock.dup())
+        for sock in sockets:
+            # bind alone permits competing SO_REUSEADDR binders on some systems.
+            # Establish exclusive listeners now, before any expensive initialization.
+            sock.listen(socket.SOMAXCONN)
+            sock.set_inheritable(True)
+        return sockets
+    except BaseException:
+        for sock in sockets:
+            sock.close()
+        raise
+    finally:
+        server.close()
+        await server.wait_closed()
+
+
+def _serve(args: argparse.Namespace, config: HindsightConfig, is_daemon: bool, sockets: list[socket.socket]):
+    global _memory
+
     # Print banner (not in daemon mode)
     if not is_daemon:
         print()
@@ -314,15 +354,11 @@ def main():
     # Load operation validator extension if configured
     operation_validator = load_extension("OPERATION_VALIDATOR", OperationValidatorExtension)
     if operation_validator:
-        import logging
-
         logging.info(f"Loaded operation validator: {operation_validator.__class__.__name__}")
 
     # Load tenant extension if configured
     tenant_extension = load_extension("TENANT", TenantExtension)
     if tenant_extension:
-        import logging
-
         logging.info(f"Loaded tenant extension: {tenant_extension.__class__.__name__}")
 
     # When using workers or reload, we must use import string so each worker can import the app
@@ -432,7 +468,27 @@ def main():
             text_search_extension=config.text_search_extension,
         )
 
-    uvicorn.run(**uvicorn_config)
+    _run_uvicorn(sockets=sockets, **uvicorn_config)
+
+
+def _run_uvicorn(*, sockets: list[socket.socket], **kwargs):
+    """Keep Uvicorn's supervisor and startup-failure behavior with prebound sockets."""
+    from uvicorn.main import STARTUP_FAILURE
+    from uvicorn.supervisors import ChangeReload, Multiprocess
+
+    config = uvicorn.Config(**kwargs)
+    server = uvicorn.Server(config)
+    try:
+        if config.should_reload:
+            ChangeReload(config, target=server.run, sockets=sockets).run()
+        elif config.workers > 1:
+            Multiprocess(config, target=server.run, sockets=sockets).run()
+        else:
+            server.run(sockets=sockets)
+    except KeyboardInterrupt:
+        pass
+    if not server.started and not config.should_reload and config.workers == 1:
+        raise SystemExit(STARTUP_FAILURE)
 
 
 if __name__ == "__main__":
--- a/hindsight-api-slim/hindsight_api/main.py
+++ b/hindsight-api-slim/hindsight_api/main.py
@@ -14,6 +14,7 @@ import argparse
 import asyncio
 import atexit
 import dataclasses
+import errno
 import logging
 import os
 import signal
@@ -288,6 +289,8 @@ def main():
         logging.error("Cannot bind %s:%s: %s", args.host, args.port, exc)
         raise SystemExit(1) from exc
     try:
+        args.port = sockets[0].getsockname()[1]
+        config = dataclasses.replace(config, port=args.port)
         _serve(args, config, is_daemon, sockets)
     finally:
         for sock in sockets:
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (commented):
The focused suite passes at `e31d75e6`, including the real retained-socket HTTP test. One boundary in the new resolver test is currently hiding a behavior change, though: `--host localhost --port 0` no longer selects one ephemeral service port. On macOS/Python 3.11, the exact head returns:

```
[(AF_INET, ('127.0.0.1', 55847)), (AF_INET6, ('::1', 55848, 0, 0))]
```

So the process serves IPv4 and IPv6 on different ports, while `args.port` remains `0`. The parametrized test only checks each assigned port is positive, so it accepts that split. `asyncio.create_server(..., port=0)` allocates independently per resolved address; duplicating the sockets preserves it.

Could the reservation pin the first assigned port and bind the remaining resolved addresses to that same port (with retry/cleanup if a later bind loses the race), or reject port 0 explicitly if it is not intended to be supported? A regression asserting one unique `getsockname()[1]` across the `localhost` sockets would keep the CLI's one-port contract explicit.

j0935586110-lgtm:
We hit this in production on the standalone `hindsight-api` path, and it matches #4281 exactly.

Environment: Linux (Ubuntu, kernel 6.8), `hindsight-api --daemon --idle-timeout 0 --port 9177` under a systemd **user** unit with `Restart=always`. A stale process was holding 9177, so the unit entered a restart loop — **146 failed starts in ~21 minutes**, each one re-running full initialization (embedded PostgreSQL + embedding/reranker models + migrations) before failing at the uvicorn bind with `[Errno 98] address already in use`.

Impact: the repeated full inits pinned a shared host that also runs several other agents, and orphaned embedded PostgreSQL instances were left behind, so the port drifted and unrelated services lost their database. Recovery required stopping the unit and killing the leftovers by hand.

So this isn't just "the port conflict is reported late" — under a restart policy it becomes a resource storm, which is why reserving the listener before initialization is the right shape.

Happy to re-test this branch on the affected setup (isolated, same systemd shape) and report back if that would be useful.

r266-tech:
Thanks for the concrete systemd report and the port-0 review. I reproduced the localhost split on the previous head and pushed `780f4279`, rebased onto current main. Ephemeral selection now reserves one port across all resolved addresses, retries a competing bind up to five times, and passes the selected port into the CLI/configuration. The final listeners stay held through initialization and serving.

The strengthened localhost and CLI-port assertions fail on the old head. All 38 entrypoint/daemon tests pass on the update, including real competing listeners, retry exhaustion/cleanup and HTTP through a retained socket; the full repository lint passes.

An isolated retest with the reported systemd setup would be useful: with the target port already occupied, confirm the process exits before engine initialization, leaves the existing listener intact, and starts normally once that conflict is removed. I have not reproduced the full model/PostgreSQL restart storm locally, so I am not claiming that deployment validation. AI-assisted implementation and self-review.

koriyoshi2041 (approved):
The updated port-0 path now preserves one service port across localhost IPv4/IPv6 listeners and retries the fixed-port reservation without leaking sockets. I re-ran the focused main-module suite at 780f4279: 22/22 passed on macOS/Python 3.11, including the strengthened localhost uniqueness and retry-exhaustion cases; diff hygiene also passes. The hosted blackbox failure is in the unchanged mental-model refresh-dedupe surface, outside this two-file API entrypoint diff. The original boundary is resolved.

j0935586110-lgtm:
Retest with the reported systemd setup — all three points verified locally.

Environment: Linux (Ubuntu, kernel 6.8), Python 3.12, isolated checkout of `780f4279`
(clean tree; `git rev-parse HEAD` = `780f42793…`; `hindsight_api/main.py` sha256 `3cb9089c…d6ac`).
The live service kept port 9177 occupied for the entire test; nothing on the host was stopped,
restarted, or killed.

**1. Occupied port → exits before engine initialization.**
`hindsight-api --idle-timeout 0 --port 9177` → exit `1` in **1.275 s**. Output is the two config
warnings plus `ERROR:root:Cannot bind 0.0.0.0:9177: [Errno 98] address already in use`. Zero
engine-init markers in every occupied run — no `Starting embedded PostgreSQL`, no `PostgreSQL
started`, no `MemoryEngine`, no `Application startup complete`. A 15-attempt restart loop behaved
identically (15/15 bind failures, 0 engine markers, 33.4 s total).

**2. Existing listener intact.**
Across 220 samples taken during that conflict loop, the pre-existing listener kept the same PID and
answered `/health` with 200 (220/220). Its `/proc` start time predates the whole test window, and it
runs from a different install than the tested checkout.

**3. Free port → starts normally.**
On a free port the process reserved it, started embedded PostgreSQL, reached
`Application startup complete.`, and served `/health` 200 for 388 consecutive samples before a
bounded SIGTERM shutdown. Raw `ss` captured during the run shows the listener belonging to the same
PID the app logs as `Started server process [...]`.

For contrast at the source level: the pre-PR revision has no `_bind_sockets` in `main()` and calls
`MemoryEngine(...)` directly, so this diff really does move socket reservation ahead of engine
construction.

**One boundary worth recording**, since it limits what the fix claims: in `--daemon` mode the
supervisor parent exits before the bind, which happens in the re-exec'd child. So with
`Restart=always` an occupied port still restarts — what changes is that each attempt no longer
initializes models/PostgreSQL (1.3 s, no engine work) instead of doing the full init every cycle.
That is exactly the reported harm (146 failed starts with repeated initialization), so the change
does remove it. If you want the restart loop itself suppressed, that is a separate mechanism
(`StartLimit*`, or refusing to start when the port is held by another process).

**Residual limitation of this run:** point 3 was exercised with the foreground invocation; a
free-port *daemon* start was not observed end-to-end (the bind path after `daemonize()` is the same
code, but I did not observe it directly).

r266-tech:
Thanks for completing the isolated retest and spelling out the boundary. Your occupied-port runs and unchanged existing listener address the reported repeated-initialization harm; the successful foreground startup adds deployment evidence beyond my retained-socket test.

Agreed on the daemon distinction: reservation happens in the re-executed child after `daemonize()`. This change prevents engine/model/PostgreSQL initialization on that child's bind failure; it does not suppress a supervisor's restart policy. I will keep free-port daemon startup listed as unverified end-to-end rather than infer it from the foreground run.

The current hosted failure is `test_33_refresh_dedupe.py::test_the_bank_does_not_accumulate_a_refresh_per_request` (5 refresh operations), with 228 other blackbox tests passing. That test and the refresh implementation are unchanged in this two-file entrypoint patch; I have not treated the red aggregate as a passing CI run.

nicoloboschi:
Thanks for digging into this and for the thorough tests. We went with a smaller fix in #4386 (now merged): it runs a pre-flight bind probe that copies uvicorn's own bind (address family, `SO_REUSEADDR`), never calls `listen()`, and leaves `uvicorn.run` untouched. We chose that because binding through asyncio sets `IPV6_V6ONLY` (so `--host ::` would lose IPv4), and listening during init changes how `hindsight-embed`'s port and `/health` checks behave while a daemon is booting. Closing in favour of #4386.

PR #4299, opened 2026-09-10
perf(recall): account for a recall's time per phase, and fix the three things that showed up

nicoloboschi:
A recall's time was mostly unattributed: `hindsight.operation.duration` gave one opaque number, and subtracting the store's own stage timings left a large residual nobody could explain. This instruments the request end to end, then fixes what the instrumentation found.

Reviewable as two halves — the `diag(...)` commits add measurement and change no behaviour; the `perf(...)` commits are the fixes. Happy to split into two PRs if you would rather, though they overlap in `metrics.py`.

## The measurement (commits 1-7)

Times every await, `async with` and `async for` on the recall path, plus the request region *outside* the handler (`mw_and_routing` / `deps_total` / `body_parse`), and adds an event-loop lag probe to tell a slow await from a busy loop.

Two instrumentation bugs of my own fixed along the way, both of which had been actively misleading:
- `on_recall_complete` has **three** call sites and only the first was wrapped, so `validate_post` read 0.0 ms while the path actually taken went through one of the other two. That hid 39% of the request.
- `recall_duration` was measured at the end of the handler, which made `post_recall` zero by construction.

The last commit adds `store_hop_overhead` — the hop minus the store's own reported stages. It has to be per-request: percentiles of separate phases are not additive, so this gap cannot be subtracted after the fact. It measures 4.8 ms mean / 10 ms p99, which is what rules the client out as the source of a ~1.9 s request tail.

## The fixes (commits 8-9)

Found with py-spy against a CPU-bound API under recall load:

**Every phase was recorded twice.** `record_recall_phase` and `record_validator_phase` wrote each measurement into a duration histogram *and* a parallel invocation counter — but the histogram already carries `_count` for the identical attribute set. OpenTelemetry's `consume_measurement` was **8.9% of the process's busy CPU**, `record_recall_phase` alone **5.35%**, and it ran twice for nothing.

⚠️ **Metric removal:** `hindsight.recall.phase.calls` and `hindsight.validator.phase.calls` are deleted rather than left unwritten — a metric that exists and is never recorded is worse than one that is gone. `hindsight_recall_phase_duration_seconds_count` has the same value. Flagging in case a dashboard scrapes them.

Also adds `HINDSIGHT_API_RECALL_DIAGNOSTIC_PHASES=false` to drop the subset phases, which are the bulk of the instruments on a busy path and only wanted while diagnosing.

**gzip on every recall response.** `GZipMiddleware` compressed anything over 1 KB, which a recall always is — **~5% of the request's CPU** for a deployment that is CPU-bound rather than bandwidth-bound. `HINDSIGHT_API_GZIP_MIN_SIZE` raises the floor; negative drops the middleware. **Default unchanged at 1024.**

**TLS setup for a plaintext embedder.** `httpx.Client()` builds an SSLContext and loads the system CA bundle whatever the scheme, and an in-cluster TEI is plain `http://`. `ssl.load_default_certs` showed up in the profile for exactly that, paid again for each thread the pool retires and recreates. Skipped when the base URL is `http://`; an `https://` TEI verifies exactly as before.

## Effect

Measured on one 2-vCPU pod, warm, with workers matched to the CPU limit. These two fixes plus that worker change took the rate holding p99 &lt; 1 s from **90/s to 150/s** and CPU from **13.6 to 11.5 ms per request**. The rest of the gap to a later 180/s came from changes outside this repo.

## Tests

`tests/` passes (435 passed, 30 skipped on the embedding/metrics selection; `test_profiling.py` 8 passed). `test_profiling` sliced the histogram's argument block to the counter that follows it, so its delimiter moves to the next instrument. ruff clean.


nicoloboschi:
Review pass pushed (2c081f10b + merge of main):

- **CI fix**: `verify-generated-files` was failing on unformatted code (ruff format).
- **Config**: the three new env vars were read ad hoc via `os.environ`. They are now `HindsightConfig` fields, documented in `configuration.md` and `.env.example` (+ embed copy):
  - `HINDSIGHT_API_GZIP_MIN_SIZE` (default 1024, negative disables)
  - `HINDSIGHT_API_RECALL_DIAGNOSTIC_PHASES` (default true)
  - ⚠️ `HINDSIGHT_API_LOOP_LAG` → **`HINDSIGHT_API_LOOP_LAG_REPORT_SECONDS`** (default 0 = off), to say what the number means.
- **Bug**: in `get_request_context` the `hs_deps_t0` line sat above the docstring, which turned the docstring into a no-op string expression. Moved below.
- `_bind_bank_id` decides once per function whether to time the recall body, and no longer swallows exceptions with a bare `except: pass`.
- Removed duplicate timers (`_t0_sem`, `_t0_swr`) in favour of the existing `semaphore_wait_start` / `backend_acquire_start`.
- `loop_lag`: the probe task is now strongly referenced (an unreferenced task can be GC'd mid-run); the noqa lambda is gone.
- Tests added for the diagnostic-phase flag, TEI `verify` over http vs https, and the loop-lag probe.

PR #4305, opened 2026-09-10
fix(reranker): release the MLX buffer cache after each jina-mlx batch

w-seok:
`JinaMLXCrossEncoder` does not release its inference memory after a batch. The other two in-process rerankers do: `LocalSTCrossEncoder` (`cross_encoder.py:333`) and `FlashRankCrossEncoder` (`cross_encoder.py:1106`) both call `release_local_inference_memory` in a `finally`.

MLX keeps freed Metal buffers in its own cache and only releases them once the cache limit is exceeded. That limit defaults to the device memory limit — measured at **121.60 GB on a 128 GB Mac** — so in practice it never releases anything. On a laptop running `jina-mlx` as the reranker, `IOAccelerator` memory reached **42.7 GB after 486 rerank calls**, about 90 MB per call. Only 3.9 GB of that was still in use. macOS cannot drop the rest, because the GPU driver marks those regions non-discardable, so it compresses them instead: system-wide compressed memory sat at 41.67 GB at a compression ratio of 1.21:1. After the process was restarted with the cache bounded, the same ratio came back to 2.44:1.

Adding the missing call alone would not have fixed it. `_empty_gpu_cache` (`local_device.py:171`) looks the backend up with `getattr(torch, device_type, None)`, and torch has no `mlx` attribute, so it returns having freed nothing — no error, no log. Any future non-torch backend has the same gap.

## The change

- `_empty_gpu_cache` gets an `mlx` branch that calls `mx.clear_cache()` and returns before the torch lookup.
- `JinaMLXCrossEncoder` sets `self._device_type = "mlx"` and wraps its scoring loop in `try/finally`, the same shape the other two use. The release runs inside `self._mlx_lock`, since MLX Metal ops are not thread safe — that is why #1113 added the lock.
- The `gc.collect()` gate from #3858 now skips `mlx` as well as `cpu`. MLX arrays are refcounted C++ objects behind Python wrappers, the same reason that gate was written for CPU, so a full cycle scan on every batch does not release anything extra.

I could not measure the cost of that last one. Running the same no-release baseline three times over the batch gave 1125.7 / 1223.0 / 1182.3 ms — a 97 ms spread. The difference I was trying to see, between calling the helper with `"mlx"` and with `"cpu"`, was 40 ms. A live instance was sharing the GPU on that machine, so a clean number would need the service stopped. This part follows #3858's reasoning rather than a measurement.

## Why clear instead of cap

Capping the cache with `mx.set_cache_limit()` also works, and is what I run locally. But a cap needs a number, and a number that fits one machine's workload does not fit another's. Clearing after each batch needs no configuration, and matches what the other two providers already do. Clearing measured 1209.4 ms against a 1182.3 ms baseline on the run above — 27 ms, inside the same 97 ms spread, so I would not read it as a cost.

## Tests

`tests/test_local_device.py` — four cases in `TestEmptyGpuCache`, one in
`TestReleaseLocalInferenceMemory`, using a `_fake_mlx()` stub built the same way as the
existing `_fake_torch()`:

- `_empty_gpu_cache("mlx")` calls `mx.clear_cache()` once
- the `mlx` branch returns before the torch lookup, so `torch.*.empty_cache` is never reached
- with `mlx.core` unimportable, nothing raises
- a `clear_cache()` that raises does not propagate
- `release_local_inference_memory("mlx")` trims the heap and clears the MLX cache, and skips `gc.collect()`

`tests/test_jina_mlx_release.py` — four cases on the provider:

- `_load_model()` sets `_device_type` to `"mlx"`
- a successful batch releases, and passes `"mlx"` as the device type
- a batch that raises still releases
- an empty batch does not release, since nothing was allocated

```
tests/test_local_device.py ............................ 27 passed
tests/test_jina_mlx_release.py ......................... 4 passed
+ test_jina_mlx_import_error.py, test_local_cross_encoder.py
                                                  54 passed in 5.4s
```

Each change reverted one at a time, to check the tests actually catch it:

```
drop the finally in _predict_sync            2 failed
drop self._device_type = "mlx"               1 failed
drop the mlx branch in _empty_gpu_cache      2 failed
restore the gc gate to `!= "cpu"`            1 failed
```

`ruff check`, `ruff format --check` and `ty check hindsight_api` pass.

Happy to change any of this, including dropping the `gc` gate change if you would rather keep it separate.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
+++ b/hindsight-api-slim/hindsight_api/engine/cross_encoder.py
@@ -1453,6 +1453,10 @@ class JinaMLXCrossEncoder(CrossEncoderModel):
         # Device::end_encoding() crash with SIGSEGV (NULL deref).
         # Serialize all reranker inference through this lock.
         self._mlx_lock = threading.Lock()
+        # Picked up by release_local_inference_memory() below, which routes "mlx" to
+        # mx.clear_cache(). Without it MLX holds every freed Metal buffer for the life
+        # of the process.
+        self._device_type = "mlx"
         logger.info("Reranker: jina-mlx provider initialized")
 
     def _predict_sync(self, pairs: list[tuple[str, str]]) -> list[float]:
@@ -1467,13 +1471,18 @@ class JinaMLXCrossEncoder(CrossEncoderModel):
         all_scores = [0.0] * len(pairs)
 
         with self._mlx_lock:
-            for query, indexed_docs in query_groups.items():
-                docs = [doc for _, doc in indexed_docs]
-                indices = [idx for idx, _ in indexed_docs]
-                results = self._reranker.rerank(query, docs)
-                for result in results:
-                    original_idx = result["index"]
-                    all_scores[indices[original_idx]] = result["relevance_score"]
+            try:
+                for query, indexed_docs in query_groups.items():
+                    docs = [doc for _, doc in indexed_docs]
+                    indices = [idx for idx, _ in indexed_docs]
+                    results = self._reranker.rerank(query, docs)
+                    for result in results:
+                        original_idx = result["index"]
+                        all_scores[indices[original_idx]] = result["relevance_score"]
+            finally:
+                # Inside the lock: MLX Metal ops are not thread-safe, which is why
+                # #1113 introduced _mlx_lock in the first place.
+                release_local_inference_memory(self._device_type)
 
         return all_scores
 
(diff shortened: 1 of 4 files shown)


nicoloboschi (approved):
Verified locally on an M3 Max with the real jina-reranker-v3-mlx: the MLX cache grew ~730 MB/call without the fix (10.9 GB after 15 calls) and stays at 0 MB over 100 calls with it, no latency cost. Only the jina-mlx path passes "mlx", so torch/CPU providers are unaffected. Thanks!

PR #4309, opened 2026-09-10
fix(api): report deleted_count from DELETE /banks/{bank}/memories

### LINKED ISSUE #4307: Bulk memories delete reports no `deleted_count`, even when it deleted rows
### What happens

`DELETE /v1/{tenant}/banks/{bank}/memories` reports success but never says how much it
removed. Measured against a real bank holding 6 memories:

```
GET    /v1/default/banks/<bank>/memories/list?limit=1   → {"total": 6, ...}
DELETE /v1/default/banks/<bank>/memories                → {"success":true}
GET    /v1/default/banks/<bank>/memories/list?limit=1   → {"total": 0, ...}
```

Six memories were deleted and the response carried no count. `deleted_count` is absent from
the body; on an earlier image it was present as `null`. Either way a caller gets nothing.
Adding a `?type=` filter makes no difference.

### What I expected

`deleted_count` populated with the number of memories removed. The delete has just done the
work that produces that number, so it is the only participant that can report it accurately.

### Why it matters

The response cannot distinguish "deleted everything" from "deleted nothing". The only
workaround is to count the scope on both sides of the delete and report the difference,
which is what we do — and it is unsatisfying in two ways:

- it costs two extra `memories/list` calls per delete;
- **it is inaccurate under concurrency.** A memory retained between the delete and the
  post-count moves the delta down by one, so the figure can under-report an erasure. We
  chose differencing over simply reporting the pre-count precisely so that it can never
  *over*-report one — telling an operator that data was destroyed when it was not is the
  worse failure — but that only bounds the error, it does not remove it. **No client-side
  workaround can close that gap**, because the exact figure only exists inside the delete's
  own transaction — where, as shown below, it is already computed and then discarded.

For a delete endpoint, "how much did I just erase?" is the question a caller most needs
answered exactly, and it is the one thing the response cannot say.

### Reproduction

```bash
BANK=repro_$(date +%s)
# create the bank and put some memories in it, then:

curl -s "$HINDSIGHT/v1/default/banks/$BANK/memories/list?limit=1" | jq .total
# → 6

curl -s -X DELETE "$HINDSIGHT/v1/default/banks/$BANK/memories"
# → {"success":true}          # no deleted_count

curl -s "$HINDSIGHT/v1/default/banks/$BANK/memories/list?limit=1" | jq .total
# → 0                          # so 6 rows were removed and the response said nothing
```

### Environment

- Image: `ghcr.io/vectorize-io/hindsight:latest`
- Postgres backend
- Observed 2026-08-02 (`deleted_count: null`) and re-confirmed 2026-09-10 (field absent)

### Suggested fix

**The count already exists and is thrown away.** In
`hindsight-api-slim/hindsight_api/api/http.py`, `api_clear_bank_memories` awaits
`app.state.memory.delete_bank(bank_id, fact_type=type, delete_bank_profile=False, ...)`
without binding the result, then returns a bare `DeleteResponse(success=True)`. That engine
method is declared `-> dict[str, int]` ("Dict with deletion counts") in
`engine/interface.py`, and the sibling route `api_delete_bank` — calling the *same* method —
already reads it:

```python
deleted_count=result.get("memory_units_deleted", 0)
          + result.get("entities_deleted", 0)
          + result.get("documents_deleted", 0),
```

So the two `delete_bank` callers disagree about whether the caller gets a number, which also
explains why `DeleteResponse.deleted_count` is `int | None`.

`MemoryEngine.delete_bank` (the only concrete implementation of that interface method)
populates `memory_units_deleted` on **both** of its branches, so binding the result and
passing it through is the whole fix:

- `if fact_type:` → `result = {"memory_units_deleted": units_count, "entities_deleted": 0}`
- `else:` → `result = {"memory_units_deleted": units_count, "entities_deleted": ...,
  "documents_deleted": ...}`

`result` is returned unmodified, and the `delete_bank_profile=False` path the clear endpoint
uses only adds a vector-index maintenance call — it does not touch `result`. Both branches
take `units_count` from a `SELECT COUNT(*)` issued **inside the same transaction** as the
`DELETE`, which is strictly more accurate than anything a client can reconstruct from
outside.

Please also use `0` rather than `null`/absent when the scope was already empty, so a missing
value keeps meaning "not reported" rather than "nothing to report". Happy to send a PR if
that direction is welcome.

L4XB:
Fixes #4307.

`DELETE /v1/{tenant}/banks/{bank}/memories` answered with a bare `{"success": true}` (or `deleted_count: null` on older builds). The engine already counts the removed units inside the delete transaction: `MemoryEngine.delete_bank` returns `memory_units_deleted` on both the type-filtered path and the whole-bank path. `api_clear_bank_memories` simply discarded that result, so a caller could not distinguish "deleted everything" from "deleted nothing" and had to difference two extra `memories/list` calls, which under-reports when a retain lands in between.

## Change

- `api_clear_bank_memories` keeps the engine result and returns `deleted_count=result["memory_units_deleted"]` plus a message naming the count and scope (`Deleted 6 memory units from bank 'x'` / `Deleted 2 world memory units from bank 'x'`).
- `DeleteResponse` already declares `deleted_count: int | None` and the generated clients already carry the field, so there is no schema or client change.

## Tests

`tests/test_clear_bank_memories_deleted_count.py` drives the endpoint through `create_app(SimpleNamespace(...))` with an `AsyncMock` engine (same pattern as `test_bank_templates.py`), and checks the count for the unfiltered delete, the `?type=world` delete (filter forwarded, `delete_bank_profile=False`), and an empty bank (`0`, not `null`). Ruff check/format pass.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -9357,11 +9357,19 @@ def _register_routes(app: FastAPI):
     ):
         """Clear memories for a memory bank, optionally filtered by type."""
         try:
-            await app.state.memory.delete_bank(
+            result = await app.state.memory.delete_bank(
                 bank_id, fact_type=type, delete_bank_profile=False, request_context=request_context
             )
-
-            return DeleteResponse(success=True)
+            # The delete is the only place that knows exactly how many units went; a caller
+            # cannot recover that number afterwards (a pre/post count races with concurrent
+            # retains), so report it instead of a bare success.
+            deleted_count = result.get("memory_units_deleted", 0)
+            scope = f"{type} memory units" if type else "memory units"
+            return DeleteResponse(
+                success=True,
+                message=f"Deleted {deleted_count} {scope} from bank '{bank_id}'",
+                deleted_count=deleted_count,
+            )
         except OperationValidationError as e:
             raise HTTPException(status_code=e.status_code, detail=e.reason)
         except (AuthenticationError, HTTPException):
--- a/hindsight-api-slim/hindsight_api/engine/memory_engine.py
+++ b/hindsight-api-slim/hindsight_api/engine/memory_engine.py
@@ -10222,16 +10222,17 @@ class MemoryEngine(MemoryEngineInterface):
                         # dependent observations AFTER the delete below. Running the
                         # stale-observation sweep post-delete ensures we also catch
                         # observations inserted concurrently by consolidation.
+                        # Both the stale-observation ids and the deleted count must come from
+                        # wherever the memories live: for a store that keeps them elsewhere the
+                        # memory_units table is intentionally empty, so reading it yields nothing
+                        # (the sweep would silently skip, and the count would report 0 rows).
+                        from .memories import get_memories as _get_memories_for_scope
+
+                        _scope_store = _get_memories_for_scope()
+                        _scope_store_owned = _scope_store.store_owned_for(bank_id)
                         unit_ids: list[str] = []
                         if fact_type in ("experience", "world"):
-                            # These ids drive the stale-observation sweep below, so they must come
-                            # from wherever the memories live: reading memory_units for a store that
-                            # keeps them elsewhere yields nothing, and the sweep would silently skip,
-                            # leaving observations behind that outlive the sources they summarise.
-                            from .memories import get_memories as _get_memories_for_scope
-
-                            _scope_store = _get_memories_for_scope()
-                            if not _scope_store.store_owned_for(bank_id):
+                            if not _scope_store_owned:
                                 unit_id_rows = await conn.fetch(
                                     f"SELECT id FROM {fq_table('memory_units')} WHERE bank_id = $1 AND fact_type = $2",
                                     bank_id,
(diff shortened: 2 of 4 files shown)


koriyoshi2041 (commented):
One backend path still reports `0` after deleting rows. In `MemoryEngine.delete_bank`, the `fact_type` branch always derives `units_count` from the SQL `memory_units` table, but for a store-owned bank that table is intentionally empty; the actual deletion happens later through `store.delete_where(...)`. The unfiltered branch already switches to `count_memories()` for store-owned banks, but the filtered branch does not.

Could the count be made backend-aware before this endpoint exposes it as authoritative? A regression using a store-owned bank with filtered memories would also keep the response contract honest; the new endpoint tests mock the engine result, so they cannot catch this case.

L4XB:
Good catch, thanks — fixed in the follow-up commit. `delete_bank`'s `fact_type` branch now resolves the store once (the same lookup the stale-observation scan already used) and, for a store-owned bank, takes the count from `store.count_memories(...)[fact_type]` instead of the SQL `memory_units` table; SQL-backed banks keep the existing `COUNT(*)`.

Regression: `tests/test_delete_bank_filtered_count_store_owned.py` drives `delete_bank(fact_type=...)` on a bare engine with a store-owned fake store (`store_owned_for → True`, `count_memories`, `scan_memories`, `delete_where`) and a connection whose SQL side is empty, for `world` and `observation`. It asserts `memory_units_deleted` equals the store's count and that `delete_where` receives `fact_types=[fact_type]`. On the previous commit the test fails with `0 == 3`. The endpoint tests stay mocked at the engine boundary, as before.

koriyoshi2041 (approved):
The follow-up closes the store-owned filtered-count gap: the count now comes from the same backend that `delete_where` mutates, while SQL-backed banks retain the existing path. I checked exact head `ec8469190` and ran both focused test files locally: 5 passed.

nicoloboschi:
Closing as a duplicate of #4312, which makes the same fix (endpoint count + the store-owned `?type=` count). Thanks for the work, @L4XB!

PR #4310, opened 2026-09-10
fix(openclaw): expose knowledge tool payloads as details for Code Mode

### LINKED ISSUE #4308: agent_knowledge_* tools return empty results in OpenClaw Code Mode (details overwritten with {})
## Bug Description

All knowledge tools provided by `@vectorize-io/hindsight-openclaw` (`agent_knowledge_recall`, `agent_knowledge_reflect`, `agent_knowledge_list_pages`, `agent_knowledge_get_page`, `agent_knowledge_ingest`, ...) silently return an empty object `{}` when called from an OpenClaw **Code Mode** session, even though the Hindsight API correctly serves the results.

In non-Code-Mode sessions the same tools work, because there the model receives the `content[0].text` payload.

**Root cause:**

1. `@vectorize-io/hindsight-agent-sdk@0.1.1` — `createKnowledgeTools()` wraps every result via `ok()`:

   ```js
   function ok(data) {
       return { content: [{ type: "text", text: JSON.stringify(data, null, 2) }] };
   }                                                                    // dist/index.js ~L29
   ```

2. `@vectorize-io/hindsight-openclaw@0.11.1` — the plugin's tool-registration wrapper (dist/index.js ~L2440-2453) spreads the SDK result and **overwrites `details` with an empty object**:

   ```js
   return { ...(await t.execute(params)), details: {} };
   ```

3. OpenClaw (2026.9.2) Code Mode bridge — for any tool result that has a `details` key, **only `details`** is passed to the code-mode guest as the structured value:

   ```js
   value = isRecord(called.result) && "details" in called.result ? called.result.details : called.result;
   ```

   (`code-mode-state-CvxJMErC.js:285` in the 2026.9.2 dist bundle)

So the guest receives `details` = `{}`, and the actual payload (a JSON string inside `content[0].text`) is dropped.

## Steps to Reproduce

1. OpenClaw 2026.9.2 with `@vectorize-io/hindsight-openclaw` 0.11.1 (`hindsight-agent-sdk` 0.1.1), `enableKnowledgeTools: true`, static `bankId` (`dynamicBankId: false`).
2. In a Code Mode session call: `agent_knowledge_recall({ query: "anything" })`.
3. Observed result: `{}` — `Object.getOwnPropertyNames(result)` is `[]`.
4. Meanwhile the hindsight-api log for the same query shows `[RECALL ...] Complete: N facts ... results=N`, and a direct `POST /v1/default/banks/<bank>/memories/recall` returns `{"results":[...], "entities":[...]}`.

## Expected Behavior

Code Mode guests should receive the actual recall/page results as a structured value.

## Actual Behavior

Every `agent_knowledge_*` tool result arrives as `{}` in Code Mode, while hindsight-api logs show the data was retrieved and returned successfully over HTTP.

## Suggested fix (plugin side)

Put the structured payload into `details` in the wrapper, e.g.:

```js
const res = await t.execute(params);
let payload = {};
try { payload = JSON.parse(res?.content?.[0]?.text ?? "{}"); } catch {}
return { ...res, details: payload };
```

(Alternatively, have the SDK's `ok()` return `{ content: [...], details: data }` directly — then the wrapper spread would already do the right thing.)

An OpenClaw-side fallback (parse `content[0].text` when `details` is empty) is possible, but the plugin-side fix is cleaner since OpenClaw's convention is that `details` carries the structured tool result.

## Version

- openclaw: 2026.9.2 (3928bad)
- @vectorize-io/hindsight-openclaw: 0.11.1
- @vectorize-io/hindsight-agent-sdk: 0.1.1
- hindsight-api: ghcr.io/vectorize-io/hindsight-api:latest-slim

## LLM Provider

Other (OpenRouter)

L4XB:
Fixes #4308.

The plugin's tool-registration wrapper in `hindsight-integrations/openclaw/src/index.ts` returned the SDK result with `details: {}`:

```ts
return { ...(await t.execute(params)), details: {} };
```

OpenClaw's Code Mode bridge hands a tool result's `details` (and only that) to the guest as the structured value, so every `agent_knowledge_*` call came back as `{}` in Code Mode while the JSON payload sat unread in `content[0].text`. Non-Code-Mode sessions kept working because the model reads the text content.

## Change

- New exported helper `knowledgeToolDetails(result)` parses the SDK's JSON text payload (`content[0].text`) back into an object for `details`. A non-object payload (array/scalar) is wrapped as `{ result }`; unparseable or missing text still yields `{}`, so nothing that worked before changes.
- The identity-error branch now reports `details: { error }` instead of `{}` for the same reason.
- The `content` text is untouched, so the non-Code-Mode path is unaffected.

I kept the change in the plugin rather than in `hindsight-agent-sdk`'s `ok()` so the SDK's MCP-style result shape stays as it is for its other consumers; if you would rather have the SDK return the structured data alongside `content`, I can move it there.

## Tests

`src/index.test.ts` gains three unit tests for `knowledgeToolDetails` (object payload, wrapped non-object payload, fallbacks). Full suite: `npx vitest run src/index.test.ts` → 208 passed; `tsc --noEmit` clean.


nicoloboschi:
Reviewed, rebased and merging once CI is green.

Two things I pushed to your branch:

1. **Rebased onto current `main`.** The branch was six days stale and, more importantly, had never run CI at all — no checks were ever registered on it. #4348 merged a few minutes ago and also appends tests to `src/index.test.ts`, so the rebase conflicted there; I kept both blocks and ran prettier over the result.
2. **Moved `knowledgeToolDetails` below the imports** (`42d10d0`). It was declared between two `import` statements, which reads as an accident. Same code, no behaviour change.

On the design question you raised: keeping the parse in the plugin rather than changing the SDK's `ok()` is the right call, and it's the complete fix — `details: {}` appears in exactly two places repo-wide, both in this file, and no other harness wraps SDK results that way.

353 tests pass, tsc and prettier clean. Thanks for the precise root-cause write-up in #4308 — the `code-mode-state` bundle reference is what made this a five-minute review.

PR #4319, opened 2026-09-11
feat(metrics): /metrics covers every worker (labelled api_worker), and event-loop lag as a histogram

nicoloboschi:
## Why

With `--workers N`, every uvicorn worker is its own process with its own metrics registry, but all
of them share one listening port. A scrape of `/metrics` therefore reaches **one worker, chosen by
the kernel**. Three consecutive scrapes of a 2-worker pod returned `process_cpu_seconds_total` of
566.42, then 582.04, then 566.44: two different processes. Two consequences:

- **Rates are wrong.** A counter that goes backwards between scrapes looks like a reset, so a rate
  or a histogram quantile over any API metric mixes the workers.
- **A saturated worker is invisible.** Each worker's event loop can use at most one core. Under
  load, both workers of a 2-vCPU pod can be pegged while the pod's total CPU still reads well under
  its limit.

**Labelling each worker's series isn't enough on its own.** A scrape would still return a single
worker's series, the others would be missing from about half the scrapes, and Prometheus marks a
missing series stale. Every scrape has to include every worker.

## What

**`HINDSIGHT_API_METRICS_WORKER_LABEL`** (default `false`). When it is on:

- Each worker claims a **slot** (`0..N-1`) with an exclusive, non-blocking `flock` on a per-slot
  lock file. The kernel releases the lock when the process exits, so a respawned worker reuses its
  predecessor's slot, and the label stays bounded to N values.
- Each worker publishes a **snapshot** of its registry every 5 s to a directory shared by that
  server's workers, which is keyed by the supervisor PID. Writes are atomic.
- **`/metrics`**, whichever worker answers, returns every live worker's series, each labelled
  `api_worker="<slot>"`. Its own series are read live and the others come from their latest
  snapshot. Series are **kept apart, never summed**: summing is a query-time decision, and a sum
  would hide the per-worker skew this exists to show.
- A snapshot older than 15 s belongs to a worker that is gone, and is skipped. A corrupt snapshot
  is skipped without breaking the scrape. Setup never stops the API from starting.

It uses one port and one scrape target, so it needs no change to charts, Services or scrape
configs. The label is `api_worker` rather than `worker`, because the task poller already uses that
key.

**`HINDSIGHT_API_LOOP_LAG_METRIC`** (default `false`). The existing loop-lag probe records every
50 ms sample in a `hindsight.event_loop.lag` histogram (seconds, with buckets from 1 ms to 5 s),
independent of its log reports. It complements `hindsight.event_loop.stall_duration`, which only
counts blocks past a threshold: a loop that is busy but never blocked shows up here.

Both are opt-in, and unset behaves exactly as before.

## Tests

- `tests/test_metrics_multiworker.py` checks that:
  - every worker appears in one scrape, labelled, with counters and histograms kept apart;
  - the answering worker's own series are live, not its last snapshot;
  - a gone worker's stale snapshot is skipped;
  - a corrupt snapshot is skipped without breaking the scrape;
  - the merged output round-trips through the Prometheus text parser;
  - slots are distinct, and a released slot is reused;
  - start publishes immediately, and a failure to set up disables the feature rather than raising.
- `tests/test_loop_lag.py` checks that with the metric on and reports off, every tick is recorded
  and nothing is logged.
- The metrics, metrics-lifecycle and config suites pass: 151 tests.

The docs (`configuration.md`, the generated skills copy, and `.env.example`) cover both settings.


koriyoshi2041 (commented):
Exact-head check at 902992c66: the focused metrics/loop-lag suite passes locally (46 tests), but two merge blockers remain.

1. Both hosted test-embed jobs fail at test_bundled_template_matches_repo_root because .env.example changed without the byte-identical hindsight-embed/hindsight_embed/env.example copy.
2. metrics_multiworker.py imports fcntl at module load. On Windows that module does not exist, so enabling HINDSIGHT_API_METRICS_WORKER_LABEL makes the lifespan import fail; the broad metrics-init catch then disables metrics entirely. The Windows CI does not exercise this because the flag is off. This needs a Windows-capable slot-lock path (or an explicit supported-platform gate/contract) plus coverage for the enabled flag.

The merge/render behavior itself looks coherent in the focused tests.

PR #4330, opened 2026-09-12
fix(coding-agents): preserve distinct non-ASCII document titles

### LINKED ISSUE #4329: hindsight_ingest_document: non-ASCII (e.g. Chinese) titles collapse to colliding document_id, causing silent overwrite
### Bug Description

`hindsight_ingest_document` (the MCP tool exposed by `hindsight-integrations/coding-agents`) derives the memory `document_id` purely from the `title` argument:

```ts
const docId = title.toLowerCase().replace(/[^a-z0-9]+/g, "-").replace(/^-+|-+$/g, "") || "doc";
```

(`hindsight-integrations/coding-agents/src/core/knowledge-tools.ts`, current `main` — same logic in the published `0.6.0` build.)

Because the regex strips every character outside `[a-z0-9]`, a non-Latin-script title (Chinese, Japanese, Korean, Cyrillic, etc.) loses all of its distinguishing characters. Two different titles that happen to reduce to the same leftover ASCII/digit substring — or to nothing at all — collapse onto the same `document_id`, and the second `hindsight_ingest_document` call **silently overwrites** the first document. No error, no duplicate warning, no merge.

This only affects `hindsight_ingest_document`. The sibling `hindsight_capture_initiative` tool appends `Date.now()` to its fallback marker id, so it doesn't collide the same way.

### Steps to Reproduce

1. Use any MCP-capable harness with `hindsight-coding-agents` installed (repro'd at `0.6.0`; same code on `main`).
2. Call `hindsight_ingest_document` with `title: "测试问题17记忆文档一"`, `content: "... content A ..."`.
3. Call `hindsight_ingest_document` with `title: "测试问题17记忆文档二"`, `content: "... content B ..."`.
4. Both calls return `{"ok": true, "doc_id": "17"}` — `"17"` is the only ASCII/digit substring either title has left after the strip.
5. `GET /v1/default/banks/{bank}/documents/17` shows only content A. A full-text search for content B's marker string across the bank returns zero results — it was never recoverable.

A minimal repro doesn't even need two similar titles: any two purely non-Latin titles both collapse to the hardcoded fallback `"doc"`. That exact fallback is covered by an existing unit test (`hindsight_ingest_document falls back to 'doc' when the title has no safe characters`), but the test only exercises ASCII punctuation (`"!!!///???"`) — never a real-world non-Latin-script title, so the collision case slipped through.

### Expected Behavior

Two `hindsight_ingest_document` calls with different titles and different content should either produce two distinct documents, or — if slugging the title is still desired for id-safety — fall back to something that stays unique per call (e.g. a hash of the raw title/content, or a timestamp suffix, matching the pattern `hindsight_capture_initiative`'s marker id already uses) instead of a shared static string.

### Actual Behavior

Different non-Latin-script titles collapse to the same `document_id` (or to the same hardcoded `"doc"` when no ASCII/digit survives at all), and a later `hindsight_ingest_document` call silently overwrites the earlier document under that id. The tool's return value (`{ ok: true, doc_id }`) gives no signal that the id already existed with different content.

### Version

`@vectorize-io/hindsight-coding-agents` 0.6.0 — verified the same code path is still present, unchanged, on `main`.

### Suggested fix direction

Not proposing a specific implementation, just flagging what's already in the codebase as a precedent:

- Reuse the "unique fallback" pattern `hindsight_capture_initiative`'s marker id already uses (append a short hash or timestamp) whenever the slugified title is empty or very short.
- Or transliterate non-Latin scripts before slugging instead of dropping them.
- Or hash the full raw title (not just the ASCII-surviving remainder) into the id, so two different non-Latin titles never collide.

Happy to share the exact repro (a couple of MCP tool calls plus `curl` against `/v1/default/banks/{bank}/documents`) if that's useful.

Shy7777:
`hindsight_ingest_document` gives `测试问题17记忆文档一` and `测试问题17记忆文档二` the same document ID, `17`, so saving the second document can overwrite the first. Titles without ASCII letters or numbers all share the `doc` fallback.

Append a SHA-256 hash of the original title when it contains non-ASCII characters or produces an empty slug. Saving the same exact title with new content keeps its ID, while existing nonempty ASCII slugs remain unchanged. Already-saved affected titles will receive a new ID when imported again; this does not migrate old documents or recover overwritten content.

Fixes #4329.

Validation: all 1,020 coding-agents tests pass across 72 files, including regressions for Chinese titles with a shared number, pure Chinese titles, accented titles, punctuation-only titles, and repeated imports. TypeScript checking, the package build, and formatting checks on the changed files pass. The new regression uses the real tool handler with a stubbed retain client; no live backend or LLM is required. The repository-wide `lint.sh` was attempted but stops during Python dependency sync because this sparse checkout lacks the `hindsight-all` workspace, so a full-workspace lint pass is not claimed.

### DIFF
--- a/hindsight-integrations/coding-agents/src/core/knowledge-tools.test.ts
+++ b/hindsight-integrations/coding-agents/src/core/knowledge-tools.test.ts
@@ -376,20 +376,40 @@ describe("buildKnowledgeTools", () => {
     );
   });
 
-  it("hindsight_ingest_document falls back to 'doc' when the title has no safe characters", async () => {
-    const client = stubClient();
-    const tools = buildKnowledgeTools(client, "repo-a");
-    const tool = findTool(tools, "hindsight_ingest_document");
-    await tool.handler({ title: "!!!///???", content: "x" });
-    expect(client.retain).toHaveBeenCalledWith(
-      "x",
-      "ingested document",
-      "doc",
-      ["source:upload"],
-      "document",
-      {} // no harness in these tests: nothing to stamp
-    );
-  });
+  it.each([
+    ["测试问题17记忆文档一", "测试问题17记忆文档二"],
+    ["记忆文档一", "记忆文档二"],
+    ["Résumé", "Rèsumé"],
+    ["!!!///???", "???///!!!"],
+  ])(
+    "hindsight_ingest_document keeps distinct lossy titles: %s / %s",
+    async (title, otherTitle) => {
+      const documents = new Map<string, string>();
+      const client = stubClient({
+        retain: vi.fn(async (content: string, _context: string, documentId: string) => {
+          documents.set(documentId, content);
+        }),
+      });
+      const tool = findTool(buildKnowledgeTools(client, "repo-a"), "hindsight_ingest_document");
+      const first = JSON.parse(
+        (await tool.handler({ title, content: "first document" })).content[0].text
+      );
+      const second = JSON.parse(
+        (await tool.handler({ title: otherTitle, content: "second document" })).content[0].text
+      );
+      expect(first.ok).toBe(true);
+      expect(second.ok).toBe(true);
+      expect(first.doc_id).not.toBe(second.doc_id);
+      expect(documents.get(first.doc_id)).toBe("first document");
+      expect(documents.get(second.doc_id)).toBe("second document");
+
+      const updated = await tool.handler({ title, content: "updated first document" });
+      expect(JSON.parse(updated.content[0].text)).toEqual(first);
+      expect(documents.size).toBe(2);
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (approved):
Verified the collision fix at exact head `b158086d`. The four new lossy-title cases pass (including stable re-ingestion), the package build/type generation passes, and `git diff --check` is clean. The full test file has one unrelated worktree-fixture failure in `hindsight_sync_status` because the detached worktree does not contain the expected `chat` ref; all 30 other tests pass.

Hashing the raw title while keeping the slug prefix preserves stable updates and avoids the shared-number collision without changing existing ASCII-only IDs.

nicoloboschi (approved):

PR #4332, opened 2026-09-12
fix(embeddings): retry once smaller when a provider rejects a capped input

### LINKED ISSUE #4331: Embedding truncation counts cl100k tokens while remote providers count their own: a cap of 8192 still 400s and hard-fails the task
## Version

`0.9.2` (self-hosted, `hindsight-api-slim`), embeddings via an OpenAI-compatible
endpoint (SiliconFlow, `BAAI/bge-m3`), `HINDSIGHT_API_EMBEDDINGS_MAX_INPUT_TOKENS=8192`
**explicitly set**.

## Summary

`HINDSIGHT_API_EMBEDDINGS_MAX_INPUT_TOKENS` is documented as an *approximate* bound
("counted with `HINDSIGHT_API_TOKENIZER_ENCODING`, approximate"). It is counted with
**cl100k_base**. A remote provider counts with **its own** tokenizer. Those two agree
closely on natural-language prose but diverge by **~1.6× on ASCII/code-heavy text** — so
text truncated to exactly the configured cap can still arrive at the provider over its
limit, with a permanent 400 that fails the owning task.

This is **not** #4165. That case was "nothing was bounding the input at all"
(`DEFAULT_EMBEDDINGS_MAX_INPUT_TOKENS = None`), and #4182 fixed it by defaulting the cap
to 8192. **This report is the residual gap that #4182 does not close**: the counting unit
is still cl100k, and the new default (8192) has *zero* headroom for a provider tokenizer
that expands ASCII.

## Observed

```
2026-09-12 14:20:01,008 ERROR hindsight_api.engine.memory_engine
  Task execution failed: consolidation, error: Exception: Failed to generate batch
  embeddings: Error code: 400 - {'code': 20015, 'message': 'The parameter is invalid.
  Please check again.'}

2026-09-12 14:20:01,037 INFO  hindsight_api.engine.memory_engine
  Consolidation d4c5b51e-af52-4778-8de0-432d54d04fd4 for bank coding-agent::doudian-export
  hit transient error; another consolidation is already pending for this bank — skipping retry.

2026-09-12 14:20:01,038 ERROR hindsight_api.worker.poller
  Task d4c5b51e-af52-4778-8de0-432d54d04fd4 failed: Exception: Failed to generate batch embeddings ...
```

Call site: `engine/consolidation/consolidator.py` → `_create_observation_directly`, which
embeds **a single text** (`generate_embeddings_batch(engine.embeddings, [observation_text])`).
The run had already completed 36 units across 10 LLM batches and died on one text.

Note the retry behaviour differs from #4165: the operation did **not** get its retries —
because another consolidation was pending for the same bank, the failure was taken
terminally (`retry_count: 0`, op left `failed`). So a payload-level 400 here is not
self-healing at all.

## Measurement (provider probed directly, Hindsight out of the loop)

Same endpoint/model, same API key, one text per request, cl100k token counts computed with
`tiktoken`:

| Input | chars | cl100k tokens | Result |
|---|---|---|---|
| Chinese prose | 8,000 | 8,400 | ✅ 200 |
| Code-style ASCII (indented `const … // …`) | 19,982 | 4,944 | ✅ 200 |
| Code-style ASCII | 21,825 | 5,400 | ❌ 400 |
| Code-style ASCII | 28,227 | 6,984 | ❌ 400 |
| `""`, `"   "`, `"\n"`, single char | 0–1 | 0–1 | ✅ 200 (ruled out) |

Two things follow:

1. **The effective bound is the provider's tokenizer, not cl100k.** The same endpoint
   accepts **8,400 cl100k tokens of Chinese** but rejects **5,400 cl100k tokens of code**.
   Its real limit is 8,192 of *its own* tokens (XLM-R for bge-m3); cl100k under-counts
   ASCII/code by **1.5×–1.7×** (boundary bracketed by 4,944 ✅ / 5,400 ❌).
2. So `8192` is not a bound at all for this provider — on code-shaped text it lands at
   roughly **12,400–13,600 of the provider's tokens, 50%–65% over its limit**. The docstring
   says "approximate"; the shipped/now-default value treats the approximation as exact.

**Why the failure is rare but recurrent:** the log carries 27
`Embeddings: truncated … to 8192 tokens (largest was ~X tokens)` warnings with
`X ∈ {8,762 … 10,756}`. Text routinely rides the cap. Chinese-dense ones pass (ratio ~0.98);
only the code-dense one crosses the provider limit. In this deployment that was `1` event,
in a bank whose content is source code.

## Impact

One oversized text inside a batch call fails the **whole** consolidation task, and the
observations it was writing are never created — silent data loss. It is only visible as an
ERROR in the API log; there is no alertable signal. Because the trigger is content-shaped
(code vs prose) rather than a one-off boundary, any deployment embedding code, logs, or
minified text against a third-party embedding provider is exposed.

## Suggested fixes (any of these closes it)

1. **Bound the payload by the provider's real limit, not by an approximation.** If the
   provider tokenizer is unknown, apply a safety factor to the cl100k count instead of
   treating it as exact — or lower the default well below 8192 (see workaround).
2. **Treat a 4xx from the embed call as a payload problem, not a task failure** — retry the
   batch with a halved budget, or skip/flag the single offending item and continue. This is
   the robustness half and matters independently: today one text kills a whole task.
3. **Don't let "another consolidation is already pending for this bank" turn a payload error
   into a terminal one** — that path skipped the retries entirely.

## Workaround

`HINDSIGHT_API_EMBEDDINGS_MAX_INPUT_TOKENS=4096` (restart required). Verified here: both
Chinese and code-shaped text pass at 4096, the truncation warning reports the new value
(`Embeddings: truncated … to 4096 tokens`), and a 25,000-char code-style query that
previously produced a 400 returns 200 through the app's own embedding path. Cost is
negligible: 6 of 41,842 stored memories are long enough to now truncate, and truncation
only shortens the vector input — the stored text is untouched.

## Repro

```python
# pip install tiktoken; needs any OpenAI-compatible embedding endpoint
import json, urllib.request, tiktoken
enc = tiktoken.get_encoding("cl100k_base")
unit = "    const base = { fact_types: PAGE_FACT_TYPES, tags_match: PAGE_TAGS_MATCH };  // merge trigger\n"

for n in (20_000, 22_000, 29_000):          # chars of code-shaped ASCII
    text = (unit * (n // len(unit) + 1))[:n]
    body = json.dumps({"model": "BAAI/bge-m3", "input": text}).encode()
    req = urllib.request.Request(URL, data=body, method="POST",
                                 headers={"Content-Type": "application/json",
                                          "Authorization": f"Bearer {KEY}"})
    try:
        urllib.request.urlopen(req)
        print(f"{n:>6} chars / {len(enc.encode(text)):>5} cl100k tokens -> 200")
    except urllib.error.HTTPError as e:
        print(f"{n:>6} chars / {len(enc.encode(text)):>5} cl100k tokens -> {e.code} {e.read()[:80]}")
```

Both Hindsight-side settings were unchanged during the probes; only the text varied.

L4XB:
Closes #4331.

## The gap

`HINDSIGHT_API_EMBEDDINGS_MAX_INPUT_TOKENS` bounds the input with our tokenizer, and the provider bounds it with its own. `token_encoding.py` already says so in writing, "the configured encoding is an approximation of any given provider's tokenizer, so set `max_tokens` with a little headroom below the model's real limit", and #4182 shipped a default of 8192, which is the limit itself and therefore has no headroom. A text cut to exactly the cap can arrive over the provider's limit.

What comes back is a permanent 4xx, and `generate_embeddings_batch` turns it into a bare `Exception` that fails the whole batch. For consolidation that is the owning task: `_embed_observation_text` has no handler, and neither do its callers.

## Change

When the cap actually truncated something and the provider answers 400, 413 or 422, the batch is retried once at half the budget. That needs no knowledge of the provider's tokenizer, which is the part we cannot have, and it is the smallest thing that turns a permanent failure into a completed one.

It deliberately does not fire when

- nothing was truncated, so the input size is not the suspect
- the status is not one a size rejection uses, so an auth or quota failure still fails fast, as `remote_retry.py` intends
- the first call succeeded, which is every ordinary run: no extra call, no behaviour change

The message body is not part of the decision. The report's own provider answers `{'code': 20015, 'message': 'The parameter is invalid. Please check again.'}`, which no substring match could classify.

The wrapper now also raises `from e`. Flattening the provider error into a string dropped the status, so a caller that wants to classify the failure has nothing left to read.

What this does not do: change the 8192 default or add a margin setting. That is a policy call about whether Hindsight should ship headroom and shrink everyone's embeddings, and it belongs to you rather than to a drive-by patch. This PR makes the failure survivable either way.

## Tests

`tests/test_embeddings_max_input_tokens.py`, next to the existing cap tests, with a `_RejectsOversizeBackend` that refuses anything above a token count and records each call:

- a capped input that the provider refuses is retried once, and the second call carries at most half the tokens
- an input the cap did not touch is not retried
- a 401 is not retried
- the provider error survives as `__cause__` with its status

```
uv run pytest tests/test_embeddings_max_input_tokens.py tests/test_token_encoding.py -q
34 passed
ruff check / ruff format --check   clean
ty check hindsight_api             no diagnostics on the changed file
```

The rest of the suite needs the `pg0` harness, which I do not have here, so I did not run it. Note that `test-api` is gated on `has_secrets` and will skip on a fork PR.

### DIFF
--- a/hindsight-api-slim/hindsight_api/engine/retain/embedding_utils.py
+++ b/hindsight-api-slim/hindsight_api/engine/retain/embedding_utils.py
@@ -6,6 +6,7 @@ import logging
 from typing import Literal, Protocol
 
 from ...config import ENV_EMBEDDINGS_MAX_INPUT_TOKENS, get_config
+from ..remote_retry import status_code_of
 from ..token_encoding import count_tokens, truncate_many_to_tokens
 
 logger = logging.getLogger(__name__)
@@ -40,6 +41,16 @@ def _prefix_tokens(backend: EmbeddingsBackend, input_type: EmbeddingInputType) -
     attr = "query_prefix" if input_type == "query" else "passage_prefix"
     prefix = getattr(backend, attr, "")
     return count_tokens(prefix) if isinstance(prefix, str) and prefix else 0
+
+
+# A provider that rejects an input for its size answers with one of these. The body is
+# not a reliable signal: some gateways say only "The parameter is invalid".
+_INPUT_SIZE_REJECTION_STATUSES = frozenset({400, 413, 422})
+
+
+def _is_input_size_rejection(exc: Exception) -> bool:
+    """True for a 4xx that plausibly means the input was too large for the provider."""
+    return status_code_of(exc) in _INPUT_SIZE_REJECTION_STATUSES
 
 
 def _truncate_inputs(
@@ -116,13 +127,35 @@ async def generate_embeddings_batch(
     # out of the individual backends so every provider (and every call path — retain,
     # recall queries, consolidation, import) gets identical, model-agnostic truncation.
     max_input_tokens = get_config().embeddings_max_input_tokens
+    original_texts = texts
     if max_input_tokens is not None and texts:
         texts = _truncate_inputs(texts, max_input_tokens, embeddings_backend, input_type)
+    capped = texts != original_texts
 
     try:
         embeddings = await _encode_with_input_type(embeddings_backend, texts, input_type)
     except Exception as e:
-        raise Exception(f"Failed to generate batch embeddings: {str(e)}")
+        if not (capped and max_input_tokens and _is_input_size_rejection(e)):
+            raise Exception(f"Failed to generate batch embeddings: {str(e)}") from e
+        # The cap counts with HINDSIGHT_API_TOKENIZER_ENCODING and the provider counts
+        # with its own, so a text cut to exactly the cap can still arrive over the
+        # provider's limit and come back as a permanent 4xx (#4331). Halving the budget
+        # once costs one extra call on a request that already failed, and needs no
+        # knowledge of the provider's tokenizer.
+        halved_budget = max(max_input_tokens // 2, 1)
+        logger.warning(
+            "Embeddings: provider %s rejected a capped input with a %s; retrying once at "
+            "%d tokens. Set %s below the model's real limit to avoid the extra call.",
+            getattr(embeddings_backend, "provider_name", "?"),
+            status_code_of(e),
+            halved_budget,
+            ENV_EMBEDDINGS_MAX_INPUT_TOKENS,
+        )
+        texts = _truncate_inputs(texts, halved_budget, embeddings_backend, input_type)
+        try:
+            embeddings = await _encode_with_input_type(embeddings_backend, texts, input_type)
+        except Exception as retry_error:
+            raise Exception(f"Failed to generate batch embeddings: {str(retry_error)}") from retry_error
 
     # Guarantee 1:1 alignment with input texts. A silent length mismatch here
     # propagates downstream as zip() drops items, eventually surfacing as an
--- a/hindsight-api-slim/tests/test_embeddings_max_input_tokens.py
+++ b/hindsight-api-slim/tests/test_embeddings_max_input_tokens.py
@@ -172,3 +172,73 @@ class TestFinalPayloadFitsTheCap:
             await embedding_utils.generate_embeddings_batch(backend, [f"User working preferences {content}"])
 
         assert count_tokens(backend.received[0]) <= 8192
+
+
+class _RejectsOversizeBackend(_FakeBackend):
+    """A provider that answers a too-large input with a permanent 4xx.
+
+    Its own tokenizer counts more than ours, so a text cut to exactly the cap
+    still arrives over its limit (#4331). It accepts anything at or below
+    ``accepts_tokens``.
+    """
+
+    def __init__(self, accepts_tokens: int, status_code: int = 400) -> None:
+        super().__init__()
+        self.accepts_tokens = accepts_tokens
+        self.status_code = status_code
+        self.calls: list[list[str]] = []
+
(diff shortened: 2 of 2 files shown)


koriyoshi2041 (approved):
Validated at exact head 883a1f42. The focused cap and token-encoding suites pass 34/34 on macOS/Python 3.11; Ruff check, format check, and diff hygiene also pass. The retry is bounded to one extra attempt, only after local truncation plus a plausible payload-size 4xx, while uncapped inputs and auth failures retain fail-fast behavior. Preserving the provider exception as the cause also keeps status-based classification available. This is a proportional recovery path for the tokenizer-mismatch failure in #4331.

nicoloboschi:
Thanks for the detailed analysis. We went with documentation instead (#4399): the cap is counted in Hindsight's tokens (`HINDSIGHT_API_TOKENIZER_ENCODING`), and there's no way to know every provider's tokenizer. The docs now explain the mismatch and tell users to convert the limit to their provider's tokenizer (e.g. `4096` for an 8192-token model when embedding code).

PR #4333, opened 2026-09-12
fix(memory-defense): avoid redacting technical numbers as credit cards

### LINKED ISSUE #4323: Memory Defense incorrectly redacts scores, timestamps, UUID segments, and port lists as credit cards
# Memory Defense incorrectly redacts scores, timestamps, UUID segments, and port lists as credit cards

## Bug Description

The built-in `MemoryDefenseRegexExtension` classifies ordinary technical data as `credit_card`. With the `sensitive_data` rule set to `redact`, this alters retained document content, not just the audit-log label.

Observed affected data includes fractional search scores, millisecond timestamps, numeric UUID segments, and space-separated port lists. The reproduction below uses only synthetic technical data and requires no LLM, database, or real payment information.

## Steps to Reproduce

Run this in a Python environment with `hindsight-api-slim==0.9.2` installed:

```python
from hindsight_api.extensions.memory_defense import apply_redaction

samples = [
    "score: 0.1234567890123456",
    "timestamp_ms: 1700000000000",
    "task_id: aaaaaaaa-bbbb-cccc-1234-123456789012",
    "ports: 8000 8080 8888 9999",
    "count: 123456",
    "score: 0.123456",
]

for content in samples:
    result = apply_redaction(content)
    print(repr(content), "=>", repr(result.content), result.matched_types)
```

This directly exercises the helper used by the default regex extension. The corresponding bank policy for the observed retention behavior is:

```json
{
  "memory_defense": {
    "enabled": true,
    "rules": [{"on": "sensitive_data", "action": "redact"}]
  }
}
```

## Expected Behavior

These non-payment technical values should remain intact instead of being classified as credit-card numbers. Genuine card detection should remain covered by positive tests.

## Actual Behavior

Verified outputs:

| Input | Output | Matched types |
| --- | --- | --- |
| `score: 0.1234567890123456` | `score: 0.[REDACTED:credit_card]` | `credit_card` |
| `timestamp_ms: 1700000000000` | `timestamp_ms: [REDACTED:credit_card]` | `credit_card` |
| `task_id: aaaaaaaa-bbbb-cccc-1234-123456789012` | `task_id: aaaaaaaa-bbbb-cccc-[REDACTED:credit_card]` | `credit_card` |
| `ports: 8000 8080 8888 9999` | `ports: [REDACTED:credit_card]` | `credit_card` |
| `count: 123456` | unchanged | none |
| `score: 0.123456` | unchanged | none |

In the live deployment, audit events report `Sensitive data pattern matched: credit_card`, and retained documents contain markers such as `0.[REDACTED:credit_card]` in search-score fields. Timestamp and port-list content also contains these markers. Original retention requests were checked to confirm score values were ordinary decimal scores.

## Version

- Installed package: `hindsight-api-slim 0.9.2`
- Container image: `ghcr.io/vectorize-io/hindsight:0.9.2`
- Python: `3.11.16`
- Container OS: Debian GNU/Linux 13 (trixie), `x86_64`
- Deployment: Docker on TrueNAS
- Also checked upstream `main` at `48b62ee08170b464f4c42b6f133b7cb798a59b21`: the entire `memory_defense.py` file is byte-for-byte identical to the deployed file used for this reproduction.
- File SHA-256: `2972984b3e7307878452b5e63ea44da1c53784fbc86f58d0dd55d52863ffc100`

## LLM Provider

Not applicable to this reproduction; the detector runs locally using regular expressions.

## Relevant Implementation and Related Work

The [credit-card pattern](https://github.com/vectorize-io/hindsight/blob/48b62ee08170b464f4c42b6f133b7cb798a59b21/hindsight-api-slim/hindsight_api/extensions/memory_defense.py#L223-L225) is:

```python
("credit_card", _ascii_token_pattern(r"(?:\d{4}[ -]?){3}\d{1,4}"))
```

It matches 13–16 digits, with optional space/hyphen separators. Its ASCII token boundaries permit matches after decimal points and across numeric UUID segments. There is no checksum or semantic-context check. The adjacent comment says 13–19 digits, which also differs from the expression's actual range.

- #4017 requests additive custom patterns per bank; this report concerns false positives in an existing built-in pattern.
- #3574 requests broader international PII coverage.
- PR #3577 proposed international PII detection and some bank-card checksum validation but was closed without merging. This report does not request that broader feature.

All reproduction values above are synthetic; no private conversations, credentials, or actual payment data are included.

kyletser:
## Summary

Fixes #4323.

The default card detector currently treats digit count as sufficient evidence. Under a redact policy, that changes technical data in retained documents; under a block policy, it can reject the item.

This change validates individual credit-card candidates before creating either a replacement or an audit hit:

- Require a Luhn checksum and reject repeated-digit placeholders.
- Exclude decimal components and candidates overlapping a UUID, including numeric UUID segments that happen to pass Luhn.
- Keep card detection beside ordinary punctuation and CJK text. Preserve the existing 13–16 digit candidate formats and other detectors.
- Preserve Unicode decimal-digit recognition, normalize digits for validation, and use consistent Unicode digit boundaries.
- Reconsider overlapping candidate starts after a rejected match instead of consuming the rejected span.
- Exercise the helper, REDACT/BLOCK decisions, and the document-body scrubber with synthetic data.

The UUID check scans only a bounded neighborhood of each candidate. There are no new dependencies, model calls, configuration fields, or API changes.

This remains a heuristic detector: Luhn does not prove a number is an issued card or distinguish every possible technical identifier. This PR addresses the reported examples and explicitly covered boundary cases, not comprehensive international payment-card detection.

## Verification

After the review fixes, on Linux / CPython 3.12.14:

```text
pytest tests/test_memory_defense.py -o addopts= -o log_cli=false -q -rs
360 passed, 14 skipped
```

The 14 existing integration tests require the unavailable local-ML extra. The new tests all run without it. Ruff 0.14.9 check and format-check pass for all three changed files; `git diff --check` passes.

The follow-up adds 44 cases covering REDACT/BLOCK decisions, Unicode and mixed-script digits, consistent decimal/long-number boundaries, repeated placeholders, overlapping starts, multiple hits, and document-body scrubbing. A supplemental runner executes the unchanged credit-card test functions with real detector modules while bypassing unrelated package initializers/conftest: 100 passed; loading the previous PR head gives 29 failures. Independently reverting overlap retry or Unicode matching gives 9 and 18 failures respectively. The full-file Linux run above uses the normal imports and conftest, without those bypasses.

Regression checks were run before the implementation: the original focused run had 12 failures. Temporarily bypassing checksum validation made five regression tests fail again; disabling the UUID guard in a separate process made two UUID regressions fail. Additional positive tests caught an overly broad punctuation guard during development; that guard was replaced before the final run.

Broader validation limitations:

- `pytest tests -x -o addopts= -o log_cli=false -q --timeout=60` stopped at the attachment-attribution LLM test because `HINDSIGHT_API_LLM_API_KEY` is not configured: 68 passed, 74 skipped, 1 failed. This is not a complete API-suite pass.
- `scripts/hooks/lint.sh` cannot run from this Windows CRLF checkout under WSL. The file was not changed; targeted checks were run directly.
- ty 0.0.8 reports one existing `invalid-argument-type` diagnostic in the generic `findall` / fingerprint path. The same diagnostic was reproduced from the unmodified base file at `4cc131c0b238c8f206def60804d7b6591f6a45e7`; that path is unchanged.

## Review path

Start with `test_credit_card_rejects_technical_values` and the positive card matrix, then review the credit-card search loop in `apply_redaction`. The policy and document-body tests verify that the filtered matches reach the existing consumers consistently.

### DIFF
--- a/hindsight-api-slim/hindsight_api/extensions/builtin/memory_defense_regex.py
+++ b/hindsight-api-slim/hindsight_api/extensions/builtin/memory_defense_regex.py
@@ -1,9 +1,10 @@
 """Memory Defense (regex) — the default extension shipping with hindsight-api-slim.
 
 Scrubs known secret/PII patterns from retained content via the
-``sensitive_data`` detector. Matching is pure regex (see ``apply_redaction``):
-no LLM call, no external dependency. A ``sensitive_data`` rule may either
-``redact`` matches in place or ``block`` the item entirely.
+``sensitive_data`` detector. Matching uses regex with a checksum for card
+candidates (see ``apply_redaction``): no LLM call, no external dependency.
+A ``sensitive_data`` rule may either ``redact`` matches in place or ``block``
+the item entirely.
 """
 
 from __future__ import annotations
--- a/hindsight-api-slim/hindsight_api/extensions/memory_defense.py
+++ b/hindsight-api-slim/hindsight_api/extensions/memory_defense.py
@@ -220,12 +220,13 @@ _REDACTION_PATTERNS: list[tuple[str, str]] = [
     ("private_key_pem", r"-----BEGIN (?:RSA |EC |DSA |OPENSSH |PGP )?PRIVATE KEY( BLOCK)?-----"),
     ("jwt", _ascii_token_pattern(r"eyJ[A-Za-z0-9_-]{10,}\.eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}")),
     # --- PII (US-centric defaults; can be tuned per deployment) ---
-    # Keep the existing 13-16 digit formats, but exclude parts of decimals.
+    # Keep the existing 13-16 Unicode decimal-digit formats, but exclude parts
+    # of decimals and longer digit runs with the same Unicode digit semantics.
     # Candidates also need a Luhn check below: digit count alone redacted
     # timestamps and port lists as if they were cards.
     (
         "credit_card",
-        _ascii_token_pattern(r"(?<![0-9]\.)(?:[0-9]{4}[ -]?){3}[0-9]{1,4}(?!\.[0-9])"),
+        _ascii_token_pattern(r"(?<!\d)(?<!\d\.)(?:\d{4}[ -]?){3}\d{1,4}(?!\d)(?!\.\d)"),
     ),
     ("ssn_us", _ascii_token_pattern(r"\d{3}-\d{2}-\d{4}")),
 ]
@@ -238,14 +239,14 @@ _UUID_PATTERN = re.compile(_ascii_token_pattern(r"[0-9a-fA-F]{8}-(?:[0-9a-fA-F]{
 
 
 def _is_luhn_card(value: str) -> bool:
-    """Validate an ASCII digit candidate already bounded by the card regex."""
-    digits = value.replace(" ", "").replace("-", "")
+    """Validate a decimal-digit candidate already bounded by the card regex."""
+    # int() normalizes Unicode decimal digits, including mixed-script placeholders.
+    digits = [int(char) for char in value if char not in " -"]
     # Repeated-digit placeholders (including all zeroes) can pass Luhn.
     if len(set(digits)) == 1:
         return False
     checksum = 0
-    for index, char in enumerate(reversed(digits)):
-        digit = int(char)
+    for index, digit in enumerate(reversed(digits)):
         if index % 2:
             digit *= 2
             if digit > 9:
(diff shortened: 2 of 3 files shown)


ekinnee:
@kyletser Thanks for the PR! I opened the Issue and waited in case there was maintainer feedback. One less thing for me to do now. :D

ekinnee:
Reviewed `bbd51891be17970b8e5f241b80b65c5fed68447b`. The reported false-positive examples are addressed, but I found two reproducible detection regressions worth fixing before merge. All examples below are synthetic test data.

### 1. [P2] Rejected candidates consume overlapping valid card starts

With input:

```text
2026 4111 1111 1111 1111
```

The regex first consumes `2026 4111 1111 1111`. Its checksum fails, so the callback returns that span unchanged. `pattern.sub()` then resumes after the consumed span and never considers the valid overlapping card `4111 1111 1111 1111`.

Observed behavior:

| Revision | REDACT policy | BLOCK policy |
| --- | --- | --- |
| Base `4cc131c` | REDACT; output `[REDACTED:credit_card] 1111` | BLOCK |
| PR head `bbd5189` | ALLOW; entire input unchanged | ALLOW |

At head, both `matched_types` and `hits` are empty. This can occur when a four-digit year or another numeric field precedes a spaced card in flattened text. The base did not identify the exact card span correctly either, but it prevented the complete number from passing through; the new validation leaves it entirely intact.

Please allow candidate discovery to reconsider overlapping starts after a failed checksum, and add REDACT/BLOCK coverage for this case.

[Callback and substitution](https://github.com/vectorize-io/hindsight/blob/bbd51891be17970b8e5f241b80b65c5fed68447b/hindsight-api-slim/hindsight_api/extensions/memory_defense.py#L279-L296)

### 2. [P2] ASCII-only digit matching removes existing Unicode decimal-digit coverage

Changing `\d` to `[0-9]` means these representations of the same synthetic test card are no longer detected:

```text
card: ４１１１１１１１１１１１１１１１
card: ٤١١١١١١١١١١١١١١١
```

The base redacts both under REDACT and blocks both under BLOCK. At the PR head, both policies return ALLOW, with unchanged content and no hits. This is a loss of existing behavior rather than a request for additional international card formats.

Please preserve Unicode decimal-digit recognition and normalize candidates for checksum validation, with positive regression tests for these representations. Decimal-boundary checks should use consistent digit semantics too.

[Changed candidate pattern](https://github.com/vectorize-io/hindsight/blob/bbd51891be17970b8e5f241b80b65c5fed68447b/hindsight-api-slim/hindsight_api/extensions/memory_defense.py#L228)

### Validation scope

I compared the base and head detector implementations and exercised the actual regex-extension `screen()` method under both policies in an isolated Python harness. Only the unrelated base `Extension` dependency was stubbed; matching, checksums, fingerprinting, policy parsing, and decision logic used the downloaded source. These probes did not persist data or exercise the HTTP/database retain pipeline.

144 isolated checks passed across the PR's positive/negative examples and the existing detector sample catalog, in addition to the regression probes above. I did not rerun the full integration suite or repository lint script.

kyletser:
Thanks for the detailed reproductions — both reports were correct. Fixed in e922905339959c7f82796a7670fbded4bcd22999.

- Rejected candidates now resume searching at the next character after their start, rather than consuming the entire span. `2026 4111 1111 1111 1111` now becomes `2026 [REDACTED:credit_card]` under REDACT and returns BLOCK under BLOCK.
- Restored Unicode decimal-digit matching and normalized the digits before checksum/placeholder validation. Fullwidth, Arabic-Indic, and mixed-script examples are covered. Decimal and longer-number boundaries use the same digit semantics; audit fingerprints still use the original text.

Added 44 regression cases covering both policies, overlapping starts, multiple hits, Unicode non-card values, and document-body scrubbing. On Linux/CPython 3.12.14, the complete `tests/test_memory_defense.py` file reports **360 passed, 14 skipped** (the existing local-ML-dependent integration tests). Ruff check/format and `git diff --check` pass.

The initial six detector probes failed before the fix and passed afterward. Supplementary mutation runs also fail when either overlap retry or Unicode matching is removed. The PR description has been updated with the verification details and limits: the full API run still stops at an unrelated missing LLM API key, and the existing CRLF lint-hook/type-check limitations remain disclosed.

nicoloboschi (approved):

PR #4342, opened 2026-09-12
fix(api): reject request bodies before clearing bank memories

### LINKED ISSUE #4337: DELETE /v1/default/banks/{bank_id}/memories silently ignores the request body and wipes the whole bank (200 + total data loss)
## What happened

`DELETE /v1/default/banks/{bank_id}/memories` is `clear_bank_memories`: its only non-path parameter is the optional query `type`. A JSON body is **silently ignored**, so a client that sends a body full of ids — the natural reading of a collection-level DELETE — gets `200 {"success": true}` and loses the **entire bank** (every fact type, all documents, all links).

Our case: **35,092 memory units / 1,279 documents / ~523k links** deleted by a single request that reported success. Because `delete_bank_profile=False`, the `banks` row survives, so afterwards the bank still carries the same `bank_id` and `created_at` and nothing looks new.

## Why the API shape invites it (three compounding signals)

1. **OpenAPI advertises no request body** for this operation: `requestBody: ABSENT`, parameters are only `bank_id` (path), `type` (query), `authorization` (header), `operationId: clear_bank_memories`. A client inspecting the schema cannot tell the body is *ignored* rather than *filtered*.
2. `DELETE /memories/{memory_id}` has only GET/PATCH → **405**, so a caller looking for a targeted delete has no working single-item route and naturally moves to the collection route (same feedback gap as #4307 "no `deleted_count`").
3. The response carries no `deleted_count` and no scope echo, so from the client side "the two things I asked for are gone" and "everything is gone" are indistinguishable — my cleanup script printed *"remaining: 0"* and I believed it.

## Reproduction (0.9.2)

```bash
BANK=scratch-bank
curl -X POST "$HOST/v1/default/banks/$BANK/memories" -H "Authorization: Bearer $KEY" \
     -d '{"items":[{"content":"fact A"},{"content":"fact B"}]}'

# intent: delete specific facts by id
curl -X DELETE "$HOST/v1/default/banks/$BANK/memories" -H "Authorization: Bearer $KEY" \
     -d '{"ids":["00000000-0000-0000-0000-000000000000"]}'
# -> 200 {"success":true}   ... and the bank is now empty
```

## Suggested fixes (any one would have prevented this)

- Reject a non-empty body on this route (`400`), or honour `ids` as a filter.
- Return `deleted_count` plus the resolved scope (`fact_type` / `ids`) in the response.
- Make the destructive scope explicit: require `type`, or a `confirm=<bank_id>` query param, and/or move the whole-bank clear to its own path (e.g. `POST /banks/{id}/clear`) so that `DELETE` on the collection route can safely mean "filtered bulk delete".
- Consider defaulting `@audited("clear_memories")` on (`audit_log_enabled`) — we had audit logging off, so the wipe left no trace whatsoever.
- The bundled control-plane UI ships a one-click "clear all memories" that hits this same route; that button deserves at least a type-the-bank-name confirmation.

## Immediate host-level mitigation we applied

A Caddy matcher in front of the API returns `403` for `DELETE` on `/v1/default/banks/*/memories`, `/observations`, and on the bank itself, for every external client *and* for the control plane (which we re-pointed through the same guarded listener). A genuine whole-bank clear now has to run on the host via `hindsight-admin`.

Environment: hindsight-api **0.9.2**, self-hosted, Postgres 18 + pgvector, request sent with `Content-Type: application/json` and body `{"ids":[...]}`.

r266-tech:
A caller sending `DELETE /banks/{bank_id}/memories` with an `ids` body currently gets 200 and clears the entire bank. Reject the first nonempty request-body chunk with 400 before calling the engine. This also covers chunked, malformed and non-JSON bodies without buffering the full payload.

Bodyless calls retain the existing `type` query filter and clear behavior. Clients sending placeholder bodies such as `{}` must omit them. This adds no bulk-delete API; #4309 and #4312 address deletion counts separately.

Fixes #4337.

Validation:
- 18 real-route cases pass; before the fix, all 15 unsafe-body cases return 200, while the three valid bodyless calls pass.
- One SDK-driven story against a real server/PostgreSQL database reproduces the original 200, then verifies 400, unchanged memory IDs/text and documents, typed clearing, full clearing and the surviving bank. LLM responses are stubbed.
- Full repository lint and documentation build pass. OpenAPI was regenerated with no changes; the SDK surface is unchanged.

AI-assisted implementation and self-review.

### DIFF
--- a/hindsight-api-slim/hindsight_api/api/http.py
+++ b/hindsight-api-slim/hindsight_api/api/http.py
@@ -9414,11 +9414,21 @@ def _register_routes(app: FastAPI):
     @audited("clear_memories", request_param=None)
     async def api_clear_bank_memories(
         bank_id: str,
+        request: Request,
         type: str | None = Query(None, description="Optional fact type filter (world, experience, observation)"),
         request_context: RequestContext = Depends(get_request_context),
     ):
         """Clear memories for a memory bank, optionally filtered by type."""
         try:
+            # A selection body must not silently become a whole-bank delete.
+            # Inspect actual bytes (including chunked bodies) without buffering them.
+            async for chunk in request.stream():
+                if chunk:
+                    raise HTTPException(
+                        status_code=400,
+                        detail="This endpoint clears bank memories and does not accept a request body. "
+                        "Use the type query parameter to filter by fact type.",
+                    )
             await app.state.memory.delete_bank(
                 bank_id, fact_type=type, delete_bank_profile=False, request_context=request_context
             )
--- a/hindsight-api-slim/tests/test_clear_memories_body.py
+++ b/hindsight-api-slim/tests/test_clear_memories_body.py
@@ -1,0 +1,62 @@
+"""A selection body must never turn into an unfiltered bank clear (#4337)."""
+
+from collections.abc import AsyncIterator
+from unittest.mock import AsyncMock, MagicMock
+
+import httpx
+import pytest
+from fastapi import FastAPI
+
+from hindsight_api.api import create_app
+from hindsight_api.engine.memory_engine import MemoryEngine
+
+
+@pytest.fixture
+def app() -> FastAPI:
+    memory = MagicMock(spec=MemoryEngine)
+    memory.audit_logger = None
+    memory.delete_bank = AsyncMock()
+    return create_app(memory, initialize_memory=False)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("body", [b'{"ids":["one"]}', b"{}", b"[]", b"null", b" ", b"invalid", b"\xff"])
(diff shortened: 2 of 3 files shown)


koriyoshi2041 (approved):
Verified at `32d33c8c68598170b7ae4cf5500f9b6160c4ff28`. The route rejects the first nonempty chunk before `delete_bank`, while bodyless and `type`-filtered clears keep their existing behavior. The focused route suite passes 18/18 locally on Python 3.11, including streamed, malformed, non-JSON, placeholder, and empty-body cases; the applicable hosted API/system/OpenAPI gates are green.

nicoloboschi:
either delete the documents or invalidate the memories via their endpoints

PR #4346, opened 2026-09-13
fix(dsh): register tools before the first request

### LINKED ISSUE #4317: Tools register inside an async ctx.inject callback, so they appear mid-session and invalidate the prompt-prefix cache
## Summary

The DSH integration registers its tools from inside an asynchronous `ctx.inject(["tools"], ...)` callback. The callback resolves after the session has already started, so the model's `tools` array **grows in the middle of a live conversation**. Because the `tools` array is part of the request prefix for provider-side prompt caching, the late registration forces a full prefix re-send (priced as cache-miss) for that session -- on the official DeepSeek API a cache-miss token costs **50x** a cache-hit token.

## Reproduction

1. Install `@vectorize-io/hindsight-coding-agents@0.5.3` in a DSH profile and start a new session.
2. Send a first message that does not use any `hindsight_*` tool.
3. In the session log (`~/.dsh/sessions/<proj>/<session>/session.v3.jsonl.zstd`, zstd multi-frame, first frame is the header), compare `request/header` events in order:

```
seq 1912->1992   tools 91->115   + hindsight_capture_initiative, hindsight_diagnose, hindsight_ingest_document,
                                  hindsight_list_knowledge_pages, hindsight_read_knowledge_page, hindsight_reflect,
                                  hindsight_search_knowledge_pages, hindsight_sync_status
```

## Current behavior

- The first requests of a session carry **no** `hindsight_*` tools; after the `inject` callback resolves all 8 appear at once.
- Measured cost of a single such transition: prompt **671,916 tokens** with **4,096** cache-hit tokens (**0.6%** hit rate) -> effectively a full-price re-send.
- The same pattern appears in other plugins (`dsh-taskboard/lib/index.js:31,37,47`; kernel `dsh-tool-subagent/lib/index.js:624-630,649-651`), so it looks like an easy mistake rather than an intentional design.

## Expected behavior

Register all tools **at load time** (synchronously during plugin mount), so a session's tool set is complete and stable from its first request. If a dependency genuinely cannot be resolved before session start, register the tool **once up front** and reject its calls with a clear error until the dependency is ready, instead of adding the tool later.

## Environment

- `@vectorize-io/hindsight-coding-agents@0.5.3`, DSH `0.1.5-rc.1`, Windows, `dsh web` profile
- Provider: `deepseek-official` / `deepseek-v4-flash`; cache-hit price is 1/50 of cache-miss
- Session log format: zstd multi-frame JSONL, first frame is exactly one header line

## Exact code location (from the installed package)

```js
// @vectorize-io/hindsight-coding-agents/dist/dsh.js
:17885-17887   ctx.inject(["tools"], (...) => { ... registerTools ... })
```

Moving `registerTools` out of the async `inject` callback and into load-time mount resolves this.

koriyoshi2041:
## Summary

- declare the DSH tool registry as a mount-time plugin dependency
- register the Hindsight tool suite directly during plugin mount
- cover the stable dependency contract with a focused regression test

This prevents the tool catalog from growing after a session has already started, which invalidates provider prompt-prefix caches.

Fixes #4317.

## Tests

- `npm test -- --run src/dsh.test.ts` (15 passed)
- `npm run build`
- `./scripts/hooks/lint.sh`
- full coding-agents unit suite: 1006 passed, 12 existing environment-sensitive git-layout tests failed because the parent home directory is itself a git repository; the focused DSH suite is green

### DIFF
--- a/hindsight-integrations/coding-agents/src/dsh.ts
+++ b/hindsight-integrations/coding-agents/src/dsh.ts
@@ -35,8 +35,8 @@ const HARNESS = "dsh";
 /** Cordis plugin name (loader diagnostics, and the `source.plugin` on everything we inject). */
 export const name = HINDSIGHT_PLUGIN;
 
-/** The agent registry owns the lifecycle events below; without it there is nothing to bind. */
-export const inject = ["agents"];
+/** Both services must exist before mount so the first model request sees the final tool catalog. */
+export const inject = ["agents", "tools"];
 
 // ── host shapes (structural: this package must not depend on dsh's packages) ─────
 
@@ -74,6 +74,7 @@ interface PreStepPayload {
 
 /** The Cordis context surface this plugin uses. */
 interface DshContext {
+  tools: DshToolContext["tools"];
   on(
     event: "agent/session-start",
     listener: (payload: { agent: DshAgent }) => void
@@ -94,7 +95,6 @@ interface DshContext {
     event: "agent/disposed",
     listener: (payload: { agent: DshAgent }) => void
   ): (() => void) | void;
-  inject(names: string[], callback: (ctx: DshToolContext) => void): void;
 }
 
 interface DshToolContext {
@@ -408,11 +408,10 @@ export function apply(ctx: DshContext): void {
   ctx.on("agent/pre-step", hooks.preStep, { prepend: true });
   ctx.on("agent/turn-stopping", hooks.turnStopping);
   ctx.on("agent/disposed", hooks.disposed);
-  // The tools registry is optional: a composition without it (a bare headless assembly) still gets
-  // recall, injection and write-back.
-  ctx.inject(["tools"], (toolCtx) => {
-    registerTools(toolCtx, process.cwd());
-  });
+  // `tools` is a static plugin dependency, so Cordis does not mount us until the registry exists.
+  // Registering here keeps the catalog stable from the first request instead of growing it later
+  // when a nested dependency callback happens to resolve.
+  registerTools(ctx, process.cwd());
 }
 
 export default { name, inject, apply };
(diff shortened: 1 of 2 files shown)


hindsight-integrations/coding-agents/src/dsh.ts
```diff
@@ -35,8 +35,8 @@ const HARNESS = "dsh";
 /** Cordis plugin name (loader diagnostics, and the `source.plugin` on everything we inject). */
 export const name = HINDSIGHT_PLUGIN;
 
-/** The agent registry owns the lifecycle events below; without it there is nothing to bind. */
-export const inject = ["agents"];
+/** Both services must exist before mount so the first model request sees the final tool catalog. */
+export const inject = ["agents", "tools"];
```
ebarkhordar, line 39: I read this at `e60451ff9e61bb216d9f653389a5803f63906265` and went looking for the composition the deleted comment names.

The old line said the tools registry was optional so that "a composition without it (a bare headless assembly) still gets recall, injection and write-back". On an unshallowed clone, `git log -G 'The tools registry is optional' --follow` puts it in `28760f62d` (#3504), the commit that added dsh support, so it is the shape this plugin shipped with rather than something that drifted in.

Moving `tools` into `inject` makes it a mount gate, which your new comment says plainly: Cordis does not mount the plugin until the registry exists. In a composition that never registers one, what goes away is not just the tool suite. `apply` is where `agent/pre-step`, `agent/session-start` and `agent/turn-stopping` get bound, so an unmounted plugin loses the recall and write-back the old comment was protecting.

What I could not settle from this repo: whether a live composition actually lacks `tools`. The one that made me ask is `src/e2e/harnesses.ts:278`, where the dsh e2e drives `dsh --profile headless`, and that profile's bundle lives in the dsh package rather than here, so I could not read it. I did not run the e2e.

So: is the `headless` profile known to carry a tools registry? If it is, this is a non-issue and only the comment was stale. If it is not, prompt-prefix cache stability is being bought with recall on that profile, which seems like the wrong side of the trade.

One thing I did not verify, since I could not see which Cordis version dsh pins: if the split matters, `inject: { required: ["agents"], optional: ["tools"] }` is the form that mounts either way while still registering at mount time when the registry is there.

PR #4347, opened 2026-09-13
fix(embed): pin the probe's decode instead of taking the locale's

### LINKED ISSUE #4344: [Bug] #2258 follow-up: _run_probe() reads netstat/PowerShell output without encoding= (reader thread dies on cp936 consoles, listening-PID detection silently returns [])
### Bug Description

`DaemonEmbedManager._run_probe()` calls `subprocess.run()` with `text=True` but no `encoding=` and no `errors=`:

```python
# hindsight-embed/hindsight_embed/daemon_embed_manager.py
#   main  : def _run_probe (395), subprocess.run( (398), text=True (401)
#   0.9.2 : def _run_probe (402), subprocess.run( (405), text=True (408)
result = subprocess.run(
    cmd,
    capture_output=True,
    text=True,
    timeout=timeout,
    creationflags=getattr(subprocess, "CREATE_NO_WINDOW", 0),
)
```

On Windows, `text=True` without `encoding=` decodes the child's stdout with `locale.getpreferredencoding(False)`. The commands this helper runs are Windows-native CLIs — `netstat`, `powershell`, `wmic` — which emit localized text in the console code page. On a zh-CN host that is cp936/GBK, so the decode raises `UnicodeDecodeError` inside `subprocess._readerthread`.

The failure does **not** surface as an exception. The reader thread dies, its traceback is printed to stderr by `threading`'s excepthook, and `subprocess.run()` returns normally with `stdout is None`. `_run_probe()` then returns `None` (the return code is 0), and `_windows_listening_pids()` takes its `if output is None: return []` branch — so on such a host the manager concludes that **no process is listening on any port**. Silently, with nothing raised anywhere.

This is the same defect class as #3747 / #3837, but in the `subprocess` text-mode path, which those fixes did not cover (they covered `open()` without `encoding=` in `profile_manager.py` and `cli.py`). The `subprocess.run` kwargs block above is the same one #2258 touched to add `CREATE_NO_WINDOW`; the encoding parameter was not added then. Both v0.9.2 (latest release) and `main` currently lack it.

### Steps to Reproduce

Environment: Windows 11 (build 26200), system locale zh-CN (console code page 936), Python 3.11.16.

1. Make the host process run in UTF-8 mode — `PYTHONUTF8=1`. This is what makes `locale.getpreferredencoding(False)` return `utf-8` instead of `cp936` (see the additional context below). Hermes Agent, which embeds `hindsight-embed` for its hindsight memory provider, sets this.

2. Reduce the probe to its essential call:

```python
import subprocess
r = subprocess.run(["netstat", "-ano", "-p", "TCP"], capture_output=True, text=True)
print(r.stdout)
```

3. Observe. The raw bytes of that command on this host are:

```
>>> subprocess.run(["netstat", "-ano", "-p", "TCP"], capture_output=True).stdout[:8]
b'\r\n\xbb\xee\xb6\xaf\xc1\xac'
>>> subprocess.run(["netstat", "-ano", "-p", "TCP"], capture_output=True).stdout[:8].decode("cp936")
'\r\n活动连接'
```

Byte 2 is `0xbb`, the lead byte of "活" — the first character of the localized header line. That is exactly the byte and position reported in the crash:

```
Exception in thread Thread-1 (_readerthread):
  File "...\Lib\subprocess.py", line 1599, in _readerthread
    buffer.append(fh.read())
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbb in position 2: invalid start byte
```

4. Or via the real path: `pip install hindsight-all==0.9.2`, then trigger any operation in which the daemon manager checks the daemon port (`_ensure_started()`, `is_running()`, `stop()`). The traceback appears on stderr and in the host application's log, while the operation itself reports success.

### Expected Behavior

The probe reads the child's output deterministically, whatever the console code page:

```python
result = subprocess.run(
    cmd,
    capture_output=True,
    text=True,
    encoding="utf-8",
    errors="replace",
    timeout=timeout,
    creationflags=getattr(subprocess, "CREATE_NO_WINDOW", 0),
)
```

With that form, `_windows_listening_pids()` returns the real listener PIDs using the **existing parser unchanged**, and `_process_command_line()` returns the real command line.

### Actual Behavior

- `stdout is None` in **12 of 12** consecutive runs of the reduced call; return code 0, no exception raised in the caller.
- `_windows_listening_pids()` returns `[]` for every port. At measurement time this host had **19** TCP listeners (135, 139, 445, 5432, 8642, 27036, 28252, 49401, 49664-49668, ...); the manager saw none of them.
- With `encoding="utf-8", errors="replace"` the same command returns 6,987 chars and the existing parser yields all **19** PIDs — the same set as decoding the raw bytes as cp936. No parser change is needed: only the localized header contains non-ASCII, and every data line is pure ASCII.

The two parameters are not interchangeable — `encoding="utf-8"` **alone still crashes** (the stream genuinely is not UTF-8), which is why `errors="replace"` is required as well.

### Version

hindsight-embed 0.9.2, installed from PyPI; this is the latest release at the time of writing (v0.9.2, 2026-08-25). Also verified against `main` on 2026-09-12: `_run_probe()` there is identical apart from line offsets — still no `encoding=`, no `errors=`.

### LLM Provider

N/A — the crash occurs in the daemon manager's local port/process probe, before any LLM call.

### Additional context: the UTF-8-mode dependency (why this is easy to miss)

The crash only occurs when the host process runs in UTF-8 mode. Measured on the same machine, same command:

| Host environment | `locale.getpreferredencoding(False)` | Result of the reduced call |
|---|---|---|
| `PYTHONUTF8=1` | `utf-8` | `stdout=None` — crash |
| `PYTHONUTF8` unset | `cp936` | 6,966 chars — works |
| `PYTHONUTF8=0` | `cp936` | 6,966 chars — works |

Without UTF-8 mode the locale default happens to match what `netstat` emits, so the bug is invisible — and it only appears once the host turns UTF-8 mode on, which is the correct thing for a host to do. The same machine therefore reproduces or does not reproduce depending on how the process was launched. This is likely why it has gone unnoticed alongside #3747 / #3837.

It is also why the fix has to be explicit rather than "leave the default alone": the default is wrong in both directions depending on the host — cf. #3747 / #3837, where the file was UTF-8 and the cp936 default was the wrong one.

### Affected scope

Any host satisfying both of:

1. the console/system code page is not UTF-8 (zh-CN 936, ja-JP 932, ko-KR 949, ru-RU 866, ...), and
2. the process running `hindsight-embed` is in UTF-8 mode (`PYTHONUTF8=1` / `-X utf8`), as Hermes Agent is.

On such hosts `_windows_listening_pids()` is blind, and `_process_command_line()` returns `None`. The comment already present on the `_process_command_line()` call site describes the consequence of the latter: "the caller reads None as 'not one of our processes' (see `_process_matches`), so the daemon manager refuses to stop a daemon it does in fact own." I have not observed that end state directly — it follows from the return-value path and that comment, and I am flagging it as inference rather than measurement.

L4XB:
Fixes #4344.

## The change

```python
result = subprocess.run(
    cmd,
    capture_output=True,
    text=True,
    encoding="utf-8",
    errors="replace",
    timeout=timeout,
    creationflags=getattr(subprocess, "CREATE_NO_WINDOW", 0),
)
```

Your diagnosis held up in full, including the part that makes it invisible: the `UnicodeDecodeError` is raised inside `subprocess`'s own reader thread, so nothing in `_run_probe` can catch it. `run()` returns normally with `stdout=None` and returncode 0, and `_windows_listening_pids()` takes its `if output is None: return []` branch. No parser change was needed.

## Testing

Three cases in `hindsight-embed/tests/test_embed_manager.py`, all portable — they drive `_run_probe` with a child that writes the exact cp936 bytes, so they exercise the decode on any host whose preferred encoding is UTF-8, which is the condition in your report:

- `test_run_probe_reads_output_that_is_not_utf8` — a localized `netstat -ano -p TCP` answer (cp936 header, ASCII data lines) comes back with `LISTENING`, the address and the PID intact.
- `test_windows_listening_pids_parses_a_localized_netstat` — the existing parser, unchanged, yields `[4242]` for 8642 and `[4]` for 445 from that same payload, and `[]` for a port with no listener. Your claim that only the header carries non-ASCII, asserted rather than taken on trust.
- `test_run_probe_still_reports_a_failed_command_as_none` — a non-zero exit is still `None`, so pinning the decode did not widen what counts as success.

Negative controls, one per parameter, confirming they are not interchangeable as you noted:

| mutation | result |
|---|---|
| remove `encoding=` and `errors=` | `test_run_probe_reads_output_that_is_not_utf8` red |
| remove `errors="replace"` only | same case red |

```
pytest hindsight-embed/tests/test_embed_manager.py     30 passed
pytest hindsight-embed/tests/                          219 passed, 3 failed
```

The three failures are in `test_profile_daemon_config.py` and fail on a stashed clean tree here as well (they assert Windows/POSIX launcher paths). Same for the one `ruff check` error, which is `E721` on `test_embed_manager.py:58`, a pre-existing line this branch does not touch.

## Scope

`daemon_client.py:209` has the same `text=True` without `encoding=`, and I left it: it runs `bash -c "curl … | bash"`, whose output is POSIX and only printed, so neither half of the condition applies. Say the word if you would rather have it pinned anyway for uniformity.


koriyoshi2041 (approved):
Verified at 8a3aaa29. The raw cp936 regression exercises the actual subprocess reader-thread failure, and the existing listener parser still recovers both PIDs after replacement decoding. Focused embed-manager tests pass 30/30 locally; the full embed suite is 219 passed with the same three environment-sensitive launcher failures documented on the PR. The change preserves non-zero exit handling and limits replacement damage to text the current ASCII-based probes do not consume.

nicoloboschi (approved):

Issue #1726, opened 2026-05-25
fix(openclaw): SessionEnd does not flush un-retained turns — short conversations lose all memory


eightHundreds:
## Summary

Issue #1145 fixed the `hindsight-memory` (Python/Claude Code) plugin so that `session_end.py` flushes any un-retained turns before stopping the daemon. The **OpenClaw plugin** (TypeScript) has the same root cause but was **not included in that fix**: when `retainEveryNTurns > 1` and a conversation ends before the next cadence boundary, all remaining turns are silently dropped.

## Reproduction

1. Configure an OpenClaw integration with `retainEveryNTurns: 5`.
2. Have a conversation of only 2 turns, then end the session.
3. Check the bank — no documents were retained for this session.

Expected: turns 1–2 are retained on session end.
Actual: zero retains fired. Turns 1 and 2 are permanently lost.

## Root Cause

In `hindsight-integrations/openclaw/src/index.ts`, the `agent_end` handler (line ~2554) is the sole entry point for retention:

```ts
if (turnCount % retainEveryN !== 0) {
    debug(`[Hindsight Hook] Turn ${turnCount}/${retainEveryN}, skipping retain...`);
    return;  // early return, no retain at all
}
```

The turn counter is incremented and preserved in an in-memory `Map` (`turnCountBySession`), but:

1. **No `session_end` event handler exists** — grep for `session_end`/`sessionEnd`/`SessionEnd` returns zero hits in the entire OpenClaw source.
2. **No final-flush mechanism** — no `forceRetain`, `flushRemaining`, or equivalent logic anywhere.
3. The retain queue (JSONL-backed `retainQueue`) only retries **failed API calls**, not cadence-skipped turns.

When the conversation ends, the in-memory turn counters are simply discarded along with the un-retained message content.

## Impact

- Any conversation shorter than `retainEveryNTurns` (e.g., quick 2–3 turn Q&A with `retainEveryNTurns: 10`) produces **zero** memory retention.
- Even longer conversations lose the tail between the last cadence boundary and session end (e.g., 27 turns with `retainEveryNTurns: 10` loses turns 21–27).
- Users have no visibility into this data loss — there is no warning or fallback.

## Proposed Fix

Mirror the approach from #1145's fix in `session_end.py`: register a session-end hook that performs a final retain, bypassing the `retainEveryNTurns` cadence check.

Candidate shape in `index.ts`:

```ts
// Register a session-end handler that flushes remaining turns
[...]

PR #4348, opened 2026-09-13
fix(openclaw): read the transcript session_end does not carry

### LINKED ISSUE #4341: session_end force-flush (#1726) is dead code: OpenClaw session_end events carry no messages
## Bug Description

The `session_end` force-flush introduced for #1726 in `@vectorize-io/hindsight-openclaw` 0.12.0 never fires, so with `retainEveryNTurns > 1` the un-retained tail of every conversation (the 1–2 turns after the last cadence boundary) is silently dropped when a session ends.

The flush is dead code because of a payload mismatch between the plugin and OpenClaw:

- The plugin's `runRetain()` (dist/index.js, shared `agent_end`/`session_end` path) requires messages and bails otherwise:

  ```js
  if (!Array.isArray(event.context?.sessionEntry?.messages ?? event.messages) ||
      (event.context?.sessionEntry?.messages ?? event.messages ?? []).length === 0) {
      debug("[Hindsight Hook] No messages in event, skipping retention");
      return;
  }
  ```

- But OpenClaw's `buildSessionEndHookPayload()` (dist/active-sessions-shutdown-tracker-*.mjs, src/auto-reply/reply/session-hooks.ts) builds the `session_end` event with **only** `sessionId`, `sessionKey`, `messageCount`, `durationMs`, `reason`, `sessionFile`, `transcriptArchived`, `nextSessionId`, `nextSessionKey` — **no `messages` array, and `context` carries only `sessionId`/`sessionKey`/`agentId`**.

So for every `session_end` (`new`, `reset`, `idle`, `daily`, `compaction`, `deleted`, `shutdown`, `restart`) the guard above always triggers and the forced flush of the un-retained tail never happens.

## Steps to Reproduce

1. Run `@vectorize-io/hindsight-openclaw` 0.12.0 on a current OpenClaw gateway (verified on 2026.9.4) with `retainEveryNTurns: 3`, `retainOverlapTurns: 1`, `autoRetain: true`.
2. Have a conversation longer than the cadence boundary, e.g. 4–5 turns (cadence retain fires at turn 3).
3. End the session: run `/new` or `/reset` (verified from Telegram, WebChat and the Control UI; also applies to idle/daily resets and gateway shutdown).
4. Inspect the bank's operations (e.g. `GET /v1/default/banks/{bank_id}/operations` or `hindsight operation list <bank>`) and filter for the session's document id.

Observed: retain/batch_retain operations exist only for `agent_end` cadence boundaries. **Zero retain operations are created at or after the `session_end` timestamp** — the turns after the last cadence boundary never reach the bank.

Example (2026-09-12): session `agent:main:telegram:group:...`:topic:44` — last cadence retain operation created 20:09:02Z; `/reset` at 20:21:12Z; no retain operation after that. Gateway journal confirms the periodic plugin stats show 0 retains in the windows covering both reset timestamps.

## Expected Behavior

On `session_end`, the plugin force-flushes the un-retained tail: a window of `turnCount % retainEveryNTurns` un-retained turns + `retainOverlapTurns`, as implemented in the `force` branch of `runRetain()` (the #1726 fix). Short sessions and the tail of long sessions land in the bank when the session closes.

## Actual Behavior

The `session_end` handler exits early with `"No messages in event, skipping retention"` (debug level). Nothing is flushed; the tail turns are only recoverable manually (e.g. via the backfill CLI).

## Suggested Fix

On `session_end`, load the transcript from `event.sessionFile` and pass it as the messages array into `runRetain()` (or give `runRetain` a `sessionFile` fallback when the event carries no messages). OpenClaw already hands the plugin everything needed:

- `event.sessionFile` (path to the transcript; `resolveStableSessionEndTranscript` also resolves archived transcripts, so `transcriptArchived: true` cases work),
- `event.messageCount` / `event.reason` for logging.

The shutdown drain shares a 2-second budget across all sessions, so the handler should only read + POST (extraction already runs asynchronously in the Hindsight bank operation queue) — with that constraint the fix fits the drain budget.

Side note: for `/new`, OpenClaw emits the previous session's `session_end` lazily on the first turn of the successor session (`isFirstSessionTurn` in get-reply) — the fix must resolve the transcript from `sessionFile` at handler time (which it would), since the old messages are no longer in the live session entry.

## Version

- `@vectorize-io/hindsight-openclaw`: 0.12.0 (latest as of 2026-09-12)
- OpenClaw: 2026.9.4 (3a9d69d)
- Hindsight API: 0.9.2
- OS: Linux (Debian 13), Node v24.18.1

## LLM Provider

Other (extraction runs via OpenRouter)

L4XB:
Fixes #4341.

## The change

The forced flush now reads the transcript the event points at, on the one path that needs it:

```ts
let eventMessages = event.context?.sessionEntry?.messages ?? event.messages;
if (force && (!Array.isArray(eventMessages) || eventMessages.length === 0)) {
  eventMessages = sessionEndMessagesFromTranscript(event);
}
```

Your diagnosis held in full. `buildSessionEndHookPayload()` carries ids and counts and no `messages`, so the guard below that line ended every session-close flush before it began, and with `retainEveryNTurns > 1` the tail after the last cadence boundary was dropped on every reason — `new`, `reset`, `idle`, `daily`, `compaction`, `deleted`, `shutdown`, `restart`.

`sessionEndMessagesFromTranscript` answers `undefined` for a missing `sessionFile`, an unreadable file, or a transcript that parses to no messages, so the existing guard skips the flush exactly as it did before. One synchronous read, inside the drain's shared 2s budget, with extraction still asynchronous in the bank's operation queue — the constraint you named.

Your `/new` note decided the shape: the resolution happens at handler time from `sessionFile`, not from the live session entry, because by the time OpenClaw emits the previous session's `session_end` on the successor's first turn the entry no longer holds those messages.

**One structural move.** `parseSessionFile` (with `readJsonLines` and `extractTextContent`) is now `src/session-file.ts`. `backfill-lib.ts` imports `./index.js`, so having `index.ts` import `backfill-lib.ts` back would have made the cycle load-order dependent; `backfill-lib` re-exports all three names, so its own surface and the parsing are unchanged.

## Testing

Five cases in `src/index.test.ts`, against real transcripts on disk:

- the messages come back from the file the event points at, in order and with content intact;
- `undefined` when the event carries no `sessionFile`;
- `undefined` when the file has been removed, rather than throwing out of a session close;
- `undefined` when the transcript parses to zero messages, so the caller's guard still skips;
- the agent id from `event.context` reaches the reader.

```
npx tsc --noEmit        clean
npx vitest run src      348 passed (13 files)
npx prettier --check    clean
```

To be straightforward about one gap: those cover `sessionEndMessagesFromTranscript`, not the call site inside `runRetain`, which is a closure in the plugin factory. Removing the call reddens nothing, because reaching the retain from a `session_end` hook needs the service, bank and client stubbed the way `local-daemon-parity.test.ts` does for the daemon. Happy to add that harness if you want the wiring pinned too — I did not want to grow the diff with a test fixture larger than the fix.


koriyoshi2041 (commented):
Reviewed exact head `ab58605961c37257b504c2d3c4fc41112f4ed191`. The OpenClaw extraction is structurally sound: the transcript parser moved without semantic changes, the forced path resolves messages before the existing guard, and the existing turn-count/overlap logic still selects only the unretained tail. I ran the integration package locally: 348/348 source tests pass, `tsc` passes, and Prettier passes on the four touched TS files.

One merge-scope item to make explicit: this PR currently also contains `8a3aaa2` (the full #4347 Windows probe change), so its diff against `main` is not isolated to #4341. If this is intentionally stacked on #4347, could you mark that dependency in the description (and rebase once #4347 lands); otherwise please drop that commit here. I did not find a correctness blocker in the #4341 change itself.

L4XB:
Rebased onto current `main` — `135b473f`. #4347 has merged since, so the commit
you spotted is now an ancestor rather than part of this diff, and the PR is the
single openclaw commit:

```
$ git diff --name-only origin/main...HEAD
hindsight-integrations/openclaw/src/backfill-lib.ts
hindsight-integrations/openclaw/src/index.test.ts
hindsight-integrations/openclaw/src/index.ts
hindsight-integrations/openclaw/src/session-file.ts
```

`hindsight-embed/` is gone from it entirely. Thanks for catching it — it was a
branch-cut slip, not intent.

koriyoshi2041 (approved):
Rechecked exact head `135b473f` after the rebase. The unrelated probe commit is gone from the GitHub diff, and the session-end change remains sound: forced flushes resolve the transcript before the existing empty-message guard, while missing, unreadable, and empty transcripts preserve the skip behavior. Local package evidence at this head: 348/348 source tests, `tsc --noEmit`, and Prettier on all four changed files pass. No remaining correctness or scope blocker from my review.

nicoloboschi:
Reviewed and merging this. I pushed one commit to your branch (`a0783ea`): the call-site test you flagged as the gap.

It turned out cheap — `src/retain-queue.integration.test.ts` already drives the plugin through its registered hooks against a fake server, so it only needed to capture `session_end` alongside `agent_end`. Two cases: the un-retained tail reaches the server on session close, and a `session_end` pointing at no readable transcript still skips the flush. Confirmed as a real regression test — the first case fails against `main`'s `index.ts` and passes with your fix.

That gap was worth closing precisely because it is how #1726 died: the hook fired, the guard above it saw a payload with no messages, and the tail was dropped in silence with every unit test still green. Thanks for the diagnosis and the clean fix.

PR #4356, opened 2026-09-14
fix(recall): anchor the default temporal window on UTC, not the server clock

ebarkhordar:
When the caller does not send `query_timestamp`, `analyze()` falls back to `datetime.now()`, so "yesterday" resolves against the server's local wall clock. `event_date` is stored in UTC and `retrieve_temporal_combined_sql` stamps a naive window as UTC, so on a box outside UTC the window lands on the wrong day for part of every day.

Same instant, same query, only `TZ` differs:

```
2026-09-14 09:21:33Z   "what happened yesterday"
TZ=UTC              -> 2026-09-13 00:00:00 .. 2026-09-13 23:59:59.999999
TZ=Pacific/Midway   -> 2026-09-12 00:00:00 .. 2026-09-12 23:59:59.999999
```

#1788 already moved the other anchor to `utcnow()` via `_recall_scoring_now`, so today one recall scores recency against UTC and builds its temporal window against local time.

The default is now naive UTC. An explicit `reference_date` is untouched, so the golden corpus in `test_temporal_extraction.py` does not move: every call in it passes a reference date.

The new test walks two zones inside one function on purpose. Kiritimati (+14) only disagrees with UTC after 10:00Z and Midway (-11) only before 11:00Z, so either one alone would pass on main for part of the day.


nicoloboschi:
Pushed 3b8e43e: the timezone tests now restore the runner's own `TZ` through a fixture that undoes monkeypatch before calling `tzset()`. Before, the `finally` unset `TZ` and a runner that had one set kept the wrong zone for the rest of the worker. The fix itself is unchanged; running full CI on an upstream branch before merging.

ebarkhordar:
Good catch on the TZ restore. My `finally` block assumed the runner had no `TZ` set to begin with, so on one that does it would have left the wrong zone in place for every test after mine.

PR #4359, opened 2026-09-14
fix(paperclip): make hindsight_recall honor the agent's query and make the request timeout configurable

sdfantini-cs:
## Problem

- `hindsight_recall` returns the memories recalled at `agent.run.started` whenever they exist, ignoring the `query` argument. Mid-run lookups ("what did we decide about hosting?") always get the issue-level memories back instead of a targeted recall.
- The HTTP client has a hardcoded 15s timeout. On self-hosted Hindsight the run-start recall (query = issue title + description) can exceed it — in our deployment 15 of 16 run-start recalls failed with an aborted request.

## Change

- The run-start result is cached together with its query and reused only when the tool is called with that same query; any other query performs a live recall.
- New optional config `requestTimeoutMs` (default `15000`, unchanged behavior when unset), applied to all Hindsight requests.
- README updated.

## Tests

- New: live recall on a different query, request aborted after `requestTimeoutMs`, default/custom/invalid timeout values.
- Existing cache test now asserts reuse for the same query.
- `npm run build`, `npm run typecheck`, `npm test` → 62 passed. The new tests fail on `main`.


koriyoshi2041 (commented):
Reviewed exact head `f2eb0c17`. The cache key now follows the actual run-start query, so a different mid-run query reaches Hindsight instead of reusing issue-level context; same-query reuse still avoids the extra request. I also checked that the configured timeout reaches every recall/retain client construction path.

Local Paperclip results: 62/62 tests passed, TypeScript typecheck passed, build passed, and `git diff --check` passed. One small pre-merge cleanup: `npx prettier --check src/client.ts src/manifest.ts src/worker.ts tests/plugin.spec.ts README.md` reports `src/worker.ts` as unformatted. The behavior change itself looks sound.

sdfantini-cs:
Thanks for the review. Formatted `src/worker.ts` with prettier (no behavior change); `prettier --check` on the five files now passes. Note: I also rewrote the author on the original commit, so the head moved from `f2eb0c17` to `36b5c9b6`; the tree of that commit is unchanged.

koriyoshi2041 (approved):
Rechecked the updated head: the follow-up commit only applies the requested Prettier formatting in `worker.ts`; the behavior and tested scope are unchanged. Looks good.

sdfantini-cs:
Friendly ping on this one — the last exchange was a week ago.

Recap: `hindsight_recall` ignored the agent's query, and the request timeout was not configurable. The review feedback was addressed on 14 September (prettier formatting on the five files), and the branch has been sitting since.

Extra data point in the meantime: we have been running the same change on a self-hosted instance for a week, against a real agent workload, with no issue. The configurable timeout in particular removes the need for a local patch on our side.

Happy to rebase if the branch has drifted, or to split it if the timeout option and the query fix are better reviewed separately. Anything else you need from me?

PR #4366, opened 2026-09-14
fix(attachments): reclaim an attachment a re-retain stops referencing

### LINKED ISSUE #4364: Re-retaining a document without an attachment never reclaims it
## Summary

Re-retaining a document without an attachment it used to reference removes the document's `document_attachments` edge but never reclaims the attachment. When that was the last reference in the bank, the `attachments` row and the stored blob stay forever: nothing references them, and nothing will delete them.

## Where

- `sync_document_attachments` (`engine/retain/fact_storage.py`) rewrites a document's edges delete-then-insert, so an attachment the new content no longer references loses its edge.
- `_reclaim_orphaned_attachments` (`engine/memory_engine.py`) is what drops an unreferenced attachment's row and blob, but it is only called from document delete and from the refused-retain cleanup — never after a retain rewrites the edges.

## Repro

1. Retain document `doc` with an image block.
2. Retain `doc` again with the same `document_id`, text only.
3. The document's `attachments` edge is gone; `SELECT * FROM attachments WHERE bank_id = …` still returns the image row, and its blob is still in file storage.

## Expected

After the edge rewrite, reclaim the hashes the document referenced before the retain and no longer does — through the same `_reclaim_orphaned_attachments`, so an image another document still references survives.

## Notes

- Wasted bytes rather than a correctness problem: the attachment is unreachable once no document references it.
- `tests/test_attachment_lifecycle.py::test_re_ingesting_without_an_attachment_drops_its_edge` asserts the edge is dropped but not that the blob is reclaimed; a test for the reclaim should sit next to it.

nicoloboschi:
Fixes #4364.

Re-retaining a document without an attachment it used to reference dropped its `document_attachments` edge but never reclaimed the attachment, so when that was the last reference the `attachments` row and stored blob leaked.

`retain_batch_async` now snapshots the attachment hashes the documents being rewritten reference before the retain, and after it runs them through the existing `_reclaim_orphaned_attachments` — so a blob another document still references survives, and store-owned banks remain untouched. The reclaim also runs for a cancelled run, since documents written before the cancellation have already rewritten their edges.

## Tests
`tests/test_attachment_lifecycle.py`:
- `test_re_ingesting_without_an_attachment_reclaims_its_blob` — the repro (failed before the fix)
- `test_re_ingesting_without_a_shared_attachment_keeps_its_blob`
- `test_re_ingesting_with_the_same_attachment_keeps_its_blob`

Attachment suites (105) and retain/delta/document suites (345) pass locally; lint clean.


nicoloboschi:
Superseded by #4376, which landed the same reclaim-on-re-retain fix for #4364.

PR #4370, opened 2026-09-14
fix(recall): cache and reuse observation source memory IDs across pipeline

Sanderhoff-alt:
## Overview

When recalling observations alongside raw facts with full provenance requested (`prefer_observations=True`, `include_chunks=True`, and `include_source_facts=True`), the recall pipeline previously re-queried observation source mappings across consecutive pipeline steps. This PR retains resolved observation `source_memory_ids` on candidate results and reuses them across subsequent pipeline stages, eliminating redundant database roundtrips while preserving backend-specific junction table semantics.

### Recall Pipeline Execution Flow

```mermaid
flowchart TD
    subgraph S48["Step 4.8: Prefer-Observations Dedup"]
        A[Recalled Observation Candidates] --> B{Candidates carry sources?}
        B -- No --> C["Fetch get_memories() from DB"]
        B -- Yes --> D["Use in-memory sources"]
        C --> E["Backfill sr.retrieval.source_memory_ids"]
        D --> F["Compute superseded raw facts"]
        E --> F
    end

    subgraph S55["Step 5.5: Chunk Resolution"]
        F --> G{uses_observation_sources_table?}
        G -- "Yes (Oracle)" --> H["SQL JOIN observation_sources"]
        G -- "No (PostgreSQL)" --> I{All sources carried in-memory?}
        I -- Yes --> J["Direct SELECT chunk_id WHERE id = ANY(source_ids)"]
        I -- No --> K["SQL JOIN memory_units"]
    end

    subgraph S6["Step 6: Source Facts Resolution"]
        J --> L{All sources carried in-memory?}
        K --> L
        H --> L
        L -- "Yes (Fast Path)" --> M["Reuse in-memory source_memory_ids without SQL read"]
        L -- "No (Fallback)" --> N["SELECT id, source_memory_ids FROM memory_units"]
        M --> O["Fetch source fact contents for returned candidates"]
        N --> O
    end
```

### Stage-by-Stage Comparison

| Stage | Operation | Before This PR | After This PR | Roundtrips / Cost Saved |
| :--- | :--- | :--- | :--- | :--- |
| **Step 4.8** (`prefer_observations`) | Deduplicate superseded raw facts | Fetched observation rows to read `source_memory_ids`, stored them in a local set, but discarded candidate mappings. | Retains `source_memory_ids` directly on `sr.retrieval` candidates without temporary dictionary wrappers. | Eliminates discard; enables downstream stages to skip database reads. |
| **Step 5.5** (`include_chunks`) | Map observations to source chunk IDs | Always executed a JOIN between `observation_sources` or `memory_units` self-join. | On PostgreSQL (`not uses_observation_sources_table`), resolves chunk IDs directly via primary key `WHERE id = ANY(...)` over in-memory source IDs. | Replaces self-join query with direct primary key lookup; preserves Oracle junction table queries. |
| **Step 6** (`include_source_facts`) | Retrieve observation source fact IDs | Evaluated `all(sr.retrieval.source_memory_ids is not None)` to False because no prior step retained them on SQL backends, triggering redundant `SELECT id, source_memory_ids FROM memory_units`. | Fast-path evaluates to True, reusing the cached in-memory source IDs directly. | **Eliminates 1 full database roundtrip** (`SELECT id, source_memory_ids`). |

### Backend Compatibility Matrix

| Backend | Junction Table Mode (`uses_observation_sources_table`) | Step 5.5 Behavior | Step 6 Behavior | Provenance Correctness |
| :--- | :--- | :--- | :--- | :--- |
| **PostgreSQL (Default)** | `False` | Resolves source chunk IDs directly via `id = ANY($1::uuid[])` | Hits in-memory fast path, skipping redundant observation source query | Verified by automated test suite |
| **External Memory Store** | `False` (`store_owned_for`) | In-memory source reuse across store reads | Hits in-memory fast path | Verified by interface contracts |
| **Oracle** | `True` | Uses `observation_sources` junction table JOIN (preserves transfer import & consolidation semantics) | Resolved via junction table if needed | Junction table source of truth preserved |

## Key Changes

1. **Step 4.8 Candidate Backfill & Simplification** (`memory_engine.py`):
   - Backfilled `sr.retrieval.source_memory_ids` onto candidates when observations are loaded from the store.
   - Removed intermediate dictionary wrapper `obs_rows`, collecting superseded IDs directly from candidates.
   - Guarded assignment with `if sr.id in obs_by_id:` to ensure unfetched observations remain `None` for clean downstream fallback.

2. **Step 5.5 Direct Source Resolution & Oracle Guard** (`memory_engine.py`):
   - Added `not self._backend.ops.uses_observation_sources_table` check to ensure junction table reads remain untouched on backends that treat `observation_sources` as the single source of truth (such as Oracle with transfer import).
   - On PostgreSQL, resolved candidate source chunks directly via `id = ANY(...)` using in-memory source IDs without joining observation records.

3. **Step 6 Fast-Path Activation & Dead Code Removal** (`memory_engine.py`):
   - Enabled the existing in-memory fast path `if all(sr.retrieval.source_memory_ids is not None for sr in observation_srs):` to execute, skipping redundant queries.
   - Removed dead backfill writes in Step 6 fallback branches since `sr.retrieval.source_memory_ids` has no downstream consumers beyond Step 6.

4. **Test Suite Hardening** (`test_recall_pg_enrichment_combo.py`):
   - Extended `test_recall_all_enrichments_together_on_default_store` with connection pool query interception.
   - Added safety assertions ensuring intercepted query collection is non-empty to guard against false green test passes.
   - Verified that Step 6 redundant queries (`SELECT id, source_memory_ids FROM memory_units`) are eliminated (`count == 0`), Step 5.5 skips observation JOIN queries (`ON mu.id = ANY(obs.source_memory_ids)` count == 0), and executes direct primary key lookups (`SELECT id, chunk_id FROM memory_units WHERE id = ANY(...)`).
   - Verified end-to-end chunk text, source facts, and entity inheritance integrity.
   - Removed redundant duplicate test file `test_recall_observation_sources_caching.py`.

## Verification

- Ran unit and integration tests against embedded PostgreSQL:
  ```bash
  pytest tests/test_recall_pg_enrichment_combo.py tests/test_recall_prefer_observations.py -v
  # 6 passed, 0 failed
  ```
- Ruff lint and formatting checks:
  ```bash
  uv run ruff check hindsight_api/engine/memory_engine.py tests/test_recall_pg_enrichment_combo.py
  uv run ruff format --check hindsight_api/engine/memory_engine.py tests/test_recall_pg_enrichment_combo.py
  # All checks passed!
  ```


nicoloboschi:
Thanks! Pushed a small follow-up: kept the Step 4.8 write-back (that's what lets the source-facts step skip its read), but reverted the new Step 5.5 SQL branch — it replaced one query with another single query, so no round trip saved for an extra code path. The test now asserts only the eliminated read.

PR #4375, opened 2026-09-14
Set the extension context on the operation validator extension

### LINKED ISSUE #4372: The operation validator extension never receives an ExtensionContext
## Summary

Extensions loaded into the `OPERATION_VALIDATOR` slot are never given an `ExtensionContext`, so
`self.context` raises `RuntimeError` for the whole life of the process. The `TENANT` slot *does*
receive one. There is no way for a validator to reach the memory engine, the database URL, or
anything else the context provides — which makes any validator whose hooks need system access
impossible to write.

## Steps to reproduce

1. Write a validator that touches `self.context`, e.g. in `validate_retain`:

   ```python
   class MyValidator(OperationValidatorExtension):
       async def validate_retain(self, ctx: RetainContext) -> ValidationResult:
           engine = self.context.get_memory_engine()   # <-- raises
           return ValidationResult.accept()
   ```

2. Configure and start the API:

   ```bash
   export HINDSIGHT_API_OPERATION_VALIDATOR_EXTENSION=mypackage.validators:MyValidator
   hindsight-api
   ```

3. Retain anything.

## Expected

`self.context` returns the context, exactly as it does for a `TENANT` extension.

## Actual

```
RuntimeError: Extension context not set. Context is available after the extension is loaded by the system.
```

Raised from `Extension.context` (`hindsight_api/extensions/base.py:62`).

## Cause

Both entry points build a `DefaultExtensionContext` and set it on the tenant extension only:

* `hindsight_api/main.py:356-363` (inside `main()`, the single-worker path)
* `hindsight_api/server.py:78-83` (module level, the `--workers`/`--reload` path)

```python
# Set extension context on tenant extension (needed for schema provisioning)
if tenant_extension:
    extension_context = DefaultExtensionContext(
        database_url=config.database_url,
        memory_engine=_memory,
    )
    tenant_extension.set_context(extension_context)
```

`load_extension("OPERATION_VALIDATOR", OperationValidatorExtension)` at `main.py:315` /
`server.py:59` is never paired with a `set_context` call, and nothing else sets one.

The `Extension.set_context` docstring says the loader calls it after instantiation, which points the
same way — it just does not happen for this slot.

## Environment

* `hindsight-api-slim` 0.9.2 (also present on `main` as of 0.10.0)
* Python 3.11.13, Linux x86_64
* Storage: embedded PostgreSQL (pg0), local embeddings/reranker

## Notes

`TENANT` is the only slot with a context today. If the intent is that only tenant extensions need
one, then `OperationValidatorExtension` exposes `context` it can never use, and that would be worth
documenting; otherwise this is a one-line omission at two call sites.

A test asserting the existing behaviour is already there for the tenant slot
(`tests/test_main_module.py:249`, `tests/test_server_module.py:217`), so the gap is easy to pin for
the validator too.

cagerber:
Set the extension context on the operation validator extension

Closes #4372

## What

`main.py` and `server.py` build a `DefaultExtensionContext` and set it on the tenant
extension only. Extensions in the `OPERATION_VALIDATOR` slot never receive one, so
`self.context` raises `RuntimeError` (`extensions/base.py:62`) for the whole life of
the process and a validator whose hooks need the engine cannot be written at all.

Build the context once and set it on both extension slots, still only when at least
one extension actually needs it — behaviour when no extension is configured is
unchanged, including the existing `len(context_created) == 1` assertions.

## Changes

| File | Change |
|---|---|
| `hindsight_api/main.py` | set the context on `operation_validator` as well as `tenant_extension` |
| `hindsight_api/server.py` | same, on the module-level path used by `--workers`/`--reload` |
| `tests/test_main_module.py` | `MockOperationValidator` records `set_context`; new test |
| `tests/test_server_module.py` | same |

The two new tests mirror `test_{main,server}_sets_extension_context_on_tenant_extension`,
which already assert this for the tenant slot, so the validator slot is pinned the same
way rather than by a new convention.

`MockOperationValidator` needed a `set_context` override for the same reason
`MockTenantExtension` has one: the base class only stores the context, and the mock needs
to record that the call happened.

## Verification

Run against `hindsight-api-slim/`, Python 3.11.13.

```
uv run pytest tests/test_main_module.py tests/test_server_module.py
    -> 15 passed
```

Both new tests were confirmed to fail against the unpatched source (`assert 0 == 1` on
`len(context_created) == 1`), so they test the fix rather than the fixture.

Regression baseline, unmodified tree vs. this branch:

| Check | Baseline | This branch |
|---|---|---|
| `ruff format --check` (4 files) | clean | clean |
| `ty check hindsight_api` | 22 diagnostics | 22 diagnostics |
| `pytest tests/test_extensions.py tests/test_extension_bank_tables.py tests/test_extension_unbundling.py` | 35 passed, 32 errors | 35 passed, 32 errors |
| `ruff check` (4 files) | 6 errors | 7 errors |

The 22 `ty` diagnostics and 32 test errors are pre-existing in this local environment
(no `local-ml` extras, so `sentence-transformers` is absent) and are unaffected by this
change.

The single added `ruff check` finding is `F401 import hindsight_api.server imported but
unused` in the new server test — the same finding the five neighbouring tests in that file
already produce, because the import is for its side effect. I left it consistent with them
rather than adding a `noqa` the surrounding code does not use; note that `ruff check --fix`
would delete the import and break the test, so it should not be autofixed here.

## Not included

While testing this I found `tests/test_server_module.py::...test_server_sets_extension_context_on_tenant_extension`
fails when run **alone** in this environment:

```
TypeError: the JSON object must be str, bytes or bytearray, not MagicMock
  hindsight_api/profiling.py:77  cfg = json.loads(raw)
```

`server.py:37` imports `hindsight_api.profiling` *before* the rest of the module, and
`profiling.py:59` binds `get_config` at its own import time. When `server.py` is first
imported inside a test that has patched `hindsight_api.config.get_config`, `profiling`
binds the mock, and `_config()` reads `mock_config.profile` — a truthy `MagicMock` — into
`json.loads`. The outcome therefore depends on import order.

That is pre-existing and orthogonal to this change, so I have not touched it. My new tests
set `mock_config.profile = None` to stay deterministic regardless of which binding is in
play. Happy to open a separate issue for it if useful.


koriyoshi2041 (approved):
Both startup paths now share the same context with the validator while preserving the no-extension path. I verified the exact head with the main/server module suites (15 passed), formatting, and diff hygiene.

nicoloboschi:
fixed by #4400

PR #4377, opened 2026-09-14
fix(coding-agents): take Codex user turns from UserMessage events

nicoloboschi:
## Summary

The Codex transcript reader treated every `role:"user"` `response_item` as the user. Codex also sends injected content that way: startup context (`<recommended_plugins>`, AGENTS.md, `<environment_context>`), `<turn_aborted>` notices and compaction summaries. Codex Desktop's startup block was retained as a user turn.

What the user actually typed is recorded separately, as an `event_msg` / `item_completed` event whose item is a `UserMessage`. It comes right after its `response_item` and before any reply. When a rollout contains those events, they are now the only source of user turns, so no injected block has to be recognised by its text. A rollout with no such events (older Codex) falls back to the existing prefix check, unchanged.

Supersedes #4154 (thanks @Korayem for the diagnosis and the retain-hook regression tests, adapted here). That PR matched the startup blocks with regexes, and used `content_item_kinds` origin labels where present. Those labels are only written by recent Codex builds (0.150+). Older Desktop builds (0.146/0.147) write a two-block startup envelope with no labels, which the text matcher did not recognise, so it was still retained.

## Verification on 88 real rollouts (Codex 0.93 → 0.153, CLI / exec / TUI / Desktop)

Old reader vs this one:

- `UserMessage` events are present in 85 of 88 rollouts, across every version, and none of them contains injected content. The other 3 have no genuine user messages at all.
- **32** `<recommended_plugins>` startup turns are removed (#4154 removed 8), plus **3** `<turn_aborted>` turns.
- No assistant or tool turn changed in any rollout.
- Screenshot turns lose their `<image path=…>` markers, because the event keeps images as separate blocks.
- Known remainder: a Codex Desktop "guardian" subagent session's input *is* the parent's compaction summary, so it is still retained, as before.

## Tests

- `transcript-codex.test.ts`: user turns come from events; injected `role:user` items are ignored; user text identical to startup markup is kept; image blocks are dropped; the existing fixture now covers the no-events fallback.
- `retain-hook.test.ts`: the end-to-end retain and append tests from #4154, adapted to event-based fixtures.
- `npm test` passes (1022 passed), and `tsc --noEmit` and prettier are clean.


Korayem:
@nicoloboschi Thanks again for carrying this forward. I checked the merged parser against 50 local Codex Desktop rollouts using the exact current `main` implementation.

The `item_completed` / `UserMessage` approach works correctly for top-level sessions: none of their startup envelopes survived.

I found a remaining subagent case:

- 37 rollouts contained a `<recommended_plugins>` startup envelope.
- 19 had `UserMessage` events and were filtered correctly.
- 18 had no `UserMessage` event; all 18 were subagent sessions.
- Those came from Codex Desktop `0.154.0-alpha.6.2` (12) and `0.155.0-alpha.9.2` (6).
- Running the exact current-`main` parser retained the startup envelope in all 18 because it entered the no-event fallback, whose prefix check does not recognize `<recommended_plugins>`.
- All 18 carried aligned `content_item_kinds` for the startup blocks; some also carried genuine `user.text` items.

So the event-based signal looks right as the primary path, but the fallback remains incomplete for current Desktop subagents. A narrow fix could use aligned `content_item_kinds` when no `UserMessage` events exist, then retain the existing prefix check for older rollouts without origin metadata.

This preserves the direction of #4377 while covering the newer subagent format. I can provide a sanitized fixture or open a focused follow-up PR if useful.

PR #4379, opened 2026-09-15
fix(control-plane): serve the UI when bound to a loopback address

### LINKED ISSUE #1926: Control plane: every page route 307-redirect-loops under Next 16 (next-intl rewrite leaked as redirect)

snevesgraca-nwmo:
## The problem

Setting `HINDSIGHT_CP_HOSTNAME` to a literal loopback address makes the control plane **307 every page to itself**. The UI is completely unreachable.

What makes this expensive is that nothing else notices. The bind is correct, `/health` answers, the entire REST API works normally — so `systemctl`/container health checks, smoke tests and a full retain→recall round trip all stay green while the browser surface is dead. I lost a while to it before thinking to open the UI at all, because every signal I had said the service was healthy.

Root cause is the one #1934 identified: Next.js standalone uses `HOSTNAME` both to bind and as the origin its `next-intl` locale rewrite is checked against. Next normalizes `127.0.0.1` → `localhost` on one side of that comparison but keeps the literal on the other, so the rewrite reads as cross-origin and leaks out as a redirect instead of being applied internally.

## Why this is being re-raised

#1926 was closed on the conclusion that *"production is unaffected"*, which is true for the `0.0.0.0` default. It doesn't hold for a deliberately loopback-only deployment — which is a natural posture for self-hosting, because a self-hosted server logs `No API key configured (public access)`, so **the bind is the access control**. The documented workaround, "bind `0.0.0.0`", is the one thing that deployment can't do: it would trade a broken UI for an unauthenticated service on every interface.

#1934 named the exact remedy — *"a small follow-up can normalize a `127.0.0.1` bind to `localhost`"* — and deferred it. That follow-up hasn't landed, and two independent reproductions sit on the closed thread. This PR is that follow-up.

**Upgrading Next is not the fix.** #1928 pinned 16.2.5 for this; #1934 reverted the pin. The 0.10.0 image ships Next **16.3.5** and still loops.

## Why it isn't a bare substitution

Substituting `localhost` alone would be a regression. `localhost` resolves through `/etc/hosts`, and Docker's generated hosts file maps `::1 localhost` — so a bare substitution silently moves a requested **IPv4** bind onto **IPv6**. Pinning Node's DNS result order to the requested family keeps the effective bind identical and changes only the spelling, which is all Next's origin comparison looks at.

Measured on 0.10.0 (Next 16.3.5), control-plane-only containers, one variable at a time:

| `HINDSIGHT_CP_HOSTNAME` | bind | `GET /dashboard` |
|---|---|---|
| `127.0.0.1` | `127.0.0.1` | **307 → itself (loop)** |
| `::1` | `::1` | **500** |
| `localhost` | `::1` | 200 — *family changed* |
| `localhost` + `ipv4first` | `127.0.0.1` | **200** |
| `localhost` + `ipv6first` | `::1` | **200** |

## The change

`resolve_cp_hostname` in `docker/standalone/start-all.sh`: when a loopback literal is requested, bind as `localhost` and append `--dns-result-order=<family>` to `NODE_OPTIONS`. Anything else — including the `0.0.0.0` default — is untouched and silent. An operator's existing `NODE_OPTIONS` is appended to, never replaced.

## Verification

- Unit tests added to `docker/standalone/test-start-all.sh` (the suite CI already runs); full suite green.
- **Falsified**, to confirm the tests would catch a regression rather than just passing: removing the family pin, replacing `NODE_OPTIONS` instead of appending, and widening the trigger to every hostname each fail a test. The last one matters most — it proves the trigger is narrow, not merely that the check works.
- **End to end in the published image**, patched script mounted over `/app/start-all.sh` in `ghcr.io/vectorize-io/hindsight:0.10.0`: `127.0.0.1` → v4 bind + 200, `::1` → v6 bind + 200, `0.0.0.0` → untouched, silent, 200.

Happy to adjust the shape of this — a hard failure with an actionable message instead of a silent correction would also be defensible, and I'd understand preferring that. Closes #1926.


nicoloboschi:
Thanks — verified end to end in the 0.10.0 image (127.0.0.1 / ::1 now serve /dashboard 200 on the same address family; 0.0.0.0 and localhost unchanged). I pushed one follow-up commit applying the same loopback handling to the npm CLI (`hindsight-control-plane/bin/cli.js`), which passed `-H 127.0.0.1` straight through and still looped, plus a vitest for it.

PR #4381, opened 2026-09-15
fix(api): expose document id to async retain validators

### LINKED ISSUE #4373: RetainContext.document_id is None on one retain path and set on the other
## Summary

The two retain paths disagree about where the document id lives, so an extension that reads only
`RetainContext.document_id` works on one path and silently does nothing on the other.

* `retain_async` sets `RetainContext.document_id`.
* `submit_async_retain` leaves it `None` and carries the id on each content dict instead.

Nothing errors. The retain succeeds. An extension hook that keyed off the context field simply never
fires on the second path, which is invisible without instrumenting the hook.

## Steps to reproduce

```python
class MyValidator(OperationValidatorExtension):
    async def validate_retain(self, ctx: RetainContext) -> ValidationResult:
        print("document_id =", ctx.document_id, "| on contents =",
              [c.get("document_id") for c in ctx.contents])
        return ValidationResult.accept()
```

Configure it, then retain the same document through both paths (`retain_async`, and the async submit
path). Expected: the same id both times.

## Actual

```
retain_async        -> document_id = 'conversation:hermes::sess-1' | on contents = ['conversation:hermes::sess-1']
submit_async_retain -> document_id = None                       | on contents = ['conversation:hermes::sess-1']
```

## Impact

Any `OPERATION_VALIDATOR` extension that needs to identify the document under retain has to read both
sources and take the first non-empty one. That is a workaround for an inconsistency rather than a
documented contract, and it is the kind of thing that gets shipped as a bug: the extension looks
correct, does nothing on one path, and reports success.

## Suggested fix

Set `RetainContext.document_id` on the `submit_async_retain` path too, so the two agree. If the field
is deliberately context-path-specific, documenting that on `RetainContext.document_id` (and pointing
at the content dicts) would be enough to stop the next extension author repeating this.

## Environment

* `hindsight-api-slim` 0.9.2 (also on `main` as of 0.10.0)
* Python 3.11.13, Linux x86_64, embedded PostgreSQL (pg0)

## Note

Found while writing an extension that substitutes alternative text into the retained body at
`validate_retain`. `tests/test_main_module.py` already pins extension loading, so a test here would be
cheap: assert both paths populate `document_id`.

koriyoshi2041:
## Problem

Single-item async retains carry a document ID in `contents`, but their operation validator receives `RetainContext.document_id=None`. The synchronous path exposes the same ID directly, so validators behave differently depending on submission mode.

## Fix

Populate `RetainContext.document_id` from the sole content item on the async path. Multi-item submissions keep the field unset because a singular context field cannot represent multiple document IDs.

## Test

- `uv run --directory hindsight-api-slim pytest tests/test_async_retain_tags.py -q -n 0` (2 passed)
- Ruff lint and format checks on the changed files
- `git diff --check`

## Risk

Low. The change only adds context already present in the single submitted item. Batch behavior remains unchanged.

Fixes #4373


L4XB:
I went to look at #4373 and ended up measuring something that I think changes what this PR should do. Sharing it rather than sitting on it.

**The issue's premise does not hold on `main`.** #4373 says `retain_async` sets `RetainContext.document_id` and `submit_async_retain` does not. `retain_async` does not set it either: it puts the id on the content dict and calls `retain_batch_async` without the batch-level kwarg (`memory_engine.py:5326-5341`). Same on the `v0.9.2` the issue names.

Driving a validator that records the context and then rejects, so each call stops right after the context is built, on clean `main`:

| entrypoint | `ctx.document_id` |
| --- | --- |
| HTTP `async=false` -> `retain_batch_async(contents=[...])` | `None` |
| engine `retain_async(document_id=...)` | `None` |
| HTTP `async=true` -> `submit_async_retain(contents=[...])` | `None` |
| worker -> `_handle_batch_retain(task)` | `None` |
| 2-item batch | `None` |
| `retain_batch_async(document_id=...)` (the deprecated batch kwarg) | the id |

So the five real entrypoints already agree. The only thing that populates the field is the batch-level `document_id=` kwarg, which `retain_batch_async`'s own docstring marks `**DEPRECATED** - Use "document_id" key in each content dict instead` (`memory_engine.py:5385`), and which nothing in the HTTP layer, the MCP tools, `retain_async` or the worker passes. `tests/test_extensions.py:473` pins that reading, with the comment "contents is copied before document_id is applied to individual items".

**With this patch applied, the validator sees two different values for one retain.** `validate_retain` fires twice per async retain: once at submit, and again when the worker re-enters `retain_batch_async` (`memory_engine.py:3019`; the worker loads the same `OPERATION_VALIDATOR` at `worker/main.py:275-288`). Same script, this PR's line applied to a clean tree:

| entrypoint | with this PR |
| --- | --- |
| HTTP `async=false` | `None` |
| engine `retain_async(document_id=...)` | `None` |
| HTTP `async=true` -> `submit_async_retain` | the id |
| worker -> `_handle_batch_retain` | `None` |
| 2-item batch | `None` |

A validator keying off `context.document_id` now fires once with the id and once without, for the same retain. And single-item *async* sets the field while single-item *sync* still does not, which is the asymmetry #4373 reported, with the paths swapped.

**On the `len(contents) == 1` guard.** The async path genuinely carries several ids per submission: there is no batch-level `document_id` parameter on `submit_async_retain` at all, it explicitly rejects duplicate document ids within one submission ("Each content item in an async batch must have a unique document_id to avoid races between the parallel workers"), `_split_contents_into_async_children` fans one submission across children by token budget, and the HTTP layer issues one `submit_async_retain` per strategy group carrying every item in that group (`api/http.py:9186`). A scalar on the shared context cannot represent that, and the guard is the concession to it.

If the goal is "an extension can see which document it is looking at", the per-item id is where it already lives on every path, so the fix might belong on whatever the hook is handed per item rather than on the batch context. I have not written that, and it is your call and the maintainers' — I just did not want the two-values-per-retain part to go unnoticed.

Method, so you can check it: clean `main` at `4b5c0c46`, then the same tree with only this PR's line applied, reverted afterwards, `git status --porcelain` empty on both sides.

koriyoshi2041:
Thanks for checking this across all entry points. You're right: I treated the deprecated batch-level field as the canonical contract and missed both the second validator pass in the worker and the existing per-item `document_id` contract. That makes this patch internally inconsistent rather than a parity fix. Closing it; I won't move the scalar onto another path without a maintainer-approved contract for item-level validation.

PR #4385, opened 2026-09-15
fix(coding-agents): resolve the auto-update version check through npm

Sanderhoff-alt:
## Problem

The auto-update check and the update it triggers read **two different registries**:

```mermaid
flowchart LR
    subgraph BEFORE["Before — two sources of truth"]
        S1["session start"] --> F["fetch registry.npmjs.org/&lt;pkg&gt;/latest (hardcoded)"]
        S1 --> N["spawn npx &lt;pkg&gt;@&lt;version&gt;"]
        F -.-> R1[("registry.npmjs.org")]
        N -.-> R2[("user npm config: mirror / proxy / private registry")]
    end
```

`npx` resolves through the user's npm config — registry mirror, proxy, private-registry auth in `.npmrc`. A hardcoded `registry.npmjs.org` fetch desynchronises the two:

| Environment | Before | After |
| --- | --- | --- |
| Mirror (`npmmirror`, corporate proxy), npmjs.org unreachable | fetch times out after 5s → check fails closed → **auto-update silently never fires**, although `npx` works fine | `npm view` goes through the mirror (~0.2–0.3s) → updates normally |
| Mirror, npmjs.org reachable | check pins `@<version>` from npmjs.org; the lagging mirror cannot resolve it → spawned update fails, no signal | check and install read the same registry by construction |
| Offline | 5s timeout, silent | ~1s npm failure, one info log line |
| Windows | check succeeded; update spawn failed with one `update spawn failed` warn | check is a *logged* ENOENT no-op (spawn without a shell cannot see `npm.cmd`) — the shell wrapper is a separate follow-up issue |

## Solution

`npm view <pkg> version --json` answers the check, so **the check and the install consult the same resolver by construction**. One new leaf function, `npmViewVersion()`, owns the entire lookup:

```mermaid
flowchart TD
    S["spawn npm view &lt;pkg&gt; version --json<br/>(cwd = hook cwd — deliberate)"] --> E{"first event wins<br/>(settled guard)"}
    E -->|"error: ENOENT / EACCES"| D1["done('', describeError(e))"]
    E -->|"stdout error: EPIPE"| D1
    E -->|"close(code ≠ 0)"| D2["done('', exited &lt;code/signal&gt;)"]
    E -->|"close(0)"| P{"parses to a JSON string<br/>matching RELEASE_RE?"}
    E -->|"nothing arrives —<br/>grandchild holds the pipe"| H["hard deadline @ +6s:<br/>SIGKILL + done('')"]
    P -->|"yes"| OK["done(version) — no log"]
    P -->|"no"| D3["done('', unparsable / not a release)"]
    D1 --> L["exactly ONE info line<br/>(reason rides done())"]
    D2 --> L
    D3 --> L
    H --> L
    L --> R["'' = 'no newer version' → lock released, session unharmed"]
```

Design points, each anchored in the code comments:

- **Failure semantics are fail-closed and total** — every way the lookup can end resolves to `""`, which reads as "no newer version", never as "unknown, try harder". `npmViewVersion` never throws.

| `npm view` outcome | Resolved | Log line (info) |
| --- | --- | --- |
| exit ≠ 0 (E404, offline, 401) | `""` | `npm view exited <code>` |
| timeout kill (spawn `timeout`) | `""` | `npm view exited SIGTERM` |
| spawn ENOENT (no npm; Windows `npm.cmd`) | `""` | `spawn npm ENOENT (ENOENT)` |
| stdout pipe error (EPIPE) | `""` | `describeError(e)` |
| `close` never arrives (grandchild holds the pipe) | `""` at +6s | `npm view never exited — hard deadline` |
| stdout not JSON / not a single string | `""` | `unparsable npm output: <first 80 chars>` |
| string rejected by the release whitelist | `""` | `not a release: <value>` |
| valid version | version | — |

- **Exactly one log line per failed lookup, structurally.** The reason rides `done(v, reason?)`, inside the settled guard. Node reports a spawn failure as `error` followed by `close(-2)`; the trailing event must not log a wronger second line ("exited -2" about a process that never ran). Without any log, a failed check was indistinguishable from "already current" — the state file's `latest: ""` carries no reason — making "auto-update never works on my machine" unfalsifiable. Cost is bounded: after the 24h gate and the update lock, at most one line per machine per day.

- **The hard deadline closes the gap the old `AbortSignal.timeout` used to cover.** The spawn `timeout` SIGTERMs npm itself, but `close` — unlike `exit` — waits for the stdio pipes to drain. A grandchild holding the inherited pipe would stall the promise past the kill while the update-lock holder stays alive, blocking every other session to `LOCK_STALE_MS`. The +1s deadline force-settles and SIGKILLs, so the lock is released on the spot. The timer is deliberately not `unref`'d: in the stall it is what keeps the hook process alive long enough to settle; the normal path clears it inside `done()` within ~1.2s.

- **cwd sensitivity is deliberate.** The lookup runs in the caller's cwd, reading the project-level `.npmrc` and `package.json` there — exactly as the `npx` spawn does — so a project-configured registry fails both the same way. Chdir-ing to a pinned homedir would reintroduce the two-source desync this PR removes.

- **PATH gate extended** to `npx` **and** `npm` (they ship together, so one verdict in practice); both halves are tested.

## Release whitelist

```ts
const RELEASE_RE = /^\d+\.\d+\.\d+(?:-[0-9A-Za-z.]+)?$/;
```

Nothing crosses to the spawn without matching. `isNewer` does **not** stop hostile input — `parseInt("3 && calc")` is `3`, so `isNewer("1.2.3 && calc", "1.0.0")` is `true` (reproduced in the tests' rationale). Today the version lands in one argv entry where it is inert, but it is destined for a command line the moment a Windows shell wrapper lands; a tainted mirror, a hijacked private registry or one MITM is enough to deliver the payload. Pre-releases (`1.2.3-rc.1`) remain valid. Side benefit: the same whitelist keeps hostile strings out of `.update-check.json`, where `latest` is persisted.

## Measured cost (macOS, once per 24h)

| Path | Time |
| --- | --- |
| `npm view`, official registry, warm cache | 1.24–1.35s |
| `npm view`, official registry, cold cache | 1.05–1.26s |
| bare fetch `/latest` (old implementation) | 0.91–0.94s |
| `npm view` via `npmmirror` | 0.23–0.31s |

The delta over a bare fetch is npm CLI startup (~0.3s), not the request — cold and warm cache measure the same. Still well inside the 5s `NPM_VIEW_TIMEOUT_MS` budget; mirror users get a check that is both *working* and faster than the hardcoded fetch ever was.

## Tests

The real lookup is exported and tested directly — the 0.5.0 406 bug shipped precisely because only the fetch *seam* was stubbed and the real URL was never exercised (asking `/<pkg>/latest` with the abbreviated-packument accept header gets a hard 406, which read as "no newer version"). The new coverage pins the spawn protocol itself, so the equivalent mistake cannot hide behind the seam again:

| Coverage | Asserts |
| --- | --- |
| spawn protocol | binary `npm`, args `view <pkg> version --json`, `timeout: 5000` |
| failure shapes | non-zero exit (with npm's error event alongside), timeout kill (`close(null)`), unparseable stdout, JSON-but-not-a-string, EPIPE, ENOENT-then-close double event |
| hard deadline | a child emitting nothing settles `""` with `SIGKILL` once the deadline fires |
| one line per lookup | `log.info` called exactly once per failed lookup (count only, not wording) |
| whitelist | `1.2.3 && calc` rejected, `1.2.3-rc.1` accepted |
| PATH gate | both the npx-missing and npm-missing halves skip the check |

Suite: 1022 passing (72 files), `lint.sh` green, `tsup` + DTS build green. The `maybeAutoUpdate` seam changes shape accordingly — `fetch` becomes `npmView: (pkg) => Promise<string>`, resolving `""` for failure the way the real implementation does rather than throwing.

## Docs

The README is the package's single source for configuration, so its `autoUpdate` row now states the npm requirement, with `skill/SKILL.md`, the docs page and the docs skill's copy of it regenerated from it (`skill:build` + `sync-coding-agents-doc.mjs` + the pre-commit docs-skill hook).

**Follow-up (separate issue):** the Windows spawn needs `shell: process.platform === "win32"` (spawning `npm.cmd` directly trips `EINVAL` on Node builds with the CVE-2024-27980 patch); this PR's scope ends at making that failure a logged, diagnosable no-op instead of a silent one.


nicoloboschi:
Rebased onto main (the README conflict was just the reference table re-padding; regenerated SKILL.md / docs page / docs skill from it) and pushed one follow-up commit trimming the auto-update comments to the essential rationale — no behaviour change. Thanks!

nicoloboschi (approved):

PR #4387, opened 2026-09-15
feat(coding-agents): add WorkBuddy and CodeBuddy harnesses

cq-guojia:
## What

Adds two hook harnesses for Tencent's coding agents, both on the shared `@genie/agent-cli` engine:

- **WorkBuddy** (`install workbuddy`) — 3 hooks in `~/.workbuddy/settings.json`, the MCP server in
  `~/.workbuddy/mcp.json`, the companion skill in `~/.workbuddy/skills`.
- **CodeBuddy** (`install codebuddy`) — the same three wiring points under `~/.codebuddy`. CodeBuddy
  Code ships the same engine (WorkBuddy only overrides `dataFolderName`), so it reuses WorkBuddy's
  hook runtime and adds only the root it writes to and the harness name it stamps.

## Why the CodeBuddy reader handles two shapes

The harness first assumed CodeBuddy's on-disk transcript is WorkBuddy's. That holds for **CodeBuddy
Code (the CLI)**, which writes `~/.codebuddy/projects/<cwd-encoded>/<uuid>.jsonl` — but not for the
**IDE**, which is CodeBuddy's main surface: the IDE hands `Stop` its own per-conversation directory,

    .../CodeBuddyExtension/Data/<uid>/CodeBuddyIDE/<uid>/history/<md5(workspace)>/<conversationId>/
      index.json                { messages: [{ id, type, role, isComplete }] }  ← order only, no prose
      messages/<messageId>.json { role, message, id, extra, createdAt }         ← message is an escaped JSON string

Read as JSONL, that index yields zero `type:"message"` records, so retain took its
`turns.length === 0` exit and returned silently — no request, no log line — and an IDE session was
never written back at all. The reader now dispatches on the shape it was handed
(`core/transcript-codebuddy-ide.ts`), so both hosts retain.

## Verified end to end (macOS, self-hosted server)

- WorkBuddy: `retain_ok turns:17`, document `conversation:67d25fa9-…` present in the bank.
- CodeBuddy IDE: `retain_ok turns:6`, `conversation:` write accepted by the server.
- `npm test` (1051 passing), `tsc --noEmit` clean, `sync-coding-agents-doc.mjs --check` green.

## Notes for review

- Host-specific wrapping stays in the host's own reader: the IDE's `<user_info>`, `<git_status>`,
  `<additional_data>`, and the underscore `<system_reminder>` spelling (absent from WorkBuddy's
  transcripts) are stripped there; the shared `<user_query>` unwrap and synthetic-user guards are
  reused from the WorkBuddy reader.
- The CLI path is unchanged — any non-`index.json` path falls through to the JSONL reader.

## Icons

Both harness icons are the vendors' own app icons, extracted from the installed app bundles as
180×180 PNG — the same size and format this directory already uses for `claude-code.png` and
`opencode.png`, since neither vendor ships a vector mark.


cq-guojia:
Following up with some context on these two harnesses, since from the outside they probably look like two more entries in an already long roster.

**Why they're worth adding.** CodeBuddy (the IDE, plus the `codebuddy` CLI) and WorkBuddy are Tencent's coding agents — both built on the same `@genie/agent-cli` engine, which is why the two harnesses share a hook runtime and differ only in the config root they write to. They are among the largest coding-agent products in China by user volume, so landing these two would put Hindsight's memory in front of a very large developer base that currently has no supported path to it.

**On the diff itself,** in case it helps with triage:

- It's purely additive. No existing harness, hook runtime or retain path is modified. The only change to shared code is exporting two helpers from the WorkBuddy reader (`<user_query>` unwrap + the synthetic-user guards) so the CodeBuddy IDE reader can reuse them.
- The non-obvious part is the IDE transcript. The CodeBuddy IDE hands `Stop` a per-conversation directory (`index.json` + `messages/<messageId>.json`, with the prose as an escaped JSON string), not a JSONL transcript. Handed that, the JSONL reader found zero `type:"message"` records, so retain took its `turns.length === 0` exit and returned silently — no request, no log line, and an IDE session was never written back at all. `core/transcript-codebuddy-ide.ts` fixes that with a reader that dispatches on the shape it's given; anything that isn't `index.json` still falls through to the JSONL path, so the CLI is unaffected.
- Relevant checks are green (`test-coding-agents`, `verify-generated-files`, `check-unused-code`, `build-docs`); `npm test` (1051 passing), `tsc --noEmit` and `sync-coding-agents-doc.mjs --check` are clean locally, and both harnesses were verified end to end against a self-hosted server.

**Happy to make this easier to land however you'd prefer:** split it into two PRs (one per harness), drop the IDE reader and keep the first pass CLI-only, or trim the icons/docs. And I'd be glad to own both harnesses going forward — triaging issues and keeping them working when either product changes its on-disk schema.

Thanks for building this — just wanted the "why" and the review surface to be explicit rather than leaving a 28-file diff to speak for itself.

cq-guojia:
Adding the actual numbers, because "a Tencent product" doesn't mean much if you're not following the Chinese market.

WorkBuddy only launched in March. Tencent's Q2 2026 results (Aug 12) singled it out for growth and retention, and Analysys counted ~20.9M PC visits in June, which put it top of the domestic AI office-agent category.

CodeBuddy is the coding side. The number I can point to is Tencent's 2025 R&D report: over 90% of Tencent engineers use it, and roughly half of new code is AI-assisted. That's internal penetration, not market share, and I don't have a clean public user count for it the way I do for WorkBuddy — so I'd rather say that plainly than inflate it.

They run the same @genie/agent-cli engine, which is the only reason one hook runtime covers both.

Not rushing anyone here. I just didn't want the case for this PR to be a 28-file diff and nothing else. If size is what's blocking review, say the word and I'll split it into two or cut it down to the CLI-only version.

cq-guojia:
Quick context on why these two are worth adding: CodeBuddy (IDE + CLI) and WorkBuddy are Tencent's agents, and they're the largest agent products in China by user base — a lot of developers who currently have no supported way to use Hindsight at all.

They share the same @genie/agent-cli engine, which is why one hook runtime covers both. The diff is purely additive; nothing existing changes.

If size is what's blocking review, say the word and I'll split it into two PRs or cut it down to the CLI-only version.

PR #4396, opened 2026-09-15
fix(openclaw): cancel recall requests and client retry backoff

Tosko4:
## Summary

Connect the OpenClaw plugin's recall deadline and service lifecycle to the TypeScript client's existing `AbortSignal`, and make the client's capacity-retry wait abortable.

The changes belong together in this repository: `hindsight-integrations/openclaw` owns the hook lifetime; `hindsight-clients/typescript` owns read retries. No OpenClaw core, server, configuration, dependency-version or retain behavior changes.

## Problem

The plugin's `scopeClient().recall()` races a request against a timer but does not pass a signal to `HindsightClient.recall()`. Once the timer wins, the underlying request continues. With the capacity retries introduced in #4253, that request can receive a delayed 429/503 and start **another** HTTP attempt after the hook has already returned.

Separately, stopping/restarting the service neither invalidates in-flight recall entries nor checks the generation before returning memory context. Even a plain slow success can therefore escape an old hook after service stop. This is a plugin-hook result, not evidence of cross-session injection by an entire running host.

The client already accepts signals for recall/reflect requests, but its retry helper sleeps with a plain `setTimeout`, so cancellation during backoff does not promptly settle the operation.

## Changes

- Forward a composed deadline/service signal through bank-scoped recall. Retain the bounded race for transports that ignore cancellation and clear its timer/listener on completion.
- Check recall generation after asynchronous readiness, bank setup and response settlement; abort/clear old work on service start and stop.
- Only remove an in-flight entry if it still holds that exact promise, so old cleanup cannot evict a successor.
- Preserve external-API lazy recall before the first service start with a recall-only lifetime, invalidated by the first start/stop. Pre-start retain gating stays unchanged.
- Thread the existing recall/reflect signals into `retryOnCapacity`; abort its wait promptly, clean up timers/listeners and check cancellation before subsequent sends.
- Preserve the current max-attempt/status/jitter policy and no-retry behavior for writes. Python's equivalent already uses cancellable `asyncio.sleep`; no API-field parity change is needed.

## Verification

- Regression-first: the original plugin fails six new cancellation/lifecycle checks; the original client fails seven cancellation checks while its 11 existing retry checks pass.
- Local review caught a pre-service lazy-recall regression; three additional tests failed before correcting it.
- Plugin suite: **359 tests** (pinned released client plus explicit in-repo source-client composition tests).
- Client deterministic suite: **79 tests**. `main_operations.test.ts` requires a live backend and was not run.
- Five committed loopback HTTP scenarios exercise the actual plugin hook and current client source together: transient 429/503 recovery, no retry after timeout/stop, and disconnecting an unanswered request at the deadline. Other committed tests cover ignored signals, late results, deduplication, restart with/without stop, pre-start lazy recall and timer/listener cleanup.
- Existing additional built-plugin/client HTTP harness: **28/28** scenarios passed, including retain queue replay and session exclusions, with synthetic loopback responses only.
- TypeScript checks/builds and repository `scripts/hooks/lint.sh` pass.

These tests exercise the plugin and client, not a fault-injected production Gateway. No real memory bank or production service was changed for this PR. No generated API artifacts are changed.


nicoloboschi:
Thanks for this — the diagnosis is right, and the cancellation gap is real. I'm closing the PR because of how it's scoped, not because of what it found.

The change spans two packages that we want to move independently: `hindsight-clients/typescript` (a published client with its own consumers) and `hindsight-integrations/openclaw` (one of those consumers). A single PR means the client's retry semantics and the plugin's recall lifecycle land, and would have to be reverted, together — and the two need different reviewers and different release cadences.

I've filed the client half as #4445, which covers exactly the part that is unambiguously a client defect: `retryOnCapacity` sleeps in a bare `setTimeout`, so an aborted caller waits out `Retry-After` and then issues **another** request. Your `throwIfAborted`-before-and-after-`send` shape with an abortable wait is what that issue asks for, and the deterministic tests you wrote for it are the right tests. Please do open that one if you'd like to — the issue is written so your existing client-side commits apply nearly as-is.

For the OpenClaw half, a separate PR against the plugin alone would be very welcome, and it can go in either order — the client already accepts a per-call signal for recall/reflect today, so threading one through `scopeClient().recall()` and clearing the timeout in a `finally` stands on its own. Worth calling out that the comment at `index.ts:200` claiming the client can't take a signal is simply stale, and your PR is what makes that visible.

One request for that PR: the recall generation work (`preServiceRecallController`, `isCurrentRecall()`, clearing `inflightRecalls` on start/stop) is a lifecycle change rather than signal plumbing, and it's the part carrying the most risk. Please give it its own PR and description too, so it gets reviewed as the lifecycle change it is rather than as a footnote to a cancellation fix.

Three smaller PRs, all three mergeable — sorry for the extra round trip, and thank you for the careful regression-first work.

Tosko4:
Thanks for the thoughtful explanation, and for opening #4445. You're right: sharing a repository doesn't make the SDK's retry semantics and an integration's lifecycle the same review or release unit. I should have kept those boundaries explicit.

The three replacements are now open, each targeting main with its own scope and regression coverage:

- #4448: client capacity-retry cancellation, fixing #4445, with Python cancellation parity tests.
- #4449: OpenClaw deadline/signal plumbing, including removal of the stale client-signal comment.
- #4450: OpenClaw recall lifecycle/generation ownership, including pre-start recall and successor-safe deduplication.

The client and plugin releases remain independent. Both plugin PRs pass against the existing released client, with no dependency bump. The plugin changes meet at the recall adapter, so there is a small textual overlap to resolve after either lands; their composed implementation also passes the regression suite.

No need to apologise for the extra round trip. Thanks for validating the diagnosis and giving such a clear path forward.

PR #4411, opened 2026-09-15
fix(extensions): give the operation validator its ExtensionContext

cdbartholomew:
## Problem

An `OperationValidatorExtension` is the one extension type that never receives an `ExtensionContext`.

`MemoryEngine.__init__` takes the validator as a parameter — so it's constructed **before** the engine exists — and, unlike the tenant, HTTP, and memory-defense extensions, it is never routed through `load_extension()` (which calls `set_context`). Validators also get no `on_startup()` call. So `set_context()` is never invoked on the validator, and `self._context` stays unset.

Any validator hook that reaches an engine API through its context — e.g. `context.get_memory_engine()` to obtain the data-plane pool — therefore raises `RuntimeError("Extension context not set")` at call time. For hooks that treat such access as best-effort (swallowing exceptions), this fails **silently**: the hook simply never does its work, with nothing in the logs at default levels.

## Fix

Set the context on the validator right after the engine builds `self._ext_ctx` (which already carries `memory_engine=self`), mirroring what the memory-defense extension does a few lines below:

```python
if self._operation_validator is not None and hasattr(self._operation_validator, "set_context"):
    self._operation_validator.set_context(self._ext_ctx)
```

- `instrument_operation_validator` returns the same instance, so this sets the context on the real validator.
- Guarded for a `None` validator and for a duck-typed validator that isn't an `Extension` subclass (no `set_context`).

## Test

`tests/test_operation_validator_context.py` builds a `MemoryEngine` with a minimal validator (no DB connection needed — `set_context` runs in `__init__`) and asserts the validator can resolve the engine via `context.get_memory_engine()`. It **fails without this change** (`RuntimeError: Extension context not set`) and passes with it. A second test confirms construction still succeeds for a validator lacking `set_context`.

`2 passed`; `ruff check` clean.

https://claude.ai/code/session_011mnzArjiCdBxa8dh1H47Fa


ebarkhordar (commented):
I ran this against 18a3dcbc in a clean python:3.11-slim container, and against the merge base d6c556f6, printing `hindsight_api.__file__` each run.

Base d6c556f6, building a MemoryEngine with an operation validator and reading `validator.context`:

```
RuntimeError: Extension context not set. Context is available after the extension is loaded by the system.
```

Head 18a3dcbc, same script:

```
validator.context             = OK
get_memory_engine() is engine : True
```

Two things I checked because the diff cannot show them.

`instrument_operation_validator` returning the same object is what makes this line land on the real validator, so I measured it instead of reading it: `engine._operation_validator is v` is True at both refs, and `validate_recall.__wrapped__` is still present, so the hooks stay timed.

On whether one line is enough, I parsed the package with ast for assignments to any `*operation_validator*` attribute rather than grepping for them. There is exactly one, `memory_engine.py:2657`, so there is no second install path this misses. `_ext_ctx` is mutated in place at `memory_engine.py:5079` rather than replaced, so the validator also picks up `webhook_manager` once `initialize()` sets it.

I did not run the suite against a database, and I did not test a validator loaded through the OPERATION_VALIDATOR env var.

nicoloboschi:
Thanks @cdbartholomew — your diagnosis was exactly right, and I've shipped it as #4433 with the wiring in `MemoryEngine.__init__` as you had it.

Two things came along with it:

- `main.py` and `server.py` each built a second, identical `DefaultExtensionContext` just for the tenant extension. Both are gone, so the engine's context is the only one in the process — and it is the better one, since its `run_migration` prefers the engine's live `db_url` (the pg0 case, where the URL is only known after init).
- #4400 had added a docstring and docs line saying validators are not given a context. That described the status quo rather than a consequence of the `current_schema` removal, so it is updated: `self.context` is process-global and fine to use for `get_memory_engine()`; the tenant and bank still come from the hook's own `ctx`.

Closing in favour of #4433.

PR #4415, opened 2026-09-15
feat(coding-agents): autoInject — choose reflect, pages or recall for the first-prompt injection

nicoloboschi:
## What

New config field `autoInject` (env `HINDSIGHT_AUTO_INJECT`, per-bank overridable) selecting what the coding-agents plugin injects once, on a session's first prompt:

| value | injects |
|---|---|
| `"reflect"` (default) | low-budget reflect synthesis — unchanged, incl. the pages → recall fallback on timeout/5xx |
| `"pages"` | knowledge-page search hits for the prompt |
| `"recall"` | recalled consolidated observations for the prompt |
| `"none"` | nothing (tool-only mode, the old `autoReflect: false`) |

`pages` / `recall` are retrieval-only (no LLM), so they are fast and fit the hook window easily.

`autoReflect` is **deprecated** but 100% backwards compatible: `false` ⇒ `"none"`, `true` ⇒ `"reflect"`, ignored when `autoInject` is set (also through `banks.<id>` overrides). Setting it logs a deprecation warning to the plugin log (`@deprecated` on the `RawConfig` field, README row marked deprecated).

## Tuning knobs, applied in one place

Two settings, both defaulting to what was previously hardcoded in three separate spots:

| setting | env | default | applies to |
|---|---|---|---|
| `pageSearchLimit` | `HINDSIGHT_PAGE_SEARCH_LIMIT` | `3` | every knowledge-page search: the `autoInject: "pages"` injection, the reflect fallback, **and** the agent's `hindsight_search_knowledge_pages` tool |
| `recallOptions` | file-only | see below | the whole body of every recall: `autoInject: "recall"` and the reflect fallback |

`recallOptions` is a JSON object merged key-by-key over the default recall body, rather than one setting per parameter:

```json
{ "recallOptions": { "types": ["world", "experience"], "max_tokens": 4000 } }
```

Default body: `{"types": ["observation"], "budget": "low", "max_tokens": 2000, "include": {"entities": null}}`. Keys are the API's own recall parameters and pass through unchanged, so a parameter recall gains later needs no new setting here. `query` is applied after the merge and is the one field config cannot replace — otherwise a config could silently recall for a fixed question every turn.

It exists because an observation-only recall returns **nothing** on a bank whose consolidation is disabled: observations are the consolidated layer, so such a bank never grows any. Those banks set `{"types": ["world", "experience"]}`, or `{"types": null}` for every type. File-only, like `retainMetadata` — an object does not flatten into an env var.

All three live on `HindsightClient` (`ClientOpts`), not at the call sites, so there is exactly one place each value is applied. `searchKnowledgePages` previously took a `limit` the tool passed as a literal `3` and the hook as its own `FALLBACK_PAGE_LIMIT`; both now pass nothing and inherit the client's. `recallObservations` takes no body parameters at all — the client owns the whole payload. The two client-construction paths that can reach these methods (`runHook`, `resolveHostMemory`) pass the config through.

## How

- `hook.ts`: the page-search and observation-recall halves of `reflectFallback` became `injectPages` / `injectObservations`; the fallback chains them, the new modes call one directly (diag events `inject_pages` / `inject_observations`, `*_failed`).
- `inject.ts`: formatters take an optional lead line, so direct injection doesn't claim "synthesis was unavailable".
- Persistent-plugin harnesses share `buildHookOutput`, so they get the modes too.
- README row + regenerated SKILL.md, docs page and docs skill.

## Tests

- `hook.test.ts`: pages mode, recall mode (each never calls the other sources, injects once), failed/empty retrieval injects nothing.
- `config.test.ts`: precedence, legacy mapping, junk value, per-bank legacy override.


koriyoshi2041 (approved):
Verified head `5954b466b`. The compatibility boundary looks sound: explicit `autoInject` wins over deprecated `autoReflect`, bank-level legacy overrides are resolved without resetting the base mode, and pages/recall are cached after the first attempted injection just like reflect.

Local checks:
- focused config/hook/session-start suite: 100 passed
- package build + declaration generation: passed
- generated skill is in sync with its sources
- `tsc --noEmit`: passed

PR #4421, opened 2026-09-16
fix(cross_encoder): pass device_map=device to avoid meta-tensor race in transformers ≥5 (#4420)

### LINKED ISSUE #4420: [bug] LocalSTCrossEncoder fails with 'Cannot copy out of meta tensor' during daemon startup (sentence-transformers ≥5 + transformers ≥5 + winloop executor path)
### Environment

- `hindsight-api` 0.9.2 (`hindsight-all-slim`)
- Python 3.11.16 (Windows 11 25H2, winloop installed by hindsight)
- `torch==2.7.0+cu128`
- `transformers==5.8.0`
- `sentence-transformers==6.0.1`
- Local reranker model: `Qwen/Qwen3-Reranker-0.6B`
- Embeddings provider: `local` (Qwen3-Embedding-0.6B) — works fine
- `HF_HUB_OFFLINE=1` (avoids HF hub URL bug)

### Summary

When `HINDSIGHT_API_RERANKER_PROVIDER=local`, the daemon starts, logs:

```
Reranker: initializing local provider with model Qwen/Qwen3-Reranker-0.6B
Reranker: FP16 inference enabled
Reranker: local provider initialized (max_concurrent=4)
```

…then **the lifespan throws `Application startup failed: NotImplementedError: Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device.`**

The error originates inside `hindsight_api/engine/cross_encoder.py` line 223 inside `LocalSTCrossEncoder.initialize()`, specifically from `CrossEncoder(...).to(device)` in `sentence_transformers/base/model.py`.

### Repro

Calling `LocalSTCrossEncoder(model_name="Qwen/Qwen3-Reranker-0.6B", device="cuda", ...).initialize()` **directly from a Python REPL works perfectly**:

```
$ python -c "from hindsight_api.engine.cross_encoder import LocalSTCrossEncoder; import asyncio; asyncio.run(LocalSTCrossEncoder(model_name='Qwen/Qwen3-Reranker-0.6B', force_cpu=False, fp16=True, max_concurrent=4).initialize())"
OK
```

It **only fails** when called via `memory_engine.py:4105`:

```python
async def init_cross_encoder():
    cross_encoder = self._cross_encoder_reranker.cross_encoder
    if getattr(cross_encoder, "blocking_init", False):   # True for LocalSTCrossEncoder
        await loop.run_in_executor(None, lambda: asyncio.run(cross_encoder.initialize()))
```

`run_in_executor` + `asyncio.run` + `winloop` somehow triggers the meta-init path in `sentence_transformers.CrossEncoder.__init__` (via `from_pretrained` + `self.to(device)`).

Same code path fails when `device="cpu"` is forced — the meta tensor issue fires before device placement even matters (the model is already on the meta device after `from_pretrained`).

### What I tried that did NOT fix it

1. `transformers==5.8.0`, `5.13.0`, `5.16.1`, `5.17.0` (latest as of 2026-09-16) — all fail with the same meta tensor error.
2. `sentence-transformers==5.1.2` — `CrossEncoder` init works alone but `SentenceTransformer` init then fails with the same error.
3. `sentence-transformers==4.1.0` — sidesteps meta tensor, but Hindsight 0.9.2 hard-requires `encode_query` / `encode_document` which only exist in sentence-transformers ≥5.0 (see `engineembeddings.py:287`).
5. `HINDSIGHT_API_RERANKER_LOCAL_FORCE_CPU=true` — error still fires; `device_map=None` is hardcoded in `cross_encoder.py:226` via `model_kwargs={"low_cpu_mem_usage": False}`.
6. Setting `device_map="cuda"` in `model_kwargs` — cannot, the kwarg is set in `cross_encoder.py:226` and not env-configurable.

### Workaround I'm using today

Switch `HINDSIGHT_API_RERANKER_PROVIDER=flashrank` (ONNX Runtime, sidesteps the entire transformers/torch meta path). I have to manually pre-download `ms-marco-MiniLM-L-12-v2.zip` to `~/.cache/flashrank/` from `hf-mirror.com` because `flashrank.Config.model_url` is hardcoded to `https://huggingface.co/...` and won't honor `HF_ENDPOINT`. Recall quality is acceptable for our use case but quality is lower than `Qwen3-Reranker` for Chinese.

### Suggested fix (in `cross_encoder.py`)

Pass `device_map` alongside `low_cpu_mem_usage=False` so accelerate handles placement:

```python
self._model = CrossEncoder(
    self.model_name,
    device=device,
    model_kwargs={"low_cpu_mem_usage": False, "device_map": None},  # add this
    trust_remote_code=self.trust_remote_code,
)
```

The Hindsight source comment at `cross_encoder.py:186-187` says:
> We do NOT use device_map because CrossEncoder internally calls .to(device) after loading, which conflicts with accelerate's device_map handling.

That assumption was correct for transformers 4.x but breaks under transformers ≥5.x + winloop executor path. Adding `device_map=None` (or `device_map=device`) should let accelerate take over placement and avoid the meta init race.

### Related upstream issues I found while debugging

- sentence-transformers #3759 — Qwen3-Reranker `input_ids` dtype bug (separate from meta tensor but same model)
- transformers #38066 — `AutoModel.from_pretrained` meta→cuda regression (closed but related)
- transformers #43950 — non-persistent buffer corruption on `from_pretrained()` with transformers 5.x
- Hindsight #71266 — native SkillOpt-style validation gate for skills (related to long-term self-improvement loop)

lhby2004:
## What

Single-line fix in `hindsight-api-slim/hindsight_api/engine/cross_encoder.py` `LocalSTCrossEncoder.initialize()`: pass `device_map=device` alongside `low_cpu_mem_usage=False` so accelerate handles model placement and avoids the meta-tensor race.

Also updates the now-stale comment block above the call that says "we do NOT use device_map because CrossEncoder internally calls .to(device)" — that assumption broke under transformers ≥5.

## Why

Fixes #4420. The `LocalSTCrossEncoder` init fails with:

```
NotImplementedError: Cannot copy out of meta tensor; no data!
Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to()
when moving module from meta to a different device.
```

…during daemon startup under:

- `transformers >=5.0.0` (required by `sentence-transformers >=5.0`)
- `sentence-transformers >=5.0`
- Hindsight 0.9.2 daemon startup path (`memory_engine.py:4105` → `loop.run_in_executor` → `asyncio.run` → `winloop` event loop)

Calling `LocalSTCrossEncoder(...).initialize()` **directly from a Python REPL works perfectly** — the bug only fires when invoked from inside `run_in_executor`. (Repro in #4420.)

## Diff

```diff
-        # Note: We do NOT use device_map because CrossEncoder internally calls .to(device)
-        # after loading, which conflicts with accelerate's device_map handling.
-        # MPS is opt-in (allow_mps) — see engine/local_device.py for why.
+        # Note: We pass device_map=device so accelerate handles placement. This avoids
+        # the meta-tensor race in transformers ≥5 + sentence-transformers ≥5 that fires
+        # when sentence_transformers.CrossEncoder.__init__ calls self.to(device) on a
+        # model that from_pretrained() just created on the meta device (ref: #4420).
+        # The original "we don't use device_map" assumption was correct for transformers
+        # 4.x; on 5.x it leaves the model on meta and the .to(device) call fails.
+        # MPS opt-in (allow_mps) is unchanged — see engine/local_device.py.
         device = select_local_device(self.force_cpu, self.allow_mps)
@@
                     self.model_name,
                     device=device,
-                    model_kwargs={"low_cpu_mem_usage": False},
+                    model_kwargs={"low_cpu_mem_usage": False, "device_map": device},
                     trust_remote_code=self.trust_remote_code,
                 )
```

## Risk

Low. Adding `device_map=device` to `model_kwargs` makes accelerate manage placement, which is the standard pattern when `device` is also passed (the original "we don't use device_map" comment was specifically about avoiding that conflict, but in transformers 5.x the meta-init race is the worse problem). If accelerate is not installed, the kwarg is ignored and behavior is unchanged. If it is installed, placement stays on `device` exactly as before.

## Verified

- Direct `CrossEncoder(...).initialize()` already worked on the broken combo (transformers 5.8 + torch 2.7 + sentence-transformers 6.0.1) — the daemon path was the only failure mode.
- The user this PR was prepared for is currently on `provider=flashrank` as a workaround. After this lands + `Hindsight >=0.9.3` ships, they can flip back to `provider=local` with `model=Qwen/Qwen3-Reranker-0.6B` and recover Chinese-optimized reranking quality.

## Related

- sentence-transformers #3759 (Qwen3-Reranker `input_ids` dtype bug — separate, may also need fix in sentence-transformers)
- transformers #38066 (AutoModel.from_pretrained meta→cuda regression — closed; related)
- transformers #43950 (non-persistent buffer corruption on from_pretrained() with transformers 5.x — related)


nicoloboschi:
Closing this one — thanks for the detailed writeup, but the fix as proposed breaks the default install.

`accelerate` is not a dependency of `hindsight-api-slim`. With transformers 5.x, passing `device_map` without it raises before the model ever loads:

```
ValueError: Using a `device_map`, `tp_plan`, `torch.device` context manager or setting
`torch.set_default_device(device)` requires `accelerate`.
  transformers/integrations/accelerate.py:134 check_and_set_device_map
```

Verified against the repo env (transformers 5.12.1, sentence-transformers 5.2.0, no accelerate). So this turns a bug that fires on one specific GPU/winloop path into a hard startup failure for every default install running `HINDSIGHT_API_RERANKER_PROVIDER=local`.

Two other gaps: `LocalSTEmbeddings.initialize()` in `engine/embeddings.py` has the identical `device=... + low_cpu_mem_usage=False` shape and is untouched, and there is no regression test.

Continuing the investigation on #4420 — the root cause still looks un-diagnosed (direct construction works on the same versions, which does not fit `from_pretrained` unconditionally landing on meta).

PR #4425, opened 2026-09-16
fix(codex): preserve rollout event timestamps

### LINKED ISSUE #4423: [coding-agents][Codex] Preserve rollout event timestamps in normalized turns
## Problem

The Codex CLI/Desktop rollout format stores the event time on the JSONL envelope (`timestamp`). The Codex transcript adapter currently discards that field while normalizing messages and tool calls into `TransportTurn` objects.

As a result, downstream retention falls back to the ingestion time. Historical rollouts therefore lose their original chronology, which can reduce the accuracy of temporal retrieval and freshness/obsolescence handling.

## Reproduction

Given a rollout containing records such as:

```json
{"type":"event_msg","timestamp":"2026-01-02T10:00:00Z","payload":{"type":"item_completed","item":{"type":"UserMessage","content":[{"type":"text","text":"inspect the service"}]}}}
{"type":"response_item","timestamp":"2026-01-02T10:00:02Z","payload":{"type":"function_call","name":"exec","arguments":"{\"command\":\"status\"}"}}
{"type":"response_item","timestamp":"2026-01-02T10:00:04Z","payload":{"type":"message","role":"assistant","content":[{"type":"text","text":"The service is healthy."}]}}
```

`readCodexTranscript()` returns turns without timestamps.

## Expected behavior

- Preserve a valid envelope timestamp on normalized user, assistant, and tool-call turns.
- Leave the field omitted when an input record has no timestamp, preserving existing behavior for older rollouts.
- Continue filtering synthetic/developer/reasoning content and avoid retaining raw tool arguments or tool output.
- Keep the change limited to the coding-agents Codex transcript adapter; no server API change is required.

## Environment

- Hindsight coding-agents integration
- Codex CLI/Desktop JSONL rollout transcripts
- Any supported Node.js runtime for the package

## Proposed validation

Add a parser regression fixture covering `event_msg` user input, assistant output, and a tool call with timestamps, plus the existing full package test suite and build.

carlosrve:
## Summary

Preserve Codex rollout envelope timestamps while normalizing JSONL records into `TransportTurn` values.

Historical Codex transcripts carry the time of each event on the outer JSONL record. Keeping that value allows downstream retention to use the time the conversation or tool activity actually occurred instead of the time a backfill happens. Records without a timestamp retain the previous behavior.

The parser continues to omit synthetic/developer/reasoning content and raw tool arguments/output.

## Changes

- Read the optional timestamp from Codex rollout envelopes.
- Attach it to user, assistant, and legacy function-call action turns.
- Add a regression test covering all three normalized turn types.

## Validation

- `npm test -- --run src/core/transcript-codex.test.ts` (10 tests passed)
- `npm test` (73 files, 1035 tests passed)
- `npm run build` (passed)

Fixes #4423


nicoloboschi:
Thanks — reviewed and pushed one follow-up commit to this branch (a91f95a).

The parse is right and it fills a real gap: Codex was the only transcript reader still dropping the envelope time, while the Claude, qwen, droid, dcode, zcode, cursor, copilot and dsh readers all preserve it.

The one thing I changed: `RolloutLine.timestamp` was typed `string` but a rollout line is an unchecked cast from `JSON.parse`, so a non-string value would pass the truthiness check and land on `TransportTurn` — which `chat.ts` writes verbatim into the retained JSONL. I typed it `unknown` and added a `stampOf()` guard (the same `typeof x === "string"` check `transcript.ts`, `transcript-qwen.ts` and `transcript-droid.ts` already do, factored out like `transcript-dsh.ts`'s `stampOf`), used it for all four turn kinds, and added a regression test for a numeric timestamp.

Merging once CI is green.

PR #4426, opened 2026-09-16
fix(codex): support current custom tool calls

### LINKED ISSUE #4424: [coding-agents][Codex] Normalize custom_tool_call rollout events
## Problem

Recent Codex CLI/Desktop rollouts represent tool invocations with `custom_tool_call` records. These records contain a tool `name` and JSON `input`, followed by a `custom_tool_call_output` record.

The Codex transcript adapter only recognizes the legacy `function_call` shape. On current rollouts, tool activity is therefore absent from normalized memory even though user and assistant messages are retained.

## Reproduction

Given a rollout containing:

```json
{"type":"response_item","payload":{"type":"custom_tool_call","name":"exec","input":"{\"command\":\"status\"}","call_id":"call_1"}}
{"type":"response_item","payload":{"type":"custom_tool_call_output","call_id":"call_1","output":[{"type":"text","text":"active"}]}}
```

`readCodexTranscript()` returns no action turn.

## Expected behavior

- Recognize both legacy `function_call` and current `custom_tool_call` records.
- Normalize either shape to the existing compact `role: "action"` representation, using the tool name and primary target only.
- Never retain raw JSON arguments/input or tool output in the normalized transcript.
- Keep the change backward compatible for older Codex rollouts and unrelated harnesses.

## Environment

- Hindsight coding-agents integration
- Codex CLI/Desktop JSONL rollout transcripts
- Any supported Node.js runtime for the package

## Proposed validation

Add a parser regression fixture containing a `custom_tool_call` and its output, assert the compact action turn, and run the full package test suite and build.

carlosrve:
## Summary

Normalize the current Codex `custom_tool_call` rollout shape as a compact action turn.

Current Codex CLI/Desktop transcripts use `custom_tool_call` with a JSON `input` field, while older transcripts use `function_call` with `arguments`. Supporting both shapes keeps tool activity available to Hindsight memory across Codex versions without storing raw arguments or tool output.

## Changes

- Accept `custom_tool_call` alongside legacy `function_call`.
- Parse the corresponding input field and reuse the existing compact action formatter.
- Keep `custom_tool_call_output` excluded from retained turns.
- Add a regression test for a call/output pair.

## Validation

- `npm test -- --run src/core/transcript-codex.test.ts` (10 tests passed)
- `npm test` (73 files, 1035 tests passed)
- `npm run build` (passed)

Fixes #4424


nicoloboschi:
Rebased this onto main (force-pushed 75e7a9a) — #4425 landed first and touched the same two files, so this was CONFLICTING and would have kept getting zero CI runs.

The conflict was purely additive: both PRs inserted a test at the same anchor, so I kept all three. The reader merged cleanly and the new `custom_tool_call` branch picked up the `stampOf(line)` guard from #4425, so current-shape tool calls get the envelope timestamp too.

No changes to your logic. I did consider guarding `p.input` the way #4425 guards `timestamp`, and decided against it: `input` only feeds `JSON.parse` inside a try/catch, so a non-string is already handled, whereas a bad timestamp reached the retained JSONL verbatim.

Merging once CI is green.

PR #4436, opened 2026-09-16
fix(retain): allow monotonic JSON conversation array appends

Tosko4:
## Summary

Allow valid JSON conversation-array appends through the oversized-append monotonicity guard without weakening its protection against losing committed content.

`append_document_body()` correctly merges arrays of message objects into one array. However, `assert_append_extends_stored_body()` only accepts a literal string prefix. The merged array moves the stored body's closing bracket, so a valid oversized append raises `AppendWouldTruncateDocument` and exhausts its retries.

This is the interaction between the JSON-array merging behavior from #2412 and the full-body safety guard from #4229. It reproduces on current `main` as well as v0.10.0.

## Fix

- Keep the existing plain-text prefix check.
- Additionally accept a strictly longer array of JSON objects only when its canonical prefix preserves all previously stored objects in order.
- Compare serialized JSON rather than Python object equality, so changes such as `true` → `1` remain rejected.
- Leave `_is_strict_append_of_stored_document()` and the oversized-replacement metadata-only shortcut unchanged. Broadening that shortcut is not needed for this fix.
- No API, schema, configuration, client or provider changes.

## Regression coverage

- Valid merged conversation arrays and empty-array first appends.
- Refusal of changed, removed, reordered, unchanged, malformed and scalar-array bodies, including boolean/number substitution.
- A blackbox system story through the published Python client: append a JSON message larger than the default batch budget, verify the complete stored array and extracted facts, then append again and verify the accumulated body. Background work is awaited using the existing suite fixtures.

## Validation

- Append invariant tests: **23 passed** with the fix. On the unpatched v0.10.0 runtime, the two new valid-array cases fail and the other 21 pass.
- Append system stories: **4 passed** with the fix.
- The new oversized JSON-array system story fails on unmodified current `main` with `AppendWouldTruncateDocument`.
- Python Ruff checks/formatting, API-project type checking and `git diff --check` pass.
- The repository-wide lint script was attempted; its frontend ESLint step could not load `@eslint/js` in the fresh checkout. No frontend files are changed by this PR. Unrelated formatter output was discarded.

The same guard fix was also validated by retrying an affected operation on a private deployment: previous messages and fact IDs were preserved, missing messages were recovered once, and a recovered fact was recallable. No private data or deployment fixtures are included here.


koriyoshi2041 (changes requested):
The structural prefix check still permits one committed-content rewrite: duplicate object keys are collapsed by `json.loads`. At this exact head, this call returns successfully:\n\n```python\nstored = '[{"role":"user","content":"keep","content":"committed"}]'\nbody = '[{"role":"user","content":"committed"},{"role":"assistant","content":"new"}]'\nassert_append_extends_stored_body(stored, body, document_id="d")\n```\n\nThe first `content` member is gone even though this guard is specifically the last defense against losing committed bytes. I also ran the focused invariant suite: 23/23 pass, so the current cases do not cover this. Could the comparison parse object pairs without collapsing duplicate names (for example `object_pairs_hook`) and reject duplicates, or otherwise preserve each original member before accepting the structural append?

Tosko4:
Thanks @koriyoshi2041, you're right. I missed the duplicate-key case, and your example makes the gap clear.

Fixed in 253d0cef. Both bodies now use an `object_pairs_hook` that rejects duplicate keys before the structural comparison, including duplicates in nested objects and escaped spellings of the same key. Those cases raise `AppendWouldTruncateDocument`; the literal-prefix path is unchanged since it preserves the stored text as-is.

I added your reproducer and five related cases. All six fail on the previous commit and pass with the fix. The full focused suite is now 30/30, and the four append API stories still pass. Python lint and type checks pass too; the full lint script still hits the existing missing `@eslint/js` dependency in my local frontend setup.

Really appreciate you taking the time to try this and share a concrete reproducer.

koriyoshi2041 (approved):
Re-reviewed exact head `253d0cefed6b1abddae2f66551e93d97c329361e`. The duplicate-key guard now runs recursively through `object_pairs_hook` on both stored and candidate JSON, covers escaped-equivalent keys and nested/tail objects, and leaves the byte-preserving literal-prefix path unchanged. The canonical comparison also keeps the `true` versus `1` distinction.

Local exact-head result: `test_append_body_invariants.py` passes 30/30. Hosted checks are green. The append-safety blocker from my prior review is resolved.

Tosko4:
Brief follow-up after checking the current deployment and review state: this remains relevant to long-running conversation documents on v0.10.0. Valid JSON-array appends can reach the size guard during ordinary use, so this is not limited to an artificial oversized-input case.

The original guard fix allowed an affected deployment to recover a rejected append while preserving the previously stored messages and fact IDs. That deployment evidence predates the duplicate-key hardening and should not be read as validation of the entire current head.

For the current head `253d0cef`, the duplicate-key finding is addressed, the exact-head re-review explicitly clears the blocker, and the focused invariant coverage is now **30 tests** rather than the 23 still listed in the original description. The four append API stories also passed in the follow-up validation, and the applicable hosted checks are green.

The change remains limited to append safety, with no schema or API migration. It looks ready for maintainer review and would be useful in the next patch release so downstream deployments can retire the local guard fix.

Issue #3449, opened 2026-08-13
`reasoning_effort` is silently discarded for any model not named gpt-5, o1, or o3


Linodex:
### Bug Description

In `engine/providers/openai_compatible_llm.py`, the reasoning parameter is gated on a
substring match against the model *name*:

```python
# openai_compatible_llm.py:703
def _supports_reasoning_model(self) -> bool:
    """Check if the current model is a reasoning model (o1, o3, GPT-5, DeepSeek)."""
    model_lower = self.model.lower()
    if "deepseek" in model_lower:
        return any(x in model_lower for x in ["v4-pro", "reasoner", "r1", "thinking"])
    return any(x in model_lower for x in ["gpt-5", "o1", "o3"])

# openai_compatible_llm.py:843
if is_reasoning_model:
    call_params["reasoning_effort"] = self.reasoning_effort
```

Every `HINDSIGHT_API_*_REASONING_EFFORT` variable is therefore a silent no-op for
self-hosted reasoning models — exactly the deployments where controlling thinking-token
volume matters most. Nothing is logged when a configured value is dropped.

### Steps to Reproduce

Any OpenAI-compatible endpoint serving a reasoning model whose name does not contain
`gpt-5`, `o1`, or `o3` reproduces this. The example below uses vLLM.

1. **Confirm the backend honours the parameter**, so the endpoint is ruled out as the
   cause. Run both calls directly against it:

   ```bash
   # baseline
   curl -s http://<vllm-host>:8000/v1/chat/completions \
     -H 'Content-Type: application/json' \
     -d '{"model":"<model>","max_tokens":4000,
          "messages":[{"role":"user","content":"List three facts about the Eiffel Tower as JSON."}]}' \
     | jq '.usage.completion_tokens, (.choices[0].message.reasoning｜length)'

   # same call with reasoning_effort
   curl -s http://<vllm-host>:8000/v1/chat/completions \
     -H 'Content-Type: application/json' \
     -d '{"model":"<model>","max_tokens":4000,"reasoning_effort":"none",
          "messages":[{"role":"user","content":"List three facts about the Eiffel Tower as JSON."}]}' \
     | jq '.usage.completion_tokens, (.choices[0].message.reasoning｜length)'
   ```

   Observed: `1073` completion tokens / `3627` reasoning chars versus `91` / `0`. The
   backend clearly supports the parameter.

2. **Point Hindsight at that endpoint** and set the reasoning variable:

   ```
[...]

PR #4438, opened 2026-09-16
fix(llm): omit non-auto tool_choice for Z.AI, direct and gateway-routed

### LINKED ISSUE #4246: [compatibility] Z.AI (Zhipu GLM) rejects any tool_choice other than "auto" with HTTP 400 — "Tool choice must be auto"
## Summary

Hindsight's Reflect agent loop forces a retrieval tool on its first turn by sending
`tool_choice: "required"` (named tool_choice is normalized to `required` with the tools
list pre-filtered to the single forced tool, per `openai_compatible_llm.py`). The Z.AI
(Zhipu) endpoint — used for GLM models, e.g. `z-ai/glm-5.3-flash` — rejects every
`tool_choice` value except `"auto"` with HTTP 400. This is behaviorally identical to
Meta Model API, which is already handled via `_NON_AUTO_TOOL_CHOICE_UNSUPPORTED_PROVIDERS`.

## Reproduction

Via OpenRouter with `provider.only: ["z-ai"]` — the error originates from the Z.AI
backend (see `metadata.provider_name` in the payload):

| Request | Result |
|---|---|
| `tool_choice: {"type":"function","function":{"name":"recall"}}` + tools | ❌ HTTP 400 |
| `tool_choice: "required"` + tools | ❌ HTTP 400 |
| `tool_choice: "auto"` + same single tool | ✅ 200, tool called 5/5 |

Error payload:

```json
{"error":{"message":"Tool choice must be auto","code":400,"metadata":{"provider_name":"Z.AI"}}}
```

The `"auto"` row was verified with 5/5 successful tool calls on memory-retrieval style
prompts (system prompt instructing mandatory tool use, single tool available), so the
model does not need the forced tool_choice to reliably call the tool.

## Impact

Every Reflect (and any other forced-tool) operation against GLM models routed to Z.AI
fails after retries. Retain/Consolidation are unaffected (no forced tool_choice).

Tested on `z-ai/glm-5.3-flash` with Hindsight 0.9.2, September 2026 — still current.

## Suggested fix

Add Z.AI (`zai`) to `_NON_AUTO_TOOL_CHOICE_UNSUPPORTED_PROVIDERS`. The downgrade to auto
is functionally safe here: named tool_choice is already normalized to `required` with the
tools list narrowed to the single forced tool, so under auto the call remains practically
forced — the same reasoning as the existing DeepSeek and Meta branches.

## Workaround currently in use

```
HINDSIGHT_API_REFLECT_LLM_EXTRA_BODY={"tool_choice":"auto"}
```

Verified working (Reflect succeeds end-to-end on Z.AI), but it relies on `extra_body`
merge precedence and is not an API contract. Alternative: route Reflect to a
tool-forcing-compatible host (e.g. Novita).

ebarkhordar:
Adding `zai` to `_NON_AUTO_TOOL_CHOICE_UNSUPPORTED_PROVIDERS` fixes `HINDSIGHT_API_LLM_PROVIDER=zai`, but not the setup in the issue. That one routes GLM through OpenRouter with `provider.only: ["z-ai"]`, so `self.provider` is `openrouter` and the membership test at `:915` is False either way. I checked both against the request that reaches the client (patched `chat.completions.create`, the shape `tests/test_meta_tool_choice.py` uses): with only the set widened, the direct route drops the field and the OpenRouter route still sends `tool_choice: "required"`.

So this does both. `zai` goes in the set, and `_rejects_non_auto_tool_choice` also matches the vendor namespace of the model id, which is where a gateway carries the endpoint: `z-ai/glm-5.3-flash`.

The namespace rather than the model name is deliberate, given #3449. A bare `glm-4.6` is left alone, because those weights can be served by a host that does enforce required tool calls, and there is a test pinning that. `zai/` is in the set as the Requesty spelling and I have not confirmed it against a live Requesty account; `z-ai/` is the one from the report.

I did not call Z.AI. The 400 and the 5/5 tool calls under `auto` are your measurements. What I checked is only which `tool_choice` value Hindsight puts on the request.

`tests/test_zai_tool_choice.py` covers the two routes plus the named-tool-choice narrowing; those six fail on main. The other four are controls for the routes that must keep sending `required`.

Fixes #4246


ebarkhordar:
The namespace match is the part I'd want a second opinion on. Keying `_rejects_non_auto_tool_choice` on the model id's vendor prefix is what lets the OpenRouter route see a Z.AI endpoint at all, since `self.provider` is `openrouter` there and the provider set never matches. The cost is that `z-ai/glm-5.3-flash` gets handled by name while a bare `glm-4.6` deliberately does not, per #3449.

The other loose end is `zai/`. It is in the set as the Requesty spelling and I have not confirmed it against a live Requesty account, so if you'd rather carry only `z-ai/` from the report, I'll drop it.

PR #4448, opened 2026-09-16
fix(client): cancel capacity-retry backoff on abort

### LINKED ISSUE #4445: TS/Python clients: capacity-retry backoff is not cancellable — an aborted caller still waits out Retry-After and re-sends
## Problem

`retryOnCapacity` (added in #4253) retries 429/503 with `Retry-After` and full jitter. Its wait is a plain timer:

`hindsight-clients/typescript/src/index.ts:116-131`
```ts
export async function retryOnCapacity<T extends { response?: Response }>(
  send: () => Promise<T>,
  maxAttempts: number,
  random: () => number = Math.random
): Promise<T> {
  let result = await send();
  for (let attempt = 1; attempt < maxAttempts; attempt++) {
    …
    await new Promise((resolve) => setTimeout(resolve, random() * wait));   // not abortable
    result = await send();
  }
  return result;
}
```

`recall` and `reflect` already accept `options.signal` and pass it to the underlying `fetch`, so the *in-flight request* is cancellable. But the signal stops at the transport: it is not passed to `retryOnCapacity`, and the helper has no `signal` parameter at all.

Consequences for a caller that aborts (deadline elapsed, service shutting down):

1. the abort does not settle the operation promptly — it waits out the remaining backoff first;
2. **worse, a new HTTP attempt is issued after the abort**, because nothing re-checks cancellation before the next `send()`. A caller that has already given up and moved on can still generate load, and receive a late result.

This matters most for integrations that race recall against a deadline. The OpenClaw plugin is the live example: its recall races a timer and returns, while the underlying request — and now potentially a retry of it — keeps going.

## What to change

- `retryOnCapacity` takes an `AbortSignal`: `throwIfAborted()` before and after each `send()`, and an abortable wait (settle on `abort`, with timer and listener cleanup) instead of the bare `setTimeout`.
- `recall` / `reflect` thread their existing `options.signal` into it.
- Keep the current max-attempt / status / jitter policy and the no-retry-on-writes rule unchanged.

## Python parity

`hindsight-clients/python/hindsight_client/` should be checked in the same change. Its equivalent wait uses `asyncio.sleep`, which is already cancellable, so it may need nothing — but the two wrapper clients are expected to expose the same surface (see `.claude/skills/code-review/SKILL.md` step 7a), so whichever way it lands should be deliberate and stated.

## Tests

Deterministic and cheap: abort during the backoff and assert (a) the returned promise settles promptly as aborted, and (b) no second request is issued. A fake timer plus a counting `send` covers both.

## Context

Split out of #4396, which implemented this together with OpenClaw-side recall lifecycle changes. This issue is the client half only.

Tosko4:
## Summary

Fixes #4445. Split out of #4396 as requested in https://github.com/vectorize-io/hindsight/pull/4396#issuecomment-5700067497.

Make the TypeScript client's capacity-retry backoff cancellable for **all SDK consumers**, independently of any integration:

- Thread the existing recall/reflect `options.signal` into `retryOnCapacity`.
- Check cancellation before and after each send.
- Interrupt backoff on abort and remove its timer/listener on either outcome.
- Preserve the max-attempt, retryable-status, Retry-After and jitter policies. Writes remain non-retrying.

No OpenClaw, server, schema, generated-client or dependency-version changes.

## Python parity

The Python wrapper already uses cancellable `asyncio.sleep` and does not swallow `CancelledError`. No Python implementation change is needed. Two added real-task cancellation tests cover abort during 429/503 backoff and verify no second send.

## Verification

- Current-main regression: the original TypeScript implementation fails all 7 added cancellation cases while its 11 existing retry tests pass. Type-check diagnostics were disabled only for this baseline run because the new optional helper argument is absent on main.
- TypeScript deterministic suite: **85 passed**; live-backend `main_operations.test.ts` excluded.
- Python retry suite: **13 passed**, including the two added cancellation parity cases.
- TypeScript build and full repository `scripts/hooks/lint.sh`: passed.

Each fix is tested at its owning package boundary. This PR can land and release independently of the separate OpenClaw deadline and lifecycle fixes. No production runtime or memory data changed.


koriyoshi2041 (approved):
Verified at `bbd9d6f3c5359e1f54304a260a19f27354000da6`. The TypeScript abort path clears the pending timer, removes its listener on both outcomes, and guards the pre-send / post-response races; recall and reflect both propagate the caller signal. Focused TypeScript retry tests pass 18/18, the client build including declarations passes, and the Python parity suite passes 13/13. This also supersedes my duplicate #4451 cleanly.

PR #4449, opened 2026-09-16
fix(openclaw): forward recall deadlines to the client transport

Tosko4:
## Summary

Split the **OpenClaw deadline/signal plumbing** out of #4396, following https://github.com/vectorize-io/hindsight/pull/4396#issuecomment-5700067497.

`scopeClient().recall()` currently races a timer without cancelling the HTTP request. The comment claiming the client cannot take a per-call signal is stale: the existing published client already accepts it.

- Create a per-call AbortController and pass its signal to the client.
- Abort it at the configured deadline with the existing `TimeoutError` contract.
- Keep the bounded race for transports that ignore cancellation.
- Clear the deadline and abort listener in `finally`, on success or failure.

**Only the OpenClaw package changes.** No service generation/controller policy, in-flight recall map behavior, client retry logic, dependency versions or retain behavior changes.

## Verification

- Both initial deadline regressions fail against current main: missing transport signal and leftover timer after success.
- **358 plugin tests pass** against the existing published `@vectorize-io/hindsight-client@0.8.6`, unchanged.
- Added coverage: timeout despite an ignoring transport, success/failure cleanup, no-deadline calls, and a real loopback HTTP request whose connection closes at the deadline.
- Plugin TypeScript build and full repository `scripts/hooks/lint.sh` pass.

This stands on the existing client signal support. Promptly aborting the newer SDK's capacity-retry *sleep* is separately addressed by #4448; this PR does not claim to fix that SDK implementation.

## Relationship to lifecycle work

The separate lifecycle PR #4450 owns start/stop invalidation, pre-start recall ownership and in-flight deduplication. Both are independently tested against main. They meet in `scopeClient().recall()`, so merging either first can require resolving that small adapter overlap in the other; neither PR contains the other's commits. The combined implementation was tested separately with 376 passing plugin/composition tests, including current-source-SDK HTTP scenarios.

No production service, configuration or memory data changed.


nicoloboschi:
Reviewed and merging this. Thanks for splitting #4396 up — this half is much easier to reason about on its own.

I verified the dependency question before anything else, since it decides whether this can ship without a client release: openclaw resolves the **published** client (`^0.8.6`, lockfile pins 0.8.6 from the registry, and this package is not in the root workspaces list so there is no symlink to the in-repo 0.10.0). That 0.8.6 build already forwards `options.signal` in `recall`, and AbortSignal support has been there since #1198 shipped in v0.6.0 — so your claim that the old comment was stale is right, and this needs no client release and no dependency bump.

I pushed four small things to your branch rather than round-tripping:

1. **Dropped `signal.throwIfAborted()` right after `new AbortController()`.** Nothing can abort it there, so it was unreachable — and it reads as if a caller-supplied signal were being honoured, which would mislead whoever adds one later.
2. **Named the concrete case the retained `Promise.race` is for.** It is not hypothetical: from client 0.10.x `recall` routes through `retryOnCapacity`, whose backoff sleeps in a plain `setTimeout`, so an abort during a 429/503 wait will not interrupt it and the race is the only thing bounding the hook. Keeping the race is right; the comment now says why.
3. **Tightened the timeout assertion** to `toBeInstanceOf(DOMException)` alongside the name check. The caller classifies with `error instanceof DOMException && error.name === "TimeoutError"`, so a plain `Error` with the name reassigned would have passed the old `toMatchObject` while turning every auto-recall timeout into a logged error instead of a quiet skip.
4. **Dropped the `Retry-After` fixture header** — 0.8.6 has no capacity-retry logic to read it and that test's recall handler never responds, so it implied behaviour the test does not exercise. Also documented `recallTimeoutMs` in the README table, including the 1000ms floor below which it is ignored, now that the setting aborts the request rather than merely abandoning it.

One thing I tried and reverted: lowering the HTTP test's `recallTimeoutMs` to widen its timing margin. `getPluginConfig` drops anything below 1000 and the hook then falls back to the 10s default, so the test stopped exercising the deadline at all. 1000 is the floor; there is now a comment saying so.

Rebased onto current `main` (it had drifted three commits). Suite is 358 green, `tsc` and prettier clean.

PR #4450, opened 2026-09-16
fix(openclaw): invalidate recall work across service generations

Tosko4:
## Summary

Split the **OpenClaw recall lifecycle/generation change** out of #4396, as requested in https://github.com/vectorize-io/hindsight/pull/4396#issuecomment-5700067497.

A recall started in an old service generation can currently return context after stop/restart, reuse old in-flight work, or let old promise cleanup remove its successor. This PR treats those as lifecycle invariants rather than a side effect of timeout handling.

## Changes and invariants

- Capture recall generation and its controller at hook entry, then check ownership after readiness, bank setup and response settlement.
- Preserve external-API lazy recall before the first service start, using a recall-only `preServiceRecallController`. First start/stop invalidates that lifetime; pre-start retain gating stays unchanged.
- Abort and clear old recall work on start/stop, including restart without a preceding stop.
- Forward the captured service signal through bank-scoped recall. Bound the hook wait even if transport ignores cancellation and clean up timers/listeners when it settles.
- Remove an in-flight entry only when it still holds that exact promise, protecting a successor with the same key.

**Only the OpenClaw package changes.** No TypeScript client retry changes, dependency bump, retain changes, or deadline-triggered transport cancellation: the separate deadline PR owns that last behavior. The existing timeout rejection contract is preserved here.

## Verification

- Against current main, 5 of the 8 initial lifecycle/lazy-recall tests fail; the 3 unaffected behavior checks pass.
- **363 plugin tests pass** with the existing published `@vectorize-io/hindsight-client@0.8.6`, unchanged.
- Coverage includes late success after stop, stop while waiting for initialization, restart with/without stop, successor deduplication/cleanup, pre-start success, first-start/stop cancellation, pre-aborted calls and never-settling transports.
- Plugin TypeScript build and full repository `scripts/hooks/lint.sh` pass; final test-only signal-spy correction passes focused tests, build and formatting.

## Independent review/release scope

This PR targets main directly. Generic SDK backoff cancellation is #4448 and is not required to suppress stale context or cancel an in-flight fetch with the published client.

The separate deadline PR #4449 also edits `scopeClient().recall()`. Merging either first can require resolving that small adapter overlap in the other; neither branch includes the other's commits. A separate combined implementation passes **376 plugin/composition tests**, including transient 429/503 recovery, no retry after timeout/stop with the fixed source SDK, and closing an unanswered HTTP request at deadline.

No production service, configuration or memory data changed.


nicoloboschi:
Reviewed and rebasing this onto #4449, which just merged. Both PRs rewrote the same `scopeClient().recall` block, so the conflict was guaranteed — I resolved it by **combining** the two rather than picking a side, since they cancel for different reasons:

```ts
const deadline = new AbortController();
const effective = signal ? AbortSignal.any([signal, deadline.signal]) : deadline.signal;
```

Your service-owned signal handles stop/restart; #4449's deadline controller handles the timeout; the composed signal goes to the transport. That preserves both reasons where callers classify on them — a `TimeoutError` for the deadline, an `AbortError` for a service stop — and it matches the `AbortSignal.any` pattern already used elsewhere in this file. The retained `Promise.race` stays as the backstop for a transport that ignores cancellation.

I also applied one review finding, folded into the rebase commit (flagging it since the commit carries your authorship):

**The shutdown path was reporting a timeout that never happened.** With generation-based cancellation, a service stop now rejects with `AbortError`, which landed in this branch:

```ts
log.warn(`Auto-recall aborted after ${recallTimeoutMs ?? DEFAULT}ms, skipping memory injection`)
```

Before your PR the only `AbortError` source was the deadline, so quoting the timeout was true. Now every shutdown with an in-flight recall would claim "aborted after 1000ms" with a number unrelated to the elapsed time. It now branches on `recallController?.signal.aborted` and logs a service-stop cancellation at debug level, keeping the timeout wording for the real deadline case.

The rest of the review came back clean — I could not construct an interleaving that defeats the generation guards, the in-flight map discipline is complete on every exit path (resolve, error, abort), pre-service lazy recall still works with a one-shot lifetime, and stop→start is not sticky. Nice work on the tests: each one fails without the fix, and they assert observable invariants rather than internals.

389 tests green across both PRs' suites, `tsc`, prettier and `./scripts/hooks/lint.sh` clean.

PR #4451, opened 2026-09-16
fix(ts): make capacity retry backoff cancellable

### LINKED ISSUE #4445: TS/Python clients: capacity-retry backoff is not cancellable — an aborted caller still waits out Retry-After and re-sends

koriyoshi2041:
## Summary

- make the TypeScript capacity-retry backoff settle immediately when its caller aborts
- prevent a new recall/reflect request from being sent after cancellation
- clean up retry timers and abort listeners on both cancellation and successful wake-up

## Python parity

The Python wrapper already waits with `asyncio.sleep`, so cancelling the calling task interrupts its backoff without an additional wrapper API. This change is limited to the TypeScript wrapper's existing `AbortSignal` surface.

## Tests

- `npm test --workspace=@vectorize-io/hindsight-client -- --runInBand tests/retry.test.ts` (18 passed)
- `npm run build --workspace=@vectorize-io/hindsight-client`
- `npx tsc --noEmit -p hindsight-clients/typescript/tsconfig.json`
- `npx prettier --check hindsight-clients/typescript/src/index.ts hindsight-clients/typescript/tests/retry.test.ts`

Fixes #4445


koriyoshi2041:
Closing as superseded by #4448, which landed first and covers the same TypeScript cancellation fix plus Python parity tests. I rechecked the implementation overlap; keeping both open would only duplicate maintainer work.

Issue #4175, opened 2026-09-07
GET stats / memories/list for a nonexistent bank returns 200 with empty stats instead of 404


bunnyfu:
## Summary

Read endpoints scoped to `{bank_id}` return HTTP 200 with empty/zero payloads for bank IDs that do not exist, instead of 404. Reproduced 2026-09-06 against a local ghcr.io/vectorize-io/hindsight deployment:

```
GET /v1/default/banks/<nonexistent-id>/stats
→ HTTP 200
{"bank_id":"<nonexistent-id>","total_nodes":0,"total_links":0,"total_documents":0,
 "nodes_by_fact_type":{},"links_by_link_type":{},"pending_operations":0,...}

GET /v1/default/banks/<nonexistent-id>/memories/list?limit=1
→ HTTP 200
{"items":[],"total":0,"limit":1,"offset":0}
```

In contrast, `GET /v1/default/banks` correctly lists only banks that actually exist.

The published OpenAPI spec declares only `200` and `422` responses on the `/v1/default/banks/{bank_id}/**` operations — no `404` is declared on any bank-scoped path — so this is consistent with the spec as published, but the behavior itself is a trap for API consumers.

## Why this matters

Any consumer that validates bank health or content by reading stats/list silently guards nothing after the bank is renamed, deleted, or recreated under a different ID: the request keeps returning 200 with empty counters, indistinguishable from "healthy but empty bank". Concretely:

- a monitoring job asserting `nodes_by_fact_type` stays within bounds passes forever against a missing bank,
- an empty-but-missing bank is indistinguishable from a real, freshly-created bank,
- typos in `bank_id` are never surfaced.

We hit this while building a health monitor for one of our banks: a stats-only check continued to pass after the bank it watched was recreated under the same ID on a fresh engine state, and the monitor had to add an explicit existence check against the bank list to detect the mismatch.

## Request

Return 404 (or at minimum a distinct body/marker, e.g. `{"exists": false}`) from bank-scoped GETs when the bank does not exist — at least for `/stats` and `/memories/list`, which are the natural monitoring surfaces. Also declare the 404 response in the OpenAPI spec so generated clients can handle it. If keeping 200 is a deliberate design choice (e.g. bank auto-creation semantics on write paths), documenting it would help consumers avoid building monitors that silently no-op.

koriyoshi2041:
[...]

PR #4456, opened 2026-09-16
fix(recall): 404 for a bank that was never created, before any retrieval

### LINKED ISSUE #4442: recall answers for a bank that was never created: full retrieval fan-out, then 200 with empty results instead of 404
### What

`recall` is the only bank-scoped read endpoint with no bank-existence check. Recalling a `bank_id` that was never created runs the **entire** retrieval pipeline against the memory tables and then returns `200` with empty results, instead of `404`.

### Why this is wrong

`MemoryEngine._require_bank_exists` already exists for exactly this case, and its docstring states the rule:

> Read endpoints must not answer for a bank nobody created. A 200 with empty counters is indistinguishable from a healthy, empty bank, so a typo'd, renamed or deleted `bank_id` silently passes any monitor built on `/stats` or `/memories/list` (#4175).

Every other bank-scoped read calls it — `/stats`, `/memories/list`, the document and chunk reads, the mental-model and directive reads (~15 call sites in `memory_engine.py`). `recall_async` does not, and neither does the HTTP handler, so a typo'd or deleted bank id looks like a healthy empty bank on the one endpoint most likely to be called in a loop.

### It is also not free

`PostgresMemories.recall_unified` has no existence check either, so for a bank that does not exist the request still pays the full fan-out (`hindsight_api/engine/memories/postgres.py:82`):

1. acquires a pool connection and runs the dense + BM25 UNION query,
2. runs the temporal query on the same connection when a window is given,
3. releases it, then runs the **graph arm per fact_type in parallel** on the pool, each acquiring its own connection.

Every arm filters `WHERE bank_id = $1` and matches nothing, so the work — vector scan, BM25, graph traversal, and several pool connections — is spent to return an empty list. With the default `types`, that is one fan-out per fact type. An unauthenticated-ish caller probing bank ids gets a fairly cheap way to occupy the pool.

### Suggested fix

Add the guard at the top of `recall_async`, alongside the other bank-scoped reads:

```python
await self._require_bank_exists(bank_id)
```

That returns `404` before any arm runs. It is the same one-line change the other read endpoints already carry, and the profile row is process-cached, so an existing bank pays no extra query.

### Note on tests

Any test asserting that recall of an unknown bank returns empty results is asserting the behaviour #4175 ruled out and would need updating to expect `404`.

### References

- `hindsight-api-slim/hindsight_api/engine/memory_engine.py:13591` — `_require_bank_exists`
- `hindsight-api-slim/hindsight_api/engine/memory_engine.py:7411` — `recall_async`, no guard
- `hindsight-api-slim/hindsight_api/engine/memories/postgres.py:82` — `recall_unified`, no guard
- `hindsight-api-slim/hindsight_api/api/http.py:5802` — HTTP recall handler, no guard

Observed on `main` at 590657901.

cdbartholomew:
## What

`recall` was the only bank-scoped read with no bank-existence check. Recalling a `bank_id` that was never created ran the **entire** retrieval fan-out and returned `200` with empty results, instead of `404`.

Adds `await self._require_bank_exists(bank_id)` at the top of `recall_async` (after auth), the same guard the other ~15 bank-scoped reads already carry.

## Why

`_require_bank_exists` already exists and states the rule (#4175): a `200` with empty counters is indistinguishable from a healthy empty bank, so a typo'd, renamed or deleted `bank_id` silently passes any monitor built on `/stats` or `/memories/list`. `recall` is the endpoint most likely to be polled in a loop, so it's the worst place to have the gap — and it was also the only one still paying the full dense + BM25 + per-fact-type-graph fan-out (several pool connections) to return nothing.

The guard declines with `404` before any arm runs. The profile row is process-cached, so an existing bank pays no extra query. The HTTP handler delegates to `recall_async`, so it's covered without a separate change.

## Test

`test_recall_returns_404_for_a_bank_that_was_never_created` — recall of a never-created bank returns `404 "Bank '...' not found"`. Existing `test_recall_empty_bank` is unaffected: it creates the bank first (`get_bank_profile`, `create_if_missing=True`), so recall on an existing-but-empty bank still returns `200` empty.

Note: any downstream test asserting that recall of an unknown bank returns empty results is asserting the behaviour #4175 ruled out and should be updated to expect `404`.

Fixes #4442.


cdbartholomew:
Superseded upstream: #4465 added the recall 404 and #4487 moved the existence check after the operation validator (the ordering this PR's later revision also converged on). Closing in favor of those merged fixes.

PR #4458, opened 2026-09-17
fix(recall): let zero disable query token limit

### LINKED ISSUE #4455: HINDSIGHT_API_RECALL_MAX_QUERY_TOKENS=0 breaks all recalls despite docs saying it disables the limit
**Bug Description**

The configuration docs state that `HINDSIGHT_API_RECALL_MAX_QUERY_TOKENS=0` disables the recall query token limit (`0` disables the limit). In hindsight-api 0.10.0 the validation instead treats `0` as a literal maximum, so every recall request is rejected.

**Steps to Reproduce**

1. Set `HINDSIGHT_API_RECALL_MAX_QUERY_TOKENS: "0"` in the environment of hindsight-api 0.10.0
2. Restart the container
3. Run any recall (e.g. via the OpenClaw integration auto-recall, or POST /v1/default/banks/{bank}/recall with a normal query)

Observed log output:

```
auto-recall error: recall failed: "Query too long: 284 tokens exceeds maximum of 0. Please shorten your query."
auto-recall error: recall failed: "Query too long: 399 tokens exceeds maximum of 0. Please shorten your query."
```

**Expected**

Per documentation, `0` should disable the limit and recalls should succeed regardless of query length.

**Actual**

All recall requests fail with `Query too long: N tokens exceeds maximum of 0`. This effectively disables memory recall entirely until the variable is set to a positive value.

**Version**

- hindsight-api: 0.10.0-slim (ghcr.io/vectorize-io/hindsight-api:latest-slim), Docker Compose, Postgres 18 (vchord extension image)
- Client integration: @vectorize-io/hindsight-openclaw 0.12.0

**LLM Provider**

OpenRouter (~z-ai/glm-flash-latest for extraction; qwen/qwen3-embedding-8b for embeddings; voyage rerank-2.5-lite + cohere rerank-4-fast for reranking). The failure is independent of the LLM/embedding provider — it is a validation error before any model call.

Workaround: set a positive value (e.g. `1024`).

koriyoshi2041:
Fixes #4455.

## Problem
`HINDSIGHT_API_RECALL_MAX_QUERY_TOKENS=0` is documented as disabling the recall query limit, and the engine's internal recall path already treats it that way. The REST handler instead compared every query against zero, so all non-empty API recalls failed with HTTP 400.

## Fix
Apply the token ceiling only when the configured maximum is positive. This keeps the REST boundary aligned with the internal recall path and the documented configuration contract.

## Test
- Added regression coverage that zero disables the REST cap.
- Kept a positive-limit counterexample to prove oversized queries are still rejected.
- `uv run pytest tests/test_recall_query_token_cap.py -n 0 -q` — 9 passed.
- Focused Ruff lint and format checks passed; `git diff --check` passed.

## Risk
Low. The change affects only the explicit zero sentinel; positive limits retain the existing rejection behavior and defaults are unchanged.


koriyoshi2041:
Closing in favor of #4459. Its production guard is equivalent, and its HTTP-route plus public-client blackbox coverage proves the documented zero sentinel across the real boundary more directly. I independently reran the route + engine cap suites (14/14) before approving it.

PR #4459, opened 2026-09-17
fix(api): honor disabled recall query token limit

### LINKED ISSUE #4455: HINDSIGHT_API_RECALL_MAX_QUERY_TOKENS=0 breaks all recalls despite docs saying it disables the limit

heyallencao:
## Summary

Fixes #4455.

Honor `HINDSIGHT_API_RECALL_MAX_QUERY_TOKENS=0` at the HTTP recall boundary, matching the documented behavior and the engine's existing zero-means-unlimited handling.

## Problem and approach

The HTTP handler compared the query's token count directly with the configured maximum. With a maximum of zero, every non-empty query therefore returned HTTP 400 instead of disabling the limit.

Apply the rejection check only when the configured maximum is positive. The production change is a single condition; the default limit and positive-limit behavior remain unchanged. This does not truncate the caller's query or change authentication, model adapters, or request/response schemas.

## Regression coverage

- Add seven HTTP route tests covering zero, the default limit, and queries below, at, and above positive limits. These use the real FastAPI app and tokenizer, with the downstream engine mocked. Accepted queries reach the engine unchanged; rejected queries never invoke it.
- Allow the system-test server launcher to accept explicit environment overrides without changing existing callers.
- Add a blackbox test that retains a memory, waits for background work, and recalls it with a 751-token query through the public Python client. It uses a separate PostgreSQL instance and a temporary port.

## Validation

- Before the fix: both zero-limit HTTP cases fail, while the other five pass. The blackbox test fails with HTTP 400 (`751 tokens exceeds maximum of 0`). After the fix, all pass.
- HTTP and engine query-limit tests: **14 passed**.
- Selected blackbox regression tests covering retain/recall, retrieval strategies, output-token budgets, recall budgets, and the new query-limit case: **9 passed**.
- Ruff lint, formatting, and type checks for all four changed files passed, as did `git diff --check`.
- Generated OpenAPI is identical before and after the change; no generated client changes are needed.

The blackbox tests run the real API and PostgreSQL with the project's existing strict LLM, embedding, and reranker stubs. They do not require external model credentials or disable background work.

## Related PR

This overlaps with #4458, which was opened while this change was being prepared. This version uses a direct guard in the handler and adds HTTP-route and client-driven blackbox coverage. The regression tests can also be adapted to that fix if preferred.


koriyoshi2041 (approved):
Verified at `5908c2e4669361ed4738a734ce5f59d7f2ef8b5b`. The direct positive-limit guard matches the documented zero sentinel while preserving default and positive-cap behavior. I ran the new HTTP route matrix together with the existing engine cap suite (14/14), plus focused Ruff/format and diff checks. The hosted blackbox/system and generated/OpenAPI gates are also green. This is the stronger landing path; I am closing my overlapping #4458 in favor of it.

nicoloboschi:
fixed by #4475

PR #4461, opened 2026-09-17
fix(worker): retry transient foreign key violations

### LINKED ISSUE #4453: retain: ForeignKeyViolationError from the delete_document race is classified non-retryable (#980 blanket), dropping retains at retry_count=0
## Environment

- hindsight-api **0.9.1**, PostgreSQL backend, pgvector, single dataplane process
- Verified the same code is present in **0.9.2** and **0.10.0** (latest)

## Summary

`_is_non_retryable_task_error` (`engine/memory_engine.py`) treats every
`asyncpg.IntegrityConstraintViolationError` as deterministic. That blanket was
introduced by #980, whose motivating case was `UniqueViolationError` on
`pk_chunks` — genuinely deterministic.

But `ForeignKeyViolationError` raised by the retain replacement path is **not**
deterministic: it is a concurrency race whose retry succeeds. Because the
blanket catches it, those retains go terminal at `retry_count = 0` and the
conversation content is dropped with no further attempt.

## The race

`delete_document` documents the race itself:

> *"Called when a document is replaced, so it races the replacement's writes: it
> must remove only what was written before this call, never the facts arriving
> moments later"*

When two retains for the same `document_id` overlap, one runs
`DELETE FROM memory_units WHERE document_id = $1 AND bank_id = $2` while the
other inserts `unit_entities` rows referencing units in that set. PostgreSQL's
RI check on the parent delete then raises, despite the constraint being
`ON DELETE CASCADE` (the cascade removes children visible at delete time; a
concurrently committed child still trips the final check):

```
ForeignKeyViolationError: update or delete on table "memory_units" violates
foreign key constraint "fk_unit_entities_unit_id_memory_units" on table "unit_entities"
DETAIL: Key (id)=(...) is still referenced from table "unit_entities".
```

Overlap is easy to reach whenever extraction is slow: retains for one long-lived
append-mode document queue up behind each other. In our case the same condition
also drove the `[delta] ... was modified by concurrent request — aborting delta,
falling back to full retain` path, which widens the delete window further.

## Evidence that it is transient, not deterministic

From one deployment's `async_operations`:

| failure class | total | `retry_count = 0` |
|---|---:|---:|
| `fk_unit_entities_unit_id_memory_units` | 690 | **682** |
| `Fact extraction failed` (LLM) | 770 | 132 |

LLM failures retry normally (638 with `retry_count > 0`); the FK class goes
terminal on first contact.

Recovery confirms transience:

- one operation retried by hand via `POST /operations/{id}/retry` → **completed
  on the first attempt**
- a controlled drain of the accumulated backlog (382 failed retains across 73
  documents, never two of the same document in flight) is running at
  **56 recovered, 0 failures** so far

Nothing about the data changed between the terminal failure and the successful
retry — only the concurrency.

## Impact

Silent memory loss. 327 retains in this deployment were discarded without a
single retry; the transcripts they carried never reached the bank, and nothing
surfaces to the caller because the retain API is `async: true`. The backlog is
recoverable only by manually driving `/operations/{id}/retry`.

## Suggested fix

Narrow the classification so a FK violation is retryable while the genuinely
deterministic siblings stay terminal:

```python
def _is_non_retryable_task_error(e: Exception) -> bool:
    """Classify deterministic task failures that should skip worker retry."""
    if isinstance(e, asyncpg.exceptions.ForeignKeyViolationError):
        # Concurrency race (see delete_document), not bad data: retry succeeds.
        return False
    return (
        isinstance(e, asyncpg.exceptions.IntegrityConstraintViolationError)
        or _is_oracledb_integrity_error(e)
        or _is_invalid_embedding_dimension_error(e)
    )
```

`UniqueViolationError`, `NotNullViolationError`, `CheckViolationError`,
`ExclusionViolationError` and `RestrictViolationError` keep their current
behaviour, so #980's original motivation is preserved.

A deeper fix would serialize retains per `document_id` end to end. Operations
carry a `serialization_key`, but ~40% of the backlog here predates it being
populated, and the guard does not cover writers outside the retain queue.

## Not a duplicate of

- **#3227** — different constraint (`fk_unit_entities_entity_id_entities`),
  cause was deadlock rollback; that one *was* retried at task level
- **#2662** — entity side, `prune_orphan_entities`
- **#4251** — `DeadlockDetectedError` on the explicit document-delete endpoint

koriyoshi2041:
## Problem

PostgreSQL foreign-key violations are currently grouped with deterministic integrity errors. The retain replacement path can instead hit one transiently when a document delete overlaps a child-row insert, causing the operation to fail permanently without using its retry budget.

## Fix

- Keep deterministic integrity violations such as unique/check/not-null failures terminal.
- Route `ForeignKeyViolationError` through the existing worker retry path.
- Update the worker-level regression to assert the operation remains pending and raises `RetryTaskAt`.

## Test

- `uv run pytest tests/test_integrity_violation_not_retried.py -q`
- `uv run ruff check hindsight_api/engine/memory_engine.py tests/test_integrity_violation_not_retried.py`
- `uv run ruff format --check hindsight_api/engine/memory_engine.py tests/test_integrity_violation_not_retried.py`
- `git diff --check`

Fixes #4453


hindsight-api-slim/hindsight_api/engine/memory_engine.py
```diff
@@ -1393,7 +1393,13 @@ def _entity_map_from_results(
...
     return (
-        isinstance(e, asyncpg.exceptions.IntegrityConstraintViolationError)
+        (
+            isinstance(e, asyncpg.exceptions.IntegrityConstraintViolationError)
+            # A parent delete can lose a race with a concurrent child insert.
+            # Retrying after the overlapping transaction finishes succeeds, so
+            # do not turn every PostgreSQL foreign-key violation terminal.
+            and not isinstance(e, asyncpg.exceptions.ForeignKeyViolationError)
+        )
         or _is_oracledb_integrity_error(e)
```
ebarkhordar: The narrowing lands on the asyncpg branch only, so this line is still reached for Oracle, and `oracledb.IntegrityError` covers the Oracle form of the same race: ORA-02292 "child record found" is what a parent delete raises there (ORA-02291 is its partner). If that race reaches the classifier on an Oracle deployment, the retain still goes terminal at `retry_count = 0`.

I ran the classifier at head `7f616846` in a clean `python:3.12-slim` container, with the three functions sliced out of `memory_engine.py` by ast rather than retyped:

```
case                                      exc class                 verdict
postgres FK   (the #4453 race)            ForeignKeyViolationError  retried
postgres UNIQUE (the #980 case)           UniqueViolationError      TERMINAL
oracle   FK ORA-02292 child record found  IntegrityError            TERMINAL
oracle   UNIQUE ORA-00001 (the #980 case) IntegrityError            TERMINAL
```

`ERR_INTEGRITY_ERROR_CODES` is `[1, 1400, 1438, 2290, 2291, 2292, 21525, 40479]`, so 2291 and 2292 are the two matching `ForeignKeyViolationError`. One edit here:

```python
or (_is_oracledb_integrity_error(e) and _ora_code(e) not in (2291, 2292))
```

One trap if you take it: the code is not an attribute of the exception. `getattr(e, "code", None)` returns None for all of them, which keeps every Oracle integrity error terminal and makes the change a silent no-op. It reads off `e.args[0].code`.

`_is_oracledb_integrity_error` also has a second caller at :21664, whose comment says only a unique violation should qualify, so narrowing at this call site rather than in the helper leaves that one alone.

I have not run this against a real Oracle database. What I measured is the classification, not the race.

koriyoshi2041: Addressed in `6b413cfe`. The worker classifier now reads the Oracle code from `e.args[0].code` and exempts ORA-02291/02292 at this call site only; the replay/unique-violation caller remains unchanged. Added counterfactual coverage for both FK codes and the six existing terminal integrity codes.

Local: 19 selected classifier/update tests passed; focused Ruff lint/format and diff check passed. The full file additionally reached 19 passes, but its three embedded-Postgres cases could not start because `pg0-embedded` is absent in this fresh worktree.

ebarkhordar: Confirmed at `6b413cfe`. I re-ran the classifier slice against the real `oracledb` 26.0.0, since the new test monkeypatches `sys.modules["oracledb"]` and so cannot tell you whether the live library really carries the code at `args[0]`. It does, and the split lands where you meant it:

```
postgres FK   ForeignKeyViolationError  retried
postgres UNIQUE UniqueViolationError    TERMINAL
oracle ORA-02292                        retried
oracle ORA-02291                        retried
oracle ORA-00001                        TERMINAL
oracle ORA-02290                        TERMINAL
```

`_oracledb_error_code` also covers two cases my suggested predicate did not: an `IntegrityError` with empty `args`, and one whose `args[0]` is a plain string rather than an `_Error`. Both return None and stay terminal, so behaviour is unchanged wherever the code cannot be read.

I still have not run this against a real Oracle database. What I measured is the classification.

koriyoshi2041:
Closing this now that #4516 has landed the fix with the production race regression and Oracle coverage. Thanks for carrying this through.

PR #4485, opened 2026-09-17
fix(mental-models): absorb a hallucinated extra field on a delta op instead of failing the refresh

### LINKED ISSUE #4443: parse_delta_operation_list rejects ops carrying an extra hallucinated field (extra_forbidden); lone-field ops make the whole delta refresh fail (delta_ops_failed)
### Bug Description

On `0.10.0`, during a mental-model delta refresh, a single operation that carries a field not declared in the op schema fails pydantic validation with `extra_forbidden`, because every op model runs with `extra="forbid"`. When a refresh round emits operations all of which carry such a field, all are dropped, `mode_fallback_reason` is set to `delta_ops_failed`, and the refresh raises `MentalModelRefreshError` instead of writing.

Concretely, `ReplaceSectionBlocksOp` declares `section_id` and `blocks` only, but the model emitted an extra `blocks_note` field:

```python
class ReplaceSectionBlocksOp(_OpBase):
    op: Literal["replace_section_blocks"] = "replace_section_blocks"
    section_id: str
    blocks: list[str] = Field(default_factory=list)
```

Model output:

```json
{
  "op": "replace_section_blocks",
  "section_id": "user-preferences",
  "blocks": ["..."],
  "blocks_note": "merged"
}
```

Validation rejects the whole operation, not just the stray field:

```
[STRUCTURED_DELTA] skipping invalid operation at index 0:
  [{'type': 'extra_forbidden', 'loc': ('replace_section_blocks', 'blocks_note'),
    'msg': 'Extra inputs are not permitted', 'input': 'merged'}]
[MENTAL_MODELS] Structured delta failed for <model-id> (all 2 delta operation(s)
    failed validation); delta operations were not applied
[MENTAL_MODELS] Refresh failed (delta_ops_failed); ... Previous content preserved in DB.
```

This is the same family as #3820 (top-level array) and #3901 (`blocks` emitted as objects): a model output shape that is valid JSON but off-schema from the op models. #3901's `_coerce_block_texts` absorbs the object-spelling of a **known** field's **value**, but nothing absorbs an **unknown extra field** — so this spelling still wedges deterministically at temperature=0 (thinking-disabled), retry after retry.

No data is lost (refusing to write is correct), but each failed round pays a full `reflect` before the delta call, then discards both.

### Steps to Reproduce

1. Configure a bank with an OpenAI-compatible provider (see LLM Provider below), `deepseek-v4-flash`, `temperature=0`.
2. Create a mental model with `trigger.mode = "delta"` and let it accumulate a populated document.
3. Trigger a refresh. `POST .../mental-models/{id}/dry-run-refresh` reproduces without persisting.

### Expected Behavior

An op with an undocumented extra field is applied with the stray field ignored (or the field is stripped before validation) — not silently dropped, and not allowed to fail the entire refresh.

### Actual Behavior

Every op carrying an unknown field is rejected wholesale; when all ops in a round carry one, the refresh fails with `delta_ops_failed` and retries forever.

### Version

`0.10.0`

### LLM Provider

DeepSeek (deepseek-v4-flash) via `HINDSIGHT_API_LLM_PROVIDER=openai`, `HINDSIGHT_API_LLM_EXTRA_BODY={"thinking":{"type":"disabled"}}`, temperature 0.

ekinnee:
## Summary

Fixes #4443.

When a reflection provider decorates an otherwise-valid delta operation with a field the op schema does not declare, every op model ran with `extra="forbid"`, so a single stray key dropped that entire operation. When *every* op in a refresh round carried such a field, all were dropped, `mode_fallback_reason` was set to `delta_ops_failed`, and the refresh raised `MentalModelRefreshError` instead of writing anything.

Change `_OpBase.model_config` from `extra="forbid"` to `extra="ignore"`. The stray key is now silently absorbed while every genuinely-wrong op still fails validation (missing required field, or a wrong-typed value). This mirrors the existing philosophy in this file where `_coerce_block_texts` absorbs the id-bearing block spelling rather than rejecting it — "predictable model spellings are absorbed in validation rather than left to fail".

## Files

- `hindsight-api-slim/hindsight_api/engine/reflect/delta_ops.py` — relax the op base's `extra` handling
- `hindsight-api-slim/tests/test_delta_operation_parse.py` — two regression tests:
  - a hallucinated extra field no longer fails the batch
  - a genuinely wrong-typed value still raises
- `hindsight-api-slim/hindsight_api/engine/providers/gemini_llm.py` + `tests/test_structured_output_unions.py` — comment-only: two docstrings still claimed "every delta operation model sets `extra=forbid`"; they now point at the `DeltaOperationList` container, which is what still carries it (schema behavior unchanged)

## Verification

- `tests/test_delta_operation_parse.py` + `tests/test_structured_output_unions.py`: **39 passed**
  - includes the Gemini schema guard `test_request_carries_no_field_the_backend_rejects` (unchanged — switching to `extra="ignore"` still emits no `additionalProperties`, so the schema path is untouched)
- Broadened regression net across `test_structured_doc.py`, `test_delta_editorial_fusion.py`, `test_delta_oversized_replacement.py`, `test_delta_prompt_schema_parity.py`, `test_reflect_document_answer.py`, `test_structured_delta_prompt_budget.py`: **203 passed, 1 skipped** (5 fixture-setup errors are environmental — missing `local-ml` extra + live DB, unrelated to this change, zero `delta_ops.py` frames in any traceback)
- `ruff check` clean, `ruff format --check` clean on all touched files


nicoloboschi:
fixed by #4601

PR #4489, opened 2026-09-17
feat(llm): inject x-opencode-session for any provider targeting opencode.ai

### LINKED ISSUE #4071: bug(opencode-go): requests omit required x-opencode-session header
### Bug Description

Hindsight's built-in `opencode-go` provider sends requests through the Python `AsyncOpenAI` client without the `x-opencode-session` header required by OpenCode Go.

OpenCode Go's documentation asks third-party clients to include this header so requests from one conversation can keep stable routing and prompt-cache affinity:
https://opencode.ai/docs/go#where-can-i-use-it

OpenCode Go has also emailed affected users that some requests are missing the header and that requests without it may start erroring on **09/06**. The reported user agent is **Python OpenAI client**, which matches the client Hindsight uses for this provider.

The provider source currently only forwards `default_headers` when an operator explicitly configures them; it does not generate or inject `x-opencode-session` for `opencode-go`:
https://github.com/vectorize-io/hindsight/blob/v0.9.2/hindsight-api-slim/hindsight_api/engine/providers/openai_compatible_llm.py#L676-L693

I also searched the `opencode-go` provider implementation in v0.9.1, v0.9.2, and current `main`; none contains `x-opencode-session`.

### Steps to Reproduce

1. Configure Hindsight with:
   ```yaml
   HINDSIGHT_API_LLM_PROVIDER: opencode-go
   HINDSIGHT_API_LLM_MODEL: deepseek-v4-flash
   HINDSIGHT_API_LLM_BASE_URL: https://opencode.ai/zen/go/v1
   ```
2. Do not set `HINDSIGHT_API_LLM_DEFAULT_HEADERS`.
3. Run an operation that invokes the LLM, such as retain, reflect, or consolidation.
4. Inspect the constructed `AsyncOpenAI.default_headers` or the outbound request.
5. The client identifies itself as `AsyncOpenAI/Python`, but `x-opencode-session` is absent.

Minimal equivalent inspection:

```python
from hindsight_api.engine.providers.openai_compatible_llm import OpenAICompatibleLLM

llm = OpenAICompatibleLLM(
    provider="opencode-go",
    api_key="dummy",
    base_url="https://opencode.ai/zen/go/v1",
    model="deepseek-v4-flash",
)
print(llm._client.default_headers)
```

### Expected Behavior

When `provider="opencode-go"`, Hindsight should automatically send:

```http
x-opencode-session: <stable-opaque-id-per-logical-conversation>
```

The value should remain stable across retries and all calls belonging to the same logical retain/reflect/consolidation operation or tool loop. Ideally it should not expose a raw bank ID or other user data.

Tests should verify that:

- the header is present for `opencode-go`;
- it is stable across retries and calls in one logical operation;
- separate logical operations can receive separate IDs;
- manually supplied headers have clearly defined precedence.

### Actual Behavior

No `x-opencode-session` header is sent unless the operator manually supplies a static `default_headers` value.

A static deployment-wide workaround is not equivalent to a stable ID per conversation: it pins unrelated Hindsight traffic to the same affinity key and does not model Hindsight's logical operations correctly.

Once OpenCode Go begins enforcing the requirement, Hindsight's primary OpenCode Go calls may return HTTP errors. Deployments with a fallback LLM may continue only after failover; deployments without one may fail retain, reflect, or consolidation operations.

### Version

Observed on Docker image `ghcr.io/vectorize-io/hindsight:0.9.1` with `openai` Python package 2.24.0. Source inspection indicates the same behavior in Hindsight v0.9.2 and current `main`.

### LLM Provider

Other — OpenCode Go (`opencode-go`)

davidneyra2804:
Fixes #4071 (the protocol-requirement half — the chat/completions path already had this; the Responses and Anthropic paths did not).

## Related PRs

- **#4489 (this PR)** — inject `x-opencode-session` for any provider targeting `opencode.ai`. Without it, the backend rejects requests with HTTP 400 `MissingSessionID`.
- **#4497** — downgrade `tool_choice` to `auto` when the host is `opencode.ai`. Without it, the reflect agent's `tool_choice="required"` fails with HTTP 400 `only "auto" is supported`.
- **#4498** — read either `messages` or `input` when deriving the session id. Without it, the start-up `verify_connection` probe (which has no trace context) still fails with `MissingSessionID`, even after #4489 lands.

```yaml
```


davidneyra2804:
plz check to use custom opencode-go responses llms (like muse spark 1.3)

provider=opencode-go dont support muse spark...

davidneyra2804:
Note: this PR is one of three that together make the opencode-go Responses endpoint work for  deployments. The other two are #4497 (tool_choice downgrade) and #4498 (read `input` key for session-id derivation). End-to-end, all three are needed for a Docker config like `provider=openai-responses` + `base_url=https://opencode.ai/zen/go/v1` + `model=muse-spark-1.3-contributor` to run cleanly — verified against a live deployment. Each PR's body now cross-references the others.

davidneyra2804:
Adding a production-use update: I am currently running Hindsight in production with the combined changes from #4489, #4497, and #4498, using `muse-spark-1.3-contributor`, and it is working as expected in my deployment.

The concrete use case is Muse Spark through OpenCode Go's **Responses API (`/v1/responses`)**, which requires the `openai-responses` provider for this setup, rather than the Chat Completions path:

```env
HINDSIGHT_API_LLM_PROVIDER=openai-responses
HINDSIGHT_API_LLM_BASE_URL=https://opencode.ai/zen/go/v1
HINDSIGHT_API_LLM_MODEL=muse-spark-1.3-contributor
```

This PR supplies the required `x-opencode-session` header based on the destination host, so requests using `provider=openai-responses` also receive it. #4497 handles the endpoint's tool-choice restriction, and #4498 makes session-ID derivation work with the Responses API's `input` payload, including the startup verification request without trace context.

Please consider these three changes together when reviewing support for this configuration: all three are needed for the working setup I am using in production. Thank you for reviewing!

nicoloboschi:
Reviewed and pushed one commit of review fixes to this branch (`1dffba8`):

- `apply_opencode_session` took a `provider` argument it never read, kept "for back-compat" for callers that don't exist — all six call sites are in-tree. Removed it, along with the `base_url=None` default no caller used.
- Dropped the redundant `bool(hostname) and` guard (`_host_matches("", ...)` is already False).
- Reworded the docstring paragraph that described the removed argument so it records what the check used to be and why it moved to the host.
- Trimmed the four call-site comment blocks that restated the docstring.
- Tests: hoisted the function-local `MagicMock`/`AnthropicLLM` imports, parametrised `None` alongside `""` instead of `base_url or None`, renamed the test that still said "regardless_of_provider_name".

The core change is right: `opencode-go` sets `https://opencode.ai/zen/go/v1` itself in `OpenAICompatibleLLM.__init__`, so moving the gate from the provider name to the host loses no existing coverage and picks up the Responses and Messages paths.

Note, not changed here: on the Responses path `cache_affinity_id` reads `request["messages"]`, which those params never carry (`input`), so an *untraced* Responses call still derives no id and sends no header. Traced calls — every retain/reflect/consolidation run — are fine. That's #4498's scope, so leaving it there.

Running the full suite on `ci/pr-4489-verify` since fork PRs skip `test-api` and the LLM jobs.

PR #4494, opened 2026-09-18
fix(recall): order the temporal arm by temporal_score before fusion

rocke2020:
We company start to use hindsight as self-hosted memory. thanks very much for hindsight! 

## Problem

Recall's cross-fact-type aggregation sorts the concatenated temporal results before handing them to RRF as the temporal arm. The sort read `combined_score` — a field of `ScoredResult`, which only exists **after** fusion. The list at that point holds `RetrievalResult` objects, so `hasattr(r, "combined_score")` was always `False`, every sort key fell back to `0`, and the sort was a stable no-op.

Consequence: concatenation order (fact-type bucket order) survived into fusion, and since RRF consumes list *order*, not raw scores, a memory's temporal rank was decided by which fact-type bucket it came from — not by its temporal score. The score was computed and then ignored for ordering.

## Fix

`hindsight-api-slim/hindsight_api/engine/memory_engine.py` (~line 8083): sort by `temporal_score` (the `RetrievalResult` field that is the temporal arm's own ranking signal; `None` sorts last). One-line behavior change.

## Verification

- **RED before fix**: `tests/test_temporal_arm_ordering.py` — drives `recall_async` with deliberately mis-ordered per-fact-type temporal results and asserts on the list handed to `reciprocal_rank_fusion`. Failed with the old key (fusion received concatenation order `[low, high]`).
- **GREEN after fix**: both tests pass; the fusion input is score-ordered.
- Regression sweep `pytest -k "temporal or recall or fusion or rerank"`: 602 passed. 8 failures, all pre-existing and environment-caused, **not** from this diff:
  - 4 visible `hs_llm_core` quality tests fail on an invalid Gemini judge API key (`400 API key not valid`, retried 3×) — credential issue in the local env.
  - `test_recall_pipeline_toggles.py::test_bank_override_disables_the_cross_encoder` — verified to fail identically on clean `origin/main` with the fix stashed.
- `./scripts/hooks/lint.sh`: all lints pass.


rocke2020:
btw, i make a comparsion and finally use hindsight, https://github.com/rocke2020/open-agent-memory-benchmark

PR #4497, opened 2026-09-18
fix(llm): downgrade tool_choice for opencode-go /v1/responses

davidneyra2804:
Builds on #4489 — host-based `x-opencode-session` injection. Fixes the second half of the opencode-go protocol gap: the `tool_choice` value the reflect agent sends (`required`) is rejected by the backend.

## Related PRs

- **#4489** — inject `x-opencode-session` for any provider targeting `opencode.ai`. Without it, the backend rejects requests with HTTP 400 `MissingSessionID`.
- **#4497 (this PR)** — downgrade `tool_choice` to `auto` when the host is `opencode.ai`. Without it, the reflect agent's `tool_choice="required"` fails with HTTP 400 `only "auto" is supported`.
- **#4498** — read either `messages` or `input` when deriving the session id. Without it, the start-up `verify_connection` probe (which has no trace context) still fails with `MissingSessionID`, even after #4489 lands.

```yaml
```


davidneyra2804:
Note: this PR is one of three that together make the opencode-go Responses endpoint work for `provider=openai-responses` deployments. The other two are #4489 (x-opencode-session injection) and #4498 (read `input` key for session-id derivation). End-to-end, all three are needed for a Docker config like `provider=openai-responses` + `base_url=https://opencode.ai/zen/go/v1` + `model=muse-spark-1.3-contributor` to run cleanly — verified against a live deployment.

hindsight-api-slim/hindsight_api/engine/providers/openai_responses_llm.py
```diff
@@ -567,6 +590,13 @@ async def call_with_tools(
         else:
             request_tool_choice = tool_choice.mode.value
 
+        # opencode-go's /v1/responses endpoint rejects every tool_choice value
+        # except the default. Downgrade to "auto" (omit the field) — same
+        # tradeoff the chat/completions path takes via
+        # OpenAICompatibleLLM._drops_tool_choice_required for the same backend.
+        if self._drops_tool_choice_required() and tool_choice.mode is not LLMToolChoiceMode.AUTO:
```
ebarkhordar, line 602: Reading this at `2769ade5`, the downgrade on line 597 catches `NONE` as well as `REQUIRED`, and those two do not degrade the same way.

It fires for any mode that is not `AUTO` and sets `request_tool_choice = None`, so the field is omitted. `responses_tools` is left alone, and line 603 still sends the full tool list. For `REQUIRED` that is the best-effort trade your docstring describes. For `NONE` it inverts the caller: "do not call a tool" reaches the backend as "here are the tools, your choice".

The two providers that also cannot put `none` on the wire enforce it by removing the tools instead:

```
github_copilot_llm.py:711  if tool_choice.mode is LLMToolChoiceMode.NONE:
                               selected_tools = []
claude_code_llm.py:572     allowed_tool_names = []; mcp_servers_config = {}
```

`NAMED` already survives the downgrade here for the same reason: line 586 filters `responses_tools` to the single named tool before line 597 drops the field, so the model stays constrained. The same move covers `NONE`, clearing `responses_tools` when the downgrade fires on that mode.

This is latent rather than live, and worth saying so: `LLM_TOOL_CHOICE_NONE` is defined at `llm_interface.py:57` and I found no call site for it anywhere in the package, so nothing sends `NONE` today. I read this rather than ran it, and I have no opencode.ai backend to check what it does with a tools list and no `tool_choice`.

davidneyra2804: Hi bro! Contact me via PM and I’ll send you an OpenCode API key for testing.

I made these changes because previously I couldn’t use Muse Spark with Hindsight through OpenCode. When using the custom configuration, I kept getting an `opencode-session not found` error.

With these changes, I can now use Muse Spark as the LLM for Hindsight through OpenCode.ai. It’s a much cheaper LLM option for processing medium/large memory banks in Hindsight.

Now everything works like a charm! I’m really happy with the result. Right now, I’m running my own fork in Docker :D

ebarkhordar: Thanks, no key needed here. The point is about which branch runs before the request goes out, not what opencode.ai does with it.

It moved to line 602 in `7893ffc9` and still holds: `NONE` takes the same downgrade as `REQUIRED`, `request_tool_choice` is dropped, and line 608 still sends the whole tool list. `NAMED` survives the downgrade only because line 591 narrows `responses_tools` first. Clearing `responses_tools` when the downgrade fires on `NONE` is the same move.

It stays latent, so it does not block the PR. I checked the package again at `7893ffc9` and nothing produces that mode: `llm_interface.py:57` defines `LLM_TOOL_CHOICE_NONE`, and the only other references are the three providers that handle it.

davidneyra2804: i send 3 pr , but 3 works as one (to make opencode.ai work with endpoints/llms uses responses endpoint)

davidneyra2804:
Adding a production-use update: I am currently using `muse-spark-1.3-contributor` in production with the combined changes from #4489, #4497, and #4498, and the setup is working as expected.

This model uses OpenCode Go's Responses endpoint in my setup (`provider=openai-responses`, `base_url=https://opencode.ai/zen/go/v1`). This PR addresses a specific blocker in the reflect agent's tool-calling path: the endpoint rejects `tool_choice="required"` with HTTP 400 (`only "auto" is supported`). Omitting the unsupported choice lets the endpoint use its default `auto` behavior. This allows the request to proceed, with tool selection left to the model.

The other two changes are also necessary for this configuration: #4489 adds the required session header for requests targeting `opencode.ai`, and #4498 derives the session ID from the Responses API's `input` when no trace context is available.

My production confirmation applies to the three changes used together. Please review this as part of that complete Responses API support path. Thank you!

davidneyra2804:
Thanks again for the careful review. I understand the `tool_choice=NONE` edge case: it is a valid improvement to make the downgrade remove the tools as well. As you noted, it is currently latent because Hindsight has no call site that produces `NONE`, so it does not affect this integration's current behavior. I am happy to prepare that small follow-up if the maintainers would prefer it before merging.

Could you please take another look at this PR and its companion #4498? Together with merged PR #4489, they complete support for OpenCode Go's Responses API with `muse-spark-1.3-contributor`.

This is a real production use case for me: I am running the three changes together in my deployment, and Muse Spark is working reliably for Hindsight memory operations. It provides a lower-cost option for processing medium and large memory banks while meeting the behavior required by my workload.

If a narrower implementation or a documented configuration example would be preferable, I am happy to adapt the remaining PRs. I would appreciate guidance on whether #4497 and #4498 can be reviewed as the remaining pieces of the complete Responses API support path.

ebarkhordar:
Agreed, and I would not hold this PR for it.

I re-parsed the tree at `7893ffc9` with ast over 1790 files, looking for `LLMToolChoiceMode.NONE` and `LLM_TOOL_CHOICE_NONE`. Five references: `llm_interface.py:57` is the definition, and `claude_code_llm.py:572`, `gemini_llm.py:921` and `github_copilot_llm.py:711` all handle the mode. Nothing in the tree produces it. So the gap is latent, it does not affect this PR's behaviour, and a separate follow-up is the right size for it.

On merge sequencing and on the rest of the set, the maintainers are better placed than me. I only looked at the `tool_choice` downgrade here.

PR #4498, opened 2026-09-18
fix(llm): read Responses-API input key when deriving session id

davidneyra2804:
Builds on #4489 (host-based `x-opencode-session` injection) and #4497 (tool_choice downgrade). Fixes the missing piece that made the start-up verification probe still fail with `MissingSessionID` against `opencode-go`'s Responses endpoint.

## Related PRs

- **#4489** — inject `x-opencode-session` for any provider targeting `opencode.ai`. Without it, the backend rejects requests with HTTP 400 `MissingSessionID`.
- **#4497** — downgrade `tool_choice` to `auto` when the host is `opencode.ai`. Without it, the reflect agent's `tool_choice="required"` fails with HTTP 400 `only "auto" is supported`.
- **#4498 (this PR)** — read either `messages` or `input` when deriving the session id. Without it, the start-up `verify_connection` probe (which has no trace context) still fails with `MissingSessionID`, even after #4489 lands.

```yaml
```


davidneyra2804:
Note: this PR is one of three that together make the opencode-go Responses endpoint work for `provider=openai-responses` deployments. The other two are #4489 (x-opencode-session injection) and #4497 (tool_choice downgrade). End-to-end, all three are needed for a Docker config like `provider=openai-responses` + `base_url=https://opencode.ai/zen/go/v1` + `model=muse-spark-1.3-contributor` to run cleanly — verified against a live deployment.

davidneyra2804:
Adding a production-use update: the combined changes from #4489, #4497, and #4498 are already running in my production deployment with `muse-spark-1.3-contributor` and are working as expected.

This PR closes the startup-verification gap for that setup. With `provider=openai-responses` and `base_url=https://opencode.ai/zen/go/v1`, requests carry `input` rather than `messages`. The startup `verify_connection` probe has no trace context, so session-ID derivation must also read `input`; otherwise, the required header is still missing and the backend returns `MissingSessionID`, even with #4489 applied.

Together, #4489 enables host-based session-header injection, #4497 handles the Responses endpoint's tool-choice restriction, and this PR covers session-ID derivation for the Responses payload. All three are needed for the Muse Spark configuration I am using in production.

I wanted to make the practical motivation explicit: these changes enable a production use case that depends on the Responses API rather than Chat Completions. Thank you for reviewing the three PRs together!

Issue #2115, opened 2026-06-10
Quickstart `pip install hindsight-all -U` silently installs 3-month-old 0.4.17 on Intel macOS (local-ml deps lack x86_64 mac wheels)


bionicbutterfly13:
### Bug Description

On Intel (x86_64) macOS, the README quickstart `pip install hindsight-all -U` succeeds but silently resolves to `hindsight-all==0.4.17` (released 2026-03-10) instead of the current release. Every release after 0.4.17 (0.4.18 through current 0.8.1) depends on `hindsight-api-slim[all]`, whose extras require two packages with no x86_64 macOS wheels:

- `mlx>=0.31.0` — publishes arm64 macOS + manylinux wheels only (the existing `sys_platform != 'win32'` marker doesn't help Intel Macs)
- `torch>=2.6.0` (CVE floor) — torch dropped Intel-mac wheels after 2.2.x

pip backtracks through ~15 releases without any error and lands on the last resolvable version. The user gets a 3-month-old install with no signal that anything went wrong.

### Steps to Reproduce

1. Intel Mac (macOS 15.7.8, Python 3.11.15): `pip install hindsight-all -U`
2. Watch pip backtrack through every release ≥ 0.4.18
3. `pip show hindsight-all` → 0.4.17

### Expected Behavior

Either the non-portable local-ML block is gated so unsupported platforms resolve the current release without it (consistent with the existing win32 gate on mlx), or installation fails loudly with a pointer to `hindsight-all-slim` (+ `local-onnx` since 0.8.x) as the supported path for x86_64. Anything but a silent stale install from the documented happy path.

### Actual Behavior

Silent resolution backtrack to 0.4.17; no error, no warning.

### Version

0.4.18–0.8.1 (verified against 0.8.1 metadata)

### Environment

macOS 15.7.8 (Intel x86_64, i7-9750H), Python 3.11.15. LLM provider: N/A (install-time).

---

Aware of `hindsight-all-slim` — the issue is that the *default documented path* degrades silently, not that a workaround is unavailable. Happy to PR the marker/docs change if you confirm which direction you prefer.

nicoloboschi:
Thanks for the precise report — confirmed, and your diagnosis is exactly right.

[...]

PR #4500, opened 2026-09-18
fix(packaging): publish per-platform mlx wheel metadata so hindsight-all resolves on glibc < 2.35 Linux

### LINKED ISSUE #4499: pip install hindsight-all silently backtracks to 0.4.17 on glibc < 2.35 Linux (local-ml deps publish no compatible wheels, no sdist)
### Bug Description

On Linux hosts with **glibc < 2.35** (e.g. RHEL/Alma/Rocky 8, Ubuntu 20.04, Debian 11, CentOS 8), the README quickstart `pip install hindsight-all` never installs a current release: it silently backtracks to **hindsight-all 0.4.17** (2026-03-10), the last release that doesn't depend on `hindsight-api-slim[all]`.

This is the Linux sibling of #2115 (same silent-backtrack failure mode on Intel macOS), with a sharper edge: the backtracked set includes **`torch-2.5.1`, below the `torch>=2.6.0` floor that current releases pin explicitly as a remote-code-execution CVE fix** — a months-old dependency set including known-vulnerable versions, installed with no error and no warning.

`uv` (which does not backtrack by default) surfaces the underlying reason instead:

```console
$ uv pip install "hindsight-all==0.10.0"
  × No solution found when resolving dependencies:
  ╰─▶ Because mlx{sys_platform != 'win32'}>=0.31.0 has no wheels with a
      matching platform tag (e.g., `manylinux_2_32_x86_64`) and
      hindsight-api-slim[all]==0.10.0 depends on mlx{sys_platform != 'win32'}>=0.31.0,
      we can conclude that hindsight-api-slim[all]==0.10.0 cannot be used.
    hint: Wheels are available for `mlx` (v0.32.2) on the following platforms:
      `manylinux_2_35_aarch64`, `manylinux_2_35_x86_64`, `macosx_14_0_arm64`, ...
```

**Root cause:** in `hindsight-api-slim`'s `local-ml` extra (pulled in by `[all]`), `mlx`/`mlx-lm` are gated with `sys_platform != 'win32'`. But mlx publishes Linux wheels **only for manylinux_2_35 (glibc ≥ 2.35)** and ships **no sdist** — so on glibc 2.28–2.34 the requirement is unsatisfiable, `[all]`/`[local-ml]` cannot resolve at all, and pip backtracks the metapackage to a pre-slim release. The supported-platforms docs currently mark Linux bare-metal as fully supported with no glibc caveat.

### Steps to Reproduce

On a glibc 2.28 host (`ldd --version` → `2.28`), Python 3.12:

```console
$ pip install --dry-run --only-binary :all: hindsight-all
...
Would install ... hindsight-all-0.4.17 hindsight-api-0.1.5 ...
... torch-2.5.1 ... sentence-transformers-3.2.1 transformers-4.45.2 ...
```

Current release is 0.10.0, and `torch 2.5.1 < 2.6.0` (the pinned CVE floor) — pip backtracks through every release since 0.4.17 with no diagnostic.

### Expected Behavior

`pip install hindsight-all` on glibc 2.28–2.34 Linux installs the **current** release. The `mlx`-based Apple-Silicon local-generation path was never usable there (no wheels exist and never have been), so the rest of the bundle — torch-backed embeddings/reranker included — should resolve without it.

### Actual Behavior

pip silently backtracks to hindsight-all 0.4.17 with a stale dependency set that includes `torch-2.5.1`, below the project's own `torch>=2.6.0` CVE floor. uv fails outright. Nothing in the docs mentions a glibc requirement.

### Version

hindsight-all current release (0.10.0) affected. Reproduced with pip 26.2.1 / uv 0.10.8, Python 3.12, glibc 2.28 (CentOS 8-based).

Happy to send a PR implementing per-platform wheel metadata (same pure-Python wheel, three tags, `mlx` metadata only where mlx actually ships wheels) if that direction sounds good.

richard950825-sys:
### Problem (#4499)

`hindsight-api-slim`'s `local-ml` extra (pulled in by `[all]`) gates `mlx`/`mlx-lm` with `sys_platform != 'win32'`. mlx publishes Linux wheels **only for manylinux_2_35 (glibc ≥ 2.35)** and ships no sdist, so on glibc 2.28–2.34 Linux the extras are unresolvable. Consequences:

- `pip install hindsight-all` **silently backtracks** to 0.4.17 with a stale dependency set that includes `torch-2.5.1`, below the `torch>=2.6.0` RCE-CVE floor (details in #4499; same failure mode as #2115, Linux edition);
- `uv` hard-fails with "no wheels with a matching platform tag";
- the docs' supported-platforms grid claims Linux bare-metal is fully supported, with no glibc caveat.

As established in #2115, a **marker can't fix this** — markers cannot express "glibc ≥ 2.35". But wheel **tags** can: `manylinux_2_35` tags are only selected on glibc ≥ 2.35, where mlx does have wheels.

### Fix

Ship the same pure-Python wheel with **per-platform metadata** (the standard multi-wheel mechanism pip/uv already use for compiled packages — only the metadata differs, the code is identical):

| Wheel | `mlx` requirement metadata | Selected on |
|---|---|---|
| `py3-none-any` | `sys_platform == 'darwin' and platform_machine == 'arm64'` | everywhere else: old-glibc Linux, musl, Intel mac, Windows |
| `py3-none-manylinux_2_35_x86_64` | `sys_platform == 'linux'` | Linux x86_64 with glibc ≥ 2.35 |
| `py3-none-manylinux_2_35_aarch64` | `sys_platform == 'linux'` | Linux arm64 with glibc ≥ 2.35 |

pip/uv prefer the most specific compatible tag, and the `manylinux_2_35` wheels are only tag-compatible where glibc ≥ 2.35 — i.e. exactly where mlx ships Linux wheels. Tag selection and metadata availability line up platform-for-platform.

Changes:

1. **`hindsight-api-slim/pyproject.toml`** — gate `mlx`/`mlx-lm` to `darwin and arm64` in the source-of-truth metadata (this is also what the sdist and locally-built wheels carry), with a comment explaining why.
2. **`scripts/build-api-slim-wheels.sh`** (new) — builds the sdist + base wheel, then a linux-metadata wheel retagged to `manylinux_2_35_x86_64` / `manylinux_2_35_aarch64` via `wheel tags`.
3. **`.github/workflows/release.yml`** — the `Build hindsight-api-slim` step calls the script; the publish step uploads all three wheels unchanged (`gh-action-pypi-publish` picks up everything in `dist/`).
4. **`hindsight-docs/docs/developer/installation.md`** — note the glibc ≥ 2.35 requirement of the `mlx` dependency in the Linux row (mirrors the Intel-mac note added in #2129).

`uv lock` is unaffected: the workspace pyproject still declares `mlx` (for darwin/arm64), so lock inputs don't change. The release workflow's LICENSE-in-wheel check still passes — all three wheels have identical file contents.

### Verified compatibility matrix

Reproduced and verified against PyPI metadata and real resolution runs (pip 26.2.1, uv 0.10.8, Python 3.12):

| Platform | Before | After | Verified via |
|---|---|---|---|
| Linux glibc 2.28 (CentOS 8, x86_64) | ❌ `hindsight-all` backtracks to 0.4.17 | ✅ resolves 0.10.0, without `mlx` | real host: `uv pip install --dry-run "hindsight-all==0.6.1"` with locally built wheels — 232 packages, no `mlx` |
| Linux glibc ≥ 2.35 (x86_64) | ✅ `mlx` installs | ✅ unchanged, `mlx` kept | `uv ... --python-platform x86_64-manylinux_2_35` → `+ mlx==0.32.2`, `+ mlx-lm==0.31.3`; pip selects the `manylinux_2_35_x86_64` wheel |
| Windows x86_64 | ✅ no `mlx` | ✅ unchanged, no `mlx` | `uv ... --python-platform x86_64-pc-windows-msvc` |
| macOS arm64 | ✅ `mlx` installs (macOS ≥ 14 required by mlx itself) | ✅ unchanged — metadata marker evaluates true, exactly the documented "Apple Silicon local inference" intent | metadata inspection; mlx publishes `macosx_14_0_arm64` wheels |
| Linux musl / Intel mac | ❌ blocked (by `torch`, not `mlx`) | unchanged — not fixed by this PR, not made worse | — |

To reproduce the resolution checks with a locally built set:

```bash
./scripts/build-api-slim-wheels.sh hindsight-api-slim hindsight-api-slim/dist
uv pip install --dry-run --find-links hindsight-api-slim/dist "hindsight-all==<version>"
uv pip install --dry-run --find-links hindsight-api-slim/dist \
  --python-platform x86_64-manylinux_2_35 "hindsight-all==<version>"
```

### Scope / what this does *not* change

- **No capability is removed for any currently-working platform.** macOS arm64 keeps mlx; glibc ≥ 2.35 Linux keeps mlx; Windows stays mlx-free (as today).
- Intel macs and musl remain blocked by **torch** (pre-existing, same as today; Intel macs are already documented "⚠️ slim only" since #2129). This PR neither fixes nor worsens that.
- Dev workflows syncing from **source** on Linux with glibc ≥ 2.35 (`uv sync --all-extras` in the monorepo) build the `any` wheel and therefore no longer pull `mlx` into the dev env; contributors exercising the mlx path there can `uv pip install mlx mlx-lm` manually (mlx-Linux requires glibc ≥ 2.35 regardless).
- On glibc 2.28–2.34, `[all]` now resolves **without** `mlx` — that capability never existed there (no wheels, no sdist). Use hosted/Ollama/llama.cpp providers for local generation, as the docs note says.

Closes #4499


nicoloboschi:
fixed by #4520

PR #4503, opened 2026-09-18
fix(storage): reclaim a renamed bank's files when the bank is deleted

### LINKED ISSUE #4502: Deleting a bank renamed with rename-bank leaves its stored files in the object store
`rename-bank` moves every column called `bank_id`. An object key is not one of those: `attachments.storage_key` and `documents.file_storage_key` spell the bank id inside the key string. After a rename the rows carry the new id and their keys still carry the old one.

The sweep at the end of `delete_bank` then misses them. It deletes `bank_storage_prefix(bank_id)`, which is the new prefix, plus any row whose key is `NOT LIKE 'tenants/%'`, which is there for the pre-tenant layout. A renamed bank's keys start with `tenants/` and sit under the old id, so they match neither and the bytes stay after the bank is gone. On S3 or GCS that is content a delete was meant to remove, still billed.

Reproduced at `bcca388` against postgres 18 + pgvector, one PNG attachment retained into `old_id`:

```
key = attachment_storage_key(old_id, "doc", png)   # tenants/public/banks/<old_id>/documents/doc/attachments/sha256-...
await _run_rename_bank(db_url, "public", old_id, new_id, dry_run=False)
await mem.delete_bank(new_id, request_context=rc)
await mem._file_storage.retrieve(key)              # returns the bytes
```

The rename reports `{'attachments': 1, 'banks': 1, 'chunks': 1, 'documents': 1, 'entities': 12, 'memory_links': 190, 'memory_units': 10}`, and `attachments.storage_key` comes back unchanged. The same script with the rename step removed raises `FileNotFoundError` on that last line, so the sweep itself works.

One thing I expected and did not get: the file stays readable. `GET /v1/default/files/download/<key>` still returns 200 after the rename, because the route parses the bank id out of the key and the old id resolves. So this is a leak rather than lost data.

I have a PR that widens the `legacy_files` predicate from `NOT LIKE 'tenants/%'` to "not under this bank's current prefix", which covers the pre-tenant keys and the renamed ones in one condition. If you want `rename-bank` to rewrite the keys instead, say so and I will do that.

ebarkhordar:
Fixes #4502.

`rename-bank` moves every column called `bank_id`, and an object key is not one of those: the id is spelled inside `attachments.storage_key` and `documents.file_storage_key`. The sweep at the end of `delete_bank` looks under `bank_storage_prefix(bank_id)`, which is the new prefix, so a renamed bank's bytes outlive the bank.

The `legacy_files` query already exists for exactly this shape, keys that sit outside the swept prefix, but its predicate names one cause (`NOT LIKE 'tenants/%'`, the pre-tenant layout) instead of the condition. Widened it to "not under this bank's current prefix", which catches both in one condition. `SUBSTR` rather than `LIKE` because the prefix percent-encodes each segment, so it can carry `%` of its own.

Test goes next to `test_deleting_the_bank_deletes_blobs_stored_under_the_tenantless_layout`, which is the same question for the other cause. It fails on main.

One thing worth knowing if you are deciding where this belongs: the files stay readable after a rename. `GET /v1/default/files/download/<key>` returns 200, because the route parses the bank id out of the key and the old id still resolves. So the renamed bank leaks storage but does not lose anything, and fixing the sweep is enough for that. If you would rather `rename-bank` rewrite the keys and copy the objects across, that is a different change and I can do it instead.


koriyoshi2041 (approved):
Verified at exact head `17f5009686976600f5bc8c0d00e5cc89c52b3b40` on macOS/arm64 with PostgreSQL:

`uv run --group dev pytest tests/test_attachment_lifecycle.py::test_deleting_a_renamed_bank_reclaims_blobs_keyed_on_the_old_id -q -n 0`

Result: **1 passed**. I also checked the prefix predicate against `bank_storage_prefix`: the helper includes the encoded bank-id boundary and trailing slash, so the `SUBSTR(..., LENGTH($2)) <> $2` comparison avoids both SQL `LIKE` wildcard semantics and adjacent-id prefix collisions. The row-key capture before deletion is the right boundary here.

ebarkhordar:
I think these two are complementary rather than competing, and the split is clean: #4573 stops keys and rows diverging, #4503 reclaims the ones that already have.

Reading the diff, `_move_bank_files` has a single call site inside `_run_rename_bank` and nothing backfills existing rows, so it re-keys a bank at the moment it is renamed. A bank renamed before it ships keeps rows spelling the old id, and `delete_bank`'s sweep still looks only under the current prefix, so deleting one of those leaves its bytes behind. Renaming it a second time afterwards does not rescue it either, since the select matches `starts_with(key, old_prefix)` and those keys spell an id older than that.

That case is the whole of #4503: it widens the existing `legacy_files` predicate from `NOT LIKE 'tenants/%'` to "not under this bank's current prefix", so the sweep catches the pre-tenant keys it already caught plus anything a rename left behind. One condition, and it needs nothing from the CLI.

We both touch `test_attachment_lifecycle.py`, so whichever lands second needs a rebase. If you would rather carry one change here, say so and I will close #4503.

nicoloboschi:
Thanks for this — the analysis was right and the repro ran verbatim on my side (rename, delete, and the PNG was still readable at its old key).

We went a different way for the root cause, in #4573 (merged): `rename-bank` now re-keys the bank's stored files as part of the rename — copy each object to the key under the new prefix, repoint the row, then clear the old prefix (which also takes the export/import archives left under the old id). That keeps the invariant the sweep relies on: every key a bank's rows name sits under that bank's current prefix.

That said, I don't think this PR is redundant, and I'd like to keep it. Your change stops the sweep *depending* on that invariant, which still matters for the seam the move can leave open: an interrupted rename (crash between two rows), or a row my move skips because its bytes were already gone. Those rows keep an old-prefix key and would leak on delete. One predicate, real coverage.

Two things before it can merge:

1. **Rebase on main** — #4573 touched the same `legacy_files` query and added a test right where yours goes.
2. **The test goes vacuous after the rebase.** With the rename now moving the bytes, `test_deleting_a_renamed_bank_reclaims_blobs_keyed_on_the_old_id` passes with or without your predicate change — the old key is already cleared at rename time, so it no longer exercises the thing it's testing. Forge the state instead, the way `test_deleting_the_bank_deletes_blobs_stored_under_the_tenantless_layout` right above it does: retain, `UPDATE attachments SET storage_key = <a key under a stale prefix>`, store the bytes there, then delete the bank and assert the object is gone.

One correction to the PR description, worth knowing because it changes the framing: `GET /v1/default/files/download/<key>` returning 200 after a rename is not the old id "still resolving" — #4465 dropped `create_if_missing=False` from `retrieve_bank_file`, so that route now **auto-creates** the missing bank and then serves the file. That's a separate bug (a GET creating banks); I'll open an issue for it rather than let it justify the leak being benign.

One thing the sweep-only fix would have gotten wrong on its own, for the record: with keys left under the old id, deleting by row key can remove bytes another bank now shares — rename `a`→`b`, create a new bank `a`, retain the same image, and the content-addressed attachment key is identical. Once the rename moves the keys that can't arise, so the two changes together are safe in a way this one alone isn't.

ebarkhordar:
Rebased, and you were right that the test went vacuous: with the rename moving the bytes, the old one passes at current HEAD whether or not the predicate change is there. Replaced it with a forged-state test on the pattern of `test_deleting_the_bank_deletes_blobs_stored_under_the_tenantless_layout`: retain, `UPDATE attachments SET storage_key` to a key under a stale bank prefix, store the bytes there, delete the bank, assert the object is gone. That key sits under `tenants/`, which is exactly the case the old `NOT LIKE 'tenants/%'` arm reads as current and the sweep never looks at.

What I checked here is the predicate rather than the test. Against PostgreSQL, with the prefix taken from `bank_storage_prefix`, the old arm returns only the tenantless key and leaves the stale-prefix one behind; the new one returns both and still excludes a key under the bank's current prefix. The same run covers why it is `SUBSTR` and not `LIKE`: the prefix percent-encodes each segment, so a bank id with a space gives `my%20bank/`, and read as a LIKE pattern that matches a foreign bank's key and quietly treats it as ours. I could not get the suite itself to run locally, pg0 returns no URI in my container, so the new test's own red and green is CI's word rather than mine.

Thanks for the `#4465` correction. I had read that 200 as the old id still resolving, and a GET that auto-creates the bank explains it better; I have taken the claim out of the description.

koriyoshi2041 (approved):
Re-reviewed at exact head  after the rebase.\n\nThe replacement regression now forges the post-interruption invariant directly: its row names an object under a different tenant-scoped bank prefix, so the old  predicate misses it and the post-commit current-prefix sweep cannot reach it. The new fixed-length prefix comparison selects that row while continuing to exclude current-prefix keys, without interpreting percent-encoded  as a SQL wildcard.\n\n passed. I also collected the exact focused test with the embedded-db extra; local execution skipped at the optional embeddings fixture because this checkout does not have the local ML stack. Hosted , system blackbox, Python 3.11–3.14 builds, extension tenant suites, generated-file and OpenAPI checks are green at this head.

koriyoshi2041 (approved):
Correction to the formatting-stripped identifiers in my immediately preceding approval: exact head d0258c79c5af00adfd1bd601f7dd4ebfc49fe716; old predicate NOT LIKE tenants/%; local hygiene command git diff --check upstream/main...HEAD; hosted test-dev is green. The substantive approval and local-skip boundary are unchanged.

PR #4514, opened 2026-09-18
fix(reflect): stop an unbounded sibling field from fanning out split-synthesis chunks (#4495)

### LINKED ISSUE #4495: split_context_history copies oversized sibling output fields into every split chunk (100+ chunks, ~20x token duplication)
**Version:** 0.10.0 — Docker image `ghcr.io/vectorize-io/hindsight@sha256:3edcb6165cefdeaa6721dd0fce43cfd13b7a9c346ce0d2c5f4b4bf7bc3c8ac0b` (built 2026-09-14, repo revision `5d46f9c8c8eb4fb96f549aa63abe1191b82a7840`).

## Summary

When a reflect run forces final synthesis over over-budget tool history, `split_context_history` can produce **10–30x more chunks than the entry size warrants** (observed: 142 / 275 / 310 chunks for entry totals of 69k / 132k / 123k tokens; expected ≈ 7–14). The same raw text is duplicated into nearly every chunk, inflating a refresh from a few dozen LLM calls to several hundred and into the millions of input tokens.

## Root cause

`tool_recall` (`api/hindsight_api/engine/reflect/tools.py`) returns

```python
{"query": ..., "memories": [...], "chunks": {k: {chunk_text, chunk_index, truncated}}}
```

where the `"chunks"` dict (raw source-chunk texts) is **not bounded by `max_tokens`**. With a modest result set this field alone reaches ~20–30k tokens.

In the item-split branch of `split_context_history` (`api/hindsight_api/engine/reflect/prompts.py`):

```python
items = output[split_key]          # "memories"
piece: list = []
for item in items:
    candidate = {**entry, "output": {**output, split_key: piece + [item]}}
    if piece and count_prompt_tokens(_render_history_block(candidate)) > budget:
        ...
```

`{**output, split_key: ...}` carries **every sibling key** of the tool output — including the ~30k-token `"chunks"` dict — into every candidate/partial/cut:

1. Every candidate render is ≥ the `"chunks"` size, i.e. always > `budget` (`0.8 * max_context_tokens`, e.g. 9830 at 12288).
2. The item loop therefore closes a chunk after one item, and each item falls into the `single_tokens > budget` branch, becoming a standalone cut chunk.
3. `_cut_entry_to_budget` cuts the serialized *wrapper* — which still contains the full `"chunks"` dict — to ≤ budget, so every chunk re-carries the (re-truncated) raw text.

Net effect: N memory items → N chunks, each containing the same raw text — N-fold duplication.

## Observed impact

Local deployment: bank with ~500 facts, `HINDSIGHT_API_REFLECT_MAX_CONTEXT_TOKENS=12288`, llama.cpp serving 4 slots of ~24.5k tokens.

- 3 parallel `refresh_mental_model` operations → 142 + 275 + 310 = **727 split chunks** (entry totals 69,496 / 131,955 / 123,572 tokens — ≈ 1 chunk per 400–490 tokens instead of per 9830).
- 92 map calls in 10 seconds already carried **1.09M input tokens**; the full run would have been ~727 × ~12k ≈ **8.7M input tokens**, serialized through 4 slots → hours, and the operation dies on timeout.
- Mental models stayed in the "Generating content…" (stale) state indefinitely, degrading every subsequent reflect that reads them.

## Minimal reproduction

Any reflect/refresh where a tool result contains a non-splittable sibling key whose rendered size exceeds `budget` — `recall` with `include_chunks=True` on a bank with a reasonable history is enough:

```python
from hindsight_api.engine.reflect.prompts import split_context_history

small = {"id": "m%d" % i, "text": "t" * 300, "mentioned_at": "2026-09-17T00:00:00+00:00"}
entry = {
    "tool": "recall",
    "input": {"tool": "recall", "query": "q"},
    "output": {
        "query": "q",
        "memories": [dict(small, id="m%d" % i) for i in range(30)],            # ~100 tokens each
        "chunks": {f"c{i}": {"chunk_text": "x" * 3000, "chunk_index": i, "truncated": False}
                   for i in range(20)},                                          # ~20k tokens
    },
}
chunks = split_context_history([entry], 12288)
print(len(chunks))   # 30 — instead of ~3
```

## Suggested fix

Two complementary parts:

1. **Defensive (split side).** In the item-split branch, drop sibling keys that cannot fit the budget on their own before building candidates/partials/cuts:

```python
slim_output = {
    k: v for k, v in output.items()
    if k != split_key
    and count_prompt_tokens(json.dumps(v, indent=2, default=str, ensure_ascii=False)) <= budget
}
# then use {**slim_output, split_key: ...} at all five spread sites
```

2. **Source side.** Make `tool_recall` budget the `"chunks"` payload (e.g. stop appending chunk texts once cumulative chunk text reaches a fraction of `max_tokens`, or add a chunk-text budget parameter) so the tool result respects the token contract its schema advertises.

Option 1 alone changes what the claims extractor sees (raw chunk text of that result no longer reaches the map calls); option 2 alone leaves the split fragile against any future tool that returns a large non-splittable field. Both together is best.

## Local mitigation applied (for reference)

We run a local image with fix 1 applied. The same 3 refreshes went from 727 chunks / timeout to 5 + 4 + 8 chunks / ~8.5 min each, all finishing `content_written`.

rishcx:
split_context_history's item-split loop built every candidate/partial/cut block as {**entry, "output": {**output, split_key: ...}} — spreading ALL sibling keys of `output`, not just the split_key list, into every single block. tool_recall's "chunks" field (raw source-chunk text, deliberately not bounded by the tool's own max_tokens) rides as exactly such a sibling alongside "memories". When "chunks" alone already exceeded the per-chunk budget, every candidate render exceeded budget regardless of item count, degenerating the packer to one output chunk per memory item — each one re-carrying a full copy of the same ~20-30k-token chunks blob.

Production impact (from the issue): 3 parallel refresh_mental_model runs produced 142/275/310 chunks (entries totaling 69k/132k/123k tokens) where ~7-14 were expected, pushing input tokens into the millions and timing out the operation; mental models stayed stuck in "Generating content...".

Two fixes, both applied:

1. Split side (prompts.py): before building candidates/partials/cuts, compute slim_output once per over-budget entry — drop any sibling key (not split_key) whose own serialized size exceeds _MAX_SPLIT_SIBLING_FRACTION (0.5) of the per-chunk budget, and use slim_output at all five spread sites instead of output. A dropped sibling is recorded under slim_output["omitted_fields"] rather than silently vanishing. Chose 0.5x budget over the issue's suggested plain "<= budget" filter: a sibling sitting near-but-under the full budget still leaves no room for a second item, so the very next item hits the single-item cut branch too — reproducing the same per-item fan-out one chunk at a time. The stricter fraction leaves real packing room. Extracted _dump_output_json() to share the json.dumps+fallback logic across _render_history_block, _cut_entry_to_budget, and this filter.

2. Source side (tools.py): added _chunks_within_budget(), which bounds tool_recall's "chunks" payload by walking it in given order and stopping once the RENDERED JSON (not raw chunk text) would exceed max_chunk_tokens. The engine already budgets raw chunk-text accumulation (memory_engine.py's chunk loop), but that (a) doesn't account for the JSON rendering overhead the prompt actually pays for, and (b) never runs at all for a store-answered recall path — so tool_recall is now self-bounding on its own output regardless of which recall path produced the chunks. Reuses the existing max_chunk_tokens parameter rather than adding a new config knob.

Deliberately out of scope: tool_search_observations's "source_facts" is structurally the same shape (an unbounded-by-the-tool dict sibling next to a splittable list) but isn't what #4495 reported and is already defensively covered by fix 1; adding fix 2's pattern there too would be scope creep without evidence it's needed.

Testing:
- Reproduced the issue's own repro pattern before making any change, confirmed the fix collapses it back to the no-sibling baseline.
- Added 3 regression tests to test_reflect_split_synthesis.py; verified 2 of them fail against the pre-fix code (temporarily reverted the fix, ran them, restored) to confirm they're real regression tests and not tautologies.
- Added 3 tests to test_recall_config.py (TestToolRecallChunkBudget) covering the oversized/bounded/untouched cases for fix 2.
- Updated two comments/docstrings (tools.py, test_reflect_tool_result_ fields.py) that claimed "chunks is not trimmed" — now true only for FIELD trimming, not size.
- Full reflect/recall/consolidation test sweep: 704 passed, 0 failed (332 errors are pre-existing missing pg0-embedded dependency in this environment, unrelated to this change — confirmed by isolating one).
- ruff check, ruff format, ty check all clean on hindsight-api-slim.

Fixes vectorize-io/hindsight#4495


rishcx:
@nicoloboschi lemme know if this solution is upto the requirements

rishcx:
any comment @nicoloboschi ?
please lemme know if any!

PR #4528, opened 2026-09-19
fix(coding-agents): single-flight the cold daemon start

Tonited:
## What

`ensureDaemon` now keeps one in-flight cold start per API URL. Callers that arrive while a start is
underway join it instead of spawning a competitor.

## Why

Every ensure point re-checks `/health` and, finding the port dead, spawns its own `daemon-start.js`.
The persistent-plugin hosts fire several of them in the same tick — a SessionStart per session plus
a Stop per finished turn — and each of them observes the same dead port before the first daemon
answers. So one cold start becomes N starters.

Those starters then serialise on the profile's file lock (`hindsight_embed/profile_manager.py`),
where the losers wait out `DEFAULT_LOCK_TIMEOUT = 300 s` and fail. That is not the harmless race the
old comment on `startDaemonDetached` described, so that comment is corrected here too.

Observed on a machine where the local daemon could not come up (no HuggingFace route, empty model
cache — an environment problem, not this bug):

- 5 `daemon_start_spawned` diag events inside 300 ms
- 36 additional 300 s `ProfileLockTimeout` failures across 25 distinct holder PIDs
- 365 spawns across 51 sessions

## How

- A module-level `coldStarts: Map<apiUrl, Promise<void>>`. The check-then-set holds no `await`, so
  the guard cannot be won twice.
- A follower awaits the shared attempt through `waitBounded`, capped by its **own** `waitMs`, so a
  40 s Stop-hook budget can never push a SessionStart hook past its 30 s timeout.
- Whatever is left of the follower's budget once the shared attempt settles goes into
  `waitForHealth`, so joining never leaves a caller with less patience than it would have had on its
  own.
- The entry is dropped as soon as the attempt settles, and only by the owner, so a later turn that
  still finds the port dead starts a fresh attempt rather than inheriting a spent one.
- `waitBounded` clears its timer on the way out. An abandoned `setTimeout` would otherwise hold a
  short-lived hook process open for up to the full budget.

## Tests

In `src/core/daemon.test.ts`:

- `collapses a concurrent burst onto a single start` — 5 concurrent `ensureDaemon` calls must call
  `spawnFn` exactly once. Fails without the fix (5 spawns).
- `spends what is left of its own budget after joining` — a follower with budget left keeps probing
  after the shared attempt is gone. Fails without the refill (2 probes instead of 3+).
- `starts again once the previous attempt has settled` — the guard does not go sticky.

Verified locally: `npm test` (74 files, 1070 passed), `npx tsc --noEmit`, `prettier --check`,
`npm run build`.

## Out of scope

- **Backoff / circuit breaker after a failed start.** Every turn still re-attempts. Worth having
  (365 spawns for 51 sessions above), but it is a separate behavioural change with its own policy
  question — how long to give up for — so it is not folded into this one.
- **`startDaemonDetached`'s two empty `catch` blocks.** Left alone deliberately: they are there so a
  failed spawn cannot break the caller, and deciding what they should report is a diagnostics
  change, not part of de-duplication.
- **Cross-process racing.** A file-lock-level fix would also cover separate hook processes (the
  Claude Code / Codex shape). This change addresses the in-process burst the persistent-plugin hosts
  produce; the profile lock still serialises the rest, as it does today.


koriyoshi2041 (commented):
The map currently tracks the health-wait attempt of a caller, not the detached daemon start itself. If a real cold start outlives the leader waitMs (the PR description says this is routine), waitForHealth settles, the finally block deletes the entry, and the next hook can spawn another starter while the first detached child is still loading. The test named starts again once the previous attempt has settled actually codifies that duplicate window with waitMs zero. Could the regression include a child that remains in-flight after the first caller budget expires, then invoke a later ensure before health comes up and assert there is still only one spawn? That likely needs the guard lifetime tied to observed health or child completion, or to a bounded startup lease longer than an individual hook budget, while each caller still keeps its own wait budget.

PR #4535, opened 2026-09-19
feat(coding-agents): add TraeCode (TRAE CN) harness

cq-guojia:
## Summary

Adds TraeCode — the agent inside TRAE CN — as a full per-prompt HOOK host (registry name `traecode`), following the same shape as the Claude Code / Factory Droid / ZCode harnesses:

- **SessionStart** injects the recall banner; **UserPromptSubmit** journals the user turn and injects recall context; **Stop** retains the turn via the journal. TraeCode keeps sessions in an encrypted local DB or the cloud — there is no transcript file — so the journal *is* the whole write-back path (same pattern as ZCode).
- TraeCode speaks Claude Code's hook protocol; the event map lives under the top-level `hooks` key of its `hooks.json` and the host-written `version` field is preserved.

### Installer targets (all verified against a live install)

| Target | Location |
|---|---|
| Hooks | `hooks.json` under the edition dot-dir |
| MCP | `mcpServers.hindsight` in the Electron userData dir (`~/Library/Application Support/Trae CN/User/mcp.json` on macOS) — a `~/.trae-cn/mcp.json` is never read by the host |
| Skill | companion skill under the same dot-dir's `skills/` |
| Sandbox | one `filesystem.readWrite` rule for `~/.hindsight` in `sandbox.json` |

The sandbox rule matters: TraeCode executes every hook inside its sandbox, and without the rule the hooks fail **silently** (exit 0, zero effect) because the sandbox blocks `~/.hindsight` writes. Network is allowed by default. With the rule, the full pipeline (recall injection, retain, skill refresh) works inside the sandbox — no need to ask users to switch hooks to unsandboxed execution.

### Edition probing

TRAE ships two editions with different brand roots — CN: `~/.trae-cn` / `Trae CN`, international: `~/.trae` / `Trae` (docs.trae.ai). Every path resolves by probing for the edition dir that exists and defaulting to the CN names. The probing lives in one place (`core/skill-dirs.ts`) so the installer and the session-start skill sync stay in agreement. Note: verified against a live CN install; the international paths follow the official docs but are not live-verified.

### Manual steps left to the user (UI state the installer cannot write)

1. Enable the hooks under TraeCode Settings > Hooks (hooks require a security confirmation toggle).
2. Add the hindsight MCP server to the agent under Settings > MCP.

## Test plan

- [x] `installer.test.ts` / `registry.test.ts` / `hook-lifecycle.test.ts` / `skill-sync` coverage: 14 new installer cases (hooks merge, userData mcp path, sandbox rule seeding/idempotency/uninstall, edition probe, skill install), 1080 tests pass on the merged branch
- [x] Live end-to-end on macOS against a real TraeCode install: hooks fire in sandbox mode, journal records both prompt and Stop turns, retain lands in the configured bank, MCP server loads and is usable by the agent
- [x] `install` is idempotent and `uninstall` restores all four targets (including removing our sandbox rule while keeping foreign ones)

Related: #4387 (WorkBuddy + CodeBuddy harnesses, independent — no shared code beyond the existing family patterns).


koriyoshi2041 (changes requested):
The latest MCP revisions leave all three TraeCode guides describing the old install contract. `README.md`, `hindsight-docs/docs-integrations/coding-agents.md`, and the mirrored skill reference still say install writes `User/mcp.json`, refuses a foreign entry there, and requires the user to add the MCP server in Settings. At this head, `install()` instead removes only our stale user-level entry, SessionStart writes `<repo>/.trae/mcp.json`, and the installer handles the workspace-MCP gate (prompt or `--enable-workspace-mcp`).

That mismatch sends a new user to configure the path the implementation deliberately treats as invalid: its home-directory cwd either disables `optInOnly` or resolves the wrong bank. Could you update the three mirrored TraeCode sections to describe the per-repo registration, workspace gate/restart behavior, second workspace-storage sandbox rule, and the remaining Hooks UI step?

I checked exact head `5dc0738c4f`; the focused installer/MCP/lifecycle/registry suite is green (233 tests), so this looks documentation-only.

cq-guojia:
Thanks for the careful read — you're right, the three guides were still describing the retired user-level contract. Fixed in 34b050200: the TraeCode sections in `README.md`, `hindsight-docs/docs-integrations/coding-agents.md` and the mirrored skill reference (re-generated from the README as the single source) now describe the per-repo registration — why user-level entries are wrong under a home cwd, `install()` migrating only our stale entry away, and the SessionStart hook maintaining `<repo>/.trae/mcp.json` — plus the `enableWorkspaceMcp` gate with its prompt / `--enable-workspace-mcp` and restart semantics, the second `workspaceStorage` sandbox rule, and the one remaining Hooks UI step. Docs-only change; the focused installer/MCP/lifecycle/registry/docs suite stays green (249 tests).

koriyoshi2041 (approved):
Re-reviewed head 31371871436ba45ba739daeb84009450784bcfee after the documentation update. The three TraeCode guides now match the implemented contract: per-repo .trae/mcp.json ownership, stale user-level entry migration, the enableWorkspaceMcp prompt/flag and restart boundary, both sandbox write rules, and the single remaining Hooks UI step. This resolves my prior documentation blockers.

koriyoshi2041 (changes requested):
The installer pre-seed direction addresses the observed Trae hook sandbox constraint, but the uninstall helper can delete unrelated top-level content. In `removeTraecodeWorkspaceMcpEntry`, after deleting our only server, `Object.keys(record).length === 0` calls `rmSync(file)` without checking the rest of `doc`. For example, `{ "mcpServers": { "hindsight": <ours> }, "inputs": [...] }` is removed wholesale, despite the ownership contract saying the rest of the document is untouched. Please delete the now-empty `mcpServers` key, then unlink only if the entire document is empty; otherwise rewrite the preserved document. A regression with unrelated top-level data and no foreign MCP server would pin this down.

koriyoshi2041 (approved):
Re-reviewed exact head 0b3d7830ea437b17b682c3dd68a5b4fb7331281b. The uninstall path now removes the empty `mcpServers` key, checks the entire document before unlinking, and rewrites unrelated top-level data otherwise. The new installer regression exercises the exact sole-owned-server plus foreign `inputs` counterexample, and the coding-agents check is green. This resolves my ownership blocker.

koriyoshi2041 (approved):
Re-reviewed exact head  after the logo-only authored commit and upstream merge. The two shipped PNG copies are byte-identical, both registries and all three rendered guide references now target , and the obsolete SVG copies are removed. The control-plane/docs builds and focused coding-agents suite are green. The new asset does not alter the previously reviewed installer or lifecycle behavior.

koriyoshi2041 (approved):
Correction to the formatting-stripped identifiers in my immediately preceding approval: exact head `a5439c7c772e`; both registries and all three rendered guide references target `traecode.png`. The substantive approval is unchanged.

Issue #4537, opened 2026-09-19
openclaw: retain silently dropped on OpenClaw 2026.9.5 due to aborted/null serviceAbortController in lazyReinit fallback


Valkster70:
### Bug Description

In OpenClaw 2026.9.5, session lifecycle management cleans up prior plugin service handles by calling service.stop() during session/tool preparation, but does not invoke service.start() on transient or session-scoped registries.

In @vectorize-io/hindsight-openclaw (v0.12.0):
1. When OpenClaw cleans up prior registrations, service.stop() runs:
   ``javascript
   serviceGeneration++;
   serviceAbortController?.abort();
   serviceAbortController = null;
   isInitialized = false;
   ``
2. When the turn runs, efore_prompt_build calls waitForReady(). Because isInitialized is false and initPromise was cleared/settled, it invokes lazyReinit().
3. lazyReinit() recreates client = new HindsightClient(...) and sets isInitialized = true, allowing **recall** to continue functioning.
4. **However, lazyReinit() does not recreate serviceAbortController or start initRetainQueue().**
5. At turn completion, 
unRetain() checks:
   ``javascript
   const retainGeneration = serviceGeneration;
   const retainController = serviceAbortController;
   const retainSignal = retainController?.signal;
   const retainLifecycleIsCurrent = () => retainController !== null &&
       serviceAbortController === retainController &&
       retainGeneration === serviceGeneration &&
       !retainSignal?.aborted;
   if (!retainLifecycleIsCurrent()) return;
   ``
   Because serviceAbortController is 
ull (or was aborted), 
etainLifecycleIsCurrent() returns alse, causing 
unRetain() to silently exit on every turn without saving or queuing memories.

Additionally, in OpenClaw 2026.9.5, the Control UI Memory section calls getActiveMemorySearchManagerCore(), which checks slots.memory ("hindsight-openclaw") expecting pi.registerMemoryCapability({ runtime: ... }). Because hindsight-openclaw does not register a memory capability with a search runtime, the UI displays "memory plugin unavailable" with a warning face (cosmetic, but confusing to users).

### Steps to Reproduce

1. Run OpenClaw gateway on version 2026.9.5 with @vectorize-io/hindsight-openclaw 0.12.0 enabled.
2. Complete an agent conversation turn.
3. Observe gateway logs: waitForReady called before service.start() — attempting lazy initialization fallback.
4. Recall injects memories properly during prompt construction.
[...]

PR #4556, opened 2026-09-19
fix(openclaw): restore retain lifecycle during lazy initialization

r266-tech:
When the external API is lazily initialized without `service.start()`, recall can succeed while `agent_end` silently skips retain because no service abort controller exists. The lazy path also leaves the persistent retry queue unopened.

Give lazy initialization a cancellable service lifecycle, preserving the first recall's controller, and open the retain queue after initialization succeeds. Check the captured generation after asynchronous initialization so a concurrent stop/start cannot publish stale readiness or reopen the queue.

Refs #4537. This addresses the lazy-initialization retention path; it does not change stopped-hook suppression or add the separately requested Control UI memory capability.

Validation: 393 tests passed across all 18 OpenClaw package test files; TypeScript build and scoped Prettier/diff checks passed. New regressions cover retain before service.start and after an explicit lazy recovery following stop, stop during initialization, and failed-retain persistence/timer replay. The first three regression cases fail on unmodified main. Tests use the plugin hooks with mocked transport, not a live OpenClaw gateway.


r266-tech:
I rechecked the current head `6a0a28ee` and the latest CI run. The OpenClaw build, docs build, smoke install, generated-file verification, and change detection checks pass; the integration jobs that do not apply to this scoped change are skipped. The lazy retain lifecycle regression coverage remains present. Please let me know if maintainers want any additional OpenClaw-specific integration run before review.

Tosko4:
Additional downstream confirmation of the failure addressed here: after moving to OpenClaw 2026.9.5 with hindsight-openclaw 0.12.0, recall continued to work while automatic retention stopped. The API stayed healthy and the operation queue was empty, so neither was a reliable indication that new conversations were reaching the backend.

An isolated reproduction against the installed plugin confirms both lazy initialization before `service.start()` and explicit stop followed by lazy recovery: the client becomes usable, but an otherwise eligible `agent_end` produces zero retain calls. Transport was mocked; this is component-level evidence, not a live validation of this PR's head.

The controller and durable-queue restoration in this PR targets that exact gap. The generation checks during asynchronous initialization are important to preserve as well. This has real downstream impact because the failure is silent rather than an observable failed operation.

Thanks for the focused fix and regression coverage. It would be valuable to include this in the next plugin release; a published version would let affected installations recover without carrying a local compatibility patch.

Issue #4557, opened 2026-09-19
Reflect returns HTTP 500 when tool-capable model gives plain text for short query


sg-shag:
## Summary

On a self-hosted Hindsight `0.10.0` instance using an OpenAI-compatible oMLX endpoint, `reflect` returns HTTP 500 when the model responds with ordinary text for a short query, even though the same model emits valid structured tool calls for a meaningful query with the identical tool schema and `tool_choice`.

The current error is:

```text
Reflect requires a tool-calling model, but openai/Qwen3.6-35B-A3B-mxfp8 produced no usable tool call
Response: 'Hello! How can I help you today?'
```

## Environment

- Hindsight: `0.10.0`
- Model: `Qwen3.6-35B-A3B-mxfp8`
- oMLX: `0.6.4`
- Endpoint: `http://msm4-wifi.lan:8018/v1`
- Hindsight operation: `reflect`

## Independent reproduction

The following exact tool schema and settings were sent directly to oMLX for both prompts:

```json
{
  "model": "Qwen3.6-35B-A3B-mxfp8",
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "search_memory",
        "description": "Search memory for relevant information.",
        "parameters": {
          "type": "object",
          "properties": {"query": {"type": "string"}},
          "required": ["query"]
        }
      }
    }
  ],
  "tool_choice": "required",
  "max_tokens": 512
}
```

### Prompt: `hi`

```json
{
  "finish_reason": "stop",
  "message": {
    "content": "Hello! How can I help you today?",
    "reasoning_content": null,
    "tool_calls": null
  }
}
```

### Prompt: `What did Sergey say about the Plavdom pilot?`

```json
{
  "finish_reason": "tool_calls",
  "message": {
    "content": null,
    "tool_calls": [
      {
        "type": "function",
        "function": {
          "name": "search_memory",
          "arguments": "{\"query\": \"Sergey Plavdom pilot\"}"
        }
      }
    ]
  }
}
```

Both requests used the same model, tools, `tool_choice`, and output limit. The only changed input was the user prompt. Both completed in under one second.

## Impact

Hindsight reflect treats a normal text response during the mandatory retrieval phase as an upstream server error. This makes short or socially conversational prompts fail with HTTP 500 instead of returning a controlled result or a clearer compatibility response.

This also makes model readiness probes incomplete: a model may pass meaningful retrieval queries but still fail on short inputs.

## Expected behavior

[...]

PR #4559, opened 2026-09-20
fix(reflect): identify tool-selection failures in compatibility errors

r266-tech:
When an endpoint returns plain text for `hi` but calls tools for another prompt (#4557), Reflect's current compatibility error does not identify the tool choice that failed or the response's finish reason. It points operators toward unsupported function calling even though the failure can be prompt-specific.

Report the initial tool-selection phase, iteration, requested canonical tool choice, and normalized `finish_reason` in `ReflectToolCallError`. Preserve the bounded response preview and add guidance to check the failing prompt on the configured endpoint. The existing HTTP 500/string-detail contract and failure behavior remain: this change diagnoses the failure and does not make text-only responses successful or add retries.

Addresses the diagnostic part of #4557; it does not close the broader compatibility/short-query behavior question. This is independent of #4469's provider wire-format configuration.

Validation:
- 12 new regression cases fail on unmodified main and pass with this change: three response shapes across mental-model, observation, recall, and auto tool selection. They also verify one provider attempt and no retrieval or final-synthesis calls.
- 55 focused existing/new Reflect tests passed (mocked agent, tool normalization, context overflow, and empty-answer guardrails). The final diagnostic wording was rechecked with all 12 new cases passing.
- Changed-file Ruff lint/format, `ty check hindsight_api/engine/reflect/agent.py`, and `git diff --check` pass.
- `./scripts/hooks/lint.sh` was attempted: dependency sync succeeded, but the frontend ESLint check failed because this checkout lacks `@eslint/js`. Unrelated auto-format edits from that script were removed. Full repository lint is not claimed.

Tests used Python 3.11 and mocked provider responses; no live oMLX/Qwen reproduction or end-to-end HTTP test is claimed. AI-assisted implementation and review.


koriyoshi2041 (approved):
The failure boundary stays evidence-first: a text-only first turn still cannot become a successful reflection, while the diagnostic now distinguishes the requested tool choice, normalized finish reason, and prompt-specific compatibility failure. I also checked that later prose after an evidence-producing tool call continues down the existing synthesis path.

Validated exact head `ae39aa0050ff06fa7487d49ca98650c7b91ed315` locally:

- `uv run pytest tests/test_reflect_agent.py::TestReflectAgentMocked -q` — 42 passed
- changed-file Ruff lint and format checks — passed
- `git diff --check` — passed

The 12-case matrix covers each initial choice (`search_mental_models`, `search_observations`, `recall`, and `auto`), response truncation/empty content, and the no-retrieval/no-synthesis negative control.

PR #4565, opened 2026-09-20
fix(clients): enforce export polling timeouts

S0rryHorizon:
The export helpers only checked their completion timeout after a status response and then slept for the full polling interval. A pending request or long interval could exceed the budget; a late `completed` response bypassed the check entirely.

This bounds the completion-polling phase in both Python and TypeScript, including HTTP requests and waits, and cancels/awaits its owned work. TypeScript's two export methods now use the existing shared helper. Submission and archive-download timeout scope is preserved, as are operation failures, transport errors and caller cancellation. Python uses `asyncio.wait` to preserve simultaneous caller cancellation on Python 3.10.

Validation:
- Real loopback HTTP regression tests through both public export methods: 32 Python and 26 TypeScript cases. Original-source reversal reproduces 12 Python and 14 TypeScript failures; an additional deterministic cancellation test rejects the unsafe Python 3.10 `wait_for` variant.
- Python offline client suite: 201 passed on 3.11, 3.13 and 3.14; all 32 new cases also passed on 3.10 and 3.12. The existing 3.12 suite has an unrelated warning-count failure, reproduced unchanged on the baseline.
- TypeScript offline suite: 118 passed on Node 20; CJS/ESM/declaration build and Python wheel/sdist build passed.
- Repository `scripts/hooks/lint.sh` passed. New Python tests pass Ruff; manual wrapper checks retain only the baseline's unrelated import-order/two formatting findings.

Live-server/provider/Oracle/Deno integration jobs were not run locally; they are separate from these offline SDK checks. No server or generated API changes.

AI disclosure: This patch and its tests were agent-authored and independently agent-reviewed under S0rryHorizon's authorization. No human pre-submission review is claimed.


koriyoshi2041 (approved):
Reviewed exact head `ed452f81004340a20eaa70685a99b81730d30ebf`. The polling deadline now covers both an in-flight status request and the inter-poll sleep in both clients, while submission and archive download keep their existing request-timeout scope. I also checked the cancellation ownership boundaries: Python cancels and awaits its polling task without replacing caller cancellation, and TypeScript detaches the caller listener and clears the deadline timer before downloading.

The focused Python loopback suite passes locally (32 tests), including late completion, simultaneous cancellation, nonpositive deadlines, reuse after timeout, and both export surfaces. `git diff --check` is clean; the hosted TypeScript build is green. No blocking issues found.

